PR #25, opened 2017-11-30
Update README.rst

a-aitken:
To avoid a name clash with the builtin "input", changed the variable names in the example as follows:
input -> x
output -> prediction
target -> y


alykhantejani:
Thanks Andy!

Issue #16, opened 2017-11-27
[Feature] validate_every_epoch may take integer


vfdev-5:
Hi, 
A feature proposition to change a little bit the behaviour of `validate_every_epoch` [here](https://github.com/pytorch/ignite/blob/master/ignite/trainer.py#L180) and accept an integer and run validation every k epoch. What do you think about this ?

HTH

alykhantejani:
This behaviour could also be achieved by handlers, as well as for running every x iterations although  I'm not opposed to changing this

alykhantejani:
After discussions in #35 and #27 I think there is some consensus on removing this parameter altogether and providing a convenient handler that calls `validate` and wraps the logic for doing this every x epochs/iterations (see #42). 

Closing this issue for now, let's redirect the conversation to #42

PR #27, opened 2017-12-02
Change validate_every_epoch (bool) -> validate_every_epoch (int)

vfdev-5:
New feature as discussed in #16

### DIFF
--- a/ignite/trainer/trainer.py
+++ b/ignite/trainer/trainer.py
@@ -176,7 +176,7 @@ class Trainer(object):
                           "Training will stop after current iteration is finished")
         self.should_terminate = True
 
-    def run(self, max_epochs=1, validate_every_epoch=True):
+    def run(self, max_epochs=1, validate_every_epoch=1):
         """
         Train the model, evaluate the validation set and update best parameters if the validation loss
         improves.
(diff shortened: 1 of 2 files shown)


alykhantejani:
Thanks, let's also get @fmassa's opinion on this before merging

fmassa:
While I'm ok with this change, I'm thinking about one of the sentences in [The Zen of Python](https://www.python.org/dev/peps/pep-0020/), which reads

> There should be one-- and preferably only one --obvious way to do it.

In this spirit, we would even remove the `validate_every_epoch` argument, and recommend using the `VALIDATION_STARTING` handler to perform the such checks based on epoch number (which would allow the user to perform arbitrarily complex checks given the epoch number).

What do you think?

vfdev-5:
@fmassa I agree to remove `validate_every_epoch`. However `VALIDATION_STARTING` handler is less obvious way to do this.

alykhantejani:
@vfdev-5 @fmassa I don't think `VALIDATION_STARTING` is the right handler to use if we remove `validate_every_epoch`. Instead you would want to use `TRAINING_ITERATION_COMPLETED` or `TRAINING_EPOCH_COMPLETED` as the `VALIDATION_STARTING` event is only triggered when `trainer.validate()` is called

alykhantejani:
After discussions in #35 and #27 I think there is some consensus on removing this parameter altogether and providing a convenient handler that calls `validate` and wraps the logic for doing this every x epochs/iterations (see #42). 

Closing this issue for now, let's redirect the conversation to #42

PR #28, opened 2017-12-02
Update setup.py

vfdev-5:
Prevent crash during `python setup.py install` if `readme.rst` once contains a non ascii character


alykhantejani:
Thanks!

PR #35, opened 2017-12-14
Adds functionality for tracking and logging avg and best loss during training

geoffroeder:
To monitor convergence when training deep learning models, it is helpful to have a log of the best validation or training loss seen so far, as well as the average loss over the all epochs so far.

This PR adds that functionality into `ignite.trainer.Trainer`.

### DIFF
--- a/ignite/trainer/trainer.py
+++ b/ignite/trainer/trainer.py
@@ -5,9 +5,7 @@ import time
 from collections import Iterable
 
 from enum import Enum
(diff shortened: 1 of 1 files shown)


alykhantejani:
Hey @geoffroeder, as discussed offline you can now achieve this functionality through `ignite.handlers.logging` or using the `simple_moving_average` etc. in the `History` object.

PR #36, opened 2017-12-16
Allows to specify the validation period in epochs by setting validate_every

farleylai:
This addresses #16 with a small change to allow for setting the validation period in epochs.
The default is set to 1, working as before with validation every epoch.


farleylai:
Seems like someone else has done the same thing.

PR #43, opened 2017-12-28
introduce Supervised wrapper

### LINKED ISSUE #38: Provide basic update + inference functions
See #37 for discussion on this. We should provide something for simple supervised training

cc @jasonkriss

jasonkriss:
Resolves #38 

@alykhantejani couple things to sort out before this can be merged...

- do you have a strong preference on docstring format? I used google style here to be consistent with pytorch itself
- how do you feel about the "wrapper" name? Is it too generic or is it alright?
- I swapped the order of y, y_pred in the return of `predict` to be consistent with loss functions in pytorch
- I haven't updated the readme or examples but i figured it might be best to do that once we wrap up the other remaining issues. Is that alright with you?

Let me know if you have any other feedback on this.

### DIFF
--- a/ignite/wrappers/supervised.py
+++ b/ignite/wrappers/supervised.py
@@ -1,5 +1,6 @@
 from torch.autograd import Variable
 
+
 class Supervised(object):
     """
     Provides training and validation update functions for standard supervised models.
--- a/ignite/wrappers/supervised.py
+++ b/ignite/wrappers/supervised.py
@@ -50,7 +50,7 @@ class Supervised(object):
         self._model.eval()
         x, y = self._prepare_batch(batch)
         y_pred = self._model(x)
-        return y_pred, y
+        return y_pred.data.cpu(), y.data.cpu()
 
     def _prepare_batch(self, batch):
         x, y = batch
--- a/ignite/wrappers/supervised.py
+++ b/ignite/wrappers/supervised.py
@@ -1,0 +1,58 @@
+from torch.autograd import Variable
+
+class Supervised(object):
+    """
+    Provides training and validation update functions for standard supervised models.
+
+    Args:
+        model (torch.nn.Module): the model to train
+        optimizer (torch.optim.Optimizer): the optimizer to use
+        loss_fn (torch.nn loss function): the loss function to use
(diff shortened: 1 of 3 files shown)


ignite/wrappers/supervised.py
```diff
@@ -0,0 +1,59 @@
...
+            float: the batch loss
+        """
+        self._model.train()
+        self._optimizer.zero_grad()
+        x, y = self._prepare_batch(batch)
+        y_pred = self._model(x)
+        loss = self._loss_fn(y_pred, y)
+        loss.backward()
+        self._optimizer.step()
+        return loss.data[0]
```
alykhantejani: probably want a `.cpu()` here too

ignite/wrappers/supervised.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def predict(self, batch):
+        """
+        Validation update function.
```
alykhantejani: nit: `Validation inference function`

alykhantejani:
Thanks @jasonkriss this looks great. 

On the docstrings, sticking with the pytorch convention is probably best

I'm not a huge fan of `ignite.wrappers` but can't think of anything better (let me think about it a bit more today). 

However, it's just occurred to me that as the `Trainer` will only take the inference and update functions in the constructor if we make the change to pass the dataloader into the `run`/`validate` functions. In this case perhaps its better if we just have a helper function that returns a trainer for supervised training instead of the functions that a user would then supply to a `Trainer` constructor. WDYT?

jasonkriss:
@alykhantejani That's an interesting idea. Hadn't thought of that before. The nice thing about that approach would be that we don't need to introduce this extra "wrapper" abstraction. Would it maybe even make sense to just have it be a `SupervisedTrainer` subclass? Or would you rather just keep it as a function. Either way I think that's a good idea. Let me know and I can make the changes.

alykhantejani:
I'd personally prefer a function. Perhaps we can put this inside `ignite/trainer/trainer.py` so it can just be `from ignite.trainer import create_supervised_trainer` or something like that

jasonkriss:
@alykhantejani I opened a new PR #48 for this

PR #47, opened 2018-01-03
add decorator for attaching event handlers

### LINKED ISSUE #41: Add annotations to attach event handlers
See #37 for details

jasonkriss:
Resolves #41 

There are a few different ways we can do this. If you want to pass arguments through, the decorator needs to accept arguments. So I chose to have a single `on` decorator that takes the event as well as any arguments to pass through. Let me know what you think.


alykhantejani:
Thanks @jasonkriss !

PR #48, opened 2018-01-04
introduce factory function for creating supervised trainers

### LINKED ISSUE #38: Provide basic update + inference functions

jasonkriss:
Resolves #38

See discussion in #43 for more context.

### DIFF
--- a/ignite/trainer/trainer.py
+++ b/ignite/trainer/trainer.py
@@ -5,9 +5,10 @@ import time
 from collections import Iterable
 
 from enum import Enum
+from torch.autograd import Variable
 from ignite.trainer.history import History
 
-__all__ = ["TrainingEvents", "Trainer"]
+__all__ = ["TrainingEvents", "Trainer", "create_supervised"]
 
 
 class TrainingEvents(Enum):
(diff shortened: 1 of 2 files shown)


ignite/trainer/trainer.py
```diff
@@ -212,3 +213,41 @@ def run(self, training_data, max_epochs=1):
...
+    Args:
+        model (torch.nn.Module): the model to train
+        optimizer (torch.optim.Optimizer): the optimizer to use
+        loss_fn (torch.nn loss function): the loss function to use
+        cuda (bool, optional): whether or not to transfer batch to GPU (default: False)
+
+    Returns:
+        Trainer: a trainer instance with supervised update and inference functions
+    """
+    def _prepare_batch(batch):
```
elanmart: Perhaps it would be nice to have a `training` flag here, to use `Variable(..., volatile=True)` (`torch` `0.3`)  or `torch.autograd.no_grad()`  (`torch` `0.4`) during inference?

jasonkriss:
@elanmart great catch. that was oversight. How's that look? We can update it to `no_grad` when  0.4.0 is released.

alykhantejani:
Thanks @jasonkriss!

PR #49, opened 2018-01-05
return relevant history from 'run' and 'validate'

jasonkriss:
This is purely for convenience but I think it is quite common that people will want to access the history right after. I'm already finding this in my own work.

### DIFF
--- a/ignite/trainer/trainer.py
+++ b/ignite/trainer/trainer.py
@@ -160,6 +160,8 @@ class Trainer(object):
 
         self._fire_event(TrainingEvents.VALIDATION_COMPLETED)
 
+        return self.validation_history
+
     def terminate(self):
         """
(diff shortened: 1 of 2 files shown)


fmassa:
Just realized that we might want to wait for #54 to be merged before merging this

jasonkriss:
@fmassa My personal use case is actually a lil different. I have a setup where models get trained to completion (a predefined number of epochs) and then are evaluated/tested once at the very end.

Currently, I do something like this,

```python
```

alykhantejani:
@jasonkriss @fmassa let's try and get #54 merged and then we can re-visit/ update this PR. Sorry for the delay on this

alykhantejani:
Closing this as the API has changed a lot and I think we can revisit this with the new State object being introduced in #80. Please let me know if you think this should be kept open and I can re-open

PR #51, opened 2018-01-10
use more consistent/explicit names

jasonkriss:
This is in response to some valuable feedback from @eifuentes.

There are two goals here:
- more consistent naming
- more semantic naming

Using `validate` throughout leads to some confusion. If someone wants to evaluate their model against multiple datasets, e.g. for training/validation curves, it feels strange to pass a "training"
 dataset through a "validation" function and then compute metrics on a "validation" history. It makes more sense when we use a generic term like "inference" here. That way you can evaluate your model by making inference passes on both "training" and "validation" sets.

This PR introduces "inference" instead of "validation" while also changing `training_history` to `update_history`. It also renames `TrainingEvents` to `Events` since `TrainingEvents` is both redundant and implies that inference events aren't included.

What do you think, @alykhantejani?


eifuentes:
thanks @jasonkriss. to add some more context, the original goal was to use ignite to plot train/validation curves using the create_supervised path.

alykhantejani:
What do you think about creating a separate `Evaluator` so we would have the `Trainer` and the `Evaluator`? Both would be initialized with a function (update or inference) and both expose a `run` function which takes a data loader. They could both expose `Events` which the user can attach to (you can also call `evaluator.run` from within a training handler).

If we go this route then they could both subclass an `Engine` which handles the adding and firing of events

WDYT?

Issue #37, opened 2017-12-18
General Roadmap?


jasonkriss:
Hey guys,

Loving the project so far. I have a handful of extensions, utilities that I'd love to discuss either adding to this project or putting in a separate library. Would it be possible to get some insight into the future directions of ignite? How are decisions being made? I'm happy to start opening PRs, but obviously wanted to discuss them first.

Some of the things I'm curious about:

- adding some common update and inference callables (e.g. a supervised updater that essentially just does what tnt did by default)
- support for multiple validation sets
- a different approach to metrics that doesn't require coordination between the update/inference function and the event handlers
- considering using a "state" based approach (like in tnt) rather than a "history" based approach (in order to minimize memory consumption for use cases like #20)
- allow validation to happen every n iterations instead of being epoch based (similar to what is suggested [here](https://github.com/pytorch/ignite/pull/27#issuecomment-350788169)
- a "callback" abstraction for cases where multiple event handlers need to coordinate and pass state around

If this is better discussed offline over email or something, just let me know.

Thanks a lot!

alykhantejani:
Hi @jasonkriss,

I think it'd be good to discuss this on github to keep it in the open and let others weigh in. To answer some of the questions please see my comments below:

_Would it be possible to get some insight into the future directions of ignite? How are decisions being made?_
 - Decisions are generally being made in the open on github (via issues + PRs) i.e. we welcome feature requests and discussions like this one :)
 - As for the future direction, this also will be driven by user feedback. Me and others have used ignite in its current form to do research for some time now, but by opening it up to a wider audience we are hoping to get more insight into what features others may want to improve the framework altogether.

_adding some common update and inference callables (e.g. a supervised updater that essentially just does what tnt did by default)_
- could you provide some more info on this? In general though, I think some common event handlers would be great, the callables - if they are indeed quite general I think would be a good addition too

_support for multiple validation sets_
- This would indeed be a welcome feature. We currently only support a single data loader for validation and a single callable. I think this could easily be extended in a non-backwards compatible way. You can also achieve this through callbacks, but it's more ugly

_a different approach to metrics that doesn't require coordination between the update/inference function and the event handlers_
- could you elaborate a bit more on the approach. In fact @apaszke had some thoughts around this too. I can maybe write up a small gist with his summary (I think there is one somewhere which I'll have to look up)

_considering using a "state" based approach (like in tnt) rather than a "history" based approach (in order to minimize memory consumption for use cases like #20)_
[...]

PR #52, opened 2018-01-10
attach current batch to trainer

jasonkriss:
This is necessary in order to support rendering input samples during training (hat tip @eifuentes).

We talked about it a bit in #37, but I wonder if we should more seriously consider introducing a `state` dict/object that gets passed to each handler. With this PR, we now have several `current_*` attributes that might make more sense in `state`.

What do you think @alykhantejani?

### DIFF
--- a/ignite/trainer/trainer.py
+++ b/ignite/trainer/trainer.py
@@ -61,6 +61,8 @@ class Trainer(object):
 
         self.training_history = History()
         self.validation_history = History()
+        self.current_batch = None
+        self.current_validation_batch = None
         self.current_iteration = 0
         self.current_validation_iteration = 0
         self.current_epoch = 0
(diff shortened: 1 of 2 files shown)


eifuentes:
:+1:

alykhantejani:
Hey @jasonkriss,

I agree the `current_*` is getting a bit messy, perhaps a state would be nicer. I'd prefer that to adding another current_batch (which we then have to clean up after each iteration). Would you mind editing this PR to add that?

Also, what is the motivation to accessing the batch within the handler?

FYI - I'll be without a laptop for the next 2 weeks so wont be able to merge this. @fmassa - if you get a chance can you review the next version?

alykhantejani:
@jasonkriss sounds good to me. I'm back now, although down with the Flu but will try to address the comments in #54 asap

jasonkriss:
Closing as this is addressed in #80

Issue #55, opened 2018-01-16
Log handlers are not up to date.


vinayakvivek:
functions like `log_training_simple_moving_average` uses `training_data` attribute of `Trainer` which does not exist.
```python
...
iterations_per_epoch = len(trainer.training_data)
...
```

alykhantejani:
Sorry about this, the API is moving quickly. Will get these things fixed this week

alykhantejani:
This is being addressed in #59

alykhantejani:
Sorry for the delay on this. The new API is in now, I will get this fixed ASAP

PR #45, opened 2017-12-29
introduce new validation approach

jasonkriss:
Resolves #39 and resolves #42 

Need to sort out things in #43 before we merge this.

@alykhantejani how do you feel about the usage of a class here for the `Validate` handler? Imo, this is easier to follow than passing the arguments to `Trainer#add_event_handler` and having it forward them along. It's also nice for decomposing the functionality into multiple methods.

PR #59, opened 2018-01-25
Fix problems related to newer API

vfdev-5:
- Fix example mnist with visdom
-- Add validation loss plot
-- Add visdom port as argument
- Fix trainer and logger

PR addresses #55

### DIFF
--- a/examples/mnist_with_visdom.py
+++ b/examples/mnist_with_visdom.py
@@ -159,7 +159,7 @@ def run(batch_size, val_batch_size, epochs, lr, momentum, log_interval, logger, 
     trainer.add_event_handler(TrainingEvents.VALIDATION_COMPLETED, get_plot_validation_loss_handler(vis))
     trainer.add_event_handler(TrainingEvents.VALIDATION_COMPLETED,
                               get_plot_validation_accuracy_handler(vis, val_loader))
-    trainer.add_event_handler(TrainingEvents.VALIDATION_COMPLETED, lambda trainer: trainer.validation_history.clear())
+    
     trainer.run(train_loader, max_epochs=epochs)
 
 
--- a/examples/mnist_with_visdom.py
+++ b/examples/mnist_with_visdom.py
@@ -52,7 +52,25 @@ def get_plot_training_loss_handler(vis, plot_every):
     return plot_training_loss_to_visdom
 
 
-def get_plot_validation_accuracy_handler(vis):
+def get_plot_validation_loss_handler(vis):
+    val_loss_plot_window = vis.line(X=np.array([1]), Y=np.array([np.nan]),
+                                      opts=dict(
+                                          xlabel='# Iterations',
+                                          ylabel='Loss',
+                                          title='Validation Loss')
+                                      )
+
+    def plot_validation_loss_to_visdom(trainer):
+        avg_loss = np.mean([loss for (loss, accuracy) in trainer.validation_history])
+        vis.line(X=np.array([trainer.current_iteration]),
+                 Y=np.array([avg_loss]),
+                 win=val_loss_plot_window,
+                 update='append')
+    return plot_validation_loss_to_visdom
+
+
+
+def get_plot_validation_accuracy_handler(vis, validation_data):
     val_accuracy_plot_window = vis.line(X=np.array([1]), Y=np.array([np.nan]),
                                         opts=dict(
                                             xlabel='# Epochs',
@@ -62,7 +80,7 @@ def get_plot_validation_accuracy_handler(vis):
 
     def plot_val_accuracy_to_visdom(trainer):
         accuracy = sum([accuracy for (loss, accuracy) in trainer.validation_history])
-        accuracy = (accuracy * 100.) / len(trainer.validation_data.dataset)
+        accuracy = (accuracy * 100.) / len(validation_data.dataset)
         vis.line(X=np.array([trainer.current_epoch]),
                  Y=np.array([accuracy]),
                  win=val_accuracy_plot_window,
@@ -70,19 +88,19 @@ def get_plot_validation_accuracy_handler(vis):
     return plot_val_accuracy_to_visdom
 
 
-def get_log_validation_loss_and_accuracy_handler(logger):
+def get_log_validation_loss_and_accuracy_handler(logger, validation_data):
     def log_validation_loss_and_accuracy(trainer):
         avg_loss = np.mean([loss for (loss, accuracy) in trainer.validation_history])
         accuracy = sum([accuracy for (loss, accuracy) in trainer.validation_history])
         logger('\nValidation set: Average loss: {:.4f}, Accuracy: {}/{} ({:.0f}%)\n'.format(
-            avg_loss, accuracy, len(trainer.validation_data.dataset),
-            (accuracy * 100.) / len(trainer.validation_data.dataset)
+            avg_loss, accuracy, len(validation_data.dataset),
+            (accuracy * 100.) / len(validation_data.dataset)
         ))
     return log_validation_loss_and_accuracy
 
 
-def run(batch_size, val_batch_size, epochs, lr, momentum, log_interval, logger):
-    vis = visdom.Visdom()
+def run(batch_size, val_batch_size, epochs, lr, momentum, log_interval, logger, visdom_port):
+    vis = visdom.Visdom(port=visdom_port)
     if not vis.check_connection():
         raise RuntimeError("Visdom server not running. Please run python -m visdom.server")
 
@@ -116,7 +134,13 @@ def run(batch_size, val_batch_size, epochs, lr, momentum, log_interval, logger):
         correct = pred.eq(target.data.view_as(pred)).sum()
         return loss, correct
 
-    trainer = Trainer(train_loader, training_update_function, val_loader, validation_inference_function)
+    def get_validation_inference_handler(validatation_data):
+        def validation_inference(trainer):
+            trainer.validate(validatation_data)
+        return validation_inference
+
+
+    trainer = Trainer(training_update_function, validation_inference_function)
     trainer.add_event_handler(TrainingEvents.TRAINING_ITERATION_COMPLETED,
                               log_training_simple_moving_average,
                               window_size=100,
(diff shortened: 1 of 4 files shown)


examples/mnist_with_visdom.py
```diff
@@ -116,7 +133,12 @@ def validation_inference_function(batch):
         correct = pred.eq(target.data.view_as(pred)).sum()
         return loss, correct
 
-    trainer = Trainer(train_loader, training_update_function, val_loader, validation_inference_function)
+    def get_validation_inference_handler(validatation_data):
```
fmassa: As this could be replaced by the `Validate` handler, I think it's worth adding a comment here saying that in most cases it can be used, but that the user is free to create its own handler

ignite/handlers/logging.py
```diff
@@ -43,10 +41,10 @@ def log_training_weighted_moving_average(trainer, window_size, weights, history_
 def log_validation_weighted_moving_average(trainer, window_size, weights, history_transform=lambda x: x,
                                            should_log=lambda trainer: True, metric_name="", logger=print):
     if should_log(trainer):
-        total_iterations = len(trainer.validation_data)
-        current_iteration = trainer.current_iteration % total_iterations
+        total_iterations = trainer.total_validation_iterations
+        current_iteration = trainer.current_iteration % (total_iterations + 1)
```
fmassa: I'm not sure the `+1` here is necessary. Do you want to make it start from 1 instead of 0? If that's the case, then the `+1` should be outside the modulo operation

ignite/handlers/logging.py
```diff
@@ -17,10 +16,10 @@ def log_training_simple_moving_average(trainer, window_size, history_transform=l
 def log_validation_simple_moving_average(trainer, window_size, history_transform=lambda x: x,
                                          should_log=lambda trainer: True, metric_name="", logger=print):
     if should_log(trainer):
-        total_iterations = len(trainer.validation_data)
-        current_iteration = trainer.current_iteration % total_iterations
+        total_iterations = trainer.total_validation_iterations
+        current_iteration = trainer.current_iteration % (total_iterations + 1)
```
fmassa: Same comment as below about the `+1`, I don't this this is necessary.

ignite/handlers/logging.py
```diff
@@ -56,11 +54,10 @@ def log_validation_weighted_moving_average(trainer, window_size, weights, histor
 def log_training_exponential_moving_average(trainer, window_size, alpha, history_transform=lambda x: x,
                                             should_log=lambda trainer: True, metric_name="", logger=print):
     if should_log(trainer):
-        iterations_per_epoch = len(trainer.training_data)
-        current_iteration = trainer.current_iteration % iterations_per_epoch
+        current_iteration = trainer.current_iteration % (trainer.iterations_per_epoch + 1)
```
fmassa: Same comment about the `+1` here.

ignite/handlers/logging.py
```diff
@@ -69,10 +66,10 @@ def log_training_exponential_moving_average(trainer, window_size, alpha, history
 def log_validation_exponential_moving_average(trainer, window_size, alpha, history_transform=lambda x: x,
                                               should_log=lambda trainer: True, metric_name="", logger=print):
     if should_log(trainer):
-        total_iterations = len(trainer.validation_data)
-        current_iteration = trainer.current_iteration % total_iterations
+        total_iterations = trainer.total_validation_iterations
+        current_iteration = trainer.current_iteration % (total_iterations + 1)
```
fmassa: And here

ignite/trainer/trainer.py
```diff
@@ -133,9 +135,9 @@ def _fire_event(self, event_name):
                 func(self, *args, **kwargs)
 
     def _train_one_epoch(self, training_data):
+        self.iterations_per_epoch = len(training_data)
```
fmassa: I'm unsure about the best way to proceed in here, if it's best to store the whole `training_data` in `self` or only it's length. @alykhantejani what are your thoughts?

ignite/trainer/trainer.py
```diff
@@ -163,6 +165,7 @@ def validate(self, validation_data):
             raise ValueError("Trainer must have a validation_inference_function in order to validate")
 
         self.current_validation_iteration = 0
+        self.total_validation_iterations = len(validation_data)
```
fmassa: Same here.

vfdev-5:
@fmassa no pb for the delay :)

Concerning the modulo and `+1`s, the reason is the following:
- if there is no modifications at all then the loggings reports on epochs or iterations looks like
`0/10` and ends with `9/10`. I think is it more natural to see `1/10` until `10/10`. This is fixed, for example, [here](https://github.com/pytorch/ignite/pull/59/files/383a39a463a95cd31236f65847917fe3ad685b04#diff-69a61daf391623d0d612bb85c22f1f9dR9)
- About `current_iteration = trainer.current_iteration % (total_iterations + 1)`, sometimes in the end of the training, logger shows `current_iteration=0` because of `trainer.current_iteration += 1` ([here](https://github.com/pytorch/ignite/blob/master/ignite/trainer/trainer.py#L147)).

Hope this answers

examples/mnist_with_visdom.py
vfdev-5: Yes, I didn't saw it before. I can change like [here](https://github.com/pytorch/ignite/pull/45#issuecomment-356774652).

ignite/trainer/trainer.py
vfdev-5: `training_data` was stored with the previous API [ref](https://github.com/pytorch/ignite/blob/707d0ecb31db4283f93dc49135b05f30b2515356/ignite/trainer/trainer.py#L68)

examples/mnist_with_visdom.py
```diff
@@ -127,10 +146,15 @@ def validation_inference_function(batch):
...
-    trainer.add_event_handler(TrainingEvents.VALIDATION_COMPLETED, get_log_validation_loss_and_accuracy_handler(logger))
-    trainer.add_event_handler(TrainingEvents.VALIDATION_COMPLETED, get_plot_validation_accuracy_handler(vis))
+    trainer.add_event_handler(TrainingEvents.EPOCH_COMPLETED, Validate(val_loader, epoch_interval=1))
+
+    trainer.add_event_handler(TrainingEvents.VALIDATION_COMPLETED,
+                              get_log_validation_loss_and_accuracy_handler(logger, val_loader))
+    trainer.add_event_handler(TrainingEvents.VALIDATION_COMPLETED, get_plot_validation_loss_handler(vis))
+    trainer.add_event_handler(TrainingEvents.VALIDATION_COMPLETED,
+                              get_plot_validation_accuracy_handler(vis, val_loader))
     trainer.add_event_handler(TrainingEvents.VALIDATION_COMPLETED, lambda trainer: trainer.validation_history.clear())
```
alykhantejani: This line can be removed as the `Validate` handler will now clear this before each run

ignite/handlers/logging.py
alykhantejani: I think @fmassa is correct here

ignite/trainer/trainer.py
alykhantejani: I think it might be better to store the ref to the whole `training_data` or `validation_data` and then in logging we can just call `len` on this. 

Once #54 is merged (after this), this will likely change again so not too fussy right now. I think once #54 is in the `Engine` will have a `dataset` field or something similar.

vfdev-5:
Do not want to insist on the story of `+1` and modulo, but here is how looks like logs before and after changes:

screen shot 2018-01-31 at 00 25 10

screen shot 2018-01-31 at 00 22 14

You can reproduce this with 
```
python mnist_with_visdom.py --log_interval 2
```

ignite/handlers/logging.py
alykhantejani: strange, this should be `trainer.current_validation_iteration` not `trainer.current_iteration`

ignite/handlers/logging.py
```diff
@@ -17,10 +16,10 @@ def log_training_simple_moving_average(trainer, window_size, history_transform=l
 def log_validation_simple_moving_average(trainer, window_size, history_transform=lambda x: x,
                                          should_log=lambda trainer: True, metric_name="", logger=print):
     if should_log(trainer):
-        total_iterations = len(trainer.validation_data)
-        current_iteration = trainer.current_iteration % total_iterations
+        total_iterations = trainer.total_validation_iterations
+        current_iteration = trainer.current_iteration % (total_iterations + 1)
         log_str = "Validation Iteration[{}/{} ({:.2f}%)]\t{}Simple Moving Average: {:.4f}" \
-            .format(current_iteration, total_iterations,
+            .format(current_iteration + 1, total_iterations,
```
alykhantejani: I don't think this +1 is necessary

ignite/handlers/logging.py
alykhantejani: Same for other validation logging functions

ignite/handlers/logging.py
alykhantejani: Same for other validation logging functions

vfdev-5:
@alykhantejani 
> I just had a look and it seems as though that current_epoch is incremented after the EPOCH_COMPLETED is fired, which is a minor inconsistency. @vfdev-5 would you mind changing this order in this PR on this line and also remove the +1 from this test
I'm not sure that we should remove `+1` from [this test](https://github.com/vfdev-5/ignite/blob/fix_new_version/tests/ignite/trainer/test_trainer.py#L204), if I understood correctly, we exchange lines 230 and [231](https://github.com/vfdev-5/ignite/blob/fix_new_version/ignite/trainer/trainer.py#L231), but `current_epoch` still will be incremented.

vfdev-5:
@alykhantejani @fmassa You were right too, actually there was a problem with `current_iteration % (total_iterations + 1)`. This gives something like this :
screen shot 2018-01-31 at 23 47 40
Final iteractions moves to the next epoch. The correct formula is 
```
(current_iteration - 1) % total_iterations + 1
```
I refactored a bit `logging.py` , let me know if it is OK for you.
@alykhantejani I exchanged the lines in the trainer and fixed test script as you asked

alykhantejani:
Hi @vfdev-5,

The refactor of logging looks good. After thinking about it a bit more, incrementing the `current_iteration` and `current_epoch` counters before the `ITERATION_COMPLETE`/`EPOCH_COMPLETE` events makes the handler logic confusing. 

Perhaps we should just revert back to incrementing them both after the event (i.e. [here](https://github.com/vfdev-5/ignite/blob/fix_new_version/ignite/trainer/trainer.py#L149) and [here](https://github.com/vfdev-5/ignite/blob/fix_new_version/ignite/trainer/trainer.py#L230))

I think if we want the counters to log from `1` to `N` instead of `0` to `N-1` (which I think makes sense) we should instead just start the `current_iteration` and `current_epoch` counters from `1` instead of `0`. This would make everything cleaner, and handler logic simpler (i.e. every 2nd iteration is current_iteration % 2). To do this, you will need to change a few tests too.

Instead, why don't I go ahead with #54 and make all of these changes and then you can open a PR on top to refactor the logging?

vfdev-5:
Hi @alykhantejani 
Yes, go ahead with #54 and then if needed I can submit a PR with refactoring etc

alykhantejani:
Closing this as fixed in #70. Please feel free to open a PR to add the port and additional log to the example :)

Issue #12, opened 2017-11-25
Add integration with tensorboard?


alykhantejani:

rooa:
+1 for this. I personally like [tensorboard-pytorch](https://github.com/lanpa/tensorboard-pytorch) very much, but I'm not sure about the policy regarding integration of an external module into ignite.

alykhantejani:
Integration with Visdom was super easy as the API for visdom worked really well with the event handlers. See [the example with MNIST](https://github.com/pytorch/ignite/blob/master/examples/mnist_with_visdom.py) for details.

Ideally tensorboard would be as simple and we can just provide examples

alykhantejani:
Closed as example of how to do this with tensorboard-pytorch has been merged in #60

PR #60, opened 2018-01-25
Add MNIST example with tensorboard

vfdev-5:
PR addresses #12.

Check with commands
```python mnist_with_tensorboard.py --log_dir=tensorboard_logs``` 
and 
```﻿python -m tensorboard.main --logdir=tensorboard_logs/```

Example writes summaries to plot : 
- Training loss
- Validation loss, accuracy
- if has `onnx` should plot model graph

### DIFF
--- a/examples/mnist_with_tensorboard.py
+++ b/examples/mnist_with_tensorboard.py
@@ -1,0 +1,172 @@
+from __future__ import print_function
+from argparse import ArgumentParser
+import logging
+
+from torch.autograd import Variable
+from torch.utils.data import DataLoader
+from torch import nn
+import torch.nn.functional as F
+from torch.optim import SGD
+from torchvision.datasets import MNIST
+from torchvision.transforms import Compose, ToTensor, Normalize
+from tensorboardX import SummaryWriter
+
+from ignite.trainer import Trainer, TrainingEvents
+from ignite.handlers.logging import log_training_simple_moving_average
+import numpy as np
+
+
+class Net(nn.Module):
+    def __init__(self):
+        super(Net, self).__init__()
+        self.conv1 = nn.Conv2d(1, 10, kernel_size=5)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by alykhantejani** [comment]
Thanks @vfdev-5 - I'll just verify by running locally - but then looks good to merge!

File: examples/mnist_with_tensorboard.py [hunk 1]
```diff
@@ -0,0 +1,172 @@
...
+import logging
+
+from torch.autograd import Variable
+from torch.utils.data import DataLoader
+from torch import nn
+import torch.nn.functional as F
+from torch.optim import SGD
+from torchvision.datasets import MNIST
+from torchvision.transforms import Compose, ToTensor, Normalize
+from tensorboardX import SummaryWriter
```
alykhantejani: can we wrap this with a try except and raise an error telling the user they need to install the dependency

examples/mnist_with_tensorboard.py
```diff
@@ -0,0 +1,172 @@
...
+                              batch_size=batch_size, shuffle=True)
+
+    val_loader = DataLoader(MNIST(download=False, root=".", transform=data_transform, train=False),
+                            batch_size=val_batch_size, shuffle=False)
+
+    model = Net()
+    optimizer = SGD(model.parameters(), lr=lr, momentum=momentum)
+
+    try:
+        writer.add_graph_onnx(model)
```
alykhantejani: what does this do?

vfdev-5: This should add a graph of the model using onnx:
See [this](https://github.com/lanpa/tensorboard-pytorch/blob/master/demo_graph.py) and [this](https://github.com/lanpa/tensorboard-pytorch/blob/master/tensorboardX/writer.py#L395)

alykhantejani:
Thanks @vfdev-5!

PR #73, opened 2018-02-11
Add ignore_exceptions flag to Trainer.run

elanmart:
`Ignite` provides a way to define a handler in case of an `Exception`, yet we still have to wrap the `run` method in a `try` / `except` because you can't really do anything inside that handler.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -138,6 +138,7 @@ The following example shows how to register a handler which will save the model 
         raise exception
 
     trainer.add_event_handler(Events.EXCEPTION_RAISED,
+                              save_and_raise,
                               my_model,
                               '/tmp/my_model.pth')
 
--- a/README.rst
+++ b/README.rst
@@ -94,7 +94,7 @@ The :code:`Trainer` emits events during the training loop, which the user can at
 - TRAINING_ITERATION_COMPLETED
 - VALIDATION_ITERATION_STARTED
 - VALIDATION_ITERATION_COMPLETED
-- EXCEPTION_RAISED
+- EXCEPTION_RAISED (see `Note on EXCEPTION_RAISED`_ below)
 
 Users can attach multiple handlers to each of these events, which allows them to control aspects of training such as 
 early stopping, or reducing the learning rate as well as things such as logging or updating external dashboards like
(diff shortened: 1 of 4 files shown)


alykhantejani:
An alternative is that we could pass the exception to the handler and not re-raise it from the trainer, then as a user, if you define a handler, you can choose to re-raise or just swallow it?

i.e. change the line `self._fire_event(Events.EXCEPTION_RAISED)` to `self._fire_event(Events.EXCEPTION_RAISED, e)` and remove the `raise`?

elanmart:
I have thought about it too, but then it's a larger change:
- ignite has to check if there is a handler for this event, raise the exception if there is not. 
- users have to know that this specific handler should accept the exception as first parameter
- we have to modify `_fire_event` to account for this one special case. 

I personally think that it would be nicer to pass the exception to the handler and let users deal with it, let me know what you think.

BTW I think the CI failure is not really related to this change?

alykhantejani:
Strange about the CI, I've restarted it as the logs looked like it couldn't pull a package to even start the build.

alykhantejani:
So currently the user has no way of getting the exception when the event is fired, which is bad and kind of makes the event useless (well not completely, you still know something went wrong). 

I'm inclined to say we should make the changes to pass the exception on to the handler (if it exists).   we can modify the signature of `_fire_event(self, event_name)`, to be   `def _fire_event(self, event_name, passthrough_args)` or something as its private so shouldn't be too much of a pain.

We should also make it clear in the README/docs that this event in particular passes the engine and the exception as the first 2 args? 

WDYT?

elanmart:
Currently you can access the exception via `Engine.exception` (https://github.com/pytorch/ignite/pull/73/files#diff-a89ca55c4b8abc2776859c78889832e4R91) but I agree this is not ideal. 

Let me update the PR tomorrow then.

jasonkriss:
@alykhantejani this is actually a question i've had for a while. What's the reasoning behind firing an event on an exception rather than just having users use the regular old `try`/`except` outside of ignite?

alykhantejani:
@jasonkriss It's because there is a try/except in the inner loop inside `_run_once_on_dataset` as well as one inside `run` - so just to remove boiler plate. But prior to this PR the exception was hidden from the user, so it was still kinda pointless :p

jasonkriss:
@alykhantejani so can we maybe just remove those try/excepts? and then users can just wrap their code in a try/except if they want?

alykhantejani:
@jasonkriss but then if you have a faulty image or something and your inner loop throws an exception, unless we catch it (and potentially the user can choose to ignore it), your whole epoch will end early.

README.rst
```diff
@@ -119,6 +119,28 @@ Event handlers are any callable where the first argument is an instance of the :
...
+.. code-block:: python
+
+    from ignite.engine import Events
+
+    def save_and_raise(engine, exception, model, path):
+        torch.save(model, path)
+        raise exception
+
+    trainer.add_event_handler(Events.EXCEPTION_RAISED,
+                              my_model,
```
alykhantejani: you forgot to add `save_and_raise` here

ignite/trainer.py
```diff
@@ -50,6 +49,11 @@ def run(self, training_data, max_epochs=1):
         max_epochs: int, optional
             max epochs to train for [default=1]
 
+        Notes
```
alykhantejani: This exception attribute is now removed right? There's no more need for it, as it is passed to handlers?

jasonkriss:
@alykhantejani I would actually argue that's the correct behavior. "fail fast" so you know what needs to be fixed. But i guess if we take that approach it makes it difficult/impossible for users who don't want it to end the whole epoch.

alykhantejani:
@jasonkriss yeah unfortunately this is a common scenario when training on huge datasets that aren't stored on your training machine, things like network errors can cause an image to fail to load for example, which you might want to swallow.

alykhantejani:
I mean ideally this is caught and handled in your data loading - but this is a nice to have "bury my head in the sand" feature

jasonkriss:
@alykhantejani 👍 seems reasonable to me. Especially the use case of loading remote datasets. I've run into this myself.

README.rst
elanmart: Thanks, the example doesn't make too much sense without this lol.

alykhantejani:
Thanks @elanmart!

PR #74, opened 2018-02-11
Add a timer to ignite/contrib

elanmart:
This PR adds a `Timer` class to `ignite.contrib`. I use it frequently to measure time it takes to process a single batch / run full validation etc.

It's based on the `timer` that was available in `torch`'s `torchnet`. It can be paused and resumed, and can 
compute average time if `unit=True` is passed.

It has an `attach` methods that can attach the timer to the `Engine`, so that you can save few lines of code (not sure if it's a good idea though).

If you think this is something you'd like to merge, I'd have to add
- [x] docs
- [ ] tests
- [x] example

The PR also sorts and removes unused imports.


alykhantejani:
Hi @elanmart, 

This looks good to me, I think it'll be a useful addition. The only thing I think we should think about is where this (and things like this) live. Right now we have `ignite.handlers` and if we add a `ignite.contrib` which is also a bunch of handlers, what's the distinction?

Maybe we can move what's in `handlers` to `contrib` or vise versa

alykhantejani:
@elanmart shall we move this to `ignite.handlers` for now. Once we sort out metrics and the evaluator we can then clean up handlers and maybe just rename it to contrib?

elanmart:
@alykhantejani yes, sorry for the delay, I'll sort out my PRs today

ignite/engine.py
```diff
@@ -47,8 +47,8 @@ def add_event_handler(self, event_name, handler, *args, **kwargs):
 
         Parameters
         ----------
-        event_name: enum
-            event from ignite.trainer.TrainingEvents to attach the
+        event_name: Events
+            event from ignite.Engine.Events to attach the
```
alykhantejani: Nice catch

ignite/handlers/timing.py
```diff
@@ -0,0 +1,141 @@
...
+
+    Args:
+        unit (bool, optional): if True, then when ``.value()`` method is called, the returned value
+            will be equal to total time measured, divided by the value of internal counter.
+
+    Attributes:
+        total (float): total time elapsed when the Timer was running (in seconds)
+        cnt (int): internal counter, usefull to measure average time, e.g. of processing a single batch.
+            Incremented with the ``.inc_unit()`` method.
+        running (bool): flag indicating if time is measuring time.
```
alykhantejani: if **timer** is measuring time*

ignite/handlers/timing.py
```diff
@@ -0,0 +1,141 @@
...
+            total += self._elapsed()
+
+        if self._unit:
+            denominator = max(self.cnt, 1.)
+        else:
+            denominator = 1.
+
+        return total / denominator
+
+    def inc_unit(self):
```
alykhantejani: should we  rename this to `setp()` to be consistent with the docstring in `attach`?

ignite/handlers/timing.py
```diff
@@ -0,0 +1,141 @@
...
+        ...              pause=Events.ITERATION_COMPLETED,
+        ...              step=Events.ITERATION_COMPLETED)
+    """
+
+    def __init__(self, unit=False):
+        self._unit = unit
+        self._t0 = time.perf_counter()
+
+        self.total = 0.
+        self.cnt = 0.
```
alykhantejani: maybe rename this to `step_count` to be clearer as its a public attribute

ignite/handlers/timing.py
```diff
@@ -0,0 +1,141 @@
...
+        >>> trainer = Trainer(trainig_update_function)
+        >>> timer = Timer(unit=True)
+        >>> timer.attach(trainer,
+        ...              start=Events.EPOCH_STARTED,
+        ...              resume=Events.ITERATION_STARTED
+        ...              pause=Events.ITERATION_COMPLETED,
+        ...              step=Events.ITERATION_COMPLETED)
+    """
+
+    def __init__(self, unit=False):
```
alykhantejani: should we rename `unit` to `mean` or `average`, I know `unit` is used in tnt but I still can't understand why

elanmart:
I agree that the torchnet naming is a little bit unfortunate. I've updated the PR.

alykhantejani:
Thanks @elanmart !

Issue #62, opened 2018-02-04
[Feature Request] Engine checkpointing


pkdogcom:
When training models, I usually would like to save a checkpoint of the model every few epochs or minutes/hours, so that I can resume training from the latest checkpoint when there are errors either in the codes or hardwares, or when I find it necessary to finetune the learning rate without losing the progress. 

In Tensorflow, there are some helper functions in the tf.train package to simplify checkpoint handling, like looking up for latest checkpoints As I was wondering if ignite can also include some of those features so that we don't need to write too many boilerplate codes.

pkdogcom:
It is also very common to keep track of the best checkpoint and it will nice if ignite has out-of-box support for this feature.

alykhantejani:
hi @pkdogcom,

Thanks for the feature request. I agree these both sound like nice features to have. I will add these to my current backlog and start working on them soon if no one else from the community pick them up.

pkdogcom:
And I think it will be also very nice if this feature is implemented with logging/plotting integrated. For example, I've been using Ryzen machine for training recently and it is not stable on Linux and the whole system may hang and require rebooting every now and then. In such case it will be nice for the training to be able to resume from the latest checkpoint, not only with the model and optimizer parameters correctly restored, but also with the state of training related to plotting, such as iteration/epoch counts, Tensorboard events (probably need to remove events logged after the last checkpoint) recovered correctly, so that the training can continue as if it hasn't stopped at all.

elanmart:
@pkdogcom what kind of API do you have in mind? In most of my workflows I simply use 

```python
@engine.on(Events.EPOCH_COMPLETED)
def checkpoint(engine):
   if engine.current_epoch % config.checkpoint_freq == 0:
        torch.save(model, config.storage / 'model.pth')
```

jasonkriss:
@elanmart @alykhantejani I think this opens up a broader question... Do we prefer to keep this library as small as possible or do we prefer to remove as much boilerplate as possible. I think @elanmart's snippet is a great example of how well setup the current API is to make things like checkpointing very easy/explicit. That being said, there may be an argument for making this even easier since it is such a common use case. For example, something like...

```python
engine.add_event_handler(Events.EPOCH_COMPLETED, Checkpoint(config.storage, config.checkpoint_freq))
```

[...]

PR #75, opened 2018-02-11
Add ModelCheckpoint to ignite/contrib

elanmart:
I think this partially solves #62 . Ultimately I agree with @alykhantejani that something more elaborate would be desirable (see https://github.com/pytorch/ignite/issues/62#issuecomment-364078176).

This is a simple handler that can be used to periodically save objects to disk. It supports saving `n` last / best models. Attach with something like
```python
trainer.add_event_handler(Events.ITERATION_COMPLETED,
                          ModelCheckpoint('/tmp/models', 
                                           'gan', 
                                           iteration_interval=1000, n_saved=5),
                          trainer, 
                          model=model)
```

If you'd decide this is something you'd like to merge, I'd add docs, tests, and example.

### DIFF
--- a/ignite/contrib/checkpoint.py
+++ b/ignite/contrib/checkpoint.py
@@ -1,0 +1,118 @@
+import os
+import tempfile
+from collections import namedtuple
+
+import torch
+
+
+class ModelCheckpoint:
+    """ Foo bar
+
+    Parameters
+    ----------
+    dirname : str
+    filename_prefix : str
+    iteration_interval : int, optional
+    epoch_interval : int, optional
+    score_function : Callable, optional
+    n_saved : int, optional
+    atomic : bool, optional
+    require_empty : bool, optional
+    create_dir : bool, optional
+    exist_ok : bool, optional
+    """
+
+    def __init__(self, dirname, filename_prefix,
+                 iteration_interval=None, epoch_interval=None,
+                 score_function=None,
+                 n_saved=1,
+                 atomic=True, require_empty=True,
+                 create_dir=True, exist_ok=True):
+
+        self._dirname            = dirname
+        self._fname_prefix       = filename_prefix
+        self._n_saved            = n_saved
+        self._iteration_interval = iteration_interval
+        self._epoch_interval     = epoch_interval
+        self._score_function     = score_function
+        self._atomic             = atomic
+        self._item_T             = namedtuple('item_T', ('priority', 'data'))
+
+        self._saved = []
+
+        if not (iteration_interval or epoch_interval or score_function):
+            raise ValueError("One of 'iteration_interval', 'epoch_interval' or "
+                             "'score_function' arguments must be provided.")
+
+        if create_dir:
+            os.makedirs(dirname, exist_ok=exist_ok)
+
+        if require_empty:
+            n_matched = [fname
+                         for fname in os.listdir(dirname)
+                         if fname.startswith(self._fname_prefix)]
+
+            if len(n_matched) > 0:
+                raise ValueError("Files prefixed with {} are already present "
+                                 "in the directory {}. If you want to use this "
+                                 "directory anyway, pass require_empty=False. "
+                                 "".format(filename_prefix, dirname))
+
+    def _save(self, obj, path):
+        if not self._atomic:
+            torch.save(obj, path)
+
+        else:
+            tmp = tempfile.NamedTemporaryFile(delete=False)
+
+            try:
+                torch.save(obj, tmp.file)
+
+            except:
+                tmp.close()
+                os.remove(tmp.name)
+                raise
+
+            else:
+                tmp.close()
+                os.rename(tmp.name, path)
+
+    def __call__(self, engine, **kwargs):
+
+        if len(kwargs) == 0:
+            raise RuntimeError("No objects to checkpoint found.")
+
+        if self._score_function is not None:
+            index    = engine.current_iteration
+            priority = self._score_function(engine)
+
+        elif self._iteration_interval is not None:
+            if engine.current_iteration % self._iteration_interval != 0:
+                return
+            index    = engine.current_iteration
+            priority = engine.current_iteration
+
+        else:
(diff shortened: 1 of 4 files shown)


ignite/contrib/checkpoint.py
```diff
@@ -0,0 +1,131 @@
...
+                os.remove(tmp.name)
+                raise
+
+            else:
+                tmp.close()
+                os.rename(tmp.name, path)
+
+    def __call__(self, engine, **kwargs):
+        if len(kwargs) == 0:
+            raise RuntimeError("No objects to checkpoint found.")
```
alykhantejani: Does the model need to be a `kwarg`? can it not just be an arg (which can also be a list) and then you get this check for free?

ignite/contrib/checkpoint.py
```diff
@@ -0,0 +1,131 @@
...
+        Directory path where objects will be saved
+    filename_prefix : str
+        Prefix for the filenames to which objects will be saved.
+        Each object is saved under '{filename_prefix}_{name}_{step_number}.pth'
+        where {step_number} is epoch or iteration number.
+    iteration_interval : int, optional
+        if not None, model will be saved to disk every 'iteration_interval' iterations
+    epoch_interval : int, optional
+        if not None, model will be saved to disk every 'epoch_interval' epochs
+    score_function : Callable, optional
```
alykhantejani: It wasn't clear to me from the docstrings how the priority is calculated and balanced between `iteration_interval`, `epoch_interval` and the `score_function`. For example, if I want to checkpoint when the lowest training score is hit, and also every 5th iteration I don't think this is possible, as the function will terminate early if it's not a 5th iteration. 

I'm ok with this not being possible, but it should be clear in the docstring and we should also raise an error if more than one of `iteration_interval`, `epoch_interval` or `score_function` are set?

ignite/contrib/checkpoint.py
```diff
@@ -0,0 +1,131 @@
...
+
+        self._dirname            = dirname
+        self._fname_prefix       = filename_prefix
+        self._n_saved            = n_saved
+        self._iteration_interval = iteration_interval
+        self._epoch_interval     = epoch_interval
+        self._score_function     = score_function
+        self._atomic             = atomic
+
+        self._item_T  = namedtuple('_item_T', ('priority', 'data'))
```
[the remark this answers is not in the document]
elanmart: It was an `item_type` for the list of saved models, but I guess it's complicating things. Will remove.

ignite/contrib/checkpoint.py
elanmart: I'm not sure how we should name the files where object will be saved, so I decided to write them to `{prefix}_{name}_{index}.pth`, where the `name` is taken from `kwargs`. 

This should be useful if you want to save multiple objects into seperate files, e.g. `encoder` and `decoder`. 

WDYT?

ignite/contrib/checkpoint.py
elanmart: I've updated the docstring.

I have also decided to combine the `iteration_interval` and `epoch_interval`. With previous solution you had to **both** attach the handler to the correct event, and provide the correct argument. I believe it is nicer now.

ignite/contrib/checkpoint.py
alykhantejani: maybe a dict of {name : model} - just so that it's explicit in the docstring?

ignite/handlers/checkpoint.py
```diff
@@ -0,0 +1,122 @@
...
+        self._dirname = dirname
+        self._fname_prefix = filename_prefix
+        self._n_saved = n_saved
+        self._save_interval = save_interval
+        self._score_function = score_function
+        self._atomic = atomic
+        self._saved = []  # list of tuples (priority, saved_objects)
+        self._iteration = 1
+
+        if (save_interval and score_function) or (not save_interval and not score_function):
```
alykhantejani: I think this can be simplified to `save_interval ^ score_function`?

alykhantejani:
Hey I think this looks good. Just need to settle on how models + model names are input to the function, then it's good to go.

ignite/contrib/checkpoint.py
elanmart: Ok, I'll also add an example in the docstring.

elanmart:
Hmm, so I used the built-in github button to resolve a single-line conflict in `handlers/__init__.py`, and it added a whole merge conflict. That's unfortunate, sorry for that.

elanmart:
@alykhantejani I've added some tests, and finally even python2 is green :)

elanmart:
@alykhantejani sure thing, I'll make changes tomorrow.

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -0,0 +1,128 @@
...
+    expected = ['{}_{}_{}.pth'.format(_PREFIX, 'name', i)
+                for i in [6, 8]]
+
+    assert sorted(os.listdir(dirname)) == expected
+
+
+def test_best_k(dirname):
+    scores = iter([1.0, -2., 3.0, -4.0])
+
+    def score_function(engine):
```
alykhantejani: nit: this param should be called `state` now

alykhantejani:
Thanks @elanmart!

vfdev-5:
@elanmart thanks for this handler 👍 ! 

What do you think about to store `priority` value in the output filename? For example, we can add a supplementary option `_fname_postfix` and if it not `None` define the filename as : 
```python
fname = '{}_{}_{}_{}.pth'.format(self._fname_prefix, name, self._iteration, self._fname_postfix)
```
and user can define its postfix with a known template name, for example like this`_val_loss={score:.4f}` and we resolve `_fname_postfix` as 
```
if 'score' in self._fname_postfix:
    self._fname_postfix = self._fname_postfix.format(score=abs(priority))
```
Absolute value is to avoid negative values as priority should be maximized.

PR #81, opened 2018-02-14
add CONTRIBUTING.md

### LINKED ISSUE #79: CONTRIBUTING questions
@alykhantejani before we start merging a bunch of new features, how do you feel about adding a CONTRIBUTING.md doc that spells things out a lil bit more? The main thing I care about right now is settling on a docstring format. I think you said you were okay with going with the google style (ala pytorch) but it'd be nice to get that down somewhere. I can then go back and update the existing docstrings.

jasonkriss:
Resolves #79 

This is basically copy/pasted from pytorch. It has what I think are the most important parts. My preference would be to get this merged and then we can flesh it out over time as we see fit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,24 @@
+# Contributing to Ignite
+
+If you are interested in contributing to Ignite, your contributions will fall
+into two categories:
+1. You want to propose a new feature and implement it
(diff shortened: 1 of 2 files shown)


alykhantejani:
Thanks @jasonkriss!

Issue #78, opened 2018-02-13
Evaluation/Metrics approach


jasonkriss:
This has been discussed a bit in other issues, but I wanted to make a dedicated issue for us to discuss this as I think it's very important we get this right.

## Background
Some previous discussions [here](https://github.com/pytorch/ignite/pull/54#issuecomment-363231705) and [here](https://github.com/pytorch/ignite/issues/72).

Throughout, I am going to use the motivating example of training a supervised model where you periodically want to compute some metrics against a validation set.

## Current Setup
Currently, in order to accomplish this, you need to do the following:

1. create an `Evaluator`
2. register the `Evaluate` handler to run the `Evaluator` on the validation set and store the predictions in the history
3. add another event handler to actually use this history to compute the metrics you care about
4. log/plot these metrics however you choose

In code, this looks something like this:

```python
model = ...
validation_loader = ...
trainer = ...
evaluator = create_supervised_evaluator(model, cuda=True)
trainer.add_event_handler(Events.EPOCH_COMPLETED, Evaluate(evaluator, validation_loader, epoch_interval=1))
@trainer.on(Events.EPOCH_COMPLETED)
def log(engine):
    print(engine.current_epoch, categorical_accuracy(evaluator.history))
```

### Pros
- keeps library code cleanly separated with minimal implicit dependencies
- user doesn't have to write much code

### Cons
- can be confusing what happens where
  - we have both an `Evaluator` and an `Evaluate` and yet neither one computes any sort of metrics
- there are a lot of ways this could go wrong
  - you have to understand the contract between what gets stored in the `Evaluator`'s history and the metrics functions
  - you have to make sure you attach the `Evaluate` handler and any logging handlers to the same event

## Goals
Evaluating a model is gonna be something that (essentially) everyone does so I think we need to have a good story here. Imo, we should make the supervised case be super easy while still making it possible for non-supervised cases. That being said, I think we want to accomplish this without removing flexibility and without adding a ton of code.

## Ideas
Working backward from what I would like the api to be, it might be nice if you could just do something like this:

```python
model = ...
validation_loader = ...
trainer = ...
@trainer.on(Events.EPOCH_COMPLETED)
def evaluate(engine):
  results = evaluate(model, {'acc': categorical_accuracy})
  # do something with those results
```

It'd be even nicer if I could do something like this:
```python
model = ...
validation_loader = ...
trainer = ...
[...]

PR #83, opened 2018-02-20
introduce Engine State object

### LINKED ISSUE #80: Engine state object
We touched on this in #37 and #52 but I wanted to make a dedicated issue to discuss it.

## Motivation
- avoid the proliferation of `current_*` attributes on `Engine`s
- allow for handlers to pass state between each other

## Proposal
- similar to torchnet approach
- we create a new `state` object at the start of `Engine#run` and it is updated throughout the run
- `Trainer` state can have the following to start: dataloader, history, iteration, epoch, max_epochs
- `Evaluator` state can have the following to start: dataloader, history, iteration
- final state is returned by `Engine.run`

## Open questions
- do we still need to pass engine to event handlers or can we just pass state?
- should state just be a dict or an actual object? (@elanmart I think you had some thoughts on this)

What do you all think?

jasonkriss:
Resolves #80 

This PR sets us up nicely for #78. In the meantime, history, logging, metrics, and the evaluate handler have been removed. 

Once this is merged, I will put together a PR addressing #78 that will introduce the new evaluation approach and stateful metrics. Once that PR is merged, things should stabilize quite a bit and we can bring the examples, docstrings, and README up to date.


tests/ignite/test_engine.py
```diff
@@ -68,7 +61,7 @@ def test_adding_multiple_event_handlers():
 
     engine.run(1)
     for handler in handlers:
-        handler.assert_called_once_with(engine)
+        handler.assert_called_once()
```
alykhantejani: This should be called with the state right?

ignite/engine.py
```diff
@@ -93,44 +99,34 @@ def decorator(f):
...
         return decorator
 
-    def _fire_event(self, event_name, *event_args):
+    def _fire_event(self, event_name, state, *event_args):
         if event_name in self._event_handlers.keys():
             self._logger.debug("firing handlers for event %s ", event_name)
             for func, args, kwargs in self._event_handlers[event_name]:
-                func(self, *(event_args + args), **kwargs)
-
-    def terminate(self):
```
alykhantejani: hmm this causes some issues. As I feel `terminate` should really be a method on the engine. I see the work around below of setting a `terminate` attribute on the state, but it really feels more natural to have it on the `Engine` paired with `run`

alykhantejani: This can be kept here if we keep the state private to the engine and expose it via a `get_state()` function, then the handlers still get passed the engine and can access the state (and the method `get_state()` is more obvious than just having it as an attribute)

tests/ignite/test_engine.py
jasonkriss: good catch

ignite/engine.py
jasonkriss: My thinking is that state is a property of the specific "run", not of the trainer itself. Similarly, you are terminating the current run rather than the engine/trainer itself. That being said, i'm definitely ok with reverting the terminate changes. I don't feel strongly one way or the other there.

I would like to keep it as the state being passed to handlers though as I think that makes the most sense.

What do you think? Should I just revert the terminate change?

alykhantejani: By just reverting the terminate change, how would a user call `terminate()` on the engine from within a handler?

jasonkriss: The way I have it being done in the tests is the handlers close over the trainer. So they can just call trainer.terminate() without needed it passed in. I recognize this is less than ideal though.

Here are the options I see:

1. Keep it the way it is in this PR. Handlers can toggle `state.terminate` in order to terminate the run.
2. add a `State.terminate()` method that essentially does what `Trainer#terminate()` did before
3. revert to previous `Trainer#terminate()` approach and require that handlers close over the trainer if they need to call this
4. revert to previous `Trainer#terminate()` approach and pass two arguments to each handler: trainer and state.
5. revert to previous `Trainer#terminate()` approach and set state as an attribute on Trainer. Handlers get passed the trainer.

Personally I like option 1. But if we feel like `terminate` belongs on the `Engine` itself, then I would vote for option 4. The only option I really would like to avoid is option 5. Since state is ephemeral and specific to that run, I don't think it should be an attribute of the trainer. Also, most handlers are gonna be interacting with the state object, so it'd be nice to avoid having to go through the trainer to do this.

What do you think?

alykhantejani: Personally I don't think terminate belongs on the state as it logically pairs with `run`. Additionally doing `state.terminate = True` is not obvious that it should effect change in the engine. 

`terminate()` will probably be called quite a bit for various early stopping handlers, so the usage pattern has to be 1) easy and 2) intuitive. The only problem with closing over the trainer is I'm not sure how intuitive this is. 

I also agree that passing 2 args to every handler when mostly you will use one or the other is not ideal, and neither is having to do `state = trainer.get_state()` at the start of every handler. 

I'm ok with 3/4/5 but neither stand out as clear choices. Also 2 could work but calling/reading `state.terminate()` still isn't super clear to me. 

I'll try and think more about this tonight - but I don't think it's a huge problem, either of them will work we just need to make it clear in examples

jasonkriss: All good points. I think you make a strong argument not to use option 1 or 2. I'm hesitant to use option 3 as I agree that it will likely be very unintuitive for users.

I actually think option 4 works quite well though. Conceptually it makes sense to me that in a handler i am passed both the engine itself and the current state of the run. Its also the easiest from user education standpoint, IMO.

alykhantejani: Let's go with 4 then, I think it's a bit more verbose but will be very clear from the API what's going on.

jasonkriss:
@alykhantejani updated the handlers to now take both engine and state. This should be ready to go pending your approval.

alykhantejani:
LGTM. 

We still need to sketch out an nicer API for users to add their own state variables to this and clear them easily. Also I guess we shouldn't let users modify things like `iteration`, `epoch` etc. or at least they should not be cleared when `clear()` is called.

alykhantejani:
Thanks @jasonkriss!

Issue #85, opened 2018-02-21
Add some tests for ignite/handlers/timing


alykhantejani:
Just to check average, step, pause, resume etc work as expected

PR #87, opened 2018-02-21
Add Timer tests

elanmart:
Solves #85
@alykhantejani sorry for doing that with the previous PR.

### DIFF
--- a/ignite/handlers/timing.py
+++ b/ignite/handlers/timing.py
@@ -3,7 +3,7 @@ from ignite.engine import Events
 try:
     from time import perf_counter
 except ImportError:
-    from time import clock as perf_counter
+    from time import time as perf_counter
 
 
 class Timer:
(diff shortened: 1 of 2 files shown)


alykhantejani:
looks like tests fail on python 2.7 as `time. perf_counter()` doesn't exist. I believe in 2.7 the preferred way was `clock()` maybe add a check and import `clock` as `perf_counter()`

elanmart:
OK, so turns out that `clock` doesn't measure wall-time, so I swapped it for `time.time`.

ignite/handlers/timing.py
```diff
@@ -1,6 +1,10 @@
-import time
 from ignite.engine import Events
 
+try:
```
alykhantejani: nit: `time.time` is available in both python 2.7 and 3 so could just use that for both

elanmart: Yeah, but I thought that if `perf_counter` has higher precision, why not use it when available.

alykhantejani: Fair enough.

alykhantejani:
Thanks @elanmart !

PR #88, opened 2018-02-22
Introduce Metrics

### LINKED ISSUE #78: Evaluation/Metrics approach

jasonkriss:
Resolves #78 

Depends on #83.

This allows us to do the following:

```python
model = ...
training_loader = ...
validation_loader = ...
trainer = ...
evaluator = create_supervised_evaluator(model, metrics={'acc': CategoricalAccuracy()})
def eval(state):
  training_acc = evaluator.run(training_loader).metrics['acc']
  validation_acc = evaluator.run(validation_loader).metrics['acc']
  print('epoch:', state.epoch, 'training acc:', training_acc, 'validation acc:', validation_acc)
```

This PR currently includes `MSE` and `CategoricalAccuracy` metrics. Once we agree everything looks good, I will add the rest of the metrics.


alykhantejani:
@jasonkriss sorry for the delay on this review - will try to get to it today

ignite/metrics/metric.py
```diff
@@ -0,0 +1,66 @@
...
+    def started(self, engine, state):
+        self.reset()
+
+    def iteration_completed(self, engine, state):
+        self.update(state.output)
+
+    def completed(self, engine, state):
+        state.metrics.append(self.compute())
+
+    def attach(self, engine):
```
alykhantejani: what about having an interface similar to the timer here ([see here](https://github.com/pytorch/ignite/blob/master/ignite/handlers/timing.py#L84)) i.e.:

```python
def attach(start=Events.STARTED, update=Events.ITERATION_COMPLETED, compute=Events.COMPLETED, reset=None)
```

alykhantejani: this way we wouldn't need `started()` and `reset()` or `iteration_completed()` and `update()`,

ignite/metrics/metric.py
```diff
@@ -0,0 +1,66 @@
...
+        pass
+
+    def started(self, engine, state):
+        self.reset()
+
+    def iteration_completed(self, engine, state):
+        self.update(state.output)
+
+    def completed(self, engine, state):
+        state.metrics.append(self.compute())
```
alykhantejani: not sure this is the best approach? What if we have several metrics, `state.metrics` should probably be a dict which is accessed with the name of the metric? which can be passed in to the constructor (and have a default value)?

jasonkriss: Yea I wasn't sure on this one. I'm def open to the dict approach. The only reason I went with the list is for the simpler API, both in the constructor and for accessing the result. I see two options if we want to use a more explicit approach:

Option 1
```python
evaluator = create_supervised_evaluator(model, metrics={'acc': CategoricalAccuracy()})
```

Option 2
```python
evaluator = create_supervised_evaluator(model, metrics=[CategoricalAccuracy(name='acc')])
```

What do you think?

ignite/metrics/metric.py
jasonkriss: I'm not sure if the flexibility of passing in the events is something we need (or even want) for metrics. Do you have a use case in mind? My goal here was to have a well-defined interface for metrics (`reset`, `update`, `compute`) that doesn't require that developers of new metrics concern themselves with events at all. Getting rid of `started` or `iteration_completed` would mean that event concepts would leak into the metric implementations. What do you think?

alykhantejani: I agree the defined interface of `reset`, `update` and `compute` is really nice and clean. 

Is it the idea that metrics will always be updated every iteration, and only computed at the end of your engine run as I imagine that metrics will also be used in training (not just evaluation) to monitor. Also, in this case when/how should `reset()` be called? 

I guess the notion of "attaching" the metric to an engine already ties it to an event model, but by adding in the flexibility it allows them to be used in training (i.e. reset every epoch/compute every epoch) and the evaluator. By defining the defaults to something sensible means for the common use case of an evaluator it's just an out-of-the-box `metric.attach(evaluator)`?

It's not super clear with the current API how to use it with a Trainer, or if this is a use case?

ignite/metrics/metric.py
alykhantejani: maybe option 1? then it's consistent with how you pass it in and how its represented on the `State`?

ignite/metrics/metric.py
jasonkriss: Ahh I see. The issue here is that I have built this out (so far at least) to be purely used with evaluators. So i guess we should first take a step back and discuss that...

Should these be just used with evaluators or should they work with trainers as well?

My personal preference is for metrics to be solely an evaluation concern. I think using them with trainers will cause quite a bit of confusion. We would need to modify `create_supervised_trainer` to take a set of metrics and change the return value of the supervised `process_fn`. I think the division of Trainer and Evaluator starts to get a lil confusing if metrics are used on both.

The general (supervised) workflow I've had in mind, is you can log the loss returned by the trainer's `process_fn` to gauge convergence every iteration. But evaluation of metrics happens against a static target (no changing weights during the computation) every so often. The evaluation can certainly be done against the training set as well as an validation set as in the example of above. 

What do you think?

ignite/metrics/metric.py
jasonkriss: 👍 I think that sounds great. I'll make the change.

ignite/evaluator.py
```diff
@@ -12,24 +12,25 @@ def add_event_handler(self, event_name, handler, *args, **kwargs):
...
 
 
-def create_supervised_evaluator(model, cuda=False):
+def create_supervised_evaluator(model, metrics={}, cuda=False):
     """
     Factory function for creating an evaluator for supervised models
 
     Args:
         model (torch.nn.Module): the model to train
+        metrics (list of Metrics): a list of metrics to compute
```
alykhantejani: This should now be dict

ignite/metrics/metric.py
alykhantejani: Ah I see what you mean, yes! I really like this approach! If the user wants to run metrics on the training set they can define an evaluator and run it on the training set. 

Perfect, this looks pretty good to me. Thanks @jasonkriss!

ignite/evaluator.py
jasonkriss: good catch

vfdev-5:
Sorry I missed previous discussions on the subject, I wonder whether we can have a wrapping on a custom loss function as a metric for the evaluator ?
In the example [mnist_with_tensorboardx.py](https://github.com/pytorch/ignite/blob/master/examples/mnist_with_tensorboardx.py) we create a trainer with a loss function nn.NLLLoss() and an evaluator that computes metrics:
```python
trainer = create_supervised_trainer(model, optimizer, nn.NLLLoss(), cuda=cuda)
                                                     'nll': NegativeLogLikelihood()}, cuda=cuda)
```
but imagine a situation when the loss function is something custom and we would like also to compute it on the validation.

jasonkriss:
@vfdev-5 I think that's a great idea and something I was considering myself. I'll open a PR right now for y'all to review.

jasonkriss:
@vfdev-5 #112

PR #92, opened 2018-03-06
Add the rest of the pre-existing metrics

### LINKED ISSUE #93: Add pairwise distance to Metrics
I think in evaluation of regression task, pairwise distance, especially norm-2 distance, as in `torch.nn.functional.pairwise_distance` is at least as frequently used as MSE, which is actually mostly used as loss rather than evaluation metrics. Therefore, I was wondering if it is worthy of being added to Metrics package as a commonly used metrics.

jasonkriss:
Resolves #93

### DIFF
--- a/ignite/metrics/root_mean_squared_error.py
+++ b/ignite/metrics/root_mean_squared_error.py
@@ -1,0 +1,29 @@
+from __future__ import division
+import math
+
+import torch
+
+from .metric import Metric
+from ignite.exceptions import NotComputableError
+
+
+class RootMeanSquaredError(Metric):
+    """
+    Calculates the root mean squared error.
+
+    `update` must receive output of the form (y_pred, y).
+    """
+    def reset(self):
+        self._sum_of_squared_errors = 0.0
+        self._num_examples = 0
+
+    def update(self, output):
+        y_pred, y = output
+        squared_errors = torch.pow(y_pred - y.view_as(y_pred), 2)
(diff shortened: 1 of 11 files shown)


ignite/metrics/root_mean_squared_error.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import division
+import math
+
+import torch
+
+from .metric import Metric
+from ignite.exceptions import NotComputableError
+
+
+class RootMeanSquaredError(Metric):
```
alykhantejani: could this not inherit from MeanSquaredError and just override the `compute` function to be:
```python
def compute(self):
    mse = super(self, RootMeanSquaredError).compute()
    return math.sqrt(mse)
```

jasonkriss: Definitely. One other option is to just have MeanSquaredError but have it take a `root` flag in the constructor. I think I prefer the more explicit approach of having both MeanSquaredError and RootMeanSquaredError though. What do you think?

jasonkriss: @alykhantejani I went ahead and made your suggested change. I can definitely update it though if you wanna go a different route.

jasonkriss:
@alykhantejani I just added MeanPairwiseDistance to this PR.

ignite/metrics/root_mean_squared_error.py
alykhantejani: 👍 I also like the explicit route

Issue #67, opened 2018-02-08
Add examples into CI tests


alykhantejani:

vfdev-5:
@alykhantejani how do you think to test it ? Something like:
```
assert os.system("python mnist.py") == 0
```
or something more sophisticated?

alykhantejani:
@vfdev-5 in travis we can just launch the script directly (no need to write tests in python). I think we should just train for a single epoch (or even less). It's basically a smoke test to make sure the examples still run (with the changing API).

jasonkriss:
@alykhantejani I think this one can be closed, right?

alykhantejani:
yup, thanks @jasonkriss

PR #101, opened 2018-03-13
Fix issue #67

vfdev-5:


.travis.yml
```diff
@@ -18,9 +18,21 @@ install:
   - source activate test-environment
   - python setup.py install
   - pip install --upgrade pytest
+  - pip install visdom torchvision tensorboardX
 
 script:
   - pytest tests/
+  # Smoke tests for examples
+  # mnist.py
+  - python examples/mnist.py --epochs=2
```
alykhantejani: any reason you went for 2 and not 1? if it's arbitrary than maybe 1 is better so that it ends quicker?

vfdev-5: Actually no particular reason, I can change to 1

alykhantejani:
Thanks @vfdev-5!

PR #112, opened 2018-03-15
add generic loss metric

jasonkriss:


ignite/metrics/loss.py
```diff
@@ -1,25 +1,33 @@
...
+        self._loss_fn = loss_fn
+
     def reset(self):
         self._sum = 0
         self._num_examples = 0
 
     def update(self, output):
         y_pred, y = output
-        loss_sum = nll_loss(Variable(y_pred), Variable(y), size_average=False).data[0]
+        loss_sum = self._loss_fn(Variable(y_pred), Variable(y), size_average=False).data[0]
```
alykhantejani: should we  pass `volatile=True` into the Variables y_pred

jasonkriss: yea good catch.

vfdev-5:
@jasonkriss I have a problem with the implementation. If loss function is generic, then `y_preds` and `y` can be anything (and tensors as a special case), so `update` should be something like

```python
    def update(self, output):
        y_pred, y = output
        # NOTE ABSENSE OF size_average=True, AS loss_fn CAN BE ARBITRARY 
        loss_sum = self._loss_fn(to_variable(y_pred, volatile=True), to_variable(y, volatile=True))        
        self._sum += loss_sum.data[0]
        # HERE DO NOT KNOW WHAT TO DO ?
        self._num_examples += y.shape[0]
```
But the problem is with `y.shape[0]` which can be not defined for `y`.

Can we just report the sum... 
Any ideas ?

jasonkriss:
@vfdev-5 hmm that seems like a pretty special case if I'm understanding you correctly. I think at that point a custom Metric is the way to go. That is, If your loss function adheres to the pytorch loss function API, then you can use the ignite `Loss` metric. But if your loss function does not adhere to that API, you can just create your own `Metric` or just subclass `Loss` and override the `update` method. Does that seem reasonable?

vfdev-5:
> That is, If your loss function adheres to the pytorch loss function API, then you can use the ignite Loss metric. But if your loss function does not adhere to that API, you can just create your own Metric or just subclass Loss and override the update method. 

@jasonkriss I agree with this, however, I think `Loss` could also accept the `criterion` used in the training phase (`nn.NLLLoss()` from the example). So we do not need to search function from `F` or recode a custom loss function. For example:
```python
criterion = nn.NLLLoss()
trainer = create_supervised_trainer(model, optimizer, criterion, cuda=cuda)
evaluator = create_supervised_evaluator(model, metrics={'nll': Loss(criterion)})
```

alykhantejani:
That should work already I believe as criterion is a callable. So `F.nll_loss == nn.NLLLoss() == criterion`

vfdev-5:
Signatures can be different. If you pass some kwargs to `loss=nn.NLLLoss()` it may fail like:
```
TypeError: forward() got an unexpected keyword argument 'XXX'
```
for example:
```python
import torch
import torch.nn as nn
import torch.autograd as autograd

loss = nn.NLLLoss()

target = autograd.Variable(torch.LongTensor([1, 0, 4]))
pred = autograd.Variable(torch.LongTensor([2, 0, 3]))
kwargs = {'size_average': True}
output = loss(pred, target, **kwargs)

```

jasonkriss:
@vfdev-5 I apologize but i think I may not be fully understanding the issue. Both of those examples should just work with the current `Loss` metric. Do you maybe have a fully worked out example that isn't working correctly?

vfdev-5:
@jasonkriss no problem, maybe I should be more explicit, here is the problem I have. If you replace in the `mnist.py` file the lines [54 - 55](https://github.com/pytorch/ignite/blob/master/examples/mnist.py#L54) by the following
```
    criterion = nn.NLLLoss()
    trainer = create_supervised_trainer(model, optimizer, criterion, cuda=cuda)
                                                     'nll': Loss(criterion)},
                                            cuda=cuda)

```
at validation phase, the code should fail with the error:
```
  File "/home/user/ignite/ignite/metrics/loss.py", line 25, in update
    loss_sum = self._loss_fn(Variable(y_pred, volatile=True), Variable(y, volatile=True), size_average=False)
  File "/usr/local/lib/python3.6/site-packages/torch/nn/modules/module.py", line 357, in __call__
    result = self.forward(*input, **kwargs)
TypeError: forward() got an unexpected keyword argument 'size_average'
```
This is due to different signatures between `nn.NLLLoss()` and `F.nll_loss`...
Hope it is more clear :)

jasonkriss:
Thanks @vfdev-5! Interesting... I had assumed that `nn.NLLLoss() == F.nll_loss`. That's a lil frustating. Looks like the class-based approach requires you pass those kwargs at instantiation time. Not sure I agree with that. 

Let me see if I can address this. I'll get back to you soon...

alykhantejani:
heh, that is annoying. @vfdev-5 do you mind opening an issue for this?

jasonkriss:
@vfdev-5 @alykhantejani I just opened #120 to address this.

alykhantejani:
awesome, thanks!

PR #113, opened 2018-03-16
Add to_tensor method and update evaluator

vfdev-5:
This covers possible issues using `create_supervised_evaluator` if targets are not simple `torch.Tensor`

### DIFF
--- a/ignite/engines/evaluator.py
+++ b/ignite/engines/evaluator.py
@@ -1,4 +1,4 @@
-from ignite._utils import to_variable
+from ignite._utils import to_variable, to_tensor
 from ignite.engines import Engine, Events, State
 
 __all__ = ["Evaluator", "create_supervised_evaluator"]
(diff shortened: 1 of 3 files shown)


jasonkriss:
@vfdev-5 Do you mind giving a concrete use case where the supervised targets are not tensors?

vfdev-5:
@jasonkriss yep, in a detection task with SSD targets are encoded  boxes and labels (location, cls). See [this](https://github.com/kuangliu/torchcv/blob/master/torchcv/loss/ssd_loss.py#L35)

vfdev-5:
@alykhantejani what do you think about this ?

ignite/engines/evaluator.py
```diff
@@ -42,7 +42,7 @@ def _inference(batch):
         model.eval()
         x, y = _prepare_batch(batch)
         y_pred = model(x)
-        return y_pred.data.cpu(), y.data.cpu()
+        return to_tensor(y_pred, cpu=not cuda), to_tensor(y, cpu=not cuda)
```
alykhantejani: nit: cpu could always be true here. I think its a no-op if it's already on the cpu

vfdev-5: Do you think it would not impact metrics computation time if it is on cpu ?

jasonkriss:
Thanks @vfdev-5! Seems totally reasonable to me.

alykhantejani:
Thanks @vfdev-5!

Issue #114, opened 2018-03-16
Invalid cross-device link with ModelCheckpoint, atomic=True


vfdev-5:
I have an issue with `ModelCheckpoint` when `atomic=True`
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/trainer.py", line 53, in run
    self._handle_exception(state, e)
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 138, in _handle_exception
    raise e
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/trainer.py", line 40, in run
    hours, mins, secs = self._run_once_on_dataset(state)
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 132, in _run_once_on_dataset
    self._handle_exception(state, e)
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 138, in _handle_exception
    raise e
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 123, in _run_once_on_dataset
    self._fire_event(Events.ITERATION_COMPLETED, state)
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 106, in _fire_event
    func(self, state, *(event_args + args), **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/handlers/checkpoint.py", line 147, in __call__
    self._save(obj=obj, path=path)
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/handlers/checkpoint.py", line 124, in _save
    os.rename(tmp.name, path)
OSError: [Errno 18] Invalid cross-device link: '/tmp/tmpe7shyxri' -> '/home/user/output/weights/_SSD300_1.pth'
```
Maybe we need to use `shutil.move` as suggested [here](https://stackoverflow.com/questions/42392600/oserror-errno-18-invalid-cross-device-link)

veugene:
[...]

PR #115, opened 2018-03-18
During atomic save in checkpoint, save temporary file to target directory rather than to default TMPDIR.

veugene:
Addresses bug #114.

> Currently, ModelCheckpoint uses the default destination in tempfile when saving a temporary file for an atomic checkpoint write. This is potentially problematic: it assumes that the default tempfile.tempdir is on the same volume as the user-specified destination directory (dirname); it also causes the issue described above. It should create the temporary file in the user-specified destination directory prior to renaming it (NamedTemporaryFile should take dir=dirname).

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -109,7 +109,7 @@ class ModelCheckpoint(object):
             torch.save(obj, path)
 
         else:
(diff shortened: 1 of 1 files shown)


alykhantejani:
Thanks @veugene!

Issue #116, opened 2018-03-19
Allow setting initial epoch in trainer.


veugene:
The initial epoch when training is assumed to always be 0; it would be useful for a user to be able to initialize this value when resuming training.

https://github.com/pytorch/ignite/blob/master/ignite/engines/trainer.py#L31

alykhantejani:
I think a better solution for this would be to implement `engine.save` and `engine.load` so that you can resume from a engine checkpoint seamlessly. There is already a ticket for this #62 so closing in favor of that. Feel free to comment on this ticket if you think that wont solve the issue

PR #119, opened 2018-03-19
Allow setting initial epoch in Trainer

veugene:
Addresses #116


alykhantejani:
Closing in favor of an implementation of `engine.save` and `engine.load` (#62 )

PR #120, opened 2018-03-19
allow Loss metric to work with non-functional losses

jasonkriss:
Addresses [this issue](https://github.com/pytorch/ignite/pull/112#issuecomment-374296668)

### DIFF
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -10,6 +10,8 @@ class Loss(Metric):
     """
     Calculates the average loss according to the passed loss_fn.
 
+    `loss_fn` must return the average loss over all observations in the batch.
+
     `update` must receive output of the form (y_pred, y).
     """
     def __init__(self, loss_fn):
@@ -23,6 +25,7 @@ class Loss(Metric):
     def update(self, output):
         y_pred, y = output
         average_loss = self._loss_fn(Variable(y_pred, volatile=True), Variable(y, volatile=True))
+        assert average_loss.shape == (1,), '`loss_fn` did not return the average loss'
         self._sum += average_loss.data[0] * y.shape[0]
         self._num_examples += y.shape[0]
 
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -22,8 +22,8 @@ class Loss(Metric):
 
     def update(self, output):
         y_pred, y = output
-        loss_sum = self._loss_fn(Variable(y_pred, volatile=True), Variable(y, volatile=True), size_average=False)
-        self._sum += loss_sum.data[0]
+        average_loss = self._loss_fn(Variable(y_pred, volatile=True), Variable(y, volatile=True))
+        self._sum += average_loss.data[0] * y.shape[0]
         self._num_examples += y.shape[0]
 
     def compute(self):
--- a/tests/ignite/metrics/test_loss.py
+++ b/tests/ignite/metrics/test_loss.py
@@ -41,6 +41,15 @@ def test_compute_on_criterion():
     assert_almost_equal(loss.compute(), 1.1253643036)  # average
 
 
+def test_non_averaging_loss():
+    loss = Loss(nn.NLLLoss(reduce=False))
+
+    y_pred = torch.Tensor([[0.1, 0.4, 0.5], [0.1, 0.7, 0.2]]).log()
+    y = torch.LongTensor([2, 2])
+    with pytest.raises(AssertionError):
+        loss.update((y_pred, y))
+
+
 def test_reset():
     loss = Loss(nll_loss)
 
--- a/tests/ignite/metrics/test_loss.py
+++ b/tests/ignite/metrics/test_loss.py
@@ -27,6 +27,20 @@ def test_compute():
     assert_almost_equal(loss.compute(), 1.1253643036)  # average
 
 
+def test_compute_on_criterion():
+    loss = Loss(nn.NLLLoss())
+
+    y_pred = torch.Tensor([[0.1, 0.4, 0.5], [0.1, 0.7, 0.2]]).log()
+    y = torch.LongTensor([2, 2])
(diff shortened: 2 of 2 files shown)


ignite/metrics/loss.py
```diff
@@ -22,8 +22,8 @@ def reset(self):
 
     def update(self, output):
         y_pred, y = output
-        loss_sum = self._loss_fn(Variable(y_pred, volatile=True), Variable(y, volatile=True), size_average=False)
-        self._sum += loss_sum.data[0]
+        average_loss = self._loss_fn(Variable(y_pred, volatile=True), Variable(y, volatile=True))
+        self._sum += average_loss.data[0] * y.shape[0]
```
vfdev-5: Maybe we can avoid this with:
```
self._sum += average_loss.data[0]
self._num_examples += 1
```

jasonkriss: @vfdev-5 Unfortunately we need to do it this way as we can't guarantee that all batches will be the same size.

vfdev-5: > we can't guarantee that all batches will be the same size.
This is sounds strange to me. @jasonkriss could you hint me an example, please?

My reasoning is that `self._loss_fn` does the average itself as `nn.NLLLoss()`. Maybe it is always the case.

jasonkriss: if your dataset has 1000 examples and you are doing batches of size 256, then your last batch will be of size 232. Therefore to get the true mean loss, we have to make sure we take into account the number of examples in the batch.

vfdev-5: @jasonkriss  I agree with what you say. Anyway, `y.shape[0]` assumes that `y` is `Tensor` and this may be a problem with multi-output...

jasonkriss: @vfdev-5 once #121 is merged, that won't be an issue any longer.

vfdev-5: @jasonkriss imagine a case when `self._loss_fn` is a sort of composite loss and works on `y_pred` and `y` being a list of tensors, so no need of `output_transform` (from #121) to apply, and `y.shape[0]` will fail...

jasonkriss: I think at that point, the recommended path should probably just be a custom `Metric` or subclass of `Loss`. Unless you can think of a simple way to support it here.

alykhantejani: Going back to 
```python
self._sum += average_loss.data[0] * y.shape[0]
self._num_examples += y.shape[0]
```
what if my criterion is `nn.NLLLoss(size_average=False)` (for whatever reason), then this will fail. 

I mean to support all cases we can get super verbose in the code i.e.
```python
if isinstance(loss_fn, nn.Loss):
    if loss_fn.size_average:
      .....
     else:
       .....
else:
      self.loss_fn(y_pred, y, size_average=False)
      .....
```

tests/ignite/metrics/test_loss.py
```diff
@@ -27,6 +27,20 @@ def test_compute():
     assert_almost_equal(loss.compute(), 1.1253643036)  # average
 
 
+def test_compute_on_criterion():
+    loss = Loss(nn.NLLLoss())
```
alykhantejani: should probably add some cases where the kwargs are set

jasonkriss: Is there a use case where someone would have `size_average=False` and then want to pass it to the `Loss` metric? The way i was looking at it is we want to allow users to pass in the loss function the are using to train their model. I'm afraid this could become a lil pathological if we try to support every permutation. What do you think?

alykhantejani:
Sounds reasonable to me. Let's be explicit in the docstring though, otherwise users will be left with obscure error messages.

We could go one step further and raise a runtime error with a useful error message in this case

jasonkriss:
Sounds good. Definitely agree on the docstring. For raising an error though... under what conditions would we raise the error? The way I'm picturing doing this would require the same checks as actually supporting every permutation.

alykhantejani:
Yeah, I guess we can check the shape of the response from calling loss_fn if it's not 1 then we might just fail silently (not at a laptop so can't double check this)

vfdev-5:
@jasonkriss  @alykhantejani what do you think if we put an additional option `reduce` to the `Loss` which indicates whether we need to reduce or not the loss result. This option should be the same value as in the loss function (user responsibility):
```python
class Loss(Metric):
    def __init__(self, loss_fn, reduce=True):
        super(Loss, self).__init__()
        self._loss_fn = loss_fn
        self.reduce = reduce

    def update(self, output):
        y_pred, y = output
        loss = self._loss_fn(Variable(y_pred, volatile=True), Variable(y, volatile=True))
        if self.reduce:
            self._num_examples += loss.shape[0]
            loss = loss.sum()
        else:
            self._num_examples += 1
        self._sum += loss.data[0]             
```
and we can treat the following cases:
```python
# when reduce is False, size_average is ignored
loss_fn1 = partial(F.nll_loss, size_average=False, reduce=False)
loss_fn2 = partial(F.nll_loss, size_average=True, reduce=False) # <- this should be the same as the previous one
# loss_fn3 = partial(F.nll_loss, size_average=False, reduce=True) # <- probably, this case we would like to avoid
loss_fn4 = partial(F.nll_loss, size_average=True, reduce=True)
```

alykhantejani:
@vfdev-5 we could go this route, but the only reduction we would support it sum. Alternatively, as @jasonkriss said we can just leave this responsibility on the user. However, we definitely need to be explicit in the docs, and a runtime error would be a nice to have - but this can also be added later. 

@jasonkriss can you add a note to the docs and then this is good to merge

jasonkriss:
@alykhantejani Added a note in the doc as well as a test and assert on the return value of the loss_fn. You were correct, that we would just silently fail otherwise.

alykhantejani:
Thanks @jasonkriss!

PR #121, opened 2018-03-19
introduce output_transform to Metrics

jasonkriss:


ignite/metrics/metric.py
```diff
@@ -11,8 +11,15 @@ class Metric(object):
...
+
+    Args:
+        output_transform (callable): an optional callable that used to transform
+            the model's output into the form expected by the metric. This can be
+            useful if, for example, you have a multi-output model and you want to
+            compute the metric with respect to one of the outputs.
     """
-    def __init__(self):
+    def __init__(self, output_transform=None):
+        self._output_transform = output_transform
```
vfdev-5: nit: maybe simlier to define self._output_transform as `lambda x: x` if no output_transform provided and avoid 
```
output = self._output_transform(state.output) if self._output_transform else state.output
```
in the loop

jasonkriss: yea good point. I was considering that. Let's do that. It's what we did before with the old `History`.

alykhantejani:
lets merge #123 first and then this will need a small update

jasonkriss:
@alykhantejani Should be all good once Travis is done.

jasonkriss:
@alykhantejani 👍 #131

Issue #117, opened 2018-03-19
Flexible access to state


veugene:
The introduction of `State` (#80, #83) which is passed to event handlers is very welcome as it is a critical requirement. The flexibility with which state is passed around, however, could still be improved. Currently:

1. Event handlers can only access the state of the engine issuing the event.
2. There is no access to the state in the training and validation functions.

An argument against making state a member of the engine was that in so not doing, the state is explicitly ephemeral. However, this enforces the assumption that an engines state should only every be accessed by that engine's event handlers.

One example of where (1) is an obstacle is on attempting to identify current epoch during training. It is useful to read the current epoch on completion of validation when saving a checkpoint, in order to record the epoch. However, the current epoch is a state attribute of the training engine; currently, there is no way to access that state, except when handling a signal emitted by the training engine.

It would be convenient to have `state` exposed in `Trainer` and `Evaluator` as `self.state` so that a user may have direct access to it, outside of callbacks that are specific to the calling engine.

If state is thus exposed, it may be cleaner to also not pass engine state by default to handlers; the user could pass any state (eg. `trainer.state`) through the *args or **kwargs of `add_event_handler`, instead -- if some state is needed by the event handler.

alykhantejani:
On this point, I was quickly trying to hack up a version of `engine.save()` and `engine.load` and found that having a `state` on the engine was desirable for a clean API.

For example, if we want the API `engine.save()` and `engine.load(file_path)` the engine will need to store and reload the state, which in it's current form can't be done as the state is ephemeral.

I think we should nail down where state belongs and how it's passed around before 0.1.0

cc @jasonkriss who I know has thoughts on this.

@veugene - even if state was a member of the engine, how would this give you access inside the update function, and what is the use case for this?

veugene:
If the state were a member of the engine, then the update function would be the one place where it would not be easy to access unless the function is passed the state as an argument.

Elsewhere, one could do something like the following to pass state to an event handler, if the handler requires a state:

```python

# Some arbitrary event handler that requires engine state.
def handler_using_state(state):
    print("Epoch {}".format(state.epoch))

trainer.add_event_handler(Events.EPOCH_COMPLETED, handler_using_state, trainer.state)
evaluator.add_event_handler(Events.STARTED, handler_using_state, trainer.state)

# A function, handling an event, that does not need engine state.
class some_custom_class(object):
    ...
    def reset(self):
        self._some_attribute = 0
obj = some_custom_class()

[...]

PR #123, opened 2018-03-19
Feature 117 state access

veugene:
Addressing #117:

- Attached `state` to `Engine`
- `Trainer` and `Evaluator` `run()` methods no longer return anything
- `Engine` is the only argument now passed by default to event handlers (not `engine, state`)
- `engine, batch` is now passed to update functions, rather than just `batch`

### DIFF
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -58,18 +58,18 @@ def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
                                             cuda=cuda)
 
     @trainer.on(Events.ITERATION_COMPLETED)
-    def log_training_loss(trainer, state):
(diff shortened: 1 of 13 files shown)


veugene:
Note that the `score_function` in `ModelCheckpoint` has been modified to take `Engine` instead of `State` as an argument.

veugene:
I stacked a bunch of commits fixing code in the tests and examples, following the above changes; I could rebase that or these could all be squashed. This will still need to be reconciled with PR #124.

alykhantejani:
Thanks for the PR @veugene, no need to worry about squashing - I can do that via github on merge.

Will take a look asap

veugene:
@alykhantejani sure, no problem

alykhantejani:
Thanks @veugene!

PR #124, opened 2018-03-19
unify trainer and evaluator into a single engine

### LINKED ISSUE #122: Unify Trainer and Evaluator to a single Engine
Following the discussions in #118

jasonkriss:
Fixes #122 

@alykhantejani @veugene I'd appreciate a review on this one.


veugene:
Looks good. Perhaps the `COMPLETED` and `STARTED` events could be removed, now.

jasonkriss:
@veugene I think there is still a use for those. For example, if you want to run some handler code only at the completion of all training.

veugene:
@jasonkriss Yeah, that makes sense.

alykhantejani:
Also needs merge conflict after #123 was merged

veugene:
Great! By the way, I noticed that the conflict resolution changes re-introduce a return value to `Engine` (`self.state`) - I've previously modified the tests to extract the state from the engine instead of getting it as a return value of `run`. There were tests that verify that the `run` method returns a state; instead, these now verify that a `State` is created in the engine after a `run`.

jasonkriss:
Is there a specific reason why we'd not want `run` to return `state`? To me, it seems like the natural thing to do.

veugene:
To me, it seems neither natural nor unnatural. I merely removed it once there was no longer a need for it, now that `state` is an accessible property of the engine.

jasonkriss:
@veugene 👍 gotcha.

PR #125, opened 2018-03-20
Avoid RecursionError in to_tensor

vfdev-5:
Avoid `RecursionError` if variable contains strings and not `autograd.Variable`

Sorry, I should add this in the previous PR

### DIFF
--- a/ignite/_utils.py
+++ b/ignite/_utils.py
@@ -33,6 +33,8 @@ def to_tensor(variable, cpu=False):
         if cpu:
             t = t.cpu()
         return t
+    elif isinstance(variable, string_classes):
+        return variable
     elif isinstance(variable, collections.Mapping):
         return {k: to_tensor(sample, cpu=cpu) for k, sample in variable.items()}
     elif isinstance(variable, collections.Sequence):


jasonkriss:
Shouldn't we have this raise a `TypeError`?

vfdev-5:
Probably, both `to_tensor` and `to_variable` should raise `TypeError` if a string is passed ?

jasonkriss:
Yea I think so. Otherwise, its just gonna end up raising an error in the code that uses `to_tensor` `to_variable`. Do you mind making that change?

vfdev-5:
@alykhantejani what is your opinion ?

alykhantejani:
Returning a string sounds reasonable to me. It's also whats done in the [dataloader collate function](https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py#L131)

jasonkriss:
@vfdev-5 returning a string does seem like the right thing to do. I was thinking this would cause issues for metrics, but with #121, that's not a problem. This should be good to merge then, right?

alykhantejani:
Yup, was just waiting for tests to re-run. 🚢

alykhantejani:
Thanks @vfdev-5!

PR #128, opened 2018-03-22
Extend accuracy for any targets shape

chicham:
Currently when the accuracy is calculated, it is assumed that the target tensor is of the shape (n_batches, 1) or n_batches. However in some cases like semantic segmentation, the target tensor is of the form (n_batches, n_categories, height, width) where all dimensions contribute to the computation of accuracy.

The following modification allows to take into account this case without changing the original behavior.


ignite/metrics/binary_accuracy.py
```diff
@@ -18,7 +18,7 @@ def reset(self):
 
     def update(self, output):
         y_pred, y = output
-        correct = torch.eq(torch.round(y_pred).type(torch.LongTensor), y)
+        correct = torch.eq(torch.round(y_pred).type(torch.LongTensor), y).view(-1)
         self._num_correct += torch.sum(correct)
         self._num_examples += correct.shape[0]
```
alykhantejani: I'm not that familiar with evaluating the output of semantic segmentation for example, but now your `num_correct` would be number in range `[0, num_categories*num_pixels]` but your `num_examples` is only adding `batch_size` to the count. Should it not be adding `batch_size*num_categories*num_pixels`?

vfdev-5:
I would say that for semantic segmentation task we need some other metrics, like Jaccard, Dice,...

jasonkriss:
I think we should we maybe define a separate `Metric` for this (and the others @vfdev-5 mentioned). Even if we can make these work for both use cases, I'm not sure we should. These are meant to be used for classification and I think it might be best to be explicit by having a separate, correctly named metric. What do you guys think?

ignite/metrics/binary_accuracy.py
chicham: In both cases, now `num_correct` will be a number in range `[0, batch_size * num_pixels]` and `batch_size * num_pixels` will be added to the count because the new shape of `correct` is `(batch_size * num_pixels,)` instead of `(batch_size, image_height, image_width)`. 

Here a gist that shows the differences between the current code and my modifications with simulated data.
https://gist.github.com/chicham/c0128d6e1405246eb99cf0b42731824e
 
Note: num_pixels = image_height * image_width

chicham:
Categorical accuracy is widely used for the evaluation of segmentation models (at least for remote sensing applications https://arxiv.org/pdf/1608.00775.pdf). Of course other metrics are needed, but this modification does not introduce any drawback (from what I could test) and allows to use this `Metric` with tensor of any shape.
Semantic segmentation can be seen as the task of classifying correctly all the pixels of an image so I don't think it is counterintuitive to use this class for this. On the other hand, it is possible that this could introduce some confusion for the user.

jasonkriss:
@chicham I see your point. Thank you for that. I'm pretty convinced myself. I'll let @alykhantejani weigh in though. Might be nice to add a test for each reflecting this use case.

alykhantejani:
Sounds good to me, I also agree with @jasonkriss about adding a test case for this new scenario, then it's good to merge :)

vfdev-5:
Maybe also add some notice in the documentation that target tensor can also have shapes like `(n_batches, n_categories, height, width)`.

chicham:
Thanks for your comments. I add the tests and documentation for `binary_accuracy` and `categorical_accuracy`

alykhantejani:
Hey @chicham how comes you closed the PR?

chicham:
Oups error on my side

alykhantejani:
Thanks @chicham!

PR #131, opened 2018-03-23
Add test for Metric output_transform

jasonkriss:
This also makes a couple changes to make sure you can use the output_transform on all metrics.


alykhantejani:
Thanks @jasonkriss!

Issue #129, opened 2018-03-23
Add filename_suffix option to ModelCheckpoint


vfdev-5:
I would like to add `filename_suffix` option to ModelCheckpoint as discussed in [here](https://github.com/pytorch/ignite/pull/75). The code is [here](https://github.com/vfdev-5/ignite/commit/f374ea83280cbbed5ed752e1e08cf9aa859ef33c).

Idea is to be able to store a score in the filename. For example, if we specify `filename_suffix="_val_loss={score:.4f}"` and `score_function` that returns `-loss`, saved models filenames will be, for example, `model_SqueezeNet_98_val_loss=1.2413.pth`. See [this example](https://github.com/vfdev-5/ignite/blob/new_examples/examples/cifar10_playground.py#L226).

@alykhantejani @jasonkriss what do you think ?

elanmart:
@vfdev-5 I'm sorry for not responding in the PR thread, I was away from the internet for few weeks. 

Personally I don't have a strong need for this feature, but I think that storing the priority can be useful.

I'm not sure about the API though. `filename_suffix="_val_loss={score:.4f}"` seems too implicit and I would surely mess that up.

How about providing just the suffix name? 

```python
if suffix is not None:
    full_suffix = f'{suffix}={abs(priority)}'
    fname = '{}_{}_{}_{}.pth'.format(self._fname_prefix, name, self._iteration, full_suffix)
else:
    fname = '{}_{}_{}.pth'.format(self._fname_prefix, name, self._iteration)

**Comment by alykhantejani** [comment]
I agree with @elanmart about the `_val_loss={score.4f}` being too complicated. I noticed that the current version will already add the loss / epoch count to the filename (see [here](https://github.com/pytorch/ignite/blob/master/ignite/handlers/checkpoint.py#L142)).

Perhaps adding a `count_or_score_prefix` (with hopefully a better name) and [this line](https://github.com/pytorch/ignite/blob/master/ignite/handlers/checkpoint.py#L142) can just change to 
```python
fname = '{}_{}_{}_{}.pth'.format(self._fname_prefix, name, self._count_or_score_prefix, self._iteration)
```

vfdev-5:
@elanmart 👍 nice with f-string :)

> I noticed that the current version will already add the loss / epoch count to the filename (see here).

@alykhantejani could you please explain ?

alykhantejani:
@vfdev Ah sorry I misread a line in the source. What I meant was 

[...]

PR #132, opened 2018-03-27
Add score_name option in ModelCheckpoint

vfdev-5:
Addresses #129 

Changes from the #129 discussion:
- `filename_suffix` -> `score_name`
-  `score_name` is activated if `score_function` is provided otherwise `ValueError`

```
suffix = ""
if self._score_name is not None:
    suffix = "_{}={:.7}".format(self._score_name, abs(priority))
```

cc @alykhantejani @elanmart 

Feel free to comment if I forgot something from the discussion.

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -26,6 +26,9 @@ class ModelCheckpoint(object):
             an `ignite.engines.Engine` object,
             and return a score (`float`). Objects with highest scores will be retained.
             Exactly one of (`save_interval`, `score_function`) arguments must be provided.
+        score_name (str, optional):
+            if `score_function` not None, it is possible to store its absolute value using `score_name`. See Notes for
+            more details.
         n_saved (int, optional):
             Number of objects that should be kept on disk. Older files will be removed.
         atomic (bool, optional):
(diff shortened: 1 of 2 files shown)


ignite/handlers/checkpoint.py
```diff
@@ -26,6 +26,9 @@ class ModelCheckpoint(object):
             an `ignite.engines.Engine` object,
             and return a score (`float`). Objects with highest scores will be retained.
             Exactly one of (`save_interval`, `score_function`) arguments must be provided.
+        score_name (str, optional):
+            if `score_function` not None, it is possible to store its absolute value using `score_name`. See Notes for
```
alykhantejani: why do we want to store the abs value and not the actual value?

ignite/handlers/checkpoint.py
```diff
@@ -84,6 +94,10 @@ def __init__(self, dirname, filename_prefix,
             raise ValueError("Exactly one of `save_interval`, or `score_function` "
                              "arguments must be provided.")
 
+        if score_function is None and score_name is not None:
+            raise ValueError("If `score_name` is setup, then `score_function` "
```
alykhantejani: nit: setup -> provided

ignite/handlers/checkpoint.py
vfdev-5: as `score_function` can return `-loss` and as objects with highest scores are retained, so to avoid sign minus in the filename

alykhantejani: Ah ok, I see your point now. cool LGTM - thanks @vfdev-5!

vfdev-5:
@alykhantejani added your suggestions

alykhantejani:
Thanks @vfdev-5!

Issue #133, opened 2018-03-27
Add early stopping


vfdev-5:
I would like to propose an early stopping handler.
Usage:
```
    from ignite.engines import Engine, Events
    from ignite.handlers import EarlyStopping

    def score_function(engine):
        val_loss = engine.state.metrics['nll']
        return -val_loss

    handler = EarlyStopping(patience=10, score_function=score_function, verbose_logger=print)
    evaluator.add_event_handler(Events.COMPLETED, handler, trainer)
```
[Code](https://github.com/pytorch/ignite/compare/master...vfdev-5:feature_early_stopping)

What do you guys think about ?

alykhantejani:
Sounds like a useful feature, please go ahead with a PR! 

My only nit would be around the logger, instead we should do something like what is done in the [engine](https://github.com/pytorch/ignite/blob/master/ignite/engines/engine.py#L41) and use python's `logging`, this way users can attach their log handlers to the root `ignite` logger

vfdev-5:
@alykhantejani Thanks!  I'll fix the logger and create a PR. Do you mean to add : 
```
self._logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
self._logger.addHandler(logging.NullHandler())
```
into the class ? So, no need to add the option `verbose`.

alykhantejani:
yup, exactly :)

alykhantejani:
closed in #136

PR #136, opened 2018-03-27
Add early stopping

vfdev-5:
Addresses #133


vfdev-5:
@alykhantejani one test fails with [timeout](https://travis-ci.org/pytorch/ignite/jobs/359095717#L700), can you restart it ?

jasonkriss:
Nice work! Love how simple you made this.

vfdev-5:
@jasonkriss thanks!

ignite/handlers/early_stopping.py
```diff
@@ -0,0 +1,50 @@
...
+    Examples:
+    ```python
+    from ignite.engines import Engine, Events
+    from ignite.handlers import EarlyStopping
+
+    def score_function(engine):
+        val_loss = engine.state.metrics['nll']
+        return -val_loss
+
+    handler = EarlyStopping(patience=10, score_function=score_function, verbose_logger=print)
```
alykhantejani: we should remove the `verbose_logger` arg from the example

ignite/handlers/early_stopping.py
```diff
@@ -0,0 +1,50 @@
...
+        assert callable(score_function), "Argument score_function should be a function"
+        assert patience > 0, "Argument patience should be positive"
+        self.score_function = score_function
+        self.patience = patience
+        self.counter = 0
+        self.best_score = None
+        self._logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
+        self._logger.addHandler(logging.NullHandler())
+
+    def __call__(self, engine, trainer):
```
alykhantejani: what do you think about passing the trainer into the constructor instead, this was users don't have to remember to add the additional arg in `add_event_handler`?

ignite/handlers/early_stopping.py
```diff
@@ -0,0 +1,50 @@
...
+        score = self.score_function(engine)
+
+        if self.best_score is None:
+            self.best_score = score
+        elif score < self.best_score:
+            self.counter += 1
+            self._logger.info("EarlyStopping: %i / %i" % (self.counter, self.patience))
```
alykhantejani: should we make this verbose instead of info? I don't have a strong preference either way

tests/ignite/handlers/test_early_stopping.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def update_fn(engine, batch):
+        pass
+
+    trainer = Engine(update_fn)
+
+    h = EarlyStopping(patience=2, score_function=score_function)
+    # Call 3 times and check if stopped
+    assert not trainer.should_terminate
+    h(None, trainer)
```
alykhantejani: nit:maybe check in between calls that `trainer.should_terminate` is false?

ignite/handlers/early_stopping.py
vfdev-5: Correct! Sorry for this!

ignite/handlers/early_stopping.py
vfdev-5: Okay, I see your point.

vfdev-5:
@alykhantejani code is updated

alykhantejani:
Thanks @vfdev-5!

satishpasumarthi:
Is the early stopping feature available in Pytorch?

PR #137, opened 2018-03-28
Use relative import throughout

tkanmae:
The codebase mixes absolute imports and relative imports. This PR replaces absolute import statements with relative ones.


alykhantejani:
Good catch, however I think we should go for absolute imports everywhere instead of relative. Relative imports used to be strongly discouraged in PEP8
```
twitter.common_internal.auth.ssh.AuthorizationError: Could not talk to SSH agent: Unable to find any signable SSH keys.  Make sure you've generated your SSH
```
now they're not so anti relative imports but still recommend absolute over relative.

Do you mind updating the PR?

vfdev-5:
@alykhantejani relative import are however widely used in pytorch, for example [here](https://github.com/pytorch/pytorch/blob/master/torch/nn/functional.py#L11). Is it a coding rules choice or it was developped like this ?

alykhantejani:
not sure what the logic/reasoning was behind doing it this way in pytorch.

tkanmae:
@alykhantejani Thanks. I wasn't aware that PEP8 favors absolute imports over relative ones. I updated the PR. Now, all intra-package import statements are absolute imports.

alykhantejani:
Thanks @tkanmae!

PR #138, opened 2018-03-30
Add Precision and Recall metrics

jasonkriss:


vfdev-5:
@jasonkriss nice work! Could you also make a notice on the type of `y_pred` (logits or probas as we can have an ambiguity from model's `forward()`)? Maybe this would be useful for all metrics...

tests/ignite/test__utils.py
```diff
@@ -36,3 +36,10 @@ def test_to_tensor():
 
     with pytest.raises(TypeError):
         to_tensor(12345)
+
+
+def test_to_onehot():
+    indices = torch.LongTensor([0, 1, 2, 3])
+    actual = to_onehot(indices, 4)
+    expected = torch.eye(4)
+    assert actual.eq(expected).sum() == 16
```
alykhantejani: nit: can just use `actual.equal(expected)` which returns a boolean

jasonkriss:
@alykhantejani fixed. Learn something new every day!

@vfdev-5 these metrics will work for both logits or probabilities. I think you are right though that all metrics could benefit from a lil more verbose docstrings.

alykhantejani:
Thanks @jasonkriss!

vfdev-5:
In my another example, I have an error with `Recall` which happens when `correct` is completely zeros (when model is completely wrong which is not as frequent):
```
correct = torch.eq(indices, y)
```
and thus `indices[correct]` is `torch.cuda.LongTensor with no dimension` and `to_onehot` fails with error:
```
RuntimeError: invalid argument 2: out of range at /pytorch/torch/lib/THC/generic/THCTensor.c:23
```

jasonkriss:
@vfdev-5 Good catches. The bug and import statements are fixed in #140.

On the other issue... If `self._all_positives` or `self._actual` are all zeros, should we maybe raise instead? That shouldn't really ever happen, right. What do you all think? @alykhantejani

vfdev-5:
@jasonkriss thanks for the fixes ! 

> On the other issue... If self._all_positives or self._actual are all zeros, should we maybe raise instead? 

`self._all_positives` can contain zeros (or can be completely zero) but `self._true_positives` contains zeros at same places, so it is `0/0` which can be regularized with `eps`. Here is an example:
```
n_classes = 10
y = torch.torch.LongTensor(5).random_(0, n_classes)
y_pred = torch.zeros(5, n_classes)

precision_handler = Precision()
precision_handler.update((y_pred, y))
res = precision_handler.compute()
print(res, precision_handler._all_positives, precision_handler._true_positives)

> (
   0
 nan
 nan
 nan
 nan
 nan
 nan
 nan
 nan
 nan
 [torch.FloatTensor of size 10], 
  5
  0
  0
  0
  0
  0
  0
  0
  0
  0
 [torch.FloatTensor of size 10], 
  0
  0
  0
  0
  0
  0
  0
  0
  0
  0
 [torch.FloatTensor of size 10])
```
I think raise an exception is too strict, I'm not sure that a metric should stop the training because of very poor model behaviour.

jasonkriss:
I guess the reason why I would say it should potentially raise is that its not a result of poor model behavior. This would only happen if you have classes with zero examples. Not if you model doesn't predict a class.

alykhantejani:
This scenario could occur in streamed datasets however, where you just go a few iterations without seeing one class and in that case if you compute metrics the training would halt. 

We could add a warn logging statement in a similar in the same way we added logging to the early_stopping handler [here](https://github.com/pytorch/ignite/blob/master/ignite/handlers/early_stopping.py#L48)

jasonkriss:
@alykhantejani great point. I'll take care of this in #140

PR #140, opened 2018-04-05
Precision/Recall Fixes

jasonkriss:


tests/ignite/metrics/test_recall.py
```diff
@@ -32,3 +32,16 @@ def test_compute():
...
+
+
+def test_compute_all_wrong():
+    recall = Recall()
+
+    y_pred = torch.FloatTensor([[1.0, 0.0], [1.0, 0.0]])
+    y = torch.ones(2).type(torch.LongTensor)
+    recall.update((y_pred, y))
+    result = list(recall.compute())
+    # assert isnan(result[0])
```
vfdev-5: nit: maybe we can remove commented code ?

vfdev-5:
@jasonkriss what do you think about to add an option to compute the mean in the compute() method : 
```python
self.agg_fn = torch.mean if option else lambda x: x

def compute(self):
    if self._all_positives is None:
        raise NotComputableError('Precision must have at least one example before it can be computed')
    return self.agg_fn(self._true_positives / (self._all_positives + eps))

```

tests/ignite/metrics/test_recall.py
jasonkriss: Good catch! Fixed.

jasonkriss:
@vfdev-5 I'd be ok with that. Would it be better to have it take an optional `agg_fn` like that or just have a `average` or `mean` flag? @alykhantejani any thoughts on this?

alykhantejani:
For precision and recall what other type of reducer functions would you want except for mean?

If there are more than that, then I'm in favor of a constructor arg `reducer` which can be `lambda x : x` by default. If there are no other (common) reducers then just a `compute_average` or `average` kwarg would do?

vfdev-5:
We choose the way of more flexibility and leave the user a choice of a custom function :)
@jasonkriss could you, please, also add `eps` in the division `self._true_positives / (self._all_positives + eps)` to avoid nans ?

jasonkriss:
@vfdev-5 yep I'll add a commit to take care of the nans and warn.

For the reduce argument, I think we should make the API as clear as possible. So unless there are clear use cases for other reductions, I'd prefer the flag. Similar to loss functions in pytorch.

jasonkriss:
@vfdev-5 @alykhantejani made a couple updates if you don't mind taking a peek.

jasonkriss:
Yea I don't think that will make much of a difference performance-wise (even with 1000s of classes).

ignite/metrics/precision.py
```diff
@@ -36,4 +47,11 @@ def update(self, output):
     def compute(self):
         if self._all_positives is None:
             raise NotComputableError('Precision must have at least one example before it can be computed')
-        return self._true_positives / self._all_positives
+        elif self._all_positives.eq(0.0).any():
+            warnings.warn('Labels with no predicted examples are set to have precision of 0.0.', UndefinedMetricWarning)
```
alykhantejani: thought: this might get annoying in the users output. Wdyt?

ignite/metrics/precision.py
```diff
@@ -1,9 +1,10 @@
 from __future__ import division
+import warnings
 
 import torch
 
-from .metric import Metric
-from ignite.exceptions import NotComputableError
+from ignite.metrics.metric import Metric
+from ignite.exceptions import NotComputableError, UndefinedMetricWarning
```
alykhantejani: Do we need to define a custom warning? Could/Should we not just use the generic RuntimeWarning?

jasonkriss: I think the argument here is the same as any custom Exception/Warning. This allows more granular control for the user. They can handle just this specific warning. That being said, I'm fine with using RuntimeWarning if you'd prefer.

ignite/metrics/precision.py
jasonkriss: Yea I can definitely see that. Do you think we maybe should just skip the warning?

alykhantejani: Yeah, or can log it with level WARNING

ignite/metrics/precision.py
alykhantejani: Well I guess if we remove the warning or swap it with a logging line then this is a moot point

ignite/metrics/precision.py
jasonkriss: I went ahead and removed it.

ignite/metrics/recall.py
```diff
@@ -23,9 +30,12 @@ def update(self, output):
         indices = torch.max(y_pred, 1)[1]
         correct = torch.eq(indices, y)
         actual_onehot = to_onehot(y, num_classes)
-        correct_onehot = to_onehot(indices[correct], num_classes)
         actual = actual_onehot.sum(dim=0)
-        true_positives = correct_onehot.sum(dim=0)
+        if correct.sum() == 0:
+            true_positives = torch.zeros(num_classes)
```
vfdev-5: same as for precision

ignite/metrics/precision.py
```diff
@@ -23,9 +30,12 @@ def update(self, output):
         indices = torch.max(y_pred, 1)[1]
         correct = torch.eq(indices, y)
         pred_onehot = to_onehot(indices, num_classes)
-        correct_onehot = to_onehot(indices[correct], num_classes)
         all_positives = pred_onehot.sum(dim=0)
-        true_positives = correct_onehot.sum(dim=0)
+        if correct.sum() == 0:
+            true_positives = torch.zeros(num_classes)
```
vfdev-5: This can fail later at 
```
  File "/usr/local/lib/python3.5/dist-packages/ignite/metrics/precision.py", line 44, in update
    self._true_positives += true_positives
```
if user uses CUDA
The error is 
```
TypeError: add_ received an invalid combination of arguments - got (torch.cuda.FloatTensor), but expected one of:
 * (float value)
      didn't match because some of the arguments have invalid types: (torch.cuda.FloatTensor)
 * (torch.FloatTensor other)
      didn't match because some of the arguments have invalid types: (torch.cuda.FloatTensor)
 * (torch.SparseFloatTensor other)
      didn't match because some of the arguments have invalid types: (torch.cuda.FloatTensor)
 * (float value, torch.FloatTensor other)
 * (float value, torch.SparseFloatTensor other)
```

vfdev-5: Maybe, we could use `true_positives = torch.zeros_like(all_positives)`

ignite/metrics/recall.py
vfdev-5: Maybe, we could use `true_positives = torch.zeros_like(actual)`

jasonkriss:
What would I do with you @vfdev-5 😄. Great catch. This should be fixed.

This also brings up a bigger question. Not specific to this PR, but in general does anyone have ideas on how we can catch these kind of bugs earlier. i.e. is there any way we can we run CI/tests on a GPU?

vfdev-5:
😄 @jasonkriss 

Probably, if you run the tests locally on your machine with GPU it can be catched :)
Otherwise, pytorch does this with jenkins [here](https://ci.pytorch.org/jenkins/job/pytorch-master/) and there are probably some machines with GPU.

alykhantejani:
Thanks @jasonkriss!

Issue #108, opened 2018-03-15
Create notes about concepts in ignite


alykhantejani:
These should be put in a sphinx compatible format like what is [done in pytorch](https://github.com/pytorch/pytorch/tree/master/docs/source/notes)

vfdev-5:
@alykhantejani what do you think to write in these notes? Today we have empty files : 
```
 	concepts.rst
	engines.rst
	exceptions.rst 
	handlers.rst 
	index.rst 
	metrics.rst
	quickstart.rst 
```
In pytorch similar notes cover some particular notions or best practices.

alykhantejani:
I think we need to cover a few things:
1) An overview of the event system
2) Defining your own event handlers
3) An overview of the metrics API

These files were put in as a stub initially, but if you fancy making a start on this feel free to re-arrange/delete as you see fit.

cc @jasonkriss in case he had some thoughts around this

jasonkriss:
Agreed with @alykhantejani. The way I had envisioned this and stubbed it out is to have a "Concepts" note that has the 3 things @alykhantejani mentioned. And then a "Quickstart" guide that just gives a barebones run through of getting a project up and running. The other .rst files will be filled in with `autodoc` but we certainly may want to add some additional context to each one at the top before the actual `autodoc` parts.

That being said, this can definitely change if you wanna get started on it and prefer a different arrangement.

vfdev-5:
I see what you mean @alykhantejani  @jasonkriss, I can start with "Quickstart". 
What do we put in `index.rst` or it will be filled with `autodoc` ?

jasonkriss:
`engines.rst` should still be `engines` because of how we have structured the code. You shouldn't have to put anything in index.rst unless you really want to. You can just add stuff to quickstart.rst. That would be awesome!

vfdev-5:
@jasonkriss you can check the first version of quickstart [here](https://github.com/vfdev-5/ignite/blob/issue_108/docs/source/quickstart.rst) ;-)

jasonkriss:
@vfdev-5 That looks like a great start!

alykhantejani:
I think this is done now @vfdev-5 @jasonkriss? Let me know if you think this should still stay open

vfdev-5:
@alykhantejani following your [comment](https://github.com/pytorch/ignite/issues/108#issuecomment-377281414) there is missing:
3. An overview of the metrics API
[...]

PR #142, opened 2018-04-05
Add quickstart guide and concepts, issue #108

vfdev-5:
I propose to address the issue 108 in multiple PRs. Here is the first one with "Quickstart" notes.
Other notes can be added later. 

@alykhantejani @jasonkriss could you review this note, please ?

### DIFF
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -1,2 +1,108 @@
 Concepts
 ========
+
+Engine
+------
+
+The **essence** of the framework is the class :class:`Engine`, an abstraction that loops given number of times over
+provided data, executes a processing and returns a result:
+
+.. code-block:: python
+
+    while epoch < max_epochs:
+        # run once on data
+        for batch in data:
+            output = process_function(batch)
+
+Thus, a model trainer is simply an engine that loops multiple times over the training dataset and updates model parameters.
+Similarly, model evaluation can be done with an engine that runs a single time over the validation dataset and computes metrics.
+For example, model trainer for a supervised task:
+
+.. code-block:: python
+
+    def update_model(trainer, batch):
+        model.train()
+        optimizer.zero_grad()
+        x, y = prepare_batch(batch)
+        y_pred = model(x)
+        loss = loss_fn(y_pred, y)
+        loss.backward()
+        optimizer.step()
+        return loss.data[0]
+
+    trainer = Engine(update_model)
+    trainer.run(data, max_epochs=100)
+
+Events and Handlers
+-------------------
+
+To improve :class:`Engine` flexibility, an event system is introduced that facilitates interaction on each step of
+the run:
+
+- *engine is started/completed*
+- *epoch is started/completed*
+- *batch iteration is started/completed*
+
+Complete list of events can be found `here <https://github.com/pytorch/ignite/blob/master/ignite/engines/engine.py#L8>`_.
+
+Thus, user can execute a custom code as an event handler. Let us consider in more detail what happens when
+:meth:`Engine.run` is called:
+
+.. code-block:: python
+
+    fire_event(Events.STARTED)
+    while epoch < max_epochs:
+        fire_event(Events.EPOCH_STARTED)
+        # run once on data
+        for batch in data:
+            fire_event(Events.ITERATION_STARTED)
+
+            output = process_function(batch)
+
+            fire_event(Events.ITERATION_COMPLETED)
+        fire_event(Events.EPOCH_COMPLETED)
+    fire_event(Events.COMPLETED)
+
+At first *engine is started* event is fired and all this event handlers are executed (we will see in the next paragraph
+how to add event handlers). Next, `while` loop is started and *epoch is started* event occurs, etc. Every time
+an event is "fired", attached handlers are executed.
+
+Attaching an event handler is simple using method :meth:`Engine.add_event_handler` or :meth:`Engine.on` decorator:
+
+.. code-block:: python
+
+    trainer = Engine(update_model)
+
+    trainer.add_event_handler(Events.STARTED, lambda engine: print("Start training"))
+    # or
+    @trainer.on(Events.STARTED)
+    def on_training_started(engine):
+        print("Another message of start training")
+
+    # attach handler with args, kwargs
+    mydata = [1, 2, 3, 4]
+
+    def on_training_ended(engine, data):
+        print("Training is ended. mydata={}".format(data))
+
+    trainer.add_event_handler(Events.STARTED, on_training_ended, mydata)
+
+
+State
+-----
+A state is introduced in :class:`Engine` to store the output of the `process_function`, current epoch, iteration and other
+helpful information. For example, in case of supervised trainer, we can log computed loss value, completed iterations and
+epochs:
+
+.. code-block:: python
+
+    trainer = Engine(update_model)
+
+    def on_iteration_completed(engine):
+        iteration = engine.state.iteration
+        epoch = engine.state.epoch
+        loss = engine.state.output
+        print("Epoch: {}, Iteration: {}, Loss: {}".format(epoch, iteration, loss))
+
+    trainer.add_event_handler(Events.ITERATION_COMPLETED, on_iteration_completed)
+
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -1,2 +1,157 @@
 Quickstart
 ==========
+
+Welcome to Ignite quickstart guide that just gives essentials of getting a project up and running.
+
+In several lines you can get your model training and validating:
+
+Code
+----
+
+.. code-block:: python
+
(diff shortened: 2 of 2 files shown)


docs/source/quickstart.rst
```diff
@@ -1,2 +1,157 @@
...
+
+    trainer = create_supervised_trainer(model, optimizer, loss)
+    evaluator = create_supervised_evaluator(model,
+                                            metrics={
+                                                'accuracy': CategoricalAccuracy(),
+                                                'nll': Loss(loss)
+                                            })
+
+    @trainer.on(Events.ITERATION_COMPLETED)
+    def log_training_loss(engine):
```
alykhantejani: nit: I would rename this input arg to `trainer` to make it clear its this engine that's passed in

docs/source/quickstart.rst
```diff
@@ -1,2 +1,157 @@
...
+        print("Epoch[{}] Loss: {:.2f}".format(engine.state.epoch, len(train_loader), engine.state.output))
+
+    @trainer.on(Events.EPOCH_COMPLETED)
+    def log_validation_results(engine):
```
alykhantejani: nit: I would rename this input arg to `trainer` to make it clear its this engine that's passed in

docs/source/quickstart.rst
```diff
@@ -1,2 +1,157 @@
...
+
+    model = Net()
+    train_loader, val_loader = get_data_loaders(train_batch_size, val_batch_size)
+    optimizer = torch.optim.SGD(model.parameters(), lr=0.01, momentum=0.8))
+    loss = torch.nn.NLLLoss()
+
+Next we define trainer and evaluator engines. The main component of Ignite is the :class:`Engine`, an abstraction over your
+training loop. Getting started with the engine is easy, the constructor only requires one things:
+
+- `update_function`: a function which is passed a batch and passes data through and updates your model
```
alykhantejani: It's now also passed the engine too

docs/source/quickstart.rst
```diff
@@ -1,2 +1,157 @@
...
+        optimizer.step()
+        return loss.data.cpu()[0]
+
+    def _inference(engine, batch):
+        model.eval()
+        x, y = _prepare_batch(batch, cuda, volatile=True)
+        y_pred = model(x)
+        return to_tensor(y_pred, cpu=not cuda), to_tensor(y, cpu=not cuda)
+
+Remark that the helper function :meth:`create_supervised_evaluator` to create an evaluator accepts an
```
alykhantejani: nit:Remark -> Note

docs/source/quickstart.rst
```diff
@@ -1,2 +1,157 @@
...
+        'accuracy': CategoricalAccuracy(),
+        'nll': Loss(loss)
+    }
+
+where we define two metrics: *categorical accuracy* and *loss* to compute on validation dataset. More information on
+metrics can be found at :doc:`metrics`.
+
+
+The most interesting part of the code snippet is adding event handlers. :class:`Engine` allows to add handlers on
+various events that fired during the run. When event is fired, attached handlers (functions) are executed. Thus, for
```
alykhantejani: When **an** event...

docs/source/quickstart.rst
```diff
@@ -1,2 +1,157 @@
...
+.. code-block:: python
+
+    def log_training_loss(engine):
+        print("Epoch[{}] Loss: {:.2f}".format(engine.state.epoch, len(train_loader), engine.state.output))
+
+    trainer.add_event_handler(Events.ITERATION_COMPLETED, log_training_loss)
+
+When an epoch ends we want to run model validation, therefore we attach another handler to the trainer on epoch complete
+event:
+
```
alykhantejani: nit: could put a note about passing other args to handlers. Or leave this for another doc/ official docstrings for `add_event_handler`

vfdev-5:
@alykhantejani thanks for the comments! I updated the text according to your comments and added a "(!) Note" as [here](http://pytorch.org/tutorials/beginner/transfer_learning_tutorial.html#convnet-as-fixed-feature-extractor) about "passing other args to handlers"

vfdev-5:
Added basic [concepts](https://github.com/vfdev-5/ignite/blob/issue_108/docs/source/concepts.rst)

docs/source/concepts.rst
```diff
@@ -1,2 +1,108 @@
 Concepts
 ========
+
+Engine
+------
+
+The **essence** of the framework is the class :class:`Engine`, an abstraction that loops given number of times over
+provided data, executes a processing and returns a result:
```
alykhantejani: processing **function**

docs/source/concepts.rst
```diff
@@ -1,2 +1,108 @@
 Concepts
 ========
+
+Engine
+------
+
+The **essence** of the framework is the class :class:`Engine`, an abstraction that loops given number of times over
```
alykhantejani: loops **a** given

alykhantejani:
Thanks @vfdev-5!

vfdev-5:
@alykhantejani do you still want me to integrate your comments ?

alykhantejani:
@vfdev-5 if you don't mind, it was very minor though so didn't want to delay merging this. I need to get the docs ready to be hosted and start building the conda tar + wheels for release

vfdev-5:
@alykhantejani no problems, go ahead with your tasks. Thanks!

jasonkriss:
@vfdev-5 Thanks a lot! These both look great. I think the only other thing I would maybe add is a mention in the "State" section of "concepts" about the ability to add arbitrary data to the state object in handlers.

vfdev-5:
@jasonkriss yes, this is an important point that completly missed, thanks! At the begining I thought we would fix the scope of the state, especially when talking about engine checkpointing and restoration. As it is more or less fixed, I can add this in another PR.

vfdev-5:
@jasonkriss rethinking about what you said:
>  the ability to add arbitrary data to the state object in handlers.

Do you talk about this (same with handler functions):
```python
import numpy as np
from ignite.engines import Engine

def update(engine, batch):
    engine.state.new_attribute = 12345

engine = Engine(update)

data = np.arange(100)
data = DataLoader(data, batch_size=10)

for attr, value in state.__dict__.items():
    print(attr, value)
```
with the output:
```
iteration 20
epoch 2
max_epochs 2
batch ...

new_attribute 12345

output None
dataloader <torch.utils.data.dataloader.DataLoader object at 0x7f124b81c518>
metrics {}
```
if yes, we can add a new attribute to `state`, `engine` and whatever object...

I thought also to add the method `__repr__` to the class `State` to visualize easily its variable content:
```python
def __repr__(self):
    s = "State:\n"
    for attr, value in self.__dict__.items():
        value = value if not torch.is_tensor(value) else "{} {}".format(type(value), value.shape)
        s += "\t{}: {}\n".format(attr, value)
    return s
```
@alykhantejani @jasonkriss what do you think about ?

jasonkriss:
Sorry for the delay on this @vfdev-5 . The `__repr__` sounds like a good idea to me. What do you think @alykhantejani.

@vfdev-5 Did you have another question there? I'm not 100% sure I understood.

vfdev-5:
@jasonkriss no problem, my question was actually about your comment:
> the ability to add arbitrary data to the state object in handlers.

Did you  talk about this ? 
```python
import numpy as np
from ignite.engines import Engine

def update(engine, batch):
    engine.state.new_attribute = 12345

engine = Engine(update)

data = np.arange(100)
data = DataLoader(data, batch_size=10)

for attr, value in state.__dict__.items():
    print(attr, value)
```

jasonkriss:
yep exactly. Setting that `new_attribute` is what i was talking about. For me, it usually happens in the handlers but setting it in the `update_fn` is completely valid as well.

vfdev-5:
Right, but actually this is not only ability of `state` but whatever object in python... So, you meant to underline that user can use `state` to add new attributes ?

alykhantejani:
@vfdev-5 yes, I think we want to note that this is the preferred way to pass state between handlers. 

When we get to serializing the engine + state, we may have to re-visit this. But for now I think if we establish this user pattern it will be easier

vfdev-5:
Talking about attaching user attributes in `state`, probably, user would like also to initialize his/her attribute before calling it in handler or update function (same as for `metrics` for example). However, today, the state is `None` before running the engine...

jasonkriss:
I've been using this feature successfully since we added it. If my attribute is a list or a dict, I will initialize it if its `None` the first time the handler is called. If you want to, you could also have an `initialize` handler that you attach to the start event. Anything more than that would be adding too much complexity IMO. I think the goal of the `state` object is to be a flexible, low-cost abstraction for passing around ad hoc information.

vfdev-5:
Sure, for me `State` is something that describes `engine`, so why state is None, when `engine` is initialized and possibly has handlers etc but not yet run... Maybe, we can initialize it on engine creation and complete with `data`, `metrics` etc on `run()`

Issue #139, opened 2018-04-04
[Feature] Add a more comprehensive traceback


vfdev-5:
When developing, it would really helpful to have some more info when something goes wrong with a fired handler. For example, see this traceback
```
Traceback (most recent call last):
  File "cifar10_train_playground.py", line 284, in <module>
    args.lr, args.gamma, args.log_interval, args.output, args.debug)
  File "cifar10_train_playground.py", line 254, in run
    trainer.run(train_loader, max_epochs=epochs)
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 150, in run
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 116, in _handle_exception
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 141, in run
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 85, in _fire_event
  File "cifar10_train_playground.py", line 213, in log_validation_results
    evaluator.run(val_loader)
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 150, in run
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 116, in _handle_exception
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 143, in run
  File "/usr/local/lib/python3.5/dist-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 85, in _fire_event
TypeError: __call__() takes 2 positional arguments but 3 were given
```
which is related with the fact that a handler signature is different from the assigned by user:
```python
    handler = EarlyStopping(patience=10, score_function=score_function, trainer=trainer)
    evaluator.add_event_handler(Events.COMPLETED, handler, TRAINER) # No more trainer in the API
```

Maybe, when adding an event handler we could try to execute it and check somehow the signature?

vfdev-5:
@alykhantejani I think we can close this one too

PR #148, opened 2018-04-22
[issue #139] Check event handler's signature before registering

sean-adler:
Addresses #139.

This adds a check to `Engine.add_event_handler()` that the handler's signature matches what it will actually be called with inside `_fire_event()`.

Adding a handler like this:
```
def handler(engine):
    pass

engine.add_event_handler(Events.STARTED, handler, 42)
```

Results in a traceback like this:
```
ValueError: Error adding handler '<function handler at 0x105323950>': takes parameters ['engine'] but will be called with [<ignite.engines.engine.Engine object at 0x107a0a8d0>, 42] (too many positional arguments)
```

Open to any and all feedback!

### DIFF
--- a/ignite/engines/engine.py
+++ b/ignite/engines/engine.py
@@ -1,5 +1,6 @@
 import inspect
 import logging
+import sys
 import time
 from enum import Enum
 
@@ -58,21 +59,39 @@ class Engine(object):
             self._logger.error("attempt to add event handler to an invalid event %s ", event_name)
             raise ValueError("Event {} is not a valid event for this Engine".format(event_name))
 
-        callable_ = handler if hasattr(handler, '__name__') else handler.__call__
-        try:
-            inspect.getcallargs(callable_, self, *args, **kwargs)
-        except TypeError as exc:
-            spec = inspect.getargspec(callable_)
-            raise ValueError("Error adding handler '{}': "
-                             "takes parameters {} but will be called with {} "
-                             "({})".format(
-                                 handler, spec.args, [self] + list(args) + list(kwargs), exc))
+        self._check_handler_signature(handler, *args, **kwargs)
 
         if event_name not in self._event_handlers:
             self._event_handlers[event_name] = []
 
         self._event_handlers[event_name].append((handler, args, kwargs))
         self._logger.debug("added handler for event %s ", event_name)
+
+    def _check_handler_signature(self, handler, *args, **kwargs):
+        exception_msg = None
+
+        if sys.version_info[0] < 3:
+            try:
+                callable_ = handler if hasattr(handler, '__name__') else handler.__call__
+                inspect.getcallargs(callable_, self, *args, **kwargs)
+            except TypeError as exc:
+                spec = inspect.getargspec(callable_)
+                handler_params = list(spec.args)
+                exception_msg = str(exc)
+        else:
+            signature = inspect.signature(handler)
+            try:
+                signature.bind(self, *args, **kwargs)
+            except TypeError as exc:
+                handler_params = list(signature.parameters)
+                exception_msg = str(exc)
+
+        if exception_msg:
+            passed_params = [self] + list(args) + list(kwargs)
+            raise ValueError("Error adding handler '{}': "
+                             "takes parameters {} but will be called with {} "
+                             "({})".format(
+                                 handler, handler_params, passed_params, exception_msg))
 
     def on(self, event_name, *args, **kwargs):
         """Decorator shortcut for add_event_handler
--- a/ignite/engines/engine.py
+++ b/ignite/engines/engine.py
@@ -63,8 +63,8 @@ class Engine(object):
         if self._process_function is None:
             raise ValueError("Engine must be given a processing function in order to run")
 
-        BATCH_PLACEHOLDER = None
-        self._check_signature(process_function, 'process_function', BATCH_PLACEHOLDER)
+        batch_placeholder = None
+        self._check_signature(process_function, 'process_function', batch_placeholder)
 
     def add_event_handler(self, event_name, handler, *args, **kwargs):
         """Add an event handler to be executed when the specified event is fired
--- a/ignite/engines/engine.py
+++ b/ignite/engines/engine.py
@@ -70,6 +70,21 @@ class Engine(object):
             *args: optional args to be passed to `handler`
             **kwargs: optional keyword args to be passed to `handler`
 
+        Notes:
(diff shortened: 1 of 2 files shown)


vfdev-5:
@sean-adler nice PR, thank you! @alykhantejani what do you think ?
Personnally, never used `inspect` and look the [documentation about `getcallargs`](https://docs.python.org/3/library/inspect.html#inspect.getcallargs) they say that the method  
> Deprecated since version 3.5: Use Signature.bind() and Signature.bind_partial() instead.

May be we need to use signature instead to avoid warnings.

sean-adler:
> May be we need to use signature instead to avoid warnings.

Thanks for catching that @vfdev-5 - I was originally using `inspect.signature()`, but it's not in Python 2.7, which caused [this test error](https://travis-ci.org/pytorch/ignite/jobs/369645260).

I switched to `inspect.getcallargs` in a5f0b0c which makes the CI pass, but I agree that using a deprecated method isn't great.

I could switch between the methods based on `sys.version_info`, if that seems alright?

vfdev-5:
@sean-adler I see your point. Yes, probably, we can intoduce a method that defines something common using `signature` for >=3.5 and `getcallargs` for others and than call it in `add_handler`. 
Another possibility, [here] (https://stackoverflow.com/questions/45946051/signature-method-in-inspect-module-for-python-2) they say about signature port to 2.7, what do you think?

sean-adler:
> I agree with @vfdev-5's approach of having a function that is called from `add_event_handler` which can do one of the other method based on `sys.version_info`.
> 
> I think we should avoid using `funcsigs` as it adds another dependency

Sounds good, thanks @alykhantejani.

ignite/engines/engine.py
```diff
@@ -57,12 +59,40 @@ def add_event_handler(self, event_name, handler, *args, **kwargs):
...
         if event_name not in self._event_handlers:
             self._event_handlers[event_name] = []
 
         self._event_handlers[event_name].append((handler, args, kwargs))
         self._logger.debug("added handler for event %s ", event_name)
 
+    def _check_handler_signature(self, handler, *args, **kwargs):
+        exception_msg = None
+
+        if sys.version_info[0] < 3:
```
vfdev-5: @sean-adler  what do you think about checking version globally and once instead of every time in `add_handler`, something like this: 
```python

if sys.version_info[0] < 3:
    
    def _check_handler_signature(self, handler, *args, **kwargs):
        exception_msg = None
        try:
            callable_ = handler if hasattr(handler, '__name__') else handler.__call__
            inspect.getcallargs(callable_, self, *args, **kwargs)
        except TypeError as exc:
            spec = inspect.getargspec(callable_)
            handler_params = list(spec.args)
            exception_msg = str(exc)
            
        if exception_msg:
            passed_params = [self] + list(args) + list(kwargs)
            raise ValueError("Error adding handler '{}': "
                             "takes parameters {} but will be called with {} "
                             "({})".format(
                                 handler, handler_params, passed_params, exception_msg))        
else:
    
    def _check_handler_signature(self, handler, *args, **kwargs):
        exception_msg = None
        signature = inspect.signature(handler)
        try:
            signature.bind(self, *args, **kwargs)
        except TypeError as exc:
            handler_params = list(signature.parameters)
            exception_msg = str(exc)

        if exception_msg:
            passed_params = [self] + list(args) + list(kwargs)
            raise ValueError("Error adding handler '{}': "
                             "takes parameters {} but will be called with {} "
                             "({})".format(
                                 handler, handler_params, passed_params, exception_msg))
```
but in a more compact way if possible :)

alykhantejani: The conditional is cheap to evaluate on every handler addition, I'm not sure it's worth complicating the code for?

alykhantejani:
LGTM. @vfdev-5 I'll wait for your review too before merging

ignite/engines/engine.py
vfdev-5: @alykhantejani I see your point, yes it is cheap to evaluate, but just esthetically I would tend to make a choice once...

A compacted version without `if` in `add_handler` as I see is like :
```python
# Define the following outside Engine
if sys.version_info[0] < 3:
    def _check_signature(handler, engine, *args, **kwargs):
        exception_msg = None
        callable_ = handler if hasattr(handler, '__name__') else handler.__call__
        try:
            inspect.getcallargs(callable_, engine, *args, **kwargs)
        except TypeError as exc:
            exception_msg = str(exc)
        return inspect.getargspec(callable_), exception_msg
else:
    def _check_signature(handler, engine, *args, **kwargs):
        exception_msg = None
        signature = inspect.signature(handler)
        try:
            signature.bind(engine, *args, **kwargs)
        except TypeError as exc:
            exception_msg = str(exc)
        return list(signature.parameters), exception_msg


# A method in Engine
def check_handler_signature(engine, handler, *args, **kwargs):
    handler_params, exception_msg = _check_signature(handler, engine, *args, **kwargs)
    if exception_msg:
        passed_params = [engine] + list(args) + list(kwargs)
        raise ValueError("Error adding handler '{}': "
                         "takes parameters {} but will be called with {} "
                         "({})".format(handler, handler_params, passed_params, exception_msg))
```
Any comments, feedbask ?

alykhantejani: both approaches are fine with me tbh :)

sean-adler: Thanks @alykhantejani and @vfdev-5 - I took the middle-ground and moved this conditional into a constant 😄

vfdev-5:
@sean-adler okay let's take this "median" solution :) 

@alykhantejani  so, to fix failing tests we need to fix pytorch version (addressed in #156) and restart CI.

vfdev-5:
@sean-adler great job! We are almost done :)
Could you add, please, the signature check on update function too, to avoid constructing a trainer that will fail on run.

sean-adler:
> Could you add, please, the signature check on update function too, to avoid constructing a trainer that will fail on run.

Good idea, added in 42a6429.

ignite/engines/engine.py
```diff
@@ -43,6 +63,9 @@ def __init__(self, process_function):
         if self._process_function is None:
             raise ValueError("Engine must be given a processing function in order to run")
 
+        BATCH_PLACEHOLDER = None
```
vfdev-5: nit: maybe lower case ?

vfdev-5:
Anyway, @sean-adler this looks great to me !
@alykhantejani if it is OK to you too, we can merge this

alykhantejani:
Thanks @sean-adler!

Issue #154, opened 2018-04-24
[Feature] PyTorch 0.4.0 Migration 


vfdev-5:
👍 🎉  [PyTorch 0.4.0](https://github.com/pytorch/pytorch/releases/tag/v0.4.0) is arrived ! 🎉 👍 
and we will need to clean-up the code according to 

**Major Changes and Potentially Breaking Changes:**

- Tensors and Variables have merged
- Some operations now return 0-dimensional (scalar) Tensors
- Deprecation of the volatile flag

Can we do this after the very first release 0.1 ?

vfdev-5:
And incompatibilities already come to the party :) 
```
>       assert model.weight.data[0, 0] == approx(0.0)
E       TypeError: eq() received an invalid combination of arguments - got (ApproxScalar), but expected one of:
E        * (Tensor other)
E             didn't match because some of the arguments have invalid types: (!ApproxScalar!)
E        * (float other)
E             didn't match because some of the arguments have invalid types: (!ApproxScalar!)
```
and 
```
>   ???
E   AssertionError: `loss_fn` did not return the average loss
```
See [here](https://travis-ci.org/pytorch/ignite/jobs/370846036) for example

alykhantejani:
hmm - perhaps we should do this before 0.1 especially if it has breaking changes

PR #157, opened 2018-04-25
PyTorch 0.4.0 migration

sean-adler:
Addresses #154.

I _think_ I hit everything necessary, but it's totally possible that I missed something. (I'm also not 100% sure yet of the use of `torch.set_grad_enabled` to replace `volatile`.) In any case, hopefully this will be a helpful start.

One really pernicious thing to note is that `Metric.compute() == <some_value>` will always be true unless that method extracts the scalar value from the new `torch.tensor` scalar. In other words, because `Tensor.data.__getindex__` returns a tensor now, `torch.Tensor.__eq__` will return another tensor like this:

```
In [1]: import torch

In [2]: not_a_float = torch.Tensor([0.0]).data[0]

In [3]: not_a_float == 0.5
Out[3]: tensor(0, dtype=torch.uint8)
```

Which is truthy in the unit tests. Kinda tricky! I added `assert isinstance(<val>, float)` checks to try and catch that (in addition to using `.item()` in the metric methods themselves), but maybe there is a better way.

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -89,7 +89,7 @@ And update functions of the trainer and evaluator are simply:
     def _update(engine, batch):
         model.train()
         optimizer.zero_grad()
-        x, y = _prepare_batch(batch, requires_grad=False, device=device)
+        x, y = _prepare_batch(batch, device=device)
         y_pred = model(x)
         loss = loss_fn(y_pred, y)
         loss.backward()
@@ -98,7 +98,7 @@ And update functions of the trainer and evaluator are simply:
 
     def _inference(engine, batch):
         model.eval()
-        x, y = _prepare_batch(batch, requires_grad=False, device=device)
+        x, y = _prepare_batch(batch, device=device)
         y_pred = model(x)
         return y_pred, y
 
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -89,18 +89,18 @@ And update functions of the trainer and evaluator are simply:
     def _update(engine, batch):
         model.train()
         optimizer.zero_grad()
-        x, y = _prepare_batch(batch, cuda)
+        x, y = _prepare_batch(batch, requires_grad=False, device=device)
         y_pred = model(x)
         loss = loss_fn(y_pred, y)
         loss.backward()
         optimizer.step()
-        return loss.data.cpu()[0]
+        return loss.item()
 
     def _inference(engine, batch):
         model.eval()
-        x, y = _prepare_batch(batch, cuda, volatile=True)
+        x, y = _prepare_batch(batch, requires_grad=False, device=device)
         y_pred = model(x)
-        return to_tensor(y_pred, cpu=not cuda), to_tensor(y, cpu=not cuda)
+        return y_pred, y
 
 Note that the helper function :meth:`create_supervised_evaluator` to create an evaluator accepts an
 argument `metrics`:
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -44,12 +44,11 @@ def get_data_loaders(train_batch_size, val_batch_size):
 
 
 def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
-    cuda = torch.cuda.is_available()
     train_loader, val_loader = get_data_loaders(train_batch_size, val_batch_size)
-
     model = Net()
-    device = None
-    if cuda:
+    device = 'cpu'
+
+    if torch.cuda.is_available():
         device = 'cuda'
         model = model.to(device)
 
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -48,14 +48,17 @@ def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
     train_loader, val_loader = get_data_loaders(train_batch_size, val_batch_size)
 
     model = Net()
+    device = None
     if cuda:
-        model = model.cuda()
+        device = 'cuda'
+        model = model.to(device)
+
     optimizer = SGD(model.parameters(), lr=lr, momentum=momentum)
-    trainer = create_supervised_trainer(model, optimizer, F.nll_loss, cuda=cuda)
+    trainer = create_supervised_trainer(model, optimizer, F.nll_loss, device=device)
     evaluator = create_supervised_evaluator(model,
                                             metrics={'accuracy': CategoricalAccuracy(),
                                                      'nll': Loss(F.nll_loss)},
-                                            cuda=cuda)
+                                            device=device)
 
     @trainer.on(Events.ITERATION_COMPLETED)
     def log_training_loss(engine):
--- a/ignite/_utils.py
+++ b/ignite/_utils.py
@@ -15,7 +15,8 @@ def to_variable(batch, cuda=False, volatile=False):
     if torch.is_tensor(batch):
         if cuda:
             batch = batch.cuda()
-        return Variable(batch, volatile=volatile)
+        with torch.set_grad_enabled(not volatile):
+            return batch
     elif isinstance(batch, string_classes):
         return batch
     elif isinstance(batch, collections.Mapping):
--- a/ignite/_utils.py
+++ b/ignite/_utils.py
@@ -20,7 +20,5 @@ def convert_tensor(tensor, requires_grad=False, device=None):
 
 
 def to_onehot(indices, num_classes):
-    onehot = torch.zeros(indices.size(0), num_classes)
-    if indices.is_cuda:
-        onehot = onehot.to('cuda')
+    onehot = torch.zeros(indices.size(0), num_classes, device=indices.device)
     return onehot.scatter_(1, indices.unsqueeze(1), 1)
--- a/ignite/_utils.py
+++ b/ignite/_utils.py
@@ -11,19 +11,17 @@ def _to_hours_mins_secs(time_taken):
     return hours, mins, secs
 
 
-def convert_tensor(input_, requires_grad=False, device=None):
+def convert_tensor(input_, device=None):
     if torch.is_tensor(input_):
         if device:
             input_ = input_.to(device=device)
-        if requires_grad != input_.requires_grad:
-            input_.requires_grad_(requires_grad)
         return input_
     elif isinstance(input_, string_classes):
         return input_
     elif isinstance(input_, collections.Mapping):
-        return {k: convert_tensor(sample, requires_grad=requires_grad, device=device) for k, sample in input_.items()}
+        return {k: convert_tensor(sample, device=device) for k, sample in input_.items()}
     elif isinstance(input_, collections.Sequence):
-        return [convert_tensor(sample, requires_grad=requires_grad, device=device) for sample in input_]
+        return [convert_tensor(sample, device=device) for sample in input_]
     else:
         raise TypeError(("input must contain tensors, numbers, dicts or lists; found {}"
                          .format(type(input_[0]))))
--- a/ignite/_utils.py
+++ b/ignite/_utils.py
@@ -11,35 +11,19 @@ def _to_hours_mins_secs(time_taken):
     return hours, mins, secs
 
 
-def to_variable(batch, cuda=False, volatile=False):
-    if torch.is_tensor(batch):
+def to_tensor(variable, cuda=False, requires_grad=False):
+    if torch.is_tensor(variable):
         if cuda:
-            batch = batch.cuda()
-        with torch.set_grad_enabled(not volatile):
-            return batch
-    elif isinstance(batch, string_classes):
-        return batch
-    elif isinstance(batch, collections.Mapping):
-        return {k: to_variable(sample, cuda=cuda, volatile=volatile) for k, sample in batch.items()}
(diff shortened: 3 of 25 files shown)


vfdev-5:
@sean-adler maybe we should adopt the merge of Variable/Tensor and merge our `to_variable`, `to_tensor` methods :)

ignite/engines/__init__.py
```diff
@@ -30,7 +30,7 @@ def _update(engine, batch):
         loss = loss_fn(y_pred, y)
         loss.backward()
         optimizer.step()
-        return loss.data.cpu()[0]
+        return loss.data.cpu().item()
```
fmassa: nit: this can be simplified as
```
loss.item()
```

sean-adler:
> @sean-adler maybe we should adopt the merge of Variable/Tensor and merge our `to_variable`, `to_tensor` methods :)

Good idea! I took a stab at combining them, mostly just by adding a check in `to_tensor()` for what was previously `volatile=True` in `to_variable()` (and adding a short test assertion for that case).

I'm still not sure that calling `.detach()` is correct in that context, or whether we should simply attempt to call `.grad_required_()`, and bubble up any exceptions that might be raised by doing that. I think it depends on how you want `to_tensor()` to be used, but please let me know what you think @vfdev-5.

(Also - I ran the full [test script](https://github.com/sean-adler/ignite/blob/95c85b9813ccec3dda134e5ce426e386f68024b0/.travis.yml#L24-L35) locally and it passed, so I think the CI failures are related to Travis installing PyTorch 0.3.1 in the install script.)

ignite/engines/__init__.py
sean-adler: Thanks!

ignite/_utils.py
```diff
@@ -11,34 +11,19 @@ def _to_hours_mins_secs(time_taken):
...
-                         .format(type(batch[0]))))
-
-
-def to_tensor(variable, cpu=False):
-    if isinstance(variable, Variable):
-        t = variable.data
-        if cpu:
-            t = t.cpu()
-        return t
+            variable = variable.cuda()
```
igormq: Wouldn't be better passing the device instead of cuda arg. in order to maintain this code agnostic to device? Then, this part of the code could be
```python
variable = variable.to(device)
```

ignite/engines/__init__.py
```diff
@@ -49,9 +49,9 @@ def create_supervised_evaluator(model, metrics={}, cuda=False):
     """
     def _inference(engine, batch):
         model.eval()
```
igormq: This code could be a little different using no_grad function and the fact that Variable=Tensor
```
with torch.no_grad():
    x, y = _prepare_batch(batch, device)
    y_pred = model(x)
    return y_pred
```

vfdev-5: I think we should keep anyway `model.eval()`

vfdev-5:
@igormq yes, it's a good point, probably, we need to replace all `cuda` variables by `device` and `.cuda()` by `.to(device)`.

ignite/engines/__init__.py
igormq: Yes, otherwise it will break some ops (e.g., BatchNorm, and Dropout) in inference mode.

examples/mnist.py
```diff
@@ -48,14 +48,17 @@ def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
     train_loader, val_loader = get_data_loaders(train_batch_size, val_batch_size)
 
     model = Net()
+    device = None
```
vfdev-5: maybe `device='cpu'`

ignite/engines/__init__.py
```diff
@@ -49,9 +48,9 @@ def create_supervised_evaluator(model, metrics={}, cuda=False):
...
-        x, y = _prepare_batch(batch, cuda, volatile=True)
+        x, y = _prepare_batch(batch, requires_grad=False, device=device)
         y_pred = model(x)
-        return to_tensor(y_pred, cpu=not cuda), to_tensor(y, cpu=not cuda)
+        return (y_pred, y)
```
vfdev-5: nit: redundant parenthesis (, )

vfdev-5:
> I think the issue is that Travis runs whatever config is set in master, cd30b15 wasn't pulled into this branch.

You need to modify `.travis.yml` (remove `==0.3.1`), remember the case when 0.4 version become available and see this #156 where the build passes fixing the version to 0.3.1

ignite/_utils.py
```diff
@@ -11,41 +11,16 @@ def _to_hours_mins_secs(time_taken):
...
+    if requires_grad != tensor.requires_grad:
+        tensor.requires_grad_(requires_grad)
+    return tensor
 
 
 def to_onehot(indices, num_classes):
     onehot = torch.zeros(indices.size(0), num_classes)
     if indices.is_cuda:
-        onehot = onehot.cuda()
+        onehot = onehot.to('cuda')
```
vfdev-5: @sean-adler here we maybe need to take the device of `indices`, not just any 'cuda' ? What do you think ?

docs/source/quickstart.rst
```diff
@@ -89,18 +89,18 @@ And update functions of the trainer and evaluator are simply:
     def _update(engine, batch):
         model.train()
         optimizer.zero_grad()
-        x, y = _prepare_batch(batch, cuda)
+        x, y = _prepare_batch(batch, requires_grad=False, device=device)
```
vfdev-5: @sean-adler here should not be `requires_grad=True` ?

sean-adler: I think the batch shouldn't require gradient tracking - if it did then `optimizer.step()` would update `x` and `y`, and if I make the following change:

```diff
    def _update(engine, batch):
        model.train()
        optimizer.zero_grad()
-       x, y = _prepare_batch(batch, requires_grad=False, device=device)
+       x, y = _prepare_batch(batch, requires_grad=True, device=device)
        loss = loss_fn(y_pred, y)
        loss.backward()
        optimizer.step()
        return loss.item()
```

Running `python examples/mnist.py --epochs=1` in my local environment raises this error:

```
  File "examples/mnist.py", line 98, in <module>
    run(args.batch_size, args.val_batch_size, args.epochs, args.lr, args.momentum, args.log_interval)
  File "examples/mnist.py", line 78, in run
    trainer.run(train_loader, max_epochs=epochs)
  File "/Users/sean/anaconda3/envs/test-environment/lib/python3.5/site-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 151, in run
  File "/Users/sean/anaconda3/envs/test-environment/lib/python3.5/site-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 117, in _handle_exception
  File "/Users/sean/anaconda3/envs/test-environment/lib/python3.5/site-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 138, in run
  File "/Users/sean/anaconda3/envs/test-environment/lib/python3.5/site-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 111, in _run_once_on_dataset
  File "/Users/sean/anaconda3/envs/test-environment/lib/python3.5/site-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 117, in _handle_exception
  File "/Users/sean/anaconda3/envs/test-environment/lib/python3.5/site-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/engine.py", line 101, in _run_once_on_dataset
  File "/Users/sean/anaconda3/envs/test-environment/lib/python3.5/site-packages/ignite-0.1.0a1-py3.5.egg/ignite/engines/__init__.py", line 34, in _update
  File "/Users/sean/anaconda3/envs/test-environment/lib/python3.5/site-packages/torch/tensor.py", line 93, in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph)
  File "/Users/sean/anaconda3/envs/test-environment/lib/python3.5/site-packages/torch/autograd/__init__.py", line 89, in backward
    allow_unreachable=True)  # allow_unreachable flag
RuntimeError: No grad accumulator for a saved leaf!
```

I'm pretty sure the issue is that `y` is a `torch.LongTensor`, and you can't backprop to that data type in 0.4.0. Seems like it's the same issue as https://github.com/pytorch/pytorch/issues/7021 (which is not a bug).

ignite/_utils.py
sean-adler: Good point, thanks!

docs/source/quickstart.rst
vfdev-5: @sean-adler I agree, so should not we just change `_prepare_batch` as :
```
def _prepare_batch(batch, requires_grad=False, device=None):
    x, y = batch
    return convert_tensor(x, requires_grad=requires_grad, device=device), 
               convert_tensor(y, requires_grad=False, device=device)
```
what do you think ?

vfdev-5:
There is something weird with your merge master commit as it shows as your modifications already merged file. Maybe we need to revert this ?

sean-adler:
> There is something weird with your merge master commit. It shows as your modifications for some of already merged files. Maybe we need to revert this ?

Sorry about that - I merged upstream changes incorrectly (fetched + merged them directly into this branch, instead of syncing `upstream/master` with my local `master` first) so that messed up the history. Will try to fix 😳

docs/source/quickstart.rst
sean-adler: I don't see why `x` would need to be updated either, but I could definitely be missing something - I'm happy to make the change if you think it's worth it.

vfdev-5: I checked an example  in 0.3.1 with `Variable` etc and it seems, you're right, we do not need `requires_grad` to be true for `x` neither. So we can simply remove it from the signature of 
`def _prepare_batch(batch, requires_grad=False, device=None)`...

ignite/_utils.py
```diff
@@ -11,41 +11,14 @@ def _to_hours_mins_secs(time_taken):
...
-        return variable
-    elif isinstance(variable, collections.Mapping):
-        return {k: to_tensor(sample, cpu=cpu) for k, sample in variable.items()}
-    elif isinstance(variable, collections.Sequence):
-        return [to_tensor(sample, cpu=cpu) for sample in variable]
-    else:
-        raise TypeError(("Argument variable must contain torch.autograd.Variable, numbers, dicts or lists; found {}"
-                         .format(type(variable))))
+def convert_tensor(tensor, requires_grad=False, device=None):
+    if device:
```
vfdev-5: @sean-adler just remarked that you removed the multiple `if-elif-else` tree as it was in `to_variable`. I think we should keep it. Otherwise you implicitly imposes that batch is composed of only tensors and not list of tensors or a dict of tensors etc.

vfdev-5:
This PR becomes very large, @sean-adler did a huge job! Anyway, it looks good to me and I think we can merge it and fix the rest if any other small problems appear with another PRs. 
@alykhantejani what do you think ?

ignite/_utils.py
```diff
@@ -11,41 +11,24 @@ def _to_hours_mins_secs(time_taken):
...
-        return variable
-    elif isinstance(variable, collections.Mapping):
-        return {k: to_tensor(sample, cpu=cpu) for k, sample in variable.items()}
-    elif isinstance(variable, collections.Sequence):
-        return [to_tensor(sample, cpu=cpu) for sample in variable]
-    else:
-        raise TypeError(("Argument variable must contain torch.autograd.Variable, numbers, dicts or lists; found {}"
-                         .format(type(variable))))
+        raise TypeError(("input must contain tensors, numbers, dicts or lists; found {}"
+                         .format(type(input_[0]))))
```
vfdev-5: Maybe  should be `input_[0]` -> `input_`, otherwise: 
```python
convert_tensor(12345)
```
fails with 
```
TypeError: 'int' object is not subscriptable
```

vfdev-5:
another nit (cannot make a comment on line): @sean-adler you can remove `from torch.autograd import Variable`  from [here](https://github.com/sean-adler/ignite/blob/2e8157f5ab533ea3c3fe98d9ee911da60489f698/ignite/_utils.py#L5)

sean-adler:
> I have one nit thought, why do we have the `requires_grad=False` param in `convert_tensor`, it doesn't seem to be used anywhere (except the test).

Good point @alykhantejani, I think it made sense initially but we ended up whittling `_prepare_batch()` down so the kwarg isn't needed anymore. Removed in 2c1cf81

vfdev-5:
> I have one nit thought, why do we have the requires_grad=False param in convert_tensor, it doesn't seem to be used anywhere (except the test).

@alykhantejani this option actually can be used in evaluator update function if user does not want to use `with torch.no_grad()`. Two similar codes:
```python
def _inference(engine, batch):
    model.eval()
    with torch.no_grad():
        x, y = _prepare_batch(batch, device=device)
        y_pred = model(x)
        return y_pred, y
    
def _inference2(engine, batch):
    model.eval()
    x, y = _prepare_batch(batch, requires_grad=False, device=device)
    y_pred = model(x)
    return y_pred, y
```

alykhantejani:
Thanks @sean-adler!

Issue #160, opened 2018-04-30
torch.save(obj) instead of torch.save(obj.state_dict())


akurniawan:
Hi guys,

As per my understanding and from [notes](https://github.com/pytorch/pytorch/blob/master/docs/source/notes/serialization.rst) page in PyTorch itself, it is recommended to create a checkpoint based on models state_dict and not saving the entire object as currently written in here
https://github.com/pytorch/ignite/blob/cd30b15db32dfa92c0469893fcf2ce889e9c4389/ignite/handlers/checkpoint.py#L120

Is there any particular discussion of why we save the model this way and not using `state_dict`?
Would love to contribute if you guys agree for this changes :)

vfdev-5:
@akurniawan I'm not sure if there was a discussion on what to save `model` or `state_dict`. [Here](https://github.com/pytorch/ignite/issues/62#issuecomment-363791423) is the comment of @elanmart who implemented `ModelCheckpoint` where he stores the model. Maybe we would like to have both ways if you want to contribute. 
cc @alykhantejani

elanmart:
Pesonally I always use `torch.save`, for two reasons:
* it is easier to load models -- you just call `th.load(path)`, instead of instantiating an empty model, and then copying the loaded state dict,
* usually I want to save the whole python object anyway, with all the attached information, so I can't use `save_dict`. 

But I agree that perhaps having a flag for this would be nice. I'd definatelly vote against changing this to using `state_dict` only.

PR #161, opened 2018-05-01
Adding flag for ModelCheckpoint to save model as state_dict

akurniawan:
Trying to address https://github.com/pytorch/ignite/issues/160 by adding flag to store the model as state_dict and keeping the old method intact.
I have never used pytest before so I don't know whether creating the dummy model is the right way, please let me know if it's not following any standard :)


ignite/handlers/checkpoint.py
```diff
@@ -87,53 +93,62 @@ def __init__(self, dirname, filename_prefix,
...
-
             else:
                 tmp.close()
                 os.rename(tmp.name, path)
 
+    def _internal_save(self, obj, path):
+        if not self._save_as_state_dict:
+            torch.save(obj, path)
+        else:
+            torch.save(obj.state_dict(), path)
```
vfdev-5: Maybe, we need to check if `obj` has `state_dict` function ?

vfdev-5:
@akurniawan thanks, looks good. Just wondering why there are many changes with line breakings only?

akurniawan:
@vfdev-5 I forgot to add new line in test file every time I write new test case 😅, sorry for that

vfdev-5:
@akurniawan okay, actually I meant the line breakings like that : 
```python
-            raise ValueError("Exactly one of `save_interval`, or `score_function` "
-                             "arguments must be provided.")
+                "Exactly one of `save_interval`, or `score_function` "
+                "arguments must be provided.")
```
There is a bunch of modifications like that... Maybe we can keep the previous line width as it has already passed lint checks. What do you think ?

akurniawan:
@vfdev-5 ah I see, that’s the linter on my vscode trying to prettify the code. Will fix that

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -133,33 +161,122 @@ def test_best_k_with_suffix(dirname):
...
+        save_as_state_dict=True)
+
+    engine.add_event_handler(Events.EPOCH_COMPLETED, handler, {name: model})
+    engine.run([0], max_epochs=4)
+
+    saved_model = os.path.join(dirname, os.listdir(dirname)[0])
+    load_model = torch.load(saved_model)
+
+    assert not isinstance(load_model, DummyModel)
+    assert isinstance(load_model, dict)
```
alykhantejani: can we also check the loaded state dict matches the model.state_dict()

alykhantejani:
LGTM if we fix revert the line break changes. Also can you add an additional check in the test that the loaded state_dict is the same as the models, then its good to merge. 

Thanks for the PR!

alykhantejani:
Thanks @akurniawan!

Issue #118, opened 2018-03-19
Metrics with multiple outputs & standardizing outputs


veugene:
The current implementation of metrics assumes a very specific set of return values from the training/validation function: (y_pred, y), as can be seen for example in https://github.com/pytorch/ignite/blob/master/ignite/metrics/mean_absolute_error.py#L20 .

This has a couple inconvenient consequences:

1. Existing metrics cannot handle a model with multiple outputs.
2. Users cannot pass around additional return values to be picked up by a custom handler via `state.outputs`

It would be reasonable to standardize the set of return values expected from training/validation functions if the user would have the option of putting custom values in the calling engine's `state`. This would require exposing state to the training/validation functions. This would solve (2) but metrics would need to be further adjusted to solve (1).

One way of standardizing the return values would be to always return only loss, model outputs, and targets. If state is ephemeral but is passed to the training/validation function, this would be of the form:
`return loss, model_outputs, targets, state`

If state were to be exposed in an Engine as described in #117, this would be of the form:
`return loss, model_outputs, targets`

Every element in `model_outputs` would be paired with each element in the targets. This would allow a Metric to optionally track a single output in a multi-output model, changing

```python
class Metric(object):
    def __init__(self):
        self.reset()

def iteration_completed(self, engine, state):
        self.update(state.output)
```

to

```python
class Metric(object):
    def __init__(self, output_index=None):
        self.output_index = output_index
        self.reset()

    def iteration_completed(self, engine, state):
        output = state.output
        target = state.target
        if self.output_index is not None:
            output = output[self.output_index]
            target = target[self.output_index]
        self.update(output, target)
```

vfdev-5:
I would say that in this special case of computing a metric on a particular index of the output/target, you can write a custom class, sort of proxy metric, `IndexedMetric` (find a better name):
```python


class ProxyMetric(Metric):

  def __init__(self, metric):
      self.metric = metric    
      super(IndexedMetric, self).__init__()

  def reset(self):
      self.metric.reset()

  def compute(self):
      return self.metric.compute()


class IndexedMetric(ProxyMetric):
    
    def __init__(self, metric, index):
        self.index = index
        super(IndexedMetric, self).__init__(metric)
    
    def update(self, output):
        y_pred, y = output
        y_pred = y_pred[self.index]
        y = y[self.index]
        self.metric.update((y_pred, y))           
```

HTH

veugene:
[...]

PR #165, opened 2018-05-03
Trainer with metrics

vfdev-5:
A proposition to update `create_supervised_trainer` to accept metrics similarly to `create_supervised_evaluator`. A log can be updated as the following
```
Epoch[1] Iteration[910/938] Loss: 0.48
Epoch[1] Iteration[920/938] Loss: 0.56
Epoch[1] Iteration[930/938] Loss: 0.34
Training Results - Epoch: 1  Avg accuracy: 0.67 Avg loss: 0.97
Validation Results - Epoch: 1  Avg accuracy: 0.94 Avg loss: 0.21
```
What do you guys think ?

In addition, added minor changes:
- Remove `Variable`
- replace tensorboardx graph creation method
- fix `F.log_softmax` warnings on unspecified dimension

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -29,6 +29,13 @@ Code
         print("Epoch[{}] Loss: {:.2f}".format(trainer.state.epoch, len(train_loader), trainer.state.output))
 
     @trainer.on(Events.EPOCH_COMPLETED)
+    def log_training_results(trainer):
+        evaluator.run(train_loader)
+        metrics = evaluator.state.metrics
+        print("Training Results - Epoch: {}  Avg accuracy: {:.2f} Avg loss: {:.2f}"
+              .format(trainer.state.epoch, metrics['accuracy'], metrics['nll']))
+
+    @trainer.on(Events.EPOCH_COMPLETED)
     def log_validation_results(trainer):
         evaluator.run(val_loader)
         metrics = evaluator.state.metrics
@@ -136,10 +143,18 @@ or equivalently without the decorator
 
     trainer.add_event_handler(Events.ITERATION_COMPLETED, log_training_loss)
 
-When an epoch ends we want to run model validation, therefore we attach another handler to the trainer on epoch complete
-event:
+When an epoch ends we want compute training and validation metrics [#f1]_. For that purpose we can run previously defined
+`evaluator` on `train_loader` and `val_loader`. Therefore we attach two additional handlers to the trainer on epoch
+complete event:
 
 .. code-block:: python
+
+    @trainer.on(Events.EPOCH_COMPLETED)
+    def log_training_results(trainer):
+        evaluator.run(train_loader)
+        metrics = evaluator.state.metrics
+        print("Training Results - Epoch: {}  Avg accuracy: {:.2f} Avg loss: {:.2f}"
+              .format(trainer.state.epoch, metrics['accuracy'], metrics['nll']))
 
     @trainer.on(Events.EPOCH_COMPLETED)
     def log_validation_results(engine):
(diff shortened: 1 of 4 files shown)


vfdev-5:
Travis CI is need to be restarted

alykhantejani:
@vfdev-5 done

vfdev-5:
Thank you @alykhantejani

docs/source/quickstart.rst
```diff
@@ -17,7 +17,11 @@ Code
     optimizer = torch.optim.SGD(model.parameters(), lr=0.01, momentum=0.8))
     loss = torch.nn.NLLLoss()
 
-    trainer = create_supervised_trainer(model, optimizer, loss)
+    trainer = create_supervised_trainer(model, optimizer, criterion,
```
jasonkriss: Small typo here... `criterion` needs to be `loss` or vice versa

vfdev-5: thanks @jasonkriss !

jasonkriss:
Thanks for taking this on @vfdev-5.

I'm still a lil hesitant on this though. I get that this is a pretty common pattern so it might makes sense to add the ability to `create_supervised_trainer` but i hesitate to put it in the quickstart as I would rather not encourage this approach.

In your lil output sample, the training results are presented as if they are comparable. However, as training metrics are computed against a moving target, they are not. I would much rather encourage users to use the evaluator on both a train set and a validation set.

For example, this is (roughly) what I have in every project:

```python
train_dataset = ...
validation_dataset = ...
train_eval_dataset = Subset(train_dataset, list(range(len(validation_dataset))))
train_loader = DataLoader(train_dataset, ...)
train_eval_loader = DataLoader(train_eval_dataset, ...)
validation_loader = DataLoader(validation_dataset, ...)

trainer = create_supervised_trainer(...)
evaluator = create_supervised_evaluator(...)

def evaluate(trainer):
  validation_metrics = evaluator.run(validation_loader).metrics
  train_metrics = evaluator.run(train_eval_loader).metrics
  print('validation', validation_metrics)
  print('train', train_metrics)
```

That way, we can actually meaningfully compare training and validation metrics.

What do you all think?

vfdev-5:
@jasonkriss probably I see what you mean. If I understand correctly your snippet than `train_eval_dataset` is not seen by model for training, I suppose. So, then your metrics show model's generalization capacity if two dataset are from different distributions?
In my case, train vs validation metrics can be used to see the overfitting.
Correct me please if I missed something

vfdev-5:
@jasonkriss I read  again the snippet, actually, I think I am wrong saying about `train_eval_dataset` is not seen by model for training. Then, some questions:
- why do you prefer to have the same size `len(train_eval_dataset) == len(validation_dataset)`, train_eval_dataset as validation dataset and not taking this randomly ?

I agree that computing train metrics during the training is a sort of average on something inexistant :)

vfdev-5:
@jasonkriss thank you for explaination ! 
I agree that computing metrics with `create_supervised_trainer` does not have much sense, so no need for such modification.
Okay, I'll see what can be kept here in this PR...

jasonkriss:
👍 sounds good. Thanks @vfdev-5. @alykhantejani might have some thoughts on this as well. I know we've discussed it a bit in the past.

vfdev-5:
Ah, okay, surely I missed this discussion.
Maybe, I could reimplement your pattern of train metrics computation in `mnist*.py` examples and tell about this in `quickstart.rst`. 
Otherwise I think I wont be the last person who would like to propose such PR then :)

vfdev-5:
@jasonkriss in your practice do you reuse the same `evaluator` for training and validation performance evaluation ? In some sense, it is better to have two evaluators because validation evaluator can have handlers like ModelCheckpoint, EarlyStopping and reduce on plateau. 
So, it seems to be tricky to verify which (train or val) metrics are in the state in the handler...

alykhantejani:
Sorry I haven't had a chance to look at this yet (have been away for a few days), I will get to this first thing tomorrow when I'm back :)

alykhantejani:
@vfdev-5 the discussion on this previously can be found [here](https://github.com/pytorch/ignite/issues/118#issuecomment-374296777). The conclusion we came to was that:
1) it is indeed useful to compute metrics on the training set. 
2) sometimes you don't want to have a separate evaluator run over your training set (or a subset of it) just to compute metrics when these can be done whilst iterating over the training set during trianing
3) with this in mind a "trainer" and an "evaluator" are essentially the same thing, i.e. an "engine"

After that merge to have a single engine, metrics can now be added to the Trainer no problem. However, as @jasonkriss  mentions this is probably not the best idea as the underlying model moves every iteration.  

So maybe it's worthwhile mentioning in one of the readme docs that whilst this is possible (as you can attach metrics to any engine), the recommended approach is to have a separate engine that is used to benchmark the model on the training set (or a subset) at regular intervals. Therefore we should not add this functionality to `create_supervised_trainer` as we don't want to encourage this pattern.

vfdev-5:
@jasonkriss @alykhantejani I removed `create_supervised_trainer` with metrics and left minor updates of examples, reimplemented metrics computation with 2nd pass and added a footnote on these two approaches.

examples/mnist_with_visdom.py
```diff
@@ -34,7 +34,7 @@ def forward(self, x):
         x = F.relu(self.fc1(x))
         x = F.dropout(x, training=self.training)
         x = self.fc2(x)
-        return F.log_softmax(x)
+        return F.log_softmax(x, dim=-1)
```
jasonkriss: Is there a reason you didn't add a handler to log validation results in this example?

jasonkriss:
@vfdev-5 just one small comment/question. Looks great! Thanks!

examples/mnist_with_visdom.py
vfdev-5: @jasonkriss do you mean no `def log_training_results(engine)` as in mnist or mnist with tensorboardX ? 
Actually, there is no particular reason except we need to create specific windows and draw lines as done in validation... :)
Yep, I will add this

vfdev-5: @jasonkriss I updated the example, if you want to take a look

jasonkriss: @vfdev-5 LGTM. Thanks!

PR #170, opened 2018-05-05
Metric Loss should be agnostic to number of outputs

igormq:
The responsibility of feed them correctly should lie on `output_transform` function. This would enable working with losses that require more than 2 outputs (e.g., CTC loss)

### DIFF
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -21,8 +21,7 @@ class Loss(Metric):
         self._num_examples = 0
 
     def update(self, output):
-        y_pred, y = output
-        average_loss = self._loss_fn(y_pred, y)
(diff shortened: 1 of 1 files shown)


vfdev-5:
@igormq actually we have already evoked this [here](https://github.com/pytorch/ignite/pull/112)
The problem is that how do you define `y.shape[0]` ?
```python
average_loss = self._loss_fn(*output)
assert len(average_loss.shape) == 0, '`loss_fn` did not return the average loss'
self._sum += average_loss.item() * y.shape[0] # y is not defined 
```
A solution is to implement a custom function in case of CTC loss. What do you think ?

igormq:
You are right, now I also think that is better to create a custom function in that case. :(

thanks @vfdev-5!

Issue #169, opened 2018-05-04
[Feature Request] Parameter scheduling


sean-adler:
Providing an abstraction to adjust optimizer parameters during training seems like it might be useful - techniques like [SGDR](https://arxiv.org/pdf/1608.03983.pdf) seem applicable to many types of models.

The [`torch.optim.lr_scheduler`](https://pytorch.org/docs/master/optim.html#how-to-adjust-learning-rate) module in PyTorch core implements some useful schedulers, but (a) can only adjust the LR, and (b) only adjusts it per-epoch.

On the other hand, the `Engine` event API seems like a really natural way to adjust parameter values, since handlers that manipulate them could be added for either `ITERATION_*` or `EPOCH_*` events, and modifying multiple parameters at once (e.g. LR and momentum) would be straightforward too.

I wrote a short [IPython notebook](https://gist.github.com/sean-adler/467321a23805736f8dc06a19157b0567) as a prototype of one way it could look to do this with the event API in a general way (plots are at the very bottom). I left most of the actual scheduler code in separate files for now to try and see if the idea is even worth it first. Would this be useful?

vfdev-5:
@sean-adler there was a discussion in pytorch issues on this and there is [CosineAnnealing](https://pytorch.org/docs/stable/optim.html?highlight=cosineannealinglr#torch.optim.lr_scheduler.CosineAnnealingLR) built-in in pytorch. However, maybe it could make sense to modify LR and momentum... 

Speaking about techniques like SGDR, maybe it could be generalized to something like LR scheduler with restarts, like I implemented [here](https://github.com/vfdev-5/ignite-examples/blob/master/classification/cifar_10/lr_schedulers.py#L6). 

HTH

sean-adler:
> Speaking about techniques like SGDR, maybe it could be generalized to something like LR scheduler with restarts, like I implemented [here](https://github.com/vfdev-5/ignite-examples/blob/master/classification/cifar_10/lr_schedulers.py#L6).

@vfdev-5 Nice! Yeah, that's pretty much exactly what I had in mind, except with batch-level adjustment too.

vfdev-5:
@sean-adler could you explain for me please why is it necessary to control at batch-level adjustment ?

sean-adler:
> @sean-adler could you explain for me please why is it necessary to control at batch-level adjustment ?

[...]

PR #171, opened 2018-05-05
Parameter scheduling handler

sean-adler:
https://github.com/pytorch/ignite/issues/169

Adds some handlers for optimizer parameter scheduling.

The proposed API looks like this:

```
model = Model()
optimizer = SGD(model.parameters(), lr=1e-3)
trainer = create_supervised_trainer(model, optimizer, F.nll_loss)

scheduler = LinearScheduler(optimizer, 'lr', 1e-3, 1e-1, len(train_loader))
trainer.add_event_handler(Events.ITERATION_COMPLETED, scheduler)

# Linearly increases LR from 1e-3 to 1e-1 and back to 1e-3 over the course of 1 epoch
```

* the `ParamScheduler` base class doesn't do much other than setting up the basic API and (optionally) saving parameter histories.
* the `CyclicalScheduler` class implements `get_cycle_progress()`, which returns a percentage representing how far we are through the current cycle. This makes it fairly easy to implement cyclical parameter schedules - for example, the cosine annealing algorithm as described in the [SGDR](https://arxiv.org/pdf/1608.03983.pdf) paper becomes 2 lines of code.
* the `CosineAnnealingScheduler` class implements cosine annealing as described in that paper, and the `LinearScheduler` class implements the "triangular" linear cyclical learning rate policy as described in [Cyclical Learning Rates for Training Neural Networks](https://arxiv.org/pdf/1506.01186.pdf).

Random side-note: I don't love the amount of OOP here (3 layers of inheritance is kind of a lot), but I wanted the user-facing API of writing new schedules (e.g. `LinearScheduler.get_param`) to be simple, so hopefully the rest of the `param_scheduler` module isn't too convoluted.


vfdev-5:
As it was discussed in #169 @sean-adler could you please put this code and tests into `contrib` module ? 
If you do not mind I'll close soon this PR in favor of a new one to `contrib`

vfdev-5:
@sean-adler I updated the code if you do not mind

vfdev-5:
@alykhantejani @jasonkriss review please

jasonkriss:
@sean-adler Nice work! I think this looks really clean. Nice work on the tests! The only thing I would request is that we add a docstring for `get_param`. I also agree that the multiple layers of inheritance is less than ideal. If anyone has a better alternative, that'd be great. Otherwise, I think it's fine for now.

vfdev-5:
@jasonkriss I added `get_param` docstring and renamed `LinearScheduler` to `LinearCyclicalScheduler`

jasonkriss:
👍thanks @vfdev-5. LGTM.

vfdev-5:
@jasonkriss could we merge this, please ?

Issue #173, opened 2018-05-08
[Bug] Bug when add an exception handler


vfdev-5:
As discussed [here](https://github.com/pytorch/ignite/pull/162#issuecomment-387418936), 
signature checking fails when adding an exception handler, especially, exception `e` argument is not taken into account.

For example, the following code fails:
```python
def update_function(engine, batch):
    raise ValueError("TEST VALUE ERROR")

engine = Engine(update_function)

class ExceptionCounter(object):
    def __init__(self):
        self.exceptions = []

    def __call__(self, engine, e):
        self.exceptions.append(e)

counter = ExceptionCounter()        
engine.add_event_handler(Events.EXCEPTION_RAISED, counter)
state = engine.run([1])
```
with 
```
ValueError: Error adding <__main__.ExceptionCounter object at 0x7f897d69bdd8> 'handler': takes parameters ['engine', 'e'] but will be called with [<ignite.engine.engine.Engine object at 0x7f897d69bc50>] (missing a required argument: 'e')
```
Test with `Mock` does not revealed the bug.

alykhantejani:
fixed in #174

PR #174, opened 2018-05-08
Fix issue #173

vfdev-5:
CI will fail in the fixed test because of another bug discussed [here](https://github.com/pytorch/ignite/pull/162#discussion_r186746525):

when test exception is raised the method `Engine._run_once_on_dataset` handles the exception and returns None but `hours, mins, secs = self._run_once_on_dataset()` fails with error.
This second exception make current test fail. 

cc @elanmart, this PR should wait until your PR to fix above bug with `hours, mins, secs = self._run_once_on_dataset()`

@alykhantejani if you know a proper manner to use `MagicMock` I'd be happy to fix my code.

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -100,7 +100,8 @@ class Engine(object):
             self._logger.error("attempt to add event handler to an invalid event %s ", event_name)
             raise ValueError("Event {} is not a valid event for this Engine".format(event_name))
 
-        self._check_signature(handler, 'handler', *args, **kwargs)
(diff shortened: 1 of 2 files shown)


vfdev-5:
@alykhantejani could you please restart CI

alykhantejani:
The tests wont pass due to #175 so we should merge that fix first

alykhantejani:
kicked off the build again. Once green this is good to merge

vfdev-5:
@alykhantejani sounds good! 

sorry, I pushed "comment" button before finished the phrase :)

PR #177, opened 2018-05-10
Nit: replace Engine._event_handlers dict by defaultdict

vfdev-5:
I was just puzzled for some seconds by the code:
```python
if event_name not in self._event_handlers:
    self._event_handlers[event_name] = []
```
and thus for ease of understanding it could be replaced by `defaultdict`.
This PR is really a code polishing, so if there are some objections, we can easily close it. 
What do you guys think ?

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -2,6 +2,7 @@ import inspect
 import logging
 import sys
 import time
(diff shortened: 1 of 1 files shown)


alykhantejani:
LGTM, not sure why I didn't use a defaultdict in the first place.

PR #182, opened 2018-05-13
[WIP] Engine checkpointing

vfdev-5:
### This PR is WIP to address the issue #62 

Implementation tries to cover the ideas from the issue to serialize
- the model => model.state_dict() in `EngineCheckpoint`
- the optimizer => optimizer.state_dict() in `EngineCheckpoint`
- the state & history => added `Engine.state_dict` to store iteration, epoch, max_epochs, seed
- allows easily resuming training from this state. => added method `Engine.resume` that takes as arguments: 1) checkpoint folder (where EngineCheckpoint stores checkpoints), 2) dataloader and 3) objects to restore: model(s), optimizer(s), all objects that have `state_dict` and `load_state_dict` attributes. 

See ` examples/mnist_save_restore_engine.py` to have an idea of proposed API.

It is not a big deal to save and restore pytorch objects as models, optimizers etc 

Major problem to resume the training from a given iteration is related to the type of given `data`.
Today, we accept everything that is iterable and have the length: list, tuple, DataLoader, etc

In this implementation Engine resuming is achieved only for `data` being pytorch `DataLoader`.
Before looping on an epoch, a seed depending on epoch value is setup => such that we can control randomness of data augs, sampling etc for a single epoch:
```python
    while self.state.epoch < self.state.max_epochs and not self.should_terminate:
        self.state.epoch += 1
        self._fire_event(Events.EPOCH_STARTED)
        self._manual_seed(self.state.seed, self.state.epoch)
```
Thus if we set the correct seed we can rerun exactly the epoch and resume from the begining of the epoch.

If we need to restore from an iteration in somewhere between two epochs, we need to rewind the data. 
Data iteration even without forward pass and update can be expensive, so if we have pytorch data loader we can work directly with batch sampler (see [_DataLoaderIter](https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py#L235)). The role of batch sampler is to return indices to load from dataset, however it is an iterator that can contain some random generator.

In order to avoid random state change before _DataLoaderIter starts to iterate over the batch sampler, idea is to transform it into something predictable and deterministic. The class `RewindableBatchSampler` handles this: takes a batch sampler as input and iterates & stores the batch indices. Thus, we can get batches for any iteration we like.
The last point is to replace original batch sampler from the input dataloader by rewindable batch sampler => we make a copy of the data loader and replace the batch sampler. 

### Major changes:
- State has `__repr__` and `seed`
- `dataloader` is now the attribute of `Engine` instead of `State` <= due to the replacement of BatchSampler the dataloader is copied, so perefer to not expose the user not his original dataloader
- class `RewindableBatchSampler`

### Need to be done:
- Resume Engine on other types of data
- More tests to check model(s) and optimizer(s), lr scheduler(s) resuming etc.

@alykhantejani @jasonkriss @elanmart @sean-adler @pkdogcom what do you guys think about this, happy to hear your feedback and comments/critics etc ?


ignite/handlers/checkpoint.py
```diff
@@ -178,3 +181,100 @@ def __call__(self, engine, to_save):
...
+    def __init__(self, dirname, to_save,
+                 save_interval=100,
+                 atomic=True, create_dir=True):
+        assert isinstance(dirname, str)
+        assert isinstance(to_save, dict), "Argument `to_save` should be a dictionary"
+        self.check_objects(to_save)
+
+        self.dirname = dirname
+        self.to_save = to_save
+        self.save_interval = save_interval
```
alykhantejani: Why do we have this attribute twice?

tests/ignite/engine/test_engine.py
```diff
@@ -198,7 +199,9 @@ def test_state_attributes():
     assert state.iteration == 9
     assert state.output == 1
     assert state.batch == 3
-    assert state.dataloader == dataloader
+    # dataloader is now attributed to engine
+    # assert state.dataloader == dataloader
```
alykhantejani: remove commented old test

alykhantejani:
Thanks for the PR, in general I think the approach is good and will be really useful once merged. I have some general comments and questions below:

I feel like the logic in `run` has become a bit complex, there's a few different functions being called which make it a bit harder to follow.

We now have an `EngineCheckpoint` and a `ModelCheckpoint` but we also use the `EngineCheckpoint` to checkpoint a model and an optimizer, this feels a bit confusing to me. 

From my understanding, this is what we want to achieve:
 - A checkpoint of an engine is really just a checkpoint of the State (iteration, epoch, any other user state etc.), the model and the optimizer. (All other funky user stuff should be contained in the state object)
- Resuming training from a checkpoint is really just loading the state back into your engine and running as usual

Right now the model and optimizer get reloaded from the checkpoint only if they are in the to_load/to_save dict, instead this should always happen. 

The comments above are just a dump of thoughts I had while reading this PR. I think it's almost there but we can simplify the usage a bit more. I think something like this would be cleaner:

For checkpointing:
```
model = ...
optimizer = ...
trainer = Engine(...)
checkpoint = EngineCheckpoint(path, interval, model=model, optimizer=optimizer)
trainer.add_event_handler(Event.ITERATION_COMPLETE, checkpoint)
trainer.run(data)
```
For resuming
```
model = ...
optimizer = ...
trainer = Engine(...)
trainer.load_from_checkpoint(path, model=model, optimizer=optimizer)
trainer.run(data)
```

To summarize my proposed changes:
- The EngineCheckpoint handler should explicitly ask for the model and optimizer (it is always passed the engine) as these are required to resume training
- We should keep a unified run interface (might simplify user code a bit)
- the internals of run can be made simpler. Once `load_from_checkpoint` has been called, it can replace the batch_sampler with a sampler that has been progressed to the correct point. `engine.state.epoch` can be reduced by 1 and the normal training loop can continue?

Other points:
- The state_dict stuff should move from the `Engine` to the `State` as this is what we are actually exporting from the engine

In general though,  I think the approach is good and will be really useful once merged. Let me know if my comments make sense? Happy to hear counters if you disagree/I've misunderstood something.

ignite/engine/engine.py
```diff
@@ -222,4 +303,46 @@ def run(self, data, max_epochs=1):
...
+            hours, mins, secs = self._run_once_on_dataset()
+            self._logger.info("Epoch[%s] Complete. Time taken: %02d:%02d:%02d", self.state.epoch, hours, mins, secs)
+            if self.should_terminate:
+                break
+            self._fire_event(Events.EPOCH_COMPLETED)
+
+    def _manual_seed(self, seed, epoch):
+        torch.manual_seed(seed + epoch)
+
+    def _copy_dataloader(self, loader):
```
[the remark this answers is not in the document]
vfdev-5: @alykhantejani maybe it can be replaced, I didn't yet tested what it could do such copies...

ignite/handlers/checkpoint.py
vfdev-5: @alykhantejani you are right my bad, copy paste

ignite/_utils.py
```diff
@@ -29,3 +29,31 @@ def convert_tensor(input_, device=None):
...
+        assert 0 <= start_batch_index < len(batch_sampler)
+        self.batch_sampler = batch_sampler
+        self.start_batch_index = start_batch_index
+        self.batch_indices = self._setup_batch_indices(batch_sampler, start_batch_index)
+
+    @staticmethod
+    def _setup_batch_indices(batch_sampler, start_batch_index):
+        batch_indices = []
+        for batch in batch_sampler:
+            batch_indices.append(batch)
```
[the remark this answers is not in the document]
vfdev-5: @alykhantejani here the situation is a bit tricky. If batch sampler uses a torch random generator the RNG state can and will change between manual_seed and once batch_sampler is called to provide batch indices. In this sense the iteration `for batch in batch_sampler:` ensure that we will see exactly the same indices if we resume or not

vfdev-5:
@alykhantejani thank you for your remarks ! I'll try to answer them:
> I feel like the logic in run has become a bit complex, there's a few different functions being called which make it a bit harder to follow.

I agree, I will try to think how to make it more transparent

> We now have an EngineCheckpoint and a ModelCheckpoint but we also use the EngineCheckpoint to checkpoint a model and an optimizer, this feels a bit confusing to me.

I would say `EngineCheckpoint` and `ModelCheckpoint` (can) play a bit different roles. I see `ModelCheckpoint`  as a handler to store the best models according to score_function (however there is also an option of basic light checkpointing vs heavy `EngineCheckpoint`). And `EngineCheckpoint` can provide complete checkpointing : engine, model(s), optimizer(s), lr_scheduler(s). 
Maybe we can split the boundary between them in this way.


> The EngineCheckpoint handler should explicitly ask for the model and optimizer (it is always passed the engine) as these are required to resume training

I totally agree with that, but we can not know a priori how many models, optimizers user would like to save (e.g. GAN vs classification) and maybe lr_scheduler too

> the internals of run can be made simpler. Once load_from_checkpoint has been called, it can replace the batch_sampler with a sampler that has been progressed to the correct point. engine.state.epoch can be reduced by 1 and the normal training loop can continue?

Good point, I have to think about


> The state_dict stuff should move from the Engine to the State as this is what we are actually exporting from the engine

I would say that today `State` is defined just as storage to pass what would we like between handlers, and imo it is not really a state of Engine...
 I try to think about nn.Module or Optimizer and their `state_dict` and compare them with `State` or `Engine` => I would say Engine would more resemble to nn.Module than State does. 

@alykhantejani me too happy to hear your point of view on these details.

alykhantejani:
> I totally agree with that, but we can not know a priori how many models, optimizers user would like to save (e.g. GAN vs classification) and maybe lr_scheduler too

That's a good point! thinking about it a bit more I see why you went this route now, and am happy with it. I think we should do a few things to neaten up the impl:

1) simplifying the logic inside the run function, I think we can have old workflow, but if loading a checkpoint just progressing the batch sampler
2) RewindableBatchSampler is a bit of a confusing name as we are actually fast forwarding it to the last point it was saved?
3) Maybe move RewindableBatchSampler (or whatever we end up calling it) inside the engine.py and making it private, as it should only be exposed to the engine?

vfdev-5:
Thoughts from slack discussion on some corner cases with patterns:
```
trainer.load_from_checkpoint(path, to_load)
trainer.run(data)
```
vs 
```
trainer.run(train_loader, max_epochs=10)
# or
trainer.resume(train_loader, restore_from, to_load=objects_to_checkpoint)
```

### Pattern *Load-Run* 
Today, Engine's behaviour is the following:
- `__init__` => no state
- `run` => state is initialiazed and updated
- second call of `run` => same as the first call 
In some sense we have a "stateless" engine.

If we add `load_from_checkpoint` we add a notion of "state" in the engine:
- initialization => we have to define a default state: `self.state=State(epoch=0, ...)`
- run => update the state
- load_from_checkpoint => update the state
- second call of run => update the state **starting from the previous state** and this is probably not what we would like: e.g. `evaluator.run(val_loader)` in a handler
- `reset` => we need to introduce a reset method to reset the state 

### Pattern *Run-or-Resume* 
In this case we keep "stateless" behaviour of Engine:
- `__init__` => no state
- `run` => state is initialiazed and updated
- second call of `run` => same as the first call 
- `resume` => state is loaded from the checkpoint and updated

Note that `engine.state` is preserved between the calls of `run`, `resume`

So it is up to us to decide what kind of behaviour we would like to adopt.

vfdev-5:
@alykhantejani I think it's time to make some updates on this :)

pkdogcom:
There seems to be several major issues including this one which have been stuck and lack of discussion for months. Is there any way to make some decisions and move forward while leaving refactoring later if necessary?

vfdev-5:
@pkdogcom I working on this in a spare time. I thought about to close this PR and open an issue on Engine Refactoring adopting : Run-or-Resume pattern.
For me, the main problem is to handle all kind of data inputs: lists, dicts, iterators, DataLoader and be able to record and rewind on them. Iterator and DataLoader are the most difficult cases. With DataLoader we can hack the batch sampler to record batch indices and reproduce batches, but with a genering iterator I do not know yet how to proceed ...

jyzhang-bjtu:
I like the pattern : Load-Run, because it separate the main running engine and the load function.

vfdev-5:
Closed in favor of #640

PR #184, opened 2018-05-21
Update _utils.py

vfdev-5:
Update error message. This method can not and is not intended to work with numbers, for example:
```python
>>> convert_tensor(1)
TypeError: input must contain tensors, numbers, dicts or lists; found <class 'int'>
```


alykhantejani:
Thanks!

PR #188, opened 2018-06-06
Update readme, add setup.cfg and bump version

vfdev-5:


alykhantejani:
🎉 we're waiting for docs to add the website right?

vfdev-5:
Yes, so once the website is known, I'll add it to this PR

assets/bare_pytorch_right_part.py
```diff
@@ -0,0 +1,55 @@
+model = Net()
```
alykhantejani: Are these python files needed? If not we should remove it from this PR

vfdev-5: Idea to keep them is to trace the sources of how side-by-side image has been created in case if we would like to change something (a minor change) without rewriting everything...

alykhantejani: I think wiki is a better idea as you mentioned on slack :)

y0ast:
Looks very nice! Well done, excellent way to prove the point.

vfdev-5:
> Looks very nice! Well done, excellent way to prove the point.
thanks @soumith :)

PR #194, opened 2018-06-18
Fix #192 setting the model on specified device if not None

### LINKED ISSUE #192: create_supervised_trainer fails if model device is diff from arg device
Feature or bug to discuss. If I modify `mnist.py` example: 
```python
def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
    model = Net()
    device = 'cpu'

    if torch.cuda.is_available():
        device = 'cuda'
#         model = model.to(device)

    trainer = create_supervised_trainer(model, optimizer, F.nll_loss, device=device)
                                                     'nll': Loss(F.nll_loss)},
                                            device=device)
```
The operation `model(x)` will fail as `x` and `model` are on different devices. Maybe we need to : 
-  check this and assert or
-  remove `device` from argument and setup it internally according to the model or
- set model on the provided `device` 

@alykhantejani @elanmart @jasonkriss  What do you guys think about ?

vfdev-5:
Also tested on a local machine with GPU

@elanmart if you could take a look

### DIFF
--- a/docs/source/engine.rst
+++ b/docs/source/engine.rst
@@ -6,6 +6,9 @@ ignite.engine
 .. autoclass:: Engine
    :members:
 
+.. automodule:: ignite.engine
+   :members:
+
 .. autoclass:: Events
    :members:
    :undoc-members:
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -50,7 +50,6 @@ def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
 
     if torch.cuda.is_available():
         device = 'cuda'
-        model = model.to(device)
 
     optimizer = SGD(model.parameters(), lr=lr, momentum=momentum)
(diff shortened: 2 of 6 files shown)


jasonkriss:
👍 Looks good!

elanmart:
@vfdev-5 LGTM, but perhaps we'd want to remove https://github.com/pytorch/ignite/blob/master/examples/mnist.py#L53 ?

vfdev-5:
@elanmart Nice catch, yes we can remove `model = model.to(device)` from examples.

Concerning a note in the doc, I think we can keep it like that and avoid docs regeneration PR :)
If you guys think if it is necessary I can do it...

elanmart:
> Concerning a note in the doc, I think we can keep it like that and avoid docs regeneration PR :)

My vote would still go to adding the note :) But hey, who am I to decide -- @alykhantejani ?

vfdev-5:
@elanmart every vote is counted in ignite autocracy :)

Issue #209, opened 2018-07-06
[Feature Request] Add option to stop batch iteration without stopping the training


vfdev-5:
As discussed in #208 we need to provide an option to be able to stop batch iteration without stopping the training.

alykhantejani:
Implemented in #210

PR #210, opened 2018-07-06
Add flag to terminate single epoch

vfdev-5:
Addresses #209


tests/ignite/engine/test_engine.py
```diff
@@ -339,6 +339,23 @@ def start_of_iteration_handler(engine):
     assert state.epoch == np.ceil(iteration_to_stop / num_iterations_per_epoch)  # it starts from 0
 
 
+def test_terminate_epoch_stops_mid_epoch():
+    num_iterations_per_epoch = 10
+    iteration_to_stop = num_iterations_per_epoch + 3  # i.e. part way through the 3rd epoch
```
alykhantejani: nit: shouldn't this say part way through the 2nd epoch

Issue #213, opened 2018-07-16
[Docs] Improve output_transform description


vfdev-5:
I think current [description of `output_transform`](https://github.com/pytorch/ignite/blob/master/ignite/metrics/metric.py#L16) is not very clear:
```
output_transform (callable): a callable that is used to transform the
            model's output into the form expected by the metric. This can be
            useful if, for example, you have a multi-output model and you want to
            compute the metric with respect to one of the outputs.
```
Probably, we need to a small example to show that `output_transform` is applied on `engine.state.output` that is eventually is the output of update function and not model's output as it could be understood from the doc string.

Any thoughts?

jasonkriss:
I agree that this is not very clear. I think a small tweak in the wording along with a minimal example would be great. Perhaps just have an update function that returns a dict and then have the output_transform access a member of the dict. I think that'd be enough to illustrate what this can be used for.

PR #214, opened 2018-07-20
Update metrics docs, added example of usage of `output_transform`

vfdev-5:
Addresses #213 

- Needed to move below ` __metaclass__ = ABCMeta` as it disables docs generation.
- Added example with `output_transform`

cc @jasonkriss

### DIFF
--- a/docs/source/metrics.rst
+++ b/docs/source/metrics.rst
@@ -4,8 +4,8 @@ ignite.metrics
 Metrics provide a way to compute various quantities of interest in an online
 fashion without having to store the entire output history of a model.
 
-In practice user needs to attach metric instance to engine. Metric value is
-computed using engine's output:
+In practice a user needs to attach the metric instance to an engine. The metric
+value is then computed using the output of the engine's `process_function`:
 
     .. code-block:: python
 
@@ -17,7 +17,8 @@ computed using engine's output:
         metric = CategoricalAccuracy()
         metric.attach(engine, "accuracy")
 
-If engine's output is not in the format `y_pred, y`, user can use `output_transform`:
+If the engine's output is not in the format `y_pred, y`, the user can
+use the `output_transform` argument to transform it:
 
     .. code-block:: python
 
--- a/docs/source/metrics.rst
+++ b/docs/source/metrics.rst
@@ -1,6 +1,42 @@
 ignite.metrics
 ==============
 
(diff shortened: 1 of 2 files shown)


vfdev-5:
I think we need also provide more info on how to work with metrics in case of custom evaluator: 
- how to attach
- what to write in the output
- etc.

Even user can just copy `create_supervised_evaluator`, this info is not mentioned anywhere

vfdev-5:
Restructured the docs, added more info on how to attach metrics etc

screen shot 2018-07-21 at 21 58 55

ignite/metrics/metric.py
```diff
@@ -4,20 +4,19 @@
...
 
-    Metrics provide a way to compute various quantities of interest in an online
-    fashion without having to store the entire output history of a model.
-
     Args:
         output_transform (callable): a callable that is used to transform the
-            model's output into the form expected by the metric. This can be
-            useful if, for example, you have a multi-output model and you want to
-            compute the metric with respect to one of the outputs.
+            :class:`ignite.engine.Engine` `process_function`'s output into the
```
alykhantejani: nit:class:`ignite.engine.Engine`**'s** `process_function`'s

docs/source/metrics.rst
```diff
@@ -1,6 +1,42 @@
 ignite.metrics
 ==============
 
+Metrics provide a way to compute various quantities of interest in an online
+fashion without having to store the entire output history of a model.
+
+In practice user needs to attach metric instance to engine. Metric value is
```
alykhantejani: nit: In practice **a** user needs to **the**  metric instance to **an** engine. **The** metric value is **then** computed using **the** output of the engine's `process_fn`:

docs/source/metrics.rst
```diff
@@ -1,6 +1,42 @@
...
+
+        def process_function(engine, batch):
+            # ...
+            return y_pred, y
+
+        engine = Engine(process_function)
+        metric = CategoricalAccuracy()
+        metric.attach(engine, "accuracy")
+
+If engine's output is not in the format `y_pred, y`, user can use `output_transform`:
```
alykhantejani: If **the** engine's output is not in the format `y_pred, y`, **the** user can use  **the** `output_transform` **argument to transform it**:

vfdev-5:
@alykhantejani thank you very much ! Will change the PR and need to improve my english :)

Issue #217, opened 2018-07-23
[Feature Request] Loops and user-defined events


AntoinePrv:
This one is more of a long shot. Today ignite provides two nested loops (epochs and batchs) with a custom callback callback function.

In some case however it requires more than two loops. For instance, in truncated backpropagation through time, one iterates over, epochs, then batches, then over time.

Another example I'm thinking about (without adding loops) is for someone working on optimization. Such a person may want to add an event between the `loss.backward()` and `optim.step()` and then swap different handles in their work without having to care bout the rest of the training loop.

One can have workarounds for both these examples, but I think it is valuable to have some way to let the user add some more loops and more events. I believe being very modular will let researchers more easily separate their code between what they are working on, and what is left untouched.

vfdev-5:
@AntoinePrv thanks for the feature request ! Even if you are talking about some specific situations that can be implemented by a custom engine's `process_function`, maybe this can be implemented in common way with what you proposed in #215 - more engines.

In a `contrib` we could eventually provide some customized `process_function`s : 
- `process_function_time_iteration` that could have an internal loop over time
- `optimization_process_function` that could fire some custom events between `loss.backward()` and `optim.step()`
- ...

What do you think about ?

AntoinePrv:
Yes, a `contrib` could be a nice solution to these two examples (and more to come hopefully!).
For the TBTT case, we can build the inner loop inside the `process_function` passed to the `Engine` and even fire custom events dedicated to this iteration.

I wonder how to fire custom events within the given framework.
More precisely:
  - [Events](https://github.com/pytorch/ignite/blob/c4563ddc67bee03fc616e87b84abd9b5e17e21cc/ignite/engine/engine.py#L13) is a `Enum`;
  - Engine [checks that an event is in the Enum](https://github.com/pytorch/ignite/blob/c4563ddc67bee03fc616e87b84abd9b5e17e21cc/ignite/engine/engine.py#L101) before adding it as an handler.

Intuitively, if `Events` was a `dict` (not implying this is what should be done), we could define new events in a `contrib` (or in user code), add it to the supported events, and then fire them and add handlers as usual. This would have no impact if the code is not used because such events would be added at runtime.

Any idea how to proceed?

vfdev-5:
@AntoinePrv thinking about custom events, actually I do not see why we need to pass by event handling, we can call directly the handler in the processing function. For example,
```
def on_custom_event(engine, *args, **kwargs):
    pass

[...]

PR #220, opened 2018-07-28
Custom events and engines

AntoinePrv:
This is work in progress for #217 

# User defined events
I'm opening the PR unfinished to discuss how to actually add user define events given that `Enum` are not mutable.
In this first commit, I swap the value of `ignite.engine.engine` with a new `Enum` in the function `add_event` from `ignite.contrib.event`.
However this does not work in the following case:

```python
# or ignite.engine.engine
from ignite.contrib.events import add_events

add_events(...)
# Or other function that end up adding event
```
This is because now `Events` is in the namespace of the current module. The `Events` object will not be changed and the user cannot access the new events to add new handles (unless weirdly making imports).

I suggest three solutions:
  1. Move the `Enum` inside the `Engine` class. Swaping the `Enum` becomes a mutation of the class so works properly.
  2. Use `aenum` package to make `Enum` mutable (+1 package dependency)
  3. Change the `Enum` structure of `Events` for something mutable

Any opinion/  suggestion welcome.

# Contrib engine
I added a contrib TBPTT trainer to have a example of creating custom events along the way.

# TODO

- [x] Add unit tests
- [x] Find a solution to add new `Events` to the original `Enum`
- [ ] Add examples
- [x] Add reference to TBPTT in docstring


vfdev-5:
to be honest, personnally i'm not so convinced about the way to change Events as you proposed. Need to think about more elegant way.

AntoinePrv:
@vfdev-5 I cannot find the comments...

> Concerning Events, to be honest, personnally i'm not so convinced about the way to change Events as you proposed. Need to think about more elegant way.

I'm not either, hopping someone has a good solution.

Of course this need testing, I'll work on that.

vfdev-5:
In github interface you should see something like that : 
screen shot 2018-07-28 at 21 54 38

AntoinePrv:
@vfdev-5 found this on [StackOverflow](https://stackoverflow.com/questions/10248666/how-to-view-line-comments-in-github):
>If you're here because you've left comments, and can see your own comments, but others cannnot, then check for a yellow pending label on your comments. Others cannot see your comment until you've completed the review process and selected whether you want to request changes. Once you click through a full review, then others will see your comments.

vfdev-5:
@AntoinePrv correct, forgot to submit the review

vfdev-5:
> Find a solution to add new Events to the original Enum

Maybe we could tweak `Engine` to accepts some custom events and add time events into `ignite.contrib.events`.

ignite/contrib/engine.py
```diff
@@ -0,0 +1,110 @@
+# coding: utf-8
```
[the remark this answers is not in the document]
AntoinePrv: Perhaps even `trainers` and leave `engine` for ignite core.
  - engine -> class
  - trainer -> specific instance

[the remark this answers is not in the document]
vfdev-5: > Perhaps even trainers and leave engine for ignite core.

In `ignite.contrib.engines` we could later add evaluators too. Even in your case, do you need a supervised evalutor or native one is ok ?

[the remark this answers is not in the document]
AntoinePrv: > we could later add evaluators too

That's a good point, even though we could have `ignite.contrib.evaluators` module.

In my case no, I'm doing without. There are cases in time series where you want to feed back your output into your input. This does not need to be defined outside of the model (i.e. could be done within the `forward` of `nn.Module` with a simple `if not self.training` because no optimization is done at this point) but if eventually someone write a code for that, it doesn't hurt to have it as an evaluator and not rewrite the code.

ignite/engine/engine.py
```diff
@@ -12,13 +12,13 @@
...
 class Events(Enum):
     """Events that are fired by the :class:`ignite.engine.Engine` during execution"""
-    EPOCH_STARTED = "epoch_started"
-    EPOCH_COMPLETED = "epoch_completed"
-    STARTED = "started"
-    COMPLETED = "completed"
-    ITERATION_STARTED = "iteration_started"
-    ITERATION_COMPLETED = "iteration_completed"
-    EXCEPTION_RAISED = "exception_raised"
+    EPOCH_STARTED = auto()
```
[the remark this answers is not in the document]
AntoinePrv: Removed it.

AntoinePrv:
As for the Events, I changed it by giving the opportunity to pass an optional `Enum` of events at engine initialization.

ignite/contrib/engines.py
```diff
@@ -0,0 +1,107 @@
...
+    device=None
+):
+    """
+    Factory function for creating a trainer for truncated backpropagation
+    through time supervised models.
+
+    Args:
+        model (`torch.nn.Module`): the model to train
+        optimizer (`torch.optim.Optimizer`): the optimizer to use
+        loss_fn (torch.nn loss function): the loss function to use
```
vfdev-5: nit: maybe just a callable ?

ignite/engine/engine.py
```diff
@@ -37,6 +37,9 @@ class Engine(object):
     Args:
         process_function (Callable): A function receiving a handle to the engine and the current batch
             in each iteration, and returns data to be stored in the engine's state
+        events (Enum, optional): Additional events that can be fired. Use this
```
vfdev-5: nit: maybe better  `events` -> `custom_events` ?

ignite/contrib/engines.py
```diff
@@ -0,0 +1,107 @@
...
+            loss_t = loss_fn(y_pred_t, y_t)
+            loss_t.backward()
+            optimizer.step()
+
+            # Setting state of engine for consistent behaviour
+            engine.state.output = loss_t.item()
+            loss_list.append(loss_t.item())
+
+            # Fire event for end of iteration
+            engine._fire_event(Tbptt_Events.TIME_ITERATION_STARTED)
```
vfdev-5: nit: `TIME_ITERATION_STARTED` -> `TIME_ITERATION_COMPLETED`

vfdev-5:
@AntoinePrv looks good:
* [x] Find a solution to add new Events to the original Enum 

cc @alykhantejani

ignite/contrib/engines.py
```diff
@@ -0,0 +1,107 @@
...
+    model,
+    optimizer,
+    loss_fn,
+    tbtt_step,
+    dim=0,
+    device=None
+):
+    """
+    Factory function for creating a trainer for truncated backpropagation
+    through time supervised models.
```
vfdev-5: Could you please also add more context and link on the tbptt

vfdev-5:
@AntoinePrv Looks good, just a flake fix needed

AntoinePrv:
Adding the example is going to take a little more time.

ignite/contrib/engines.py
```diff
@@ -36,10 +37,12 @@ def _detach_hidden(hidden):
     Auxillary function to cut the backpropagation graph by detaching the hidden
     vector.
     """
-    if isinstance(hidden, collections.Iterable):
+    if isinstance(hidden, torch.Tensor):
+        return hidden.detach()
+    elif isinstance(hidden, tuple):
```
vfdev-5: nit: `elif isinstance(hidden, tuple):` -> `elif isinstance(hidden, (list, tuple)):`

AntoinePrv: I removed `collections.Iterable` because it turns out that `Tensor` implements that (in an unexpected way).

In Pytorch, the hidden vectors for recurrent models are either a `Tensor` or a `tuple` of `Tensor`. I'll mimic the function you mentioned so that this can generalize to user-defined structures if needed.

ignite/engine/engine.py
```diff
@@ -37,6 +37,9 @@ class Engine(object):
     Args:
         process_function (Callable): A function receiving a handle to the engine and the current batch
             in each iteration, and returns data to be stored in the engine's state
+        additional_events (Enum, optional): Additional events that can be fired.
```
vfdev-5: Just wonder whether it makes sense to have a list of custom events ? 
For example, I write another custom Tbptt trainer and want to use your Tbptt_Events plus another CustomEvent...
Maybe a solution will be to inherit of Tbptt_Events ...

vfdev-5:
LGTM.

@alykhantejani @jasonkriss could you please take a look and give your opinion

I think we can agree on merge this without examples and provide an example in another PR.

tests/ignite/contrib/test_engines.py
```diff
@@ -0,0 +1,98 @@
...
+    assert not model.weight_hh_l0.item() == pytest.approx(0)
+
+
+def test_create_supervised_tbptt_trainer_with_cpu():
+    _test_create_supervised_tbptt_trainer("cpu")
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="Skip if no GPU")
+def test_create_supervised_tbptt_trainer_with_gpu():
+    _test_create_supervised_tbptt_trainer("gpu")
```
vfdev-5: nit: "gpu" -> "cuda"

vfdev-5: Actually, I can fix it by myself (seen a lot of times, but never tried and it works :)

jasonkriss:
I have two minor discussion points...

1. Is it worth the complicated API to have event validation when attaching a handler? I wonder if it might be better to just allow firing/handling arbitrary events. New engines can implement their own event enums to make this easier but it wouldn't require changes to the core engine class. I know there's downsides to this, so I don't feel super strongly about it. Wanted to mention it as an option though.

2. If new engines are gonna be firing events, we should maybe consider changing `_fire_event` to `fire_event`.

vfdev-5:
@jasonkriss thanks !
> If new engines are gonna be firing events, we should maybe consider changing _fire_event to fire_event. 

Nice catch. Yes, it should become public 

Concerning the first point, if I understand correctly your point, you propose to remove original event check:
```python
if event_name not in Events.__members__.values():
    self._logger.error("attempt to add event handler to an invalid event %s ", event_name)
    raise ValueError("Event {} is not a valid event for this Engine".format(event_name))
```
such that we can pass to `add_event_handler` any type object. I see your point it will much more flexible and without the overhead with `additional_events`, however the fact that there is not restriction on type of `event_name` makes me hesitating...
Maybe we could impose event_name to be a `str`...

elanmart:
Hi, I also have two minor comments
1. Is there any reason why we want to fire custom `TIME_ITERATION_STARTED` in the propsed bptt-trainer? If this is only serving as an example, then perhaps we should put it into `examples`? 
2. @vfdev-5 how about having `accepted_events` as a property of the `Engine`? If you want to fire custom events, then maybe you should inherit `Engine`, and use something like.
```python
@property
def accepted_events(self):
    return super().accepted_events + [my, events, here]
```

vfdev-5:
Hey @elanmart thanks for switching from pytorch and passing by with your comments :)

We discussed the reason to fire custom event [here](https://github.com/pytorch/ignite/issues/217#issuecomment-408243272) and idea here is to provide a contrib module with some specific trainers: an the first example tbptt trainer to solve however a list of specific tasks with RNNs and be reusable. That's why this is not in examples.

>  how about having accepted_events as a property of the Engine? If you want to fire custom events, then maybe you should inherit Engine, and use something like.

Sure that this can be a solution, but I agree with Jason Kriss that current `Engine` and event_name verification can be simplified without introducing additional attributes or inheritance...

What do you think ?

jasonkriss:
@vfdev-5 Yea i think its ultimately a tradeoff between simplicity and safety. I can see an argument for each side. I subjectively would prefer the simpler side as I'd like to keep `Engine` as minimal as possible.

elanmart:
@vfdev-5 I see, thanks! 

Well I'd probably vote for some form of validation (because I'm paranoid and often make stupid mistakes), but I'm certainly not a fan of having an extra arg in `Engine.__init__`.

vfdev-5:
Okay, thanks you guys, let's wait for Aly's opinion. So, now we opt to remove `additional_events` and replace `event_name` type check by check if it is `str`:
```python
if not isinstance(event_name, str):
    self._logger.error("attempt to add event handler to an invalid event {} ".format(event_name))
    raise ValueError("Event name should be a string, but given {}".format(type(event_name)))
```

alykhantejani:
@AntoinePrv thanks for the nice PR!

I have some comments around adding extra events. So I agree with @elanmart, I don't think we should add an extra arg to the constructor of the `Engine`. What about this:

Inside Engine's constructor we make a variable `self._allowed_events = list(Events.__members__.values())`

Then we have a method on `Engine`, `register_event(event_name)` which just does `self._allowed_events.append(event_name)`. This way `event_name` can be an enum value, a string or whatever. 

The check in `_fire_event` then becomes `if event_name in self._allowed_events:`

And finally, we should change `_fire_event` to `fire_event` (make it public)

vfdev-5:
@alykhantejani really nice solution ! Thanks !

AntoinePrv:
I made the discussed changes, which I think are the right direction. However, I think the interface to `fire_event` is still confusing:
  - IMHO, having it totally public can be confusing: a user not familiar with ignite may not know if he is responsible for firing the core events `ignite.engines.Events` (should it be allowed?)
  - The extra args passed to __all__ handlers (initially for handling the exception event I believe) are passed to all handlers.

ignite/engine/engine.py
```diff
@@ -60,27 +57,58 @@ def train_and_store_loss(engine, batch):
...
-    def __init__(self, process_function, additional_events=None):
+    def __init__(self, process_function):
         self._event_handlers = defaultdict(list)
         self._logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
         self._logger.addHandler(logging.NullHandler())
         self._process_function = process_function
         self.should_terminate = False
         self.should_terminate_single_epoch = False
         self.state = None
+        self._allowed_events = list()
```
vfdev-5: `list()` -> `[]`

vfdev-5:
> IMHO, having it totally public can be confusing: a user not familiar with ignite may not know if he is responsible for firing the core events ignite.engines.Events (should it be allowed?)

@AntoinePrv I agree with this point. We can point out in the doc string the intended usage of the method, saying explicitly that "if no custom events in the engine, no need to call the method".

> The extra args passed to all handlers (initially for handling the exception event I believe) are passed to all handlers.

As you wrote in the docs, event *args, **kwargs are optional and its true that all handlers will receive them in case of a custom event.

So we have two types of args/kwargs:
- static = set during handler function setup
- dynamic = set during `fire_event` if case of custom events only

[Previously in the docs](https://pytorch.org/ignite/concepts.html#state) we suggested to use `engine.state` to exchange "dynamic" data between `process_function` and `handlers`.

To avoid double usage we can wrap `_fire_event` by a more user friendly `fire_event` without args/kwargs. 

What do you think ?

ignite/engine/engine.py
```diff
@@ -212,17 +280,17 @@ def run(self, data, max_epochs=1):
         try:
             self._logger.info("Engine run starting with max_epochs={}".format(max_epochs))
             start_time = time.time()
-            self._fire_event(Events.STARTED)
+            self.fire_event(Events.STARTED)
```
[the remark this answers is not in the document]
AntoinePrv: To clarify: you say we  should leave `_fire_event` in all `Engine` internal calls?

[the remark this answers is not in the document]
vfdev-5: Yes, I think we could leave internally `self._fire_event`

ignite/engine/engine.py
```diff
@@ -65,17 +65,50 @@ def __init__(self, process_function):
...
+        """
+        for name in event_names:
+            self._allowed_events.append(name)
+
     def add_event_handler(self, event_name, handler, *args, **kwargs):
         """Add an event handler to be executed when the specified event is fired
 
         Args:
-            event_name (Events): event from ignite.engine.Events to attach the handler to
+            event_name: event from ignite.engine.Events to attach the handler to
```
alykhantejani: should also add something like "or any event_name added by `register_events`"

ignite/engine/engine.py
```diff
@@ -138,7 +171,7 @@ def on(self, event_name, *args, **kwargs):
         """Decorator shortcut for add_event_handler
 
         Args:
-            event_name (Events): event to attach the handler to
+            event_name: event to attach the handler to
```
vfdev-5: nit: same as said Aly on `add_event_handler`: 
event from `ignite.engine.Events` or any event_name added by `register_events` to attach the handler to

vfdev-5:
@AntoinePrv thank you !

vfdev-5:
@AntoinePrv FIY there is a discussion on TBPTT implementations [here](https://discuss.pytorch.org/t/implementing-truncated-backpropagation-through-time/15500/9).

AntoinePrv:
Thanks for the heads up. I implemented on purpose  `k1 == k2` because it is much simpler and achieves most of the improvement.
I'm happy to work on a `k1 != k2` implementation, as well as providing the missing example in the near future.

vfdev-5:
@AntoinePrv sounds good! So, send a PR when you have time to advance on this, as you know PR's are very welcome :)

PR #222, opened 2018-08-03
Neural style example

anmolsjoshi:
Pull Request for PyTorch Ignite implementation of Fast Neural Style. 

To run on Travis, the code creates a random style image, test image, content image and creates a model with CIFAR10 dataset. 

To correctly create a style transfer model, the MSCOCO dataset and an appropriate style image should be used. 

 For training
  - python examples/fast_neural_style/neural_style.py train --epochs 1 --cuda 0 --style_image random --test_image random
  
For evaluation
  - python examples/fast_neural_style/neural_style.py eval --content_image random --output_image /tmp/test.png --cuda 0 --model /tmp/checkpoints/checkpoint_net_1.pth


vfdev-5:
@anmolsjoshi Thanks for the PR! 

I left some comments in the code on how to improve it. Next I'll check it on my machine etc and keep you updated.

Binary files :  examples/fast_neural_style/images/stylized_test/stylized_sample_epoch_000X.png are
produced by the example ? If yes we need to remove them too.

.idea/workspace.xml
```diff
@@ -0,0 +1,198 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
anmolsjoshi: Incorporated.

examples/fast_neural_style/neural_style.py
```diff
@@ -0,0 +1,291 @@
...
+    trainer = Engine(step)
+    checkpoint_handler = ModelCheckpoint(args.checkpoint_model_dir, 'checkpoint',
+                                         save_interval=args.checkpoint_interval,
+                                         n_saved=10, require_empty=False, create_dir=True)
+
+    @trainer.on(Events.ITERATION_COMPLETED)
+    def update_logs(engine):
+        for k, v in engine.state.output.items():
+            old_v = running_avgs.get(k, v)
+            new_v = alpha * old_v + (1 - alpha) * v
```
vfdev-5: Can we use ignite metrics for that ?

examples/fast_neural_style/README.md
```diff
@@ -0,0 +1,76 @@
+# fast-neural-style
+
+### Introduction
+This example is ported over from [pytorch-examples](https://github.com/pytorch/examples/tree/master/fast_neural_style).
+
+It uses pytorch-ignite to implement an algorithm for artistic style transfer as described in [Perceptual Losses for Real-Time Style Transfer and Super-Resolution](https://arxiv.org/abs/1603.08155).
+
+This repo gives examples on how to create a Progress Bar with `pytorch-ignite`.
```
vfdev-5: What do you mean by : "how to create a Progress Bar with `pytorch-ignite`" ?

examples/fast_neural_style/neural_style.py
anmolsjoshi: Yes, we can. I got that from the dcgan example. I'll look into that this weekend.

examples/fast_neural_style/README.md
anmolsjoshi: My bad, I can remove that. I added some functions in the code to create a progress bar to print metrics related to training. It was an additional example of how to use Events.

examples/fast_neural_style/neural_style.py
anmolsjoshi: @vfdev-5 Looking into this more, I'm not exactly sure how to incorporate Ignite Metrics into this. 

From what I have seen of the Metrics documentation, metrics are calculated using the output of the `step function` of the engine. In this case the metrics are the content_loss, style_loss and total_loss (MSE) that are needed for the optimization step inside the `step(engine, batch)` function. 

There are no metrics calculated at the end of Epoch, like in a classification problem.

This is why I used the `update_logs(engine)` function as shown. 

Happy to run with any ideas you may have.

vfdev-5: @anmolsjoshi I see your point, true that here there are running loss function values and we can not use Metrics in this case.

examples/fast_neural_style/neural_style.py
```diff
@@ -0,0 +1,291 @@
...
+import random
+from PIL import Image
+import torch
+from torch.optim import Adam
+from torch.utils.data import DataLoader
+from torchvision import datasets
+from torchvision import transforms
+
+from ignite.engine import Engine, Events
+from ignite.handlers import ModelCheckpoint, Timer
```
vfdev-5: If you are not using Timer, it seems, we can remove it

examples/fast_neural_style/neural_style.py
```diff
@@ -0,0 +1,291 @@
+# coding: utf-8
+from __future__ import print_function, division
+
+import argparse
+import os
+import sys
+import time
```
vfdev-5: Same here, it seems that time, re are not used, according to PyCharm

examples/fast_neural_style/neural_style.py
```diff
@@ -0,0 +1,291 @@
...
+            sub = 0
+
+        else:
+            sub = 1
+            progress = 1
+            equal_to = np.max(np.where([percent < percent_seen for percent in percentages])[0])
+
+        bar = '[' + '=' * equal_to + '>' * progress + ' ' * (10 - equal_to - sub) + ']'
+
+        message = 'Epoch {epoch} | {percent_seen:.2f}% | {bar}'.format(epoch=engine.state.epoch,
```
vfdev-5: Concerning your progress bar, looks cool on the execution. Could you please refactor it in a separate function ? Such that it could be reused by users if they would like to

examples/fast_neural_style/neural_style.py
```diff
@@ -0,0 +1,291 @@
...
+        utils.save_image(args.output_image, output[0])
+
+
+def main():
+    main_arg_parser = argparse.ArgumentParser(description="parser for fast-neural-style")
+    subparsers = main_arg_parser.add_subparsers(title="subcommands", dest="subcommand")
+
+    train_arg_parser = subparsers.add_parser("train", help="parser for training arguments")
+    train_arg_parser.add_argument("--epochs", type=int, default=2, help="number of training epochs, default is 2")
+    train_arg_parser.add_argument("--batch_size", type=int, default=8,
```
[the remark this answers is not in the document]
anmolsjoshi: Incorporated

examples/fast_neural_style/neural_style.py
anmolsjoshi: Incorporated

examples/fast_neural_style/neural_style.py
anmolsjoshi: Incorporated

examples/fast_neural_style/neural_style.py
anmolsjoshi: Added handlers.py with Progbar class. The code is now a lot cleaner. And I was wondering if it could be added to the ignite handlers? Maybe that's a separate issue.

vfdev-5: @anmolsjoshi Look good ! Yes, I thought about to propose a sort of RunningMetrics handler. However in this case we would need to separate running metric computing and progress bar logic. For a progress bar reporting, maybe we could take a look at tqdm direction.

vfdev-5:
Looks good to me!  Tested locally, looks ok

test

@anmolsjoshi nice solution for travis!

cc @alykhantejani @elanmart could you give your opinion

elanmart:
I haven't reviewed the algorithm, but the code and docs look good to me!

examples/fast_neural_style/README.md
```diff
@@ -0,0 +1,74 @@
+# fast-neural-style
+
+### Introduction
+This example is ported over from [pytorch-examples](https://github.com/pytorch/examples/tree/master/fast_neural_style).
+
+It uses pytorch-ignite to implement an algorithm for artistic style transfer as described in [Perceptual Losses for Real-Time Style Transfer and Super-Resolution](https://arxiv.org/abs/1603.08155).
```
alykhantejani: nit: can we change `pytorch-ignite` to `ignite`. `pytorch-ignite` is the name on pip but the package is called `ignite`

examples/fast_neural_style/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Requirements
+
+The code can be executed using both `Python 2.7` and `Python 3.5`.
```
alykhantejani: this can be removed as is an assumption made by pytorch + ignite in general

examples/fast_neural_style/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+* `torch`
+* `torchvision`
+* `pytorch-ignite`
```
alykhantejani: should be `ignite` as it's the package name`

examples/fast_neural_style/neural_style.py
```diff
@@ -0,0 +1,249 @@
...
+                                 help="path for saving the output image")
+    eval_arg_parser.add_argument("--model", type=str, required=True,
+                                 help="saved model to be used for stylizing the image.")
+    eval_arg_parser.add_argument("--cuda", type=int, required=True,
+                                 help="set it to 1 for running on GPU, 0 for CPU")
+
+    args = main_arg_parser.parse_args()
+
+    if args.subcommand is None:
+        print("ERROR: specify either train or eval")
```
alykhantejani: these should be changed to `raise ValueError` instead of `print` and `sys.exit`

examples/fast_neural_style/neural_style.py
```diff
@@ -0,0 +1,249 @@
...
+    trainer.add_event_handler(event_name=Events.ITERATION_COMPLETED, handler=progress_bar)
+    trainer.run(train_loader, max_epochs=args.epochs)
+
+
+def stylize(args):
+    device = torch.device("cuda" if args.cuda else "cpu")
+
+    if args.content_image != 'test':
+        content_image = utils.load_image(args.content_image, scale=args.content_scale)
+    else:
```
alykhantejani: we should remove this testing code from here. We don't actually need to do the smoke test and see results

examples/fast_neural_style/neural_style.py
```diff
@@ -0,0 +1,249 @@
...
+    device = torch.device("cuda" if args.cuda else "cpu")
+
+    transform = transforms.Compose([
+        transforms.Resize(args.image_size),
+        transforms.CenterCrop(args.image_size),
+        transforms.ToTensor(),
+        transforms.Lambda(lambda x: x.mul(255))
+    ])
+
+    if args.dataset == 'test':
```
alykhantejani: we should remove the testing code

alykhantejani:
Thanks for the PR.

In general this looks ok, I have some comments though:

1. We should remove the test code. I don't think its necessary to run a test on some fake data - we never look at the output images when this is run on travis.
2. I don't think we should do much work on travis either. I think the whole point of the smoke test is to run the code once (1 iteration max) just to make sure any API changes haven't broken the example. Anything longer than a few seconds is too long
3. We should remove the committed images, they are currently not displayed in the README and we should remove the test code anyway. Alternatively, we can display them in the README like the [original example](https://github.com/pytorch/examples/tree/master/fast_neural_style)
4. I'm not sure the ProgressBar is the best way to do this (we should be able to find a way to use tqdm), but as its for an example I'm ok with this

vfdev-5:
@alykhantejani as discussed on slack, the time consumption on travis by this example is ~35 secs which is not the longest one among others.  So, we should find out a solution to all of the `if test` in the code as it serves only to run example on travis. Points 1) and 2) are to reconsider.

anmolsjoshi:
@vfdev-5 @alykhantejani I've incorporated all the comments. Quick question before I push another commit. Removing the testing code will have this fail on Travis because MSCOCO dataset or the Fake Dataset is needed to run the code, and downloading MSCOCO will take a while on Travis. 

If I do remove the test code, will it still have to pass Travis tests before completing the pull request? Or can it be incorporated without passing travis?

apologies if this is a novice question.

vfdev-5:
@anmolsjoshi Sorry, I haven't yet time to writeup a guide on what we have decided on travis tests. 
Accepted solution is not to remove smoke tests and change the code such that to copy what is done in dcgan example with [`check_dataset`](https://github.com/pytorch/ignite/blob/master/examples/gan/dcgan.py#L156) to avoid urgly `if test` conditions in `train`. In check_dataset here we can similary support 'folder', 'mscoco', 'fake'.
As  `eval` does not use ignite we can avoid its testing.

examples/fast_neural_style/README.md
anmolsjoshi: Incorporated

examples/fast_neural_style/README.md
anmolsjoshi: Incorporated

examples/fast_neural_style/README.md
anmolsjoshi: Incorporated

examples/fast_neural_style/neural_style.py
anmolsjoshi: Incorporated

examples/fast_neural_style/neural_style.py
anmolsjoshi: Incorporated

examples/fast_neural_style/neural_style.py
anmolsjoshi: Incorporated

vfdev-5:
@anmolsjoshi in the README.md when I commited some of my modifications to this PR, I broke the correct representation of MSCOCO tree view. Sorry. Could you please restore it. Thanks

anmolsjoshi:
@vfdev-5 I fixed it! Here is a summary of changes:

- Updated README.md as requested. Added images as shown in pytorch examples.
- Removed function to stylize on epoch completion, I can add it back in, didn't think it was necessary. 
- Included check_dataset as in dcan.py example
- test works by creating Fake Dataset sized of one batch, thus test only uses one iteration
- Removed eval test from .travis.yml 
- Incorporated remaining comments as requested by @vfdev-5  and @alykhantejani

examples/fast_neural_style/neural_style.py
```diff
@@ -0,0 +1,216 @@
...
+    running_avgs = OrderedDict()
+
+    def step(engine, batch):
+
+        x, _ = batch
+        x = x.to(device)
+
+        n_batch = len(x)
+
+        transformer.zero_grad()
```
alykhantejani: `transformer.zero_grad()` and `optimizer.zero_grad()` should be the same AFAIK, but can we change this to `optimizer.zero_grad()` to be consistent with the other examples?

alykhantejani:
LGTM - @vfdev-5 if the changes are good with you lets go ahead and merge.

Thanks @anmolsjoshi!

Issue #224, opened 2018-08-05
`pack_sequence` capability


kl2792:
Is it possible for this to be implemented in Ignite, or should we just call pack_sequence in our LSTM nn.Modules?

Any help appreciated!

vfdev-5:
@kl2792 unfortunately, we have not yet worked on use-cases with RNN/LSTM. So I would say we can call pack_sequence in your model.

I'm not expert in RNNs but if word embeddings are not trainable, so probably this can be done in a custom [`_prepare_batch`](https://github.com/pytorch/ignite/blob/master/ignite/engine/__init__.py#L7) for example if one wants to use [`create_supervised_trainer`](https://github.com/pytorch/ignite/blob/master/ignite/engine/__init__.py#L12).

Anyway, we can discuss here your use-case and see if we can improve Ignite to cover more RNN/LSTM use-cases. 

And as always PR are very welcome :)

kl2792:
Let me know how my PR's code style/content looks! I think this could help resolve this issue with an extra argument:

``` python3
create_supervised_trainer(..., prepare_batch=lambda batch: pack_padded_sequence(batch[0]), batch[1])
```

alykhantejani:
In this case I would also recommend rolling your own version of `create_supervised_trainer` as the current one is meant to serve as a minimalist example (which fits many use cases). 

I would however be interested to know if this was the only change you needed.

Also a RNN example would be great if you want to send a PR :)

kl2792:
Sounds good. I thought about it it a bit more too and realized that I can use the `collate_fn` argument in DataLoader to do this:

``` python3
def collate_fn(batch):
    l, i, t = zip(*sorted((len(x), i, (x, y)) for i, (x, y) in enumerate(batch)), reverse=True))
    x, y = zip(*t)
    return (torch.LongTensor(np.array(l)), pack_sequence(x)), y

DataLoader(..., collate_fn=collate_fn)
```

Thanks for your help!

PR #225, opened 2018-08-06
Add batch preprocessing option

kl2792:
In reference to #224 -- allows `pack_sequence` preprocessing capabilities.


alykhantejani:
As discussed in #224 I think this can be incorporated in your data loading pipeline and doesn't belong in this function

Issue #232, opened 2018-08-09
[Docs] Add a page "Examples" to the docs


vfdev-5:
Idea is to add a page "Examples" to the docs regrouping links to existing examples. Otherwise users have not direct reference on ignite portfolio

PR #234, opened 2018-08-13
Add a page "Examples" to the docs #232

vfdev-5:
Addresses #232 

- Add page mentioning examples in github 
- update links
- improve visualization of side-by-side image when click on it

### DIFF
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -43,7 +43,7 @@ the run:
 - *epoch is started/completed*
 - *batch iteration is started/completed*
 
-Complete list of events can be found `here <https://github.com/pytorch/ignite/blob/master/ignite/engine/engine.py#L13>`_.
+Complete list of events can be found at :class:`ignite.engine.Events`.
 
 Thus, user can execute a custom code as an event handler. Let us consider in more detail what happens when :meth:`Engine.run` is called:
 
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -1,0 +1,15 @@
+Examples
+========
+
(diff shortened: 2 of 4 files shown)


docs/source/concepts.rst
```diff
@@ -43,7 +43,7 @@ the run:
 - *epoch is started/completed*
 - *batch iteration is started/completed*
 
-Complete list of events can be found `here <https://github.com/pytorch/ignite/blob/master/ignite/engine/engine.py#L13>`_.
+Complete list of events can be found at :class:`ignite.engine.Events`.
```
alykhantejani: we should keep the link no?

docs/source/examples.rst
```diff
@@ -0,0 +1,15 @@
+Examples
+========
+
+There are several examples ported from `pytorch/examples <https://github.com/pytorch/examples>`_ using `ignite`
+to display how it helps to write compact and full-featured training loops in a few lines of code:
+
+- Mnist
```
alykhantejani: maybe add individual links here - will be a nicer user experience

docs/source/concepts.rst
vfdev-5: I think this is bad to keep track of the line in the repository as we could add some more code in the file and this link will become obsolete. 
This rst notation :class:`ignite.engine.Events` creates a link to the doc like https://pytorch.org/ignite/engine.html#ignite.engine.Events

docs/source/examples.rst
vfdev-5: Good point! Will add individual links to the examples

docs/source/concepts.rst
alykhantejani: ah ok that's  perfect then

Issue #207, opened 2018-07-03
[Feature Request] Epoch-wise metric computation structure 


vfdev-5:
Following the discussion in #168, here idea is to provide a base structure to compute a metric cumulating model's output during whole epoch. 

However, model's output and targets are restricted : 
-  to be of shape `(batch_size, n_classes)` or `(batch_size, )` <=> avoid to store a lot of data
-  if target shape is `(batch_size, n_classes)` and `n_classes > 1` than it should be binary: e.g. `[[0, 1, 0, 1], ]`  <=> a use-case for mAP 

The implementation is rather straight forward, I think this should be an abstract class like `Metric` but it implements the method `update`:
```python
class EpochMetric(Metric):

    def reset(self):
        self._predictions = torch.tensor([], dtype=torch.float32)
        self._targets = torch.tensor([], dtype=torch.long)

    def update(self, output):
        y_pred, y = output

        assert 1 <= y_pred.ndimension() <= 2, "Predictions should be of shape (batch_size, n_classes)"
        assert 1 <= y.ndimension() <= 2, "Targets should be of shape (batch_size, n_classes)"

        if y.ndimension() == 2:
            assert torch.equal(y ** 2, y), 'Targets should be binary (0 or 1)'

        if y_pred.ndimension() == 2 and y_pred.shape[1] == 1:
            y_pred = y_pred.squeeze(dim=-1)

        if y.ndimension() == 2 and y.shape[1] == 1:
            y = y.squeeze(dim=-1)

        y_pred = y_pred.type_as(self._predictions)
        y = y.type_as(self._targets)

        self._predictions = torch.cat([self._predictions, y_pred], dim=0)
        self._targets = torch.cat([self._targets, y], dim=0)

    @abstractmethod
    def compute(self):
        pass

```
and user can implement something like this:
```python
from sklearn.metrics import roc_auc_score
from sklearn.metrics import average_precision_score


class AUC(EpochMetric):
    
    def compute(self):
        y_true = self._targets.numpy()
        y_pred = self._predictions.numpy()
        return roc_auc_score(y_true, y_pred)
    
    
class mAP(EpochMetric):
    
    def compute(self):
        y_true = self._targets.numpy()
        y_pred = self._predictions.numpy()
        return average_precision_score(y_true, y_pred)
```
Not sure about `mAP` and `average_precision_score` but anyway.

What do you guys think about this ? 

cc @pkdogcom

jasonkriss:
[...]

PR #235, opened 2018-08-14
Epoch metric and contrib ROC_AUC/AveragePrecision metrics

vfdev-5:
Addresses #207 

Added `EpochMetric` with `compute_fn` as discussed in the issue. Signature checking is done once in the update function.

In the contrib module added ROC_AUC and AveragePrecision metrics. Added activation option if user would like to apply `F.softmax` or something else.

@jasonkriss @pkdogcom Could you please review ?

Thanks


jasonkriss:
I know its in contrib but i'm not a fan of bringing in sklearn as a dependency. I'd personally rather ROC_AUC and AveragePrecision exist as examples (until we implement them in pytorch itself). Up to you guys though.

I think we can/should eventually refactor a lot of the existing metrics to use EpochMetric. We can worry about that later though.

Other than that, Looks great to me! Thanks @vfdev-5!

vfdev-5:
@jasonkriss yes you're right about `sklearn` as dependency. Me too, I did not want to add it as new dependency to ignite. I thought that in the contrib module we could have more dependencies vs core ignite so it is up to user to have necessary modules and I forgot to check sklearn import.
What do you think about just to do 
```python
try:
    from sklearn.metrics import average_precision_score
except ImportError:
    raise RuntimeError("This contrib module requires sklearn to be installed") 
```

Concerning 
> I think we can/should eventually refactor a lot of the existing metrics to use EpochMetric. 

I'm not that sure about this, because `EpochMetric` metric imposes types on the input `y_pred` and `y_true` and then stores all the data. If user uses precision metric on images, than EpochMetric's behaviour will store entire dataset of masks...

pkdogcom:
It looks great to me! Thanks @vfdev-5

jasonkriss:
@vfdev-5 on the refactor suggestion... Good point! You are absolutely correct. My mistake!

On the sklearn stuff... Personally I'd rather not even import sklearn in this project. Even if its only in contrib and even if its not a hard dependency. IMO, the correct thing to do would be to implement these metrics using pytorch. That being said, until we have somebody take that on, EpochMetric allows users to easily piggyback on sklearn metrics. And realistically there will always be metrics that people want that ignite does not provide. So EpochMetric will always be a useful helper. But I don't think the actual combination of sklearn/EpochMetric should be part of ignite. I think an example showing how they can be used together is the better way to go. But again, this is pretty subjective and ultimately just my personal opinion.

@vfdev-5 @alykhantejani What do you guys think?

vfdev-5:
@jasonkriss I see your point and I agree that core ignite should remain simple and less 3rd party dependent. 
However, I think a combination of sklearn/EpochMetric in an example showing how they can be used together will lead to a copy-pasting pattern. I'd prefer to avoid such behaviour.
Similarly I think a lot of users already copy-paste the way to integrate tensorboard displayed in the examples. In this sense I would like that contrib provides more useful features and integrates with useful 3rd parties.

I think we need also to discuss somewhere what will be the purpose of the contrib module in general.
Maybe we could make contrib as another python module that user can install like `pip install ignite[contrib]`

vfdev-5:
@alykhantejani what is your opinion ?

ignite/metrics/metric.py
```diff
@@ -1,5 +1,7 @@
 from abc import ABCMeta, abstractmethod
 
+import torch
```
alykhantejani: is this import necessary? It's already imported in epoch_metric

tests/ignite/metrics/test_epoch_metric.py
```diff
@@ -0,0 +1,93 @@
...
+    with pytest.raises(AssertionError):
+        output = (torch.rand(4, 3, 1), torch.rand(4, 3))
+        em.update(output)
+
+    # Target is not binary
+    with pytest.raises(AssertionError):
+        output = (torch.rand(4, 3), torch.randint(0, 5, size=(4, 3)))
+        em.update(output)
+
+    torch.manual_seed(12)
```
alykhantejani: why is this needed? The compute function currently does nothing

tests/ignite/metrics/test_epoch_metric.py
```diff
@@ -0,0 +1,93 @@
...
+    em.update(output1)
+    output2 = (torch.rand(4, 3), torch.randint(0, 2, size=(4, 3), dtype=torch.long))
+    em.update(output2)
+
+    assert em._predictions.device.type == 'cpu' and em._targets.device.type == 'cpu'
+    assert torch.equal(em._predictions[:4, :], output1[0])
+    assert torch.equal(em._predictions[4:, :], output2[0])
+    assert torch.equal(em._targets[:4, :], output1[1])
+    assert torch.equal(em._targets[4:, :], output2[1])
+    assert em.compute() == 0.0
```
alykhantejani: might be better to make compute_fn a mock and assert its called with the expected tensors

alykhantejani: ah I see you have some examples  like this below. So this is not so important here

alykhantejani:
@vfdev-5 This PR is great - thanks! I think its ok for now to leave the sklearn dependency in contrib for now. Between now and 0.2 we can see how many more contrib metrics get added, and we can decide before 0.2 whether its worth just adding sklearn as a dependency vs implementing them in pytorch (I suspect it might just be worth having sklearn as a dependency on core if we get a ton of metrics for free (and bugfree))

alykhantejani:
Left a few nit comments - happy for you to merge once you've addressed/decided not to address them

vfdev-5:
@alykhantejani I fixed the nits you commented. I also added contrib dependency in .travis and the following check on import of sklearn in both contrib metric modules:
```
try:
    from sklearn.metrics import average_precision_score
except ImportError:
    raise RuntimeError("This contrib module requires sklearn to be installed") 
```

tests/ignite/metrics/test_epoch_metric.py
vfdev-5: okay, so i ignore this

ignite/contrib/metrics/average_precision.py
```diff
@@ -0,0 +1,22 @@
...
+def average_precision_compute_fn(y_preds, y_targets, activation=None):
+    y_true = y_targets.numpy()
+    if activation is not None:
+        y_preds = activation(y_preds)
+    y_pred = y_preds.numpy()
+    return average_precision_score(y_true, y_pred)
+
+
+class AveragePrecision(EpochMetric):
+
```
vfdev-5: I'll some doc string to contrib metrics.

vfdev-5:
@alykhantejani I'm still working on the docs, so please do no merge.

vfdev-5:
@alykhantejani I would say it OK with the docs. Please review the doc strings :)

alykhantejani:
Thanks, Victor!

Issue #231, opened 2018-08-07
[Feature] Possibility to pass `non_blocking` attribute to `convert_tensor`


vfdev-5:
Following torch docs on [CUDA memory management](https://pytorch.org/docs/stable/notes/cuda.html#use-pinned-memory-buffers) if user configures DataLoader with `pin_memory=True` then
he/she can use asynchronous GPU copies, just passing an additional `non_blocking=True` argument to a `cuda()` call or tensor method [`.to()`](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.to).

In ignite this can be done with utils method `convert_tensor` that accepts `device` argument: `cpu`, `cuda` or `cuda:N`. It would be great to have a possibility to pass `non_blocking` as kwargs.

In addition, this should be also propaged until `create_supervised_trainer` and `_prepare_batch`.

PS: it is okay to call `x.to('cpu', non_blocking=True)`, no error raised.

vfdev-5:
@anmolsjoshi could you please take a look on this one, once #220 is merged ?

anmolsjoshi:
@vfdev-5 sure!

alykhantejani:
Fixed in #239

vfdev-5:
Forgot to say that `create_supervised_evaluator` needed this option too. Will add by myself

PR #239, opened 2018-08-17
Added non_blocking capability to convert_tensor, prepare_batch, create_supervised_trainer

anmolsjoshi:
Pull request for this [issue](https://github.com/pytorch/ignite/issues/231)

Adds **kwargs to prepare_batch, create_supervised_trainer, convert_tensor

convert_tensor has a check for non_blocking in kwargs, otherwise assigns False.

### DIFF
--- a/ignite/engine/__init__.py
+++ b/ignite/engine/__init__.py
@@ -4,12 +4,13 @@ from ignite.engine.engine import Engine, State, Events
 from ignite._utils import convert_tensor
 
 
-def _prepare_batch(batch, device=None, **kwargs):
+def _prepare_batch(batch, device=None, non_blocking=False):
     x, y = batch
-    return convert_tensor(x, device=device, **kwargs), convert_tensor(y, device=device, **kwargs)
+    return (convert_tensor(x, device=device, non_blocking=non_blocking),
+            convert_tensor(y, device=device, non_blocking=non_blocking))
 
 
-def create_supervised_trainer(model, optimizer, loss_fn, device=None, **kwargs):
+def create_supervised_trainer(model, optimizer, loss_fn, device=None, non_blocking=False):
     """
     Factory function for creating a trainer for supervised models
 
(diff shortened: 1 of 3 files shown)


vfdev-5:
There tests that are missing. That's why CI fails.

vfdev-5:
@anmolsjoshi may I ask you why you think it is better to provide kwargs instead of just `non_blocking` attribute ?

anmolsjoshi:
@vfdev-5 thanks for your comments. I misunderstood the discussion on the issue about using kwargs, reading it now it was meant as more of a general statement. I have incorporated that comment. I had initially written it without kwargs, sorry! 

Is there a method to test non_blocking?

vfdev-5:
@anmolsjoshi  no problem, I just wonder what else can we pass in kwargs to the method `.to()` 
```
to(device=None, dtype=None, non_blocking=False) → Tensor
```
[ref](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.to)
We are already passing device, about dtype I'm not sure and remains `non_blocking` :)

> Is there a method to test non_blocking?

actually you just need to have tests that passes all cases of 
```python
    if 'non_blocking' in kwargs:
        non_blocking = kwargs.pop('non_blocking')
    else:
        non_blocking = False
```
so just one test with `non_blocking=True` and another with `non_blocking=None` in your current implementation.

anmolsjoshi:
@vfdev-5 given that only other parameters we pass to .to() is device, dtype and non_blocking. I'm going to change the code to pass non_blocking directly as previously suggested.

ignite/engine/__init__.py
```diff
@@ -4,12 +4,13 @@
...
-def _prepare_batch(batch, device=None):
+def _prepare_batch(batch, device=None, non_blocking=False):
     x, y = batch
-    return convert_tensor(x, device=device), convert_tensor(y, device=device)
+    return (convert_tensor(x, device=device, non_blocking=non_blocking),
+            convert_tensor(y, device=device, non_blocking=non_blocking))
 
 
-def create_supervised_trainer(model, optimizer, loss_fn, device=None):
+def create_supervised_trainer(model, optimizer, loss_fn, device=None, non_blocking=False):
```
alykhantejani: Can we add `non_blocking` to the docstring

anmolsjoshi: Incorporated and committed.

alykhantejani:
I guess this build fail is down to visdom on travis @vfdev-5 ?

vfdev-5:
@alykhantejani yes you are right. I added an issue on this #241

vfdev-5:
Sorry for the delay with failed CI. As it is fixed now, we can merge the PRs.
@anmolsjoshi thanks !

PR #242, opened 2018-08-20
docstring fix #237

### LINKED ISSUE #237: Update the "Concepts" part in the documentation
Recently we've added a possibility to handle custom events, so we need to keep "Concepts" documentation updated.

There is also a bad docstring rendering of parameters on methods
- [add_event_handler](https://pytorch.org/ignite/engine.html#ignite.engine.Engine.add_event_handler)
- [fire_event](https://pytorch.org/ignite/engine.html#ignite.engine.Engine.fire_event)
- [on](https://pytorch.org/ignite/engine.html#ignite.engine.Engine.fire_event)
- [register_events](https://pytorch.org/ignite/engine.html#ignite.engine.Engine.register_events)

rpatrik96:
- ModelCheckpoint: added missing argument

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -42,6 +42,9 @@ class ModelCheckpoint(object):
             in the directory 'dirname'
         create_dir (bool, optional):
             If True, will create directory 'dirname' if it doesnt exist.
+        save_as_state_dict (bool, optional):
+            If True, will save only the `state_dict` of the object specified in the dict argument,
+            otherwise the whole object will be saved.
 
     Notes:
           This handler expects two arguments: an `Engine` object and a `dict`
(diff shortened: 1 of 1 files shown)


ignite/handlers/checkpoint.py
```diff
@@ -42,6 +42,9 @@ class ModelCheckpoint(object):
             in the directory 'dirname'
         create_dir (bool, optional):
             If True, will create directory 'dirname' if it doesnt exist.
+        save_as_state_dict (bool, optional):
+            If True, will save only the `state_dict` of the object specified in the dict argument,
```
alykhantejani: Thanks - it's a bit confusing can we change it to
```
save_as_state_dict (bool, optional):
            If True, will save only the `state_dict` of the objects specified ~~in the dict argument~~, otherwise the whole object will be saved
```

rpatrik96: @alykhantejani The fix is on the way. But did You really mean ~~in the dict argument~~?

vfdev-5: @rpatrik96 yes, remove `in the dict argument`

rpatrik96: @vfdev-5 Ok, thanks! I was not 100% sure.

alykhantejani:
Thanks @rpatrik96!

Issue #248, opened 2018-08-23
Automatic deployment on tag


vfdev-5:
Travis with PyPI: https://docs.travis-ci.com/user/deployment/pypi/

vfdev-5:
PyPI wheels+tars build and upload can be done as following:
```
# PyPI Deployment: https://docs.travis-ci.com/user/deployment/pypi/
deploy:
  provider: pypi
  user: vfdev-5
  # If password contains non alphanumeric characters
  # https://github.com/travis-ci/dpl/issues/377
  # pass it as secured variable
  password: $PYPI_TOKEN
  # otherwise, follow "How to encrypt the password": https://docs.travis-ci.com/user/encryption-keys/
  # `travis encrypt deploy.password="password"`
#    secure: "secured_password"

  # !!! This can be removed if deploy on PyPI
  server: https://test.pypi.org/legacy/
  skip_cleanup: true
  distributions: "sdist bdist_wheel"
  on:
    tags: true
    python: "3.5"
```
A test result is [here](https://test.pypi.org/manage/project/pytorch-ignite/release/0.1.0.2/)


Note that user and password should be specified and password is passed as secure Travis variable. I did not manage to handle password encryption.

Conda build and upload can be done in the phase "after_deploy". Need to be scripted and tested

vfdev-5:
Example projects 
- [tnt deploys on PyPI with travis](https://github.com/pytorch/tnt/blob/master/.travis.yml)
- [bokeh build conda](https://github.com/bokeh/bokeh/tree/master/conda.recipe) and [deploy](https://github.com/bokeh/bokeh/blob/master/scripts/ci/deploy)

PR #249, opened 2018-08-23
Add auto PyPI deployement on tag

vfdev-5:
Partly addresses #248 to upload wheels+tars on PyPI


vfdev-5:
Will try out another way to deploy on PyPI and Conda

Issue #202, opened 2018-06-22
[Feature Request] More examples


karandwivedi42:
It would be good to have more examples from the official [examples repo](https://github.com/pytorch/examples) reproduced in ignite.

Specifically, I would like to see if ignite can lead to compact but full-featured code in:

1. Imagenet
2. Reinforcement Learning
3. Neural Style

before I feel confident enough to use it.

alykhantejani:
Thanks for the suggestion. I think more examples would be great and something we'd like to get to. Of course  PR's are welcome!

vfdev-5:
@karandwivedi42 good proposition!

Personnally, I started some more examples [here](https://github.com/vfdev-5/ignite-examples) and find out that classification tasks can be easily implemented, so ImageNet example won't be very different from MNIST. 
Concerning detection or segmentation examples, we will need to rewrite all the stuff: data augmentation, data transformations (bounding box encoding/decoding) around the task depending on the architecture used, e.g. Faster-RCNN, RetinaNet for detection or a basic UNet for segmentation.

I do not know whether RL examples and Neural Style should suffer of the same "problem"...

alykhantejani:
with regards to detection, I think the transforms etc. should live in `torchvision` or a detection specific package. Beyond that it's just a supervised training task I believe so don't see any difficulty of using ignite for this.

Although, I still think examples of how to use ignite for these would be cool

vfdev-5:
@alykhantejani  I agree. We can create github project on this issue.

And maybe can port to ignite more examples from [pytorch/examples](https://github.com/pytorch/examples), where there are imagenet, RL...

elanmart:
Language modelling example from `pytorch/examples` could use a refactor...

jasonkriss:
Agreed that this sounds like a good idea. Neural Style, detection, Imagenet etc should work no problem. The only thing we may want is some more task-specific metrics. RL will probably not work super well, though it will depend on the algorithm used.

vfdev-5:
In a previous issue users asked about hyperparameter tuning, maybe we could also provide an example using [ray library](http://ray.readthedocs.io/en/latest/tune.html). 
Upvote if you think this can be interesting.

elanmart:
I'll add a language modelling example hopefully this week

vfdev-5:
[...]

Issue #64, opened 2018-02-07
[Feature Request] Templates for logging/plotting


pkdogcom:
I think although right now the logging/plotting are straightforward to add using the event handler as in the examples, it can potentially be made even easier by adding some templates or some level of abstractions as logging/plotting are usually pretty standard and most people simply plot mostly the same types of loss, accuracy, learning rates, weights, etc. for the same type of task. 

For example, there can be a logging/plotting class which uses the low-level event handler API underlyingly but abstracts away implementation details like type of logger (file/tensorboard/wisdom), logging formating and provide standard implementation for certain common logging/ploting handlers like in the ignite examples or even task like classification/detection/regression by wrapping up logging/plotting handlers as a package, while still allows users to customize the behaviors of certain handlers or add extra handlers.

alykhantejani:
@pkdogcom sounds like an interesting idea, what do you envisage the API looking like?

pkdogcom:
I think the code snippets elanmart provides in #63 look like a good idea to me. And if the user wants to customize certain behaviors, they can either use the low-level event handler directly or subclass the template class and override certain types of logging/plotting inside.

alykhantejani:
Agreed, this can also live in the contrib area

Robin-des-Bois:
Is anyone working on this? If not I might give it a go, as I already have written some related code for myself.

vfdev-5:
@Robin-des-Bois thanks for the proposition. Yes, I think you can work on this issue.
I think @pkdogcom talks about [`Timer`](https://github.com/pytorch/ignite/blob/master/ignite/handlers/timing.py#L9). 

I think a template for logging with Tensorboard can be useful, e.g. metrics. The difficulty is that it can be rather difficult to cover a generic data. 

Anyway let's discuss what can we add to ignite to simpify the workflow.

anmolsjoshi:
Closing this in favor of #455

PR #253, opened 2018-08-31
Example of MNIST with tqdm

rave78:
I have updated the basic example of MNIST with tqdm.  This example has two progress bars:  one progress bar counts the iterations whereas the other progress bar counts the epoch.


vfdev-5:
@rave78 could you please fix flake8 problems in the modifications: https://travis-ci.org/pytorch/ignite/jobs/423061955

vfdev-5:
Concerning the code itself, maybe it would be more interesting to replace `print` operations by tqdm.

rave78:
Hi @vfdev-5,  
There is not print, only tqdm.write

vfdev-5:
@rave78 sorry, I wanted to say to replace `mnist.py` example by `mnist-with-tqdm.py`, replace `print`s by tqdm...

vfdev-5:
@rave78 tested your modification and gives me something like 
screen shot 2018-09-02 at 23 34 13

Could you please integrate "Loss: {:.2f}" into the progress bar and remove `tqdm.write` ?

rave78:
@vfdev-5 sure, I can do those changes.  I will try to PR today

vfdev-5:
@rave78 thanks, and please do not forget to add `tqdm` [here](https://github.com/pytorch/ignite/blob/master/.travis.yml#L28) as dependency

examples/mnist/mnist.py
```diff
@@ -58,30 +60,64 @@ def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
                                                      'nll': Loss(F.nll_loss)},
                                             device=device)
 
+    @trainer.on(Events.ITERATION_STARTED)
+    def log_training_loss_start(engine):
+        iter = (engine.state.iteration - 1) % len(train_loader) + 1
+        if iter == 1:
+            # engine.state.pdbar_iter = tqdm(
```
vfdev-5: Maybe we could remove commented code

examples/mnist/mnist-with-tqdm.py
```diff
@@ -0,0 +1,131 @@
+from __future__ import print_function
```
vfdev-5: Maybe we can remove this file ?

examples/mnist/mnist.py
rave78: Sure, I will do it now (both changes)

examples/mnist/mnist-with-tqdm.py
vfdev-5: @rave78 here I meant to remove the file `examples/mnist/mnist-with-tqdm.py` and not the line 
`from __future__ import print_function`. But you are right we do not need the same line in `mnist.py` neither

examples/mnist/mnist.py
```diff
@@ -58,30 +59,61 @@ def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
                                                      'nll': Loss(F.nll_loss)},
                                             device=device)
 
+    @trainer.on(Events.ITERATION_STARTED)
+    def log_training_loss_start(engine):
```
vfdev-5: @rave78 I think it would be better have a single log function `log_training_loss` instead of two `log_training_loss_start` and `log_training_loss_end`. What do you think ?

rave78: I think that makes total sense to merge both functions.  There is no difference starting the tqdm pbar at the ITERATION_STARTED or ITERATION_COMPLETED.

vfdev-5:
@rave78 sorry about to request you modifications again and again, but if I look the [code difference : 
I have an impression that tqdm integration makes basic example with `print` more complex with tqdm. 
The modifications with tqdm specific handler function that you propose may perfectly be integrated as a separate tqdm handler in contrib, but here I think we should keep things as simple as with `print`. 
Let me define what we would like to keep and what we can replace : 
- please keep the same number of handlers. For example `pbar_epoch_start(engine)` should be avoided
- We can just use the progress bar for iterations (instead of a stack of lines as previously) and training/validation results can be printed with `tqdm.write` ?

What do you think ?

rave78:
@vfdev-5 don't worry.  I am happy contributing with a few lines of code :).  Yes, it is possible to keep the original structure of mnist-example (in terms of handler functions).  Removing pdbar_epoch and using tqdm.write simplify a lot the code.  I will push those changes.

examples/mnist/mnist.py
```diff
@@ -61,27 +62,39 @@ def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
     @trainer.on(Events.ITERATION_COMPLETED)
     def log_training_loss(engine):
         iter = (engine.state.iteration - 1) % len(train_loader) + 1
-        if iter % log_interval == 0:
-            print("Epoch[{}] Iteration[{}/{}] Loss: {:.2f}"
-                  "".format(engine.state.epoch, iter, len(train_loader), engine.state.output))
+        desc = "ITERATION - loss: {:.2f}".format(engine.state.output)
+
+        if iter == 1:
```
vfdev-5: @rave78 could we avoid if check like this ?

vfdev-5:
@rave78 this looks much better. I left a comment on explicit `if iter == 0` checking

examples/mnist/mnist.py
rave78: Another option is to use `if not hasattr(engine.state, 'pdbar_iter')` instead of `if iter == 1`.  Note that `tqdm` is required for the first iteration for each epoch and `tqdm.close()` for the last iteration.

vfdev-5: is it necessary to add `pdbar_iter` to `engine.state` ? Maybe you can define a `pbar = tqdm(...)` outside the handler like it is done for the [writer in tensorboard example](https://github.com/pytorch/ignite/blob/master/examples/mnist/mnist_with_tensorboardx.py#L80).

rave78: I see.  Yes, it is possible reducing more the interaction between `tqdm` and the handler.

examples/mnist/mnist.py
```diff
@@ -58,32 +59,47 @@ def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
...
         metrics = evaluator.state.metrics
         avg_accuracy = metrics['accuracy']
         avg_nll = metrics['nll']
-        print("Validation Results - Epoch: {}  Avg accuracy: {:.2f} Avg loss: {:.2f}"
-              .format(engine.state.epoch, avg_accuracy, avg_nll))
+        tqdm.write(
+            "Validation Results - Epoch: {}  Avg accuracy: {:.2f} Avg loss: {:.2f}"
+            .format(engine.state.epoch, avg_accuracy, avg_nll))
+
+        pdbar.n = log_interval
```
vfdev-5: @rave78 I wonder why we need the following code ?
```
pdbar.n = log_interval
pdbar.last_print_n = log_interval
pdbar.refresh()
```
And a nit: why `pdbar` and not just `pbar` ?

vfdev-5:
I ran the code it looks good:
screen shot 2018-09-04 at 13 51 52

examples/mnist/mnist.py
rave78: @vfdev-5 the three lines of code set the starting state for the first iteration of the next epoch.  It is not a strong argument to use `pdbar` instead of `pbar`

examples/mnist/mnist.py
```diff
@@ -58,32 +59,47 @@ def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
                                                      'nll': Loss(F.nll_loss)},
                                             device=device)
 
+    desc = "ITERATION - loss: {:.2f}"
+    pdbar = tqdm(
+        initial=log_interval, leave=False, total=len(train_loader),
```
vfdev-5: Shouldn't we this `initial=log_interval` and start from 0 ?

examples/mnist/mnist.py
vfdev-5: I would prefer to to compact this into 
```python
pdbar.n = pdbar.last_print_n = 0
```
starting from 0 instead of `log_interval`

examples/mnist/mnist.py
rave78: Yes, you are right.  I have set the initial value to zero.

examples/mnist/mnist.py
rave78: I have done both changes

vfdev-5:
@rave78 LGTM thanks! I let @alykhantejani comment and merge if it is OK

rave78:
@vfdev-5  sure :D.  Is there any other task that I can collaborate with?

vfdev-5:
@rave78 thanks for the proposition ! Let's say we have two principal [projects](https://github.com/pytorch/ignite/projects) now : contrib module and portfolio examples.
Please take a look [here](https://github.com/pytorch/ignite/issues/64) or [here](https://github.com/pytorch/ignite/issues/202).
- a TensorboardLogger handler in contrib could be of interest. Something similar to #256. Main difficulty is to be enough generic and be simple :)
- port an example from pytorch/examples or something more "interesting" like segmentation example.
- we are also trying to find a solution to `create_supervised_*` methods to be able to configure `_prepare_batch`. Let me create an issue for that where I explain what we would like to have

alykhantejani:
@rave78 thanks for the PR - sorry for the delay in merging!

PR #256, opened 2018-09-04
TQDM Progress Bar

miguelvr:
adds a tqdm progress bar to the contrib module

### DIFF
--- a/ignite/contrib/handlers.py
+++ b/ignite/contrib/handlers.py
@@ -1,0 +1,34 @@
+from tqdm import tqdm
+
+
+class ProgressBar:
+    def __init__(self, loader, metrics):
+        self.num_iterations = len(loader)
+        self.pbar = None
+        self.metrics = metrics
+        self.alpha = 0.98
+
+    def _calc_running_avg(self, engine):
+        for k, v in engine.state.output.items():
+            old_v = self.metrics.get(k, v)
+            new_v = self.alpha * old_v + (1 - self.alpha) * v
+            self.metrics[k] = new_v
+
+    def _reset(self):
+        self.pbar = tqdm(
+            total=self.num_iterations,
+            bar_format='{desc}[{n_fmt}/{total_fmt}] {percentage:3.0f}%|{bar}{postfix} [{elapsed}<{remaining}]')
+
+    def _format_metrics(self):
+        formatted_metrics = {}
+        for key in self.metrics:
+            formatted_metrics[key] = '{:.2e}'.format(self.metrics[key])
+        return formatted_metrics
+
+    def __call__(self, engine):
+        if (engine.state.iteration - 1) % self.num_iterations == 0:
+            self._reset()
+        self._calc_running_avg(engine)
+        self.pbar.set_description('Epoch {}'.format(engine.state.epoch))
+        self.pbar.set_postfix(**self._format_metrics())
+        self.pbar.update()
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ readme = read('README.rst')
 
 VERSION = find_version('ignite', '__init__.py')
 
-requirements = ['enum34;python_version<"3.4"', 'torch']
+requirements = ['enum34;python_version<"3.4"', 'torch', 'tqdm']
 
 setup(
     # Metadata


miguelvr:
I just noticed I am assuming a dictionary is returned from the engine output... We might want a more robust solution. I accept suggestions

vfdev-5:
@miguelvr thanks for the PR !
Concerning the code, I think we should keep `setup.py` without `tqdm` dependency and check the import as it is done for [AUC metrics](https://github.com/pytorch/ignite/blob/master/ignite/contrib/metrics/roc_auc.py#L4).
And let me think about the implementation, maybe we would like to separate the display and running average logic.

miguelvr:
Sure, that separation would make sense. But then there's the question of where to same these metrics. Should we access `engine.state.metrics` directly and compute the running average from there?

miguelvr:
Btw, what test suite did you adopt for the repo? I will add some tests as well

vfdev-5:
> Btw, what test suite did you adopt for the repo? I will add some tests as well

It is `pytest`

vfdev-5:
Previously I thought about to simplify the stuff as it is done in [this PR](https://github.com/pytorch/ignite/pull/253).

miguelvr:
@vfdev-5 do you mean the style of the output? I agree it is a bit cleaner... I will try it out

miguelvr:
@vfdev-5 what do you think of this? Basically, the tqdm progress bar stays on top, being replaced after each epoch. At the end of an epoch, the requested metrics are logged with a simple style: 

e.g.
`Epoch 1 | total_loss=5.47e-01 | loss_autoencoder=5.38e-01 | loss_clustering=9.32e-03`

ignite/contrib/handlers.py
```diff
@@ -0,0 +1,68 @@
...
+                into a dictionary of format {name: value}
+
+    Example:
+        (...)
+        pbar = ProgressBar(trainer, train_loader, output_transform=lambda x: {'loss': x})
+        trainer.add_handler(Events.ITERATION_COMPLETED, pbar)
+    """
+
+    def __init__(self, engine, loader, output_transform=lambda x: x):
+        self.num_iterations = len(loader)
```
vfdev-5: nit: maybe we can setup `num_iterations` directly instead of `loader` ?

ignite/contrib/handlers.py
```diff
@@ -0,0 +1,68 @@
...
+        formatted_metrics = {}
+        for key in self.metrics:
+            formatted_metrics[key] = '{:.2e}'.format(self.metrics[key])
+        return formatted_metrics
+
+    def __call__(self, engine):
+        self._calc_running_avg(engine)
+        self.pbar.set_description('Epoch {}'.format(engine.state.epoch))
+        self.pbar.set_postfix(**self._format_metrics())
+        self.pbar.update()
```
rave78: Hi @miguelvr.  I have one quick question.  How are you considering to show the results for a specific condition? Let's say after 10 iterations/epoch.

ignite/contrib/handlers.py
miguelvr: sure, but I don't see the advantage of that... the user will have the loader more easily accessible, and from an efficiency point of view, it is the same

ignite/contrib/handlers.py
miguelvr: I didn't thought about that, but I guess it would be possible to change it here:

`engine.add_event_handler(Events.EPOCH_COMPLETED, self._log_message)`, by changing the event to a custom one triggered every X iterations?

This is the minimal change solution, but I also accept suggestions.

ignite/contrib/handlers.py
vfdev-5: For me if handler require data loader than it will work with its content. So, without looking into the code user can have a misleading a priori on the handler

ignite/contrib/handlers.py
rave78: You are right. I think that it is a good/fast solution

miguelvr: check the latest commit, I think it is an even better solution

ignite/contrib/handlers.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+
+    def __init__(self, engine, loader, output_transform=lambda x: x, mode='epoch', log_frequency=1):
+        self.num_iterations = len(loader)
+        self.metrics = {}
+        self.alpha = 0.98
+        self.output_transform = output_transform
+        self.pbar = None
+        self.mode = mode
+        self.log_frequency = log_frequency
```
vfdev-5: nit: `log_frequency` -> `log_interval`

ignite/contrib/handlers.py
miguelvr: sure, I don't mind going with num_iterations instead but bear in mind that from the user point of view this is more confusing because it's not the total number of iterations, but rather the number of batches in the loader

ignite/contrib/handlers.py
```diff
@@ -0,0 +1,87 @@
...
+
+        assert log_frequency >= 1, 'log_frequency must be positive'
+        assert mode in {'iteration', 'epoch'}, \
+            'incompatible mode {}, accepted modes {}'.format(mode, {'iteration', 'epoch'})
+
+        log_event = Events.EPOCH_COMPLETED if mode == 'epoch' else Events.ITERATION_COMPLETED
+
+        engine.add_event_handler(Events.EPOCH_STARTED, self._reset)
+        engine.add_event_handler(Events.EPOCH_COMPLETED, self._close)
+        engine.add_event_handler(log_event, self._log_message)
```
vfdev-5: I think you have a collision with `engine.add_event_handler(Events.EPOCH_COMPLETED, self._close)`

miguelvr: why? the progress bar is still going to display the information, but it will flush it
I've tested it and it works well

vfdev-5: @miguelvr maybe you are right, just to be sure, did you tested in mode `epoch`?

miguelvr: I've tested both, but I can add another unit test

vfdev-5: Okay, let me play with this handler too :)

vfdev-5:
@miguelvr could you please add a file `handlers.rst` into [docs](https://github.com/pytorch/ignite/tree/master/docs/source/contrib)

miguelvr:
@vfdev-5 sure, on it!

ignite/contrib/handlers.py
```diff
@@ -0,0 +1,87 @@
...
+        log_interval (int): interval of which the metrics information is displayed (default=1)
+
+    Example:
+        (...)
+        pbar = ProgressBar(trainer, train_loader, output_transform=lambda x: {'loss': x})
+        trainer.add_handler(Events.ITERATION_COMPLETED, pbar)
+    """
+
+    def __init__(self, engine, loader, output_transform=lambda x: x, mode='epoch', log_interval=1):
+        self.num_iterations = len(loader)
```
vfdev-5: Coming back to this one. So I run the following code: 
```
trainer = Engine(process_function)
pbar = ProgressBar(trainer, train_loader)
```
by default its epoch mode but I see:
```
[0/105]   0%|           [00:00<?]
```
so 105 is number of batches

miguelvr: that's intentional, because the mode only affects the `tqdm.write()` operation

vfdev-5: Otherwise, here is the code I'm running in the notebook
```python
from __future__ import print_function
from argparse import ArgumentParser

from torch import nn
from torch.optim import SGD
import torch
from torchvision.transforms import Compose, ToTensor, Normalize
from torchvision.datasets import FakeData
from torchvision.models.resnet import resnet34

from ignite.engine import Events, Engine, _prepare_batch
from ignite.metrics import CategoricalAccuracy, Loss, TopKCategoricalAccuracy

from ignite.contrib.handlers import ProgressBar


def get_data_loaders(train_batch_size, val_batch_size):
    data_transform = Compose([ToTensor()])

    train_loader = DataLoader(FakeData(size=10000, image_size=(3, 224, 224), num_classes=10, 
                                       transform=data_transform, target_transform=lambda t: t.long()),
                              batch_size=train_batch_size, shuffle=True)

    val_loader = DataLoader(FakeData(size=10000, image_size=(3, 224, 224), num_classes=10, 
                                     transform=data_transform, target_transform=lambda t: t.long()),
                            batch_size=val_batch_size, shuffle=False)
    return train_loader, val_loader


def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
    train_loader, _ = get_data_loaders(train_batch_size, val_batch_size)
    model = resnet34(pretrained=False, num_classes=10)
    device = 'cpu'

    if torch.cuda.is_available():
        device = 'cuda'

    loss_fn = F.cross_entropy
    
    if device:
        model.to(device)

        model.train()
        optimizer.zero_grad()
        x, y = _prepare_batch(batch, device=device)
        y_pred = model(x)
        loss = loss_fn(y_pred, y)
        loss.backward()
        optimizer.step()
        return [loss.item(), y_pred, y]

    trainer = Engine(process_function)
    
    pbar = ProgressBar(trainer, train_loader)
    
    output_transform = lambda x: [x[1], x[2]]
    metrics={
        'accuracy': CategoricalAccuracy(output_transform=output_transform),
        'topk': TopKCategoricalAccuracy(output_transform=output_transform),
        'xentropy': Loss(loss_fn, output_transform=output_transform)
    }
    for name, metric in metrics.items():
        metric.attach(trainer, name)


    def log_training_loss(engine):
        iter = (engine.state.iteration - 1) % len(train_loader) + 1
#         if iter % log_interval == 0:
#             print("Epoch[{}] Iteration[{}/{}] Loss: {:.2f}"
#                   "".format(engine.state.epoch, iter, len(train_loader), engine.state.output[0]))

    trainer.run(train_loader, max_epochs=epochs)
    print(trainer.state.metrics)
```
and I get 
```
[0/105]   0%|           [00:00<?]
Epoch 1
[0/105]   0%|           [00:00<?]
Epoch 2
[0/105]   0%|           [00:00<?]
Epoch 3
[0/105]   0%|           [00:00<?]
Epoch 4
```

miguelvr: you need to pass the `output_transform` argument:

Not sure what is the format in your code, but this transform is supposed to return a dictionary in the format `{'metric_name1': metric_value1, 'metric_name2': metric_value2, (...)}`

miguelvr: you can even do: `lambda x: {'xentropy': engine.state.metrics['xentropy']}`

miguelvr: also you need to add: 
`trainer.add_handler(Events.ITERATION_COMPLETED, pbar)`

vfdev-5: Okay, let's say the previous example is rather specific. 
Here is another test and I do not understand the intension to the output.
```python
from __future__ import print_function
from argparse import ArgumentParser

from torch import nn
from torch.optim import SGD
import torch
from torchvision.transforms import Compose, ToTensor, Normalize

from ignite.engine import Events, create_supervised_trainer, create_supervised_evaluator
from ignite.metrics import CategoricalAccuracy, Loss

from ignite.contrib.handlers import ProgressBar


class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.conv1 = nn.Conv2d(1, 10, kernel_size=5)
        self.conv2 = nn.Conv2d(10, 20, kernel_size=5)
        self.conv2_drop = nn.Dropout2d()
        self.fc1 = nn.Linear(320, 50)
        self.fc2 = nn.Linear(50, 10)

    def forward(self, x):
        x = F.relu(F.max_pool2d(self.conv1(x), 2))
        x = F.relu(F.max_pool2d(self.conv2_drop(self.conv2(x)), 2))
        x = x.view(-1, 320)
        x = F.relu(self.fc1(x))
        x = self.fc2(x)
        return F.log_softmax(x, dim=-1)


def get_data_loaders(train_batch_size, val_batch_size):

    train_loader = DataLoader(MNIST(download=True, root=".", transform=data_transform, train=True),
                              batch_size=train_batch_size, shuffle=True)

    val_loader = DataLoader(MNIST(download=False, root=".", transform=data_transform, train=False),
                            batch_size=val_batch_size, shuffle=False)
    return train_loader, val_loader


def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_interval):
    model = Net()
    device = 'cpu'

    if torch.cuda.is_available():
        device = 'cuda'

    trainer = create_supervised_trainer(model, optimizer, F.nll_loss, device=device)
                                                     'nll': Loss(F.nll_loss)},
                                            device=device)
    
    pbar = ProgressBar(engine=trainer, loader=train_loader, mode='iteration', log_interval=log_interval)

    def log_training_loss(engine):
        iter = (engine.state.iteration - 1) % len(train_loader) + 1
#         if iter % log_interval == 0:
#             print("Epoch[{}] Iteration[{}/{}] Loss: {:.2f}"
#                   "".format(engine.state.epoch, iter, len(train_loader), engine.state.output))

        evaluator.run(train_loader)
        avg_accuracy = metrics['accuracy']
        avg_nll = metrics['nll']
#         print("Training Results - Epoch: {}  Avg accuracy: {:.2f} Avg loss: {:.2f}"
#               .format(engine.state.epoch, avg_accuracy, avg_nll))

    def log_validation_results(engine):
        evaluator.run(val_loader)
        avg_accuracy = metrics['accuracy']
        avg_nll = metrics['nll']
#         print("Validation Results - Epoch: {}  Avg accuracy: {:.2f} Avg loss: {:.2f}"
#               .format(engine.state.epoch, avg_accuracy, avg_nll))

    trainer.run(train_loader, max_epochs=epochs)
```
and output:
```
[0/938]   0%|           [00:00<?]

Iteration 10
Iteration 20
Iteration 30

[0/938]   0%|           [00:00<?]

Iteration 40
Iteration 50
Iteration 60

[0/938]   0%|           [00:00<?]

Iteration 70
Iteration 80
Iteration 90

[0/938]   0%|           [00:00<?]

Iteration 100
Iteration 110
Iteration 120

[0/938]   0%|           [00:01<?]

Iteration 130
Iteration 140
Iteration 150

[0/938]   0%|           [00:01<?]

Iteration 160
Iteration 170
Iteration 180

[0/938]   0%|           [00:01<?]

Iteration 190
Iteration 200
Iteration 210

[0/938]   0%|           [00:01<?]

Iteration 220
Iteration 230
Iteration 240

[0/938]   0%|           [00:01<?]

Iteration 250
Iteration 260
Iteration 270

[0/938]   0%|           [00:02<?]

Iteration 280
Iteration 290
Iteration 300
```

I would like to see in this case a plain progress barm just showing batch iterations without `tqdm.write`. What do you think ?

miguelvr: you are still not adding the pbar to the trainer

add this line:
`trainer.add_handler(Events.ITERATION_COMPLETED, pbar)`

> I would like to see in this case a plain progress barm just showing batch iterations without tqdm.write. What do you think?

I could check if `log_interval` is `None`, and if so I won't output the `tqdm.write()` message. Would this do?

ignite/contrib/handlers.py
```diff
@@ -0,0 +1,91 @@
...
+                into a dictionary of format {name: value}
+        mode (str): 'iteration' or 'epoch' (default=epoch)
+        log_interval (int or None): interval of which the metrics information is displayed.
+                            If set to None, only the progress bar is shown and not
+                            the metrics. (default=1)
+
+    Example:
+        (...)
+        pbar = ProgressBar(trainer, train_loader, output_transform=lambda x: {'loss': x})
+        trainer.add_handler(Events.ITERATION_COMPLETED, pbar)
```
vfdev-5: `add_handler` -> `add_event_handler`

ignite/contrib/handlers.py
vfdev-5: > but for both modes, the triggered event should still be ITERATION_COMPLETED

If it is necessary part, you can add it directly in the `__init__` ?

miguelvr: yes, I can, but is that good practice?

vfdev-5:
Just remarked that this progress bar in the file `handlers.py`, I think it would be better to have a folder and a file `tqdm_logger.py` or similar.

miguelvr:
If we remove it from `handlers.py`. I could make a `pbar.attach()` method, just like what you do with the metrics, and add the `__call__` method to the `Events.ITERATION_COMPLETED` there as well. What do you think?

vfdev-5:
In the previous message I mean to have a module `handlers` (as in core `ignite`) and a separate file `*tqdm*logger.py` (whatever easy to understand the purpose).

> I could make a pbar.attach() method, just like what you do with the metrics, and add the __call__ method to the Events.ITERATION_COMPLETED there as well.

I would like a solution that is simple of usage, like
```python
trainer = ...
pbar = ProgressBar(trainer, mode="iterations", log_interval=100)
```
what do you gain by calling `attach` ?

miguelvr:
@vfdev-5 I just don't like the idea of having an unused variable in my code. That is why doing the explicit call to `trainer.add_event_handler(Events.ITERATION_COMPLETED, pbar)` still sounds like the best approach to me

vfdev-5:
@miguelvr I see your point, but I thought that we could use pbar for other engines (e.g. evaluator). 

Okay, let's redefine the purpose of this handler. IMO, idea is to provide a more simple solution for [this](https://github.com/pytorch/ignite/pull/253/files). So progress bar iterates over batches with `log_interval` for printing and HMI. I'm not sure whether it should be the same bar for evaluator or different (would be better if it is the same).

miguelvr:
As it is, it really needs to be a different one, not only because of the `engine` argument but also because of the `metrics` attribute.

miguelvr:
For the evaluator, you probably want a slightly different progress bar anyway, and thus the `output_transform` argument would also be different.

vfdev-5:
Another remark on metrics (a part of average running computation), it can happen that a metric value is not just a float, but a `torch.tensor` (e.g. categorical accuracy per class), we need to think about how to cover a generic metric value

miguelvr:
@vfdev-5 what we could also do would be to add another argument to process the metrics, but default it to the current method. That way, if you need something more advanced, you can always override it.

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -0,0 +1,92 @@
...
+
+from ignite.engine import Events
+
+
+class ProgressBar:
+    """
+    TQDM progress bar handler to log training progress and computed metrics
+
+    Args:
+        engine (ignite.Engine): an engine object
```
vfdev-5: I think `ignite.Engine` -> `Engine`. Please check by rendering the docs:
```
cd docs && make html && cd build && python3 -m http.server 1234
```

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -0,0 +1,92 @@
...
+                into a dictionary of format {metric_name: metric_value}
+        mode (str): 'iteration' or 'epoch' (default=epoch)
+        log_interval (int or None): interval of which the metrics information is displayed.
+                            If set to None, only the progress bar is shown and not
+                            the metrics. (default=1)
+
+    Example:
+        (...)
+        pbar = ProgressBar(trainer, train_loader, output_transform=lambda x: {'loss': x})
+        trainer.add_event_handler(Events.ITERATION_COMPLETED, pbar)
```
vfdev-5: What do you think to change a little the API as 
```
pbar = ProgressBar(num_iterations, metric_names)
pbar.attach(trainer, mode='epoch', log_interval=1)
# or
pbar.attach(trainer, mode='iteration', log_interval=10)
```

vfdev-5:
> I was wondering if I should adapt the code so that the user can more easily log metrics without having to define an output_transform

Good idea ! Please go ahead.

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -0,0 +1,92 @@
...
+    def _reset(self, engine):
+        self.pbar = tqdm(
+            total=self.num_iterations,
+            leave=False,
+            bar_format='{desc}[{n_fmt}/{total_fmt}] {percentage:3.0f}%|{bar}{postfix} [{elapsed}<{remaining}]')
+
+    def _close(self, engine):
+        self.pbar.close()
+
+    def _log_message(self, engine):
```
vfdev-5: What do you think about to provide a public method `log_message` that users can use in their custom event to log something particular. For example,
```
pbar = ...

@trainer.on(...)
def log_validation_scores(engine):
    evaluator.run(validation_dataset)
    pbar.log_message("Val loss : {}".format(metrics['loss']))

```

ignite/contrib/handlers/tqdm_logger.py
miguelvr: looks good to me!

miguelvr: but I wouldn't pass `num_iterations` if we are attaching it to the trainer anyway

miguelvr:
@vfdev-5 would you prefer the progress bar itself didn't log any metrics on the `tqdm` and would just use the ``self.log_message` method to write?

vfdev-5:
> @vfdev-5 would you prefer the progress bar itself didn't log any metrics on the tqdm and would just use the self.log_message method to write?

No I mean, I would like the API with `attach` and an additional method to write some validation results:
```
pbar = ProgressBar(num_iterations, metric_names)
pbar.attach(trainer, mode='iteration', log_interval=10)

@trainer.on(...)
def log_validation_scores(engine):
    evaluator.run(validation_dataset)
    pbar.log_message("Val loss : {}".format(metrics['loss']))
```

@miguelvr how do you see to log the results of other engines in a common way ? Or should we use two bars for training and validation engines ?

miguelvr:
@vfdev-5 I did some changes already, let me test and commit them. I made the pbar reusable between engines, but still didn't account for custom log messages. Let me commit first and then we re-evaluate

vfdev-5:
As we discuss here on API of a basic logger, @anmolsjoshi I think it would be good that you take a look too and we provide a common API between tqdm logger and tensorboardx logger (even if the latter is more sophisticated)

vfdev-5:
@miguelvr some more precisions on API (sorry if I repeat already said or implemented):
```python
# metric_names can be None -> only output is printed as running average.
pbar = ProgressBar(num_iterations, metric_names=None)
# if metric_names are provided than output and metric_names are printed as running average
pbar = ProgressBar(num_iterations, metric_names=metric_names)
pbar.attach(trainer, mode='iteration', log_interval=10)
```

miguelvr:
@vfdev-5 this is my suggested API:

```python
pbar = ProgressBar()
pbar.attach(engine=trainer,
            num_iterations=len(data_loader),
            metric_names=['loss'],
            mode='iteration',
            log_interval=1)
```

vfdev-5:
And what about the output of the processing function?

miguelvr:
> And what about the output of the processing function?

do you want to keep that? with the `RunningAverage` metric, it seems pointless

vfdev-5:
Okay, you are right. I see your point.

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -0,0 +1,85 @@
...
+
+            tqdm.write(message)
+
+    def _update(self, engine, metric_names):
+        metrics = {name: '{:.2e}'.format(engine.state.metrics[name]) for name in metric_names}
+        self.pbar.set_description('Epoch {}'.format(engine.state.epoch))
+        self.pbar.set_postfix(**metrics)
+        self.pbar.update()
+
+    def attach(self, engine, num_iterations, metric_names, mode='epoch', log_interval=1):
```
vfdev-5: Just wonder what happens if user attaches to multiple engines:
```python
pbar = ProgressBar()
pbar.attach(trainer1, len(data_loader1), ['loss'], mode='iteration', log_interval=1)
pbar.attach(trainer2, len(data_loader2), ['loss'], mode='iteration', log_interval=1)
```
maybe we need two tqdm progress bars ?

miguelvr: If they are running concurrently, we would need two. If not, there would be no problem.

miguelvr:
@vfdev-5 More flexible proposed API:

Like this you can just attach the progress bar and get real-time feedback on the values you want to monitor and, in addition, you can periodically monitor other metrics with a different mode and log interval.

## Example:

Monitoring the training loss on the progress bar, logging it every 10 iterations, and at the end of each epoch log the average loss and the accuracy.

```python
pbar = ProgressBar()
pbar.attach(trainer, len(data_loader), ['loss'])
pbar.attach_metrics(trainer, ['loss'], mode='iteration', log_interval=10)
pbar.attach_metrics(trainer, ['loss', 'accuracy'], mode='epoch', log_interval=1)
```

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -0,0 +1,113 @@
...
+            engine (Engine): engine object
+            num_iterations (int): number of iterations of one epoch
+            metric_names (list): list of the metrics names to log
+        """
+
+        engine.add_event_handler(Events.EPOCH_STARTED, self._reset, num_iterations)
+        engine.add_event_handler(Events.EPOCH_COMPLETED, self._close)
+        engine.add_event_handler(Events.ITERATION_COMPLETED, self._update, metric_names)
+
+    def attach_metrics(self, engine, metric_names, mode='epoch', log_interval=1):
```
vfdev-5: IMO method name is unclear as `attach` and `atach_metrics` both requires `metric_names`...

miguelvr: I wondered the same but didn't come up with a better one... Suggestions?

vfdev-5:
So, resuming on what provides the API: 
- progress bar display with running values <-> `attach`
- custom periodic printing <-> `attach_metrics`

In case of multiple trainers (for example use-case of GANs) I think we need to cover properly the use-case with multiple progress bars.

ignite/contrib/handlers/tqdm_logger.py
vfdev-5: > Suggestions?

`add_periodic_logging` or samething like that ?

miguelvr: I thought of `log_metrics`, but didn't sound very solid to me

miguelvr:
> So, resuming on what provides the API:
> 
> * progress bar display with running values <-> `attach`
> * custom periodic printing <-> `attach_metrics`
> 
> In case of multiple trainers (for example use-case of GANs) I think we need to cover properly the use-case with multiple progress bars.

Correct. 

You can alternatively to `attach_metrics` define your own logging handler, as long as you use the `pbar.log_message` method (or `tqdm.write`)

For the GANs problem, maybe we can create an example as proof of concept?

ignite/contrib/handlers/tqdm_logger.py
miguelvr: changed it to `add_logging`, sounds good to me this way

vfdev-5:
> For the GANs problem, maybe we can create an example as proof of concept?

@miguelvr  go ahead, there is already [one](https://github.com/pytorch/ignite/blob/master/examples/gan/dcgan.py), we can update it if this progress bar is easily integrated and provide same or more useful info

miguelvr:
> > For the GANs problem, maybe we can create an example as proof of concept?
> 
> @miguelvr go ahead, there is already [one](https://github.com/pytorch/ignite/blob/master/examples/gan/dcgan.py), we can update it if this progress bar is easily integrated and provide same or more useful info

@vfdev-5 done with the last commit

vfdev-5:
@miguelvr  could you please provide a screenshot with new dcgan.py and progresbar?

miguelvr:
> @miguelvr could you please provide a screenshot with new dcgan.py and progresbar?

pbar

vfdev-5:
Looks nice. Do you think we could remove all `print` with functionallities of `pbar` or we suggest to use both ?

miguelvr:
> Looks nice. Do you think we could remove all `print` with functionallities of `pbar` or we suggest to use both ?

we can still use both, it the prints occur after every epoch. I'm going to run the DCGAN with a GPU now, and will test it further

miguelvr:
@vfdev-5 can confirm, the prints look good!

vfdev-5:
IMO it would be nicer to replace this print:
```
    # adding handlers using `trainer.on` decorator API
    def print_times(engine):
        print('Epoch {} done. Time per batch: {:.3f}[s]'.format(engine.state.epoch, timer.value()))
        timer.reset()
```

vfdev-5:
> can confirm, the prints look good!

Can you share a screenshot with multiple epochs and PRINT_FREQ by default, please

miguelvr:
> > can confirm, the prints look good!
> 
> Can you share a screenshot with multiple epochs and PRINT_FREQ by default, please

pbar-gpu

miguelvr:
> IMO it would be nicer to replace this print:
> 
> ```
>     # adding handlers using `trainer.on` decorator API
>     @trainer.on(Events.EPOCH_COMPLETED)
>     def print_times(engine):
>         print('Epoch {} done. Time per batch: {:.3f}[s]'.format(engine.state.epoch, timer.value()))
>         timer.reset()
> ```

I would do that because the function also resets the timer

vfdev-5:
I meant replace only `print` like this
```python
    # adding handlers using `trainer.on` decorator API
    def print_times(engine):
        pbar.log_message('Epoch {} done. Time per batch: {:.3f}[s]'.format(engine.state.epoch, timer.value()))
        timer.reset()
```

miguelvr:
@vfdev-5 done

32 commits and about 80 comments later, would you reckon it is time to merge? :sweat_smile:

vfdev-5:
24 days after the PR submission too :) This PR is the most discussed one 😅

But at least we can be proud of the proposed API which is simple and powerful. Anyway, thank you @miguelvr for your work !

LGTM ! 

@alykhantejani @jasonkriss, your opinion please and if you are OK, merge it

alykhantejani:
@miguelvr Thanks a lot for your efforts on this. It LGTM, but I wonder if we can not just merge the attach and `add_logging` functions? i.e. add `mode` and `log_interval` to `attach` with sensible defaults (`None` and `0`)?

vfdev-5:
@alykhantejani they have a little bit different behaviours:

    tqdm progress bar display with running values <-> `attach`
    custom periodic printing <-> add_logging

miguelvr:
> @miguelvr Thanks a lot for your efforts on this. It LGTM, but I wonder if we can not just merge the attach and `add_logging` functions? i.e. add `mode` and `log_interval` to `attach` with sensible defaults (`None` and `0`)?

@alykhantejani They have different purposes: `attach` attaches the progress bar, displaying the desired metrics on it in real time, while `add_logging` refers to a periodic message logging, that can display a completely different set of metrics. By keeping it separate you have more flexibility, imo.

miguelvr:
@vfdev-5 I just noticed in the generated docs that the index for the `ignite.contrib.handlers` does not stay in the sidebar after I click it. I have no idea how to fix this (first time handling docs).

vfdev-5:
@miguelvr I had this previously, check if there is no errors while `make html` and leave it as is when deployed at https://pytorch.org/ignite/ seems no problem with that.

miguelvr:
> @miguelvr I had this previously, check if there is no errors while `make html` and leave it as is when deployed at https://pytorch.org/ignite/ seems no problem with that.

No errors... It seems to happen only if open a subset of tabs, but not for all of them. I guess it will work fine

alykhantejani:
Thanks for the explanation. LGTM!

jasonkriss:
@miguelvr Thanks a lot for your patience on this! This is gonna be super useful. I have a couple questions/suggestions, though.

Can we maybe remove the need for users to pass the num_iterations? We [set the dataloader on the state object](https://github.com/pytorch/ignite/blob/master/ignite/engine/engine.py#L284) so we should be able to just do a len(engine.state.dataloader). It's one less thing for the user to have to worry about and it removes the possibility of passing in the wrong num_iterations.

I find the metric logging kind of confusing. I understand now, after reading your response to @alykhantejani and reading through the code, but it was not obvious at first. It also is error-prone in its current state. That is, if you pass any metric other than a `RunningAverage` to `attach` it fill fail as the metric will not yet be in engine.state.metrics.

Maybe it's just a documentation issue but I wonder if maybe it'd be better to remove the ability to pass metrics at `attach` time and only support the `add_logging` approach. That is, only support onw way of logging any sort of metric. I think it'd be a lot less confusing. Also, maybe change `add_logging` to `log_metrics`?

What do you all think?

miguelvr:
> Can we maybe remove the need for users to pass the num_iterations? We [set the dataloader on the state object](https://github.com/pytorch/ignite/blob/master/ignite/engine/engine.py#L284) so we should be able to just do a len(engine.state.dataloader). It's one less thing for the user to have to worry about and it removes the possibility of passing in the wrong num_iterations.

Removing the `num_iterations` will be no problem, my original approach was to pass the `dataloader` anyway, but I changed it because @vfdev-5 suggested so. I'm fine with either way. 

> I find the metric logging kind of confusing. I understand now, after reading your response to @alykhantejani and reading through the code, but it was not obvious at first. It also is error-prone in its current state. That is, if you pass any metric other than a RunningAverage to attach it fill fail as the metric will not yet be in engine.state.metrics.

Why is it error prone? You only pass the name of the metric, not the metric itself, so it will retrieve whatever value is saved in `engine.state.metrics`.

> Maybe it's just a documentation issue but I wonder if maybe it'd be better to remove the ability to pass metrics at attach time and only support the add_logging approach. That is, only support onw way of logging any sort of metric. I think it'd be a lot less confusing. 

I think it's just a matter of flexibility. You don't need to log any metric on the progress bar, but it is nice to have the ability to do so.

> Also, maybe change add_logging to log_metrics?

This was the previous name and I changed it because @vfdev-5 said it was confusing the other way around 😅 

Maybe we can improve the docs to make sure the user understand its correct usage? I'm open to suggestions

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -35,19 +35,22 @@ class ProgressBar:
...
         self.pbar.close()
 
-    def _update(self, engine, metric_names):
-        metrics = {name: '{:.2e}'.format(engine.state.metrics[name]) for name in metric_names}
+    def _update(self, engine, metric_names=None):
         self.pbar.set_description('Epoch {}'.format(engine.state.epoch))
-        self.pbar.set_postfix(**metrics)
+
+        if metric_names is not None:
+            metrics = {name: '{:.2e}'.format(engine.state.metrics[name]) for name in metric_names}
```
vfdev-5: Following the comment of @jasonkriss here: `engine.state.metrics[name]` can be the problem if `_update` is called on not `RunningAverage` metric.

miguelvr: but if you set it in any other way, wouldn't you be able to access it? The metric must be set the progress bar, though

miguelvr:
@vfdev-5 @jasonkriss removed the `num_iterations` argument and made the `metric_names` in the attach method to be optional, in order to get a simple progress bar with no postfix

ignite/contrib/handlers/tqdm_logger.py
miguelvr: you got me confused with the `_update` method call, since there is no such method. Also the `update` method in `RunningAverage` doesn't do anything

miguelvr: Ah I got it... you mean the `_update` from the progress bar itself. 

That would not be called if there is no `ITERATION_COMPLETED` event fired...

vfdev-5:
@miguelvr sorry for being confusing about `_update` function. Here the snippet of code to reproduce the problem with non-`RunningAverage` metrics:
```python
import numpy as np
import torch

from ignite.metrics import Loss, CategoricalAccuracy


n_iters = 100
batch_size = 10
n_classes = 10
data = list(range(n_iters))
y_true_batch_values = iter(np.random.randint(0, n_classes, size=(n_iters, batch_size)))
y_pred_batch_values = iter(np.random.rand(n_iters, batch_size, n_classes))
loss_values = iter(range(n_iters))


def update_fn(engine, batch):
    loss_value = next(loss_values)
    return loss_value, torch.from_numpy(y_pred_batch), torch.from_numpy(y_true_batch)


trainer = Engine(update_fn)

accuracy = CategoricalAccuracy(output_transform=lambda x: (x[1], x[2]))
accuracy.attach(trainer, "avg_accuracy")

pbar = ProgressBar()
pbar.attach(trainer, ['avg_accuracy'])
```
and the traceback:
```
<ipython-input-1-96906d255d1a> in _update(self, engine, metric_names)
     49 
     50         if metric_names is not None:
---> 51             metrics = {name: '{:.2e}'.format(engine.state.metrics[name]) for name in metric_names}
     52             self.pbar.set_postfix(**metrics)
     53 

<ipython-input-1-96906d255d1a> in <dictcomp>(.0)
     49 
     50         if metric_names is not None:
---> 51             metrics = {name: '{:.2e}'.format(engine.state.metrics[name]) for name in metric_names}
     52             self.pbar.set_postfix(**metrics)
     53 

KeyError: 'avg_accuracy'
```

vfdev-5:
@miguelvr I tried the progress bar on the above example without attaching any metric. When training is completed, `close` method is called which closes the progress bar and the progress is cleaned up: 
without closing pbar:
screen shot 2018-09-28 at 23 52 16
vs current behaviour:
screen shot 2018-09-28 at 23 52 00

I'm not sure about such behaviour, maybe we could avoid of closing tqdm pbar ?

miguelvr:
> @miguelvr I tried the progress bar on the above example without attaching any metric. When training is completed, `close` method is called which closes the progress bar and the progress is cleaned up:
> without closing pbar:
> screen shot 2018-09-28 at 23 52 16
> vs current behaviour:
> screen shot 2018-09-28 at 23 52 00
> 
> I'm not sure about such behaviour, maybe we could avoid of closing tqdm pbar ?

the default tqdm on notebooks usually doesn't work that well.

miguelvr:
> About `metrics`, it becomes a dictionary in the `state` once `engine.run` is called

how would you solve this then?

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def attach(self, engine, metric_names=None):
+        """
+        Attaches the progress bar to an engine object
+
+        Args:
+            engine (Engine): engine object
+            metric_names (list): (Optional) list of the metrics names to log as the bar progresses
+        """
+
```
vfdev-5: please add a check on the type `list` here. This could avoid silent errors like `pbar.attach(trainer, "avg_accuracy")`

vfdev-5:
@jasonkriss maybe it was not clear on how to use both methods in the API, here is an example (not yet perfect) that shows, IMO, the intended usage:
```python
# MNIST example

metrics={
    'accuracy': CategoricalAccuracy(),
    'loss': Loss(criterion)
}

evaluator = create_supervised_evaluator(model, metrics=metrics, device=device)
    
pbar = ProgressBar()

def run_train_evaluation(engine):
    pbar.log_message("Run evaluation on training dataset:")
    evaluator.run(train_loader)

def run_val_evaluation(engine):
    pbar.log_message("Run evaluation on validation dataset:")        
    evaluator.run(val_loader)

loss = RunningAverage(output_transform=lambda x: x)
loss.attach(trainer, "running_avg_loss")
# Display progress bar and running average loss value
pbar.attach(trainer, ["running_avg_loss", ])

# Display evaluation loss and accuracy at the end of epoch
pbar.add_logging(evaluator, ['loss', 'accuracy'])
```
and the output:
screen shot 2018-09-29 at 10 29 00


It would be perfect to be able somehow to pass a custom message here:
```python
pbar.add_logging(evaluator, ['loss', 'accuracy'])
```
and override the message `Epoch 1` by something else

miguelvr:
@vfdev-5 you can create your custom logging message if you use `pbar.log_message`, you don't need to use the `add_logging` method. In that case, I wouldn't use it. I'm open to suggestions for the custom message.

> What would we miss out on if we removed the ability to pass metric_names at attach time?

@jasonkriss We would miss the real time metrics logging in the progress bar (in the postfix of the tqdm)

vfdev-5:
@miguelvr seems we still discuss on the API.

> you can create your custom logging message if you use `pbar.log_message`, you don't need to use the `add_logging` method. In that case, I wouldn't use it.

In this case we need to create a special handler:
```python
pbar.add_logging(evaluator, ['loss', 'accuracy'])
```
vs 
```python
@evaluator.on(Events.EPOCH_COMPLETED)
def log_val_matrics(engine):
    msg = "Loss: {} | Accuracy: {}".format(engine.state.metrics["loss"], engine.state.metrics["accuracy"])
    pbar.log_message(msg) 
```
I'm not fan of that.


In this case me too, neither, I do not see the practical usage of `add_logging`. Just writing metrics that we already see with the progressbar as [here](https://github.com/pytorch/ignite/pull/256#issuecomment-425426034) is not of the interest, IMO.

vfdev-5:
@miguelvr I let you decide: either remove it, either make it cover the use-case proposed [here](https://github.com/pytorch/ignite/pull/256#issuecomment-425627388).

miguelvr:
@vfdev-5 @jasonkriss removed `add_logging` method from pbar

right now it should be very straightforward to understand the handler from the user point of view

vfdev-5:
@miguelvr seems OK for me, except the following in `_update`:
```python
        if metric_names is not None:
            metrics = {name: '{:.2e}'.format(engine.state.metrics[name]) for name in metric_names}
```
if `name` not yet in `engine.state.metrics` (case of ordinary metrics, not `RunningAverage`). 


Please, fix the CI problem

miguelvr:
@vfdev-5 are this changes good enough? Or is there any other workaround for the remainder of metrics?

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -0,0 +1,80 @@
...
+    def _update(self, engine, metric_names=None):
+        if self.pbar is None:
+            self._reset(engine)
+
+        self.pbar.set_description('Epoch {}'.format(engine.state.epoch))
+
+        if metric_names is not None:
+            if not engine.state.metrics \
+                    or not all(metric in metric_names for metric in engine.state.metrics):
```
jasonkriss: I think this needs to be reversed. engine.state.metrics will always exist as well. So i think this can be simplified to:

```python
```

jasonkriss:
@miguelvr thank you! I like this more. I think it's more clear and less "magical."

ignite/contrib/handlers/tqdm_logger.py
vfdev-5: Okay, I can push it myself and merge then

miguelvr:
@vfdev-5 for some reason the ProgressBar docs are not showing up in the website

vfdev-5:
@miguelvr yep correct. Here is the problem: 
https://travis-ci.org/pytorch/ignite/jobs/435957238

PR #257, opened 2018-09-04
Fix TBPTT trainer

AntoinePrv:
Corrected the bug and added a couple of tests.
Do you feel it's enough?


vfdev-5:
@AntoinePrv Looks good. Have you tested your new tests whether they catch the error if bug is not fixed ?

tests/ignite/contrib/test_engines.py
```diff
@@ -83,11 +80,59 @@ def _test_create_supervised_tbptt_trainer(device):
...
     # If tbptt is not use (one gradient update), the hidden to hidden weight
     # should stay zero
     assert not model.weight_hh_l0.item() == pytest.approx(0)
 
+    # Cheking forward calls
+    assert forward_mock.call_count == 3
+    for i in range(3):
+        inputs = forward_mock.call_args_list[i][0][1]
+        if i == 0:
+            assert len(inputs) == 1
```
AntoinePrv: @vfdev-5 This is the first call to the model, without `hidden`

tests/ignite/contrib/test_engines.py
```diff
@@ -83,11 +80,59 @@ def _test_create_supervised_tbptt_trainer(device):
...
+        else:
+            assert len(inputs) == 2
```
AntoinePrv: @vfdev-5 Here the number of arguments call to the model is 2 for input and hidden

AntoinePrv:
Yes, test fails without the fix. Also [this line](https://github.com/pytorch/ignite/pull/257/files#diff-0f9d88f685966cb40ff94faaa0e520cdR89) would fail.

Both `_test_create_supervised_tbptt_trainer` and `test_create_supervised_tbptt_trainer_callcounts`  fail without the fix.

Many aspects are cover by both test, `test_create_supervised_tbptt_trainer_callcounts` is based on mocking many objects, and ` _test_create_supervised_tbptt_trainer` uses real PyTorch computations.

vfdev-5:
Okay, thanks @AntoinePrv! 
I let @alykhantejani  review and merge

vfdev-5:
@AntoinePrv could you also put this in a module `engines` and in a separate file, e.g. `tbptt_trainer.py` in case of other contrib engines.

alykhantejani:
@AntoinePrv LGTM although I'm not too familiar with truncated bp through time. I have one nit though, the function `_detach_hidden` is a private function but used everywhere in the tests. 

Is this function intended to be used by end users? If so then we should make it public, if not we could re-write the tests to not use this. 

Anyway - merging this for now as its a bug fix and a nit comment but food for thought. 

Thanks for the PR!

AntoinePrv:
No, it is not meant to be used by the end user, it's tested with `RNN` and `LSTM` because it's two cases where the Module don't return the same type: `RNN` has a `Tensor` as hidden, and `LSTM` a tuple of `Tensor`s).
It's also an important part of the TBPTT.

It's actually only used once, and is only one line, so we could remove it.

PR #261, opened 2018-09-08
[WIP] Add language modelling example

elanmart:
This PR still needs some work, mainly adding docs and testing different options.

It adds a language model training, based on https://github.com/salesforce/awd-lstm-lm .

Let me know if the structure and features look OK, and I'll proceed with adding the missing docs / testing it a bit more.

Oh, and currently this is `python 3.6+` only, should it be adapted to py-2 as well?

### DIFF
--- a/examples/language_modelling/README.md
+++ b/examples/language_modelling/README.md
@@ -1,0 +1,3 @@
+Trains an `awd-lstm` model on a `wiki-text` dataset.
+
+Credit goes mostly to https://github.com/salesforce/awd-lstm-lm
--- a/examples/language_modelling/data.py
+++ b/examples/language_modelling/data.py
@@ -1,0 +1,125 @@
+import hashlib
+import os
+from collections import Counter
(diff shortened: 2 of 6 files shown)


vfdev-5:
@elanmart very interesting and modern example. 

A question, can we more separate the logic between what is done by `salesforce` and where we use `ignite`? Let me explain, can we just ask the user to git clone the repository and reuse common files like `utils.py`, `model.py` etc? If it is the case we could rewrite only the training part. 

What do you think ?

> Oh, and currently this is python 3.6+ only, should it be adapted to py-2 as well?

Yes, I think we should be python2 compatible :)

vfdev-5:
@elanmart maybe we can close this PR ?

elanmart:
Hey @vfdev-5 , yes, I'm really sorry for being so ridiculously unreliable. I'll close the PR, and hopefully one day will get myself together to contribute something complete. 

Cheers, 
Marcin

Issue #260, opened 2018-09-07
Tensorboard Logger


anmolsjoshi:
Working on a Tensorboard Handler for Ignite. 

Currently the code does the following
- Creates a graph of the model on initialization of the event handler
- Creates distributions and histograms for the model at the end of each epoch
- Creates distributions and histograms for the gradients to the parameters of the model at the end of each epoch
- Plots metrics related to training and validation (if available) at the end of each epoch

Features I'm working to add are
- View embedding
- Visualize weights of the model

What other features do you think are necessary for this handler? 
@rave78 @vfdev-5

anmolsjoshi:
I am trying to make the event handler general. Looking through examples, these are the variations I can see arising:
- Logging evaluator metrics (separate engine for evaluating) (MNIST)
- Training engine.state.output, this can be a dictionary, a single value or nothing at all (Fast Neural Style, DCGAN)
- A new attribute to engine.state (see RL example)

Any other variations we can think of?

anmolsjoshi:
Just another idea, we could have the TensorBoard event handler as a Base class that people can add functions that will be called at the end of each epoch. For example:
 
- Stylizing an image at the end of each epoch to see improvement
- Feature maps of a single image through convolution layers
- Generating text, audio, images

Thoughts?

rave78:
@anmolsjoshi sorry for my late reply.  I have not started working on TensorBoard.  However, let me know how I can help you

anmolsjoshi:
@rave78 what are your thoughts on the things listed above? any things you disagree with or think need added?

anmolsjoshi:
@vfdev-5 @rave78 Currently facing issues plotting histograms and writing images at the same time. Will post an issue on tensorboardX for workarounds. Let me know if you guys are interested in looking at the code!

rave78:
@anmolsjoshi I think your list of features is complete.

PR #265, opened 2018-09-17
Tensorboard Logger

anmolsjoshi:
PR for [Tensorboard Logger](https://github.com/pytorch/ignite/issues/260) 

Posted code works with all examples in ignite/examples, I can add those files to this PR as well. Would appreciate any comments and feedback. 

Currently, handler can do the following:

- use an evaulator to log metrics at the end of each epoch
- plot scalars from engine.state.output 
- plot scalars from custom engine.state attributes such as reward
- plot histograms of weights of models
- plots gradients of histograms of weights of models
- handler can be used as a base class to create custom functions like create fake image, create stylized image, etc and called at the end of each epoch (want to improve upon this)

Would like to add
- embedding projections
- some method to add custom functions to correct Event


vfdev-5:
Thanks @anmolsjoshi ! Let me check it locally and review the code. Just a question whether you think it would be interesting to replace mnist with tensorboard x example using this handler directly ?

anmolsjoshi:
@vfdev-5 I updated my post above, I have added an example on using this handler to all the examples in ignite. I can add them to the PR if you would like!

vfdev-5:
@anmolsjoshi which post are you mentioning ? 
IMO, I think if we have a good tensorboard handler we can replace our example with the newer one with the handler :)

anmolsjoshi:
@vfdev-5 sorry, I was referring to this PR's description.

vfdev-5:
> I have added an example on using this handler to all the examples in ignite. I can add them to the PR if you would like!

@anmolsjoshi so yes please add `mnist_with_tensorboard.py` example to this PR too :)

anmolsjoshi:
@vfdev-5 I've updated the code and added 2 examples. I changed the mnist example as discussed and added an example for dcgan. 

Would really appreciate any feedback! Trying to improve my skills :)

ignite/contrib/handlers/tensorboardX.py
```diff
@@ -0,0 +1,185 @@
...
+            self._log_custom_state(engine)
+
+        # Use engine.state.output for example mnist.py example
+        if self.use_output:
+            self._log_engine_output(engine)
+
+    def _on_complete(self, engine):
+        self.writer.close()
+
+    def _difference(self):
```
vfdev-5: @anmolsjoshi could you please detail what does this method ? Name of the method is not clear

anmolsjoshi: @vfdev-5 are you referring to `_difference`? it can be named better, it basically finds user defined functions that need to added to Tensorboard Summary if they were to use this handler as a base class. These functions could be creating images, audio, video, etc.

I showed that use here in [dcgan_with_tensorboardX.py](https://github.com/anmolsjoshi/ignite/blob/tensorboard/examples/gan/dcgan_with_tensorboardX.py#L145)

anmolsjoshi:
@vfdev-5 quick update, my code for visualizing layers works now if you wanted that added.

anmolsjoshi:
@vfdev-5 any more changes needed?

anmolsjoshi:
@vfdev-5 I updated the code to fit the suggested API. It works well with the MNIST, I can extend it to GAN as well. 

I can add the following, just wanted to make sure we were on the same page for structure:
* add images of weights
* add embeddings
* add a method to run user created functions, if not that, an example how to use the summary writer for custom functions.

anmolsjoshi:
- Agree on changing the name to TensorboardLogger

- My understanding was that evaluator was an engine. for example: create_supervised_evaluator. The reason I have train_evaluator and validation_evaluator so that once metrics are computed, it plots those metrics from the training set and validation set. Here is a suggestion, we can include parameters evaluator (engine to evaluate model with metrics attached), train_loader and validation_loader. The handler can run evaluator on said datasets and plot results. Thoughts?

- If we use train_loader/validation_loader as suggested above, then we can remove the parameter write_graph and keep that as a default. The graph can be written using train_loader and model for every run. This way we won't need input_shape.

- state_keys is used mainly for the RL example. engine.state.reward is something that could be plotted using this. 

The mnist example works well! and I've written out tests that I will be committing some time tomorrow. 

I'm going to repurpose the new dcgan example to include tensorboard as well, to show how to use the handler for custom functions.

miguelvr:
@vfdev-5 @anmolsjoshi - What is preventing this to move forward, and how can I help?

vfdev-5:
> and how can I help?

@miguelvr sound great! 

> What is preventing this to move forward,

IMO we are still missing a good API as we've done for `ProgressBar`. 

We need to think how to simplify it and that it remains rather generic. I exposed some problems in above messages.

anmolsjoshi:
@miguelvr @vfdev-5 sorry for the delay. I was on work travel for the last few weeks, so my contribution was delayed. I'll push my latest commit later tonight that should hopefully address the comments from the last review.

anmolsjoshi:
@vfdev-5 @miguelvr changes have been made. please review updated handler, example and tests.

anmolsjoshi:
cc @vfdev-5

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Examples:
+
+        Create a TensorBoard summary that visualizes metrics related to training,
+        validation, and model parameters, by simply attaching the handler to your engine.
+
+        ..code-block:: python
+
+            tbLogger = TensorboardLogger()
+            tbLogger.attach(engine=trainer,
```
vfdev-5: Seems the example is out of date

amitibo:
> @anmolsjoshi take a look at #329 and we need to discuss how to unify the API between these loggers.

Unified API is a good idea, though currently the two PR's have different approaches to the task. #329 is more low level. So it is not clear to me if it is possible to join APIs at the current stage.

On the other hand I just noticed that TensorboardX is supposed to support visdom:
https://github.com/lanpa/tensorboardX/blob/9adaa96ffb8d0fb4ed4a2ae43a0be2ea2a739773/tensorboardX/torchvis.py#L15
So maybe this might be a solution to having a joint API.

anmolsjoshi:
@vfdev-5 are we thinking of unifying the two APIs then? it seems that what can be achieved by visdom is different than what tensorboardX.

anmolsjoshi:
I'll push code later tonight with a more concrete example and better documented code.

anmolsjoshi:
@vfdev-5 @amitibo @miguelvr I pushed some code for possibly a cleaner API for tensorboard logger. Please review. I think we put our heads together we can clean this up further. 

Results are shown below. Here you can see the following:

* training loss from engine.state.output is printed
* results from evaluating on the training set and validation set are shown on the same graph. 

screenshot from 2019-01-07 20-17-43

TheCodez:
@vfdev-5 any update on this?

vfdev-5:
@TheCodez IMO it is rather complex to findout a good API for such logger to have basics and more complex stuff. I tend to use directly tensorboardX writer in events. Maybe a solution could be to cover basics and expose writer for other stuff...

amitibo:
@vfdev-5 @anmolsjoshi, what about the following idea:
A general logger with the following features
1) The logger is attached to some engine and by default logs all its metrics.
2) The logging method will be configured by environment variables, e.g. having an environment variable named "TENSORBOARD_LOGDIR", will result in configuring `tensorboardx.SummaryWriter`. In a similar way, having "VISDOM_SERVER" environment variable, will result in configuring `visdom.Visdom` etc.
3) Logging will happen each time a metric is updated (might require adding a callback mechanism to `Metric`) or at the end of each epoch (doesn't require a new callback mechanism).
4) Grouping of plots can be done according to hierarchical names. For example, having two metrics, one called `loss/train` and the other `loss/validation` will result in a single `tensorboard` or `visdom` window called `loss` and two lines, one for each metric.
5) Logging of training loss can be done by introducing something like an `IterationMetric` i.e. a metric that summarizes the output of the traininer Engine, similar to the way a standard `Metric` summarizes the output of a validator Engine. This will unify the interface to the logger but will also simplify the understanding of training and validation engines (especially for people using `ignite` for the first time) as it will make them more similar to each other.
6) Simple API can expose other capabilities of the logging backend, e.g. `log_embedding` etc.
7) The logger can offer further customization (graph visualization) but its defaults (as described above) should satisfy most use cases of most users.

vfdev-5:
@amitibo I need some precisions on what you proposed :

>  2\.  The logging method will be configured by environment variables, e.g. having an environment variable named "TENSORBOARD_LOGDIR", will result in configuring `tensorboardx.SummaryWriter`. In a similar way, having "VISDOM_SERVER" environment variable, will result in configuring `visdom.Visdom` etc.

You would like to propose a unified logger that uses either tensorboardx or visdom depending on env. variable ?

>     5\. Logging of training loss can be done by introducing something like an `IterationMetric` i.e. a metric that summarizes the output of the traininer Engine, similar to the way a standard `Metric` summarizes the output of a validator Engine. This will unify the interface to the logger but will also simplify the understanding of training and validation engines (especially for people using `ignite` for the first time) as it will make them more similar to each other.

What would be the difference between `IterationMetric` and `RunningAverage` ?

amitibo:
> @amitibo I need some precisions on what you proposed :
> 
> > 2.  The logging method will be configured by environment variables, e.g. having an environment variable named "TENSORBOARD_LOGDIR", will result in configuring `tensorboardx.SummaryWriter`. In a similar way, having "VISDOM_SERVER" environment variable, will result in configuring `visdom.Visdom` etc.
> 
> You would like to propose a unified logger that uses either tensorboardx or visdom depending on env. variable ?
> 

Yes. This will enable seamless logging to multiple destinations: visdom+mlflow or tensorboard+mlflow etc. The purpose is to hide details from the typical user, while allowing flexibility. 

> 
> What would be the difference between `IterationMetric` and `RunningAverage` ?

I forgot about `RunningAverage`. It should serve the purpose.

vfdev-5:
@amitibo I'm not fan of mixing both loggers. Actually, I propose to have a very basic wrappers (2 classes) for TensorboardX or Visdom, that as suggested by @amitibo : 

    1. The logger is attached to some engine and by default logs all its metrics with a certain period.
    2. Exposes internal writer to make possible more complex logging stuff directly done by user
    3. Closes properly internal writer when wrapper is deleted.
    4. Plot model's graph and other features that we can do in a single way. For example, in contrast, we can plot model's weights as histogram, or as scalar of weight norm etc.

@anmolsjoshi @amitibo any other suggestions ?

@amitibo concerning the integration with mlflow, in our case when you setup env var "TENSORBOARD_LOGDIR"  or visdom, do you make an assumption that the storage is file storage ?

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -0,0 +1,305 @@
...
+        writer = self.writer[learning]
+
+        if mode == 'epoch':
+            if use_metrics:
+                self._plot_metrics(engine, writer, mode='epoch')
+
+            if state_keys:
+                self._plot_state_keys(engine, writer, state_keys, mode='epoch')
+
+        if histogram_freq > 0 and self.engine.state.epoch % histogram_freq == 0:
```
Bibonaut: I expect histogram_freq to be 1/histogram_interval. Then the following should be correct from my understanding:
```suggestion
        if histogram_freq > 0 and self.engine.state.epoch % 1./histogram_freq == 0:
```

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -0,0 +1,305 @@
...
+    def _write_graph(self, engine, model, dataloader):
+        """
+        Create computational graph of model.
+
+        Args:
+            model (nn.Module): model to be written
+            dataloader (torch.utils.DataLoader): data loader for training data
+        """
+        writer = self.writer['train']
+        x, y = next(iter(dataloader))
```
Bibonaut: Assuming that seeds are set manually at the beginning of an experiment, this line of code can break reproducibility. Running the same experiment with `write_graph=True` and `write_graph=False` can result in different batch data passed to the engines `process_function` because the dataloader is called once here. I personally do not use the `write_graph` option but if `batch_size` or number of samples passed to the model to write a graph is unimportant, one can use 
```
x, y = dataloader.collate_fn(dataloader.dataset.__getitem__(0))
```
instead?

PR #266, opened 2018-09-17
Add prepare batch as an optional argument to create_supervised*

vfdev-5:
As discussed in slack, let's add `prepare_batch` as argument into `create_supervised*`

cc @jasonkriss 

@AntoinePrv did you have an issue with this missing option ?

### DIFF
--- a/ignite/engine/__init__.py
+++ b/ignite/engine/__init__.py
@@ -10,7 +10,9 @@ def _prepare_batch(batch, device=None, non_blocking=False):
             convert_tensor(y, device=device, non_blocking=non_blocking))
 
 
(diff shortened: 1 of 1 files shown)


AntoinePrv:
I remembered having to do something specific, that I solved first by hacking
```python
import ignite
ignite._prepare_batch = new_func
```
Then moved on to write my own trainer, because I ended needing more than the two arguments in 
```
x, y = _prepare_batch(batch, device=device, non_blocking=non_blocking)
```

So in the end, I wouldn't qualify it as missing for my uses.

jasonkriss:
@vfdev-5 LGTM. Thanks for taking care of that!

vfdev-5:
@alykhantejani merge this please if you are OK too

PR #267, opened 2018-09-19
Add RunningAverage handler to compute running average value of a metr…

### LINKED ISSUE #264: Running average metric
Following the discussion in #256 on running average values in the TQDM progress bar, IMO it would be better to split running average computation and display logic.

Here is an implementation on how to compute running average value from an instance of Metric:
```python


class RunningAverage(Metric):
    """Compute running average of a metric
    """
    
    def __init__(self, metric, alpha=0.98):
        assert isinstance(metric, Metric), "Argument metric should be an instance of Metric"
        assert 0.0 < alpha <= 1.0, "Argument alpha should be a float between 0.0 and 1.0"
        self.metric = metric
        self.alpha = alpha
        super(RunningAverage, self).__init__()
        
    def reset(self):
        self._value = None

    def iteration_completed(self, engine):        
        self.metric.iteration_completed(engine)

    def compute(self):
        if self._value is None:
            self._value = self.metric.compute()
        else:
            self._value = self._value * self.alpha + (1.0 - self.alpha) * self.metric.compute()
        return self._value
    
        # restart average every epoch
        # reset metric at each iteration        
        engine.add_event_handler(Events.ITERATION_COMPLETED, self.metric.started)
        # compute metric
        # apply running average
        engine.add_event_handler(Events.ITERATION_COMPLETED, self.completed, name)
```

Usage of the proposed API is following:
```python
import numpy as np
import torch

from ignite.metrics import Loss, CategoricalAccuracy


n_iters = 100
batch_size = 10
n_classes = 10
data = list(range(n_iters))
y_true_batch_values = iter(np.random.randint(0, n_classes, size=(n_iters, batch_size)))
y_pred_batch_values = iter(np.random.rand(n_iters, batch_size, n_classes))
loss_values = iter(range(n_iters))


def update_fn(engine, batch):
    loss_value = next(loss_values)
    return loss_value, torch.from_numpy(y_pred_batch), torch.from_numpy(y_true_batch)


trainer = Engine(update_fn)

alpha = 0.98
acc_metric = RunningAverage(CategoricalAccuracy(output_transform=lambda x: [x[1], x[2]]), alpha=alpha)
acc_metric.attach(trainer, 'running_avg_accuracy')

loss_metric = RunningAverage(Loss(torch.nn.CrossEntropyLoss(), output_transform=lambda x: [x[1], x[2]]), alpha=alpha)
loss_metric.attach(trainer, 'running_avg_loss')

def log_running_avg_metrics(engine):
    print("running avg accuracy:", engine.state.metrics['running_avg_accuracy'])
    print("running avg loss:", engine.state.metrics['running_avg_loss'])    
    
```

There is however still missing the part of the running average value based on `state.output`.
Maybe we could find a solution together by improving the previous code...

What do you guys think about ?
Thanks 

cc @miguelvr

vfdev-5:
…ic or engine.state.output

Here is an implementation of running average for a metric or engine.state.output that is frequently used in our examples and in PR with TQDM.

Any feedback on the implementation, please
Thanks

cc @miguelvr

PS.: CI fails on python 2, will fix soon


ignite/handlers/running_average.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __init__(self, src, alpha=0.98, output_transform=None):
+        if not (isinstance(src, Metric) or src == 'output'):
+            raise TypeError("Argument src should be a Metric or string 'output'")
+        if not (0.0 < alpha <= 1.0):
+            raise ValueError("Argument alpha should be a float between 0.0 and 1.0")
+
+        if isinstance(src, Metric):
+            if output_transform is not None:
+                raise ValueError("Argument output_transform should be None if src is 'output'")
```
alykhantejani: This error message is confusing as this  is the case where `isinstance(src,Metric)` is `True`

ignite/handlers/running_average.py
```diff
@@ -0,0 +1,66 @@
...
+    Args:
+        src (str or Metric): input source: "output" string or an instance of :class:`ignite.metrics.Metric`
+        alpha (float, optional): running average decay factor, default 0.98
+        output_transform (Callable, optional): a function to use to transform the output if `src` is "output".
+            Otherwise it should be None
+
+    """
+
+    def __init__(self, src, alpha=0.98, output_transform=None):
+        if not (isinstance(src, Metric) or src == 'output'):
```
alykhantejani: Should we make the default `output` (or `None` meaning output)? I have no strong feeling about this, just an idea

ignite/handlers/running_average.py
```diff
@@ -0,0 +1,66 @@
+from ignite.metrics import Metric
+from ignite.engine import Events
+
+
+class RunningAverage(object):
```
alykhantejani: Can we not make this inherit from `Metric`? It feels as though the running average is also a metric (of metrics)?

vfdev-5: yes you are right, as discussed in the issue at first it was implemented as a class inherited from Metric. But then I did not find any advantage of doing that...
See #264

ignite/handlers/running_average.py
vfdev-5: Good idea

jasonkriss: I like this idea

ignite/handlers/running_average.py
jasonkriss: I think maybe we can just drop this entirely? I can see valid use cases where you way want to compute the running average of just some part of the output in which case we should allow `output_transform`.

ignite/handlers/running_average.py
jasonkriss: Agreed. Conceptually, I think it makes sense for this to be with other "metrics."

ignite/handlers/running_average.py
vfdev-5: In case of `src` is `Metric` is it redundant to have `output_transform` as metric itself should take care of output transform. In case of src being engine.state.output it is better to be able to transform the output. This check somehow controls the redundant case in the way if user setups both `output_transform`s the code stops. But I'm OK to remove it

jasonkriss: I see you point about the redundancy in the `Metric` case. In that case, we can just update the error message, right? It currently reads "Argument output_transform should be None if src is 'output'" but should be "Argument output_transform should be None if src is a Metric"?

vfdev-5: Oh, yes, my bad. Error message is just wrong :)

vfdev-5:
Moved RunningAverage to metrics. Tests should pass also

ignite/metrics/running_average.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def __init__(self, src=None, alpha=0.98, output_transform=None):
+        if not (isinstance(src, Metric) or src is None):
+            raise TypeError("Argument src should be a Metric or None")
+        if not (0.0 < alpha <= 1.0):
+            raise ValueError("Argument alpha should be a float between 0.0 and 1.0")
+
+        if isinstance(src, Metric):
+            if output_transform is not None:
+                raise ValueError("Argument output_transform should be None if src is a Metric")
```
vfdev-5: Updated message

ignite/metrics/running_average.py
```diff
@@ -0,0 +1,72 @@
...
+            self.src = src
+            self._get_src_value = self._get_metric_value
+            self.iteration_completed = self._metric_iteration_completed
+        else:
+            if output_transform is None:
+                raise ValueError("Argument output_transform should not be None if src corresponds "
```
vfdev-5: Added similar check if src is "output"

miguelvr:
@vfdev-5 should I adapt the TQDM Progress Bar accordingly?

PR #269, opened 2018-09-21
Create PULL_REQUEST_TEMPLATE.md

vfdev-5:
I would like to propose a PR template with a check list to be verified when a PR is to be merged.
PR template will help us to avoid merging PR with only code and less docs updated (for example, my last PR on RunningAverage, etc)

So when we create a PR we will see something similar

image

[More info](https://help.github.com/articles/about-issue-and-pull-request-templates/#pull-request-templates)

What do you think ?

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,9 @@
+Fixes # 
+
+Description:
+
+
+Check list:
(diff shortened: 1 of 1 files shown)


alykhantejani:
sounds like a good idea to me

vfdev-5:
@jasonkriss if you are OK, please merge :)

Issue #271, opened 2018-09-21
Early stopping handler should fire on Events.EPOCH_COMPLETED


pbamotra:
Hi guys,

I was going through the handlers section and I feel that early stopping should be called on Events.EPOCH_COMPLETED to check if validation loss / accuracy hasn't improved unlike documentation where it says to fire on Events.COMPLETED?

Am I right?

pbamotra:
nvm, it's already asked in closed issues.

vfdev-5:
Thanks for the feedback @pbamotra. Yes it is duplicate of #258.

However we need to update the docs

Issue #258, opened 2018-09-06
Wrong event -- EarlyStopping Example


rave78:
This is the example in the Ignite documentation about EarlyStopping
```
from ignite.engine import Engine, Events
from ignite.handlers import EarlyStopping

def score_function(engine):
    val_loss = engine.state.metrics['nll']
    return -val_loss

handler = EarlyStopping(patience=10, score_function=score_function, trainer=trainer)
evaluator.add_event_handler(Events.COMPLETED, handler)
```
I was wondering if `Events.COMPLETED` is wrong. I think the evaluator should be fired either `Events.EPOCH_COMPLETED` or `Events.ITERATION_COMPLETED`

vfdev-5:
@rave78 this is a correct event in the docentation. Actually this handler is attached to an evaluator that runs a single epoch on a validation dataset. Once it is done, run completed we get results and check for early stopping.

rave78:
I see.  thank you for your reply

PR #272, opened 2018-09-21
Update EarlyStopping doc string and contrib dependency

vfdev-5:
- Same question was asked multiple times #258 #271 
- Fix problem when build the docs for contrib metrics

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ jobs:
         - pip install -r docs/requirements.txt
         - pip install http://download.pytorch.org/whl/cpu/torch-0.4.1-cp35-cp35m-linux_x86_64.whl
         # Add contrib dependencies (otherwise doc is not built)
-        - pip install scikit-learn
+        - pip install scikit-learn scipy
         # `pip install .` vs `python setup.py install` : 1st works better to produce _module/ignite with source links
         - pip install .
       script:
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -25,6 +25,7 @@ class EarlyStopping(object):
             return -val_loss
 
         handler = EarlyStopping(patience=10, score_function=score_function, trainer=trainer)
+        # Note: the handler is attached to an *Evaluator* (runs one epoch on validation dataset)
         evaluator.add_event_handler(Events.COMPLETED, handler)
 
     """


jasonkriss:
@vfdev-5 maybe we should add a comment in the codeblock that says something like
```python
# Note: the handler is attached to an *Evaluator*
```
What do you think?

vfdev-5:
@jasonkriss so the whole block is like this ?
```
    .. code-block:: python

        from ignite.handlers import EarlyStopping

        def score_function(engine):
            val_loss = engine.state.metrics['nll']
            return -val_loss

        handler = EarlyStopping(patience=10, score_function=score_function, trainer=trainer)
        # Note: the handler is attached to an *Evaluator* (runs one epoch)
```

jasonkriss:
@vfdev-5 👍exactly. What do you think?

vfdev-5:
Yes, I like it. Let me commit the change

pbamotra:
Also, I think trainer=trainer is causing the Early Stopping handler to track the training accuracy when I really want the training to stop when the improvements in validation accuracy vanishes beyond patience.

```

                                                     'nll': Loss(criterion)},
                                            device=device)

def score_function(engine):
        return engine.state.metrics['accuracy']

earlystop_handler = EarlyStopping(patience=5, score_function=score_function, trainer=trainer)
evaluator.add_event_handler(Events.COMPLETED, earlystop_handler)
```

**Am I doing anything wrong here?**

Also, I modified the EarlyStopping handler to print messages

```
    def __init__(self, patience, score_function, trainer):
        assert callable(score_function), "Argument score_function should be a function"
        assert patience > 0, "Argument patience should be positive"
        assert isinstance(trainer, Engine), "Argument trainer should be an instance of Engine"
        self.score_function = score_function
        self.patience = patience
        self.trainer = trainer
        self.counter = 0
        self.best_score = None
        self._logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
        self._logger.addHandler(logging.NullHandler())

    def __call__(self, engine):

        if self.best_score is None:
            self.best_score = score
        elif score < self.best_score:
            self.counter += 1
            tqdm.write("EarlyStopping: %i / %i\n" % (self.counter, self.patience))
            if self.counter >= self.patience:
                tqdm.write("EarlyStopping: Stop training\n")
                self.trainer.terminate()
        else:
            tqdm.write("Resetting counter. New best score: {}\t Prev best: {}\n".format(score, self.best_score))
            self.best_score = score
            self.counter = 0
```

And as I see it, the scores it track are the training data scores.

vfdev-5:
@pbamotra could you please post a complete snippet on gist ?

Following your code:
```python
def score_function(engine):
        return engine.state.metrics['accuracy']


class EarlyStopping:
     # ....
     def __call__(self, engine):


evaluator.add_event_handler(Events.COMPLETED, earlystop_handler)
```

So as `earlystop_handler` is attached to `evaluator` so when `self.score_function(engine)` is called, the `engine` should be `evaluator`. Could you please check it in your code?  Trainer passed in the `EarlyStopping` is just needed to stop the training and nothing else.
HTH

PR #275, opened 2018-09-25
Update Precision, Recall, add Accuracy (Binary and Categorical combined)

### LINKED ISSUE #262: Problems with binary classifier & metrics
Hi,

I am new to PyTorch and am trying to get started with Ignite. I want to train a simple binary classifier:

```
class NeuralNet(nn.Module):
  def __init__(self, input_size, hidden_size, num_classes):
    super(NeuralNet, self).__init__()
    self.fc1 = nn.Linear(input_size, hidden_size)
    self.relu = nn.ReLU()
    self.fc2 = nn.Linear(hidden_size, num_classes)
    self.out_act = nn.Sigmoid()

  def forward(self, x):
    out = self.fc1(x)
    out = self.relu(out)
    out = self.fc2(out)
    out = self.out_act(out)
    return out
```

First I had (plus some events):

```
trainer = create_supervised_trainer(model, optimizer, nn.BCELoss(), device=device)

evaluator = create_supervised_evaluator(
    model, metrics={'precision': Precision(),
                    'recall': Recall(),
                    'accuracy': BinaryAccuracy()}, device=device)
```

This failed with:

```
~/.local/share/virtualenvs/bot2-Z_RSwUyv/lib/python3.6/site-packages/ignite/metrics/precision.py in update(self, output)
     29         num_classes = y_pred.size(1)
     30         indices = torch.max(y_pred, 1)[1]
---> 31         correct = torch.eq(indices, y)
     32         pred_onehot = to_onehot(indices, num_classes)
     33         all_positives = pred_onehot.sum(dim=0)

RuntimeError: Expected object of type torch.LongTensor but found type torch.FloatTensor for argument #2 'other'
```

Then I found the `output_transform` parameter and changed it to:

```
def output_transform(output):
  y_pred, y = output
  return y_pred.gt(0.5).long(), y.long()

evaluator = create_supervised_evaluator(
    model, metrics={'precision': Precision(output_transform=output_transform),
                    'recall': Recall(output_transform=output_transform),
                    'accuracy': BinaryAccuracy(output_transform=output_transform)}, device=device)
```

Even if this worked, I do not think it was a good solution. In my opinion, it should be possible to do the transformation in just one place instead of having to pass it to each metric. Is this somehow possible? Anyway, this fails with:

```
~/.local/share/virtualenvs/bot2-Z_RSwUyv/lib/python3.6/site-packages/ignite/metrics/precision.py in update(self, output)
     35             true_positives = torch.zeros_like(all_positives)
     36         else:
---> 37             correct_onehot = to_onehot(indices[correct], num_classes)
     38             true_positives = correct_onehot.sum(dim=0)
     39         if self._all_positives is None:

IndexError: too many indices for tensor of dimension 1
```

Am I on the right track?

anmolsjoshi:
Fixes #262 

Description:
This PR updates Precision and Recall, and adds Accuracy to handle binary and categorical cases for different types of input. 

Check list:
* [x] New tests are added. 
* [x] Updated doc string RST format
* [x] Edited metrics.rst to add information about Accuracy.


ignite/metrics/accuracy.py
```diff
@@ -0,0 +1,45 @@
...
+from ignite.metrics.metric import Metric
+from ignite.exceptions import NotComputableError
+
+
+class Accuracy(Metric):
+    """
+    Calculates the accuracy.
+
+    - `update` must receive output of the form `(y_pred, y)`.
+    - `y_pred` must be in the following shape (batch_size, num_categories, ...)
```
vfdev-5: nit: may be we need to update this `(batch_size, num_categories, ...)` too

vfdev-5:
@anmolsjoshi thanks a lot for taking this ! 
Just a small precision on how to handle old `BinaryAccuracy` and `CategoricalAccuracy`:
> merge the classes into Accuracy and still keep `BinaryAccuracy` and `CategoricalAccuracy` and have their constructors just create the appropriate instance of `Accuracy` and throw a warning that these will be removed in 0.1.2

ignite/metrics/accuracy.py
```diff
@@ -0,0 +1,45 @@
...
+    - `y` must be in the following shape (batch_size, ...)
+    """
+    def reset(self):
+        self._num_correct = 0
+        self._num_examples = 0
+
+    def update(self, output):
+        y_pred, y = output
```
vfdev-5: Maybe we need also to compare shapes of `y_pred` and `y` and raise error if they are different ?

anmolsjoshi: I was planning on adding that!

vfdev-5:
@anmolsjoshi no problem, they could be something like [this](https://github.com/pytorch/vision/blob/f91a1823109393b6d521ebde6aa5f930f62f383c/torchvision/transforms/transforms.py#L194)

anmolsjoshi:
Thanks for the example. Will get this worked tomorrow.

anmolsjoshi:
@vfdev-5 could you clarify [this test](https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/test_binary_accuracy.py#L30) ?

y has 4 elements with shape of (1, 2, 2), while y_pred has 8 elements with shape of (2, 2, 2). correct is equal to `tensor([0, 1, 0, 1, 0, 1, 0, 1], dtype=torch.uint8)`

Given the docs, y_pred should be of shape (batch_size, ...) or (batch_size, num_classes, ...), which is not the case here. Given that this is a test for binary accuracy, shouldn't y_pred have a shape (1, 2, 2) or (1, 2, 2, 2). If it is the second case, this test should be handled in CategoricalAccuracy. 

Or am I understanding this incorrectly?

vfdev-5:
@anmolsjoshi this is the case to compute accuracy between images (2x2) (segmentation task).

ignite/metrics/accuracy.py
anmolsjoshi: resolved. code is verbose, let me know if you need it trimmed down.

ignite/metrics/accuracy.py
anmolsjoshi: resolved. let me know if that works.

anmolsjoshi:
@vfdev-5 I have updated all the tests appropriately and incorporated your comments. Please let me know your thoughts!

vfdev-5:
> any suggestions on the codecov failures?

@anmolsjoshi take a look [here](https://codecov.io/gh/pytorch/ignite/compare/a437a814fa72c7b861fce6004a0c2008c3a02681...bb7bef970fd46a4d71469fcb51173f602bca2d2f/diff). In red the lines that are not covered by tests (code is not executed).

vfdev-5:
@anmolsjoshi I was thinking on your implementation of unified binary and categorical accuracy and maybe we can unify it more (I'm not a fan of huge if cases like `if y.ndimension() > 1:` and `if is_categorical:`). Let me explain: 
We would like to cover the following cases:

Binary classification/regression:
- `y_pred` can be `(batch_size, )`
- `y_true` can be `(batch_size, )`

Binary segmentation on ND array:
- `y_pred` can be `(batch_size, H, W, ...)`
- `y_true` can be `(batch_size, H, W, ...)`

Categorical classification/regression:
- `y_pred` can be `(batch_size, C)`
- `y_true` can be `(batch_size, )`

Categorical segmentation on ND array:
- `y_pred` can be `(batch_size, C, H, W, ...)`
- `y_true` can be `(batch_size, H, W, ...)`

As you may already noticed that the only thing that change in current implementations is `indices`:
Here is the update method of binary accuracy (indices variable is made explicit):
```python
    def update(self, output):
        y_pred, y = output
        indices = torch.round(y_pred).type(y.type())
        correct = torch.eq(indices, y).view(-1)
        self._num_correct += torch.sum(correct).item()
```
and here is the update method of categorical accuracy:
```python
    def update(self, output):
        y_pred, y = output
        indices = torch.max(y_pred, 1)[1]
        correct = torch.eq(indices, y).view(-1)
        self._num_correct += torch.sum(correct).item()
```

So, there are two questions for unified update function:
- how to validate the input data ?
- how to understand in which case we are: categorical or binary ?
- how to compute `indices` in an unified way ?

1) how to validate the input data ?

Following above supported case, we can check that 
```python
assert y.ndimension() >= 1 and y_pred.ndimension() >= 1
assert y.ndimension() == y_pred.ndimension() or y.ndimension() + 1 == y_pred.ndimension()
y_shape = y.shape
y_pred_shape = y_pred.shape
if y.ndimension() + 1 == y_pred.ndimension():
    y_pred_shape = (y_pred_shape[0], ) + y_pred_shape[2:]
assert y_shape == y_pred_shape
```

2) how to understand in which case we are: categorical or binary

I think this is just:
- if `y.ndimension() == y_pred.ndimension()` - binary
- if `y.ndimension() + 1 == y_pred.ndimension()` - categorical

3) how to compute `indices` in an unified way ?

Here I propose to map the binary case into 2 classes categorical:
```python
y_pred = y_pred.unsqueeze(dim=1)
```
Sure that in terms of performances we need to benchmark this.

And finally, in the ideal way we should also store somehow what type of data (categorical or binary) was used in the previous `update` and raise an error if user mixes the types.

What do you think about ?

jasonkriss:
@vfdev-5 I like that a lot.

rpatrik96:
@anmolsjoshi  @vfdev-5 
Hi, are you concerning also tasks with multiple target labels? I currently have a problem where it could come handy and I am eager to help.

vfdev-5:
@rpatrik96, thanks for the proposition! What kind of metrics do you think would be useful with multi-label targets?

rpatrik96:
@vfdev-5 
In my opinion, it would be great if the `Precision, Recall` could be used for these tasks too `BinaryAccuracy` is OK for them. I wonder how to generalize the current metrics. 
The easiest would be to specify a flag for multitarget, but maybe it van be done automagically, I mean if we have multiple labels, the target is a one-hot vector, we should compare only for that. What do you think?

vfdev-5:
@rpatrik96 how would you threshold `y_pred` to choose which tag are on and which off ?

rpatrik96:
@vfdev-5 I think the following way would be logical: if `y_` is a one-hot vector with `N` 1s in it, then I would sort the `y_pred` vector and use the top `N` entries like in the case of `TopKCategoricalAccuracy`, but this time the mapping is not many-to-one, but many-to-many.

vfdev-5:
@rpatrik96 maybe we could provide a custom function to override `torch.round` when transforming probas `y_pred` to predicted tags and then continue as in [scikit-learn](http://scikit-learn.org/stable/modules/model_evaluation.html#accuracy-score) ?

rpatrik96:
@vfdev-5
I like your proposal, just one question : why would we need to override `torch.round`? That part is not crystal clear for me.

rpatrik96:
@vfdev-5 Oh, fine, I thought there is something more complicated for it :). So how should we proceed?

anmolsjoshi:
@vfdev-5 I updated accuracy, precision and recall with the suggested shape check.

accuracy also handles mutlilabel cases. 

There is code in Precision and Recall that transforms y or y_pred if there are three dimensions ( N * num_classes * max_len), this the case of Named Entity Recognition, I can take it a step further for Segmentation, although I don't think precision and recall are used for this task. 

I also updated how precision and recall are calculated, the basic idea is to convert y and y_pred to have dimensions of [-1, C] and take an element wise product. 

I also added a threshold to recall and precision - this can be removed and these can be unified like accuracy.

ignite/metrics/accuracy.py
```diff
@@ -0,0 +1,54 @@
...
+
+        assert y.ndimension() >= 1 and y_pred.ndimension() >= 1
```
vfdev-5: @anmolsjoshi please use the pattern `if ... then raise"

anmolsjoshi: @vfdev-5  Done! Do you want this done in tests as well?

vfdev-5:
@anmolsjoshi thanks for covering other cases and sorry for the delay ! Let me review the code and get back to you soon.

@rpatrik96 as multilabels are now tried to being covered, could you please also test it in your cases and comment out. Thanks !

ignite/metrics/accuracy.py
```diff
@@ -0,0 +1,57 @@
...
+        y_shape = y.shape
+        y_pred_shape = y_pred.shape
+
+        if y.ndimension() + 1 == y_pred.ndimension():
+            y_pred_shape = (y_pred_shape[0], ) + y_pred_shape[2:]
+
+        if not (y_shape == y_pred_shape):
+            raise ValueError("y and y_pred must have compatible shapes.")
+
+        if y_pred.ndimension() == y.ndimension():
```
vfdev-5: Please insert a comment here, something like that
```
# Binary case: map into 2 classes categorical
```

tests/ignite/metrics/test_accuracy.py
```diff
@@ -0,0 +1,161 @@
...
+    y = torch.Tensor([[1., 0., 1., 0.],
+                      [0., 1., 0., 0.],
+                      [0., 0., 1., 0.],
+                      [1., 0., 0., 1.]]).type(torch.LongTensor)
+
+    acc.update((y_pred, y))
+    assert acc.compute() == 0.875
+
+
+def test_ner_example():
```
vfdev-5: Could you please explain what this test is testing ?

vfdev-5:
@anmolsjoshi I'm not really sure about multilabelling implementation. Should not this case be 
```
    y_pred = torch.eye(4)
    y = torch.Tensor([[1., 0., 1., 0.],
                      [0., 1., 0., 0.],
                      [0., 0., 1., 0.],
                      [1., 0., 0., 1.]]).type(torch.LongTensor)
```
equal to 
```
0.5
```
as only 2 predictions are exact ?

anmolsjoshi:
@vfdev-5 thanks for the review! I do agree that the result of the test should be 0.5 and not 0.875. I checked using sklearn just now. I followed guidance of a stackoverflow post that I suspect is wrong now... 

We may have to discuss this further - multilabel accuracy and binary accuracy are cases where y and y_pred have the same shape. in multilabel accuracy, we can't do torch.cat([1.0 - y_pred, y_pred], dim=1) because y and y_pred are already of shape (batch_size, n_classes, ...). 

Using torch.eq may not work, as it checks for equality element wise. I suggest we include a parameter if the user wants to calculate multilabel accuracy and use torch.round on y_pred and torch.equal to compare y and y_pred, when both are reshaped to [-1, C]. 

Or we can create another metric called MultiLabelAccuracy, just as we have TopKCategoricalAccuracy. That might be the cleanest way to do this as we'll always know that y and y_pred have the shape (batch_size, num_classes, ...). 

Thoughts?

tests/ignite/metrics/test_accuracy.py
anmolsjoshi: @vfdev-5 discussed below! this test is wrong.

ignite/metrics/accuracy.py
anmolsjoshi: Will do! This can't be done for multilabel. Discussed further below!

anmolsjoshi:
@rpatrik96 @vfdev-5 Thoughts on the above?

rpatrik96:
@anmolsjoshi 
Personally I would prefer to have only one class for accuracy calculations. E.g. we could use, that when multiple classes are assigned to a sample, the target is a one-hot vector, i.e. the last dimension of the shape is 1. 
Furthermore, if we sum the tensor, we get the number of classes needed, thus we could select the best `n` predictions.

ignite/metrics/accuracy.py
vfdev-5: Please add the comment
```suggestion
        # Binary case: map into 2 classes categorical
        if y_pred.ndimension() == y.ndimension():
```

ignite/metrics/precision.py
```diff
@@ -26,16 +27,48 @@ def reset(self):
...
+        y_shape = y.shape
+        y_pred_shape = y_pred.shape
+
+        if y.ndimension() + 1 == y_pred.ndimension():
+            y_pred_shape = (y_pred_shape[0],) + y_pred_shape[2:]
+
+        if not (y_shape == y_pred_shape):
+            raise ValueError("y and y_pred must have compatible shapes.")
+
+        if y_pred.ndimension() == y.ndimension() + 1:
```
vfdev-5: Can't we use the same trick as in Accuracy here ?
I think it would be rather weird to have `threshold_function` for Precision and Recall and not for Accuracy...
IMO, we do not need `threshold_function` for these metrics. Correct me if I'm wrong.

vfdev-5:
@anmolsjoshi good work! We need to advance on it, however, finally, I would prefer to split this PR into 2: leave multi-label part for another PR and firstly merge `BinaryAccuracy` and `CategoricalAccuracy`. Multi-label case should be also introduced in Precision and Recall. Later once we merged 2 accuracies we could add multi-label case with an option param of the initializer... What do you think about all of this ?
I left other comments in the code too.

tests/ignite/metrics/test_accuracy.py
vfdev-5: Finally, do we keep this test ?

ignite/metrics/accuracy.py
anmolsjoshi: Incorporated!

ignite/metrics/precision.py
anmolsjoshi: Incorporated! Changed to similar to accuracy

tests/ignite/metrics/test_accuracy.py
anmolsjoshi: NER example we should keep because it is another type of problem users might encounter. I have removed any multi label tests.

vfdev-5:
@anmolsjoshi it seems that you rebased your branch as there are other commits now appear in this PR. Could you please revert ?

anmolsjoshi:
@vfdev-5 to clarify, I'm going to follow instructions of this [link](https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-a-pull-request) to remove the commits that were already on this PR. Is that the correct way of doing what you want?

vfdev-5:
@anmolsjoshi I'm not sure about this link.
you can do something like
```
git reset --hard HEAD~2
git merge origin master
... save marge commit ..
git push <fork> updatemetrics -f
```

tests/ignite/metrics/test_accuracy.py
vfdev-5: Actually what means NER ?

anmolsjoshi: Named Entity Recognition.

anmolsjoshi:
@vfdev-5 commits are structured correctly now. sorry for the confusion, there were some technical difficulties on my end.

vfdev-5:
@anmolsjoshi look good for me, but I would like to add more various tests, current tests looks a bit very simple. This is to make sure that we are computing things perfectly correct and not because we had a too naive tests. 
For example, take a random predictions, compute locally the result with sklearn, set it in hard in the test and check our computation against the result.

anmolsjoshi:
@vfdev-5 thanks for the review! I'll write a few more tests as you have suggested.

anmolsjoshi:
@vfdev-5 updated the tests to prevent that sklearn warning, updated mnist and test_pbar also. 

Getting [depreciation warnings](https://travis-ci.org/pytorch/ignite/jobs/448384999#L1198), hope that its ok since it is the point of this PR.

vfdev-5:
@anmolsjoshi thanks ! 

> Getting depreciation warnings, hope that its ok since it is the point of this PR.

Yes, this is OK.

We are almost done, concerning sklearn tests, I thought about to add them to all tests : test_binary_compute, test_binary_compute_batch_images, test_categorical_compute, test_categorical_compute_batch_images. Let me add them by myself.

anmolsjoshi:
@vfdev-5 feel free to use this latest commit, I can remove it. but it replaces all the accuracy tests to compare with sklearn and using random values for y_pred.

anmolsjoshi:
@vfdev-5 I have changed most tests of Precision and Recall to add a comparison with sklearn, I kept the naive examples because they are good sanity checks. let me know if any changes are needed. 

I think we should add some information in the doc string about Precision and Recall not being batch wise metrics. See discussion [here](https://github.com/keras-team/keras/issues/5794).

Basically, it shouldn't be used for the Progress Bar or as a Running Average and should be computed at the end of epoch. Happy to discuss!

vfdev-5:
@anmolsjoshi thanks a lot ! LGTM

> I think we should add some information in the doc string about Precision and Recall not being batch wise metrics.

I think this better to be added to `RunningAverage` and even there what happens is that metric is computed batchwise and then smoothed.


@jasonkriss or @alykhantejani please review and merge if OK

ignite/metrics/binary_accuracy.py
```diff
@@ -1,30 +1,13 @@
...
-        correct = torch.eq(torch.round(y_pred).type(y.type()), y).view(-1)
-        self._num_correct += torch.sum(correct).item()
-        self._num_examples += correct.shape[0]
-
-    def compute(self):
-        if self._num_examples == 0:
-            raise NotComputableError('BinaryAccuracy must have at least one example before it can be computed')
-        return self._num_correct / self._num_examples
+    def __init__(self, *args, **kwargs):
+        warnings.warn("The use of ignite.metrics.BinaryAccuracy is deprecated, it will be removed in 0.1.2")
```
jasonkriss: Should we maybe mention in the warning it self that this is being deprecated in favor of Accuracy? Same with the CategoricalAccuracy warning

anmolsjoshi: thanks for the review. I'll update this.

anmolsjoshi:
@vfdev-5 @jasonkriss is there a way to ignore the W504 flake8 error? If I kept the warning as one line, it would be too long and throw another error lol.

vfdev-5:
@anmolsjoshi I think it is OK like you did with a line break inside a string.

jasonkriss:
Yea looks good to me. Thanks @anmolsjoshi!

alykhantejani:
This PR looks great, thanks @anmolsjoshi! Thanks Victor for the thorough review

PR #276, opened 2018-09-25
fix deprecation warning Python 3.7

AntoinePrv:
```
.../ignite/ignite/_utils.py:37: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  elif isinstance(input_, collections.Sequence):

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

### DIFF
--- a/ignite/_utils.py
+++ b/ignite/_utils.py
@@ -1,7 +1,14 @@
-import collections
+import sys
 
 import torch
 from torch._six import string_classes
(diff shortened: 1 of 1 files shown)


vfdev-5:
@AntoinePrv thanks!

PR #282, opened 2018-10-02
Fix contrib.handlers documentation

miguelvr:
added tqdm to docs dependencies to fix the problem reported [here](https://github.com/pytorch/ignite/pull/256#issuecomment-426192327)


vfdev-5:
@miguelvr thanks !

PR #286, opened 2018-10-11
tqdm logger: allow no metric_names

### LINKED ISSUE #285: tqm_logger: metric_names is currently not optional
Hi,

https://github.com/pytorch/ignite/blob/master/ignite/contrib/handlers/tqdm_logger.py#L75
This line should be modified to make `metric_names` optional. Here is a suggestion:
```
if metric_names is not None and not isinstance(metric_names, list):
    raise TypeError("metric_names should be a list, got {} instead".format(type(metric_names)))
```

Thanks

miguelvr:
Fixes #285

### DIFF
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -72,7 +72,7 @@ class ProgressBar:
             engine (Engine): engine object
             metric_names (list): (Optional) list of the metrics names to log as the bar progresses
         """
-        if not isinstance(metric_names, list):
+        if metric_names is not None and not isinstance(metric_names, list):
             raise TypeError("metric_names should be a list, got {} instead".format(type(metric_names)))
 
         engine.add_event_handler(Events.EPOCH_COMPLETED, self._close)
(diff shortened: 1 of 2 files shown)


vfdev-5:
@miguelvr can you also add a test for the argument ?

miguelvr:
Sure thing

miguelvr:
@vfdev-5 I'm trying to do something similar to what I've done with the other unit tests.

However, I'm getting weird behaviours.

The snippet for the code is like this:
```python
def test_pbar_no_metric_names(capsys):

    n_epochs = 2
    loader = [1, 2]
    engine = Engine(update_fn)

    pbar = ProgressBar()
    pbar.attach(engine)

    engine.run(loader, max_epochs=n_epochs)

    captured = capsys.readouterr()
    stderr = captured.err.split('\r')
    err = list(map(lambda x: x.strip(), stderr))
    err = list(filter(None, err))
    actual = err[-1]
    expected = u'Epoch 2: [1/2]  50%|█████      [00:00<00:00]'
    assert actual == expected
```

The comparison with the stderr string is behaving in a weird way and I'm obtaining a bunch of hexadecimal characters. However, if I debug and do everything manually the assertion checks out...

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -47,6 +47,7 @@ def _update(self, engine, metric_names=None):
 
         if metric_names is not None:
             if not all(metric in engine.state.metrics for metric in metric_names):
+                self._close(engine)
```
miguelvr: this indeed seems to solve the problem, thanks!

miguelvr:
LGTM!

PR #291, opened 2018-10-17
Import metrics and exceptions

zasdfgbnm:


vfdev-5:
@zasdfgbnm thanks for the PR, could you explain more what we gain with this modification ?

zasdfgbnm:
@vfdev-5 This allows doing something like
```python
import ignite
print(ignite.metrics.Metric)
```
Currently this is not available, we have to do
```python
import ignite.metrics
print(ignite.metrics.Metric)
```

vfdev-5:
@alykhantejani @jasonkriss I let you decide whether this modification is pertinent. 
If yes maybe `contrib` should be added too ?

alykhantejani:
LGTM, thanks for the PR @zasdfgbnm.

Can you make the same changes to ignite.contrib?

zasdfgbnm:
@alykhantejani Done

PR #294, opened 2018-10-20
Batch size loss metric

### LINKED ISSUE #293: [feature request]Changing how the loss metric get the shape information
Recently I was working with multi-output models and I was using a pattern that I believe to
be common. The model returns a tuple, for instance `y_pred = (ypred_1, ypred_2)` with the correspondent y of the form `y = (y_1, y_2)` with the `loss_fn` accepting as arguments `ypred` and  `y` (`loss_fn(ypred, y)`).

However I have run into problems  when using the Loss metric due to the batch size used on the update function of this metric calling the attribute shape of the, presumed, torch.Tensor directly.

I have as suggestion to change the Loss metric class to accept an extra function, for instance `batch_size = lambda x: x.shape[0]` to recover the current behavior and to permit to access the shape attribute from a tuple/list.

labor00:
Fixes #293 

Description:
The Loss metric now takes a optional argument batch_size that is used to access
the  batch size from the target variable y.

Check list:
* [ ] New tests are added (if a new feature is modified) (uses the same test)
* [ ] New doc strings: text and/or example code are in RST format (is necessary to add an example?)

### DIFF
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -20,6 +20,8 @@ class Loss(Metric):
             The output is is expected to be a tuple (prediction, target) or
             (prediction, target, kwargs) where kwargs is a dictionary of extra
             keywords arguments.
+        batch_size (callable): a callable taking a target tensor that returns the
+            first dimension size (usually the batch size).
 
     """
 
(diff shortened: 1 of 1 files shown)


vfdev-5:
LGTM!

PR #295, opened 2018-10-21
RunningAverage : allow alpha = 0 (+ ProgressBar discussion)

milesial:
**Description:**

Hi, I was trying to create an `Identity` metric that was computed at each iteration, in order to use `contrib.handlers.ProgressBar` to display a training loss. I decided to simply use a running average with alpha=0 (ignore all past values, and take 100% of the new value).

```python
trainer = engine.create_supervised_trainer(model, optimizer, loss)

metric = metrics.RunningAverage(output_transform=lambda x: x, alpha=0.0)
metric.attach(trainer, 'loss')
```

I ran into a `ValueError` because the value 0 for alpha is not allowed, I see no reason why it couldn't be.
=> I modified the check to be inclusive for zero.

No new test or documentation is needed.

**Check list:**

---

PS : The fact that `ProgressBar` only works with `RunningAverage` _(or generally all metrics that rewires `self.computed` on `Events.ITERATION_COMPLETED` instead of `Events.EPOCH_COMPLETED`)_ is not documented and that took me some time to figure out. (see https://github.com/pytorch/ignite/pull/256 for issues on non-`RunningAverage` metrics).

### DIFF
--- a/ignite/metrics/running_average.py
+++ b/ignite/metrics/running_average.py
@@ -33,7 +33,7 @@ class RunningAverage(Metric):
     def __init__(self, src=None, alpha=0.98, output_transform=None):
         if not (isinstance(src, Metric) or src is None):
             raise TypeError("Argument src should be a Metric or None")
-        if not (0.0 < alpha <= 1.0):
+        if not (0.0 <= alpha <= 1.0):
             raise ValueError("Argument alpha should be a float between 0.0 and 1.0")
 
         if isinstance(src, Metric):


vfdev-5:
@milesial first of all thanks for the feedback! I agree with you that the doc of `ProgressBar` may appear unclear about metrics. And probably we need to improve it.
Concerning the PR I think `RunningAverage` functionality does not make sense with `alpha=0`. And as you precised that your problem is to display a value without running average in the `ProgressBar`, a PR should be related to `ProgressBar`, IMO.

In your case I think the following code can make what you would like. This may look like a hack. So you can directly insert your value in the metric dictionary which is read by `ProgressBar`
```python
     # usual stuff before return loss value
     engine.metrics['loss'] = loss_value
     return loss_value

pbar = ProgressBar()
pbar.attach(trainer, ['loss'])

engine.run(...)
```

Anyway, please create an issue to improve `ProgressBar` and let's discuss how we can make it better.
For example, extend it for the [custom variables](https://pytorch.org/ignite/concepts.html#state) stored in the `State`.

milesial:
Okay thanks for your response, I will open an issue detailing the problem with `ProgressBar`. The support of custom variables is a good idea. As I'm using `engine.create_supervised_trainer`, the process function is already built in so I went with this workaround :

```python
class Identity(metrics.Metric):
    def reset(self):
        self.output = None

    def update(self, output):
        self.output = output

    def compute(self):
        return self.output

        engine.add_event_handler(Events.ITERATION_COMPLETED, self.completed, name)

trainer = engine.create_supervised_trainer(model, optimizer, loss)
Identity().attach(trainer, 'loss')
ProgressBar().attach(trainer, ['loss'])
```

vfdev-5:
@milesial just a remark, IMO, instead of specific classes as Identity and using create_supervised*, it is better to define a custom processing function and define an engine.

milesial:
I disagree, I think `engine.create_supervised_trainer` is the most intuitive way to do things, for beginner and for clarity. I put more trust in a built-in engine than in a custom one. With the `Identity` class in a separate file, the 3 last lines are self explanatory (except maybe the 2nd one) and concise. For a custom engine, one has to read the processing function in order to understand that is is indeed a simple training loop. `create_supervised_trainer` is clear and documented.

PR #298, opened 2018-10-22
tqdm_logger update (#296)

### LINKED ISSUE #296: Discussion about ProgressBar and metrics display
Hello, I would like to discuss the feature in `contrib.handlers.ProgressBar` that displays metrics on each `Events.ITERATION_COMPLETED` as a postfix of the bar. For some context, please read https://github.com/pytorch/ignite/pull/295.

I think the `ProgressBar` and this feature are essential to the package, as it provides a concise way to have a nice training and evaluation experience (replacing the old way of spamming numbers in the console), and is appealing to beginners.

Here is some of the issues I found.

### Problem
On the doc, the given example is
```python
pbar = ProgressBar()
pbar.attach(trainer, ['loss'])
```

The `metrics` parameter (here `['loss']`) works with `Metric`s that have been attached to the engine, **AND that rewired `self.computed` (which updates the `state.metrics` dict of the engine) on `Events.ITERATION_COMPLETED` instead of `Events.EPOCH_COMPLETED`**. The second part is unintuitive, and not documented. The current document suggests that this might work with all metrics, but this raises a `KeyError` when the dict has not been updated before the end of the first iteration.
Currently, this freature only works with `RunningAverage`.

The problem has been raised by @vfdev-5 in https://github.com/pytorch/ignite/pull/256, but I did not see the discussion go anywhere, I would like for it to continue.

### Proposed solutions

1. The documentation should be updated to reflect the fact that this feature will not work on all of the built-in metrics but `RunningAverage`.

2. The error message for the `KeyError` is `metrics not found in engine.state.metrics` and could be modified to hint for what the issue might be.

3. a Maybe the `Metric` class hierarchy should be updated, to distinguish metrics that update the engine state every epoch, and those that update it every iteration. The `ProgressBar` could then only accept the latter one. 

3. b Or maybe all of the current built-in metrics could update the state of the engine at each iteration (on `self.update` instead of on `self.completed`. This will add computation at each epoch. The progress bar would then work on all metrics.

### Problem
Have a look at the doc example again :
```python
pbar = ProgressBar()
pbar.attach(trainer, ['loss'])
```
This suggests that we can print the training loss, that updates at each iteration.

However, we can't. When we are using a trainer created with `engine.create_supervised_trainer`, the `process_function` returns the loss for the processed batch. I think this is the most intuitive way to do it even for custom training engines.
The `Metric`s that can be attached are computed from the output of this function (`y_pred, y` for an evaluator for example), which may pass through a transform function. You see where I am going: the output of the trainer is already the loss itself, we do not need any further computation.

If we want to print the training loss, we have to attach a `Metric` that does nothing but moves the already computed loss into the `state.metrics` dict of the engine at every iteration. See https://github.com/pytorch/ignite/pull/295 for the `Identity` metric. This is not intuitive, we should be able to display the training loss out of the box.

### Proposed solutions

1. Built-in Identity metric, see https://github.com/pytorch/ignite/pull/295 for the code.

2. Modify `ProgressBar` to allow the display in the postfix the value of `engine.state.output`
    The `attach` method could then be 
    ```python
    attach(self, engine, metric_names=None, display_output=False)
    ```
    with the output looking like 
    `Epoch 1: [59/689]   9%|▊                 , output=3.54e-02 [00:04<00:50]`
3. Modify `ProgressBar` to allow the display in the postfix the value of any custom variables in the `engine.state` object, including the output.

---
Thanks for reading, sorry for the long issue and the english mistakes.

miguelvr:
Fixes #296 

Description:
added `persist` option and `output_transform` argument

Check list:
* [x] New tests are added
* [x] Documentation is updated

### DIFF
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -11,7 +11,7 @@ class ProgressBar:
     TQDM progress bar handler to log training progress and computed metrics.
 
     Args:
-        persist (bool): (Optional) set to ``True`` to persist the progress bar after completion (default = ``False``)
+        persist (bool, optional): set to ``True`` to persist the progress bar after completion (default = ``False``)
 
     Examples:
 
@@ -107,7 +107,7 @@ class ProgressBar:
 
         Args:
             engine (Engine): engine object
-            metric_names (list): (Optional) list of the metrics names to log as the bar progresses
+            metric_names (list, optional): list of the metrics names to log as the bar progresses
             output_transform (Callable, optional): a function to select what you want to print from the engine's
                 output. This function may return either a dictionary with entries in the format of ``{name: value}``,
                 or a single scalar, which will be displayed with the default name `output`.
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -82,8 +82,7 @@ class ProgressBar:
             output_dict = output_transform(engine.state.output)
 
             if not isinstance(output_dict, dict):
-                raise ValueError("output_transform function should return a dict, got {} instead"
-                                 .format(type(output_dict)))
+                output_dict = {"output": output_dict}
 
             metrics.update({name: '{:.2e}'.format(value) for name, value in output_dict.items()})
 
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -108,7 +108,7 @@ class ProgressBar:
         Args:
             engine (Engine): engine object
             metric_names (list): (Optional) list of the metrics names to log as the bar progresses
-            output_transform (Callable, optional) a function to select what you want to print from the engine's
+            output_transform (Callable, optional): a function to select what you want to print from the engine's
                 output. This function may return either a dictionary with entries in the format of ``{name: value}``,
                 or a single scalar, which will be displayed with the default name `output`.
         """
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -24,7 +24,7 @@ class ProgressBar:
             pbar = ProgressBar()
             pbar.attach(trainer)
 
-        Attach metrics that already has been computed (such as `RunningAverage`)
+        Attach metrics that already have been computed at `ITERATION_COMPLETED` (such as `RunningAverage`)
 
         .. code-block:: python
 
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -108,8 +108,9 @@ class ProgressBar:
         Args:
             engine (Engine): engine object
             metric_names (list): (Optional) list of the metrics names to log as the bar progresses
-            output_transform (Callable, optional) a function to transform the engine output into a
-                dictionary with entries in the format of ``{name: value}``
+            output_transform (Callable, optional) a function to select what you want to print from the engine's
+                output. This function may return either a dictionary with entries in the format of ``{name: value}``,
+                or a single scalar, which will be displayed with the default name `output`.
         """
         if metric_names is not None and not isinstance(metric_names, list):
             raise TypeError("metric_names should be a list, got {} instead".format(type(metric_names)))
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -10,6 +10,9 @@ class ProgressBar:
     """
     TQDM progress bar handler to log training progress and computed metrics.
 
+    Args:
+        persist (bool): (Optional) set to True to persist the progress bar after completion (default=False)
+
     Examples:
 
         Create a progress bar that shows you some metrics as they are computed,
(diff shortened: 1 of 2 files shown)


miguelvr:
@vfdev-5 please have a look :)

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -10,47 +10,84 @@ class ProgressBar:
...
+        Simple progress bar
+
+        .. code-block:: python
+
+            trainer = create_supervised_trainer(model, optimizer, loss)
+
+            pbar = ProgressBar()
+            pbar.attach(trainer)
+
+        Attach metrics that already has been computed (such as `RunningAverage`)
```
milesial:
```suggestion
        Attach metrics that already have been computed at `ITERATION_COMPLETED` (such as `RunningAverage`)
```

This better describes it IMO, as the metric isn't already computed at attach time, but should have updated the `state.metrics` of the engine before the bar updates, at ITERATION_COMPLETED

miguelvr: just fixed that :) thanks!

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -65,16 +102,22 @@ def log_message(message):
...
 
-    def attach(self, engine, metric_names=None):
+    def attach(self, engine, metric_names=None, output_transform=None):
         """
         Attaches the progress bar to an engine object
 
         Args:
             engine (Engine): engine object
             metric_names (list): (Optional) list of the metrics names to log as the bar progresses
+            output_transform (function): (Optional) function to transform the engine output into a
```
vfdev-5:
```suggestion
            output_transform (Callable, optional) a function to transform the engine output into a
```

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -10,47 +10,84 @@ class ProgressBar:
...
         if metric_names is not None:
             if not all(metric in engine.state.metrics for metric in metric_names):
                 self._close(engine)
                 raise KeyError("metrics not found in engine.state.metrics")
 
-            metrics = {name: '{:.2e}'.format(engine.state.metrics[name]) for name in metric_names}
+            metrics.update({name: '{:.2e}'.format(engine.state.metrics[name]) for name in metric_names})
+
+        if output_transform is not None:
+            output_dict = output_transform(engine.state.output)
```
vfdev-5: IMO, to force the output as dict is a bit too strict and too verbose if user would like to display a single scalar, e.g. `loss.item()`. However this can be more interesting to display multiple losses without smoothing. Maybe we can let two possibilities for the output: dict and "not dict" which is then wrapped into a dict as : 
```python
output_dict = {"output": output_dict}
```

miguelvr: it is harder to check for the proper output type, however. My bet would be to accept either a scalar or a `dict`

miguelvr:
> @miguelvr what about custom variables ?

what do you mean by that?

vfdev-5:
> what do you mean by that?

@miguelvr take a look [here](https://github.com/pytorch/ignite/issues/296#issuecomment-431697231) at display custom variables.

Actually, I think we will need to do a similar stuff for RunningAverage to support custom variables too...

ignite/contrib/handlers/tqdm_logger.py
vfdev-5: Actually, you are not checking if the output is printable, I think this is OK. So you can just additionally compare if the output is not dict and wrap it to the dict.

miguelvr: > > what do you mean by that?
> 
> @miguelvr take a look [here](https://github.com/pytorch/ignite/issues/296#issuecomment-431697231) at display custom variables.
> 
> Actually, I think we will need to do a similar stuff for RunningAverage to support custom variables too...

I did look at that, but I don't understand the usage of it. Can you give me an example of it being necessary?

miguelvr:
> > what do you mean by that?
> 
> @miguelvr take a look [here](https://github.com/pytorch/ignite/issues/296#issuecomment-431697231) at display custom variables.
> 
> Actually, I think we will need to do a similar stuff for RunningAverage to support custom variables too...

I looked at that but still don't understand the usage. Can you give me an example of such a thing being necessary?

vfdev-5:
@miguelvr for example in a custom processing_function user store aux output in the state:
```python
def processing_function(engine, batch):
    # ...
    engine.state.aux_out = aux_out
    # ...
    return loss_value
```
and would like to display it too. With such option in progress bar we do not force the user to return this aux output with loss value and then dispatch it with the progress bar.
What do you think ?

miguelvr:
In that example you would solve the problem simply by returning the value itself... To me, it seems a bit too cumbersome to support that, since it is not preventing the user to accomplish the same result.

vfdev-5:
I see your point. Okay, let's leave it without custom variable support.

But still I'm not convinced on the output_transform to return directly a `dict`. I think it is better to have two options as discussed earlier...

miguelvr:
> I see your point. Okay, let's leave it without custom variable support.
> 
> But still I'm not convinced on the output_transform to return directly a `dict`. I think it is better to have two options as discussed earlier...

and what is the name displayed then?

vfdev-5:
@miguelvr as I proposed above:
```python
if output_transform is not None:
     output_dict = output_transform(engine.state.output)
    if not isinstance(output_dict, dict): 
        output_dict = {"output": output_dict}
```

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -65,16 +101,23 @@ def log_message(message):
...
         tqdm.write(message)
 
-    def attach(self, engine, metric_names=None):
+    def attach(self, engine, metric_names=None, output_transform=None):
         """
         Attaches the progress bar to an engine object
 
         Args:
             engine (Engine): engine object
             metric_names (list): (Optional) list of the metrics names to log as the bar progresses
```
vfdev-5: @miguelvr could you please change this too
```
metric_names (list): (Optional)
```
to 
```
metric_names (list, optional): 
```
thanks

vfdev-5:
@jasonkriss could you please review and merge if OK

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -10,47 +10,83 @@ class ProgressBar:
...
-            metrics = {name: '{:.2e}'.format(engine.state.metrics[name]) for name in metric_names}
+            metrics.update({name: '{:.2e}'.format(engine.state.metrics[name]) for name in metric_names})
+
+        if output_transform is not None:
+            output_dict = output_transform(engine.state.output)
+
+            if not isinstance(output_dict, dict):
+                output_dict = {"output": output_dict}
+
+            metrics.update({name: '{:.2e}'.format(value) for name, value in output_dict.items()})
```
TheCodez: Why is `{:.2e}` used here? Every _ignite_ examples uses the `{:.2f}` formatting, right?

miguelvr: Exponential notation generalizes better regardless of the type of experiment. If you keep .2f you risk not having enough resolution to check the output

TheCodez: Yes that true, that's why I use `{:.4f}` most of the time as I ike it more than the exponential notation. Maybe there should be a way to specify the format?

vfdev-5: @miguelvr I think we can change to `{:.4f}`. I would prefer.

amitibo:
@miguelvr and @vfdev-5 what about adding the ability to direct the output of the progress bar to a logging stream: https://stackoverflow.com/a/41224909

vfdev-5:
@amitibo sounds good, thanks for the idea! Please create a new issue for that. As we have some handlers (Engine, EarlyStopping, ...) that have their own loggers, we need to think about if their integration is necessary etc.

ignite/contrib/handlers/tqdm_logger.py
miguelvr: @vfdev-5 I currently have a project where even that resolution wouldn't cover one of my variables. I still vouch for the exponential notation as it is what guarantees resolution over all the cases.

vfdev-5: @miguelvr ok, let's keep it `{:.2e}` to guarantee the resolution.

jasonkriss:
LGTM. Thanks @miguelvr!

PR #304, opened 2018-10-30
Plot metrics handler

### LINKED ISSUE #301: Discussion about Metrics plotting
Hello, I would like to discuss the automation the metric plotting.

## Motivation

Currently a big part of (my?) ignite scripts go into the boiler plate of logging the training and testing results, be it a set of **visdom** windows, **mlflow** experiments, logging files or any combination of these. Each time a metric is added or removed, all the logging mechanism needs to be updated accordingly.

## Solution

A possible solution is to have a mechanism similar to the `ProgressBar` (#296), where you create a `Plotting` object (e.g. a wrapper to a **visdom** window or **mlflow** server) and `attach` metrics to it. This object will take care of plotting the metric in the correct context.

An example use case is to have a **visdom** window that plots loss values every iteration or metric values every epoch. Another use case is to log the metrics to an **mlflow** experiment. 

The benefit of this idea is that it will reduce much of the plotting boiler plate that currently has to be "manually" added to event handlers.

## Implementation Discussion

The problem that I see with the `attach` approach (used in the `ProgressBar`) is that it is not clear whether the `EPOCH_COMPLETED` will trigger first the `self.completed` method of the metric, or the plotting callback of the `Plotting` object. In the second case, the metric value will not be ready for plotting.

In my opinion a better method is to add a callback mechanism to the `Metric` class that enables calling arbitrary callbacks (**visdom** plot etc) at the end of the `completed` method.

Another option is to add a wrapper metric (like `RunningAverage`) that will add the callback functionality.

Will appreciate any ideas.

amitibo:
Fixes #301 

Description:
Handlers to plot losses and metrics to [Visdom](https://github.com/facebookresearch/visdom) and [mlflow](https://mlflow.org/) tracking server.

Check list:


vfdev-5:
@amitibo thanks for the PR. 
Some remarks, your code is 3.7 only but we need a code executable compatible 2.7, 3.5...

Concerning mlflow, I thought about a bit different integration, related with a config file runner, [ref](https://github.com/vfdev-5/IgniteConfRunner).

ignite/contrib/handlers/visdom_plotter.py
```diff
@@ -0,0 +1,212 @@
...
+
+    Examples:
+
+    Plotting of trainer loss.
+
+    .. code-block:: python
+
+        trainer = create_supervised_trainer(model, optimizer, loss)
+
+        visdom_plotter = VisdomPlotter(vis, env=env)
```
vfdev-5: Visdom initialization can fail the whole run in the middle of the code. Maybe we could initialize the connection to the server in the file itself "ignite/contrib/handlers/visdom_plotter.py" (and ensure that connection is established on import of visdom_handler)...

amitibo: In most situations, visdom is supposed to run on a separate server. Therefore trying to establish connection on import of the module will fail as there is a need for the server url. The constructor of the handler has the option to pass the server url and establish the connection.

miguelvr:
I would call them VisdomLogger and MLFlowLogger, just to keep consistency with the other handlers

ignite/contrib/handlers/visdom_plotter.py
vfdev-5: Yes, you are right, I forgot about this. Maybe we can define them as environment variables ?

amitibo:
@vfdev-5, does the failure in codecov checks relates to not having tests for these loggers? Any idea how to test these handlers (they require external visdom/mlflow servers)?

vfdev-5:
@amitibo yes the failure is due to code coverage change of `- 8%`.
To test visdom we have a smoke test like this:
https://github.com/pytorch/ignite/blob/4b7481e33d3f1ed75af0533bf1a22064f0005f0c/.travis.yml#L44-L49

ignite/contrib/handlers/visdom_plotter.py
vfdev-5: @amitibo coming back to this one again. IMO user is not intended to reconfigure `VisdomPlotter` on multiple servers. There is a single server running in background and user should know a priori the url. We can thus require to define it as `VISDOM_SERVER_URL` environment variable and check if everything is connectable during import and avoid problems in the middle the code execution. What do you think ?

amitibo:
I am rewriting the `mnist_with_visdom.py` example to use the suggested `VisdomLogger`. Doing so, caused me to think that the logger API is not flexible enough. For example, it doesn't enable plotting plots from different engines (trainer and evaluator) on the same window. I suggest the following API.
```
    avg_loss_win = visdom_logger.create_window(
        window_title="Average Loss",
        ylabel="Loss"
    )
    avg_loss_win.attach(
        engine=train_evaluator,
        metric_names={'train': 'nll'}
    )
    avg_loss_win.attach(
        engine=evaluator,
        metric_names={'val': 'nll'}
    )
```
The API is separated into window creation and engine/event attaching. Note also, the mapping between metric name ('nll') and plot label ('train'/'val').

Also I think of adding plotting of `ParamScheduler`. The API is based on the parameter history, i.e.:
```
    lr_scheduler = CosineAnnealingScheduler(
        optimizer=optimizer,
        param_name='lr',
        ...
        save_history=True
    )
    momentum_scheduler = CosineAnnealingScheduler(
        optimizer=optimizer,
        param_name='momentum',
        ...
        save_history=True
    )
    ...
    trainer.add_event_handler(Events.EPOCH_COMPLETED, lr_scheduler)
    trainer.add_event_handler(Events.EPOCH_COMPLETED, momentum_scheduler)
    ...
    visdom_logger.create_window(
        window_title="Training Parameters"
    ).attach(
        engine=trainer,
        param_history=True
   )
```
The handler will plot all parameters whose history is saved (we can make it the default behavior of the scheduler). The plots will be labeled according to the parameter names.

I will appreciate your opinion.

vfdev-5:
@amitibo I like proposed API for Visdom ! In the case of training you suppose that metrics are computed on epoch end ? Is it possible to indicate a mode (like we wish in tqdm progress bar) of iteration/epoch, if it makes sense (for average running loss during training).

Concerning logging ParamScheduler during the runtime, I think maybe it would be more useful to draw directly optimizer group parameters: lr, momentum etc. What is your opinion ?

amitibo:
Output of `examples/mnist/mnist_with_visdom.py`:
graphs

amitibo:
Should I add a test for the mlflow handler, or remove it from this PR?

vfdev-5:
Yes, maybe better without mlflow for now.
And there is something weird with changed files, added old commits. git rebase ?

amitibo:
> Yes, maybe better without mlflow for now.
> And there is something weird with changed files, added old commits. git rebase ?

Yes, I did rebase in order to merge the new changes from master. Guess it messed up things. Maybe it is better to open a new PR just for the visdom handler (and close this one)?

vfdev-5:
@amitibo you can reset --hard and then cherry pick you commit, no ?
And to remove mlflow maybe manually... Actually, I let you decide what is more simple for you.

amitibo:
> @amitibo you can git reset --hard and then cherry pick your commits, no ?
> And to remove mlflow maybe manually... Actually, I let you decide what is more simple for you.

I opened a new PR (#329). This one (#304) can be closed.

PR #300, opened 2018-10-28
Add ignite.contrib.metrics.MaximumAbsoluteError

zasdfgbnm:
closes: https://github.com/pytorch/ignite/issues/292

PR #305, opened 2018-10-30
move MaximumAbsoluteError to core

zasdfgbnm:
This is a follow up of discussion at:
https://github.com/pytorch/ignite/pull/300#issuecomment-434001627

Feel free to merge or close this, depending on whether `MaximumAbsoluteError` is useful enough to get into core.


zasdfgbnm:
cc: @jasonkriss

vfdev-5:
@zasdfgbnm could you please add some article references or similar where we use such metric and such that user understands better the context ?

zasdfgbnm:
@vfdev-5 I don't think there is an article or something for MaxAE. MaxAE is just something with obvious definition. It is the thing that naturally come to people's mind when they want to know about worst outliers in regression problem. The use case can be very diverse.

For example, below paper list a set of metrics, and MaxAE is included:

vfdev-5:
@zasdfgbnm this paper can be a good reference 👍 
MAE, MSE are rather frequently cited in textbooks on regression tasks.

PS. There is even another citation inside the paper where MaxAE was used previously: Zhang et al, 2015 "A suite of metrics for assessing the performance of solar power forecasting".

zasdfgbnm:
@vfdev-5 Yes, I saw the "A suite of metrics for assessing the performance of solar power forecasting", but I think "solar power forecasting" is not a general enough area (compared with vision, NLP, etc.) to put into ignite and it might confuse users.

vfdev-5:
@zasdfgbnm does it make sens to implement other metrics from 1809 paper, like MdAE, MARE, etc ?

zasdfgbnm:
@vfdev-5 I think it would be nice if we could have as much of these metrics in ignite. These should be a very comprehensive list of metrics used most in machine learning. And these are mostly rather easy to implement.

zasdfgbnm:
So how about this PR? Does this wait for https://github.com/pytorch/ignite/issues/307? And is it necessary to like to that paper in the docstring of this metric, or maybe better somewhere at the the doc of `ignite.metrics`

vfdev-5:
I propose to wait for #307 and yes later link the paper in the docs of `ignite.metrics` or `ignite.metrics.regression` (if we decide to group them in a module).

zasdfgbnm:
I think we will finally move all metrics in a single PR, instead of one by one like this.

Issue #311, opened 2018-11-02
RunningAverage malfunction when attached to multiple outputs


miguelvr:
I just verified in `examples/gan/dcgan.py` that if we attach multiple RunningAverages in a for loop, the value updated to all the metrics is the one corresponding to the last value in the for loop.

The code snippet:
```python
# attach running average 
monitoring_metrics = ['errD', 'errG', 'D_x', 'D_G_z1', 'D_G_z2']
for metric in 
    RunningAverage(alpha=alpha, output_transform=lambda x: x[metric]).attach(trainer, metric)
```

The values being printed in the progress bar (all the same):
```
D_G_z1=5.85e-03, D_G_z2=5.85e-03, D_x=5.85e-03, errD=5.85e-03, errG=5.85e-03 
```

I've check and the output values are correct. What is corrupted is the `engine.state.metrics` values

If you add the outputs without a for loop, it works fine...

```python
RunningAverage(alpha=alpha, output_transform=lambda x: x['errD']).attach(trainer, 'errD')
RunningAverage(alpha=alpha, output_transform=lambda x: x['errG']).attach(trainer, 'errG')
RunningAverage(alpha=alpha, output_transform=lambda x: x['D_x']).attach(trainer, 'D_x')
RunningAverage(alpha=alpha, output_transform=lambda x: x['D_G_z1']).attach(trainer, 'D_G_z1')
RunningAverage(alpha=alpha, output_transform=lambda x: x['D_G_z2']).attach(trainer, 'D_G_z2')
```

vfdev-5:
Thanks for catching this @miguelvr !

vfdev-5:
Fixed by #312

miguelvr:
@vfdev-5 my PR doesn't fix the issue. We should still look into it

vfdev-5:
The problem is with `output_transform=lambda x: x[metric]` and especially with `metric` variable. Here is a snippet:
```python
funcs = []


monitoring_metrics = ['errD', 'errG', 'D_x', 'D_G_z1', 'D_G_z2']
for metric in monitoring_metrics:
    funcs.append(lambda x: metric)

for i, _ in enumerate(monitoring_metrics):
    print(funcs[i](0))

> D_G_z2
> D_G_z2
> D_G_z2
> D_G_z2
> D_G_z2
```

PR #312, opened 2018-11-02
Fix dcgan example

miguelvr:
Makes sure the dcgan example is functioning properly due to `RunningAverage` bug (see #311)


vfdev-5:
Thanks @miguelvr

PR #318, opened 2018-11-06
ParamScheduler to support multiple optimizers/multiple parameter groups

### LINKED ISSUE #317: ParamScheduler doesn't support multiple optimizers/multiple parameter groups
Currently `ParamScheduler` (in the `contrib` module) does not support per-parameter scheduling (https://pytorch.org/docs/stable/optim.html), i.e. the scheduled parameter will be updated uniformly in all the parameter groups of the optimizer. It will be beneficial to be able to update only a specific parameter group.

It also doesn't fully support the use of multiple optimizers. For example if two param_schedulers will be set to update the same parameter of two different optimizers. Both will write to the same `param_history` field stored in the `engine.state`.

Will appreciate any ideas on how to extend the `ParamScheduler` API to support both these cases.

amitibo:
Fixes #317

Description:
Allow parameter scheduler to control a specific parameter group of an optimizer.
Also allow giving a parameter scheduler a different name than the parameter it controls. This way different parameter schedulers will not overwrite the same field in `engine.state.param_history`.

Check list:
* [x ] New doc strings: text and/or example code are in RST format


ignite/contrib/handlers/param_scheduler.py
```diff
@@ -8,28 +8,38 @@ class ParamScheduler(object):
     training.
 
     Args:
-        optimizer (`torch.optim.Optimizer`): the optimizer to use
+        optimizer (`torch.optim.Optimizer`/dict): the optimizer or parameters group to use
```
vfdev-5:
```suggestion
        optimizer (`torch.optim.Optimizer` or dict): the optimizer or parameters group to use
```

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -8,28 +8,38 @@ class ParamScheduler(object):
...
         param_name (str): name of optimizer's parameter to update
         save_history (bool, optional): whether to log the parameter values
             (default=False)
     """
     def __init__(self, optimizer, param_name, save_history=False):
+
         self.optimizer = optimizer
```
vfdev-5: `self.optimizer` becomes unused ?

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -8,28 +8,38 @@ class ParamScheduler(object):
...
+        if name is None:
+            name = self.param_name
+
         if self.save_history:
             if not hasattr(engine.state, 'param_history'):
                 setattr(engine.state, 'param_history', {})
-            engine.state.param_history.setdefault(self.param_name, [])
-            values = [pg[self.param_name] for pg in self.optimizer.param_groups]
-            engine.state.param_history[self.param_name].append(values)
+            engine.state.param_history.setdefault(name, [])
```
vfdev-5: Just wonder whether it should not be 
```
if not hasattr(engine.state, 'param_history'):
    setattr(engine.state, 'param_history', {})
    engine.state.param_history.setdefault(name, [])
```

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -44,7 +54,7 @@ class CyclicalScheduler(ParamScheduler):
     cycle of some size.
 
     Args:
-        optimizer (`torch.optim.Optimizer`): the optimizer to use
+        optimizer (`torch.optim.Optimizer`/dict): the optimizer or parameters group to use
```
vfdev-5:
```suggestion
        optimizer (`torch.optim.Optimizer` or `dict`): the optimizer or parameters group to use
```

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -77,21 +87,21 @@ def __init__(self,
...
+        return super(CyclicalScheduler, self).__call__(engine, name)
 
 
 class LinearCyclicalScheduler(CyclicalScheduler):
     """Linearly adjusts param value to 'end_value' for a half-cycle, then linearly
     adjusts it back to 'start_value' for a half-cycle.
 
     Args:
-        optimizer (`torch.optim.Optimizer`): the optimizer to use
+        optimizer (`torch.optim.Optimizer`/dict): the optimizer or parameters group to use
```
vfdev-5:
```suggestion
        optimizer (`torch.optim.Optimizer` or dict): the optimizer or parameters group to use
```

vfdev-5:
@amitibo please remove not merged updates in ROC_AUC and EpochMetric

ignite/contrib/handlers/param_scheduler.py
amitibo: It is used in ConcatScheduler

ignite/contrib/handlers/param_scheduler.py
amitibo: I think not. `param_history` might already exist (added by another scheduler), but `name` can still be missing.

ignite/contrib/handlers/param_scheduler.py
vfdev-5: @amitibo Okay, I see the dependency, we need to setup the optimizer inside ConcatScheduler. But we can then set `self.optimizer = optimizer` in ConcatScheduler only.

Issue #321, opened 2018-11-07
Arithmetic of metrics?


zasdfgbnm:
Is it good to support arithmetics of metrics?

For example, `metric1 + metric2` gives a new metric whose update function is updating metric1 and metric2, and compute function is the sum of the compute of metric1 and metric2.

The idea comes from wiki: https://en.wikipedia.org/wiki/Sensitivity_and_specificity

There are many things to do with confusion matrix, see the table in https://en.wikipedia.org/wiki/Sensitivity_and_specificity#Definitions

It might be nice to add a metric for counting positive, true positive, negative, true negative, and allow users to combine these values using arithmetics to form their own metric.

vfdev-5:
@zasdfgbnm excelent idea, thanks ! I was thinking about a similar stuff previously in a very simple case of `accuracy` -> `error = 1.0 - accuracy`. 
The API I was thinking of is something like:
```
precision = Precision()
recall = Recall()


def compute_f1(p, r):
    return 2 * p * r / (p + r)

f1= MetricsLambda(compute_f1, precision, recall) 
f1.attach(evaluator, "f1")
```

> It might be nice to add a metric for counting positive, true positive, negative, true negative, and allow users to combine these values using arithmetics to form their own metric.

👍

amitibo:
I am already using a reduction metric:
```
class ReductionMetric(Metric):
    """
    Apply reduction to a list of metrics.

    Args:
        metrics (list of metrics): The metrics to apply reduction to.
        reduction (callable): The reduction operation to apply to the metrics.
        output_transform (callable): a callable that is used to transform the
            model's output into the form expected by the metric. This can be
            useful if, for example, you have a multi-output model and you want to
            compute the metric with respect to one of the outputs.
    """
    def __init__(self, metrics: Union[Tuple, List], reduction=mean, output_transform=lambda x: x):

        self.metrics = metrics
        self.reduction = reduction
        super(ReductionMetric, self).__init__(output_transform=output_transform)

    def reset(self):
        for metric in self.metrics:
            metric.reset()

    def update(self, output):
        for metric in self.metrics:
            o = metric._output_transform(output)
            metric.update(o)

    def compute(self):
        return self.reduction([m.compute() for m in self.metrics])
```
I can create a PR, let me know about the API.

jasonkriss:
Love this idea!

zasdfgbnm:
done

PR #323, opened 2018-11-07
[Step 3/4: Ready for review] ignite.metrics.MetricsLambda

zasdfgbnm:
https://github.com/pytorch/ignite/issues/321

Arithmetics can be be implemented using `MetricsLambda` and will be implemented in a following PR


zasdfgbnm:
https://travis-ci.org/pytorch/ignite/builds/451937490

The build is done, don't know why travis does not report status to github.

vfdev-5:
@zasdfgbnm should be another bug on github :)

ignite/metrics/metrics_lambda.py
```diff
@@ -0,0 +1,34 @@
+from ignite.metrics.metric import Metric
+
+
+class MetricsLambda(Metric):
+    """
+    Apply a function to other metrics to obtain a new metric.
+    The result of the new metric is defined to be the result
+    of applying the function to the result of argument metrics.
+    For example, ``MetricsLambda(lambda x, y: x + y, metric1, 2)``
```
vfdev-5:
```suggestion
    For example, ``MetricsLambda(lambda x, y: x + y, metric1, metric2)``
```

zasdfgbnm:
@vfdev-5 resolved

vfdev-5:
@zasdfgbnm sorry i misunderstood your example at first reading, it was perfect as it was :)

zasdfgbnm:
@vfdev-5 I don't know which is better, we want to make sure that the user understand that we can fuse multiple metrics together with numbers to get a new number, maybe a better example is:
`MetricsLambda(lambda x,y,z: x+y+z, metric1, metric2, 3)`
What do you think?

vfdev-5:
Maybe we can directly setup F2 as example from precision and recall. 
In some sense, using `lambda x, y: 3 + x + y`  on a single input metric can do the job without setting `3` explicitly...

zasdfgbnm:
@vfdev-5 I agree that `lambda x, y: 3 + x + y` is better for this specific application. But I think we should have a constant here, to make sure the user knows they can use constant.

The case in my mind where constant is useful is:

```python
def f(x, y, z, param1, param1):
    pass  # something complicated
m00 = MetricsLambda(f, x, y, z, 0, 0)
m01 = MetricsLambda(f, x, y, z, 0, 1)
m10 = MetricsLambda(f, x, y, z, 1, 0)
m11 = MetricsLambda(f, x, y, z, 1, 1)
```
This would be cleaner than using a lambda for each construct.

zasdfgbnm:
@vfdev-5 Yes, let me also add that example F2

vfdev-5:
> But I think we should have a constant here, to make sure the user knows they can use constant.

In that case add Fbeta with beta parameter as input to lambda :)

zasdfgbnm:
@vfdev-5 How does it look right now?

ignite/metrics/metrics_lambda.py
```diff
@@ -0,0 +1,44 @@
+from ignite.metrics.metric import Metric
+
+
+class MetricsLambda(Metric):
+    """
+    Apply a function to other metrics to obtain a new metric.
+    The result of the new metric is defined to be the result
+    of applying the function to the result of argument metrics.
+    For example, ``MetricsLambda(lambda x, y, z: x + y, metric1, metric2)``
```
jasonkriss: Where is the `z` coming from here? Should we maybe just remove this example in favor of the actual example below?

zasdfgbnm:
@jasonkriss Done!

zasdfgbnm:
@vfdev-5 @jasonkriss reset was not tested previously. Now I add test coverage to it. But this seems to be not trivial because it raises some subtle issue and needs discussion. I'm not sure if what I'm doing now is the best way to do it:

Usually people like to write code like

```python
precision = Precision()
recall = Recall()

def Fbeta(r, p, beta):
     return (1 + beta ** 2) * p * r / (beta ** 2 * p + r)

F1 = MetricsLambda(Fbeta, recall, precison, 1)

precision.attach(precision, "precision")
recall.attach(recall.attach, "recall.attach")
F1.attach(trainer, "F1")
```

This is natural, but wrong, because `precision` and `F1` are using the same precision, and `precision.update` will be called twice! This might cause lots of hard-to-debug problems to users. Currently what I'm doing is, put a sentence in the doc saying that you should copy it manually if this is not your desired behavior, but do you think it's better that we force make copies on `MetricsLambda.__init__`?

vfdev-5:
@zasdfgbnm this is a good point. In case of precision, recall, double update is cancelled as there is a division of double cumulated variables. 
I think a solution can be to remove `update` from `MetricLambda`. So that the API is 
```python
precision = Precision()
recall = Recall()

def Fbeta(r, p, beta):
     return (1 + beta ** 2) * p * r / (beta ** 2 * p + r)

F1 = MetricsLambda(Fbeta, recall, precision, 1)

precision.attach(trainer, "precision")
recall.attach(trainer, "recall")
F1.attach(trainer, "F1")
```
in all cases if user would like or not to compute recall/precision. 

Another solution is to do an introspection during `F1.attach(trainer, "F1")` call and remove all attached update functions for all `MetricsLambda.args` if is Metric.

zasdfgbnm:
@vfdev-5 IMHO, I don't think removing `update` is a good idea, since the user might want only want to keep track of F1 and are not interested in precision and recall.

Also, doing an introspection during `F1.attach(trainer, "F1")` might have similar issue. Consider that the user is interested in F1 and F2, but not precision and recall, it is hard to figure out what F1 and F2 are sharing...

vfdev-5:
>  I don't think removing update is a good idea, since the user might want only want to keep track of F1 and are not interested in precision and recall.

If we remove `update`, we impose to the user even if he/she is not interested by computing recall, precision to use everything like:
```python
precision.attach(trainer, "precision")
recall.attach(trainer, "recall")
F1.attach(trainer, "F1")
```

zasdfgbnm:
> > I don't think removing update is a good idea, since the user might want only want to keep track of F1 and are not interested in precision and recall.
> 
> If we remove `update`, we impose to the user even if he/she is not interested by computing recall, precision to use everything like:
> 
> ```python
> precision.attach(trainer, "precision")
> recall.attach(trainer, "recall")
> F1.attach(trainer, "F1")
> ```

Is this a good design? This is like asking the user to manually manage dependencies. Maybe we can do something for the user, like:

When attaching, these dependencies will be attached automatically. Is this what you mean doing an introspection during `F1.attach(trainer, "F1")`? The point is, during attach, we do not remove anything, but instead we we attach dependencies that is not attached yet.

vfdev-5:
Yes, this seems good ! Maybe we can even generalize (1) behaviour on other `attach` methods in other classes. 
In the implementation, I think engine can have an introspection method like `has_handler(func) : bool`...

zasdfgbnm:
Waiting for https://github.com/pytorch/ignite/pull/325

zasdfgbnm:
@vfdev-5 @jasonkriss This is ready now.

jasonkriss:
LGTM! @vfdev-5 feel free to merge when ready.

ignite/metrics/metrics_lambda.py
```diff
@@ -0,0 +1,55 @@
...
+        f (callable): the function that defines the computation
+        args (sequence): Sequence of other metrics or something
+            else that will be fed to ``f`` as arguments.
+
+    Examples::
+        >>> precision = Precision()
+        >>> recall = Recall()
+        >>> def Fbeta(r, p, beta):
+        >>>     return (1 + beta ** 2) * p * r / (beta ** 2 * p + r)
+        >>> F1 = MetricsLambda(Fbeta, recall, precison, 1)
```
vfdev-5:
```suggestion
        >>> F1 = MetricsLambda(Fbeta, recall, precision, 1)
```

ignite/metrics/metrics_lambda.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class MetricsLambda(Metric):
+    """
+    Apply a function to other metrics to obtain a new metric.
+    The result of the new metric is defined to be the result
+    of applying the function to the result of argument metrics.
+
+    When update, this metric does not recursively update the metrics
+    it depend on. When reset, all its dependency metrics would be
```
vfdev-5:
```suggestion
    it depends on. When reset, all its dependency metrics would be
```

ignite/metrics/metrics_lambda.py
```diff
@@ -0,0 +1,55 @@
...
+    resetted. When attach, all its dependencies would be automatically
+    attached.
+
+    Arguments:
+        f (callable): the function that defines the computation
+        args (sequence): Sequence of other metrics or something
+            else that will be fed to ``f`` as arguments.
+
+    Examples::
```
vfdev-5:
```suggestion
    Examples:
```

zasdfgbnm: @vfdev-5 I think this should be double colon, just like in:
https://pytorch.org/docs/stable/_modules/torch.html#set_default_dtype

vfdev-5: Maybe both works but in `ModelCheckpoint` we have a similar case with single colon:
https://pytorch.org/ignite/_modules/ignite/handlers/checkpoint.html#ModelCheckpoint

zasdfgbnm: http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#literal-blocks

> A paragraph consisting of two colons ("::") signifies that the following text block(s) comprise a literal block. The literal block must either be indented or quoted (see below). No markup processing is done within a literal block. It is left as-is, and is typically rendered in a monospaced typeface:

I think both works. Do you want to change to `:`?

zasdfgbnm:
@vfdev-5 done

ignite/metrics/metrics_lambda.py
```diff
@@ -0,0 +1,55 @@
...
+        args (sequence): Sequence of other metrics or something
+            else that will be fed to ``f`` as arguments.
+
+    Examples:
+        >>> precision = Precision()
+        >>> recall = Recall()
+        >>> def Fbeta(r, p, beta):
+        >>>     return (1 + beta ** 2) * p * r / (beta ** 2 * p + r)
+        >>> F1 = MetricsLambda(Fbeta, recall, precision, 1)
+        >>> F2 = MetricsLambda(Fbeta, recall, precison, 2)
```
vfdev-5: nit: could you please fix typo in all `precison` -> `precision`, lines 26-28

zasdfgbnm:
@vfdev-5 all done

vfdev-5:
@zasdfgbnm thank you !

It is almost perfect and to make it brilliant I would like to integrate the following integration test:
```python
np.random.seed(1)
from pytest import approx

n_iters = 10
batch_size = 10
n_classes = 10

y_true = np.arange(0, n_iters * batch_size) % n_classes
y_pred = np.zeros((n_iters * batch_size, n_classes))
for i in range(n_iters * batch_size):
    if np.random.rand() > 0.7:
        y_pred[i, y_true[i]] = 1.0
    else:
        i = np.random.randint(0, n_classes)
        y_pred[i, i] = 0.7

y_true_batch_values = iter(y_true.reshape(n_iters, batch_size))
y_pred_batch_values = iter(y_pred.reshape(n_iters, batch_size, n_classes))


def update_fn(engine, batch):
    return torch.from_numpy(y_pred_batch), torch.from_numpy(y_true_batch)


evaluator = Engine(update_fn)

recall = Recall(average=True)

precision_orig = Precision(average=True)
recall_orig = Recall(average=True)


def Fbeta(r, p, beta):
     return (1 + beta ** 2) * p * r / (beta ** 2 * p + r)


F1 = MetricsLambda(Fbeta, recall, precision, 1)

precision_orig.attach(evaluator, "precision_orig")
recall_orig.attach(evaluator, "recall_orig")

precision.attach(evaluator, "precision")
recall.attach(evaluator, "recall")
F1.attach(evaluator, "f1")

data = list(range(n_iters))
state = evaluator.run(data, max_epochs=1)

print(state.metrics)

precision_true = precision_score(y_true, np.argmax(y_pred, axis=-1), average='macro')
recall_true = recall_score(y_true, np.argmax(y_pred, axis=-1), average='macro')
f1_true = f1_score(y_true, np.argmax(y_pred, axis=-1), average='macro')

assert precision_true == approx(state.metrics['precision_orig']), "{} vs {}".format(precision_true, state.metrics['precision_orig'])
assert precision_true == approx(state.metrics['precision']), "{} vs {}".format(precision_true, state.metrics['precision'])
assert recall_true == approx(state.metrics['recall_orig']), "{} vs {}".format(recall_true, state.metrics['recall_orig'])
assert recall_true == approx(state.metrics['recall']), "{} vs {}".format(recall_true, state.metrics['recall'])
assert f1_true == approx(state.metrics['f1']), "{} vs {}".format(f1_true, state.metrics['f1'])
```

However, `f1_true` is not equal to `state.metrics['f1']`. `state.metrics['f1']` is really equal to `2 * P * R / (R + P)`. Let me found out where is the problem and if you could then insert this in test it would be super :)

vfdev-5:
Here is the correct test:
```python
import torch

from ignite.metrics import Precision, Recall


np.random.seed(1)
from pytest import approx

n_iters = 10
batch_size = 10
n_classes = 10

y_true = np.arange(0, n_iters * batch_size) % n_classes
y_pred = 0.2 * np.random.rand(n_iters * batch_size, n_classes)
for i in range(n_iters * batch_size):
    if np.random.rand() > 0.4:
        y_pred[i, y_true[i]] = 1.0
    else:
        j = np.random.randint(0, n_classes)
        y_pred[i, j] = 0.7

y_true_batch_values = iter(y_true.reshape(n_iters, batch_size))
y_pred_batch_values = iter(y_pred.reshape(n_iters, batch_size, n_classes))


def update_fn(engine, batch):
    return torch.from_numpy(y_pred_batch), torch.from_numpy(y_true_batch)


evaluator = Engine(update_fn)

def Fbeta(r, p, beta):
     return torch.mean((1 + beta ** 2) * p * r / (beta ** 2 * p + r)).item()


F1 = MetricsLambda(Fbeta, recall, precision, 1)

precision.attach(evaluator, "precision")
recall.attach(evaluator, "recall")
F1.attach(evaluator, "f1")

data = list(range(n_iters))
state = evaluator.run(data, max_epochs=1)

precision_true = precision_score(y_true, np.argmax(y_pred, axis=-1), average=None)
recall_true = recall_score(y_true, np.argmax(y_pred, axis=-1), average=None)
f1_true = f1_score(y_true, np.argmax(y_pred, axis=-1), average='macro')

assert precision_true == approx(state.metrics['precision']), "{} vs {}".format(precision_true, state.metrics['precision'])
assert recall_true == approx(state.metrics['recall']), "{} vs {}".format(recall_true, state.metrics['recall'])
assert f1_true == approx(state.metrics['f1']), "{} vs {}".format(f1_true, state.metrics['f1'])
```

Actually we should not average precision and recall before computing F1. We need to change the example by precising `average=False`:
```

def Fbeta(r, p, beta):
     return torch.mean((1 + beta ** 2) * p * r / (beta ** 2 * p + r)).item()

```

vfdev-5:
> The code given has problem of comparing two arrays using ==, I change it to use `numpy.testing.assert_allclose`.

@zasdfgbnm normally, `approx` works, tested on my side. Seems strange ...

ignite/metrics/metrics_lambda.py
```diff
@@ -0,0 +1,55 @@
...
+    resetted. When attach, all its dependencies would be automatically
+    attached.
+
+    Arguments:
+        f (callable): the function that defines the computation
+        args (sequence): Sequence of other metrics or something
+            else that will be fed to ``f`` as arguments.
+
+    Examples:
+        >>> precision = Precision()
```
vfdev-5:
```suggestion
        >>> precision = Precision(average=False)
```

ignite/metrics/metrics_lambda.py
```diff
@@ -0,0 +1,55 @@
...
+        >>> recall = Recall()
```
vfdev-5:
```suggestion
        >>> recall = Recall(average=False)
```

ignite/metrics/metrics_lambda.py
```diff
@@ -0,0 +1,55 @@
...
+        >>> def Fbeta(r, p, beta):
+        >>>     return (1 + beta ** 2) * p * r / (beta ** 2 * p + r)
```
vfdev-5:
```suggestion
        >>>     return torch.mean((1 + beta ** 2) * p * r / (beta ** 2 * p + r)).item()
```

zasdfgbnm:
@vfdev-5 I get:

```
  File "tests/ignite/metrics/test_metrics_lambda.py", line 127, in <module>
    test_integration()
  File "tests/ignite/metrics/test_metrics_lambda.py", line 120, in test_integration
    print(approx(state.metrics['precision']))
  File "/home/gaoxiang/.virtualenvs/pt/lib/python3.7/site-packages/_pytest/python_api.py", line 526, in approx
    raise _non_numeric_type_error(expected, at=None)
TypeError: cannot make approximate comparisons to non-numeric values: tensor([0.7500, 0.7778, 0.7500, 0.4706, 0.6667, 0.5714, 0.7778, 0.6667, 0.8889,
        0.7273], dtype=torch.float64)
```

ignite/metrics/metrics_lambda.py
zasdfgbnm: @vfdev-5 Shall we update the unit test accordingly?

vfdev-5: It is the case, not ?

vfdev-5:
@zasdfgbnm the error you have probably is OK, as I have an error too on calling
```python
print(approx(state.metrics['precision']))
```
but what it gives you on comparision:
```python
print(precision_true == approx(state.metrics['precision']))
```

zasdfgbnm:
@vfdev-5 But I'm getting the same error for compare:

```
  File "tests/ignite/metrics/test_metrics_lambda.py", line 121, in <module>
    test_integration()
  File "tests/ignite/metrics/test_metrics_lambda.py", line 118, in test_integration
    assert precision_true == approx(state.metrics['precision']), "{} vs {}".format(precision_true, state.metrics['precision'])
  File "/home/gaoxiang/.virtualenvs/pt/lib/python3.7/site-packages/_pytest/python_api.py", line 526, in approx
    raise _non_numeric_type_error(expected, at=None)
TypeError: cannot make approximate comparisons to non-numeric values: tensor([0.7500, 0.7778, 0.7500, 0.4706, 0.6667, 0.5714, 0.7778, 0.6667, 0.8889,
        0.7273], dtype=torch.float64) 
```

vfdev-5:
@zasdfgbnm OK, I think I can merge it, thank you for the PR !

PR #324, opened 2018-11-07
[Step 4/4: Ready for review] metrics arithmetic

zasdfgbnm:
Depend on: https://github.com/pytorch/ignite/pull/323
See also: https://github.com/pytorch/ignite/issues/321


jasonkriss:
Very cool! Let's wait til #323 is merged and then we'll take care of this.

zasdfgbnm:
Finally all tests pass. I hate python2 compatibility :-(

zasdfgbnm:
@vfdev-5 Final step here, and then we will have metric arithmetics :)

docs/source/metrics.rst
```diff
@@ -37,6 +37,13 @@ use the `output_transform` argument to transform it:
         metric = Accuracy(output_transform=output_transform)
         metric.attach(engine, "accuracy")
 
+Metrics could be combined together to form a new metric through arithmetics,
+for example:
+    .. code-block:: python
+        precision = Precision()
+        recall = Recall()
+        F1 = precision * recall * 2 / (precision + recall)
```
vfdev-5: According to the integration test of #323 this computation is not a real F1. Maybe we need replace this example by something else.

zasdfgbnm: @vfdev-5 We can just add `average=False` and then this will be the F1 for each class. I just added an integration test, you can take a look.

vfdev-5: @zasdfgbnm looks good !

tests/ignite/metrics/test_metric.py
```diff
@@ -1,7 +1,12 @@
-from ignite.metrics import Metric
+from ignite.metrics import Metric, Precision, Recall
 from ignite.engine import Engine, State
 import torch
 from mock import MagicMock
+
+from pytest import approx
```
vfdev-5:
```suggestion
from pytest import approx
import sys
```

docs/source/metrics.rst
```diff
@@ -37,6 +37,13 @@ use the `output_transform` argument to transform it:
         metric = Accuracy(output_transform=output_transform)
         metric.attach(engine, "accuracy")
 
+Metrics could be combined together to form a new metric through arithmetics,
+for example:
+    .. code-block:: python
+        precision = Precision(average=False)
+        recall = Recall(average=False)
+        F1 = precision * recall * 2 / (precision + recall)
+
```
vfdev-5: Maybe we could add suplementary note point out that user should carefully think about combining metrics as F1 example and averaging as this can silently produce wrong result instead of expected.

vfdev-5:
@jasonkriss , @alykhantejani review please and merge if OK

PR #325, opened 2018-11-08
[Step 1/4: Ready for review] Add Engine.has_event_handler

zasdfgbnm:
As discussed in https://github.com/pytorch/ignite/pull/323#issuecomment-436819733

@vfdev-5 Which do you think we should have?
`has_handler(func)`, or `has_event_handler(event_name, func)`, or both?

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -142,16 +142,24 @@ class Engine(object):
         self._event_handlers[event_name].append((handler, args, kwargs))
         self._logger.debug("added handler for event %s ", event_name)
 
-    def has_event_handler(self, event_name, handler):
+    def has_event_handler(self, handler, event_name=None):
         """Check if the specified event has the specified handler.
 
         Args:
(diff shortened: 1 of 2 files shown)


zasdfgbnm:
Where is the CI?

zasdfgbnm:
@vfdev-5 I finally make it `has_event_handler(func, event_name=None)`. I choose `has_event_handler` instead of `has_handler` just to be consistent with `add_event_handler`. Let me know if you think `has_handler` is better.

vfdev-5:
@zasdfgbnm I think it is ok for the name. Looks good.

@alykhantejani @jasonkriss please review. This feature is needed as discussed in #323 (the link in the description above) to verify if a handler is already attached and to be able to attach a handler only once.

zasdfgbnm:
Kindly ping @alykhantejani , no hurry, just in case you forget this

zasdfgbnm:
Next step is: https://github.com/pytorch/ignite/pull/326

PR #326, opened 2018-11-08
[Step 2/4: Ready for review] Allow attaching metric to engine multiple times

zasdfgbnm:
Depends on https://github.com/pytorch/ignite/pull/325
See also: https://github.com/pytorch/ignite/pull/323#issuecomment-436815481

### DIFF
--- a/docs/source/contrib/engines.rst
+++ b/docs/source/contrib/engines.rst
@@ -8,3 +8,4 @@ Contribution module of engines
 .. automodule:: ignite.contrib.engines
    :members:
    :undoc-members:
+   :imported-members:
--- a/docs/source/contrib/handlers.rst
+++ b/docs/source/contrib/handlers.rst
@@ -8,3 +8,4 @@ Contribution module of handlers
 .. automodule:: ignite.contrib.handlers
    :members:
    :undoc-members:
+   :imported-members:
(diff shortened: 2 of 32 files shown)


zasdfgbnm:
Why the CI build is canceled?

vfdev-5:
> Why the CI build is canceled?

This must be a bug in travis. I just restarted previously failed test job on py35

vfdev-5:
@jasonkriss please review and merge if OK

jasonkriss:
Looks good! Thanks @zasdfgbnm!

vfdev-5:
@zasdfgbnm sorry for revealing this one. I'm just wondering in which case actually user would like to add the same handler to the same event multiple times ?

zasdfgbnm:
@vfdev-5  `add_event_handler` has `args` and `kwargs` which would be passed to handler function. The user might create handlers with same function, but different arguments. Such as:

```python
def write_result_to_file(engine, index):
    result = engine.state.xxx[index]
    result = .....  # do some computation
    open('result-{}.txt'.format(index), 'w').write(result)

trainer.add_event_handler(write_result_to_file, 'result1')
trainer.add_event_handler(write_result_to_file, 'result2')
trainer.add_event_handler(write_result_to_file, 'result3')
```

vfdev-5:
@zasdfgbnm Okay, this can make sense. Let's leave as is.

PR #327, opened 2018-11-08
Added D1 Regression Metrics

### LINKED ISSUE #307: More metrics for regression tasks and forecasting

D1 = magnitude of error
- Mean Error
- Mean Normalized Bias
- Fractional Bias

D2 = absolute error
- Median Absolute Error
- Wave Hedges Distance
- Fractional absolute error
- Canberra Metric

based on the paper: 
See appendix 2, page 30.

What do you think ?

anmolsjoshi:
Related to #307 

Description:
Adds D1 regression metrics.

Check list:
* [x] New tests are added. 


vfdev-5:
@anmolsjoshi I edited your message : `Fixes` -> `Related to` as this PR fixes partly the issue and if we keep Fixes when PR is accepted, the issue will be closed.

vfdev-5:
@anmolsjoshi could you put the code into contrib please.

anmolsjoshi:
@vfdev-5 metrics have been moved to contrib and tests have been added. 

For Fractional Bias, I was thinking of adding a test to check if any element of y_pred + y is 0. Thoughts?

docstrings need to be updated as in #331 - will commit that later tonight.

ignite/contrib/metrics/fractional_bias.py
```diff
@@ -0,0 +1,47 @@
...
+class FractionalBias(Metric):
+    r"""
+    Calculates the Fractional Bias.
+
+    It has been proposed in `Performance Metrics (Error Measures) in Machine Learning Regression, Forecasting and
+    Prognostics: Properties and Typology`.
+
+    More details can be found in `here`_.
+
+    :math:`\text{FB} = \frac{1}{n}\sum _j^n\frac{2 * (A_j - P_j}{A_j + P_j})`
```
vfdev-5:
```suggestion
    :math:`\text{FB} = \frac{1}{n}\sum _j^n\frac{2 * (A_j - P_j)}{A_j + P_j}`
```

ignite/contrib/metrics/mae.py
```diff
@@ -4,11 +4,25 @@
...
-    Calculates the maximum absolute error.
+    r"""
+    Calculates the Maximum Absolute Error.
+
+    It has been proposed in `Performance Metrics (Error Measures) in Machine Learning Regression, Forecasting and
+    Prognostics: Properties and Typology`.
+
+    More details can be found in `here`_.
+
+    :math:`\text{MaxAE} = \max_{j=1,n} \left(|A_j-P_j\right|)`
```
vfdev-5:
```suggestion
    :math:`\text{MaxAE} = \max_{j=1,n} \left( \lvert A_j-P_j \rvert \right)`
```

ignite/contrib/metrics/mae.py
```diff
@@ -4,11 +4,25 @@
 
 
 class MaximumAbsoluteError(Metric):
```
vfdev-5: Could you please also rename the file `mae.py` -> `maximum_absolute_error.py` to have a coherence with others.

ignite/contrib/metrics/fractional_bias.py
anmolsjoshi: Done!

ignite/contrib/metrics/mae.py
anmolsjoshi: Done!

ignite/contrib/metrics/mae.py
anmolsjoshi: Done!

anmolsjoshi:
@vfdev-5 all comments have been incorporated. 

- docstrings have been updated as requested.
- mean_percentage_error metric and tests have been deleted. 
- regression module has been added, containing all regression related metrics. 
- mae has been renamed as maximum_absolute_error.
- all the tests have been updated appropriately.

vfdev-5:
@jasonkriss or @alykhantejani review and merge please

tests/ignite/contrib/metrics/regression/test_fractional_bias.py
```diff
@@ -0,0 +1,46 @@
...
+
+def test_fractional_bias():
+    a = np.random.randn(4)
+    b = np.random.randn(4)
+    c = np.random.randn(4)
+    d = np.random.randn(4)
+    ground_truth = np.random.randn(4)
+
+    m = FractionalBias()
+    m.reset()
```
zasdfgbnm: No need to do `reset` here, it would be called automatically by `__init__`. As in https://github.com/pytorch/ignite/blob/master/ignite/metrics/metric.py#L23

Same for other metrics.

tests/ignite/contrib/metrics/regression/test_manhattan_distance.py
```diff
@@ -0,0 +1,38 @@
+from ignite.exceptions import NotComputableError
+from ignite.contrib.metrics.regression import ManhattanDistance
+import torch
+import numpy as np
+import pytest
+
+
+def test_zero_div():
```
zasdfgbnm: There is no div at all in `ManhattanDistance`, so this shouldn't be called `test_zero_div`

ignite/contrib/metrics/regression/fractional_bias.py
```diff
@@ -0,0 +1,45 @@
...
+        self._sum_of_errors = 0.0
+        self._num_examples = 0
+
+    def update(self, output):
+        y_pred, y = output
+
+        denominator = y_pred + y.view_as(y_pred)
+
+        if (denominator == 0).any():
+            raise NotComputableError('The denominator has 0.')
```
zasdfgbnm: @vfdev-5 for this case, different from https://github.com/pytorch/ignite/pull/331, should we just leave a warning and return an `inf` or `nan` whichever the result is?  The comment in https://github.com/pytorch/ignite/pull/331#issuecomment-437721992 no longer applies, because here the `y_pred` is the value of model output, which is not under the control of user's hand. During training, `y_pred` could be any ridiculous number. Users might be interested in this metric (because when the model is good enough, this is well-defined), and doesn't care if occasionally this is ill-defined during training. In this case, raising `NotComputableError` might be annoying to the user because this stops their training.

vfdev-5:
@zasdfgbnm thanks for the review !

> Also, @vfdev-5 what is the best way to reference the paper in docstring? My personal opinion is this way: Botchkarev 2018, that is use name + year as link text, and use the abstract page of arXiv as link. Whichever we choose, we should be consistent in #331 as well. cc: @weihuangxu

I have no special preference for this, `Botchkarev 2018` seems ok for me. The problem could be with with sphinx docs rendering. Maybe we can do the same as in PyTorch, for example, [citation of ADADELTA](https://pytorch.org/docs/stable/_modules/torch/optim/adadelta.html#Adadelta).

ignite/contrib/metrics/regression/manhattan_distance.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Where, :math:`A_j` is the ground truth and :math:`P_j` is the predicted value.
+
+    - `update` must receive output of the form `(y_pred, y)`.
+    - `y` and `y_pred` must be of same shape.
+
+    """
+    def reset(self):
+        self._sum_of_errors = 0.0
+        self._num_examples = 0
```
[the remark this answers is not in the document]
vfdev-5: @zasdfgbnm correct, unused variable, thanks for catching this !

tests/ignite/contrib/metrics/regression/test_manhattan_distance.py
vfdev-5:
```suggestion
def test_no_elements():
```

ignite/contrib/metrics/regression/manhattan_distance.py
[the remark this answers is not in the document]
zasdfgbnm: @vfdev-5 What is your opinion on the second part of my comment? i.e. about

> All metrics that has not been updated yet after last reset should raise `NotComputableError` when `compute`

ignite/contrib/metrics/regression/fractional_bias.py
vfdev-5: @zasdfgbnm I agree, thanks for pointing on this ! Let's use warnings.warn('The denominator is 0. Error becomes NaN') and set `_sum_of_errors` to np.nan.

ignite/contrib/metrics/regression/manhattan_distance.py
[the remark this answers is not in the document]
zasdfgbnm: @vfdev-5 here for this case, instead of using `_num_examples`, it might be better to use a boolean variable called `_updated`, or even just remove the check (I don't know which choice is better).

ignite/contrib/metrics/regression/fractional_bias.py
zasdfgbnm: @vfdev-5 how about `inf`? Would it be better to leave it an `inf` if denominator is 0 but numerator is not?

ignite/contrib/metrics/regression/manhattan_distance.py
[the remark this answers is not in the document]
vfdev-5: @zasdfgbnm this is a correct assumption. How would you implement this ?
A resetable counter of `update` method that is checked in `compute`... :|

ignite/contrib/metrics/regression/fractional_bias.py
vfdev-5: Mathematically correct would be plus or minus infinity. Maybe we can do similarly to torch:
```
import torch

torch.ones(2) / torch.eye(2)

tensor([[1., inf],
        [inf, 1.]])
```

vfdev-5: > I feel here we are a bit over-worried about divide by zero case. 

Yes, probably this is the best path to take :) Let's see others opinion. But I think you are right, we probably should just remove this `raise`.

ignite/contrib/metrics/regression/manhattan_distance.py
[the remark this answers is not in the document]
zasdfgbnm: @vfdev-5 My suggestion is:
1. For the purpose of this PR, just remove `_num_examples` and don't worry about not being updated. If we are worried about this metric, then we should also be worried about more other metrics, so it worth a separate PR to correct all of them.
2. If we are using metrics only with engines, then it should be safe to assume a metric will never be computed before updated.
3. If we still want to check if it is updated before compute, some change in `Metric` like this might work:

```python
class Metric:
    def  __init__(self, ...):
        def wrapped_reset():
            self._updated = False
            self.reset()
        self.reset = functool.wraps(self.reset)(wrapped_reset)

        def wrapped_update(output):
            self._updated = True
            self.update(output)
        self.update =  functool.wraps(self.update)(wrapped_update)

        def wrapped_compute():
            if not self._updated:
                raise NotComputableError('not updated before compute')
            return self.compute()
        self.compute =  functool.wraps(self.compute)(wrapped_compute)
```

tests/ignite/contrib/metrics/regression/test_fractional_bias.py
anmolsjoshi: Incorporated this comment.

ignite/contrib/metrics/regression/manhattan_distance.py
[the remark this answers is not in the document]
anmolsjoshi: For ManhattanDistance, since there is no division, it might be best to just return whatever the current value of self._sum_of_errors is. I have removed NotComputableError for this metric.

tests/ignite/contrib/metrics/regression/test_manhattan_distance.py
anmolsjoshi: Please see comment above about removing NotComputableError entirely.

anmolsjoshi: Please see comment above.

ignite/contrib/metrics/regression/fractional_bias.py
anmolsjoshi: I agree with setting this as inf and removing the raise error.

anmolsjoshi:
@vfdev-5 @zasdfgbnm thanks for the review and discussion! Please see me latest commit.

- removed m.reset() from all the tests per this [comment](https://github.com/pytorch/ignite/pull/327#discussion_r233275118)
- changed docstring to simple author + year, with link to abstract per this [comment](https://github.com/pytorch/ignite/pull/327#issuecomment-438498035)
- removed raise error of denominator == 0 in fractional_bias per this [comment](https://github.com/pytorch/ignite/pull/327#discussion_r233286181)
- removed manhattan_distance test and raise error for 0 iterations per this [comment](https://github.com/pytorch/ignite/pull/327#discussion_r233265126)

vfdev-5:
@jasonkriss or @alykhantejani please review and merge if OK

anmolsjoshi:
@vfdev-5 @zasdfgbnm please note that I have updated the metrics to calculate e = A - P. 

Last few commits had the opposite, tests have been updated also.

jasonkriss:
@anmolsjoshi Can you resolve the conflicts? Sorry about that. We'll get this merged as soon as that happens.

anmolsjoshi:
@jasonkriss I have resolved the conflict. let me know if there are any issues.

jasonkriss:
Thanks @anmolsjoshi!

PR #329, opened 2018-11-09
Added visdom logger handler.

### LINKED ISSUE #301: Discussion about Metrics plotting

amitibo:
Handler that plots to [Visdom](https://github.com/facebookresearch/visdom).

graphs

Fixes #301
Replaces #304 

Description:
This handler introduces API to create attach visdom plots to metrics and parameters of the training.

Modified the mnist_with_visdom.py to use the new handler.
Added installation of typing module to .travis.yml

Check list:


amitibo:
Any ideas how I can make `codecov` happy?

vfdev-5:
@amitibo take a look [here](https://codecov.io/gh/pytorch/ignite/compare/2f48a098cae1208ecf1a5c0e78f3583e6b3b4d2d...f9e866b66d05eda2cb3a79376af9ff9fe37652ef/tree/ignite/contrib/handlers), it says that `visdom_logger.py ` is covered by 15%.
and similar message is shown in github:
```
codecov/patch — 16% of diff hit (target 90%)
```
Maybe we need to improve the data sent to codecov to see which part of code is not covered. The same stuff you can do locally by running something like :
```
py.test --cov ignite --cov-report term-missing tests/ignite/contrib/handlers/visdom_logger.py
```

vfdev-5:
@amitibo maybe we could use `mock` to emulate visdom server behaviour instead of launching real server ?

vfdev-5:
@amitibo I would like to discuss about how to make unified API between this logger and tensorboard logger. 

In some sense, at first glance window abstraction seems to be unnecessary if we do not want to draw in the same window, e.g. train loss, learning rate, ...
On the other hand, drawing metrics of train and val in the same window looks great and this is achieved with this window abstraction. In tensorboard it seems to me we can not draw two curves (train, val) like this in the same window. But in tensorboard we can compare curves of different experiments. Is it possible with Visdom ?


cc @anmolsjoshi

amitibo:
> @amitibo maybe we could use `mock` to emulate visdom server behaviour instead of launching real server ?

I agree but it will have to wait till I have more time.

amitibo:
> @amitibo I would like to discuss about how to make unified API between this logger and tensorboard logger.
> 
> In some sense, at the first glance window abstraction seems to be unnecessary if we do not want to draw in the same window, e.g. train loss, learning rate, ...
> On the other hand, drawing metrics of train and val in the same window looks great and this is achieved with this window abstraction.

This is the reason I added the windows abstraction.

> In tensorboard it seems to me we can not draw two curves (train, val) like this in the same window. But in tensorboard we can compare curves of different experiments. Is it possible with Visdom ?

Yes, but it is part of Visdom, not the handler. I think that it also requires that the windows share the same name, i.e. you compare (visdom) environments, not specific graphs.

vfdev-5:
> Did `codecov` see the test file I added (`tests/ignite/contrib/handlers/visdom_logger.py`)?

Seems like it does not see. [ref](https://travis-ci.org/pytorch/ignite/jobs/453737518#L1127)

ignite/contrib/handlers/visdom_logger.py
```diff
@@ -0,0 +1,305 @@
...
+from typing import Callable, Dict, List, Union
+
+
+class _VisdomWindows:
+    """Object encapsulating a Visdom Window.
+
+    Args:
+        vis (Visdom object): Visdom client.
+        env (str): Name of Visdom environment for the graphs.
+        save_by_default (bool): The graphs will be saved by default by the server.
```
vfdev-5: This option name is not clear. According to the API https://github.com/facebookresearch/visdom#vissave
this saves some environment of the server. Maybe we could leave it by default as true ?

vfdev-5:
@amitibo I updated travis and the code of the `test_visdom_logger.py` so that the tests pass but not the coverage.

amitibo:
> @amitibo I updated travis and the code of the `test_visdom_logger.py` so that the tests pass but not the coverage.

Does this require more tests to the visdom logger?

ignite/contrib/handlers/visdom_logger.py
amitibo: This option saves the environment (your current plots) on the server side. Without this option if the server is closed (crashes) and opened again, all past work is lost.
I agree that it is better to have it True by default.

vfdev-5:
Yes, more tests needed

amitibo:
> Yes, more tests needed

Maybe its better to first decide on the final API (considering also #265), before tests. What do you think @vfdev-5 ?

anmolsjoshi:
Closing this in favor of #455

PR #331, opened 2018-11-11
Implement Mean Absolute Relative Error

### LINKED ISSUE #307: More metrics for regression tasks and forecasting

weihuangxu:
Fixes #307 

Description:
Add mean absolute relative error to metrics. 

Check list:


vfdev-5:
@weihuangxu thanks for the PR, could you please put PR template back.

ignite/contrib/metrics/MARE.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import division
+from ignite.exceptions import NotComputableError
+from ignite.metrics import Metric
+import torch
+
+
+class MeanAbsoluteRelativeError(Metric):
+    """
+    Calculate Mean Absolute Relative Error
+
```
vfdev-5: In case of #307 maybe we can put some metrics mathematical definitions and link to the article https://arxiv.org/ftp/arxiv/papers/1809/1809.03006.pdf

weihuangxu:
@vfdev-5 Sure. I have fixed the template. There is an issue regarding the devision in mean absolute relative error. If the ground truth is 0, it may raise numerical issue. I am not sure if we should raise an error or just show Nan in the output. What do you think?

ignite/contrib/metrics/MARE.py
weihuangxu: That's good idea. I can add equations and link to the paper.

weihuangxu: @vfdev-5 It's done. Let me know if there is more modifications needed.

ignite/contrib/metrics/MARE.py
```diff
@@ -0,0 +1,35 @@
...
+    def update(self, output):
+        y_pred, y = output
+        absolute_error = torch.abs(y_pred - y.view_as(y_pred)) / torch.abs(y.view_as(y_pred))
+        self._sum_of_absolute_relative_errors += torch.sum(absolute_error).item()
+        self._num_samples += y.shape[0]
+
+    def compute(self):
+        if self._num_samples == 0:
+            raise NotComputableError('MeanAbsoluteError must have at least one example before it can be computed')
```
TheCodez: Should be `MeanAbsoluteRelativeError` instead.

vfdev-5:
@weihuangxu maybe we could add a flag ~ `percent=True` to return Mean Absolute Percentage Error = MARE * 100 ?

> There is an issue regarding the devision in mean absolute relative error. If the ground truth is 0, it may raise numerical issue. I am not sure if we should raise an error or just show Nan in the output. What do you think?

@weihuangxu good question, not sure how to regularize this properly. 
Maybe we can add something like:
```
a = 0
if error < tol (1e-20):
     MARE = 0
else:
     MARE = Inf
``` 
@zasdfgbnm any ideas ?

ignite/contrib/metrics/MARE.py
vfdev-5: @TheCodez thank you :)

vfdev-5:
```suggestion
            raise NotComputableError('MeanAbsoluteRelativeError must have at least one example before it can be computed')
```

zasdfgbnm:
@vfdev-5 Considering metric arithmetic (https://github.com/pytorch/ignite/issues/321) is coming soon, would it be helpful to complicate the API by adding a `percent` argument? Builtin metrics like `Precision` does not have such thing.

For the divide by zero error, I would suggest just ignore the case that ground truth is zero, doing so the final result would automatically becomes `inf`,  and the user should know  preprocess their dataset if they want MARE.

vfdev-5:
@zasdfgbnm good point with `percent`. Ok, let's skip this flag.

> For the divide by zero error, I would suggest just ignore the case that ground truth is zero, doing so the final result would automatically becomes inf, and the user should know preprocess their dataset if they want MARE.

So you suggest to ignore the cases with gt = 0 ? 

I did not get your point about "the user should know preprocess their dataset if they want MARE." ...

zasdfgbnm:
@vfdev-5 as from the definition, MARE = sum(|error/gt|), gt is something given in the dataset, which I assume the users should but might not know well. If something in the dataset has gt=0, why do they insist on using ill-defined metrics like MARE? So all we need to do is to notify the user that they should not use this metric, by returning a `inf` or `nan` or raise an Exception.

So, considering that if we don't add any special handle of case gt=0, the user would know that they are doing things wrongly by getting a `inf` or `nan` result. But if we add 
```python
if error < tol (1e-20):
     MARE = 0
else:
     MARE = Inf
```
the user might sometimes get 0 instead of `nan`, which I consider as a silent failure.

vfdev-5:
@zasdfgbnm Ok, I see your point. Let's raise an exception `NotComputableError` with a message that the ground truth is zero.

weihuangxu:
@TheCodez Thanks for pointing out. 

@vfdev-5 I added the error message in update step when the ground truth has 0 value.

vfdev-5:
@weihuangxu thanks! Some comments in the code and could you also please check if docs are correctly rendered with formula:
```
cd docs
pip install -r docs/requirements.txt
make html
cd build && python3 -m http.server 1234
```
and go in the browser to 0.0.0.0:1234

ignite/contrib/metrics/MARE.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    def reset(self):
+        self._sum_of_absolute_relative_errors = 0.0
+        self._num_samples = 0
+
+    def update(self, output):
+        y_pred, y = output
+        if (y == 0).any():
+            raise NotComputableError('The ground truth has 0')
```
vfdev-5: Could you please also add a test on this

weihuangxu: Sure. I will work on this soon.

weihuangxu:
@vfdev-5  Thanks for reviewing the code. Here are a couple modifications:

- Add the test for raising error for zero division.

- Add the test for raising error for zero sample.

- Use random values instead of fixed values in test_MARE.

- Change the reference name to Botchkarev 2018 and successfully build the docs.

ignite/contrib/metrics/MARE.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import division
```
vfdev-5: nit: maybe it'd be better to rename the file as `mean_absolute_relative_error` instead of `MARE`

ignite/contrib/metrics/test_MARE.py
```diff
@@ -0,0 +1,65 @@
+from ignite.contrib.metrics import MeanAbsoluteRelativeError
```
vfdev-5: Test file need to be with all other test files :)

ignite/contrib/metrics/MARE.py
weihuangxu: Sure. It is important to keep consistent. If other metrics use full name, it would be better to change it.

ignite/contrib/metrics/test_MARE.py
weihuangxu: Fixed.

ignite/contrib/metrics/MARE.py
weihuangxu: Fixed.

tests/ignite/contrib/metrics/test_mean_absolute_relative_error.py
```diff
@@ -0,0 +1,65 @@
...
+    sum_samples += num_samples_d
+    MARE_d = sum_error / sum_samples
+    assert m.compute() == approx(MARE_d.item())
+
+
+def test_zero_div():
+    a = torch.tensor([2.0, -1.0, -1.0, 2.0])
+    ground_truth = torch.tensor([0.0, 0.5, 0.2, 1.0])
+
+    m = m = MeanAbsoluteRelativeError()
```
vfdev-5:
```suggestion
    m = MeanAbsoluteRelativeError()
```

tests/ignite/contrib/metrics/test_mean_absolute_relative_error.py
```diff
@@ -0,0 +1,65 @@
...
+
+def test_mean_absolute_relative_error():
+    a = torch.rand(4)
+    b = torch.rand(4)
+    c = torch.rand(4)
+    d = torch.rand(4)
+    ground_truth = torch.rand(4)
+
+    m = MeanAbsoluteRelativeError()
+    m.reset()
```
vfdev-5: I was discussed in some of previously proposed PRs that `reset` is not needed after the constructor. We can remove.

tests/ignite/contrib/metrics/test_mean_absolute_relative_error.py
```diff
@@ -0,0 +1,65 @@
...
+    MARE_d = sum_error / sum_samples
+    assert m.compute() == approx(MARE_d.item())
+
+
+def test_zero_div():
+    a = torch.tensor([2.0, -1.0, -1.0, 2.0])
+    ground_truth = torch.tensor([0.0, 0.5, 0.2, 1.0])
+
+    m = m = MeanAbsoluteRelativeError()
+    m.reset()
```
vfdev-5: Same here. No need to reset

tests/ignite/contrib/metrics/test_mean_absolute_relative_error.py
```diff
@@ -0,0 +1,65 @@
...
+    with raises(NotComputableError):
+        m.update((a, ground_truth))
+
+
+def test_zero_sample():
+    m = MeanAbsoluteRelativeError()
+    m.reset()
```
vfdev-5: Here too, reset is useless

ignite/contrib/metrics/mean_absolute_relative_error.py
```diff
@@ -0,0 +1,39 @@
...
+
+class MeanAbsoluteRelativeError(Metric):
+    r"""
+    Calculate Mean Absolute Relative Error
+
+    :math:`\text{MARE} = \frac{1}{n}\sum _j^n\frac{\left|A_j-P_j\right|}{\left|A_j\right|}`
+
+    where, :math:`A_j` is the ground truth and :math:`P_j` is the predicted value.
+
+    More details can be found in the reference `Botchkarev 2018`_ .
```
zasdfgbnm: What is the difference between referencing as 
```rst
`text`__
__ https://link.to/your/target
```

vs 

```rst
`text`_
.. _ `text`:
    https://link.to/your/target
```
?
If we use the latter one for different metrics but with same text and url, would it create errors like in https://github.com/pytorch/ignite/pull/331#issuecomment-438177435

PR #333, opened 2018-11-14
Multilabel Metrics

### LINKED ISSUE #310: Multilabel Option for Accuracy, Precision and Recall
### Proposal
As previously discussed #275 , we should incorporate multilabel cases for precision, recall and accuracy. 

### How
Add a multilabel flag to each of the metrics, listed above. For multilabel cases, y and y_pred should be shape (batch_size, n_classes, ...) where n_classes is more than 1 (not possible with binary). 

Convert y_pred to 0's and 1's using the method below or round y_pred.

```
y_pred = y_pred.unsqueeze(1)
y_pred = torch.max(y_pred, dim=1)[1]
```

Convert y_pred and y to a shape of (-1, n_classes), to handle classification, NER, etc. 

For accuracy, use torch.eq to compare every item of y and y_pred. Count number of True. 

For precision and recall, keep the calculations the same i.e. calculating y * y_pred

Thoughts?

anmolsjoshi:
Related to #310
Description:
Adds multilabel support for metrics.

Check list:


anmolsjoshi:
@vfdev-5 I updated the structure of the code and added recall. All tests are passing!

anmolsjoshi:
cc @rpatrik96

ignite/metrics/precision.py
```diff
@@ -10,15 +11,21 @@
...
+
+    For multilabel cases `average` is True and returns the unweighted average across all samples.
     """
-    def __init__(self, average=False, output_transform=lambda x: x):
+
+    def __init__(self, average=False, threshold_function=lambda x: torch.round(x), output_transform=lambda x: x):
         super(Precision, self).__init__(output_transform)
         self._average = average
+        self._threshold = threshold_function
+        self._is_multi = False
```
vfdev-5: Do we need `_is_multi` in `self` ?

vfdev-5:
@anmolsjoshi let me test your implementation locally. 
IMO it would be better to have MultilabelPrecision as MultilabelAccuracy. This could simplify some checks between modes in the single Precision.
Let us set this requirement in a standby until I test all this and have a more clear idea

tests/ignite/metrics/test_recall.py
```diff
@@ -158,3 +156,63 @@ def test_sklearn_compute():
...
+
+def test_multilabel_example():
+    recall = Recall(average=True)
+
+    # N x C case
+    y_pred = torch.round(torch.rand(4, 4))
+    y = torch.ones(4, 4).type(torch.LongTensor)
+
+    recall.update((y_pred, y))
+    assert recall.compute() == pytest.approx(recall_score(y.data.numpy(), y_pred.data.numpy(), average='samples'))
```
vfdev-5: Do we still need to use `tensor.data.numpy()` vs `tensor.numpy()` ?

vfdev-5:
@anmolsjoshi a way I see to mix up two implementations without a bunch `if else` and keeping it clear/clean can be something like it is done in `RunningAverage` for different outputs: 
https://github.com/pytorch/ignite/blob/11928a3b02eacaed1f312c225434195adf9e5082/ignite/metrics/running_average.py#L39-L50

In the same way we can implement some methods : 
- `update_multiclass`
- `update_multilabel`
- others if needed
and setup the computation logic at the initialization.

anmolsjoshi:
@vfdev-5 thanks for your input. I like the idea of separating the metric functions into `_update_multiclass` and `_update_multilabel`. I updated the code as such and added tests. Let me know if there are any other issues.

ignite/metrics/accuracy.py
```diff
@@ -14,11 +14,48 @@ class Accuracy(Metric):
     - `y_pred` must be in the following shape (batch_size, num_categories, ...) or (batch_size, ...)
```
vfdev-5: We need to update the docs and explain what do all new arguments.
Update info on input shapes in all cases.

ignite/metrics/accuracy.py
```diff
@@ -14,11 +14,48 @@ class Accuracy(Metric):
...
     - `y` must be in the following shape (batch_size, ...)
     """
+    def __init__(self, is_multilabel=False, threshold_function=lambda x: torch.round(x), output_transform=lambda x: x):
```
```suggestion
    def __init__(self, output_transform=lambda x: x, is_multilabel=False, threshold_function=lambda x: torch.round(x)):
```

ignite/metrics/accuracy.py
```diff
@@ -14,11 +14,48 @@ class Accuracy(Metric):
...
+        super(Accuracy, self).__init__(output_transform)
+        self._threshold = threshold_function
+        if is_multilabel:
+            self.update = self._update_multilabel
```
vfdev-5: We also need to check whether self._threshold is callable in case of multi label.

ignite/metrics/accuracy.py
```diff
@@ -14,11 +14,48 @@ class Accuracy(Metric):
...
+        else:
+            self.update = self._update_multiclass
```
vfdev-5: Here we need to raise error if user specifies `self._threshold`. So maybe better to set `threshold_function=None` by default and then setup as `torch.round(x)` if needed in multi-label case.

ignite/metrics/accuracy.py
anmolsjoshi: Will incorporate.

ignite/metrics/accuracy.py
anmolsjoshi: Will incorporate. added in latest commit.

ignite/metrics/accuracy.py
anmolsjoshi: Will incorporate. added in latest commit.

anmolsjoshi:
@vfdev-5 I updated the code with your comments. 
- Updated BinaryAccuracy and CategoricalAccuracy with correct __init__ call, and updated tests. 
- Updated Precision, Recall, Accuracy with _update_multilabel, _update_mutliclass, threshold_function check, average check. Updated tests also.
- Updated Precision and Recall multi class update to have binary to categorical mapping and not use torch.round.

Thanks for your review

anmolsjoshi:
@vfdev-5 just looked, it seems that there is an issue with the docstrings according to travis. will update that.

ignite/metrics/recall.py
```diff
@@ -10,34 +11,116 @@
...
+            # Converts y and y_pred to (-1, num_classes) from N x C x L
+            y_pred = y_pred.transpose(2, 1).contiguous().view(-1, y_pred.size(1))
+            y = y.transpose(2, 1).contiguous().view(-1, y_pred.size(1))
+
+        elif y_pred.ndimension() == 4:
+            # Converts y and y_pred to (-1, num_classes) from N x C x H x W
+            y_pred = y_pred.permute(0, 2, 3, 1).contiguous().view(-1, y_pred.size(1))
+            y = y.permute(0, 2, 3, 1).contiguous().view(-1, y_pred.size(1))
+
+        else:
```
vfdev-5: Can we remove this ?

ignite/metrics/precision.py
```diff
@@ -10,33 +11,116 @@
...
+            # Converts y and y_pred to (-1, num_classes) from N x C x L
+            y_pred = y_pred.transpose(2, 1).contiguous().view(-1, y_pred.size(1))
+            y = y.transpose(2, 1).contiguous().view(-1, y_pred.size(1))
+
+        elif y_pred.ndimension() == 4:
+            # Converts y and y_pred to (-1, num_classes) from N x C x H x W
+            y_pred = y_pred.permute(0, 2, 3, 1).contiguous().view(-1, y_pred.size(1))
+            y = y.permute(0, 2, 3, 1).contiguous().view(-1, y_pred.size(1))
+
+        else:
```
vfdev-5: Can we remove this too ?

vfdev-5:
@anmolsjoshi looks better, thanks !
Concerning the following code 
```python
        if y_pred.ndimension() == 3:
            # Converts y and y_pred to (-1, num_classes) from N x C x L
            y_pred = y_pred.transpose(2, 1).contiguous().view(-1, y_pred.size(1))
            y = y.transpose(2, 1).contiguous().view(-1, y_pred.size(1))
        elif y_pred.ndimension() == 4:
            # Converts y and y_pred to (-1, num_classes) from N x C x H x W
            y_pred = y_pred.permute(0, 2, 3, 1).contiguous().view(-1, y_pred.size(1))
            y = y.permute(0, 2, 3, 1).contiguous().view(-1, y_pred.size(1))
```
Can we generalize it to any shape of `(N, C, ...)` as it is decribed in the docs ?
For example, we can permute like this `N, C, ...` => `C, N, ...` => `C, -1` => `-1, C` ?

vfdev-5:
In tests you have a lot of `y.data.numpy()` which can be directly done `y.numpy()`

ignite/metrics/precision.py
anmolsjoshi: Incorporated. Generalized as suggested below.

ignite/metrics/recall.py
anmolsjoshi: Incorporated. Generalized as suggested below.

anmolsjoshi:
@vfdev-5 thanks for the review. 

I've made the following changes:
- Generalized conversion of y and y_pred to (-1, C) in Accuracy, Precision, Recall.
- Updated each of the listed metrics test with this generalization. 
- Updated each of the listed metrics by updating the numpy conversion.

anmolsjoshi:
@vfdev-5 I added additional testing after reading the codecov report. There were missing test cases for multilabel cases. Here are the tests I added for Precision and Recall - 

- All examples wrong
- Batch Updates
- Incorrect Shape
- test for incorrect y (1's and 0's) was wrong, I have now corrected it

vfdev-5:
@anmolsjoshi thanks for the update! Let me check this on my side. Hope not to take a lot of time for this

vfdev-5:
> All examples wrong

@anmolsjoshi could you please detail what do you mean by that ?

anmolsjoshi:
Every item of y and y_pred are wrong or do no match.  [Here](https://github.com/pytorch/ignite/pull/333/files#diff-ae3f7f6de363a79b30d3d20e5f9dd649R232) y is ones and y_pred is zeros. So precision and recall should be 0 for this. 

The test was included to run this [line of code](https://github.com/pytorch/ignite/pull/333/files#diff-b5da37c1a57f328606cb00cf5ef683c6R89). 

the tests existed in the original version of Precision and Recall, before we combined binary and multiclass cases. I just wanted to make sure we had coverage for _update_multilabel.

vfdev-5:
Correct me if I'm wrong but I think there is a problem with the current `Precision` implementation in multilclass/binary case:
```python

y_pred = torch.Tensor([[0.4], [0.6], [0.7], [0.8]])
y = torch.Tensor([0, 0, 0, 0]).type(torch.LongTensor)
# or 
# y = torch.Tensor([0, 0, 1, 1]).type(torch.LongTensor)

precision.update((y_pred, y))

np_y = y.numpy()
np_y_pred = (y_pred.numpy().ravel() > 0.5).astype('int')

precision_score(np_y, np_y_pred), precision.compute()
```

vfdev-5:
I checked the implementation in the master and there is the same problem.

anmolsjoshi:
@vfdev-5 thanks for your review. 

Please see this [comment](https://github.com/pytorch/ignite/issues/348#issuecomment-442926728)

I think if a user is using Precision and Recall, they should not be providing logits to the metric. We can make that clear in the documentation to use output_transform to convert the logits to probabilities.

Or we can include an activation parameter to the metric. If activation is None, the metric will assume the user is using probabilities, else logits. activation will be a callable function that is applied to y_pred.

I personally prefer the first option, what do you think?

anmolsjoshi:
@vfdev-5 there is an error arising from cases where the probability is 0.5

In pytorch-cpu
```python
y = torch.Tensor([[0.5]])
print (torch.max(torch.cat([1.0 - y, y], dim=1), dim=1)[1]
# 0

print (torch.round(y))
# tensor([0.])
```

In pytorch-nightly-cpu
```python
y = torch.Tensor([[0.5]])
print (torch.max(torch.cat([1.0 - y, y], dim=1), dim=1)[1]
# 1

print (torch.round(y))
# tensor([0.])
```

I think this might be more of a reason to use torch.round for binary cases. 

Thoughts?

vfdev-5:
@anmolsjoshi thank you a lot for taking this ! 

> Or we can include an activation parameter to the metric. If activation is None, the metric will assume the user is using probabilities, else logits. activation will be a callable function that is applied to y_pred.

Set activation IMO is not a good idea.

> I think if a user is using Precision and Recall, they should not be providing logits to the metric. We can make that clear in the documentation to use output_transform to convert the logits to probabilities.

I agree, for instance we should make it clear in documentation.

vfdev-5:
About your latest [comment](https://github.com/pytorch/ignite/pull/333#issuecomment-442970894), let me check

anmolsjoshi:
@vfdev-5 I was thinking maybe integrating this as proposed [here](https://github.com/pytorch/ignite/issues/348#issuecomment-442724298) might be a good idea. It requires refinement for sure. 

My idea is to basically check if `update` has been called, if it hasn't been then check and store the type of problem. After that continue checking in update if type changes, then raise an error.

I'll have a more robust example pushed later tonight.

vfdev-5:
@anmolsjoshi actually, I prefer to separate problems: a) bug with binary case, b) multilabel
When you proposed this new implementation, I remarked that in multilabel case, accumulation goes with `torch.cat` and this seems a bit strange. Maybe this is the only way to do this, need to check. But at first I would like to fix the bug and improve tests, before merging multilabel.

anmolsjoshi:
@vfdev-5 of course. I'll send in a separate PR.

anmolsjoshi:
@vfdev-5 You are correct, we need is_multilabel flag. There is no method to differentiate between binary and multilabel cases in some situations. For example, if y has shape (1, 2, 2) and y_pred has shape (1, 2, 2) - you could think of this as batch_size=1 and a 2x2 image or batch_size =1, num_classes=2, and two predictions for that batch for example two words.

Technically, sklearn does have a is_multilabel flag i.e. average=samples. sklearn also has the advantage of being limited to shapes of 1 or 2 dimensions, it does not handle exotic dimension types like sentences, images, etc.

anmolsjoshi:
To help unify accuracy calculations - 
correct  = torch.cat([torch.Tensor([1]).type_as(y_pred).byte() if torch.equal(true, pred) else torch.Tensor([0]).type_as(y_pred).byte() for true, pred in zip(y, y_pred)])

`type_as` is needed to keep the same device. 
`.byte()` is used to keep the calculation of self._num_correct the same.

anmolsjoshi:
@vfdev-5 it seems that we can use binary to categorical mapping for multilabel cases so we don't have to use threshold_function.  See below.

As you had mentioned, we can include a note in the docstring that if a specific threshold is needed, users should use output_transform.

```python
import torch
y_pred = torch.rand(10, 4)
y = torch.randint_like(y_pred, 0, 2).type(torch.LongTensor)
y_pred_ = y_pred.unsqueeze(dim=1)
y_pred_ = torch.cat([1.0 - y_pred_, y_pred_], dim=1)
y_pred_ = torch.max(y_pred_, dim=1)[1]

print (torch.equal(torch.round(y_pred), y_pred_.type_as(y_pred)))
# True
```

ignite/metrics/binary_accuracy.py
```diff
@@ -9,6 +9,7 @@ class BinaryAccuracy(Accuracy):
     Note: This metric is deprecated in favor of Accuracy.
     """
     def __init__(self, *args, **kwargs):
+        super(BinaryAccuracy, self).__init__(*args, **kwargs)
         warnings.warn("The use of ignite.metrics.BinaryAccuracy is deprecated, it will be "
                       "removed in 0.2.0. Please use ignite.metrics.Accuracy instead.", DeprecationWarning)
         super(Accuracy, self).__init__(*args, **kwargs)
```
vfdev-5: There are two calls of `super`, this one to remove ?

ignite/metrics/categorical_accuracy.py
```diff
@@ -9,6 +9,7 @@ class CategoricalAccuracy(Accuracy):
     Note: This metric is deprecated in favor of Accuracy.
     """
     def __init__(self, *args, **kwargs):
+        super(CategoricalAccuracy, self).__init__(*args, **kwargs)
         warnings.warn("The use of ignite.metrics.CategoricalAccuracy is deprecated, it will be "
                       "removed in 0.2.0. Please use ignite.metrics.Accuracy instead.", DeprecationWarning)
         super(Accuracy, self).__init__(*args, **kwargs)
```
vfdev-5: Same here

vfdev-5:
Travis CI says https://travis-ci.org/pytorch/ignite/jobs/467876689#L1147 
Strange error, I will restart one of jobs to see if we can reproduce it.

ignite/metrics/binary_accuracy.py
anmolsjoshi: Done!

ignite/metrics/categorical_accuracy.py
anmolsjoshi: Done!

anmolsjoshi:
@vfdev-5 I'm unsure of what's happening, google has not really been helpful about this error haha

anmolsjoshi:
@vfdev-5 Figured it out! Working on a fix, will push something later tonight.

anmolsjoshi:
@vfdev-5 problem solved! tests are passing, I also added tests in precision, recall and accuracy to check for batched updates - i think it's fairly useful to have and makes the testing more robust.

vfdev-5:
> Figured it out! Working on a fix, will push something later tonight.

@anmolsjoshi what was the problem actually ?

anmolsjoshi:
@vfdev-5 for binary and multilabel cases we are using randint, since the expected input for y and y_pred are 0's and 1's. The default tensor type for randint is torch.LongTensor. See code block below.

```python
import torch

y = torch.randint(0, 2, size=(4, ))
print (y)
# tensor([1, 0, 0, 0])

print (y.type())
# torch.LongTensor
```

I ran some code below for binary test, the error is thrown off at compute() for Precision and Recall. 

```python
import warnings
from sklearn.metrics import precision_score
import pytest
from sklearn.exceptions import UndefinedMetricWarning

average = True
pr = Precision(average=average)
y_pred = torch.randint(0, 2, size=(10, 1))
y = torch.randint(0, 2, size=(10,)).type(torch.LongTensor)
pr.update((y_pred, y))
np_y = y.numpy().ravel()
np_y_pred = y_pred.numpy().ravel()
assert pr._type == 'binary'
assert isinstance(pr.compute(), float if average else torch.Tensor)
pr_compute = pr.compute() if average else pr.compute().numpy()
assert precision_score(np_y, np_y_pred, average='binary') == pytest.approx(pr_compute)

# <ipython-input-10-e863c9fab78d> in compute(self)
#      56 
#      57         if self._average:
# ---> 58             return result.mean().item()
#      59         else:
#      60             return result

# RuntimeError: Can only calculate the mean of floating types. Got Long instead.
```

Another error is that division of torch.LongTensor throws an error when the denominator is 0. See code below.

```python
import torch
a = torch.LongTensor([1])
b = torch.LongTensor([0])

a/b
# Throws a kernel died error in Jupyter or floating point exception in python terminal
```

So the fix is to convert _positives and _true_positives to float before division. It would solve both problems listed above.

vfdev-5:
@anmolsjoshi I made some edits in multilabel accuracy implementation and updated tests, please take a look, what do you think ?

vfdev-5:
> As for the testing of batches, I deleted code in45ae0f5, do you think we should add it to all tests?

This can be more exaustive 

> I was getting these assertion errors on my local machine too. These errors were also on pytorch-nightly. What are your thoughts on adding a tolerance when using pytest.approx(answer, tolerance)?

@anmolsjoshi let me check this.

vfdev-5:
@anmolsjoshi I made some code simplification for precision/recall. Any comments ?
The split on reset/collect functions you made in `_BasePrecisionRecall` was OK but it would be better not to store the whole history. This step has been replaced by less evident precomputing step in `update`:
```python
        if self._type == "multilabel":
            true_positives = true_positives.type(torch.float) / (all_positives.type(torch.float) + self.eps)
            all_positives = torch.ones_like(true_positives)
```
and thus in `compute` we have an artificial:
```python
        result = self._true_positives.type(torch.float) / (self._positives.type(torch.float) + self.eps)
```

anmolsjoshi:
@vfdev-5 I added batched updated to all the accuracy tests. It seems that precision and recall tests are still failing due to precision of the numbers see [here](https://travis-ci.org/pytorch/ignite/jobs/471446467#L1242) and [here](https://travis-ci.org/pytorch/ignite/jobs/471446466#L1266). 

I'm not sure how your commits passed earlier. I haven't made any changes to the calculation of precision and recall in this last commit.

vfdev-5:
@anmolsjoshi I'll take a look

vfdev-5:
Current implementation of precision/recall for multilabel case is wrong:
- when batches not of the same size
```python
pr = Precision(average=True, is_multilabel=True)

y_pred = torch.randint(0, 2, size=(100, 4))
y = torch.randint(0, 2, size=(100, 4)).type(torch.LongTensor)

batch_size = 16

for i in range(n_iters):
    idx = i * batch_size
    pr.update((y_pred[idx:idx + batch_size], y[idx:idx + batch_size]))

np_y = y.numpy()
np_y_pred = y_pred.numpy()


pr._type == 'multilabel', isinstance(pr.compute(), float), precision_score(np_y, np_y_pred, average='samples'), pr.compute()
```

anmolsjoshi:
@vfdev-5 

>  I think we need also to get back data to cpu ...

Is this to prevent wasting GPU memory on storing metrics? If so, it might be good to include that in that in other metrics as well or somehow include it somehow in the Metric class. 

> I have still some problems locally with number accuracy... Need to investigate.

Yea, I'm confused on why tests are passing online but not locally. facing the same error on my machine.

vfdev-5:
@anmolsjoshi yes, to prevent wasting GPU. However, IMO only precision and recall can return `torch.Tensor` (to verify) and thus all other metrics are "save".

I found the problem why my tests are failling locally. It is related to random_state and random input. 
For example, I have precision tests are failling when executed only this test. In case of multiclass data
I have a situation when `y` and `y_pred` both do not contain class `4` and `precision_score` outputs a vector of size smaller that our computation as it's never encountered class `4` but in our implementation as we apply argmax we know the total number of classes... 
Need to fix the assert part

vfdev-5:
@anmolsjoshi I agree

anmolsjoshi:
tests passing locally. will push results in the next few minutes

vfdev-5:
@alykhantejani or @jasonkriss please review and merge if OK

vfdev-5:
@anmolsjoshi there is a more compact way to code this to store the whole history...
I implemented this for my project, but I'm still hesitating whether we should not just provide `F-beta` metric...

anmolsjoshi:
@vfdev-5 I just pushed a change regarding your concerns. 

Tested MetricsLambda locally. There might be an issue. Please see below.

```python
from ignite.metrics import MetricsLambda

y_pred = torch.randint(0, 2, size=(15, 10, 4))
y = torch.randint(0, 2, size=(15, 10, 4)).long()

def update_fn(engine, batch):
    y_pred, y = batch
    return y_pred, y

evaluator = Engine(update_fn)

precision = Precision(average=False, is_multilabel=True)
recall = Recall(average=False, is_multilabel=True)

F1 = precision * recall * 2 / (precision + recall)

precision.attach(evaluator, "precision")
recall.attach(evaluator, "recall")
F1.attach(evaluator, "f1")

def data(y_pred, y):
    for i in range(y_pred.shape[0]):
        yield (y_pred[i], y[i])
        
d = data(y_pred, y)

state = evaluator.run(d, max_epochs=1)

print (state.metrics.keys())
# ['f1', 'f1[0][0][0][1]', 'f1[0][0]', 'f1[0]', 'recall', 'f1[0][1][1]', 'f1[0][0][0]', 'precision', 'f1[0][1]', 'f1[0][1][0]', 'f1[0][0][0][0]']

print (state.metrics['f1'])
# nan
```

Not sure what all those f1 keys are in. 

The reason ['f1'] is nan because there are precision and recall that contain 0's. So I think we should update the docs to include a function similar to the one in tests/metrics/test_metric_lambda.py.

See below:

```python
def Fbeta(r, p, beta):
        return torch.mean((1 + beta ** 2) * p * r / (beta ** 2 * p + r + 1e-12)).item()
```

anmolsjoshi:
@vfdev-5 

> @anmolsjoshi there is a more compact way to code this to store the whole history...

Can you share that method? I'd be interested in seeing that implementation.

vfdev-5:
@anmolsjoshi 

concerning the strange keys in `metrics`, just posted the issue #381  :)

Yes, you are correct, the `NaN` is due to precision and recall are both zero which can be overcomed with `1e-10` in denominator.

Concerning the implementation, actually nothing special. Here is [gist](https://gist.github.com/vfdev-5/7559dcc1de9ef868531ee5612a652d41)

vfdev-5:
> concerning f-beta, would that be part of this PR?

@anmolsjoshi probably not. I still have a doubt whether we should open an issue for that...

anmolsjoshi:
@vfdev-5 sounds good. I updated the metrics with your simplified code. tests are passing. 

Let me know if any other changes are needed at this time.

ignite/metrics/precision.py
```diff
@@ -9,21 +10,26 @@
 
 class _BasePrecisionRecall(_BaseClassification):
 
-    def __init__(self, output_transform=lambda x: x, average=False):
+    def __init__(self, output_transform=lambda x: x, average=False, is_multilabel=False):
         self._average = average
-        super(_BasePrecisionRecall, self).__init__(output_transform=output_transform)
+        if is_multilabel:
+            if not self._average:
+                warnings.warn("Average argument should be True for multilabel case. "
```
vfdev-5: Not sure about this warning. Maybe docstring is enough ?

ignite/metrics/precision.py
```diff
@@ -38,38 +44,73 @@ class Precision(_BasePrecisionRecall):
...
+    .. warning::
+
+        If average is False, current implementation stores all input data (output and target) in as tensors before
+        computing a metric. This can potentially lead to a memory error if the input data is larger than available RAM.
+
+    .. code-block:: python
+
+        precision = Precision(average=False, is_multilabel=True)
+        recall = Recall(average=False, is_multilabel=True)
+        F1 = precision * recall * 2 / (precision + recall)
```
vfdev-5:
```suggestion
         F1 = precision * recall * 2 / (precision + recall + 1e-20)
```

ignite/metrics/recall.py
```diff
@@ -11,39 +14,74 @@ class Recall(_BasePrecisionRecall):
...
+    .. warning::
+
+        If average is False, current implementation stores all input data (output and target) in as tensors before
+        computing a metric. This can potentially lead to a memory error if the input data is larger than available RAM.
 
+    .. code-block:: python
+
+        precision = Precision(average=False, is_multilabel=True)
+        recall = Recall(average=False, is_multilabel=True)
+        F1 = precision * recall * 2 / (precision + recall)
```
vfdev-5:
```suggestion
         F1 = precision * recall * 2 / (precision + recall + 1e-20)
```

anmolsjoshi: Incorporated.

ignite/metrics/precision.py
anmolsjoshi: Incorporated.

anmolsjoshi:
@vfdev-5 all changes have been incorporated.

jasonkriss:
Looks great. @anmolsjoshi @vfdev-5 great work! This was a big one.

Issue #337, opened 2018-11-16
Metaclass for Metric is not working on python3


zasdfgbnm:
```python
import ignite
m = ignite.metrics.Metric()
```

The above code should report an error, but on python3, it doesn't.

This is because the following way of setting metaclass is python2 only:

```python
class Metric(object):
    __metaclass__ = ABCMeta
```

PR #338, opened 2018-11-20
Fix metaclass of metric for python3

zasdfgbnm:
Fixes https://github.com/pytorch/ignite/issues/337

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -18,13 +18,11 @@ requirements:
     - setuptools
     - pytorch
     - enum34 # [py < 34]
-    - six
 
   run:
     - python
     - pytorch
     - enum34 # [py < 34]
-    - six
 
 test:
   imports:
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -18,11 +18,13 @@ requirements:
     - setuptools
     - pytorch
     - enum34 # [py < 34]
+    - six
 
   run:
     - python
     - pytorch
     - enum34 # [py < 34]
+    - six
 
 test:
   imports:
--- a/ignite/_six.py
+++ b/ignite/_six.py
@@ -1,0 +1,126 @@
+# Copyright (c) 2010-2017 Benjamin Peterson
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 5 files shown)


zasdfgbnm:
@vfdev-5 Do you think it worth a new test case for this?

vfdev-5:
@zasdfgbnm yes, please add a test what would fail without this fix with assert raise when an abstract class is instanciated.

zasdfgbnm:
@vfdev-5 test added

vfdev-5:
@zasdfgbnm please update [meta.yaml](https://github.com/pytorch/ignite/blob/3f57ee52d233b39df926c0aef4d3c23730f2a821/conda.recipe/meta.yaml) too

vfdev-5:
@alykhantejani, @jasonkriss we add `six` as new dependency, any objections ?

zasdfgbnm:
@vfdev-5 done

jasonkriss:
Can we maybe utilize https://github.com/pytorch/pytorch/blob/master/torch/_six.py#L54 rather than adding `six` as a dependency?

zasdfgbnm:
@jasonkriss Do you mean something like:

```
from torch._six import with_metaclass
```

Or like pytorch did, just copy that file to ignite? `_six` is not in pytorch public API and might be removed in the future.

alykhantejani:
Let's not rely on pytorch's import as its a private module `_six`. However, we can follow there approach and have an `ignite._six` and just copy the relevant functions we need (and the copyright notice)?

alykhantejani:
we can, but seems overkill (also keeps ignite lightweight). We could just cp pytorch/_six.py as it contains what we need. I'm open to including other dependencies in general, but I don't think we need/use much of six, and pytorch has already copied this function across

zasdfgbnm:
I also prefer not copying file, but use dependency instead. But I'm OK with any decision.

jasonkriss:
If we start to need a lot of functionality from `six`, then I think it's fine to add it as a dependency. That being said, in this case, it's 5 lines of code, so I think it'd preferable to add an `ignite._six` with just the `with_metaclass` implementation.

vfdev-5:
Okay, let's merge it like this

alykhantejani:
Thanks @zasdfgbnm!

PR #341, opened 2018-11-21
Fixes #340 and added tests against pytorch-nightly

### LINKED ISSUE #340: Terminate on Nan tests fail with pytorch-nightly
https://travis-ci.org/vfdev-5/ignite/builds/458179313

vfdev-5:
Fixes #340

Description:

1) In 0.4.1 `torch.isfinite(1.0 / torch.arange(-1, 2))` raises `RuntimeException` and in 1.0-preview it is finite. Terminate on nan catches `RuntimeException` and thus tests passed for 0.4.1 and fail with 1.0-preview. This PR properly transforms int to float when 1.0 / int_value

2) Added tests for python 3.6 and {pytorch-cpu, pytorch-nightly-cpu}

Check list:
* [x] New tests are added

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ language: python
 python:
   - "2.7"
   - "3.5"
+  - "3.6"
+
+env:
+  - PYTORCH_PACKAGE=pytorch-cpu
+  - PYTORCH_PACKAGE=pytorch-nightly-cpu
 
 stages:
   - Lint check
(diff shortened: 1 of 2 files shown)


vfdev-5:
@jasonkriss , @alykhantejani what do you think about if we extend testing grid to 
- py27, py35, py36
-  pytorch 0.4.1 , 1.0-preview

I would say this extends the testing by ~ 8 mins as in Tests travis allows to run on 5 machines in parallel and we have 6 cases.

alykhantejani:
hmm, running on 1.0-preview is probably good. 8 minutes seems crazy long though. Do we need to test py35 and py36?

vfdev-5:
I agree that this testing grid becomes too long. I thought that we need to be more modern than ubuntu 16.04 and its built-in python 3.5. But okay, we can choose one of python3 
Let me vote for py36 :)

.travis.yml
```diff
@@ -3,6 +3,11 @@ language: python
 python:
   - "2.7"
   - "3.5"
+  - "3.6"
```
vfdev-5:
```suggestion

```

.travis.yml
```diff
@@ -4,6 +4,10 @@ python:
   - "2.7"
   - "3.5"
```
vfdev-5:
```suggestion
  - "3.6"
```

PR #343, opened 2018-11-25
[1/2] Added nightly build stage into .travis

### LINKED ISSUE #306: Master branch nightly build depending on torch_nightly
We need a build of ignite to allow people try nightly version of PyTorch together with ignite. Since the PyTorch is nightly built, it might be better to call this package `pytorch-ignite-nightly` which is also a nightly build of master branch, but with the dependencies on `torch` changed to `torch_nightly`.

cc: @vfdev-5 

NB: I'm not sure the dependency should be `torch_nightly` or `torch-nightly`, because the install command has `torch_nightly` but the output of pip says `torch-nightly`:
```
gaoxiang@pc ~ $ sudo pip install torch_nightly -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
Looking in links: https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
Collecting torch_nightly
  Downloading https://download.pytorch.org/whl/nightly/cpu/torch_nightly-1.0.0.dev20181029-cp37-cp37m-linux_x86_64.whl (85.9MB)
    100% |████████████████████████████████| 85.9MB 698kB/s 
Installing collected packages: torch-nightly
Successfully installed torch-nightly-1.0.0.dev20181029
```
@soumith Do you think we should change these two names to make them consistent?

vfdev-5:
Related to #306 

The second part of the PR : https://github.com/pytorch/ignite/pull/344 

Description:
Idea is to build ignite master code every day. There is a possibility to build & upload some code with [travis cron jobs](https://docs.travis-ci.com/user/cron-jobs/). 
We need to upload the builds at another package storage: pytorch-ignite-nightly for PyPI and ignite-nightly for Conda.

Basic idea is to create a new branch `nightly` with all needed modifications in package name and dependencies. But the problem is how to synchronize the branch with master. So the idea is to create a special folder `nightly` that exists only in `nightly` branch, all other files are same as in master. At each cron build job, we can do the following:
```
        # Merge master into nightly
        - git fetch origin master
        - git merge FETCH_HEAD --no-edit
        # Copy "nightly" files
        - cp nightly/setup_nightly.py setup.py
        - cp -R nightly/conda.recipe .
        # Push to nightly branch
        - git remote rm origin && git remote add origin https://$GITHUB_TOKEN@github.com/${TRAVIS_REPO_SLUG}.git
        - git push origin nightly
```
We merge master branch into nightly and then push to the nightly. As all `nightly` modifications are separated into a folder `nightly`, most probably we'll never have a conflict to solve (exception if someone put `nightly` folder into master etc).

In this PR: 
Updated .travis.yml with new stage `Nightly` that merges master branch, replaces existing setup.py and meta.yml: version is YYYYMMDD and pytorch -> pytorch-nightly ; and finally git push to nightly branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ stages:
   - Test
   - Docs
   - Deploy
+  - Nightly
 
 before_install: &before_install
   - sudo apt-get update
@@ -83,7 +84,7 @@ jobs:
     # GitHub Pages Deployment: https://docs.travis-ci.com/user/deployment/pages/
     - stage: Docs
       python: "3.5"
-
+      if: branch != nightly
       # Use previously defined before_install
       before_install: *before_install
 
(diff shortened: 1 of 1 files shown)


alykhantejani:
why do we need a nightly folder, it's unclear to me. And what would be the difference between nightly/setup.py and master setup.py?

alykhantejani:
Ah I see, we want to depend on torch-nightly

alykhantejani:
(could probably do this with some grep sed magic as part of the travis job?)

vfdev-5:
@alykhantejani there are package name, dependencies and versions to change for pipy and conda. 

> could probably do this with some grep sed magic as part of the travis job?

I agree. I thought it would be more simple to copy files which content is directly visible. No pain to debug if there is a change to integrate.

alykhantejani:
cool - LGTM then

jasonkriss:
I may be missing something, but would it be possible to instead just have ignite master depend on pytorch-nightly and then build ignite-nightly directly from master? Then, when we actually want to cut a versioned release of ignite, we can update the pytorch dependency. 

If that doesn't work, then I'm fine with this PR.

vfdev-5:
@jasonkriss interesting idea! So with such approach when we want to create a tag from master, we need to bump the version and setup the good dependency... Probably, this can be error prone.

jasonkriss:
@vfdev-5  By error-prone, do you mean it could be easy to forget to switch the dependency? If so, I agree, that's a good point.

I'm fine with the approach in this PR. My only reservation is that it feels quite complex just to get a nightly build. Have you seen other projects do this? Does torchvision do something like this? I'm just wondering if there is maybe a simpler approach that we're missing at the moment.

vfdev-5:
> By error-prone, do you mean it could be easy to forget to switch the dependency?

yes, exactly.

Concerning torchvision-nightly, it is unofficial build by fastai: https://anaconda.org/fastai/torchvision-nightly/files , there was even a discussion at slack about it. If you take a look at conda package, this nightly was build 1 month and 24 days ago. Not so nightly. Honestly, I do not know how they build it, should be a separate repository, I suppose.

> My only reservation is that it feels quite complex just to get a nightly build. 

A particularity with this "nightly" is that we also change the dependencies and we would like to upload as another package with its own version.

jasonkriss:
@vfdev-5 sounds good! I'm good with this.

vfdev-5:
Okay, I'll merge this and we'll change it if a better way is found

PR #346, opened 2018-11-27
Update checkpoint.py

willfrey:
Change `ignite.handlers.ModelCheckpoint` to call `os.path.expanduser()` on the `dirname` argument before it is assigned as an attribute.

@apaszke said this was small enough that it did not warrant a new issue.

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -1,5 +1,4 @@
 import os
-import os.path
 import tempfile
 
 import torch
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -1,4 +1,5 @@
 import os
+import os.path
 import tempfile
 
 import torch
@@ -83,7 +84,7 @@ class ModelCheckpoint(object):
                  create_dir=True,
                  save_as_state_dict=False):
 
-        self._dirname = dirname
+        self._dirname = os.path.expanduser(dirname)
         self._fname_prefix = filename_prefix
         self._n_saved = n_saved
         self._save_interval = save_interval


vfdev-5:
@willfrey thanks for the PR ! 
However if user would like to apply `os.path.expanduser()` on a path, he/she can do it directly before assigning dirname to `ModelCheckpoint` or am I missing something ?

willfrey:
@vfdev-5 That's totally reasonable. The default behavior in `torchvision.datasets` appears to be to expand `~` but the default behavior of `torch.load` appears to not expand `~`.

I totally get it if you want to avoid mutating `dirname`! If that's the case, just scrap the PR.

willfrey:
I don't think you need the `if` statement guard. I believe the default behavior of `os.path.expanduser` is to only expand a leading `~`.

For example:

```python
>>> import os.path
>>> os.path.expanduser("~")
'/home/willfrey'
>>> os.path.expanduser("/~")
'/~'
```

vfdev-5:
Yes, you are right, it does not change other path if no `~` and its execution if no tilde is almost equivalent to run `if tilde not in path: return path`. The implementation in 3.7 looks like this:
```python
def expanduser(path):
    path = os.fspath(path)
    if isinstance(path, bytes):
        tilde = b'~'
    else:
        tilde = '~'
    if not path.startswith(tilde):
        return path
    sep = _get_sep(path)
    ....
```

jasonkriss:
I think this is totally reasonable. I'd say let's try to be as consistent with the pytorch ecosystem as we can here. Seems like torchvision and at least some parts of pytorch itself expand, so I'd say lets merge this.

ignite/handlers/checkpoint.py
```diff
@@ -1,4 +1,5 @@
 import os
+import os.path
```
vfdev-5: Maybe we can skip this import

vfdev-5:
@willfrey thanks !

PR #347, opened 2018-11-28
Added D2 metrics requiring entire history of training

### LINKED ISSUE #307: More metrics for regression tasks and forecasting

anmolsjoshi:
Related to #307 

Description:
* Median Absolute Error
* Median Relative Absolute Error
* Median Absolute Percentage Error
* Geometric Mean Relative Absolute Error

Check list:


anmolsjoshi:
cc @zasdfgbnm

anmolsjoshi:
cc @vfdev-5 

I have completed implementations of the metrics that require the history of training, all the tests, and I added tests for _base.py for code coverage.

Let me know if you have any comments.

ignite/contrib/metrics/regression/geometric_mean_relative_absolute_error.py
```diff
@@ -0,0 +1,31 @@
...
+    n = y_pred.shape[0]
+    e = torch.log(torch.abs(y.view_as(y_pred) - y_pred) / torch.abs(y.view_as(y_pred) - torch.mean(y)))
+    return torch.exp(torch.mean(e)).item()
+
+
+class GeometricMeanRelativeAbsoluteError(_BaseRegressionEpoch):
+    r"""
+    Calculates the Geometric Mean Relative Absolute Error:
+
+    :math:`\text{GMRAE} = TODO`,
```
vfdev-5: We need to replace TODO

vfdev-5:
@anmolsjoshi thank you for working on this PR !
Could you please search how `np.median` is implemented and can we insprire of this in our code ?

anmolsjoshi:
@vfdev-5 I looked into [numpy median](https://github.com/numpy/numpy/blob/v1.13.0/numpy/lib/function_base.py#L4113-L4117) implementation. 

NumPy implementation is closer to what is the known definition of median, for odd number of elements - it's the middle element of the sorted array, for even number of elements - its the average of the two middle elements of the sorted array. 

PyTorch does not intend to change there implementation, see [here](https://github.com/pytorch/pytorch/issues/1837#issuecomment-314992655).

I ran some code locally, here are the results.

```python
import numpy as np
import torch

a = np.random.rand(4)
a.sort()
print (a)
# [0.15983365 0.20260213 0.40420191 0.43118176]
print (np.median(a))
# 0.3034020154974049
print (torch.from_numpy(a).median())
# tensor(0.2026, dtype=torch.float64)

a = np.random.rand(5)
a.sort()
print (a)
# [0.14675148 0.18874507 0.64389564 0.72931892 0.75430595]
print (np.median(a))
# 0.6438956443116628
print (torch.from_numpy(a).median())
# tensor(0.6439, dtype=torch.float64)
```

Instead of calculating torch.median, we can use the code below:

```python
def median(tensor):
    tensor = tensor.sort()[0]
    length = tensor.shape[0]

    if length % 2 == 0:
        szh = length // 2
        kth = [szh - 1, szh]
    else:
        kth = [(length - 1) // 2]
    return tensor[kth].mean()

a = np.random.rand(4)
print (median(torch.from_numpy(a)).item())
# 0.6643668870940789
print (np.median(a))
# 0.6643668870940789

a = np.random.rand(5)

print (median(torch.from_numpy(a)).item())
# 0.42556139026154693
print (np.median(a))
# 0.42556139026154693
```

If we want to be more in line with PyTorch, we can choose to not incorporate this, but if we want to be closer to the definition of median it might be useful to implement the above code. Let me know what you think.

vfdev-5:
@anmolsjoshi thanks for digging that ! So both actually works (evidently on the whole data history). Maybe we would like also to know whether there exists a running median implementation ...
However if we should store the complete historym yes we could use such implementation you provided.

anmolsjoshi:
@vfdev-5 thanks! Looking into a 'running median' implementation, I only methods that retained the history of the entire stream of data. let me know if you find something different! I'll wait on what you decide. 

As for GMRAE, we can try a running average for `y`, this will prevent us from keeping the entire history of the training data.

anmolsjoshi:
cc @vfdev-5

vfdev-5:
> As for GMRAE, we can try a running average for y, this will prevent us from keeping the entire history of the training data.

yes, I agree, we can deal with it as for an average mean.

vfdev-5:
@anmolsjoshi please review the updates.

anmolsjoshi:
> It remains to reimplement GMRAE.

@vfdev-5 does that mean this PR is complete after GMRAE is converted to running mean method?

vfdev-5:
@anmolsjoshi I would say yes, but we need to review.

anmolsjoshi:
@vfdev-5 I have updated the GMRAE implementation and tests.

The metric now keeps a running total of y and number of examples, thus for each batch a mean is calculated and a GMRAE is calculated and summed to a running total. At compute, the total GMRAE is divided by the number of examples, and it's exponential is calculated.

please review

anmolsjoshi:
@vfdev-5 changes to docs and tests LGTM!

vfdev-5:
@jasonkriss please review and merge if OK

PR #349, opened 2018-11-29
Bug Related to Calculation of Binary Metrics

### LINKED ISSUE #348: Bug in binary precision
Basic tests of binary precision seem to fail:
```python

y_pred = torch.rand(10, 1)
y = torch.randint(0, 2, size=(10,)).type(torch.LongTensor)

precision.update((y_pred, y))

np_y = y.numpy().ravel()
np_y_pred = (y_pred.numpy().ravel() > 0.5).astype('int')

precision_score(np_y, np_y_pred), precision.compute()
```

@jasonkriss could you please confirm this ?

EDIT: Another failing test: https://github.com/pytorch/ignite/pull/333#issuecomment-442643530

Probably, the error is at binary to categorical mapping and counting class-0 similarly to class-1. But in binary case we should ignore class-0.

There is also input "binary or categorical" checking is missing if user tries to mix both in several updates.

anmolsjoshi:
Fixes #348  

Description:
Bug in Binary Precision/Recall maps binary cases into 2 classes and then averages the metrics of both. This is an incorrect method of calculating precision and recall for Precision and Recall. It should be treated as a one person class only. 

I have included the following in the code:
- Created _check_shape to process and check the shapes of y, y_pred
- Created _check_type to determine the type of problem - binary or multiclass - based on y and y_pred, also raises error if the problem type changes during training. Type is decided on first update, and then checked for each subsequent update. 
- Calculates binary precision using threshold function, torch.round default
- Includes check of binary output eg: torch.equal(y, y ** 2)
- Only inputs torch.round as default is problem is binary
- Appropriate checks for threshold_function
- Added better tests - improved binary tests, incorrect threshold function, incorrect y, changing type in between updates. 

Check list:


anmolsjoshi:
@vfdev-5 please see the following PR. If we stick with this method, it'll be easier to incorporate multilabel as well.

vfdev-5:
@anmolsjoshi thanks for the PR !
Refactoring huge `update` into `_check_shape`, `_check_type` definitly makes sense. I have an impression that `_check_shape`, `_check_type` are the same for Precision and Recall ? Maybe even Accuracy could benefit from one of them ?

anmolsjoshi:
@vfdev-5 

> Refactoring huge update into _check_shape, _check_type definitly makes sense. I have an impression that _check_shape, _check_type are the same for Precision and Recall ? 

You are correct, they are the same for Precision and Recall. I took a note from how sklearn calculates precision and recall. 

> Maybe even Accuracy could benefit from one of them ?

Definitely! I can work on that later tonight.

vfdev-5:
Could you please post a reference on sklearn here ?

vfdev-5:
Tell me when you are done with the code, I'll update the tests in the same way as for Accuracy.

anmolsjoshi:
@vfdev-5 not sure what went wrong with the pytorch-nightly-cpu 2.7

anmolsjoshi:
Here is the [sklearn reference](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/metrics/classification.py#L1094-L1316)

vfdev-5:
@anmolsjoshi seems like that pytorch nightly is broken for 2.7

anmolsjoshi:
> Maybe we could create a common class for Precision, Recall in order to not to copy these methods ?

That's an interesting idea. So a base class with a common `_check_type`, `_check_shape`? looks like we can make `compute` the same too.

anmolsjoshi:
@vfdev-5 I believe that I'm done with all the tests and code. Feel free to incorporate this into Accuracy.

> @anmolsjoshi seems like that pytorch nightly is broken for 2.7

what are next steps?

vfdev-5:
Let's ignore this failing test for a moment, if this is pytorch problem, it will be solved soon. We can ask to be sure.

Next step is to refactor classes to avoid code copying.

anmolsjoshi:
@vfdev-5 ok I'll start refactoring the classes. Did you want me to work on Accuracy as well?

vfdev-5:
> Did you want me to work on Accuracy as well? 

@anmolsjoshi yes please :)

anmolsjoshi:
@vfdev-5 I refactored the code, let me know if this works for you! 

_classification_support.py
- ClassificationSupport - this will be used in Accuracy, Precision, Recall - contains type check and shape check
- PrecisionRecallSupport - this will be used to calculate Precision/Recall - contains common update, reset, compute

precision.py
- Precision - uses PrecisionRecallSupport as a base class with self._precision_vs_recall True

recall.py
- Recall - uses PrecisionRecallSuport as a base class with self._precision_vs_recall False.

vfdev-5:
@anmolsjoshi I think we can create these classes directly in `accuracy.py`, `precision.py` files.
`ClassificationSupport` -> ~ `_BaseClassification`, leave it abstract (do not define `update`, we should not instantiate this class). 
`PrecisionRecallSupport` -> ~ `_(Base)PrecisionRecall(Support)`
I'm not fan of `_precision_vs_recall` as true/false. Maybe better a function to override ?
Initialization may be specific to Accurary, Precision, Recall, IMO should not be done in `ClassificationSupport` ?

anmolsjoshi:
@vfdev-5 

> ClassificationSupport -> ~ _BaseClassification, leave it abstract (do not define update, we should not instantiate this class).

Agree, will update. 

>  I'm not fan of _precision_vs_recall as true/false. Maybe better a function to override ?
Could you explain that further? The calculation is entirely the same, the only difference is what type of positives to calculate i.e. positives of y_pred or y. precision_vs_recall might not be the cleanest way of doing this. 

> Initialization may be specific to Accurary, Precision, Recall, IMO should not be done in ClassificationSupport
Referring to [this comment](https://github.com/pytorch/ignite/pull/333#issuecomment-442970894), it might be best stick with threshold_function for binary calculation. I think initialization of _BaseClassification should stay the same, only thing that is added for Precision and Recall is the term average, which is introduced in _BasePrecisionRecallSupport (child of _BaseClassification).

>  think we can create these classes directly in accuracy.py, precision.py files. No need of _classification_support.py.

I'll move _BaseClassification to accuracy.py, and _BasePrecisionRecallSupport to precision.py, and will call it in recall.py as well.

Thoughts?

anmolsjoshi:
@vfdev-5 Here is a summary of the most up to date commit:

- accuracy.py - contains _BaseClassification (still abstract) with improved binary shape check (last few commits had incorrect logic)

- recall.py - deleted

- precision.py - renamed to precision_recall.py

- precision_recall.py - contains _BasePrecisionRecallSupport (child of _BaseClassification and still abstract), Precision and Recall. _Base..Support contains two functions _calculate_correct and _sum_positives that both Precision and Recall use. Here's how:

```python
class Precision(_BasePrecisionRecallSupport):

...

    def update(self, output):
        correct, y_pred, y = self._calculate_correct(output)
        all_positives = y_pred.sum(dim=0)
        self._sum_positives(correct, all_positives)
``` 

- __init__.py - changed for updated Precision and Recall calls

- tests_accuracy.py - Contains your modified tests from Multilabel PR and include tests similar to test_precision/recall testing threshold_function, _check_type.

Please note that Binary Accuracy is now being calculated using torch.round, I think it is the preferred method due to inconsistent results in between PyTorch versions. Discussed in this [comment](https://github.com/pytorch/ignite/pull/333#issuecomment-442970894).

anmolsjoshi:
@vfdev-5 all tests are now passing. I updated precision and recall so that they have there own unique updates. Let me know if you have any other issues!

ignite/metrics/accuracy.py
```diff
@@ -6,47 +6,98 @@
...
+class Accuracy(_BaseClassification):
+    """
+    Calculates the accuracy.
+
+    - `update` must receive output of the form `(y_pred, y)`.
+    - `y_pred` must be in the following shape (batch_size, num_categories, ...) or (batch_size, ...)
+    - `y` must be in the following shape (batch_size, ...)
+    """
+
+    def __init__(self, output_transform=lambda x: x, threshold_function=None):
```
vfdev-5: Probably, no need of constructor ?

vfdev-5: But I'm OK with that if this help user to understand better how to construct the object

ignite/metrics/precision_recall.py
```diff
@@ -0,0 +1,136 @@
+from __future__ import division
+
+import torch
+
+from ignite.metrics.accuracy import _BaseClassification
+from ignite.exceptions import NotComputableError
+from ignite._utils import to_onehot
+
+
+class _BasePrecisionRecallSupport(_BaseClassification):
```
vfdev-5: Maybe we can rename it to `_BasePrecisionRecall` ?
I understood that 'support' was coming from sklearn ...

vfdev-5:
@anmolsjoshi I'm analysing some corner cases: Precision binary input with average=False
```python

y_pred = torch.rand(10, 1)
y = torch.randint(0, 2, size=(10,))
precision.update((y_pred, y))

np_y = y.numpy()
np_y_pred = (y_pred.numpy().ravel() > 0.5).astype('int')
print(np_y)
print(np_y_pred)
precision_score(np_y, np_y_pred, average=None), precision.compute().numpy()
```
and it gives
```
[1 0 0 0 1 0 0 1 0 1]
[1 1 0 0 0 0 1 0 1 1]
(array([0.6, 0.4]), array(0.4, dtype=float32))
```
In `precision_score` when `average=None` it interprets this example as 2-classes case and the score is not averaged in the end.

ignite/metrics/accuracy.py
```diff
@@ -6,47 +6,98 @@
...
-            # Maps Binary Case to Categorical Case with 2 classes
-            y_pred = y_pred.unsqueeze(dim=1)
-            y_pred = torch.cat([1.0 - y_pred, y_pred], dim=1)
+        return y_pred, y
+
+    def _check_type(self, output):
+        y_pred, y = output
+
+        if y.ndimension() + 1 == y_pred.ndimension():
+            update_type = 'multiclass'
```
vfdev-5: Maybe we should warn user that he/she uses threshold function but input data is multiclass.

vfdev-5:
I left some comments in the code to make some changes concerning threshold function. I'm hesitating on the usage of threshold function as it is specific for binary case only.

Again, why it was decided to introduce `threshold_function`. There was a problem with binary to 2-classes mapping: `y_pred -> [1.0 - y_pred, y_pred]` which have implicit assumption that `y_pred` is between 0 and 1 and `torch.max` used as argmax gives different results for 0.4.1 and 1.0-preview.
Using `threshold_function` we handles the first limitation on `y_pred` range and we do not use torch max to compute the number of TP for the class 1 -> no more different results between torch versions. 

I have another objection agains `threshold_function`, the binarization of `y_pred` can be also done by `output_transform`. And as it is specific for binary case only, we can update the docs and require binary `y_pred` instead of logits, probas or whatever and apply a custom threshold operation. 

@jasonkriss any ideas on this ?

ignite/metrics/precision_recall.py
anmolsjoshi: Agree!

anmolsjoshi:
@vfdev-5 the use of threshold_function is for binary at this time, but in the future will be used for multilabel cases. So it might be useful to keep it. 

You are correct, we'll have these issues in multiclass cases also, eg: [0.333, 0.333, 0.333]. Although its not incorrect what argmax selects, our tests can't reflect tests that involve such cases because these tests are used for multiple versions of PyTorch, and will fail dependent on what version it is tested on. 

I don't mind switching to torch.max method. But if we want people to rely on the output_transform, when we try to add multilabel cases, are we going to ask people to use output_transform as well? My thoughts are that we should use the same method for both binary and multilabel cases. Although I think we can use torch.max for multilabel cases as well. 

We have had a user point out issues with writing their own rounding code before, see [this comment](https://github.com/pytorch/ignite/issues/262#issuecomment-421625931).

Regarding this [comment](https://github.com/pytorch/ignite/pull/349#issuecomment-443457033), problem will be solved if we used torch.max method. We'll need to modify to include an if else statement in compute for binary cases (we can think of better methods). 

```python
if self._average:
    if self._type == 'binary':
        return results[1].item()
    else:
        return result.mean().item()
else:
        return result
```

It might be helpful to include a note about `1` being the primary class in binary cases. Just so that users don't have to look into the code to figure this out.

ignite/metrics/accuracy.py
anmolsjoshi: I have now removed threshold_function. So no need to warn!

ignite/metrics/accuracy.py
anmolsjoshi: Same as above, removed threshold. Agree about removing constructor. It doesn't serve a purpose.

ignite/metrics/precision_recall.py
anmolsjoshi: Incorporated.

anmolsjoshi:
> sorry for the delay, haven't much time these days :(

@vfdev-5 no worries!

I agree with your approach and have removed threshold_function entirely. The following changes have been made. 

- Removed threshold_function from all metrics, all metrics now include Binary to Categorical Mapping
- All tests have been updated accordingly
- Binary case outputs Precision/Recall for 2 classes, when it times to compute average it only outputs the last class, if average False it returns score for both classes
- Added test for when average=None in sklearn scores, binary case
- Added test for binary cases when average=None
- I updated Precision/Recall tests similar to Accuracy to include random numbers instead of hard code values
- Removed constructor from all final metrics, Base classes handles everything
- Renamed _BasePrecisionRecallSupport to _BasePrecisionRecall

It seems that tests are passing! 

I wanted to remove [this test[(https://github.com/anmolsjoshi/ignite/blob/precisionbug/tests/ignite/metrics/test_precision.py#L75-L80) from both precision and recall. I don't think there is a case where one would average the results of Binary classes. Given that, it might be useful to include a line of code like this:

```python
if y_pred.shape[1] == 2:
    self._type = 'binary'
```

Let me know if you have any comments!

vfdev-5:
@anmolsjoshi I just looked over the PR, thanks for working on this, really good work especially with the tests !

A minor suggestion in `_Base` classes you propose two attributes
```
self._type
self._updated
```
I think we can have one `self._type`, at initialization it can be equal `None` and once specified it plays the same role as updated.

anmolsjoshi:
> I think we can have only one self._type, at initialization it can be equal None and once specified it plays the same role as updated.

@vfdev-5 I agree, thanks for a thorough review!

I have pushed this change.

anmolsjoshi:
@vfdev-5 Made some additional changes.

- Updated warning on binary accuracy and categorical accuracy to DepreciationWarning
- Updated binary accuracy and categorical accuracy tests to catch warning on said metric instantiation 
- Added assert of self._type in accuracy, binary accuracy, categorical accuracy tests

ignite/metrics/accuracy.py
```diff
@@ -6,47 +6,79 @@
...
+        self._check_type((y_pred, y))
+
         if y_pred.ndimension() == y.ndimension():
-            # Maps Binary Case to Categorical Case with 2 classes
             y_pred = y_pred.unsqueeze(dim=1)
             y_pred = torch.cat([1.0 - y_pred, y_pred], dim=1)
 
         indices = torch.max(y_pred, dim=1)[1]
-        correct = torch.eq(indices, y).view(-1)
+        correct = torch.eq(indices.view_as(y), y.type_as(indices)).view(-1)
```
vfdev-5: Why we need to do `indices.view_as(y), y.type_as(indices)` ?

ignite/metrics/__init__.py
```diff
@@ -7,8 +7,7 @@
 from ignite.metrics.mean_squared_error import MeanSquaredError
 from ignite.metrics.metric import Metric
 from ignite.metrics.epoch_metric import EpochMetric
-from ignite.metrics.precision import Precision
-from ignite.metrics.recall import Recall
+from ignite.metrics.precision_recall import Precision, Recall
```
vfdev-5: I'm not sure about to unify the files, this is a breaking change. The common class `_Base*` class we can store in `precision.py` and import then from `recall.py`

ignite/metrics/accuracy.py
anmolsjoshi: My bad! Once I fixed check_shape, this was no longer needed. I have fixed this.

ignite/metrics/__init__.py
anmolsjoshi: Done!

anmolsjoshi:
cc @jasonkriss

anmolsjoshi:
Updated docstring to mention that y_pred has to be probabilities, user should use output_transform.

vfdev-5:
@anmolsjoshi I think I badly explained the idea when I talked about to remove `threshold_function` and force user to binarize output with `output_transform`. Idea was to move `threshold_function` into `output_transform` and directly compute the metric without `binary -> categorical` mapping.

Another thing, I'm trying to list all cases and write all possible tests:

Case | y shape | y_pred | y**2 = y | y_pred**2=y_pred 
---|---|---|---|---
Binary a) | (N,) or (N, 1) | (N,) or (N, 1) | Yes | Yes 
Binary b) | (N, L) | (N, L) | Yes | Yes 
Binary c) | (N, H, W, ...) | (N, H, W, ...) | Yes | Yes 
Multiclass a) | (N,) or (N, 1) | (N, C) | No | No
Multiclass b) | (N, L) | (N, C, L) | No | No
Multiclass c) | (N, H, W, ...) | (N, C, H, W, ...) | No | No

The current implementation makes the following code ambigious:
```python
acc.update((torch.randint(0, 2, size=(10, 5)).type(torch.LongTensor),
                    torch.randint(0, 2, size=(10,)).type(torch.LongTensor)))
```
so, possible interpretations:
- bad shapes in binary case
- correct data in multiclass case: target data has only 0, 1 classes among 0-4 classes

anmolsjoshi:
@vfdev-5 could you explain why the code is ambiguous? 

Also could you give an example of the threshold function with output transform? I don’t think I understands.

vfdev-5:
> could you explain why the code is ambiguous?

the code 
```python
acc.update((torch.randint(0, 2, size=(10, 5)).type(torch.LongTensor),
             torch.randint(0, 2, size=(10,)).type(torch.LongTensor)))
```
is now interpreted as multiclass. However I supposed to test it as wrong shapes in binary. But OK probably in a sequence of binary data, when this case is encountered it will raise the problem as multiclass instead of binary.

> Also could you give an example of the threshold function with output transform? I don’t think I understand

Something like this
```python
        def thresholded_output_transform(output):
            y_pred, y = output
            y_pred = torch.round(torch.softmax(y_pred, dim=1))
            return y_pred, y

        binary_accuracy = Accuracy(thresholded_output_transform)
```

anmolsjoshi:
@vfdev-5 

> acc.update((torch.randint(0, 2, size=(10, 5)).type(torch.LongTensor),
             torch.randint(0, 2, size=(10,)).type(torch.LongTensor)))

Looking at this problem, y_pred.ndimension() == y.ndimension() + 1, currently that's how we check is something is binary or multiclass. And in the case shape[1] == 1, we squeeze y and y_pred to better decide if the data is multiclass or binary. 

Given the docs, if you want to pass an example that is a sequence of binary data it should be of the form, y=(batch_size, seq_len) and y_pred = (batch_size, num_classes=1, seq_len) or (batch_size, seq_len). The current example assumes each example of y_pred is of shape (1,5), while y is of shape (1,), thus is not setup of sequence classification of binary data. 

Given the current example you wrote up, for sequence classification if you're classifying each element of the sequence, each example of both y and y_pred should be of shape (1, 5). If the entire sequence has to be classified, then both y and y_pred should of shape (1,). 

Sorry if I'm not understanding the issue correctly.

anmolsjoshi:
@vfdev-5 as for threshold, if you want to move that to output_transform, will we provide that as the default for Accuracy, Precision and Recall?

vfdev-5:
@anmolsjoshi actually I'm not sure that it is an issue with the previous example.

> We are not using the values of y and y_pred to decide if a problem is binary or multiclass. I think we shouldn't do that because not every batch of data will have the same number of classes, meaning you could get a batch of just 1 class. 

Correct

When I have more time I'll continue with other tests on `Accuracy` and others according to the table from the previous message. Important thing is `y_pred**2=y_pred` for binary case.

vfdev-5:
> will we provide that as the default for Accuracy, Precision and Recall?

Did not get your point on this.

anmolsjoshi:
How do you think we should include thresholded_output_transform in the current API. 

The code you provided above, will we include that as an example and remove the binary to categorical mapping?

vfdev-5:
> How do you think we should include thresholded_output_transform in the current API.

As in the previous example:
```python
def thresholded_output_transform(output):
     y_pred, y = output
     y_pred = torch.round(torch.softmax(y_pred, dim=1))
     return y_pred, y

binary_accuracy = Accuracy(output_transform=thresholded_output_transform)
```
As it is shown thresholding is integrated into `output_transform`.

> The code you provided above, will we include that as an example and remove the binary to categorical mapping?

Yes

anmolsjoshi:
What are your thoughts on adding `thresholded_output_transform` to ignite._utils.py? So that users can access it.

vfdev-5:
> What are your thoughts on adding thresholded_output_transform to ignite._utils.py? So that users can access it.

For instance I'm not sure that this is necessary... Let's see

anmolsjoshi:
@vfdev-5 the original binary_accuracy had torch.round implemented within update, see [here](https://github.com/jasonkriss/ignite/blob/5db45c2427322dfd63f5fe0865aed6a81c9bc178/ignite/metrics/binary_accuracy.py#L24). 

Is there a chance we might inconvenience people by having them use output_transform? I guess if the plan is to cut a release after this is merged, we can point this out.

vfdev-5:
@anmolsjoshi this is a good point ! Actually, just remarked that in the docs it was directly stated that `y_pred` should be between 0 and 1. I think we lost this when merged binary and categorical accuracies.

anmolsjoshi:
Yes, that's probably my bad when I sent in PR #275, I missed including that docstring. 

How do you think we should proceed? Still include binary to categorical mapping? Or introduce this change in the release notes?

vfdev-5:
@anmolsjoshi no problems with that. We'll take care of it this time :) I'm readjusting the tests now.

> How do you think we should proceed? Still include binary to categorical mapping? Or introduce this change in the release notes?

Thanks for asking! So, as we discussed #348 in we need to fix the bug and cut the release `0.1.2` that should be compatible with `0.1.1`. Thus, we should include `torch.round` in binary case and update the docs  saying that `y_pred` is between 0 and 1 (in binary case).

In another PR we can remove `torch.round` and propose user to apply binarization inside `output_transform`.

vfdev-5:
I'll commit my modifications on `Accuracy` and I let you review it.

ignite/metrics/precision.py
```diff
@@ -1,60 +1,90 @@
...
-        super(Precision, self).__init__(output_transform)
+    def __init__(self, output_transform=lambda x: x, average=False):
         self._average = average
+        super(_BasePrecisionRecall, self).__init__(output_transform=output_transform)
+
+    def _check_type(self, output):
+        y_pred, y = output
+
+        if y.ndimension() + 1 == y_pred.ndimension():
+            if y_pred.shape[1] == 2:
```
vfdev-5: @anmolsjoshi could you please recall me why you separate this case and raise a warning ?

anmolsjoshi: @vfdev-5 if num_classes=2, it is a binary case that is being fed as a categorical case. I think it'll be helpful to warn the user that only precision for the positive class if calculated in this case.  Because in the case of binary, we shouldn't be average the precision of 0 and 1.

vfdev-5: But we can also have multiclass 2-classes case which should be computed as N-classes too

anmolsjoshi: True, I'll remove the binary_multiclass entirely and treat it as multiclass.

vfdev-5: I'll take it in new modifications

vfdev-5:
@anmolsjoshi I did yet not finished the code and tests on precision. Seems binary case of (N, L) is not well coded. If you would like to continue, feel free (wont touch it at least for 7-8h :)

anmolsjoshi:
@vfdev-5 thanks, this looks fantastic! I'll work on it

anmolsjoshi:
@vfdev-5 all tests are passing, let me know what you think! 

Currently, there are warnings for Precision and Recall using sklearn for cases where number of predicted or actual positives is 0 for a specific class. If we want to test average, it might be best to ignore these warnings.

vfdev-5:
@anmolsjoshi thanks a lot ! That starts look much better than the previous version :) 
I'll check out the new version and make some local tests (I had a doubt on some combinations)

vfdev-5:
@anmolsjoshi I updated the doc.

@alykhantejani @jasonkriss any comments, review please

anmolsjoshi:
@jasonkriss thanks for the review! My thought about adding the `_` for the `_Base*` classes was that we didn't want users using these base classes and just accessing the metrics only. 

This is done over [here](https://github.com/pytorch/ignite/blob/9748defd7ba19d56e1e4f1e6c87cabc1e331bfef/ignite/engine/engine.py#L8) in ignite as well.

vfdev-5:
@jasonkriss as @anmolsjoshi says the idea is to explicitly indicate that these classes are private (and abstract) and shouldn't be used by users. Another solution could be to put them into a `_basesomething.py` file and import from there as `BaseSomething`, but IMO having a private file in metrics looks strange...

jasonkriss:
@vfdev-5 @anmolsjoshi 👍 I'm fine keeping it as is.

vfdev-5:
@anmolsjoshi thank you for the PR

anmolsjoshi:
@vfdev-5 thanks for all your help! 

Is it ok to send in a PR for the thresholded_output_transform as discussed in this PR?

Once that is merged, we can incorporated all this into Multilabel metrics PR #333 

Thoughts?

vfdev-5:
@anmolsjoshi let's work directly in #333

anmolsjoshi:
@vfdev-5 sounds good

Issue #226, opened 2018-08-06
[Feature Request] Setting # train steps per epoch


kl2792:
In the case that a dataset is very large (>500k training samples) and you want to see validation loss periodically throughout the dataset.

Happy to put in a pull request implementing this -- any pointers on where to start?

vfdev-5:
@kl2792 this can be done by setting the validation handler on `Events.ITERATION_COMPLETED`:
```
trainer = ...
evaluator = ...

val_interval = 10000

@trainer.on(Events.ITERATION_COMPLETED)
def log_validation_results(engine):
    if engine.state.iteration % val_interval == 0:
        evaluator.run(val_loader)
```

alykhantejani:
Closing as this can be handled with the ITERATION_COMPLETED event as @vfdev-5 says. @kl2792 if this doesn't fully answer please feel free to comment back on this issue

vfdev-5:
Here is an example implementation what I'm talking about:
```python
from enum import Enum
from ignite.engine import Engine, Events

loss_values = list(range(500, 0, -1))
loss_iter = iter(loss_values)

def update(*args, **kwargs):
    return next(loss_iter)


engine = Engine(update)

### THE FOLLOWING CAN BE WRAPPED IN A SPECIAL HANDLER ####

class CustomEpochEvents(Enum):
    CUSTOM_EPOCH_STARTED = "custom_epoch_started"
    CUSTOM_EPOCH_COMPLETED = "custom_epoch_completed"

engine.register_events(*CustomEpochEvents)
custom_epoch_length = 3

@engine.on(Events.STARTED)
def on_started(engine):
    engine.state.custom_epoch = 0


@engine.on(Events.ITERATION_STARTED)
def on_custom_epoch_started(engine):
    if engine.state.iteration % custom_epoch_length == 1:
        engine.state.custom_epoch += 1
        engine.fire_event(CustomEpochEvents.CUSTOM_EPOCH_STARTED)


@engine.on(Events.ITERATION_COMPLETED)
def on_custom_epoch_started(engine):
    if engine.state.iteration % custom_epoch_length == 0:
        engine.fire_event(CustomEpochEvents.CUSTOM_EPOCH_COMPLETED)

### end of THE FOLLOWING CAN BE WRAPPED IN A SPECIAL HANDLER ####        

data = list(range(16))

@engine.on(Events.ITERATION_STARTED)
def on_iteration_started(engine):
[...]

PR #351, opened 2018-12-02
Added custom periodic events creator and handler

vfdev-5:
Addresses a potential feature request

Description: 
Based on the [comment](https://github.com/pytorch/ignite/issues/226#issuecomment-443426169), idea is to provide a possibility to interact in a clean way with a custom epoch started/ended events.

Check list:
* [x] New tests are added


amitibo:
@vfdev-5 my comments:
1) The name might be a bit misleading as I think most people think of `eopch` in terms of training the model on the complete training set. A possible alternative is `CustomPeriod` etc.
2) Currently one cannot define more than one `CustomEpoch`, e.g. you cannot have event fired every 1000 iterations and another one every 2000 iterations. A possible cure is to have the new code return a new custom event object to which the user can attach handlers.

vfdev-5:
@amitibo thanks for the feedback! 

Concerning the name, I agree that we probably need to have something better than `CustomEpochLength` or `CustomPeriod`. The latter one does not reflect the fact that user would like 
redefine the epoch length in terms of number of iterations.

> Currently one cannot define more than one CustomEpoch, e.g. you cannot have event fired every 1000 iterations and another one every 2000 iterations. A possible cure is to have the new code return a new custom event object to which the user can attach handlers.

Actually, I do not fully understand the purpose this. Let's say my original dataset is composed of 10k batches, I would like to validate on it more frequently, so I define a custom epoch length to 1k batches. Then I add a handler on custom event.
What the user would like to do on 2k iterations ?

amitibo:
> Actually, I do not fully understand the purpose this. Let's say my original dataset is composed of 10k batches, I would like to validate on it more frequently, so I define a custom epoch length to 1k batches. Then I add a handler on custom event.
> What the user would like to do on 2k iterations ?

I also don't have a concrete example, but the current implementation might confuse the user, i.e. it is not clear from the name/docs/example that it is possible to have only one custom period.

vfdev-5:
@amitibo yes you are right, we need to warn user in the code and in the doc that he/she can add such handler only once

amitibo:
> Actually, I do not fully understand the purpose this. Let's say my original dataset is composed of 10k batches, I would like to validate on it more frequently, so I define a custom epoch length to 1k batches. Then I add a handler on custom event.
> What the user would like to do on 2k iterations ?

@vfdev-5 on a second thought, you cannot anticipate what users will do with your code. To my understanding the purpose of `ignite` is to make the code simpler to read and write, i.e. more elegant. So what is the purpose of this new feature, a new custom epoch, or a new periodic event? I think that a new periodic event might be a more elegant description. If so, why limit it to a single period per program?

amitibo:
@vfdev-5  one scenario that I use frequently and that can benefit from multiple periodic events, is for example, logging the training loss every 10 iterations (frequent logging with visdom can slow training due to network) and evaluating the validation loss every 1000 iterations.

vfdev-5:
@amitibo I understand your point, so to achieve the following goal:

> logging the training loss every 10 iterations

actually user just needs to implement a handler:
```python
def log_every_10_iterations(engine):
    if engine.state.iteration % 10 == 0:
        log_something(engine)
```
and you propose instead to implement this using custom events:
```
CustomPeriodicEvent(n_iterations=10).attach(trainer)

@engine.on(CustomPeriodicEvent.ITERATIONS_10_COMPLETED)
def log_every_10_iterations(engine):
        log_something(engine)
```

Sure that we gain in factoring `if engine.state.iteration % 10 == 0` with several handlers using this custom event. But IMO implementing `CustomPeriodicEvent` to replace `if iteration % 10 == 0` looks like a bit heavy...
What do you think ?

amitibo:
> Sure that we gain in factoring `if engine.state.iteration % 10 == 0` with several handlers using this custom event. But IMO implementing `CustomPeriodicEvent` to replace `if iteration % 10 == 0` looks like a bit heavy...
> What do you think ?

I agree that it looks a bit heavy. Though having custom events can simplify the specific code to:

```
CustomPeriodicEvent(n_iterations=10).attach(trainer)

trainer.add_event_handler(CustomPeriodicEvent.ITERATIONS_10_COMPLETED, log_something)
```

I am not sure though that it is general enough.

vfdev-5:
@amitibo another use case to support `CustomPeriodicEvent`  instead of `CustomEpochEvent` :
```python
CustomPeriodicEvent(n_iterations=1000).attach(trainer)
checkpointer = ModelCheckpoint(checkpoint_dir, 'my_model', save_interval=1, create_dir=True)
trainer.add_event_handler(Events.ITERATIONS_1000_COMPLETED, checkpointer, {'mymodel': model})
```
vs original version
```python
checkpointer = ModelCheckpoint(checkpoint_dir, 'my_model', save_interval=1000, create_dir=True)
trainer.add_event_handler(Events.ITERATION_COMPLETED, checkpointer, {'mymodel': model})
```

vfdev-5:
New API:
```python
        from ignite.handlers import CustomPeriodicEvent

        # Let's define an event every 1000 iterations
        cpe1 = CustomPeriodicEvent(n_iterations=1000)
        cpe1.attach(trainer)

        # Let's define an event every 10 epochs
        cpe2 = CustomPeriodicEvent(n_epochs=10)
        cpe2.attach(trainer)

        @trainer.on(cpe1.Events.ITERATIONS_1000_COMPLETED)
        def on_every_1000_iterations(engine):
            # run a computation after 1000 iterations
            # ...
            print(engine.state.iterations_1000)

        @trainer.on(cpe2.Events.EPOCHS_10_STARTED)
        def on_every_10_epochs(engine):
            # run a computation every 10 epochs
            # ...
            print(engine.state.epochs_10)

```

ignite/contrib/handlers/custom_events.py
```diff
@@ -0,0 +1,103 @@
...
+    - `Events.EPOCHS_<K>_STARTED`
+    - `Events.EPOCHS_<K>_COMPLETED`
+
+
+    Examples:
+
+    .. code-block:: python
+
+        from ignite.engine import Engine, Events
+        from ignite.handlers import CustomPeriodicEvent
```
anmolsjoshi: ignite.contrib.handlers

vfdev-5: Done :)

PR #352, opened 2018-12-03
Pytorch schedulers

amitibo:
Description:

Add wrappers that allow using "native" `torch.optim lr_scheduler` objects as (or together with)`ignite.contrib.handlers.param_scheduler` objects. For example this allowing attaching `ReduceLROnPlateau` to a `trainer` object or concatenate it with other `ParamScheduler` objects.

Check list:

### DIFF
--- a/ignite/contrib/handlers/param_scheduler.py
+++ b/ignite/contrib/handlers/param_scheduler.py
@@ -284,3 +284,165 @@ class ConcatScheduler(ParamScheduler):
                 self._next_scheduler()
 
         return self._scheduler(engine, name)
+
+
+class LRScheduler(ParamScheduler):
+    """A wrapper class to call `torch.optim.lr_scheduler` objects as `ignite` handlers.
+
+    Args:
+        lr_scheduler (subclass of `torch.optim.lr_scheduler._LRScheduler`): The lr_scheduler object to wrap.
+        save_history (bool, optional): whether to log the parameter values
+            (default=False)
+
+    .. code-block:: python
+
+        from ignite.contrib.handlers.param_scheduler import LRScheduler
+        from torch.optim.lr_scheduler import StepLR
+
+        step_scheduler = StepLR(optimizer, step_size=3, gamma=0.1)
+        scheduler = LRScheduler(step_scheduler)
+
+        trainer.add_event_handler(Events.ITERATION_COMPLETED, scheduler)
+    """
+    def __init__(self, lr_scheduler, save_history=False):
+
+        self.lr_scheduler = lr_scheduler
+        super(LRScheduler, self).__init__(
+            optimizer=self.lr_scheduler.optimizer,
+            param_name='lr',
+            save_history=save_history
+        )
+
+    def __call__(self, engine, name=None):
+
+        #
+        # Call the pytorch scheduler step method.
+        #
+        self.lr_scheduler.step(engine.state.epoch)
+
+        #
+        # Call the ignite object __call__ method.
+        # Note:
+        # The 'lr' parameter will be updated for the second time but with the same value.
+        #
+        super(LRScheduler, self).__call__(engine, name)
+
+    def get_param(self):
+        """Method to get current optimizer's parameter value
+        """
+        return self.lr_scheduler.get_lr()
+
+
+class ReduceLROnPlateau(ParamScheduler):
+    """An adapter class to call `torch.optim.lr_scheduler.ReduceLROnPlateau` object as `ignite` handlers.
+
+    Args:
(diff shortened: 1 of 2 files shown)


ignite/contrib/handlers/param_scheduler.py
```diff
@@ -284,3 +284,165 @@ def __call__(self, engine, name=None):
...
+            param_name='lr',
+            save_history=save_history
+        )
+
+    def __call__(self, engine, name=None):
+
+        #
+        # Call the pytorch scheduler step method.
+        #
+        self.lr_scheduler.step(engine.state.epoch)
```
vfdev-5: I think we should not attach the scheduler to the epochs.
Why can't we simply do 
```python
self.lr_scheduler.last_epoch += 1
```
without any update by pytorch and update params from ParamScheduler side ?

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -284,3 +284,165 @@ def __call__(self, engine, name=None):
...
+        # Call the ignite object __call__ method.
+        # Note:
+        # The 'lr' parameter will be updated for the second time but with the same value.
+        #
+        super(LRScheduler, self).__call__(engine, name)
+
+    def get_param(self):
+        """Method to get current optimizer's parameter value
+        """
+        return self.lr_scheduler.get_lr()
```
vfdev-5: This can be tricky, `lr_scheduler.get_lr()` return a list of values and not a single value.

vfdev-5:
@amitibo I'm trying to check `LRScheduler` for some basic schedulers e.g. `StepLR`.
I left some comments in the code.

Should not we insure that the optimizer passed to an lr_scheduler should have a single param group ?
Let me explain:
```python
t1 = torch.zeros([1], requires_grad=True)
optimizer = torch.optim.SGD([t1], lr=0.1)
lr_scheduler = StepLR(optimizer=optimizer, step_size=2)
lr_scheduler.get_lr()
> [0.1]
```
and 
```python
t1 = torch.zeros([1], requires_grad=True)
t2 = torch.zeros([1], requires_grad=True)
t3 = torch.zeros([1], requires_grad=True)

optimizer = torch.optim.SGD(
    [
        {"params": t1, 'lr': 0.001},
        {"params": t2, 'lr': 0.01},
        {"params": t3, 'lr': 0.0001}
    ]
)
lr_scheduler = StepLR(optimizer=optimizer, step_size=2)
lr_scheduler.get_lr()
> [0.01, 0.1, 0.001]
```

Code of `lr_scheduler.step()` of `_LRScheduler` assign values as 
```python
for param_group, lr in zip(self.optimizer.param_groups, self.get_lr()):
            param_group['lr'] = lr
```
But in our case, 
```python
value = self.get_param()

for param_group in self.optimizer_param_groups:
    param_group[self.param_name] = value
```

amitibo:
> @amitibo I'm trying to check `LRScheduler` for some basic schedulers e.g. `StepLR`.
> I left some comments in the code.
> 
> Should not we insure that the optimizer passed to an lr_scheduler should have a single param group ?
> Let me explain:
> 
> ```python
> t1 = torch.zeros([1], requires_grad=True)
> optimizer = torch.optim.SGD([t1], lr=0.1)
> lr_scheduler = StepLR(optimizer=optimizer, step_size=2)
> lr_scheduler.get_lr()
> > [0.1]
> ```
> 
> and
> 
> ```python
> t1 = torch.zeros([1], requires_grad=True)
> t2 = torch.zeros([1], requires_grad=True)
> t3 = torch.zeros([1], requires_grad=True)
> 
> optimizer = torch.optim.SGD(
>     [
>         {"params": t1, 'lr': 0.001},
>         {"params": t2, 'lr': 0.01},
>         {"params": t3, 'lr': 0.0001}
>     ]
> )
> lr_scheduler = StepLR(optimizer=optimizer, step_size=2)
> lr_scheduler.get_lr()
> > [0.01, 0.1, 0.001]
> ```
> 
> Code of `lr_scheduler.step()` of `_LRScheduler` assign values as
> 
> ```python
> for param_group, lr in zip(self.optimizer.param_groups, self.get_lr()):
>             param_group['lr'] = lr
> ```
> 
> But in our case,
> 
> ```python
> value = self.get_param()
> 
> for param_group in self.optimizer_param_groups:
>     param_group[self.param_name] = value
> ```

@vfdev-5, sorry for the late reply, I think it should be the other way round, `ParamScheduler` should support a per param group parameter. This way ignite can support the full capabilities of `pytroch.lr_scheduler` instead of downgrading it. I also think that it can be done without changing the current API.

vfdev-5:
@amitibo could you please create an issue for that ?

PR #356, opened 2018-12-06
Use collections.abc for isinstance calls

### LINKED ISSUE #355: Use collections.abc in ProgressBar instead of checking isinstance() against list or dict
Expecting a `list` or `dict` is overly restrictive when we really just need a sequence or mapping. I have a patch that checks against `collections.abc.Sequence` instead of `list` and `collections.abc.Mapping` instead of `dict`. This will work for both Python 3 and 2.7.

If you're cool with this change, I'll submit a PR for the patch.

Thanks.

willfrey:
Fixes #355

### DIFF
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -130,9 +130,9 @@ class ProgressBar:
                 output. This function may return either a dictionary with entries in the format of ``{name: value}``,
                 or a single scalar, which will be displayed with the default name `output`.
         """
-        if (metric_names is not None
-                and not isinstance(metric_names, _collections_abc.Sequence)
-                or isinstance(metric_names, str)):
+        if (metric_names is not None and
+           not isinstance(metric_names, _collections_abc.Sequence) or
+           isinstance(metric_names, str)):
             raise TypeError("metric_names should be a sequence of strings, got {} instead".format(type(metric_names)))
 
         if output_transform is not None and not callable(output_transform):
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -1,4 +1,9 @@
 try:
+    import collections.abc as _collections_abc
+except ImportError:  # Python 2.7 compatibility
+    import collections as _collections_abc
+
+try:
     from tqdm import tqdm
 except ImportError:
     raise RuntimeError("This contrib module requires tqdm to be installed")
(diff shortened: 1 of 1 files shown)


ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -1,3 +1,8 @@
+try:
+    import collections.abc as _collections_abc
```
alykhantejani: we could do 
```
from collections.abc import Mapping, Sequence 
```
it would avoid the need for `_collections_abc`

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -125,8 +130,8 @@ def attach(self, engine, metric_names=None, output_transform=None):
                 output. This function may return either a dictionary with entries in the format of ``{name: value}``,
                 or a single scalar, which will be displayed with the default name `output`.
         """
-        if metric_names is not None and not isinstance(metric_names, list):
-            raise TypeError("metric_names should be a list, got {} instead".format(type(metric_names)))
+        if metric_names is not None and not isinstance(metric_names, _collections_abc.Sequence):
```
alykhantejani: This will also allow incorrect types like string to be passed through (which is why the build is failing as we have a test for this)

ignite/contrib/handlers/tqdm_logger.py
willfrey: I personally tend to avoid importing symbols that aren't packages or modules unless they're from the `typing` module, but that's just personal preference.

If we import those specifically like you have and someone did want to add typing annotations in the future, you'd get a namespace clash.

That's just personal preference though.

ignite/contrib/handlers/tqdm_logger.py
willfrey: I've modified the guard statement to account for strings. Would this be better with a `try`/`except` block though? That is, assume that the client gave us something valid for `metric_names` and catch any exceptions that are raised and raise our own `TypeError`?

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -125,8 +130,10 @@ def attach(self, engine, metric_names=None, output_transform=None):
                 output. This function may return either a dictionary with entries in the format of ``{name: value}``,
                 or a single scalar, which will be displayed with the default name `output`.
         """
-        if metric_names is not None and not isinstance(metric_names, list):
-            raise TypeError("metric_names should be a list, got {} instead".format(type(metric_names)))
+        if (metric_names is not None and
+           not isinstance(metric_names, _collections_abc.Sequence) or
```
vfdev-5: I'm not fan of this, maybe we can just replace `list` -> `(list, tuple)`

Issue #289, opened 2018-10-15
[Feature Request]  tqdm_logger update on every epoch instead of every iteration


PsiPhire:
Currently `tqdm_logger` creates a progress bar for every epoch and updates it on every iteration. I would like to have a single progress bar for the entire run that updates every epoch instead.

miguelvr:
@vfdev-5 I could add back the `mode` argument to support this. Should be relatively straightforward

vfdev-5:
@miguelvr yes, go ahead.

alykhantejani:
Thanks @miguelvr

vfdev-5:
@miguelvr any updates on this :)

miguelvr:
I started it already, but I have been quite busy with work. Will work on it soon, sorry for the delay

vfdev-5:
@miguelvr no problem, take your take :) I had a doubt whether it was already done in a PR on tqdm

vfdev-5:
@miguelvr any updates on this ?

vfdev-5:
@miguelvr I tried to implement this one by myself. Seems to be not a trivial one without changing the API and keeping all simple

vfdev-5:
I was also thought about how to itegrate this one with [this feature](https://github.com/pytorch/ignite/pull/351): to support custom events

PR #360, opened 2018-12-11
Unified tensorboard handler API

amitibo:
A new tensorboard handler API (following #265)

Description:
Following https://github.com/pytorch/ignite/pull/265#issuecomment-437720047 this PR suggests a modification to the TensorboardLogger that is more similar to the ProgressBar and VisdomLogger API.

Check list:

### DIFF
--- a/examples/mnist/mnist_with_tensorboardx.py
+++ b/examples/mnist/mnist_with_tensorboardx.py
@@ -13,9 +13,9 @@
     python mnist_with_tensorboardx.py --log_dir=/tmp/tensorboard_logs
     ```
 """
-
 from __future__ import print_function
 from argparse import ArgumentParser
+
 import torch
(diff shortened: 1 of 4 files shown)


amitibo:
@vfdev-5 I will separate these, though I still think that the API is somewhat cumbersome. For example, maybe it is better to attach the metrics to the loggers, as you actually log the metrics not the engine. This way there is no need to set the logging event, as the logger can log whenever the metrics are updated. Another example is to attach the model to the logger to get the histograms and graph.

vfdev-5:
@amitibo I separated tqdm part into #365 , please take a look and comment out if any other suggestions.

vfdev-5:
> This way there is no need to set the logging event, as the logger can log whenever the metrics are updated. 

@amitibo there is probably a similar problem in #289 with tqdm

PR #364, opened 2018-12-11
[ci skip][DO NOT MERGE] Discussion on deploy nightly

zasdfgbnm:
https://github.com/pytorch/ignite/issues/306

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -167,7 +167,7 @@ jobs:
       python: "3.5"
       env:
         - PYTORCH_PACKAGE=pytorch-nightly-cpu
-      if: branch = master
+      if: branch = master & type = cron
       # Use previously defined before_install
       before_install: *before_install
       install:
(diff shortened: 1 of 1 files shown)


vfdev-5:
I thought that this could be a solution if nightly is a separate root as `gh-pages`. 
But actually, I'm not sure about how to fetch the code from it and put it into master. If we keep `nightly` as a branch from master, then how we keep it updated ?

zasdfgbnm:
@vfdev-5 Do we need to keep nightly updated in this case? As long as there is no merge conflict, it could be merged automatically into master. Just
```
git pull origin nightly
```
would work.


Since all changes in nightly branch are in a separate directory, it is unlikely to have conflict. We only need to update nightly branch manually if there is a conflict.

The `Nightly` stage in this design is on master branch and is only activated on cron, and there is no CI on nightly branch. All nightly does is to store the nightly build files.

vfdev-5:
@zasdfgbnm need to try this solution

PR #365, opened 2018-12-12
Added kwargs into progressbar

vfdev-5:
This is the second part of PR #360 split

Description:
Added kwargs to contrib `tqdm_logger`

Check list:

### DIFF
--- a/examples/contrib/mnist_with_tqdm_logger.py
+++ b/examples/contrib/mnist_with_tqdm_logger.py
@@ -1,0 +1,108 @@
+from argparse import ArgumentParser
+
+from torch import nn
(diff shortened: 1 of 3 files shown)


amitibo:
Looks good.
I suggest to update the examples with the progress bar. This way users can copy paste into their code.

miguelvr:
hey guys, I've been away from this for a bit and I'm amazed by the number of changes that have been going on.

What is the use of the kwargs in ProgressBar?

amitibo:
@miguelvr  I use it for configuring the tqdm progressbar, e.g. set its description or change the progressbar fill (my terminal doesn't show unicode chars).

PR #366, opened 2018-12-13
Updating Fast Neural Style Example

anmolsjoshi:
Updated Progress Bar in Fast Neural Style Example

Description:
Integrated Progress Bar from ignite.contrib.handlers to example.


anmolsjoshi:
@vfdev-5 maybe this needs to be moved to `examples/contrib` just as `mnist_with_tqdm_logger.py` has been.

vfdev-5:
@anmolsjoshi yes, I think so

anmolsjoshi:
@vfdev-5 I moved the example, similarly i think dcgan.py needs to be moved also, as it uses tqdm_logger.py and we need to add a smoke_test for the mnist_with_tqdm_logger.py.

vfdev-5:
Actually I have doubt now. The first idea was to provide some examples which use `contrib` module. No smoke tests as it will become too heavy. However I would also like to keep base dcgan and fast neural style tutorial in offical non-contrib examples. 
In case of `mnist_with_tqdm_logger.py`, this one is a copy of `mnist.py` + adaptions.
If we just move fast neural style to contrib this will look like a regression...

miguelvr:
@vfdev-5 why not move the `ProgressBar` out of contrib?

vfdev-5:
@miguelvr this was another idea (not exposed in the previous message) which requires to add a new dependency to ignite. Need to discuss about this. Especially, in the very beginning we thought about that some helpful stuff could migrate from contrib to core. Today there is still one issue on progressbar to solve and probably this will take time to confirm the API ...

miguelvr:
what is the issue?

anmolsjoshi:
@vfdev-5 should i close this PR until #289 is resolved?

Issue #373, opened 2018-12-14
Issue with traced models


vfdev-5:
If user passes a traced model into `create_supervised*` methods and specifies a device, there will be the following error:
```python
traced_model = torch.jit.trace(model, z)

evaluator = create_supervised_evaluator(traced_model, metrics, device='cuda')
> RuntimeError: to is not supported on TracedModules
```

We need to check if model have the method `.to`...

vfdev-5:
Probably, we should wait until [this PR](https://github.com/pytorch/pytorch/issues/6008) is done.

TheCodez:
> the test should pass for pytorch-nightly build (can be check with torch.__version__)

Do you mean only executing the test if on the nightly, or check if it throws a `RuntimeError` if not on the nightly?

vfdev-5:
@TheCodez yes something like this.

TheCodez:
@vfdev-5 which `torch.__version__` to check for nightly?

vfdev-5:
@TheCodez you can do it like this:
```python
# Check if torch-nightly
if "dev" in torch.__version__:
     ...
else:
    ...
```

PR #374, opened 2018-12-14
Simple fix for #373

TheCodez:
Is this enough or should we also check if 'to' is a method?

### DIFF
--- a/ignite/engine/__init__.py
+++ b/ignite/engine/__init__.py
@@ -33,7 +33,7 @@ def create_supervised_trainer(model, optimizer, loss_fn,
     Returns:
         Engine: a trainer engine with supervised update function
     """
-    if device:
+    if device and hasattr(model, 'to'):
         model.to(device)
 
     def _update(engine, batch):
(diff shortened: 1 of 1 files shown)


vfdev-5:
@TheCodez thanks for the PR, but #373 is more subtle than that. Traced/Scripted models has attribute `to` as they derive from `nn.Module`:
```python
hasattr(traced_model, "to"), hasattr(scripted_model, "to")
> (True, True)
```

Probably one can trace a model when its on GPU and would like to run it on a data being on GPU...
We need to think about all these cases

TheCodez:
> RuntimeError: to is not supported on TracedModules

Instead of checking if we have the `to` method, we should probably just check if the model passed in is a `ScriptModule`.

vfdev-5:
@TheCodez yes, we can differentiate whether model is `ScriptModule` or not with `isinstance`.

> In that case we should warn the user, if the device is different from the one of the traced model, right?

I'm not sure if we can get the device of the traced/scripted model...

Another point to confirm, can we train traced/scripted models ? Otherwise we should raise an error if `ScriptModule` is passed to `create_supervised_trainer`.

TheCodez:
According to https://github.com/pytorch/pytorch/issues/14696 they can be trained

TheCodez:
I did some changes, which for now just throws a `RuntimeException `if a device is passed when using traced modules.

TheCodez:
Closing this as it should be solved when https://github.com/pytorch/pytorch/pull/15340 is merged.

PR #383, opened 2018-12-26
[WIP] Helper function to create engines with fp16 using nvidia/apex

vfdev-5:
New feature to create engines with fp16 using nvidia/apex

Check list:

### DIFF
--- a/ignite/contrib/engines/fp16.py
+++ b/ignite/contrib/engines/fp16.py
@@ -1,0 +1,102 @@
+import torch
+import torch.nn as nn
+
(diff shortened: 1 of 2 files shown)


khoa-ho:
Hi @vfdev-5, is this still WIP?

vfdev-5:
@khoa-ho nvidia/apex has recently changed the API and I think the current implementation is not valid. 
Their new API is very simple and I need to think more on what could we gain while providing a helpers for fp16 in ignite.

PR #385, opened 2018-12-26
Updated MetricsLambda to only attach Lambda Metric to engine.state.metrics

### LINKED ISSUE #381: Improve MetricLambda implementation
User can get the following output in the `state.metrics`:
```
{'accuracy': 0.37127896829244644,
 'error': 0.6287210317075536,
 'error[0]': -0.6287210317075536,
 'error[0][0]': 0.37127896829244644,
```
when launch the below code:
```python
accuracy_metric = Accuracy()
error_metric = (accuracy_metric - 1.0) * (-1.0)

metrics = {
    "accuracy": accuracy_metric,
    "error": error_metric,
}

validator = create_supervised_evaluator(model, metrics=metrics)
validator.run(val_loader, max_epochs=1)
print(validator.state.metrics)
```

This is due to 

https://github.com/pytorch/ignite/blob/d9820451da779e0d0c393804db381e5483240b1c/ignite/metrics/metrics_lambda.py#L50-L54

and 

https://github.com/pytorch/ignite/blob/d9820451da779e0d0c393804db381e5483240b1c/ignite/metrics/metric.py#L68

IMO, user is not interested of this internal info used to compute `error_metric`. We can add some special characters to ignore the result of compute when insert to `engine.state.metrics[name]`.

cc @zasdfgbnm

anmolsjoshi:
Fixes #381 

Description:
Adds special characters to metric names used in calculation of MetricsLambda

Check list:


vfdev-5:
@anmolsjoshi thanks for the PR, however I thought about something like this : 
```python
class Metric(with_metaclass(ABCMeta, object)):

    name_to_ignore = "~MetricNameToIgnore~"
    
    def completed(self, engine, name):
        if self.name_to_ignore not in name:
            engine.state.metrics[name] = self.compute()


class MetricsLambda(Metric):

        # recursively attach all its dependencies
        for index, metric in enumerate(self.args):
            if isinstance(metric, Metric):
                metric.attach(engine, name + '{}[{}]'.format(self.name_to_ignore, index))
        super(MetricsLambda, self).attach(engine, name)
```

anmolsjoshi:
@vfdev-5 your changes have been incorporated.

vfdev-5:
@anmolsjoshi could you please write a test checking that dictionary contains only what we need. Take a test example with a metric composed of several internal metrics, e.g F1

ignite/metrics/metrics_lambda.py
```diff
@@ -51,5 +52,8 @@ def attach(self, engine, name):
         # recursively attach all its dependencies
         for index, metric in enumerate(self.args):
             if isinstance(metric, Metric):
-                metric.attach(engine, name + '[{}]'.format(index))
-        super(MetricsLambda, self).attach(engine, name)
+                if not engine.has_event_handler(metric.started, Events.EPOCH_STARTED):
+                    engine.add_event_handler(Events.EPOCH_STARTED, metric.started)
+                if not engine.has_event_handler(metric.iteration_completed, Events.ITERATION_COMPLETED):
+                    engine.add_event_handler(Events.ITERATION_COMPLETED, metric.iteration_completed)
+            super(MetricsLambda, self).attach(engine, name)
```
vfdev-5: Wrong indent here, IMO

anmolsjoshi: Corrected!

anmolsjoshi:
@vfdev-5 Added a test!

vfdev-5:
@anmolsjoshi thanks !

anmolsjoshi:
@amitibo the above code would only work if you attached accuracy_1 and accuracy_2, since you aren't currently using MetricsLambda. 

Even with the previous implementation of MetricsLambda, the above code would not work since MetricsLambda is not used. 

As shown in the [docs](https://pytorch.org/ignite/metrics.html#ignite.metrics.MetricsLambda), MetricsLambda requires a callable function and the related arguments. 

You can modify your code like this:

```python
accuracy_1 = Accuracy()
accuracy_2 = Accuracy()

def mean_accuracy_function(accuracy_1, accuracy_2):
    return (accuracy_1 + accuracy_2) / 2

mean_accuracy = MetricsLambda(mean_accuracy_function, accuracy_1, accuracy_2)

metrics = {
    "mean accuracy": mean_accuracy ,
}

validator = create_supervised_evaluator(model, metrics=metrics)
validator.run(val_loader, max_epochs=1)
print(validator.state.metrics)
```

If you choose not to use MetricsLambda, both accuracy_1 and accuracy_2 need to be attached to the Engine.

amitibo:
@anmolsjoshi are your sure? As far as I understand, the `+` and `/` operators call the `__add__` and `__div__` methods of `Metric` which return a `MetricsLambda`.

anmolsjoshi:
@amitibo you are correct. will look into this further.

amitibo:
@anmolsjoshi @vfdev-5 I prefer your original solution (add special characters or name_to_ignore)

vfdev-5:
@amitibo we will investigate the bug and try to repare it instead of revert to the previous version. IMO, in your example the previous version would give `state.metrics` something like
```python
{
'mean accuracy': X + Y,
 'mean accuracy[0]': X,
 'mean accuracy[0]': Y
}
```
and those `mean accuracy[i]` lead to some confusion that we need to avoid. What do you think ?

vfdev-5:
@anmolsjoshi the bug is with recursive attach of `MetricsLambda` that we broken by this PR.
For example, the following will work
```
accuracy_1 = Accuracy()
accuracy_2 = Accuracy()
mean_accuracy = accuracy_1 + accuracy_2
```

PR #386, opened 2018-12-27
Updated pbar description: custom desc + [epoch/n_epochs]

vfdev-5:
Description:
Minor pbar update to display current epoch / num_epochs with custom description:
```python
    pbar = ProgressBar(desc="Validation")
    pbar.attach(engine, output_transform=lambda x: x)
    engine.run(loader, max_epochs=n_epochs)
    expected = u'Validation [10/10]: [4/5]  80%|████████  , output=1.00e+00 [00:00<00:00]'
```

Check list:
* [x] Tests updated

cc @amitibo @miguelvr

### DIFF
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -20,9 +20,9 @@ class ProgressBar(object):
             r_bar='| {n_fmt}/{total_fmt} [{elapsed}<{remaining}, {rate_fmt}{postfix}]'. For more details on the
             formatting, see `tqdm docs <https://tqdm.github.io/docs/tqdm/>`_.
         **tqdm_kwargs: kwargs passed to tqdm progress bar.
-            By default, progress bar description displays "Epoch [5/10]". If tqdm_kwargs defines
-            `desc`, e.g. "Predictions", than the description is "Predictions [5/7]" if number of epochs is more
-            than one otherwise it is simply "Predictions".
+            By default, progress bar description displays "Epoch [5/10]" where 5 is the current epoch and 10 is the
+            number of epochs. If tqdm_kwargs defines `desc`, e.g. "Predictions", than the description is
+            "Predictions [5/10]" if number of epochs is more than one otherwise it is simply "Predictions".
 
     Examples:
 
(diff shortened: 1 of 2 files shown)


amitibo:
@vfdev-5 I think you can write:
```
desc = "Epoch" if 'desc' not in self.tqdm_kwargs else self.tqdm_kwargs['desc']
```
easier this way:
```
desc = self.tqdm_kwargs.get("desc", "Epoch")
```

If I am correct this PR forces the description to include the [epoch/max_epoch] construct, whether the user wants it or not. Is this what you want? If so, maybe it is better to describe this in the docstring.

vfdev-5:
@amitibo yes, you are correct. Will fix the code and the docstring.

vfdev-5:
@amitibo updated docs and the code such that append to desc `[curr_epoch/num_epoch]` if more than one.

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -17,14 +17,12 @@ class ProgressBar(object):
...
-              percentage, rate, rate_fmt, rate_noinv, rate_noinv_fmt,
-              rate_inv, rate_inv_fmt, elapsed, remaining, desc, postfix.
-            Note that a trailing ": " is automatically removed after {desc}
-            if the latter is empty.
-        **tqdm_kwargs: kwargs passed to tqdm progress bar
+            r_bar='| {n_fmt}/{total_fmt} [{elapsed}<{remaining}, {rate_fmt}{postfix}]'. For more details on the
+            formatting, see `tqdm docs <https://tqdm.github.io/docs/tqdm/>`_.
+        **tqdm_kwargs: kwargs passed to tqdm progress bar.
+            By default, progress bar description displays "Epoch [5/10]". If tqdm_kwargs defines
+            `desc`, e.g. "Predictions", than the description is "Predictions [5/7]" if number of epochs is more
```
jasonkriss: can we change this to "Predictions [5/10]"? Don't want anyone getting confused over why it went from "5/10" to "5/7"

vfdev-5: Yes, you are right. It is unclear what are these numbers 5, 7 and 10. I'll improve this.

jasonkriss: @vfdev-5 looks perfect now! I'll let you merge when you are ready.

PR #397, opened 2019-01-09
Add start_epoch to Engine.run

willprice:
**Description:**
This is useful for those wishing to resume training an existing
partially trained model. If you've got code set up to use
`engine.state.epochs` ideally you want to be able to resume training
without having to change this code to take into account the fact the
model has already been partially trained. The reason for adding the argument
after `max_epochs` is to maintain compatibility with existing code that uses the
position of `max_epochs` rather than kwarg.

**Check list:**

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -290,18 +290,19 @@ class Engine(object):
         else:
             raise e
 
-    def run(self, data, max_epochs=1):
+    def run(self, data, max_epochs=1, start_epoch=0):
         """Runs the process_function over the passed data.
 
         Args:
             data (Iterable): Collection of batches allowing repeated iteration (e.g., list or `DataLoader`)
             max_epochs (int, optional): max epochs to run for (default: 1)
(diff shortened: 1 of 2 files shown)


vfdev-5:
@willprice thanks for the PR, however engine checkpointing and resuming from a state still remains as open issue in ignite. Please, take a look at #182 and tell me what you think ?

TheCodez:
@vfdev-5 I think this could still be useful for now until a complete checkpointing solution is implemented. What do you think?

vfdev-5:
@TheCodez I agree with what you're saying. We have not yet scheduled when we end up with #182  but the proposed workaround changes the API that will be finally removed. I think the above behaviour can be also achieved by tweaking `state.epoch` at `Events.STARTED`:
```

trainer = ...

@trainer.on(Events.STARTED)
def setup_state(engine):
    engine.state.epoch = start_epoch
```

willprice:
I'll close this given that the current version of ignite has a suitable workaround until #182 lands.
Thanks for the workaround @vfdev-5 :+1:

PR #399, opened 2019-01-15
State docfix

### LINKED ISSUE #393: Improve State documentation
Hi,

I think the documentation of states and the way they're used by engines is not detailed enough.
Currently the only way to know what attributes are available in the default State object, used by the engines generated by convenience methods such as `create_supervised_trainer`, is to look at the source code.  
Some of them are mentioned in the [Concept](https://pytorch.org/ignite/concepts.html#state) part of the documentation but there is still a lot of room for confusion about the way they work and some are missing.
For example, no information about the `batch` attribute which without documentation some might assume to correspond to the batch index (kinda like `epoch` for the epoch index) which is actually `iteration`. Same thing for `epoch` which is used implicitly in examples but no mention of it in the default `State` documentation, and you might think it's obvious but then someone will try to create a default `State` and assume `epoch` is a built-in attribute set to 0 at instantiation when in fact you'll have to search through the source code to figure out that it's not a default attribute and is only created by the default `Engine` through a kwarg passed to the `State` class.

Additionally this kind of explicit documentation of the state should be done every time a custom `State` is created otherwise the user is left to guess what attributes are available in it, what they mean and what format they're in. For example, is `output` the return from the forward pass of your model? The loss?
You'll have to look at `Engine` then find out it's the output of the process_function argument. Then figure out that in the case of the  `create_supervised_trainer` that function is `_update` which returns `loss.item()` and in the case of `create_supervised_evaluator` returns `(y_pred, y)`.

Thank you, and sorry for the verbosity but without it I was afraid this would be rejected as trivial.

anmolsjoshi:
Fixes #393  

Description:
Updates concepts.rst and descriptions of create_supervised_trainer and create_supervised_evaluator to provide explanations and examples of engine.state.output

Check list:

### DIFF
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -103,11 +103,10 @@ State
 A state is introduced in :class:`~ignite.engine.Engine` to store the output of the `process_function`, current epoch,
 iteration and other helpful information. Each :class:`~ignite.engine.Engine` contains a :class:`~ignite.engine.State`, 
 which includes the following:
-* `engine.state.epoch`: Number of epochs the engine has completed. Initializated as 0. 
-* `engine.state.iteration`: Number of iterations the engine has completed. Initialized as 0. Note that it is not
-    reset at every epoch.
-* `engine.state.output`: The output of the `process_function` defined for the :class:`~ignite.engine.Engine`. See
-    examples below.
+
+- *`engine.state.epoch`: Number of epochs the engine has completed. Initializated as 0. 
+- *`engine.state.iteration`: Number of iterations the engine has completed. Initialized as 0.
+- *`engine.state.output`: The output of the `process_function` defined for the :class:`~ignite.engine.Engine`. See below.
 
 In the code below, `engine.state.output` will store the batch loss. This output is used to print the loss at 
 every iteration.
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -4,7 +4,7 @@ Concepts
 Engine
 ------
 
-The **essence** of the framework is the class :class:`ignite.engine.Engine`, an abstraction that loops a given number of times over
+The **essence** of the framework is the class :class:`~ignite.engine.Engine`, an abstraction that loops a given number of times over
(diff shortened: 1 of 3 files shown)


docs/source/concepts.rst
```diff
@@ -116,9 +130,66 @@ completed iterations and epochs:
 
     trainer.add_event_handler(Events.ITERATION_COMPLETED, on_iteration_completed)
 
+There are many ways of creating a `process_function`, this is why Ignite provides output_transform parameter for its
```
vfdev-5: I would rephrase this as 
```suggestion
Since there is no restrictions on `process_function`, Ignite provides `output_transform` argument for its 
```
@anmolsjoshi what do you think ?

docs/source/concepts.rst
```diff
@@ -116,9 +130,66 @@ completed iterations and epochs:
 
     trainer.add_event_handler(Events.ITERATION_COMPLETED, on_iteration_completed)
 
+There are many ways of creating a `process_function`, this is why Ignite provides output_transform parameter for its 
+:class:`~ignite.metrics` and :class:`~ignite.handlers`. `output_transform` is a function used to transform `engine.state.output` for intended use. Below we'll see different types of `engine.state.output` and how to transform them.
```
vfdev-5:
```suggestion
:class:`~ignite.metrics` and :class:`~ignite.handlers`. Argument `output_transform` is a function used to transform `engine.state.output` for intended use. Below we'll see different types of `engine.state.output` and how to transform them.
```

vfdev-5:
Visdom is [broken](https://github.com/facebookresearch/visdom/issues/548)

docs/source/concepts.rst
anmolsjoshi: It reads better!

docs/source/concepts.rst
anmolsjoshi: Reads better!

anmolsjoshi:
@vfdev-5 thanks! I've incorporated your comments. please review and merge.

PR #400, opened 2019-01-15
Fix examples/mnist/mnist*.py 

ndronen:
Fixes # 2

Description:

The scripts wouldn't even run because

* Someone split `Accuracy` into `BinaryAccuracy` and `CategoricalAccuracy` without updating the client code! 
* On a machine with a GPU, the model wasn't being put to the device, but the inputs were.

Check list:


vfdev-5:
I'll close this PR as proposed modifications are not correct. Please create an issue if you still experience any problem with running mnist examples

PR #416, opened 2019-01-25
Change default argument save_as_state_dict in ModelCheckpoint to True

### LINKED ISSUE #406: ModelCheckpoint parameter `save_as_state_dict` should default to True
Given that the recommended way to save model checkpoints is to only serialize the `state_dict` rather than the whole object (see https://pytorch.org/docs/stable/notes/serialization.html), shouldn't `save_as_state_dict` in `ModelCheckpoint` default to `True`?

Maybe I'm nitpicking but I think this would be a better default.

Here's the line: https://github.com/pytorch/ignite/blob/862ab0073e461a32bf5b53fa015f88c143ae9079/ignite/handlers/checkpoint.py#L84

fabianschilling:
Fixes #406

Description:

ModelCheckpoint should default to saving the `state_dict` rather than the entire object (as is recommended in https://pytorch.org/docs/stable/notes/serialization.html).

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -81,7 +81,7 @@ class ModelCheckpoint(object):
                  n_saved=1,
                  atomic=True, require_empty=True,
                  create_dir=True,
-                 save_as_state_dict=False):
+                 save_as_state_dict=True):
 
         self._dirname = os.path.expanduser(dirname)
         self._fname_prefix = filename_prefix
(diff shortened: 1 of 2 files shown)


vfdev-5:
@fabianschilling could you please adapt the tests and mention in the docsting that by default the argument is True

fabianschilling:
For now, I left the docstring as is since the defaults are not explicitly mentioned for the other arguments. Let me know if I should add it anyways.

vfdev-5:
Thank you @fabianschilling !

PR #417, opened 2019-01-25
Added mean and sum ops to MetricsLambda

### LINKED ISSUE #414: Add helper ops for metrics arithmetics
Following the discussion from [here](https://github.com/pytorch/ignite/pull/410#issuecomment-457105415)

> I was wondering whether it is possible to introduce a metric that performs mean, Mean, and compact the following piece of code:

```python
```
> to:
```python
    metrics = Mean(precision * recall * 2 / (precision + recall + 1e-20))
```

Idea is to provide some helper classes like `Mean`, `Sum` and others.

cc @amitibo

vfdev-5:
Fixes #414  

Description:
Added two basic helper ops `mean` and `sum`. To enable the below code:
```python
f1 = (precision * recall * 2 / (precision + recall + 1e-20)).mean()

custom_metric = (precision + recall).sum()
```
Check list:

### DIFF
--- a/ignite/metrics/metrics_lambda.py
+++ b/ignite/metrics/metrics_lambda.py
@@ -1,3 +1,5 @@
+import torch
+
 from ignite.metrics.metric import Metric
 from ignite.engine import Events
 
@@ -68,3 +70,31 @@ class MetricsLambda(Metric):
         self._internal_attach(engine)
         # attach only handler on EPOCH_COMPLETED
         engine.add_event_handler(Events.EPOCH_COMPLETED, self.completed, name)
+
+    def mean(self):
+        """Helper operation to average the output metric
+
+        Example:
+
+        .. code-block:: python
+
+            precision = Precision(average=False)
+            recall = Recall(average=False)
+            f1 = (precision * recall * 2 / (precision + recall + 1e-20)).mean()
+
+        """
+        return MetricsLambda(lambda t: torch.mean(t).item(), self)
+
+    def sum(self):
+        """Helper operation to sum the output metric
+
+        Example:
+
+        .. code-block:: python
+
+            precision = Precision(average=False)
+            recall = Recall(average=False)
+            custom_metric = (precision + recall).sum()
+
+        """
+        return MetricsLambda(lambda t: torch.sum(t).item(), self)
--- a/tests/ignite/metrics/test_metrics_lambda.py
+++ b/tests/ignite/metrics/test_metrics_lambda.py
@@ -285,3 +285,68 @@ def test_recursive_attachment():
         return 2.0 + 0.2 * (p1 * p2 + p1 - p2) ** 0.5
 
     _test(some_metric, "some metric", compute_true_somemetric)
+
+
+def test_helper_ops():
+
+    def _test(composed_metric, metric_name, compute_true_value_fn):
+
+        metrics = {
+            metric_name: composed_metric,
+        }
+
+        y_pred = torch.rand(15, 10, 5).float()
+        y = torch.randint(0, 5, size=(15, 10)).long()
+
+        def update_fn(engine, batch):
+            y_pred, y = batch
+            return y_pred, y
+
+        validator = Engine(update_fn)
+
+        for name, metric in metrics.items():
+            metric.attach(validator, name)
+
+        def data(y_pred, y):
+            for i in range(y_pred.shape[0]):
+                yield (y_pred[i], y[i])
+
+        d = data(y_pred, y)
+        state = validator.run(d, max_epochs=1)
+
+        assert set(state.metrics.keys()) == set([metric_name, ])
+        np_y_pred = np.argmax(y_pred.numpy(), axis=-1).ravel()
+        np_y = y.numpy().ravel()
+        assert state.metrics[metric_name] == approx(compute_true_value_fn(np_y_pred, np_y))
+
+    precision_1 = Precision(average=False)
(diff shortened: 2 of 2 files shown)


amitibo:
@vfdev-5 I think u should add these methods to Metric. MetricsLambda will inherit it. You can also add dim or axis as arguments to the methods

vfdev-5:
> I think u should add these methods to Metric.

@amitibo I'm not sure that this make sense. A single metric if it needs to be averaged or summed, probably, should have such arguments on initialization. 
Composed metrics are `MetricsLambda` and that's why I added these methods to `MetricsLambda`.
Could you please develop your idea why we need to add these methods to `Metric`.

> You can also add dim or axis as arguments to the methods

Yes, this make sense, I can add.

vfdev-5:
@jasonkriss @alykhantejani what do you think ?

amitibo:
IMHO the problem is remembering what 'average' means, e.g. what is averaged, the samples or labels (in the case of multilabels). It might even be more  confusing in the case of precision as average precision (AP) has a different meaning. Why was the 'average' argument introduced? If the reason was to allow creating complicated MetricsLambda expressions then it might be to set automatically on all MetricsLambda objects except the topmost.

vfdev-5:
I see what you mean about remembering what is averaged: samples or labels. That's why we are trying to keep the docstring up to date. 


> Why was the 'average' argument introduced? 

@amitibo I think the primary reason was to be able to compute precision/recall per class. And secondary I happend such that we can create complicated expressions using precision/recall. 

> If the reason was to allow creating complicated MetricsLambda expressions then it might be to set automatically on all MetricsLambda objects except the topmost.

If you take a look at all available metrics, there are only precision and recall that involves multiclass option and allow to compute values per class. Other metrics are for regression tasks. There are also some accuracy classes. Yes, we can ensure that if ingredient metric has attribute `average` to check if it is `False`. But this is make sense only for `F-beta` and similar metrics. However, probably we could have an expression where `average=True` required... I have no any example for the moment.

amitibo:
There is a delicate balance between adding functionality and keeping things simple to use. I think that in some cases `ignite` is missing this balance. For example I have a feeling that having `Accuracy` handle both the binary, multiclass and multilabel cases implicitly might have made it less intuitive. I had to resort to reading the code in order to understand how to use it. I think that the `average` argument makes it  more confusing.

vfdev-5:
@amitibo your feedback is really appreciated, let's discuss how could we improve it.

If you have access to pytorch slack, we have `pytorch-ignite` channel where we could talk in more fluent manner.

Anyway, what do you think about `average` argument in `sklearn.metrics.precision_score`, is it less confusing ? In our case, what could make it more obvious/transparent ?

amitibo:
The `average` parameter confused me because it sounds like `average precision` (AP). Maybe calling the argument `mean` or `classwise` might have made it clearer. For me at least :-).
Though, I agree that following the standards of `sklearn` is a good advice, so maybe it is better to leave it as it is now.

vfdev-5:
@amitibo yes I agree that we can follow example of `sklearn`.

> Doesn't the mean() method make the average argument redundant? i.e. aren't the following two expressions equivalent?
```python
p1 = Precision(average=True)
p2 = Precision(average=False).mean()
```
Yes, they should be similar.

Actually, from the beginning we decided to introduce `.mean()` as synthetic sugar to easily write some composite metrics. That's why I added `.mean()` to `MetricsLambda`. 
On the other hand, if we are thinking of `.mean()` as something helpful as arithmetic ops and adding it to `Metric` we can have such redundent situations as `Precision(average=False).mean()`.

We need to discuss more on it. @alykhantejani @jasonkriss what do you think ?

alykhantejani:
I'm don't have a strong preference on where/if this should go in but if we do add it to `Metric` then we should still maintain the old API  i.e. the `average` param to `Precision` etc.  but maybe add deprecation warnings?

vfdev-5:
@zasdfgbnm as you implemented metrics arithmetics, i'd like to know what is your opinion on where to implement `mean`, `sum` methods ? I'd appreciate your point of view :)

zasdfgbnm:
I have no idea on whether this should be on Metric or MetricLambda. I didn't see strong pros/cons on either decision.

The most natural way for me is:
```python
import ignite.metrics.functional as F
# F contains all torch operators, not just mean and sum
# these operators works on all metrics, not just metric lambdas
f1 = F.mean(precision * recall * 2 / (precision + recall + 1e-20))
```

possible implementations are:
```python
# ignite/metrics/functional.py
import torch

def __getattr__(self, attr):
    if hasattr(torch, attr):
        fn = getattr(torch, attr)
    elif hasattr(F, attr):
        fn = getattr(F, attr)
    else:
        raise AttributeError('Unknown')
    def wrapper(*args, **kwargs):
        return MetricLambda(fn, *args, **kwargs)
```

zasdfgbnm:
@vfdev-5 My opinion is: if we are planning to support all ops as above, then better on `Metric`, otherwise if we just want to support a few of commonly used operators, then I prefer put them into `MetricLambda`. I don't know why I like this way, but just because this seems "natural" to me. This might be a very personal preference.

vfdev-5:
@zasdfgbnm thank you for your comments ! I agree with your reasoning. For now we can keep these two ops in `MetricsLambda` and move out either in `Metric` or as `functional`. 

To merge this one it remains to add kwargs.

vfdev-5:
Closed in favor of #427

PR #418, opened 2019-01-26
Fix typos and improve docs of Precision/Recall

vfdev-5:
Description: Fix typos and improve documentation of Precision/Recall

Check list:

### DIFF
--- a/ignite/metrics/precision.py
+++ b/ignite/metrics/precision.py
@@ -51,13 +51,8 @@ class Precision(_BasePrecisionRecall):
 
         binary_accuracy = Precision(output_transform=thresholded_output_transform)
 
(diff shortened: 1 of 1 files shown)


amitibo:
@vfdev-5 I might have missed something but I didn't notice anything specifically related to `average` that makes these metrics memory consuming. If I am not wrong, `multilabel` will keep all input data regardless what the value of `average` is.

vfdev-5:
@amitibo yes, you are right, in multilabel case the input data is kept regardless `average`.

anmolsjoshi:
@amitibo #420 has updated implementation for precision and recall such that is_multilabel=True, average=True no longer stores all the data.

anmolsjoshi:
@vfdev-5 LGTM! just need to resolve the merge conflicts from merging of #420.

Issue #424, opened 2019-02-07
Fix CI : bug with python 2.7 opengym


vfdev-5:
Currently, CI is [broken](https://travis-ci.org/pytorch/ignite/builds/490068197), smoke tests of RL examples fail using opengym. Need to investigate.

anmolsjoshi:
@vfdev-5 looked into this, this is an issue with the super call in discrete.py 

This was recently fixed on the master branch using [this PR](https://github.com/openai/gym/pull/1312). I built gym from the master branch and the tests in travis.yml passed. 

One option is to wait till a new release is cut, or we can just build from master. Thoughts?

vfdev-5:
Thanks @anmolsjoshi, maybe we can set now the version of opengym to the previous stable one and once the bug is fixed, remove the version fix.

PR #427, opened 2019-02-08
Metric getattrs

### LINKED ISSUE #414: Add helper ops for metrics arithmetics

zasdfgbnm:
Fixes #414

Description:
As discussed in https://github.com/pytorch/ignite/pull/417#issuecomment-461652286

Check list:


tests/ignite/metrics/test_functional.py
```diff
@@ -0,0 +1,78 @@
...
+
+        assert set(state.metrics.keys()) == set([metric_name, ])
+        np_y_pred = np.argmax(y_pred.numpy(), axis=-1).ravel()
+        np_y = y.numpy().ravel()
+        assert state.metrics[metric_name] == approx(compute_true_value_fn(np_y_pred, np_y))
+
+    precision_1 = Precision(average=False)
+    precision_2 = Precision(average=False)
+    mean_summed_precision1 = F.mean(precision_1 + precision_2).item()
+    mean_summed_precision2 = (precision_1 + precision_2).mean().item()
```
zasdfgbnm: @vfdev-5 Unfortunately the API here are slightly different from those on https://github.com/pytorch/ignite/pull/417

zasdfgbnm:
@vfdev-5 How does this looks like? If the API looks good, then I will update docs.

vfdev-5:
@zasdfgbnm thank you, really awesome code ! 

I'm getting the understanding of what it does and it opens the boundaries and not sure how could we use it in our applications but feels great :) 
For example, chain operations on tensors:
```python
m = (precision_1 + precision_2).pow(2.0).mean()
```
Thinking more about this, I have a slight impression that we "kill a fly with a bazooka" :) 
At the begining, we needed to reduce vectors into scalars and now we have any attribute of torch applied on metrics...

About `ignite.metrics.functional`, I think this is pure private stuff and we could keep inside `Metric` ?

Maybe we can provide some more magic and override somehow the wrapper to produce `.item()` at the end of the operation. Such that we provide needed reduction and do not make all this stuff too complex and error prone.

zasdfgbnm:
@vfdev-5 I changed `Metric.completed` and now `item` is handled automatically. What do you mean keeping functional inside `Metric`? do you mean we just add `__getattr__` for `Metric` and discard the whole `functional.py`? I think we need this, because not all pytorch operators are methods, some of them you need to access it through `torch.xxx(tensor)` but not `tensor.xxx()`

zasdfgbnm:
Seems that module __getattr__ was introduced in [PEP562(]https://www.python.org/dev/peps/pep-0562/), available only on python 3.7...

zasdfgbnm:
I temporarily disabled the two failing RL examples from CI. We could reopen this after https://github.com/pytorch/ignite/issues/424 is resolved.

zasdfgbnm:
@vfdev-5 I updated the document, looks good to myself now.

vfdev-5:
@zasdfgbnm, I'll take a look at the code  in more details asap.

amitibo:
@zasdfgbnm and @vfdev-5, I like this approach very much. I wonder if it enables passing arguments to the functions, .e.g `axis` etc.

zasdfgbnm:
@amitibo It does

vfdev-5:
@zasdfgbnm I checked out the code, it is really cool stuff you proposed here 👍  and I think we will keep the part added into `Metric`. Concerning `import ignite.metrics.functional as F` and `_pep562.py` it is probably too early to introduce that. As it just gives only an equivalent usage and for that we need to create a helper file,  so I propose to remove that and to keep the API as
```
F1 = (precision * recall * 2 / (precision + recall)).mean()
```
without `F.*`. 

What do you think, @zasdfgbnm  @amitibo ?

zasdfgbnm:
@vfdev-5 If `import ignite.metrics.functional as F` was just an alias of metric methods, the I had no preference on having it or not. My only concern is that, there are some operators not accessible using method, for example `torch.meshgrid` has no equivalent `tensor.meshgrid` version. So functional would be a super set of methods.  But I don't know if there will be any user interested in torch operators that are inside functional but not a method.

Anyway, just let me know your decision. I will just follow.

vfdev-5:
@zasdfgbnm that's true, functional part gives more available operators. 
OK, for this PR we can leave it with `Metric.__getattr__` only and if we see any interest into functional part we'll add it after. So, let's adapt the current PR that way.

zasdfgbnm:
@vfdev-5 done

vfdev-5:
@anmolsjoshi review and merge please

PR #432, opened 2019-02-12
Added ParamGroupScheduler

vfdev-5:
Fixes # 
New feature "ParamGroupScheduler" to group multiple param schedulers into a single one.

Description:
```
        names = []
        lr_schedulers = []
        for i, param_group in enumerate(optimizer.param_groups):
            names.append("param_group_{}".format(i))
            lr_schedulers.append(get_lr_scheduler(param_group, param_group_scheduling_conf[i]))
        scheduler = ParamGroupScheduler(schedulers=lr_schedulers, names=names)
        # Attach single scheduler to the trainer
        trainer.add_event_handler(Events.ITERATION_STARTED, scheduler)

```

Check list:


anmolsjoshi:
@vfdev-5 this looks interesting! 

is the purpose of this scheduler to having different optimization for different parameters of a network? any examples of this used in practice?

vfdev-5:
@anmolsjoshi yeah, idea is that you can train the network with different lr schedulings for different parameter groups. For example, a segmentation NN being composed of a pretrained encoder and a decoder can be trained with smaller lr for the encoder part and all groups are trained for example with a warmup + multistep lr schedulling.
The class `ParamGroupScheduler` allows to group separated schedulers into one call.

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -553,6 +553,47 @@ def create_lr_scheduler_with_warmup(lr_scheduler, warmup_start_value, warmup_end
...
+        schedulers (list/tuple of ParamScheduler): list/tuple of parameter schedulers.
+        names (list of str): list of names of schedulers.
+
+    .. code-block:: python
+
+        names = []
+        lr_schedulers = []
+        for i, param_group in enumerate(optimizer.param_groups):
+            names.append("param_group_{}".format(i))
+            lr_schedulers.append(get_lr_scheduler(param_group, param_group_scheduling_conf[i]))
```
anmolsjoshi: Maybe add a comment giving an example of get_lr_scheduler? or just an example of what this line might look like. This can give users a starting point.

vfdev-5: @anmolsjoshi improved the example. Take a look, plz

PR #443, opened 2019-02-21
Adding Exponential Annealing to contrib handlers

zippeurfou:
Fixes # 

Description:
This add exponential annealing to the contrib handler. The goal being to be able to mimic fast.ai learning rate ability. The [LRFinder](https://github.com/fastai/fastai/blob/master/fastai/callbacks/lr_finder.py#L9) [uses](https://github.com/fastai/fastai/blob/master/fastai/callbacks/lr_finder.py#L14) [annealing_exp](https://github.com/fastai/fastai/blob/3ff819262ff0896e4b66febbc2ffabf56e8e95f2/fastai/callback.py#L320) to find the optimal learning rate.
This PR allows to do it. 

Check list:
* [ ] New tests are added (if a new feature is modified) => Not done but could not find any for cosineannealing
* [ ] New doc strings: text and/or example code are in RST format => TODO, since it is a new suggestion I'd like to wait before doing the work

### DIFF
--- a/ignite/contrib/handlers/__init__.py
+++ b/ignite/contrib/handlers/__init__.py
@@ -1,5 +1,6 @@
 
 from ignite.contrib.handlers.param_scheduler import LinearCyclicalScheduler, CosineAnnealingScheduler, \
-    ConcatScheduler, LRScheduler, create_lr_scheduler_with_warmup, PiecewiseLinear
+    ConcatScheduler, LRScheduler, create_lr_scheduler_with_warmup, PiecewiseLinear, \
+    ExponentialAnnealingScheduler
 
 from ignite.contrib.handlers.tqdm_logger import ProgressBar
--- a/ignite/contrib/handlers/param_scheduler.py
+++ b/ignite/contrib/handlers/param_scheduler.py
@@ -207,6 +207,69 @@ class LinearCyclicalScheduler(CyclicalScheduler):
         return self.end_value + (self.start_value - self.end_value) * abs(cycle_progress - 0.5) * 2
 
 
+class ExponentialAnnealingScheduler(CyclicalScheduler):
+    """Anneals 'start_value' to 'end_value' over each cycle.
+
+    Exponentially anneal following what Fast.ai uses for their learning rate scheduler.
+
+    Args:
+        optimizer (`torch.optim.Optimizer` or dict): the optimizer or parameters group to use.
+        param_name (str): name of optimizer's parameter to update.
+        start_value (float): value at start of cycle.
+        end_value (float): value at the end of the cycle.
+        cycle_size (int): length of cycle.
+        cycle_mult (float, optional): ratio by which to change the cycle_size
+            at the end of each cycle (default=1).
+        start_value_mult (float, optional): ratio by which to change the start value at the
+            end of each cycle (default=1.0).
+        end_value_mult (float, optional): ratio by which to change the end value at the
+            end of each cycle (default=1.0).
+        save_history (bool, optional): whether to log the parameter values to
+            `engine.state.param_history`, (default=False).
+
+    Note:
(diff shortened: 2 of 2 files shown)


zippeurfou:
The error from travis seems independent from the code.

vfdev-5:
@zippeurfou can we provide a generic annealing with any scheduler ?

zippeurfou:
Maybe something like a lambda annealing?

vfdev-5:
Sorry, I do not know a lambda annealing. If you are speaking about the implementation we can do it probably like:
https://github.com/pytorch/ignite/blob/e8bdf67312a36f77d275f652e139ea0a0481329b/ignite/contrib/handlers/param_scheduler.py#L489

zippeurfou:
I am sorry I am unsure to understand the question/problem.
Are you asking if the exponential annealing can be used with the concat scheduler?

vfdev-5:
Actually, yes you are right, I was speaking about a sort of lambda annealing, and exp annealing is a special case.

zippeurfou:
Sadly, looking at the formulas I don't think we can really make something generic.
We could potentially make a function that is called Annealing that have a parameter that can take "cos" or "exp" and build on top of it when we want to add more annealing.
Looking at [fast.ai](https://github.com/fastai/fastai/blob/3ff819262ff0896e4b66febbc2ffabf56e8e95f2/fastai/callback.py#L320) the other one they have is poly.
However, I am unsure this would really be a "clean" approach..

vfdev-5:
@zippeurfou I tried to prototype such class quickly and it's true that this can be more complex to support all type of schedulers...

Idea is something like that:
```python
class LambdaAnnealing(ParamScheduler):
    
    
    def __init__(self, scheduler, cycle_size, cycle_mult=1.0, save_history=False):
        super(LambdaAnnealing, self).__init__(optimizer={}, param_name="", save_history=save_history)        
        self.scheduler = scheduler 
        self.optimizer_param_groups = self.scheduler.optimizer_param_groups
        self.param_name = self.scheduler.param_name
        
        self.cycle_size = cycle_size
        self.cycle_mult = cycle_mult
        self.cycle = 0

    def get_param(self):
        
        if self.event_index > 0 and (self.event_index % self.cycle_size) == 0:
            # Restart wrapped scheduler
            # !!! THIS WONT WORK FOR EVERY SCHEDULER !!!
            self.scheduler.event_index = 0
            self.cycle_size *= self.cycle_mult
            self.cycle += 1
        
        value = self.scheduler.get_param()
        # Maybe we could scale the value here
        return value        
```

zippeurfou:
How can I help with this one?
I wonder if after this one I should do a lr_finder as a contrib. thought?

vfdev-5:
> How can I help with this one?

@zippeurfou you can go on with the prototype class `LambdaAnnealing` I provided in the previous message. However, this will be tricky.
Here is a code to check the implementation:
```python
scheduler = PiecewiseLinear(optimizer, "lr",
                            milestones_values=[(0, 0.0), (10, 0.4), (29, 0.0)])

la_scheduler = LambdaAnnealing(scheduler, cycle_size=25)

lrs = []

def save_lr(engine):
    lrs.append(optimizer.param_groups[0]['lr'])

trainer = Engine(lambda engine, batch: None)
trainer.add_event_handler(Events.ITERATION_STARTED, la_scheduler)
trainer.add_event_handler(Events.ITERATION_COMPLETED, save_lr)
trainer.run([0] * 10, max_epochs=5)


import numpy as np
import matplotlib.pylab as plt

plt.title("Lambda annealing")
plt.plot(lrs, label="learning rate")
plt.xlabel("events")
plt.ylabel("values")
plt.legend()
```

After that we need also the tests to write.

> I wonder if after this one I should do a lr_finder as a contrib. thought?

I have no idea of any usefullness of this `lr_finder`, as everybody I read about it at fast.ai but never tested explicitly. If you have more experience with that we can follow your advice.

PR #444, opened 2019-02-21
Allow duration ConcatScheduler to be from numpy fixes #442

### LINKED ISSUE #442: Allow the duration parameter to be a np.int in contrib.handler
I had some issue with concat scheduler using numpy integer for durations instead of int:
image 2019-02-20 at 1 54 36 pm
this seems pretty restrictive when you could just do something along theses lines:
```
all([isinstance(t, (int,np.integer)) for t in durations])
```
Happy to submit a PR if you think it makes sense.

zippeurfou:
Fixes #442

Description:

This PR allows number in concat scheduler to be from numpy which might be useful when you use it to calculate floor/ceiling to relate to your cycle length.

### DIFF
--- a/ignite/contrib/handlers/param_scheduler.py
+++ b/ignite/contrib/handlers/param_scheduler.py
@@ -3,6 +3,7 @@ from __future__ import division
 from copy import deepcopy
 
 import math
+import numbers
 
 from abc import ABCMeta, abstractmethod
 from ignite._six import with_metaclass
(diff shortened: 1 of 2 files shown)


vfdev-5:
Thanks @zippeurfou ! How about other checks `isinstance(xx, int)`, for example:
https://github.com/pytorch/ignite/blob/037e2ecccf63554bea5a6cbc41765fe91046dae7/ignite/contrib/handlers/param_scheduler.py#L598

vfdev-5:
@zippeurfou how about my previous message ?

zippeurfou:
Sorry been pretty busy. Just changed it.

vfdev-5:
Actually, we also need to add some tests that fail without the fix and pass with the fix.

vfdev-5:
@zippeurfou would you please write the tests as asked in the previous message ?

PR #449, opened 2019-03-01
Added yet another tensorboard logger

vfdev-5:
Yet another attempt to provide an API for TensorBoard internal logger.

- Simplier API
- Customisable logging

```python
    from ignite.contrib.handlers.tensorboard_logger import *

    tb_logger = TensorboardLogger(log_dir=log_dir)

    tb_logger.attach(trainer,
                     log_handler=output_handler(tag="training", output_transform=lambda loss: {'loss': loss}),
                     event_name=Events.ITERATION_COMPLETED)

    tb_logger.attach(train_evaluator,
                     log_handler=output_handler(tag="training",
                                                metric_names=["loss", "accuracy"],
                                                another_engine=trainer),
                     event_name=Events.EPOCH_COMPLETED)

    tb_logger.attach(validation_evaluator,
                     log_handler=output_handler(tag="validation",
                                                metric_names=["loss", "accuracy"],
                                                another_engine=trainer),
                     event_name=Events.EPOCH_COMPLETED)

    tb_logger.attach(trainer,
                     log_handler=optimizer_params_handler(optimizer),
                     event_name=Events.ITERATION_COMPLETED)

    x, _ = next(iter(train_loader))
    tb_logger.log_graph(model, x.to(device))

    tb_logger.attach(trainer,
                     log_handler=weights_scalar_handler(model),
                     event_name=Events.ITERATION_COMPLETED)

    tb_logger.attach(trainer,
                     log_handler=weights_hist_handler(model),
                     event_name=Events.EPOCH_COMPLETED)

    tb_logger.attach(trainer,
                     log_handler=grads_scalar_handler(model),
                     event_name=Events.ITERATION_COMPLETED)

    tb_logger.attach(trainer,
                     log_handler=grads_hist_handler(model),
                     event_name=Events.EPOCH_COMPLETED)
```

cc @anmolsjoshi , @amitibo any comments please

Check list:


ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -0,0 +1,227 @@
...
+        Args:
+            model (torch.nn.Module): input model
+            input_to_model (torch.Tensor or torch.utils.data.DataLoader): input to the model
+            **kwargs: kwargs to `tensorboardX.SummaryWriter.add_graph`
+
+        """
+        from torch.utils.data import DataLoader
+
+        if isinstance(input_to_model, DataLoader):
+            input_to_model, _ = next(iter(input_to_model))
```
Bibonaut: Here is my comment copied from the *old* tensorboard_logger (#265). I just realized that this new logger exists 👍:
Assuming that seeds are set manually at the beginning of an experiment, this line of code can break reproducibility. Running the same experiment with with and without logging the graph can result in different batch data passed to the engines `process_function` because the dataloader could be called once here. Perhaps something like this works:
```
x, y = input_to_model.collate_fn(input_to_model.dataset.__getitem__(0))
```

vfdev-5: @Bibonaut yes you're right about the batches. Concerning your proposition, probably should be something like:
```
x, _ = input_to_model.collate_fn([input_to_model.dataset[0], ])
```
Honestly, I think we should remove such option to take a dataloader as an input...

anmolsjoshi: What if we take input shape as a parameter instead of the dataloader? We can create a tensor of zeros to pass through the model

vfdev-5: This would work if we have a model working directly on tensors, but it could have more than one entry. I think we need to remove this method `log_graph` as it adds nothing more special to `tb_logger.writer.add_graph(model, inputs)`

vfdev-5:
I'm closing this in favor of #455

PR #450, opened 2019-03-03
Increase early stopping counter on plateau

maaario:
Description:
In the current implementation, the early stopping never terminates training if the score function returns the same value (e.g. `loss == 0`). With the new fix, if the score plateaus, the counter is increased and the training is eventually terminated.

### DIFF
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -55,7 +55,7 @@ class EarlyStopping(object):
 
         if self.best_score is None:
             self.best_score = score
-        elif score < self.best_score:
+        elif score <= self.best_score:
             self.counter += 1
             self._logger.debug("EarlyStopping: %i / %i" % (self.counter, self.patience))
             if self.counter >= self.patience:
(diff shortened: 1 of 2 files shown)


maaario:
The `loss == 0` situation is maybe particular and I wrote it just to give a simple example. To give one more example, in my case I tried to use this early stopping to optimize a grid-search hyperparmeter selection. Since I also used [ReduceLROnPlateau](https://pytorch.org/docs/stable/optim.html#torch.optim.lr_scheduler.ReduceLROnPlateau) scheduler, when the parameter combination was not great, the learning rate gradually decreased and the learning virtually stopped. One possible solution would be to change `eps` parameter of the schduler, but the early stopping approach seemed to me as more adaptive solution and I hoped it would stop the training.

I will add the test with constant as the score function.

vfdev-5:
Yes, that makes sense. Okay, let's go this path. So, please provide a test and we will merge it.

maaario:
Ok, the tests (I added 2 tests to follow the pattern of the test file = simple test + test with engine) are passing, but on python-3.6 CI spuriously failed on `conda create envirinment`. Can you re-run the checks or is it even ok this way? I think this should be all from my side.

vfdev-5:
@maaario thanks for the tests. Yes, travis fails sometimes like that. I restarted the test.

maaario:
Happy to help :)

PR #455, opened 2019-03-18
Generic logger: TensorboardLogger and VisdomLogger

vfdev-5:
Description: Added a logger generic implementation and two specializations:
- TensorboardLogger
- VisdomLogger

Check list:

cc @amitibo


vfdev-5:
- VisdomLogger
  - training/validation curves
  - model weights/gradients on custom events

Screen Shot 2019-03-19 at 00 03 28

vfdev-5:
This can be a starting base for the features proposed in https://github.com/pytorch/ignite/issues/260#issuecomment-419795837

amitibo:
@vfdev-5 very impressive work!
Regretfully in my new workplace we use a different deep-learning framework therefore currently I don't have a way to test pytorch code.
One issue I can comment about: from my experience the overhead of communicating with the visdom server becomes high when using extensive logging, especially when the logging is done to a remote server like when running on a cluster. In solved this by incorporating multi-threading:
https://github.com/amitibo/ignite/blob/private/ignite/contrib/handlers/visdom_logger.py

anmolsjoshi:
@vfdev-5 this looks amazing! will test locally on a few different problems. will report back.

vfdev-5:
@amitibo @anmolsjoshi thanks for the feedback !

> Regretfully in my new workplace we use a different deep-learning framework therefore currently I don't have a way to test pytorch code.

This is said :( 
What is the framework you are using now ?

> One issue I can comment about: from my experience the overhead of communicating with the visdom server becomes high when using extensive logging, especially when the logging is done to a remote server like when running on a cluster.

@amitibo yes I had the same problem with visdom and handled it by reducing the number of call. Thanks for the link, yes I saw your code using `ThreadPoolExecutor`, maybe we could integrate this too. I wonder whether this will be safe regarding DataLoader etc... Let me experiment with that.

vfdev-5:
Thinking more on the solution of @amitibo using the ThreadPoolExecutor, the only drawback is that we need to install `futures` when using python 2.7. 
As python2 support is going down in future, maybe we can allow to force user to install the package instead rewriting things.

anmolsjoshi:
@vfdev-5 I agree on forcing the user to install `futures`, would it be as simple as adding the requirement [here](https://github.com/pytorch/ignite/blob/master/setup.py#L25)? 

I have reviewed the code, and used it on a few examples of my own, results look great! Also, great job on the tests.

vfdev-5:
@anmolsjoshi actually thinking more about this, maybe we could replace `ThreadPoolExecutor` by `multiprocessing.Pool/Queue` ? Such that we do not need to manage additional dependency...

anmolsjoshi:
That's a good idea! Adding another dependency might not be a good idea, especially with support for python 2.7 being removed in the future

vfdev-5:
I improved the docs/docstring and tests for Visdom, integrated ThreadPoolExecuter (as done by @amitibo) as multiprocessing failed to pickle visdom instance...

Here is how the docs look like now:
Screen Shot 2019-04-01 at 17 13 17


@anmolsjoshi if OK for you we can merge it

vfdev-5:
Anyway, I added this possibility to use a logger as a context manager.

anmolsjoshi:
@vfdev-5 thanks for the updates, going to test it out on a few experiments. Will review tonight!

PR #476, opened 2019-04-04
Remove extra memory usage in batch processing.

IlyaOvodov:
Description:

Output of previous process_function call was not deleted yet when process_function is called next time and returns new output.
If output use much memory (say, in segmentation tasks), it results in 'out of memory' error for batch size smaller then it could be if previous output is released before new process_function call.

### DIFF
--- a/ignite/engine/__init__.py
+++ b/ignite/engine/__init__.py
@@ -76,6 +76,7 @@ def create_supervised_evaluator(model, metrics={},
         model.to(device)
 
     def _inference(engine, batch):
+        engine.state.output = None
         model.eval()
         with torch.no_grad():
             x, y = prepare_batch(batch, device=device, non_blocking=non_blocking)
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -302,7 +302,6 @@ class Engine(object):
                 self.state.batch = batch
                 self.state.iteration += 1
                 self._fire_event(Events.ITERATION_STARTED)
-                self.state.output = None
                 self.state.output = self._process_function(self, batch)
                 self._fire_event(Events.ITERATION_COMPLETED)
                 if self.should_terminate or self.should_terminate_single_epoch:
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -302,6 +302,7 @@ class Engine(object):
                 self.state.batch = batch
                 self.state.iteration += 1
                 self._fire_event(Events.ITERATION_STARTED)
+                self.state.output = None
                 self.state.output = self._process_function(self, batch)
                 self._fire_event(Events.ITERATION_COMPLETED)
                 if self.should_terminate or self.should_terminate_single_epoch:


vfdev-5:
@IlyaOvodov thanks for the PR. Seems reasonable, but let me recheck this. I was also thinking whether we can cover this with some testing...

Some more details on `out of memory` error, does it happens when output is on GPU and contains probable `(y_pred, y)` ?
Probably, we need to perform also  `gc.collect(); torch.cuda.empty_cache()` ?

IlyaOvodov:
@vfdev-5 , you're right. I used my own Engine class which pleces in output not only loss but (y_pred, y) also. And these tensors were destroyed only after process function call, while assigning its result to state.output. After my modification CUDA error memory disappeared. No gc.collect() or torch.cuda.empty_cache() are required. But I consider that in normal trainer engine (that returns only loss) it is not a probem, but fix may cause problems if some process_function really expects previous output available. So it is better to fix the problem in a process_function itself.

vfdev-5:
@IlyaOvodov I see what you mean.

I agree that between training and validation we would like to clean the memory. And in this case we can add a handler like this : 
```python
    def empty_cuda_cache(engine):
        engine.state.output = None
        import gc
        gc.collect()
        torch.cuda.empty_cache()

    trainer.add_event_handler(Events.EPOCH_COMPLETED, empty_cuda_cache)
```
What do you think ?

ignite/engine/__init__.py
```diff
@@ -76,6 +76,7 @@ def create_supervised_evaluator(model, metrics={},
         model.to(device)
 
     def _inference(engine, batch):
+        engine.state.output = None
```
vfdev-5: @IlyaOvodov here `engine` is evaluator and not the trainer. Is it OK ?

IlyaOvodov: Yes, see my message above. Traines output is just loss value and it does not produse such a problem. It's a pity that standard trainer does not return pred_y and y, it forces to use customizes one. But I don't see a way to change it without pain for existing users :)

IlyaOvodov:
@vfdev-5, my point is that previous output (if it is large) eats memory not only between train and validation but also from iteration to iteration. So cleaning it on EPOCH_COMPLETED does not fix it. 
You option can help to clean some memory for validation, but usually validation is called also on EPOCH_COMPLETED, so it becames importent in what order handlers are attached to event. It does not seem to be good.
On the other hand, I have doubts that gc.collect() and torch.cuda.empty_cache() are needed, because they influence only memory accessability for other applications. Inside the application engine.state.output = None seems to be enougth.

ignite/engine/__init__.py
vfdev-5: I see. Maybe we can introduce an optional `output_transform` which by default does 
```python
def default_output_transform(x, y, y_pred, loss):
    return loss.item()
```
And user could override the function...

vfdev-5:
@IlyaOvodov I see your point. So, yes, there are two situations where we would like to clean `engine.state.output` between epochs as we can potentially execute a validation (optional clean) and between iterations (any engine: trainer/evaluator) if `engine.state.output` can contain consuming data.
Okay, let me check all this from my side.

ignite/engine/__init__.py
IlyaOvodov: @vfdev-5 ... and pass it as parameter to create_supervised_trainer? Sounds good...
But I think make 2 functions `output_loss` and `output_y_pred_y` and:
```python
def create_supervised_trainer(... , output_transform = output_loss)
def create_supervised_evaluator(... , output_transform = output_y_pred_y)
```
- it is more meaningful, IMHO

vfdev-5:
@IlyaOvodov I confirm on my side the problem you raised. Here is a snippet to reproduce
```python
import torch

data = [0, 1, 2]

def update_fn(engine, batch):

    y_pred = torch.rand(16, 50, 256, 256).to('cuda')
    y = torch.randint(0, 50, size=(16, 256, 256)).to('cuda')
    cuda_mem = torch.cuda.memory_allocated()
    print("inside update_fn: GPU mem {} MB".format(cuda_mem / 1024 /1024))    
    engine.state.cuda_mem_consumption.append(cuda_mem)
    return y_pred, y

engine = Engine(update_fn)

@engine.on(Events.STARTED)
def initialize(engine):
    engine.state.cuda_mem_consumption = []
    torch.cuda.empty_cache()


@engine.on(Events.ITERATION_STARTED)
def log_cuda_mem(engine):
    cuda_mem = torch.cuda.memory_allocated()
    print("outside update_fn: GPU mem {} MB".format(cuda_mem / 1024 /1024))
    engine.state.cuda_mem_consumption.append(cuda_mem)

    
@engine.on(Events.ITERATION_STARTED)
def clear_state_output(engine):
    engine.state.output = None


print(state.cuda_mem_consumption)
```

If `clear_state_output` is not attached, we have
```
outside update_fn: GPU mem 0.0 MB
inside update_fn: GPU mem 208.0 MB
outside update_fn: GPU mem 208.0 MB
inside update_fn: GPU mem 416.0 MB
outside update_fn: GPU mem 208.0 MB
inside update_fn: GPU mem 416.0 MB
outside update_fn: GPU mem 208.0 MB
inside update_fn: GPU mem 416.0 MB
outside update_fn: GPU mem 208.0 MB
inside update_fn: GPU mem 416.0 MB
outside update_fn: GPU mem 208.0 MB
inside update_fn: GPU mem 416.0 MB
[0, 218103808, 218103808, 436207616, 218103808, 436207616, 218103808, 436207616, 218103808, 436207616, 218103808, 436207616]
```
and with clearing output 
```
outside update_fn: GPU mem 208.0 MB
inside update_fn: GPU mem 416.0 MB
outside update_fn: GPU mem 416.0 MB
inside update_fn: GPU mem 416.0 MB
outside update_fn: GPU mem 416.0 MB
inside update_fn: GPU mem 416.0 MB
outside update_fn: GPU mem 416.0 MB
inside update_fn: GPU mem 416.0 MB
outside update_fn: GPU mem 416.0 MB
inside update_fn: GPU mem 416.0 MB
outside update_fn: GPU mem 416.0 MB
inside update_fn: GPU mem 416.0 MB
[218103808, 436207616, 436207616, 436207616, 436207616, 436207616, 436207616, 436207616, 436207616, 436207616, 436207616, 436207616]
```

To fix this I thought about attaching internally `clear_state_output` handler:
```python
class Engine
    def __init__(self, process_function):

        def clear_state_output(engine):
            engine.state.output = None
        
        self.add_event_handler(Events.ITERATION_STARTED, clear_state_output)  
```

and adding a test (similar on my snippet) on GPU with the decorator:
```python
def test_clear_state_output():
```

What do you think ?

IlyaOvodov:
@vfdev-5 , I don't think that clearing output at event handler is a good idea. My point is that a) it should be possible to turn this behaviour on/off if needed b) it should be clear what behaviour is default c) I'm in doubts can we make output=None as default behaviour (it seems more reasonable but can break somebody's code). When using event handler is not clear enougth is it on or off and how to manage it. If you attach this handler automatically in constructor, if will run before any other ITERATION_STARTED events and output will not be available there, i is not good. It is not clear also that this thing is done at all and removing this handler is not straightforward task.

IlyaOvodov:
I consider more preferable 2 options:
a) do output=None in Engine code just before process_function call and switch it on/off by parameter, passed to Engine constructor (default = ?)
b) do nothing and possibly just document this aspect someware. At least, currnet situation is normal for pyton and if assignment is memory critical, extra means should be done by hands. Normally output is just loss.item(), and if somebody change it and is going to keep huge images over iteration, it's up to him. For inference (evaluation), where output = (pred_y,y), it is fixed in this PR

vfdev-5:
@IlyaOvodov thanks for comments! Yes, I agree that we can break someone's code with this. Maybe in LSTM folks could store hidden state results in the output...

Maybe we can reopen this PR and update the documentation as you suggested in b) ?

IlyaOvodov:
I closed id by mistake.
By the way, tha same problem is with `for batch in self.state.dataloader: ...`, if batch is large: the previous one is released only after new one is created. And batch is also stored in engine.state...

vfdev-5:
> And batch is also stored in engine.state...

Maybe we can remove it.

vfdev-5:
> By the way, tha same problem is with for batch in self.state.dataloader: ...

Actually, batch is not on GPU yet so we just double RAM... 
Batch is converted to GPU with `prepare_batch` method

vfdev-5:
This PR can be closed as we implemented a similar thing here: https://github.com/pytorch/ignite/pull/1919

PR #478, opened 2019-04-04
Fix #477: output_transform parameter added to create_supervised... functions

### LINKED ISSUE #477: Improve create_supervised_trainer with optional output_transform
Following [the discussion](https://github.com/pytorch/ignite/pull/476#discussion_r272108999), idea is to give more flexibility to users who are using `create_supervised_trainer`:
```python
def default_output_transform(x, y, y_pred, loss):
    return loss.item() 


def create_supervised_trainer(model, optimizer, loss_fn,
                              device=None, non_blocking=False, prepare_batch=_prepare_batch, 
                              output_transform=default_output_transform):
    if device:
        model.to(device)

    def _update(engine, batch):
        model.train()
        optimizer.zero_grad()
        y_pred = model(x)
        loss = loss_fn(y_pred, y)
        loss.backward()
        optimizer.step()

    return Engine(_update)    
```

cc @IlyaOvodov

IlyaOvodov:
'output_transform' parameter added to create_supervised_trainer(...) and create_supervised_evaluator(...) functions.

It can be used to change value being assigned to engine.state.output
For instance, it can be changed from loss.item() to (y, pred_y) if one need to attach metrics to training process.

Fixes #477

### DIFF
--- a/ignite/engine/__init__.py
+++ b/ignite/engine/__init__.py
@@ -15,7 +15,8 @@ def _prepare_batch(batch, device=None, non_blocking=False):
 
 def create_supervised_trainer(model, optimizer, loss_fn,
                               device=None, non_blocking=False,
-                              prepare_batch=_prepare_batch):
+                              prepare_batch=_prepare_batch,
+                              output_transform = lambda x, y, y_pred, loss: loss.item() ):
     """
     Factory function for creating a trainer for supervised models.
 
@@ -29,6 +30,10 @@ def create_supervised_trainer(model, optimizer, loss_fn,
             with respect to the host. For other cases, this argument has no effect.
         prepare_batch (callable, optional): function that receives `batch`, `device`, `non_blocking` and outputs
             tuple of tensors `(batch_x, batch_y)`.
+        output_transform (callable, optional): function that receives 'x', 'y', 'y_pred', 'loss' and returns value
+            to be assigned to engine's state.output after each iteration. Default is returning loss.item(). If you
+            are going to attach metrics to trainer, you should pass
+            'output_transform = lambda x, y, y_pred, loss: (y_pred, y,)'
 
     Note: `engine.state.output` for this engine is the loss of the processed batch.
 
@@ -46,14 +51,15 @@ def create_supervised_trainer(model, optimizer, loss_fn,
         loss = loss_fn(y_pred, y)
         loss.backward()
         optimizer.step()
-        return loss.item()
+        return output_transform(x, y, y_pred, loss)
 
     return Engine(_update)
 
 
 def create_supervised_evaluator(model, metrics={},
                                 device=None, non_blocking=False,
-                                prepare_batch=_prepare_batch):
+                                prepare_batch=_prepare_batch,
+                                output_transform = lambda x, y, y_pred: (y_pred, y,) ):
     """
     Factory function for creating an evaluator for supervised models.
 
@@ -66,6 +72,8 @@ def create_supervised_evaluator(model, metrics={},
             with respect to the host. For other cases, this argument has no effect.
         prepare_batch (callable, optional): function that receives `batch`, `device`, `non_blocking` and outputs
             tuple of tensors `(batch_x, batch_y)`.
+        output_transform (callable, optional): function that receives 'x', 'y', 'y_pred' and returns value
+            to be assigned to engine's state.output after each iteration. Default is returning '(y_pred, y,)'
 
     Note: `engine.state.output` for this engine is a tuple of `(batch_pred, batch_y)`.
 
@@ -80,7 +88,7 @@ def create_supervised_evaluator(model, metrics={},
         with torch.no_grad():
             x, y = prepare_batch(batch, device=device, non_blocking=non_blocking)
             y_pred = model(x)
-            return y_pred, y
+            return output_transform(x, y, y_pred)
 
     engine = Engine(_inference)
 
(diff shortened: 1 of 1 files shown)


ignite/engine/__init__.py
```diff
@@ -29,6 +30,10 @@ def create_supervised_trainer(model, optimizer, loss_fn,
             with respect to the host. For other cases, this argument has no effect.
         prepare_batch (callable, optional): function that receives `batch`, `device`, `non_blocking` and outputs
             tuple of tensors `(batch_x, batch_y)`.
+        output_transform (callable, optional): function that receives 'x', 'y', 'y_pred', 'loss' and returns value
+            to be assigned to engine's state.output after each iteration. Default is returning loss.item(). If you
```
vfdev-5:
```suggestion
            to be assigned to engine's state.output after each iteration. Default is returning `loss.item()`. If you
```

ignite/engine/__init__.py
```diff
@@ -29,6 +30,10 @@ def create_supervised_trainer(model, optimizer, loss_fn,
...
+            are going to attach metrics to trainer, you should pass
```
vfdev-5: nit: I would move "If you are going to attach metrics to trainer, you should pass `output_transform = lambda x, y, y_pred, loss: (y_pred, y,)`" to `create_supervised_evaluator`. We are not intended to attach metrics to the trainer as metrics as somewhat running averages of running model.

ignite/engine/__init__.py
```diff
@@ -66,6 +72,8 @@ def create_supervised_evaluator(model, metrics={},
             with respect to the host. For other cases, this argument has no effect.
         prepare_batch (callable, optional): function that receives `batch`, `device`, `non_blocking` and outputs
             tuple of tensors `(batch_x, batch_y)`.
+        output_transform (callable, optional): function that receives 'x', 'y', 'y_pred' and returns value
+            to be assigned to engine's state.output after each iteration. Default is returning '(y_pred, y,)'
```
vfdev-5:
```suggestion
            to be assigned to engine's state.output after each iteration. Default is returning `(y_pred, y,)`.
```

ignite/engine/__init__.py
```diff
@@ -66,6 +72,8 @@ def create_supervised_evaluator(model, metrics={},
...
 
     Note: `engine.state.output` for this engine is a tuple of `(batch_pred, batch_y)`.
```
vfdev-5: Probably we need to remove this line

ignite/engine/__init__.py
IlyaOvodov: No need to move it to create_supervised_evaluator because it's already OK there by default.
Changing output to (y_pred, y,) is required here and main motivation for it is attaching metrics to trainer. If loss calculation is time consuming, calculating metrics on train set while trainig is better choice then calculating them using evaluator because it means repeated run on the train set. Although the second option is more precise of cause, it's cost can be too high.

ignite/engine/__init__.py
IlyaOvodov: Do you mean  line
Note: `engine.state.output` for this engine is a tuple of `(batch_pred, batch_y)`.
Yes. But better to update it.

ignite/engine/__init__.py
vfdev-5: @IlyaOvodov we do not encourage attaching metrics to the trainer : https://pytorch.org/ignite/master/quickstart.html#id1 
So, please remove this phrase. 
And in case of `create_supervised_evaluator` we can just recall that to compute metrics we need to return also `y_pred` and `y`.

ignite/engine/__init__.py
vfdev-5: Yes, please update it

ignite/engine/__init__.py
IlyaOvodov: Done. Also added comment on 'create_supervised_evaluator'.

ignite/engine/__init__.py
```diff
@@ -66,8 +71,12 @@ def create_supervised_evaluator(model, metrics={},
             with respect to the host. For other cases, this argument has no effect.
         prepare_batch (callable, optional): function that receives `batch`, `device`, `non_blocking` and outputs
             tuple of tensors `(batch_x, batch_y)`.
+        output_transform (callable, optional): function that receives 'x', 'y', 'y_pred' and returns value
+            to be assigned to engine's state.output after each iteration. Default is returning `(y_pred, y,)` with fits
```
vfdev-5:
```suggestion
            to be assigned to engine's state.output after each iteration. Default is returning `(y_pred, y,)` which fits
```

ignite/engine/__init__.py
```diff
@@ -66,8 +71,12 @@ def create_supervised_evaluator(model, metrics={},
...
+            output expected by metrics. If you change it you should use `output_transform` in metrcis.
```
vfdev-5:
```suggestion
            output expected by metrics. If you change it you should use `output_transform` in metrics.
```

ignite/engine/__init__.py
IlyaOvodov: By the way. What if to change default output_transform for metrics from lambda x: x to lambda x: x[:2]?
As metrics now experct tuple (y_pred, y) it will not change current functioning but extend it: tuple of (y_pred, y, somth else,...) will become acceptable, so adding some information to state.output using means of this PR will not require using output_transform in metrics.

vfdev-5: Good idea, maybe something like `lambda x: x[:2] if len(x) > 2 else x` just to ensure

vfdev-5: I think in v0.3.0 we should simplify this stuff and write a dict like `{'y_pred': y_pred, 'y': y}` to the `state.output`

IlyaOvodov: :) 
Need to think about it. From one point of view it's good (I started with it), but somebody can use some custom metric inherited fromMetric class and expect to receive that output that he returned from process function, proposed sofistication will be suurprise...

PR #481, opened 2019-04-05
New metric ConfusionMatrix, IoU, mean IoU 

vfdev-5:
Description:
In this PR new metrics : ConfusionMatrix, IoU, mIoU etc are proposed into core module.

Check list:


TheCodez:
@vfdev-5 do I have to attach ConfusionMatrix or mIoU to the trainer?

vfdev-5:
@TheCodez this should work as ordinary metric, I guess :)

TheCodez:
@vfdev-5 thanks, so this usage would be correct?

```python
cm = ConfusionMatrix(num_classes=num_classes, output_transform=lambda x: x['semantics'])
mIoU(cm, ignore_index=0).attach(train_evaluator, 'mIoU')
```

vfdev-5:
@TheCodez yep. Tell me if it does not work as intended :)

PR #488, opened 2019-04-11
Add `epoch_bound` parameter to `RunningAverage`.

Evpok:
This optional parameter controls whether the running average should be reset after each epoch.

Check list:

### DIFF
--- a/ignite/metrics/running_average.py
+++ b/ignite/metrics/running_average.py
@@ -11,6 +11,8 @@ class RunningAverage(Metric):
         alpha (float, optional): running average decay factor, default 0.98
         output_transform (callable, optional): a function to use to transform the output if `src` is None and
             corresponds the output of process function. Otherwise it should be None.
+        epoch_bound (boolean, optional): whether the running average should be reset after each epoch (defaults
+            to True).
 
     Examples:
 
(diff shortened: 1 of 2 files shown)


ignite/metrics/running_average.py
```diff
@@ -67,8 +70,9 @@ def compute(self):
         return self._value
 
     def attach(self, engine, name):
-        # restart average every epoch
-        engine.add_event_handler(Events.EPOCH_STARTED, self.started)
+        if self.epoch_bound:
+            # restart average every epoch
+            engine.add_event_handler(Events.EPOCH_STARTED, self.started)
```
vfdev-5: I think we need to call started at least once. So you can define the event for `self.started` handlers as 
```python
event = Events.STARTED if self.epoch_bound else Events.EPOCH_STARTED
engine.add_event_handler(event, self.started)
```
What do you think ?

vfdev-5:
@Evpok as discussed at slack, yes, I think one integration test is needed.

I also thinking about, shouldn't `epoch_bound` be an argument of `attach` instead of `__init__` ?

Evpok:
Yes, it does make more sense

Evpok:
Are other metrics with additional parameters to `attach` though @vfdev-5 ? Not sure how to document it :-)

vfdev-5:
@Evpok thinking more on that, maybe we can leave `epoch_bound` argument in constructor such that it defines the way to compute a metric for any engine.

If you can add an integration test, it would awesome :)

Evpok:
Right, working on it

vfdev-5:
@Evpok could you please fix flake8 error

PR #493, opened 2019-04-12
[WIP] Dali Engine

### LINKED ISSUE #447: Ways or Features to improve data loading performance
I was wondering if there is any planned feature to improve data loading performance. The standard Pytorch Dataset and DataLoader implementation rely on random I/O, which can be bottleneck for image classification for small input, even with a M2 SSD. It's possible to convert the dataset into a binary data file like LMDB but I was wondering if there is a standard or perhaps better way in Pytorch, ideally supported out-of-box by high level training framework like Ignite.

chicham:
Fixes #447 

Description:
- Add a new engine based on nvidia-dali:
- Add a new metric for distributed computing
- Example script of distributed training with nvidia-dali

Check list:

### DIFF
--- a/examples/contrib/dali/README.md
+++ b/examples/contrib/dali/README.md
@@ -1,0 +1,54 @@
+# Basic Multi-processing Distributed Data Parallel Training with Dali
+
+
+## Requirements:
+
+- [nvidia-dali](https://github.com/NVIDIA/DALI): see the [README](https://github.com/NVIDIA/DALI/blob/master/README.rst)
+- [torchvision](https://github.com/pytorch/vision/): `pip install torchvision`
+- [tqdm](https://github.com/tqdm/tqdm/): `pip install tqdm`
+
+
+## Single node, multiple GPUs:
+
+Run the example:
+```
+ python -m torch.distributed.launch --nproc_per_node=N_GPUS imagefolder_with_dali.py --root PATH/TO/DATASET --model MODEL_NAME --epochs EPOCHS --batch_size SIZE
+```
+
+## Usage
+
+```
+usage: imagefolder_with_dali.py [-h] [--root ROOT] [--model MODEL]
+                                [--batch_size BATCH_SIZE]
+                                [--val_batch_size VAL_BATCH_SIZE]
+                                [--epochs EPOCHS] [--lr LR]
+                                [--momentum MOMENTUM] [--val_ratio VAL_RATIO]
+                                [--requires_grad REQUIRES_GRAD]
+                                [--pretrained PRETRAINED]
+                                [--local_rank LOCAL_RANK]
+
+optional arguments:
+  -h, --help            show this help message and exit
+  --root ROOT           Path to the root of the dataset
+  --model MODEL         Name of the model in <alexnet|densenet121|densenet161|
+                        densenet169|densenet201|inception_v3|resnet101|resnet1
+                        52|resnet18|resnet34|resnet50|squeezenet1_0|squeezenet
+                        1_1|vgg11|vgg11_bn|vgg13|vgg13_bn|vgg16|vgg16_bn|vgg19
+                        |vgg19_bn>
+  --batch_size BATCH_SIZE
+                        input batch size for training (default: 64)
+  --val_batch_size VAL_BATCH_SIZE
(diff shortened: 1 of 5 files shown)


vfdev-5:
@chicham wow, huge work ! Thanks for the PR. I'll need some time to review it.

vfdev-5:
@chicham digging into the code, I think it would be better to split this PR into at least 2:
- `DistributedMetric` which we can put into `core`
  - method `reduce_tensor` we can put into `ignite.utils` and reuse it within `DistributedMetric`.
- the rest

I have some questions on introduced classes in `ignite/contrib/engines/dali.py`:
- `TransformPipeline` seems to be a specific to a task ?
- `create_supervised_dali_evaluator` seems to be the same as `create_supervised_evaluator` (except your specific `prepare_batch` method) ?
- `create_supervised_dali_trainer` is more like a distributed trainer. IMO some assert should be done on model too. I think we could separately provide a distributed(with fp16 or fp32) trainer.

What do you think ?

Data-drone:
@vfdev-5 how can a contribute to getting this in?

vfdev-5:
@Data-drone thanks for the proposition and sorry for the delay. I think a clean example of usage of DALI can be helpful. Today branch distrib is intended to adapt metrics to work in distributed conf, so this example will be better to PR to distrib.

dnola:
This is amazing stuff! I was working on a side project, and this ended up being what I needed. It basically mimics the standard PyTorch transforms interface - I really like the Compose style DALI construction.

I have a clean example here - specifically for the single GPU side, not the distributed stuff, it basically shows that even for the most basic possible image workflow (loading images from folders and normalizing), you can still get a really large speedup just by doing JPG decode on the GPU - I got about 4x on Colabs with the GPU they gave me:

https://colab.research.google.com/drive/1uwE0VjP6MzHyCekzaV_rNAi6GSZgnGZk

I would love to see this as a part of ignite, as it is super useful. Let me know what you think!

vfdev-5:
@dnola thanks for the colab link, looks interesting! I'll take a look in details a bit later as I get more time. 
There are few things I'm thinking here. Data loader prep stuff is not intended to be handled by ignite. Maybe by DALI itself or or other lib like albumentations... Meanwhile we can keep this code in an example folder.

So, if you would like to provide a PR with this example as multiple .py files into "examples/contrib/dali" it would be awesome! We can iterate in the PR on its structure.

dnola:
Hi, thank you for your prompt feedback!

I'm working on this now. Some thoughts I'd be interested to hear your feedback on:

1. For the purposes of the example I'm preparing, I'm currently sticking to the non-distributed single-GPU case. This means that I can pretty much just put Compose, TransformPipeline, DALIIterator, and the make_supervised_trainer update into a self contained helper script, package it with a .ipynb, and have the whole thing self contained in examples.contrib - independent of this PR.

How does that sound for now? Would you like me to open a separate PR for this?

2. This means I would leave the distributed work to you guys to figure out, as I don't have time to work that one out quite immediately yet.

3. A. In terms of changes to ignite.contrib.engine, one sensible one that might make sense is to just include the create_supervised_dali_trainer and evaluator's logic, but separate it from the DALI effort. That is to say, exposing the _prepare_batch method and letting the user customize it is useful independently of DALI. It basically lets you replace the internal logic with an adapter that can handle whatever arbitrary iterable you throw at it. I would be happy to open this as a non-DALI PR if that would help! Let me know what you think - it could maybe be like ignite.contrib.engine.create_supervised_trainer_generic or something like that.

B. Kinda unrelated - but on that note (and this one I recognize will be more controversial) - the docs basically say 'we don't include moving averages because moving averages are misleading' - but it provides a copy-pasteable trainer that allows you to do just that. I find moving averages quite helpful, especially when you are doing sub-epoch HPO (think things like Hyperband) Instead of having that be something we copy paste from the docs, maybe roll that logic into the generic/custom create_supervised_trainer above as a flag? This lets us put it into contrib so we can import directly instead of copy paste from docs. This could be together, separate from, or not at all with the above points - happy to open or not open PRs here however you prefer!

I'll also follow that albumentation thread and see if we can find a more permanent home for the DALI compose/connector logic.

Let me know your thoughts on this organization and I can move forward with the PR(s)!

PR #496, opened 2019-04-19
R-Squared Metric

### LINKED ISSUE #453: [Contrib metrics] Add R2 score metric
Maybe adding [R2 metric](https://en.wikipedia.org/wiki/Coefficient_of_determination) to `ignite.contrib.metrics.regression` could make sense. Thoughts ?

anmolsjoshi:
Fixes #453  

Description:
Added implementation of R2 score, with tests.

Please note that R2 Score is an Epoch Metric, this is because the true mean of y is required for an accurate calculation, this allows for a true comparison against sklearn. 

R-squared should be handled as stateful.

Check list:

### DIFF
--- a/ignite/contrib/metrics/regression/r2_score.py
+++ b/ignite/contrib/metrics/regression/r2_score.py
@@ -6,7 +6,7 @@ from ignite.contrib.metrics.regression._base import _BaseRegressionEpoch
 
 
 def r2_score_compute_fn(y_pred, y):
-    e = torch.sum((y - y_pred) ** 2)/torch.sum((y - y.mean()) ** 2)
+    e = torch.sum((y - y_pred) ** 2) / torch.sum((y - y.mean()) ** 2)
     return 1 - e.item()
 
 
--- a/ignite/contrib/metrics/regression/r2_score.py
+++ b/ignite/contrib/metrics/regression/r2_score.py
@@ -7,12 +7,13 @@ from ignite.contrib.metrics.regression._base import _BaseRegression
 
 
 class R2Score(_BaseRegression):
-    r"""
+    """
         Calculates the R-Squared:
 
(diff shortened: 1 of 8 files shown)


ignite/contrib/metrics/regression/r2_score.py
```diff
@@ -0,0 +1,31 @@
...
+
+from ignite.contrib.metrics.regression._base import _BaseRegressionEpoch
+
+
+def r2_score_compute_fn(y_pred, y):
+    e = torch.sum((y - y_pred) ** 2) / torch.sum((y - y.mean()) ** 2)
+    return 1 - e.item()
+
+
+class R2Score(_BaseRegressionEpoch):
```
vfdev-5: @anmolsjoshi yes we can do this while saving the complete history, however it would be better to compute with running parts:
```
SS_tot = \sum_i y_i^2 - (2*n - 1) * y_mean^2
```

anmolsjoshi: @vfdev-5 could you clarify the SS_tot function?

vfdev-5: Actually, I took as a reference this wiki article: https://en.wikipedia.org/wiki/Coefficient_of_determination
or we can take this : https://scikit-learn.org/stable/modules/model_evaluation.html#r2-score-the-coefficient-of-determination

vfdev-5: Numerator can computed with an accumulative method. Denominator should be expanded and represented as something like in above message. Such approach will avoid saving whole history. What do you think about ?

anmolsjoshi: so we keep a running total of y and number of samples, and keep calculating the mean?

anmolsjoshi:
@vfdev-5 fixed implementation of r2, it no longer stores the entire training history.

ignite/contrib/metrics/regression/r2_score.py
anmolsjoshi: Fixed implementation

PR #508, opened 2019-05-01
Use Python built-in length function for batch_size in Loss metric

JeroenDelcour:
Description:
The Loss metric now gets the batch_size using the Python built-in function `len` by default. This provides more general support for iterables (such as lists and tuples) without breaking existing functionality. Previously, only objects with a `shape` method like Torch Tensors or NumPy Arrays were supported.

For example: I'm working on an object detection model where the `y` argument for the loss function is a list. In this case, `y` can not be a tensor or array since the number of target objects is not constant across inputs. Currently, I override the default with `batch_size = len`, which seems cumbersome.

Related to #294

I ran `pytest tests` -  no new failures compared to the current master branch.

Check list:

### DIFF
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -26,7 +26,7 @@ class Loss(Metric):
     """
 
     def __init__(self, loss_fn, output_transform=lambda x: x,
-                 batch_size=lambda x: x.shape[0]):
+                 batch_size=lambda x: len(x)):
         super(Loss, self).__init__(output_transform)
         self._loss_fn = loss_fn
         self._batch_size = batch_size


vfdev-5:
@JeroenDelcour thanks for the PR ! Ok, the change does not seem to be critical.
However, in your use-case of object-detection I do not quite understand how `len(y)` can help to get the batch size. I may be wrong but, for me, for example `y = [y_locs, y_labels]` so `len(y) = 2` but batch size is `len(y_locs)`...

JeroenDelcour:
Sorry for the confusion! In my case `y = [[y_locs, y_labels], [y_locs, y_labels]]` for a batch size of 2.

Unless I'm mistaken, your case only works if the number of ground truth bounding boxes is the same for all images. You could pass a `loc` and `label` for each prior, including negative examples, then the shape of `y_loc` would be `num_batches, num_priors, 4`. In my case, however, I only pass positive bounding boxes.

vfdev-5:
@JeroenDelcour ok, thanks, I see. I checked in `maskrcnn-benchmark` and it is similar as you said:
https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/engine/trainer.py#L57
```python
images.tensors.shape
> torch.Size([8, 3, 800, 1245])
len(targets)
> 8
```

We can merge this PR as soon as we fix our CI.

Issue #513, opened 2019-05-02
Label-wise metrics (Accuracy etc.) for multi-label problems


jphdotam:
Hi,

I've made a multi-label classifier using `BCEWithLogitsLoss`. In summary a data sample can be one of 3 binary classes, which aren't mutually eclusive, so y_pred and y can look something like [0, 1, 1].

My metrics include `Accuracy(output_transform=thresholded_output_transform, is_multilabel=True)` and `Precision(output_transform=thresholded_output_transform, is_multilabel=True, average=True)}`.

However, I'm interesting in having label-specific metrics (i.e. having 3 accuracies etc.). This is important because it allows me to see what labels are compromising my overall accuracy the most (a 70% accuracy be a 30% error in a single label, or a more modest error scattered across 3 labels).

There is no option to disable averaging for `Accuracy()` as with the others, and setting `average=False` for `Precision()` does not do what I expected (it yields a binary result per datum, not per label, so I end up with a tensor of size 500, not 3, if my dataset n=500).

Is there a way to get label-wise metrics in mutlilabel problems? Or a plan to introduce it?

P.S. I'd love to get an invite to the slack workspace if possible? How do I go about doing that?

vfdev-5:
@jphdotam thanks for the feedback! You are correct, multi-label case is always averaged for now for Accuracy, Precision, Recall.

> Is there a way to get label-wise metrics in mutlilabel problems? Or a plan to introduce it?

There is an issue with a similar requirement https://github.com/pytorch/ignite/issues/467
For instance we have not much bandwidth to work on that. If you can send a PR for that, we'll be awesome.

> P.S. I'd love to get an invite to the slack workspace if possible? How do I go about doing that?

You can find a link for that here : https://pytorch.org/resources

jphdotam:
Many thanks, I've made a pull request here: https://github.com/pytorch/ignite/pull/516

I'm quite new to working on large projects so apologies if I have gone about this inappropriately.

jphdotam:
In the mean time whilst the core team decide how best to implement this, this is a custom class I've made for the task which inherits from Accuracy:

```python
class LabelwiseAccuracy(Accuracy):
    def __init__(self, output_transform=lambda x: x):
        self._num_correct = None
        self._num_examples = None
        super(LabelwiseAccuracy, self).__init__(output_transform=output_transform)

    def reset(self):
        self._num_correct = None
        self._num_examples = 0
        super(LabelwiseAccuracy, self).reset()

[...]

Issue #467, opened 2019-03-31
Top-K precision/recall multilabel metrics for ranking task


vfdev-5:
Following the discussion from https://github.com/pytorch/ignite/issues/466#issuecomment-478339986 it would be nice to have such metric in Ignite.

In the context of a multilabel task, compute a top-k precision/recall per label (treating all labels independently).

Data-drone:
any update on this?

vfdev-5:
cc @anmolsjoshi

vfdev-5:
@Data-drone in case, maybe you could be interested in [this temporary solution](https://github.com/pytorch/ignite/issues/467#issuecomment-488983281)

Tanmay06:
Hi! I would like to work on this. Would you please assign this to me ?

vfdev-5:
@Tanmay06 sure !

vfdev-5:
@Tanmay06 any updates on this issue ?

Tanmay06:
I've almost completed it, but I'm getting some bugs. I'll fix those and mostly raise a PR by this weekend.

vfdev-5:
Sounds good! Do not hesitate to send a draft PR such that we could iterate faster.

julien-blanchon:
Just to be sure I understand well this issues.

1 - By multilabel (in the context of classification), you means that each data point can belong to multiple classes at the same time. For example, in an image classification task, a single image may contain multiple objects, and the model needs to predict all the objects present in the image.

2 - Following the reference given by @RoyHirsch in #466, this look like `MultilabelPrecision` and `MultilabelRecall` of `torchmetrics.classification`.

https://github.com/Lightning-AI/metrics/blob/54a06013cdac4895bf8e85b583c5f220388ebc1d/src/torchmetrics/classification/precision_recall.py#LL492

vfdev-5:
> By multilabel (in the context of classification), you means that each data point can belong to multiple classes at the same time. 

Yes, exactly, non-exclusive class labels, similar to tags.  For 3 classes, ground truth can be `y=[0, 1, 1]` or `y=[1, 1, 1]` or `y=[0, 0, 0]` etc.

> 2 - Following the reference given by @RoyHirsch in https://github.com/pytorch/ignite/issues/466, this look like MultilabelPrecision [1] and MultilabelRecall [2] of torchmetrics.classification.

I do not know what they are doing those classes MultilabelPrecision [1] and MultilabelRecall [2].
[...]

PR #516, opened 2019-05-02
Label-wise accuracy metrics implemented for multilabel classification.

jphdotam:
Signed-off-by: James P Howard <github@jph.am>

Fixes: Feature added - label-wise accuracy option for Accuracy metric.

Description: `Accuracy()` metrics with `is_multilabel=True` can now be passed `labelwise=True`. When present, the metric returns a tensor of accuracies for each class. For example:

```
    metrics={'loss': Loss(loss),
    'accuracy': Accuracy(output_transform=thresholded_output_transform, is_multilabel=True),
    'precision': Precision(output_transform=thresholded_output_transform, is_multilabel=True, average=True),
    'label_acc':Accuracy(output_transform=thresholded_output_transform, is_multilabel=True, labelwise=True)},
    device=device)

    evaluator.run(train_loader)
    acc, loss, precision, label_acc = metrics['accuracy'], metrics['loss'], metrics['precision'], metrics['label_acc']
    print(f"\rEnd of epoch {engine.state.epoch:03d}")
    print(f"TRAINING Accuracy: {acc:.3f} | Loss: {loss:.3f} | Precision: {precision:.3f} | Label-wise accuracy: {label_acc}")
    writer.add_scalar("training/accuracy", acc, engine.state.epoch)

def log_validation_results(engine):
    evaluator.run(test_loader)
    acc, loss, precision, label_acc = metrics['accuracy'], metrics['loss'], metrics['precision'], metrics['label_acc']
    print(f"TESTING  Accuracy: {acc:.3f} | Loss: {loss:.3f} | Precision: {precision:.3f} | Label-wise accuracy: {label_acc}\n")
    writer.add_scalar("testing/loss", loss, engine.state.epoch)
    writer.add_scalar("testing/accuracy", acc, engine.state.epoch)

trainer.run(train_loader, max_epochs=30)

```

Yields:

```
End of epoch 001
TRAINING Accuracy: 0.753 | Loss: 0.334 | Precision: 0.212 | Label-wise accuracy: tensor([0.8662, 0.8662], device='cuda:0')
TESTING  Accuracy: 0.725 | Loss: 0.341 | Precision: 0.221 | Label-wise accuracy: tensor([0.8302, 0.8755], device='cuda:0')

End of epoch 002
TRAINING Accuracy: 0.748 | Loss: 0.363 | Precision: 0.160 | Label-wise accuracy: tensor([0.8134, 0.9022], device='cuda:0')
TESTING  Accuracy: 0.672 | Loss: 0.537 | Precision: 0.087 | Label-wise accuracy: tensor([0.7057, 0.8792], device='cuda:0') 

```
Check list:

### DIFF
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -121,7 +121,9 @@ class Accuracy(_BaseClassification):
         self._num_correct = None
         self._num_examples = None
         self._num_correct_labelwise = None
-        super(Accuracy, self).__init__(output_transform=output_transform, is_multilabel=is_multilabel, labelwise=labelwise)
+        super(Accuracy, self).__init__(output_transform=output_transform,
+                                       is_multilabel=is_multilabel,
+                                       labelwise=labelwise)
 
     def reset(self):
         self._num_correct = 0
@@ -145,17 +147,17 @@ class Accuracy(_BaseClassification):
             last_dim = y_pred.ndimension()
             y_pred = torch.transpose(y_pred, 1, last_dim - 1).reshape(-1, num_classes)
             y = torch.transpose(y, 1, last_dim - 1).reshape(-1, num_classes)
-            correct = torch.all(y == y_pred.type_as(y), dim=-1) # Sample-wise
-
+            correct = torch.all(y == y_pred.type_as(y), dim=-1)  # Sample-wise
 
         self._num_correct += torch.sum(correct).item()
         self._num_examples += correct.shape[0]
         if self._labelwise:
             if self._num_correct_labelwise is not None:
-                self._num_correct_labelwise = torch.add(self._num_correct_labelwise, torch.sum(y == y_pred.type_as(y), dim=0))
+                self._num_correct_labelwise = torch.add(self._num_correct_labelwise,
+                                                        torch.sum(y == y_pred.type_as(y),
+                                                                  dim=0))
             else:
                 self._num_correct_labelwise = torch.sum(y == y_pred.type_as(y), dim=0)
-
 
     def compute(self):
         if self._num_examples == 0:
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -8,11 +8,15 @@ from ignite.exceptions import NotComputableError
 
 class _BaseClassification(Metric):
 
-    def __init__(self, output_transform=lambda x: x, is_multilabel=False):
+    def __init__(self, output_transform=lambda x: x, is_multilabel=False, labelwise=False):
         self._is_multilabel = is_multilabel
+        self._labelwise = labelwise
         self._type = None
         self._num_classes = None
         super(_BaseClassification, self).__init__(output_transform=output_transform)
+        if labelwise and not is_multilabel:
+            raise ValueError("labelwise=True is only appropriate for multi-label classificiation, "
+                             "i.e. is_multilabel=True")
 
     def reset(self):
         self._type = None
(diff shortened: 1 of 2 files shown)


anmolsjoshi:
@jphdotam thanks for the PR! Could you add a few tests? 

Have a look [here](https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/test_accuracy.py), we tests accuracy against scikit-learn's implementation. 

Let us know if you get stuck or have any questions!

vfdev-5:
@jphdotam thanks for the PR ! To merge it I think we need to discuss about the API. I'm not a fan of introducing another flag. Maybe we can opt something like in torch the arguments of `nn.CrossEntropy`: deprecated `reduce` and new arg `reduction` which can have text values.
Can we generalize this PR to cover two issues : https://github.com/pytorch/ignite/issues/513 and https://github.com/pytorch/ignite/issues/467 ?

@jphdotam could you please provide a very simple example of manually computing such accuracy score labelwise. For example, I have `y_true = [(1, 1, 0), (0, 0, 0,), (1, 1, 1)]` and `y_pred=[(1, 0, 1), (0, 0, 1), (0, 1, 1)]` what is the score and how it is computed in details ?

jphdotam:
Hi @vfdev-5.

Your example shows a batch size of 3 for a binary classifier with 3 labels.
The label-wise accuracy is essentially an accuracy for each position within the tuples.
If I expand your example to a batch size of 4 (just to make samples versus labels clearer), this is essentially how it works:

```
y_true = np.array([(1, 1, 0), (0, 0, 0), (1, 1, 1), (0, 1 ,0)])
y_pred = np.array([(1, 0, 1), (0, 0, 1), (0, 1, 1), (0, 1, 0)])
correct = y_true == y_pred
correct
Out[38]: 
array([[ True, False, False],
       [ True,  True, False],
       [False,  True,  True],
       [ True,  True,  True]])
np.mean(correct, axis=0)
Out[39]: array([0.75, 0.75, 0.5 ])
```

So there is 75% accuracy for the 1st category, 75% accuracy for the second, and 50% for the third.

It's very useful if one wishes to see which label in a multi-label classifier is compromising the overall accuracy.

Re: merging, would you rather I instead created a new separate metric from Accuracy called MultilabelAccuracy, or something? And submit it to either ignite.metrics or ignite.contrib.metrics?

vfdev-5:
@jphdotam thanks for the explanation. Now it is clear that we speak about the same computation method.

> Re: merging, would you rather I instead created a new separate metric from Accuracy called MultilabelAccuracy, or something?

Previously, we had `BinaryAccuracy` and `CategorialAccuracy` that we merged into a single class. Next we added the support of multilabel same as in sklearn. IMO we should keep a single class.

Let me think about the new API and I comment out here. If you have other ideas on the API we can discuss about.

jphdotam:
Ok great. In the mean time I will just use it as a new class as I've posted here, since that's probably easier until we decide.

anmolsjoshi:
@jphdotam thanks for providing the code base. In discussion with @vfdev-5, we were thinking the following:
- Add a labelwise parameter, so the constructor would be Accuracy(is_multilabel=True, labelwise=True)
- Add a check that labelwise can only be True with multilabel cases, maybe with a warning and no error raising
- The same would need to be applied to Precision and Recall, as these metrics are closely related in the way they are written. 

We might be working towards a minor release for now, so we shouldn't do major API changes. For the next major release (0.3.0), we can introduce a new_multilabel_arg with options None (binary/multiclass), multilabel (single accuracy value), labelwise (each label for multilabel). 

What are your thoughts?

Oktai15:
`Description: Accuracy() metrics with is_multilabel=True can now be passed labelwise=True`

@jphdotam why do you want to add this feature only for `multilabel` case? It can be useful also with multiclass training too, isn't it?

PR #523, opened 2019-05-12
Added event_to_attr in register_events

### LINKED ISSUE #501: event_to_attr should be extensible for register_event
There should be a way to extend `event_to_attr` when registering new events.

anmolsjoshi:
Fixes #501  

Description:
* Added parameter in register_events to update state with Custom events
* Update State and Engine to initialize values using State.event_to_attr (removed hard coded epoch=0, iteration=0 in initialization of State)

Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -34,11 +34,13 @@ class State(object):
     }
 
     def __init__(self, **kwargs):
-        self.iteration = 0
         self.output = None
         self.batch = None
         for k, v in kwargs.items():
             setattr(self, k, v)
+
+        for value in self.event_to_attr.values():
+            setattr(self, value, 0)
 
     def get_event_attrib_value(self, event_name):
         if event_name not in State.event_to_attr:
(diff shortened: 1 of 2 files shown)


ignite/engine/engine.py
```diff
@@ -89,18 +91,19 @@ def __init__(self, process_function):
...
         This opens the door to make the :meth:`~ignite.engine.Engine.run` loop even more
         configurable.
 
-        By default, the events from :class:`~ignite.engine.Events` are registerd.
+        By default, the events from :class:`~ignite.engine.Events` are registered.
 
         Args:
             *event_names: An object (ideally a string or int) to define the
                 name of the event being supported.
+            event_to_attr (dict/object): A dictionary or object to map an event to a state attribute.
```
vfdev-5: > dict/object): A dictionary or object to map an event to a state attribute.

This is not clear what is the type of event_to_attr.

ignite/engine/engine.py
```diff
@@ -114,10 +117,15 @@ class Custom_Events(Enum):
 
             engine = Engine(process_function)
             engine.register_events(*Custom_Events)
```
vfdev-5: Maybe we can provide another example here with `event_to_attr`

ignite/engine/engine.py
anmolsjoshi: Updated, event_to_attr should only be a dictionary, code includes check. 

Python 2.7 does not allow for regular parameter after * parameter, so had to update the code accordingly.

ignite/engine/engine.py
anmolsjoshi: Provide example in docstring.

PR #524, opened 2019-05-13
Added accumulative single variable metrics, Fixes #512

### LINKED ISSUE #512: Average value metric
Discussed on the slack:
> Joost van Amersfoort [7:14 PM]
I have a very simple question, how do I compute an average over a validation set?I have an evaluator engine and I am considering the EpochMetric, and pass `None` as the label and make the compute_fn only average over the first column, or the `RunningAverage` metric with `alpha=1`. However both feel a bit gross for this very simple problem.

> Interesting, looking at the `Loss` source then it indeed seems that it’s very close to what I want and more scalable than EpochMetric. But the problem is that it’s setup for supervised learning and forces len(output) >= 2

Idea is to provide some metric class similar to `Loss` but with no restrictions, just to compute average on provided iteration values.

vfdev-5:
Fixes #512

Description: 
A generic and 2 basic accumulators as metrics are added.

Example with Average accumulator :
```python
evaluator = ...

custom_var_mean = Average(output_transform=lambda output: output['custom_var'])
custom_var_mean.attach(evaluator, 'mean_custom_var')

state = evaluator.run(dataset)
# state.metrics['mean_custom_var'] -> average of output['custom_var']
```

Check list:

cc @y0ast , any feedback on this whether this could solve your needs

### DIFF
--- a/docs/source/metrics.rst
+++ b/docs/source/metrics.rst
@@ -99,3 +99,9 @@ Metrics also support indexing operation (if metric's result is a vector/matrix/t
 .. autofunction:: IoU
 
 .. autofunction:: mIoU
+
+.. autoclass:: VariableAccumulation
+
+.. autoclass:: Average
+
+.. autoclass:: GeometricAverage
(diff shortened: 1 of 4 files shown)


y0ast:
I think this looks perfect!

Thanks @vfdev-5 !

Issue #519, opened 2019-05-07
Improve OutputHandler's API


vfdev-5:
Following a dicsussion from the slack:
> trying to set up a regular evaluation (on an evaluation dataset) along the training (every 500 steps for instance) and log that on Tensorboard with the training step number. Not exactly sure if I can do that with 0.2, I've tried:
> - to log metrics after the evaluation (event `EPOCH_COMPLETED` for the evaluation engine) but with the event `ITERACTION_COMPLETED` of the training engine => doesn't seem possible

cc @thomwolf

anmolsjoshi:
@vfdev-5 here is code I mentioned on slack - https://github.com/anmolsjoshi/ignite/blob/tensorboard_fix/ignite/contrib/handlers/base_logger.py#L30-L55

In this branch, I have updated mnsit_with_tensorboard_logger.py to evaluate at CustomPeriodicEvent at every 500 iterations. 

Thoughts?

Edge cases?

PR #531, opened 2019-05-21
TensorboardLogger - Improved  

### LINKED ISSUE #530: Tensorboard logger won't work with tensorboardX 1.7 (latest release)
New release of tensorboardX 1.7 has some [breaking changes](https://github.com/lanpa/tensorboardX/pull/422/files#diff-0900f72201599ce7d24575f21ca88a85) that will result in a bug of `TensorboardLogger` init : https://travis-ci.org/pytorch/ignite/jobs/534870310#L2664

anmolsjoshi:
Fixes #530 #519 

Description:
- Updated TensorboardLogger to remove another_engine and add global_step_transform
- Updated SummaryWriter input as logdir, this seemed to be the breaking change in v1.7
- Updated tests accordingly
- Updated example to use updated API

Check list:

### DIFF
--- a/examples/contrib/mnist/mnist_with_tensorboard_logger.py
+++ b/examples/contrib/mnist/mnist_with_tensorboard_logger.py
@@ -93,6 +93,9 @@ def run(train_batch_size, val_batch_size, epochs, lr, momentum, log_dir):
         train_evaluator.run(train_loader)
         validation_evaluator.run(val_loader)
 
+    def global_step_transform(*args, **kwargs):
+        return trainer.state.epoch
+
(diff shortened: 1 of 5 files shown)


vfdev-5:
Thinking more about this BC change, IMO it would be better to keep until `0.2.1` release the previous argument `another_engine` and show a deprecation. After the `0.2.1` release we'll remove it from the master.

anmolsjoshi:
@vfdev-5 I agree, I added another_engine back to the OutputHandler for each logger, reverted tests and examples as well.

PR #539, opened 2019-05-28
Issue 538

### LINKED ISSUE #538: TypeError: type object got multiple values for keyword argument 'logdir'
>from ignite.contrib.handlers.tensorboard_logger import TensorboardLogger
>tb = TensorboardLogger(log_dir=None) 

gives the following error:

>     self.file_writer = FileWriter(logdir=log_dir, **kwargs)
> TypeError: type object got multiple values for keyword argument 'logdir'

This also happens when I run

>from tensorboardX import SummaryWriter
>sw = SummaryWriter(log_dir=None)

>     self.file_writer = FileWriter(logdir=log_dir, **kwargs)
> TypeError: type object got multiple values for keyword argument 'logdir'

However, when I do don't put in logdir=log_dir, and simply pass in log_dir, it works fine.

I was looking to see if there's any conflicting reference with `self` or anything like that, but it doesn't seem like that's the problem, since all the `self`'s seem to be in place, so I'm not sure.

Any ideas?

ignite version: 1.1.0
tensorboard version: 1.13.1
tensorboardX version: 1.6

johnstill:
Fixes #538 

I'm not sure about this warning text.  I think it's what was called for in the issue discussion though.

### DIFF
--- a/ignite/contrib/handlers/tensorboard_logger.py
+++ b/ignite/contrib/handlers/tensorboard_logger.py
@@ -407,14 +407,14 @@ class TensorboardLogger(BaseLogger):
 
         try:
             self.writer = SummaryWriter(logdir=log_dir)
-        except TypeError:
-            from pkg_resources import get_distribution
-            version = get_distribution('tensorboardX').version
-            if version < '1.7':
+        except TypeError as err:
+            if "type object got multiple values for keyword argument 'logdir'" == str(err):
+                self.writer = SummaryWriter(log_dir=log_dir)
+                from pkg_resources import get_distribution
+                version = get_distribution('tensorboardX').version
                 warnings.warn('tensorboardX version {} will not be supported '
                               'after ignite 0.3.0; please upgrade'.format(version),
                               DeprecationWarning)
-                self.writer = SummaryWriter(log_dir=log_dir)
             else:
                 raise
 
--- a/ignite/contrib/handlers/tensorboard_logger.py
+++ b/ignite/contrib/handlers/tensorboard_logger.py
@@ -405,7 +405,17 @@ class TensorboardLogger(BaseLogger):
             raise RuntimeError("This contrib module requires tensorboardX to be installed. "
                                "Please install it with command: \n pip install tensorboardX")
 
-        self.writer = SummaryWriter(logdir=log_dir)
+        try:
+            self.writer = SummaryWriter(logdir=log_dir)
+        except TypeError:
+            from pkg_resources import get_distribution
+            version = get_distribution('tensorboardX').version
+            if version < '1.7':
+                warnings.warn('tensorboardX version {} is outdated, please upgrade'.format(version),
+                              DeprecationWarning)
+                self.writer = SummaryWriter(log_dir=log_dir)
+            else:
+                raise
 
     def close(self):
         self.writer.close()
--- a/ignite/contrib/handlers/tensorboard_logger.py
+++ b/ignite/contrib/handlers/tensorboard_logger.py
@@ -411,7 +411,8 @@ class TensorboardLogger(BaseLogger):
             from pkg_resources import get_distribution
             version = get_distribution('tensorboardX').version
             if version < '1.7':
-                warnings.warn('tensorboardX version {} is outdated, please upgrade'.format(version),
+                warnings.warn('tensorboardX version {} will not be supported '
+                              'after ignite 0.3.0; please upgrade'.format(version),
                               DeprecationWarning)
                 self.writer = SummaryWriter(log_dir=log_dir)
             else:
--- a/ignite/contrib/handlers/tensorboard_logger.py
+++ b/ignite/contrib/handlers/tensorboard_logger.py
@@ -410,13 +410,11 @@ class TensorboardLogger(BaseLogger):
         except TypeError as err:
             if "type object got multiple values for keyword argument 'logdir'" == str(err):
                 self.writer = SummaryWriter(log_dir=log_dir)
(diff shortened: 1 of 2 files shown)


ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -405,7 +405,18 @@ def __init__(self, log_dir):
             raise RuntimeError("This contrib module requires tensorboardX to be installed. "
                                "Please install it with command: \n pip install tensorboardX")
 
-        self.writer = SummaryWriter(logdir=log_dir)
+        try:
+            self.writer = SummaryWriter(logdir=log_dir)
+        except TypeError:
+            from pkg_resources import get_distribution
```
vfdev-5: Should we install `setuptools` to have this package ?
Can't we just call
```
import tensorboardX
if tensorboardX.__version__ < "1.7":
    ...
```

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -405,7 +405,18 @@ def __init__(self, log_dir):
             raise RuntimeError("This contrib module requires tensorboardX to be installed. "
                                "Please install it with command: \n pip install tensorboardX")
 
-        self.writer = SummaryWriter(logdir=log_dir)
+        try:
+            self.writer = SummaryWriter(logdir=log_dir)
+        except TypeError:
```
vfdev-5: Probably we would like check if this TypeError is really related to `logdir` vs `log_dir`. Otherwise, warning can be misleading...

ignite/contrib/handlers/tensorboard_logger.py
johnstill: I tried that, and for some reason `tensorboardX` doesn't expose a `__version__` attribute.

ignite/contrib/handlers/tensorboard_logger.py
johnstill: Ok, I'm checking against the exact error message now so we'll only catch the one.

ignite/contrib/handlers/tensorboard_logger.py
vfdev-5: That's true : https://github.com/lanpa/tensorboardX/commit/28cbf7f5a82067662f0f983080ee30baa0411ca2#commitcomment-33048905
and this was fixed in `1.7`.

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -405,7 +405,18 @@ def __init__(self, log_dir):
...
+        except TypeError as err:
+            if "type object got multiple values for keyword argument 'logdir'" == str(err):
+                self.writer = SummaryWriter(log_dir=log_dir)
+                from pkg_resources import get_distribution
+                version = get_distribution('tensorboardX').version
+                warnings.warn('tensorboardX version {} will not be supported '
+                              'after ignite 0.3.0; please upgrade'.format(version),
+                              DeprecationWarning)
+            else:
+                raise
```
vfdev-5:
```suggestion
                raise err
```

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -405,7 +405,18 @@ def __init__(self, log_dir):
             raise RuntimeError("This contrib module requires tensorboardX to be installed. "
                                "Please install it with command: \n pip install tensorboardX")
 
-        self.writer = SummaryWriter(logdir=log_dir)
+        try:
+            self.writer = SummaryWriter(logdir=log_dir)
+        except TypeError as err:
+            if "type object got multiple values for keyword argument 'logdir'" == str(err):
+                self.writer = SummaryWriter(log_dir=log_dir)
+                from pkg_resources import get_distribution
```
vfdev-5: What do you think about removing `pkg_resources` (it seems like an additional?) and changing the warning message to
"tensorboardX version <1.7 will not be supported after ignite 0.3.0; please upgrade"

ignite/contrib/handlers/tensorboard_logger.py
johnstill: Then we could use the attribute error to infer that the version < 1.7, which is kind of ironic (the lack of a version gives us the version).  That would let us remove the `pkg_resources` import - you want me to commit that?

vfdev-5: Yes, please! That was my suggestion below, actually :)

vfdev-5:
@johnstill looks good to me! 
I think our `coverage` tool will complain about not covered change, so probably I'll need to provide a small test of that. I can do it later. But if you would like to contribute, it would be awesome!
Anyway, let's wait for CI and all other checks finished.

anmolsjoshi:
@vfdev-5 @johnstill maybe we should add something in the docstring on how to access the log_dir? Mainly pointing out the difference between v1.7 and versions below

Some users might use tb_logger.writer.log_dir to get the directory name for ModelCheckpoint. 

Thoughts?

vfdev-5:
@anmolsjoshi yes made we can just give a small note about that as it is up to TensorboardX API. 

> Some users might use tb_logger.writer.log_dir to get the directory name for ModelCheckpoint.

yes, it is the case of https://github.com/huggingface/transfer-learning-conv-ai
but a better way to do this IMO
```
path = ...
tb_logger = TensorboardLogger(path)
checkpoint = ModelCheckpoint(path)
```

anmolsjoshi:
@vfdev-5 I think we're ok without it. I'm good with this merge.

PR #542, opened 2019-06-05
[WIP ]Added labelwise metrics with tests

### LINKED ISSUE #513: Label-wise metrics (Accuracy etc.) for multi-label problems

anmolsjoshi:
Fixes #513 and #516 

Description:
Added tests

Check list:

### DIFF
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -8,8 +8,12 @@ from ignite.exceptions import NotComputableError
 
 class _BaseClassification(Metric):
 
-    def __init__(self, output_transform=lambda x: x, is_multilabel=False):
+    def __init__(self, output_transform=lambda x: x, is_multilabel=False, labelwise=False):
+        if labelwise and not is_multilabel:
+            raise ValueError("labelwise is only applicable for multilabel cases. If labelise=True, is_multilabel=True.")
+
         self._is_multilabel = is_multilabel
+        self._labelwise = labelwise
         self._type = None
         self._num_classes = None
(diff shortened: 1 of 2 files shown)


anmolsjoshi:
@vfdev-5 I have at least added Accuracy with labelwise option. Wanted to further discuss Precision and Recall before implementing.

ignite/metrics/accuracy.py
```diff
@@ -8,8 +8,12 @@
 
 class _BaseClassification(Metric):
 
-    def __init__(self, output_transform=lambda x: x, is_multilabel=False):
+    def __init__(self, output_transform=lambda x: x, is_multilabel=False, labelwise=False):
+        if labelwise and not is_multilabel:
+            raise ValueError("labelwise is only applicable for multilabel cases. If labelise=True, is_multilabel=True.")
```
```suggestion
            raise ValueError("labelwise is only applicable for multilabel cases. If labelwise=True, is_multilabel=True.")
```

vfdev-5 (changes requested):
I'm not a fan of another bool flag `labelwise`. Idea is to introduce a new variable, e.g `multilabel` that takes values for example `exact`, `labelwise`, `top-K` with variable K (`top-3`, `top-5` etc):
- `exact` is current `is_multilabel=True` behaviour
- `labelwise` is your implementation behaviour with `labelwise=True`
- `top-K` with parsing K into an integer and computes top-k multi-label accuracy.

@anmolsjoshi what do you think ?

Old argument `is_multilabel` becomes deprecated and will be removed since 0.3.0

PR #549, opened 2019-06-26
Added Fashion MNIST jupyter notebook

ANUBHAVNATANI:
Description:
Added Fashion MNIST jupyter notebook

Check list:


vfdev-5:
@ANUBHAVNATANI looks good, thanks, but maybe we can improve some points:
- typos : `Infrencing the model`, #moving model to cpu for infrencing 
- nits: I would prefer the comments with a space between # symbol and the following text:
```
#transform to normalize the data
->
# transform to normalize the data
```
- in the text all variables, modules etc it would be better to put into the ticks. For example,
> We import torch, nn and functional modules to create our models.
```
-> We import `torch`, `nn` and `functional` modules to create our models.
```
- Training curves looks a bit strange with accuracy (train/val) going down. Can we improve this ?
- I'd like to see also the usage of `ConfusionMatrix` how well classifier is doing, can you please implement this ?

ANUBHAVNATANI:
@vfdev-5 I had made all the changes that you had mentioned above.

vfdev-5:
@ANUBHAVNATANI thanks, that looks better. However, concerning the confusion matrix I wasn't enough clear. In ignite we provide [`ConfusionMatrix`](https://github.com/pytorch/ignite/blob/master/ignite/metrics/confusion_matrix.py#L10) metric, so I was thinking to compute it as attached metric.

About the training, so now we are clearly overfitting, any ideas how to get val accuracy ~> 93-95 % (it's a MNIST-like dataset anyway) ?

ANUBHAVNATANI:
@vfdev-5  I think by changing the model architecture or doing some hyper parameter tuning we can achieve better accuracy.
For the confusion matrix I was not able to fully understand the Confusion Matrix of ignite library and it was not working on my laptop I was not able to import it that's why I used sklearn implementation.

vfdev-5:
>  I think by changing the model architecture or doing some hyper parameter tuning we can achieve better accuracy.

Yes that would be nice to see in the notebook.

> For the confusion matrix I was not able to fully understand the Confusion Matrix of ignite library and it was not working on my laptop I was not able to import it that's why I used sklearn implementation.

Can post your code which didn't work. Confusion matrix should normally work as any other metric by attaching to an evaluator...

ANUBHAVNATANI:
@vfdev-5  I had made further changes to the notebook to get high accuracy and encompassing confusion matrix of ignite library.

vfdev-5:
@ANUBHAVNATANI nice! Result is better.  
just some nits:

1) Missing spaces after comma
```
from torchvision import datasets,transforms ->
from torchvision import datasets, transforms
```
and 
```
from ignite.metrics import Accuracy,Loss,RunningAverage,ConfusionMatrix
from ignite.handlers import ModelCheckpoint,EarlyStopping
```
and maybe others

2) In "CNN Model", "Explanation of Model Architecture", I'm not sure about whether we need to provide the following:
```
- The output of Convolutional layer is determined by W'(new dimension) = (W-F+2P)/S + 1 here W is dimension 28(28 x 28) and F is the number of filters and P is the amount of padding and S is the amount of stride.
- After second convolutional layer and max pooling we get the output as 64x6x6
- We convert this 64x6x6 output to the flat tensor by using pytorch tensor view method
- Then we pass this flat tensor to the fully connected layers
```
Maybe we could better explain ignite part rather than NN stuff. What do you think ?

3) Can you please make another pass to check that the functions are between ticks...
For example, below "EarlyStopping - Tracking Validation Loss" is not the case. Please, fix also the cases like 
```
Now we will setup a `Early Stopping` handler for this training process. 
```
to 
```
Now we will setup a `EarlyStopping` handler for this training process. 
```

4) Small code refactoring
```
evaluator = create_supervised_evaluator(model,metrics={'accuracy':Accuracy(),'nll':Loss(criterion),'cm':ConfusionMatrix(num_classes=10)},device=device)
```
to
```
metrics = {
# ...
}
evaluator = create_supervised_evaluator(model, metrics=metrics, device=device)
```
and explain how and which  we setup the metrics, please.

ANUBHAVNATANI:
@vfdev-5  Thanks for the suggestion I had made all the changes to the notebook as you have mentioned before.

PR #563, opened 2019-07-24
Updated _check_shape for BaseClassification

### LINKED ISSUE #555: y and y_pred shape error in _BaseClassification._check_shape()
Hello,

In a single-label, multiple-class classification task, suppose my model gives the output logits `y_pred` of shape (batch, num_classes, 1) and the correct results are `y` of shape (batch, 1), which are the integer labels, then the _check_shape() gives an error
`y must have shape of (batch_size, ...) and y_pred must have "
                             "shape of (batch_size, num_categories, ...) or (batch_size, ...), "
                             "but given {} vs {}.`

which happens because the code squeezes the second dimension of `y`. See https://github.com/pytorch/ignite/blob/c92838c950699839c37f57bb50f79f84ecb71fc3/ignite/metrics/accuracy.py#L24

It seems like the shapes are legal according to the message. Am I doing something wrong here, or what is the intended usage?

anmolsjoshi:
Fixes #555  

Description:
Added check to squeeze last dimension if equals to 1. 

Check list:

### DIFF
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -18,21 +18,8 @@ class _BaseClassification(Metric):
         self._type = None
         self._num_classes = None
 
-    def _process_shape(self, y):
-        # squeeze_dims = [i for i in range(len(y.shape)) if y.shape[i] == 1 and i != 0]
-        # for dim in reversed(squeeze_dims):
-        #     y = y.squeeze(dim=dim)
-        if y.shape[0] == 1:
-            y = y.squeeze().unsqueeze(dim=0)
-        else:
-            y = y.squeeze()
-        return y
-
     def _check_shape(self, output):
         y_pred, y = output
-
-        y_pred = self._process_shape(y_pred)
-        y = self._process_shape(y)
 
         if not (y.ndimension() == y_pred.ndimension() or y.ndimension() + 1 == y_pred.ndimension()):
             raise ValueError("y must have shape of (batch_size, ...) and y_pred must have "
@@ -50,8 +37,6 @@ class _BaseClassification(Metric):
 
         if self._is_multilabel and not (y.shape == y_pred.shape and y.ndimension() > 1 and y.shape[1] != 1):
             raise ValueError("y and y_pred must have same shape of (batch_size, num_categories, ...).")
-
-        return y_pred, y
 
     def _check_type(self, output):
         y_pred, y = output
@@ -127,8 +112,8 @@ class Accuracy(_BaseClassification):
         super(Accuracy, self).reset()
 
     def update(self, output):
-
-        y_pred, y = self._check_shape(output)
+        y_pred, y = output
+        self._check_shape((y_pred, y))
         self._check_type((y_pred, y))
 
         if self._type == "binary":
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -20,22 +20,8 @@ class _BaseClassification(Metric):
 
     def _check_shape(self, output):
         y_pred, y = output
-
-        if y.ndimension() > 1 and y.shape[1] == 1:
-            # (N, 1, ...) -> (N, ...)
-            y = y.squeeze(dim=1)
-
-        if y.ndimension() > 1 and y.shape[-1] == 1:
-            # (N, ..., 1) -> (N, ...)
-            y = y.squeeze(dim=-1)
-
-        if y_pred.ndimension() > 1 and y_pred.shape[1] == 1:
-            # (N, 1, ...) -> (N, ...)
-            y_pred = y_pred.squeeze(dim=1)
-
-        if y_pred.ndimension() > 1 and y_pred.shape[-1] == 1:
-            # (N, 1, ...) -> (N, ...)
-            y_pred = y_pred.squeeze(dim=-1)
+        y_pred = y_pred.squeeze()
+        y = y.squeeze()
 
         if not (y.ndimension() == y_pred.ndimension() or y.ndimension() + 1 == y_pred.ndimension()):
             raise ValueError("y must have shape of (batch_size, ...) and y_pred must have "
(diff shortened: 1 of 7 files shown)


vfdev-5:
@anmolsjoshi I thought about to remove this squeeze operations. Actually, I'm not sure about their need...

anmolsjoshi:
@vfdev-5 were you thinking of doing shape check like this - y.squeeze().shape == y_pred.squeeze().shape ?

vfdev-5:
@anmolsjoshi I was thinking to remove these lines :

https://github.com/pytorch/ignite/blob/f50a3f31336dcf16a128f0d1917d6ed28b296a46/ignite/metrics/accuracy.py#L24-L30

anmolsjoshi:
@vfdev-5 processing it like that, allows for the [_check_type](https://github.com/pytorch/ignite/blob/f50a3f31336dcf16a128f0d1917d6ed28b296a46/ignite/metrics/accuracy.py#L51) function to work correctly. Allowing the distinguishing between binary and multiclass cases.

vfdev-5:
OK, that's good to know. 
Anyway, the logic to remove a dimension with 1 can either generalized to remove all dimensions with 1 or either keep it as is and update our further logic.
Imagine another case, where user has  somehow `y_pred` as `(B, C, 1, S)` and `y` as `(B, 1, S)`, current code will fail again...

ignite/metrics/accuracy.py
```diff
@@ -20,14 +20,8 @@ def reset(self):
...
-
-        if y.ndimension() > 1 and y.shape[1] == 1:
-            # (N, 1, ...) -> (N, ...)
-            y = y.squeeze(dim=1)
-
-        if y_pred.ndimension() > 1 and y_pred.shape[1] == 1:
-            # (N, 1, ...) -> (N, ...)
-            y_pred = y_pred.squeeze(dim=1)
+        y_pred = y_pred.squeeze()
+        y = y.squeeze()
```
vfdev-5: nice :)

anmolsjoshi: @vfdev-5 I just realized this fails in case of batch size = 1.

I think we should squeeze all dimensions except for the 1st. Thoughts?

vfdev-5: make sense, we can try like that

anmolsjoshi: @vfdev-5 have a look at the updated code, let me know your thoughts.

ignite/metrics/accuracy.py
```diff
@@ -18,16 +18,17 @@ def reset(self):
         self._type = None
         self._num_classes = None
 
+    def _process_shape(self, y):
```
vfdev-5: This method is ugly, sorry

vfdev-5: I would prefer not to squeeze at all and adapt our logic and raise an error if we can not understand the case.

anmolsjoshi: I updated the function. 

What logic would you use to check the shapes?

vfdev-5: I'm trying to analyse here. 
As documentation states:
```
    - `y_pred` must be in the following shape (batch_size, num_categories, ...) or (batch_size, ...).
    - `y` must be in the following shape (batch_size, ...).
```
we must have batch size of `y` and batch size of `y_pred` to be equal and all dimensions hided behind `...` to be equal for `y` and `y_pred`.

This logic should be sufficient ?

Next, we need to determine the case : multi-class or (binary/multi-label). I think `y.ndimension() + 1 == y_pred.ndimension()` should hold even if we do not squeeze 1's.

anmolsjoshi: @vfdev-5 I mostly agree. Here's a situation.

Let's take an example of a binary case of batch size N - model output shape is (N, 1) and training set shape is (N, ). Given logic of check_type, this case will be considered a multiclass case, because y_pred.ndimension() == y.ndimension() + 1.

Should we have users account for situations using output_transform? I personally think we should account for this situation in the metrics. Most updated way of handling 1's in the shape, I think works well.

vfdev-5: Or we can assign a multi-class case when second dimension is larger than one ?

anmolsjoshi: Clarifying, if a user inputs (N, 10) and (N, 1) for multiclass cases. That is considered wrong by the new logic right? We should raise an error for this?

vfdev-5: We can follow this decision "tree" (https://github.com/pytorch/ignite/pull/275#issuecomment-425396475) and update it to the current state.

anmolsjoshi:
@vfdev-5 please have a look at the updated branch. I changed the metrics, tests per your comment.

anmolsjoshi:
@vfdev-5 we currently don't have restrictions on that, do you want to throw an error or have the metric handle it?

vfdev-5:
That's fine without restrictions. I want just to make sure that such corner cases are well handled.

anmolsjoshi:
I updated the logic for check_type to handle this case, and updated how binary accuracy was calculated to prevent any errors. 

These are the errors I wanted to avoid:
1. _type is recorded as multiclass
2. the binary checks are not implemented.

ignite/metrics/accuracy.py
```diff
@@ -122,12 +121,12 @@ def reset(self):
...
     def update(self, output):
-
-        y_pred, y = self._check_shape(output)
+        y_pred, y = output
+        self._check_shape((y_pred, y))
         self._check_type((y_pred, y))
 
         if self._type == "binary":
-            correct = torch.eq(y_pred.type(y.type()), y).view(-1)
+            correct = torch.eq(y_pred.view(-1).type(y.type()), y.view(-1))
```
vfdev-5: Maybe we can change the following
```
y_pred.view(-1).type(y.type())
```
by 
```
y_pred.view(-1).to(y)
```
https://pytorch.org/docs/stable/tensors.html#torch.Tensor.to

anmolsjoshi: agreed and updated.

vfdev-5:
Joshi, can you please add the tests from the original issue such that the PR fixes the issue too

anmolsjoshi:
Victor, I added 1 example and another test for the new check_type logic.

Did you want me to add to tests (described in the issue) to all other test cases and other metrics?

vfdev-5:
I thinks it is OK for now. We can merge this once travis is green

PR #572, opened 2019-08-02
Fixes issue #543

### LINKED ISSUE #543: Problem with IoU
I'm training a FCN on the Cityscapes dataset. All ignored classes are mapped to 255. This works perfectly fine for the loss function using ignore_index.

Using the Ignite IoU metric however results in this error:

```/pytorch/aten/src/THC/THCTensorScatterGather.cu:188: void THCudaTensor_scatterFillKernel(TensorInfo<Real, IndexType>, TensorInfo<long, IndexType>, Real, int, IndexType) [with IndexType = unsigned int, Real = float, Dims = 2]: block: [36,0,0], thread: [0,0,0] Assertion `indexValue >= 0 && indexValue < tensor.sizes[dim]` failed.```

triggered here:
```
y_pred_ohe = to_onehot(indices.reshape(-1), self.num_classes)
  File "/usr/local/lib/python3.6/dist-packages/ignite/utils.py", line 48, in to_onehot
    onehot = torch.zeros(indices.shape[0], num_classes, *indices.shape[1:], device=indices.device)
RuntimeError: cuda runtime error (59) : device-side assert triggered at /pytorch/aten/src/THC/generic/THCTensorMath.cu:26
```

It's probably because there are only 19 classes and some values are 255.

### PR DESCRIPTION
**Pull request by vfdev-5**
PR fixes #543 and probably #554  

Description:
Previous CM implementation suffered from the problem if target contains non-contiguous indices.
New implementation is almost taken from torchvision's https://github.com/pytorch/vision/blob/master/references/segmentation/utils.py#L75-L117

In addition, this implementation is more robust when trained in fp16. Previous implementation used torch.matmul which is patched by nvidia/apex confusion matrix goes easily to inf/nan.

**!!! This commit also removes the case of targets as (batchsize, num_categories, ...)** where num_categories excludes background class.
Confusion matrix computation is possible almost similarly for (batchsize, ...), but when target is all zero (0, ..., 0)  = no classes (background class), then confusion matrix does not count any true/false predictions.

- PR breaks BC

Check list:

--- 

Target indices ouside of `[0, num_classes]` are ignored as a mask `(y >= 0) & (y < num_classes)` is applied. This can be generalized with a custom method `isin` method :
```python
import torch
a = torch.randint(0, 255, size=(10000, ))
b = list(range(5, 25))

def isin(x, y):
    max_value = x.max() + 1
    x_mask = torch.zeros(max_value, dtype=torch.uint8)
    x_mask[x] = 1
    y_mask = torch.zeros(max_value, dtype=torch.uint8)
    y_mask[y] = 1
    mask = x_mask & y_mask
    return mask[x]

mask = isin(a, b)
a[mask]
```

### DIFF
--- a/ignite/metrics/confusion_matrix.py
+++ b/ignite/metrics/confusion_matrix.py
@@ -4,7 +4,6 @@ import torch
 
 from ignite.metrics import Metric, MetricsLambda
 from ignite.exceptions import NotComputableError
-from ignite.utils import to_onehot
 
 
 class ConfusionMatrix(Metric):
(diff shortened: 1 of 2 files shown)


vfdev-5:
cc @TheCodez

PR #575, opened 2019-08-04
[WIP] Possibility to alter the input batch

vfdev-5:
Description:

Idea is to provide new functionality to alter the input batch of the processing function using an event handler attached on Events.ITERATION_STARTED

```python

trainer = ...

@trainer.on(Events.ITERATION_STARTED)
def switch_batch(engine):
    engine.state.batch = alter_batch(engine.state.batch)
```

This can be useful, if we would like to switch dataloader etc

Check list:


vfdev-5:
Close this PR in favor of #641

PR #588, opened 2019-08-23
Return removable handle from Engine.add_event_handler().

### LINKED ISSUE #586: [Feature Request] "Removable" event handlers.
_I'm happy to open a PR for this immediately if the feature seems useful._

Ignite's event interface provides both add and remove operations however the "remove" operation requires a reference to the original event handler. This makes temporary or one-off event handlers somewhat tricky to implement.

Pytorch's existing hook interface provides a solution to this issue by returning a "RemovableHandle" handle from `register_hook`, holding a weakref to the target hook to support later removal.

Ignite's `add_event_handler` should implement a similar interface, returning a "RemovableEventHandle" rather than `None`. For example, in an attrs-style syntax:

```python
@attr.s(auto_attribs=True)
class RemovableEventHandle:
    """Removable weak reference to an added event handler."""


    event_name: str
    handler: weakref.ref = attr.ib(converter=_as_weakref)
    engine: weakref.ref = attr.ib(converter=_as_weakref)

    def remove(self):
        """Remove handler from engine."""
        handler = self.handler()
        engine: Engine = self.engine()

        if handler and engine and engine.has_event_handler(handler):
            engine.remove_event_handler(handler, self.event_name)

    def __enter__(self):  # noqa: D
        return self

    def __exit__(self, type, value, tb):  # noqa: D
        self.remove()
```

Would support features like:

```python

def print_epoch(engine):
    print("Epoch: {}".format(engine.state.epoch))

# Train with a temporary event handler.
with engine.add_event_handler(Events.EPOCH_COMPLETED, print_epoch):
    engine.run(dataloader, 10)
```

asford:
Fixes #586

Description:

As per #586, return a weakref-based `RemovableEventHandle` from calls to `Engine.add_event_handler`. Handle is patterned on torch's `register_hook` interface, providing an explicit `remove()` method to remove the attached event as-well-as implementing the context manager protocol to provide support for `with`-statement based event scopes.

Check list:

### DIFF
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -103,7 +103,7 @@ reference returned by :meth:`~ignite.engine.Engine.add_event_handler`. This can 
         print("Epoch: {} - {} accuracy: {:.2f}"
                .format(trainer.state.epoch, title, engine.state.metrics['acc']))
 
-    @trainer.on(Events.EPOCH_COMPLETED):
+    @trainer.on(Events.EPOCH_COMPLETED)
     def evaluate(trainer):
         with evaluator.add_event_handler(Events.COMPLETED, log_metrics, "train"):
             evaluator.run(train_loader)
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -88,6 +88,33 @@ Attaching an event handler is simple using method :meth:`~ignite.engine.Engine.a
 
     trainer.add_event_handler(Events.COMPLETED, on_training_ended, mydata)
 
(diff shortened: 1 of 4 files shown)


vfdev-5:
@asford code looks good, thanks !
I thought we could provide some more useful examples of its usage in FAQ or somewhere. What do you think ?

asford:
I've added an example to the "concepts" doc section, matching the existing documentation for `add_event_handler`. @vfdev-5, can you take a peek and let me know if you think this is suitable?

PR #596, opened 2019-08-29
Issue 590 - FastaiLRFinder

### LINKED ISSUE #590: [FR] LR Finder
There is a repository which already has ignite version of [LRFinder](https://github.com/ItamarWilf/super-convergence/blob/master/super_convergence/lr_finder_engine.py#L13). It would be helpful to put this class to ignite/contrib/engines.

cc @ItamarWilf

ItamarWilf:
Fixes #590 

Description:
Added a handler to conrib.handlers that performs the learning rate find scheme implemented in fast.ai, on any trainer with a single model and single optimizer. The handler includes lr_suggestion and loss vs lr curve plot.

Also added an example notebook showing some improvement on MNIST training using lr_suggestion instead of default lr.

Check list:


ignite/contrib/handlers/lr_finder.py
```diff
@@ -0,0 +1,425 @@
...
+        self.detach()
+
+
+class AlreadyAttachedError(Exception):
+    """
+    Exception class to raise if trying to attach an already attached LRFinder.
+    """
+
+
+class NotEnoughIterationWarning(Warning):
```
vfdev-5: Previously, we used `UserWarning` as warning types, I think can go without custom warnings.

ignite/contrib/handlers/lr_finder.py
```diff
@@ -0,0 +1,425 @@
...
+    """
+
+
+class NotDivergedWarning(Warning):
+    """
+    Warning class to raise when loss doesn't diverge during lr_finder search.
+    """
+
+
+class _LinearLR(_LRScheduler):
```
vfdev-5: Can we use https://github.com/pytorch/ignite/blob/01f2f5f96f55b8358edfe280dc6852bbaf807425/ignite/contrib/handlers/param_scheduler.py#L618 or https://github.com/pytorch/ignite/blob/01f2f5f96f55b8358edfe280dc6852bbaf807425/ignite/contrib/handlers/param_scheduler.py#L173 for this ?

ignite/contrib/handlers/lr_finder.py
ItamarWilf: No Problem

ignite/contrib/handlers/lr_finder.py
ItamarWilf: I think so, I'll look into it

ignite/contrib/handlers/lr_finder.py
```diff
@@ -0,0 +1,425 @@
...
+            lr_finder = LRFinder(model, optimizer)
+
+            # Attach the lr_finder to the trainer
+            lr_finder.attach(trainer)
+
+            # Run trainer
+            trainer.run(dataloader)
+
+            # Detach lr_finder
+            lr_finder.detach()
```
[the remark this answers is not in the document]
ItamarWilf: I think we can detach the lr_finder automatically.

Anyway, right now the user can't use the `lr_finder` multiple times with different arguments since they are all passed during `__init__`. maybe we can move some of those parameters to `attach` instead.

ignite/contrib/handlers/lr_finder.py
```diff
@@ -0,0 +1,425 @@
...
+        self._output_transform = output_transform
+        self.num_iter = num_iter
+        self._end_lr = end_lr
+        self._step_mode = step_mode
+        self._smooth_f = smooth_f
+        self._diverge_th = diverge_th
+        self._memory_cache = memory_cache
+        self._cache_dir = cache_dir
+
+        self._engine = None
```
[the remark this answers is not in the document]
ItamarWilf: Sure, it can be remove for now.

ignite/contrib/handlers/lr_finder.py
ItamarWilf: Done, used `PiecewiseLinear` instead.

ignite/contrib/handlers/lr_finder.py
ItamarWilf: Done

examples/contrib/notebooks/FastaiLRFinder_MNIST.ipynb
```diff
@@ -0,0 +1,259 @@
...
+    "%autoreload 2\n",
+    "\n",
+    "import torch.nn as nn\n",
+    "import torch.nn.functional as F\n",
+    "import torch.optim as optim\n",
+    "import torchvision.transforms as transforms\n",
+    "from torch.utils.data import DataLoader\n",
+    "from torchvision.datasets import MNIST\n",
+    "\n",
+    "import sys\n",
```
vfdev-5: Maybe this is not necessary if ignite is installed with pip

examples/contrib/notebooks/FastaiLRFinder_MNIST.ipynb
```diff
@@ -0,0 +1,259 @@
...
+   "execution_count": 1,
+   "metadata": {
+    "pycharm": {
+     "is_executing": false
+    }
+   },
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
+    "%reload_ext autoreload\n",
```
vfdev-5: Autoreload can be skipped in the final version :)

ignite/contrib/handlers/lr_finder.py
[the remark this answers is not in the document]
vfdev-5: Actually, detach can be purely internal method, IMO. Once, attached we can add detach on `Events.COMPLETED` and execute it automatically in all cases (context manager or not).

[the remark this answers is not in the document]
ItamarWilf: OK, no problem, I will change detach to be purely internal. If we do it than the context manger won't really do anything, so it can be removed as well. sounds ok?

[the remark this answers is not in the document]
vfdev-5: @ItamarWilf I thought more about this, and maybe should keep only context manager API, such that it is more clear for the user that all attachments are removed outside the context. Otherwise in the following code 
```python
# Attach the lr_finder to the trainer
lr_finder.attach(trainer)
# Run trainer
trainer.run(dataloader)
```
it is not clear that all added handlers by lr finder are removed after the run. Thoughts ?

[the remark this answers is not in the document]
ItamarWilf: Sounds good, "Explicit is better than implicit" :)
I will keep detach as an external method and remove the automatic detaching from `Event.COMPLETED`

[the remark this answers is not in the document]
vfdev-5: Actually, I thought about providing the API only with context manager and without `detach`
```python
with lr_finder.attach(trainer):
    trainer.run(dataloader)
```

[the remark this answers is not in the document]
ItamarWilf: OK, I will try to find a way to enforce usage of the `attach` method only from within a context manager.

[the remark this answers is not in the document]
ItamarWilf: Well, it seems there isn't a clean way to enforce usage of `attach` with a `with` statement. 
I suggest using `attach` just to clarify to which engine, model and optimizer the lr_finder is connected, while all the actual handler attaching and detaching happens inside the `with` statement. For example:
```python
lr_finder.attach(dummy_engine, model, optimizer)
with lr_finder:
    dummy_engine.run(dataloader)
```
This way we can keep `detach` as an internal method and we enforce the handler clean up and  model + optimizer reset.

Thoughts?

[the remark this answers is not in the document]
vfdev-5: @ItamarWilf thanks for the proposition, let me think more about the API and we'll decide which is more convenient...

marrrcin:
What is the status of this? Is it going to be integrated?

vfdev-5:
@marrrcin yes, it will be integrated. Currently, there are some API problems that blocks the advances...

marrrcin:
Is there anything that you would like to have help with?

vfdev-5:
@marrrcin well, a help could be to think and propose intuitive API. As far as I remember, the problem for me is to make clear API about how to attach and detach handlers from lr finder. Personnally, I'm not fan of the option with explicit `lr_finder.attach(trainer)` and `lr_finder.detach()` and prefer a context manager version:
```python
with lr_finder.attach(trainer):
    trainer.run(dataloader)
```
however it seems it is not possible to enforce such usage. Internal implementation of `FastaiLRFinder` seems to me can be also improved... Issue is not that simple.

marrrcin:
Shouldn't thing like `LRFinder` be a separate abstraction like the trainer and evaluator instead being a part of the trainer?

vfdev-5:
@marrrcin it is a part of handlers. See #590 for discussions on this subject/

ignite/contrib/handlers/lr_finder.py
```diff
@@ -0,0 +1,405 @@
...
+    def _reached_num_iterations(self, engine, num_iter):
+        if engine.state.iteration > num_iter:
+            engine.terminate()
+
+    def _warning(self, engine):
+        if not self._diverge_flag:
+            warnings.warn("Run completed without loss diverging, increase end_lr, decrease diverge_th or look"
+                          " at lr_finder.plot()", UserWarning)
+
+    def attach(self, engine, model, optimizer, output_transform=lambda output: output, num_iter=None, end_lr=10,
```
justusschock: Shouldn't it be possible to achieve the context manager behaviour with the `context manager` decorator from [contextlib](https://docs.python.org/3/library/contextlib.html) like the following:

```python
from context lib import contextmanager

@contextmanager
def attach(self, engine, model, optimizer, output_transform=lambda output: output, num_iter=None, end_lr=10, step_mode="exp", smooth_f=0.05, diverge_th=5):
# DO ALL THE ATTACHING STUFF
    yield self
    self.detach(self._engine)

Comment (vfdev-5) [reply ignite/contrib/handlers/lr_finder.py]: @justusschock could you please test if we could achieve that with contextlib ?
Yes, we can yield engine and detail it as :
```python
with lr_finder.attach(trainer) as trainer_to_find_lr:
    trainer_to_find_lr.run(dataloader)
```

ItamarWilf: Oh, didn't know about `contextlib` until now... Thanks @justusschock !!!
I tried it now and it seems to work. I can finish it and fix the tests and documentation if you want

vfdev-5: @ItamarWilf yes, please go ahead

ignite/contrib/handlers/lr_finder.py
```diff
@@ -0,0 +1,390 @@
+# coding: utf-8
+
+from __future__ import division
```
vfdev-5: We can now remove this line since no more support for 2.7

ItamarWilf: No problem

ignite/contrib/handlers/lr_finder.py
```diff
@@ -0,0 +1,389 @@
...
+
+        .. code-block:: python
+
+            from ignite.contrib.handlers import LRFinder
+
+            # Create a lr_finder
+            lr_finder = LRFinder(model, optimizer)
+
+            # Attach the lr_finder to the trainer
+            with lr_finder.attach(trainer, model, optimizer) as
```
vfdev-5:
```suggestion
            with lr_finder.attach(trainer, model, optimizer) as trainer_with_lr_finder
```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -0,0 +1,389 @@
...
+            trainer_with_lr_finder:
```
vfdev-5:
```suggestion
```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -0,0 +1,389 @@
...
+
+    References:
+        Cyclical Learning Rates for Training Neural Networks:
+        https://arxiv.org/abs/1506.01186
+
+        fastai/lr_find: https://github.com/fastai/fastai
+    """
+
+    def __init__(self, memory_cache=True, cache_dir=None):
+        try:
```
vfdev-5: IMO, it would be better to move this to `plot` method.

vfdev-5:
@ItamarWilf sorry for delay on this PR. I simplified the implementation. Remains tests to adapt and it will be merged then...

vfdev-5:
@justusschock can you please review ?

cc @ItamarWilf

ignite/contrib/handlers/lr_finder.py
```diff
@@ -0,0 +1,366 @@
...
+    While attached, the handler increases the learning rate in between two
+    boundaries in a linear or exponential manner. It provides valuable
+    information on how well the network can be trained over a range of learning
+    rates and what can be an optimal learning rate.
+
+    Examples:
+
+    .. code-block:: python
+
+        from ignite.contrib.handlers import LRFinder
```
justusschock: This is inconsistent with the actual class name of `FastaiLRFinder`

examples/contrib/notebooks/FastaiLRFinder_MNIST.ipynb
```diff
@@ -0,0 +1,677 @@
...
+    "criterion = nn.NLLLoss()\n",
+    "model = Net()\n",
+    "optimizer = optim.SGD(model.parameters(), lr=3e-4, momentum=0.9)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [
```
justusschock: Can we remove the state of this notebook from git?

vfdev-5: What do you mean by the state ? Outputs ?

ignite/contrib/handlers/lr_finder.py
vfdev-5: Nice catch, will fix!

examples/contrib/notebooks/FastaiLRFinder_MNIST.ipynb
justusschock: Yes, outputs and execution count

vfdev-5: I think it is intersting to see the results of the execution, even if jupyter widgets are badly rendered on github...

justusschock: The point why I ask to remove them is not that they are badly rendered (although this is true as well), but that results increase the repository size. And since they can basically be or arbitrary size, I would prefer to have no results at all in this repo and maybe link to colab instead... But if you want to keep them, that's fine

vfdev-5: That's a good point! Let's keep them for now, but remove everywhere once colab are ready.

vfdev-5:
Finally merged with our common efforts! Thanks @ItamarWilf for the PR and sorry for such huge delay !

PR #599, opened 2019-08-30
Deploy nightly builds as prerelease (#595)

### LINKED ISSUE #595: Update ignite's master nightly builds deployment
Following 
> Since torch-nightly is now deprecated (and stuck at 2019-08-05) in favour of using pre-releases of torch, ignite nighlty should stop depending on it.

I think we can go the same way as pytorch. Ignite's master nightly builds on pypi can be similar as pytorch with `--pre` and we will need to pass to `pytorch-nightly` channel for conda. 

On both I think it'll be user's responisibility to install correct pytorch with `--pre` or `-c pytorch-nightly`. As `pip install --pre pytorch-ignite`/`conda install ignite -c pytorch-nightly` will install master ignite which depends on any torch/pytorch.

cc @Evpok

Evpok:
Fixes #595 

Description: change the nightly job to release from master branch

- On Pypi as ignite-x.x.x.devYYYYMMDD
- On conda via the pytorch-nightly channel


vfdev-5:
@Evpok thank you for the PR !

A small fix for conda is needed in 187 
```
- ls conda_build/*/*.tar.bz2 | xargs -I {} anaconda -v -t $ANACONDA_TOKEN upload -u pytorch {}
```
should be 
```
- ls conda_build/*/*.tar.bz2 | xargs -I {} anaconda -v -t $ANACONDA_TOKEN upload -u pytorch-nightly {}
```

Then I need to ask @soumith to have be able to upload to this channel.

.travis.yml
```diff
@@ -169,23 +166,13 @@ jobs:
       python: "3.6"
       env:
         - PYTORCH_CHANNEL=pytorch-nightly
-      if: branch = nightly
```
vfdev-5: We still need to keep if cron otherwise every PR, push etc there is a Nightly stage : https://travis-ci.org/pytorch/ignite/builds/578850726?utm_source=github_status&utm_medium=notification

It should be something to deal with if TRAVIS_EVENT_TYPE=cron : https://docs.travis-ci.com/user/cron-jobs/#detecting-builds-triggered-by-cron

Maybe something like 
```
env(TRAVIS_EVENT_TYPE) = cron
```

vfdev-5:
A question on PyPI deployement and what defines that we can fetch this version with 
```
pip install --pre ignite
```
?

.travis.yml
Evpok: Yes, that's why this PR is still a draft, I still have to figure out if a cron job can pass more information apart that it is a cron. Of course if the only cron there is for building nightly this could be enough, but I find it irking :-)

Evpok:
The ".dev…" segment of the version identifier as per https://www.python.org/dev/peps/pep-0440/#public-version-identifiers and https://pip.pypa.io/en/stable/reference/pip_install/#pre-release-versions

.travis.yml
vfdev-5: I see, maybe we can add another env variable as 
```
      env:
        - NIGHTLY_BUILD=1
```
and check if it is present

Evpok: Well apparently you can't set an environment variable differently in a cron job in Travis yet, so I guess we'll have to make do with just checking the event type indeed :-(

Evpok:
Thanks for you patience :-)

vfdev-5:
@Evpok I merged the PR to the master. Let's check the deployment on the next cron job.
Soumith created ignite package at conda : https://anaconda.org/pytorch-nightly/ignite

Once everything is working as expected we can update README.rst and documentation to inform users.

vfdev-5:
@Evpok seems like `if: env(TRAVIS_EVENT_TYPE) = cron` does not work : https://travis-ci.org/pytorch/ignite/builds/579720684

vfdev-5:
PyPI upload does not work : 
```
HTTPError: 400 Client Error: '0.2.0.dev0+20190903t072040' is an invalid value for Version. Error: Can't use PEP 440 local versions. See https://packaging.python.org/specifications/core-metadata for url: https://upload.pypi.org/legacy/
```

vfdev-5:
@Evpok could you please investigate where is the problem.

Evpok:
Ah, it is because of the `T` of the timestamp. I'll remove it, we don't need more than one release per day anyway

PR #602, opened 2019-09-02
TensorboardLogger init - forwarding kwargs to SummaryWriter

### LINKED ISSUE #597: TensorboardLogger init - forwarding kwargs to SummaryWriter
### Problem 
When a TensorboardLogger object is created, it also initialises a SummaryWriter object. 
At the moment, the TensorboardLogger init function accepts the only argument `log_dir` that is passed to SummaryWriter but SummaryWriter accepts a lot more keyword arguments I want to make use of. 

[Link to current implementation](https://github.com/pytorch/ignite/blob/01f2f5f96f55b8358edfe280dc6852bbaf807425/ignite/contrib/handlers/tensorboard_logger.py#L402)
### Solution/Proposal
Forwarding all positional and keyword arguments to SummaryWriter. 

Would this also solve the tensorboardX compatibility issue #538?

-----------
Are there any restrictions I missed, why this is not implemented? 
If wanted, I can create a pull request.

Bibonaut:
Fixes #597 

Description:
Class TensorboardLogger creates a SummaryWriter object during initialization but does not allow to pass keyword arguments other than `log_dir`. Now, TensorboadLogger fowards all positional and keyword arguments to SummaryWriter to enable full configuartion.

Check list:

### DIFF
--- a/ignite/contrib/handlers/tensorboard_logger.py
+++ b/ignite/contrib/handlers/tensorboard_logger.py
@@ -322,7 +322,8 @@ class TensorboardLogger(BaseLogger):
 
 
     Args:
-        log_dir (str): path to the directory where to log.
+        *args: Positional arguments accepted from :class:`~tensorboardx.SummaryWriter`.
+        **kwargs: Keyword arguments accepted from :class:`~tensorboardx.SummaryWriter`.
 
     Examples:
 
--- a/ignite/contrib/handlers/tensorboard_logger.py
+++ b/ignite/contrib/handlers/tensorboard_logger.py
@@ -324,7 +324,7 @@ class TensorboardLogger(BaseLogger):
     Args:
         *args: Positional arguments accepted from :class:`~tensorboardx.SummaryWriter`.
         **kwargs: Keyword arguments accepted from :class:`~tensorboardx.SummaryWriter`, for example,
-        `log_dir` to setup path to the directory where to log.
+            `log_dir` to setup path to the directory where to log.
 
     Examples:
 
--- a/ignite/contrib/handlers/tensorboard_logger.py
+++ b/ignite/contrib/handlers/tensorboard_logger.py
@@ -399,23 +399,14 @@ class TensorboardLogger(BaseLogger):
 
     """
 
-    def __init__(self, log_dir):
+    def __init__(self, *args, **kwargs):
         try:
             from tensorboardX import SummaryWriter
         except ImportError:
             raise RuntimeError("This contrib module requires tensorboardX to be installed. "
                                "Please install it with command: \n pip install tensorboardX")
 
-        try:
-            self.writer = SummaryWriter(logdir=log_dir)
-        except TypeError as err:
-            if "type object got multiple values for keyword argument 'logdir'" == str(err):
-                self.writer = SummaryWriter(log_dir=log_dir)
-                warnings.warn('tensorboardX version < 1.7 will not be supported '
-                              'after ignite 0.3.0; please upgrade',
-                              DeprecationWarning)
-            else:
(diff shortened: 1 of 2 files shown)


vfdev-5:
@Bibonaut thanks for the PR. I think we also need to update docsting part `Args:` something like:
```
Args:
    *args: positional arguments for `SummaryWriter`, for example, `log_dir` to setup path to the directory where to log.
    **kwargs: kwargs for `SummaryWriter`
```

We need also to remove [`test_init_tb1p6`](https://github.com/pytorch/ignite/blob/master/tests/ignite/contrib/handlers/test_tensorboard_logger.py#L567).

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -322,7 +322,8 @@ class TensorboardLogger(BaseLogger):
 
 
     Args:
-        log_dir (str): path to the directory where to log.
+        *args: Positional arguments accepted from :class:`~tensorboardx.SummaryWriter`.
```
vfdev-5:
```suggestion
        *args: Positional arguments accepted from :class:`~tensorboardx.SummaryWriter`, for example, `log_dir` 
            to setup path to the directory where to log.
```

Bibonaut: Thanks for your very fast reply :) Right at the moment, I was searching the web on how to write documentation for args and kwargs.

Bibonaut: Shouldn't the `log_dir` hint be added to the kwargs description?
```python
Args:
        *args: Positional arguments accepted from :class:`~tensorboardx.SummaryWriter`.
        **kwargs: Keyword arguments accepted from :class:`~tensorboardx.SummaryWriter`, for example, `log_dir` to setup path to the directory where to log.
```

Bibonaut:
@vfdev-5  As far as I understand, this test must be removed:

```python
def test_init_tb1p6(mock_tb_module):

    def side_effect_v1p6(*args, **kwargs):
        if 'logdir' in kwargs:
            raise TypeError("type object got multiple values for keyword argument 'logdir'")

    mock_tb_module.SummaryWriter = Mock(name='tensorboardX.SummaryWriter', side_effect=side_effect_v1p6)

    with pytest.warns(DeprecationWarning, match=r'tensorboardX version < 1.7 will not be supported'):
        TensorboardLogger(log_dir=None)
```

But I am a bit unsure, because I do not understand, what effect `side_effect_v1p6` has on this test since it seems to be unused?

ignite/contrib/handlers/tensorboard_logger.py
vfdev-5: @Bibonaut yes, you are right, we can add it to kwargs

vfdev-5:
@Bibonaut I agree, we can remove `test_init_tb1p6` test.

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -323,7 +323,8 @@ class TensorboardLogger(BaseLogger):
 
     Args:
         *args: Positional arguments accepted from :class:`~tensorboardx.SummaryWriter`.
-        **kwargs: Keyword arguments accepted from :class:`~tensorboardx.SummaryWriter`.
+        **kwargs: Keyword arguments accepted from :class:`~tensorboardx.SummaryWriter`, for example,
+        `log_dir` to setup path to the directory where to log.
```
vfdev-5: You need to indent this line 4 spaces in order to continue as the previous in the rendered doc

vfdev-5:
@Bibonaut thanks !

PR #606, opened 2019-09-04
[WIP] TensorboardLogger - tag/prefix for weights, grads and optimizer logging

### LINKED ISSUE #598: TensorboardLogger - tag/prefix for weights, grads and optimizer logging
### Problem
I have to log multiple models/optimizers with TensorboardLogger and I want them to be displayed in tensorboard in separate *subpanels* with distinct names/tags. 
At the moment, it is not possible to provide a custom tag or prefix for some handlers that log model parameters, because the tag is deduced from `model.named_parameters()`. 

As a consequence, parameters of multiple models are plottet to one graph.

### Solution
The affected handlers allow passing a custom prefix/tag to the function that is prepended to the deduced tag. 

E.g. for `GradsHistHandler` (snippet):
```python
logger.writer.add_histogram(tag="grads/{}".format(name),
                                        values=p.grad.detach().cpu().numpy(),
                                        global_step=global_step)
```

It can be changed to 
```python
logger.writer.add_histogram(tag=self.prefix + "grads/{}".format(name),
                                        values=p.grad.detach().cpu().numpy(),
                                        global_step=global_step)
```

Affected handlers in `ignite.contrib.handlers.tensorboard_logger`:
- ~~`OptimizerParamsHandler`~~
- `WeightsScalarHandler`
- `WeightsHistHandler`
- `GradsScalarHandler`
- `GradsHistHandler`

What do you think? 
If wanted, I can work on this.

Bibonaut:
Fixes #598  

Description:
Introduces optional tags for weights and gradients handlers for tensorboard logging. Tag property is added to base class handlers. 

Check list:

### DIFF
--- a/ignite/contrib/handlers/base_logger.py
+++ b/ignite/contrib/handlers/base_logger.py
@@ -130,7 +130,7 @@ class BaseWeightsScalarHandler(BaseHandler):
     Helper handler to log model's weights as scalars.
     """
 
-    def __init__(self, model, reduction=torch.norm):
+    def __init__(self, model, reduction=torch.norm, tag=None):
         if not isinstance(model, torch.nn.Module):
             raise TypeError("Argument model should be of type torch.nn.Module, "
                             "but given {}".format(type(model)))
@@ -149,6 +149,7 @@ class BaseWeightsScalarHandler(BaseHandler):
 
         self.model = model
         self.reduction = reduction
+        self.tag = tag
 
 
 class BaseWeightsHistHandler(BaseHandler):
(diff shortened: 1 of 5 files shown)


Bibonaut:
@vfdev-5 What is the best way to create tests with optional tag but without full duplication of affected existing tests? Do you have a good idea? Otherwise I would just clone them and add optional tags.

vfdev-5:
@Bibonaut Visdom logger reimplements `BaseWeightsScalarHandler`, could you please add `tag` to its `WeightsScalarHandler` and `GradsScalarHandler`.

About tests factorization, maybe a solution like that could work :
```
def test_something():
   
    def _test(tag):
         # test with/without tag

    _test(tag=None)
    _test(tag="some tag")
```

Bibonaut:
@vfdev-5 I've added tags to Visdom logger and added tests for Tensorboard logger and Visdom logger. Is there anything to do?

vfdev-5:
@Bibonaut I think it is good for the issue. Thanks for the PR !

PR #605, opened 2019-09-03
Fix Pypi nightly releases

Evpok:
Picked up from #599

PR #607, opened 2019-09-04
Fix pypi deployment for good

Evpok:
Follows from #605 

This uploads successfully on test pypi: https://test.pypi.org/project/pytorch-ignite/0.2.0.dev20190904 so fingers crossed. Sorry again for the fumbling, I really wish there was an easier way to test these things.


vfdev-5:
it worked 🎉

Evpok:
Perfect, we just need a bump of ignite's version (since 0.2.0 > 0.2.0.devN for pip, which is a bit absurd imo). This can be 0.2.0.post1 if you don't want to chose what type of release the next one should be, I made some tests on test pypi, see https://test.pypi.org/project/pytorch-ignite/#history

vfdev-5:
Actually, I already bumped the version to 0.3.0 to be similar to pytorch :) and restarted cron

Evpok:
Perfect !

PR #608, opened 2019-09-04
Added nightly releases info

vfdev-5:
Updated README and docs index.rst

cc @Evpok


Evpok:
Why did I not see this. I swear I work with half a brain these days, now you have a choice with #609

PR #609, opened 2019-09-04
Announce new nightlies

Evpok:
As requested, to complete the nightly transition.

BTW is there a way to preview the generated doc?


vfdev-5:
Thanks for the PR, but I think to merge #608

PR #612, opened 2019-09-06
WeightsScalarHandler-random-side-effects

### LINKED ISSUE #611: BaseWeightsScalarHandler has side effects on random number generation
### Problem
BaseWeightsScalarHandler uses `torch.rand` to test the reduction function that is passed as an argument. 
This is very confusing, because experiments are not reproducible when tensorboard logging is switched on or off, e.g. in debug sessions. 

### Solution
Replace the random tensor with a non random tensor.

Bibonaut:
Fixes #611  

Description:
Replaces the random tensor with a non random tensor (ones) to ensure experiment reproducibility whether `BaseWeightsScalarHandler` is used or not. 

Check list:

### DIFF
--- a/ignite/contrib/handlers/base_logger.py
+++ b/ignite/contrib/handlers/base_logger.py
@@ -143,7 +143,7 @@ class BaseWeightsScalarHandler(BaseHandler):
             return isinstance(t, torch.Tensor) and t.ndimension() == 0
 
         # Test reduction function on a random tensor
-        o = reduction(torch.rand(4, 2))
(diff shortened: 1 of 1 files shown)


ignite/contrib/handlers/base_logger.py
```diff
@@ -143,7 +143,7 @@ def _is_0D_tensor(t):
             return isinstance(t, torch.Tensor) and t.ndimension() == 0
 
         # Test reduction function on a random tensor
```
Bibonaut: I will fix the comment tomorrow ;)

PR #617, opened 2019-09-09
contrib|add tensorboardX import fallback

### LINKED ISSUE #553: support for torch.utils.tensorboard when using `tensorboard_logger`
With PyTorch 1.1.0, TensorBoard is now [natively supported](https://pytorch.org/docs/stable/tensorboard.html) in PyTorch. 

Can we add a new option to switch `torch.utils.tensorboard` or `tensorboardX`?

I found that `tensorboard_logger` only need `SummaryWriter` to implement `add_histogram` and `add_scalar`. It seems that we can just import `torch.utils.tensorboard` when import tensorboardX failed.

```python
try:
    from tensorboardX import SummaryWriter
except ImportError:
    from torch.utils.tensorboard import SummaryWriter
```

For now, tensorboardX:
> Support scalar, image, figure, histogram, audio, text, graph, onnx_graph, embedding, pr_curve, mesh and video summaries.

`torch.utils.tensorboard`:
> Scalars, images, histograms, graphs, and embedding visualizations are all supported for PyTorch models and tensors as well as Caffe2 nets and blobs.

bosr:
Fixes #553 

Description: If TensorboardX is not present, then fall back on importing PyTorch's
new SummaryWriter. 

Check list:
* [x] New tests are added (no new feature is added)

### DIFF
--- a/ignite/contrib/handlers/tensorboard_logger.py
+++ b/ignite/contrib/handlers/tensorboard_logger.py
@@ -414,7 +414,12 @@ class TensorboardLogger(BaseLogger):
         try:
             from tensorboardX import SummaryWriter
         except ImportError:
-            from torch.utils.tensorboard import SummaryWriter
+            try:
+                from torch.utils.tensorboard import SummaryWriter
+            except ModuleNotFoundError:
+                raise RuntimeError("This contrib module requires either tensorboardX or torch >= 1.2.0. "
+                                   "You may install tensorboardX with command: \n pip install tensorboardX \n"
+                                   "or upgrade PyTorch using your package manager of choice (pip or conda).")
 
         self.writer = SummaryWriter(*args, **kwargs)
 
--- a/ignite/contrib/handlers/tensorboard_logger.py
+++ b/ignite/contrib/handlers/tensorboard_logger.py
@@ -322,13 +322,13 @@ class TensorboardLogger(BaseLogger):
     """
     TensorBoard handler to log metrics, model/optimizer parameters, gradients during the training and validation.
 
-    This class favors `tensorboardX <https://github.com/lanpa/tensorboardX>`_ package if installed:
+    By default, this class favors `tensorboardX <https://github.com/lanpa/tensorboardX>`_ package if installed:
 
     .. code-block:: bash
 
         pip install tensorboardX
 
-    else, it falls back to using PyTorch's new SummaryWriter.
+    otherwise, it falls back to using PyTorch's SummaryWriter (>=v1.2.0).
 
     Args:
         *args: Positional arguments accepted from :class:`~tensorboardx.SummaryWriter`.
--- a/ignite/contrib/handlers/tensorboard_logger.py
+++ b/ignite/contrib/handlers/tensorboard_logger.py
@@ -416,7 +416,7 @@ class TensorboardLogger(BaseLogger):
         except ImportError:
             try:
                 from torch.utils.tensorboard import SummaryWriter
-            except ModuleNotFoundError:
+            except ImportError:
                 raise RuntimeError("This contrib module requires either tensorboardX or torch >= 1.2.0. "
                                    "You may install tensorboardX with command: \n pip install tensorboardX \n"
                                    "or upgrade PyTorch using your package manager of choice (pip or conda).")
--- a/ignite/contrib/handlers/tensorboard_logger.py
+++ b/ignite/contrib/handlers/tensorboard_logger.py
@@ -322,12 +322,13 @@ class TensorboardLogger(BaseLogger):
     """
     TensorBoard handler to log metrics, model/optimizer parameters, gradients during the training and validation.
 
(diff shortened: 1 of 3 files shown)


bosr:
Note: I took the liberty to remove the `test_no_tensorboardX` test as it did not seem relevant anymore. As an alternative, we could check that the `SummaryWriter` class is indeed from PyTorch?

vfdev-5:
@bosr thanks for the PR ! 

> As an alternative, we could check that the SummaryWriter class is indeed from PyTorch?

Yes, I think we will need this, otherwise coverage wont be "happy" with the code coverage delta...

vfdev-5:
@bosr do you need any help with tests ?

bosr:
Ah ah, why not ;) I’ll write to you this evening. You saw my multiple failures on my fake branch? 

Romain

vfdev-5:
@bosr no problems.

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -413,8 +414,7 @@ def __init__(self, *args, **kwargs):
         try:
             from tensorboardX import SummaryWriter
         except ImportError:
-            raise RuntimeError("This contrib module requires tensorboardX to be installed. "
-                               "Please install it with command: \n pip install tensorboardX")
+            from torch.utils.tensorboard import SummaryWriter
```
vfdev-5: Actually, thinking more about this, maybe we should warn user instead of `ImprotError` if it uses `torch<1.2.0`

bosr:
Hi, I've just rebased and included some support for the test.

Basically, my problem is that, if I re-establish the test, like for instance:

    with TensorboardLogger(log_dir=dirname) as tb_logger:
        assert(isinstance(tb_logger.writer, torch.utils.tensorboard.SummaryWriter), \
            "in the absence of tensorboardX, SummaryWriter should come from torch.utils.tensorboard"

then there is another `ImportError` about missing `tensorboard` module in `torch.utils`. I'm honestly surprised by this error. You can see it here: https://travis-ci.org/bosr/ignite/jobs/583263107

Do you have an idea for fixing this? Or do you prefer to cancel this and just write a warning?

Thanks for your support!

ignite/contrib/handlers/tensorboard_logger.py
bosr: Yes, I've just updated the PR with what I tried to go further with testing the fallback, but this indeed may be sufficient.

vfdev-5: Here, I would like to cover quickly the case when users are using `torch==1.0.1`  for example:
```
# python -c "import torch; print(torch.__version__)"
1.0.1
# python -c "from torch.utils.tensorboard import SummaryWriter"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'torch.utils.tensorboard'
```
Instead of `ModuleNotFoundError` we could inform user to install tensorboardX or upgrade torch to the latest.

vfdev-5:
> there is another ImportError about missing tensorboard module in torch.utils. I'm honestly surprised by this error. 

This is because `tensorboard` is not a declared module : https://github.com/pytorch/pytorch/blob/master/torch/utils/__init__.py

Maybe you can do something like : 
```python

from torch.utils.tensorboard import SummaryWriter
with TensorboardLogger(log_dir=dirname) as tb_logger:
        assert isinstance(tb_logger.writer, SummaryWriter), type(tb_logger.writer)
```

bosr:
Oops, I got confused with the import, thanks. I’ll update the PR.

bosr:
I've finally installed the test suite locally, so I'll keep iterating and come back here. Thanks!

bosr:
Thanks a lot for the fixes! The code is nicer now, and one more case covered. This incentivized me to push it a little bit further and reduce to:

```python
def test_no_tensorboardX_package(dirname):
    from torch.utils.tensorboard import SummaryWriter
    with patch.dict('sys.modules', {'tensorboardX': None}):
        tb_logger = TensorboardLogger(log_dir=dirname)
        assert isinstance(tb_logger.writer, SummaryWriter), type(tb_logger.writer)

def test_no_torch_utils_tensorboard_package(dirname):
    from tensorboardX import SummaryWriter
    with patch.dict('sys.modules', {'torch.utils.tensorboard': None}):
        tb_logger = TensorboardLogger(log_dir=dirname)
        assert isinstance(tb_logger.writer, SummaryWriter), type(tb_logger.writer)


def test_no_tensorboardX_nor_torch_utils_tensorboard():
    with patch.dict('sys.modules', {'tensorboardX': None, 'torch.utils.tensorboard': None}):
        with pytest.raises(RuntimeError, match=r'This contrib module requires either tensorboardX or torch'):
            TensorboardLogger(log_dir=None)
```
replacing all related fixtures. Locally it passes on both 2.7 and 3.6+. Let's see if CI agrees ;)

I kept the previous code as comments for discussion. Tell me what you think about it.

PR #621, opened 2019-09-10
Fixes #620

### LINKED ISSUE #620: Improve TQDM, progress bar value formatting
Idea is to use tqdm's value renderer instead of using current implementation. Newer version will looks like:
```python
import tqdm
with tqdm.tqdm() as pbar:
    pbar.set_postfix({'a': 123, 'b': 12.3456788, 'd': 0.12345, 'c': torch.tensor(123).item(), 'e': 'text', 'f': lambda x: x})
```
out:
```
0it [00:00, ?it/s, a=123, b=12.3, d=0.123, c=123, e=text, f=<function <lambda> at 0x1234>]
```

This will help to better display integer values as `123` instead of `1.23e+02`.

cc @miguelvr

vfdev-5:
Uses tqdm's value renderer instead of ours

Fixes #620

Check list:

### DIFF
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import numbers
 import warnings
 
 import torch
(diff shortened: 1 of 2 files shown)


vfdev-5:
@miguelvr thanks for the review !

PR #624, opened 2019-09-14
Proposal - Support of periodic event handler execution

Bibonaut:
Fixes nothing -> proposal

Description:
I want to propose and discuss an enhancement to the `ignite.engine.Engine` class to support intervalled event handler execution. I am commonly facing the need to execute event handlers not every event but every *n-th* event, e.g. when
- the model is large and you want to log the model every n-th epoch or n-th iteration to keep log files smaller
- you are working on a remote machine and want to sync your logs every n-th epoch
- ....

The current solution is to wrap the handler into an intervalled executed handler with something like this (not tested):
```python
def handler_interval_wrapper(handler, interval=1):
    def _intervalled_handler(engine, *args, **kwargs):
        if (engine.State.epoch-1) % interval == 0:
            handler(engine,*args,**kwargs)
    return _intervalled_handler
```

Another solution is to use `ignite.contrib.handlers.custom_events.CustomPeriodicEvent` which requires much more code to achieve the same behaviour and requires code changes as changing the period/interval changes the event name. 

To make things easier, I think that `add_event_handler` can be enhanced to accept a keyword argument (e.g. named `interval`), that is stored in `Engine._event_handlers` and is used in `Engine._fire_event`. 
* This enables to add custom defined execution intervals to every event handler and is backward compatible except that the keyword argument `interval` is reserved now. 
* For custom events, it requires to register events and provide the `event_to_attr` dict. If not, fallback solution is to ignore `interval` and execute the handler every time the event occurs.

**Comparison of using `CustomPeriodicEvent` and proposed intervalled/periodic event handler execution:**
*Periodic Events with `CustomPeriodicEvent` (Example from docs):*
```python
trainer = ignite.engine.Engine(process_fun)

cpe1 = CustomPeriodicEvent(n_iterations=1000)
cpe1.attach(trainer)

# Let's define an event every 10 epochs
cpe2 = CustomPeriodicEvent(n_epochs=10)
cpe2.attach(trainer)

@trainer.on(cpe1.Events.ITERATIONS_1000_COMPLETED)
def on_every_1000_iterations(engine):
    # run a computation after 1000 iterations
    # ...
    print(engine.state.iterations_1000)

@trainer.on(cpe2.Events.EPOCHS_10_STARTED)
def on_every_10_epochs(engine):
    # run a computation every 10 epochs
    # ...
    print(engine.state.epochs_10)
```

*Intervalled/Periodic event handler execution with proposed feature using decorators:*
```python
trainer = ignite.engine.Engine(process_fun)

@trainer.on(Events.ITERATION_COMPLETED, interval=1000)
def on_every_1000_iterations(engine):
    # run a computation after 1000 iterations
    # ...
    print(engine.state.iteration)

@trainer.on(Events.EPOCH_STARTED, interval=10)
def on_every_10_epochs(engine):
    # run a computation every 10 epochs
    # ...
    print(engine.state.epoch)
```
or using `engine.add_event_handler` instead of decorators:

```python
trainer = ignite.engine.Engine(process_fun)

def on_every_1000_iterations(engine):
    # run a computation after 1000 iterations
    # ...
    print(engine.state.iteration)

trainer.add_event_handler(Events.ITERATION_COMPLETED, on_every_1000_iterations, interval=1000)

def on_every_10_epochs(engine):
    # run a computation every 10 epochs
    # ...
    print(engine.state.epoch)

trainer.add_event_handler(Events.EPOCH_STARTED, on_every_10_epochs, interval=10)
```
The keyword `interval` equals the argument `n_iterations` from `CustomPeriodicEvent` in case of iteration events and equals `n_epochs` in case of epoch events. Perhaps `perod`, `period_length` or `exec_interval` is a better argument name than `interval`.

@vfdev-5  What do you think about this? I created a PR for better discussion and I wanted to see, if this feature can be implemented with little changes to `Engine`. 
I also added a first test, to see if it works in general. But the PR isn't complete. 

~~At the moment, I have problems to run all tests locally without errors to investigate side-effects but `tests/ignite/engine/test_engine.py` passes.~~ Tests are all passing.

Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -127,6 +127,7 @@ class Engine(object):
         # Loss value is now stored in `engine.state.output`.
 
     """
+
     def __init__(self, process_function):
         self._event_handlers = defaultdict(list)
         self._logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
@@ -208,6 +209,8 @@ class Engine(object):
                 or any `event_name` added by :meth:`~ignite.engine.Engine.register_events`.
             handler (callable): the callable event handler that should be invoked
             *args: optional args to be passed to `handler`.
+            interval (int, optional): Execute handler in intervals (default: 1). For custom events, requires
+                event_to_attr to be set with :meth:`~ignite.engine.Engine.register_events`.
             **kwargs: optional keyword args to be passed to `handler`.
 
         Note:
@@ -232,6 +235,8 @@ class Engine(object):
             engine.add_event_handler(Events.EPOCH_COMPLETED, print_epoch)
 
         """
+        interval = kwargs.pop("interval") if "interval" in kwargs else 1
+
         if event_name not in self._allowed_events:
             self._logger.error("attempt to add event handler to an invalid event %s.", event_name)
             raise ValueError("Event {} is not a valid event for this Engine.".format(event_name))
@@ -239,7 +244,7 @@ class Engine(object):
         event_args = (Exception(), ) if event_name == Events.EXCEPTION_RAISED else ()
         self._check_signature(handler, 'handler', *(event_args + args), **kwargs)
 
-        self._event_handlers[event_name].append((handler, args, kwargs))
+        self._event_handlers[event_name].append((handler, args, kwargs, interval))
         self._logger.debug("added handler for event %s.", event_name)
 
         return RemovableEventHandle(event_name, handler, self)
@@ -259,7 +264,7 @@ class Engine(object):
         else:
             events = self._event_handlers
         for e in events:
-            for h, _, _ in self._event_handlers[e]:
+            for h, _, _, _ in self._event_handlers[e]:
                 if h == handler:
                     return True
         return False
@@ -275,8 +280,8 @@ class Engine(object):
         if event_name not in self._event_handlers:
             raise ValueError("Input event name '{}' does not exist".format(event_name))
 
-        new_event_handlers = [(h, args, kwargs) for h, args, kwargs in self._event_handlers[event_name]
-                              if h != handler]
+        new_event_handlers = [(h, args, kwargs, interval) for h, args, kwargs, interval in
+                              self._event_handlers[event_name] if h != handler]
         if len(new_event_handlers) == len(self._event_handlers[event_name]):
             raise ValueError("Input handler '{}' is not found among registered event handlers".format(handler))
         self._event_handlers[event_name] = new_event_handlers
@@ -317,9 +322,10 @@ class Engine(object):
             **kwargs: optional keyword args to be passed to `handler`.
 
         """
-        def decorator(f):
-            self.add_event_handler(event_name, f, *args, **kwargs)
+        def decorator(f, interval=1):
+            self.add_event_handler(event_name, f, *args, interval=interval, **kwargs)
             return f
+
         return decorator
 
     def _fire_event(self, event_name, *event_args, **event_kwargs):
@@ -340,9 +346,13 @@ class Engine(object):
         """
         if event_name in self._allowed_events:
             self._logger.debug("firing handlers for event %s ", event_name)
-            for func, args, kwargs in self._event_handlers[event_name]:
-                kwargs.update(event_kwargs)
-                func(self, *(event_args + args), **kwargs)
+            event_attrib_value = self.state.get_event_attrib_value(event_name) \
+                if event_name in self.state.event_to_attr \
+                else 1
+            for func, args, kwargs, interval in self._event_handlers[event_name]:
+                if (event_attrib_value - 1) % interval == 0:
+                    kwargs.update(event_kwargs)
+                    func(self, *(event_args + args), **kwargs)
 
     def fire_event(self, event_name):
         """Execute all the handlers associated with given event.
--- a/tests/ignite/engine/test_engine.py
+++ b/tests/ignite/engine/test_engine.py
@@ -579,7 +579,7 @@ def test_terminate_epoch_stops_mid_epoch():
     state = engine.run(data=[None] * num_iterations_per_epoch, max_epochs=max_epochs)
     # completes the iteration but doesn't increment counter (this happens just before a new iteration starts)
     assert state.iteration == num_iterations_per_epoch * (max_epochs - 1) + \
-           iteration_to_stop % num_iterations_per_epoch
+        iteration_to_stop % num_iterations_per_epoch
 
 
 def _create_mock_data_loader(epochs, batches_per_epoch):
--- a/tests/ignite/engine/test_engine.py
+++ b/tests/ignite/engine/test_engine.py
@@ -106,18 +106,47 @@ def test_add_event_handler():
 
     def handle_iteration_started(engine, counter):
         counter.count += 1
+
     engine.add_event_handler(Events.STARTED, handle_iteration_started, started_counter)
 
     completed_counter = Counter()
 
     def handle_iteration_completed(engine, counter):
         counter.count += 1
+
     engine.add_event_handler(Events.COMPLETED, handle_iteration_completed, completed_counter)
 
     engine.run(15)
 
     assert started_counter.count == 15
     assert completed_counter.count == 15
+
+
+def test_add_event_handler_with_interval():
+    engine = Engine(process_func)
+
+    class Counter(object):
+        def __init__(self, count=0):
+            self.count = count
+
+    started_counter = Counter()
+
+    def handle_iteration_started(engine, counter):
+        counter.count += 1
+
+    engine.add_event_handler(Events.EPOCH_STARTED, handle_iteration_started, started_counter, interval=3)
+
+    completed_counter = Counter()
+
+    def handle_iteration_completed(engine, counter):
+        counter.count += 1
+
+    engine.add_event_handler(Events.EPOCH_COMPLETED, handle_iteration_completed, completed_counter, interval=3)
+
+    engine.run(iter(range(15)), max_epochs=15)
+
+    assert started_counter.count == 5
+    assert completed_counter.count == 5
 
 
 def test_adding_multiple_event_handlers():
@@ -132,7 +161,6 @@ def test_adding_multiple_event_handlers():
 
 
 def test_event_removable_handle():
-
     # Removable handle removes event from engine.
     engine = DummyEngine()
     handler = MagicMock()
@@ -551,7 +579,7 @@ def test_terminate_epoch_stops_mid_epoch():
     state = engine.run(data=[None] * num_iterations_per_epoch, max_epochs=max_epochs)
     # completes the iteration but doesn't increment counter (this happens just before a new iteration starts)
     assert state.iteration == num_iterations_per_epoch * (max_epochs - 1) + \
-        iteration_to_stop % num_iterations_per_epoch
+           iteration_to_stop % num_iterations_per_epoch
 
 
 def _create_mock_data_loader(epochs, batches_per_epoch):
@@ -796,7 +824,6 @@ def test_create_supervised_with_metrics():
 
 
 def test_reset_should_terminate():
-
     def update_fn(engine, batch):
         pass
 


vfdev-5:
@Bibonaut thanks for the proposal! 
If I understand correctly, you would like to setup custom events directly without using CustomPeriodicEvents ?
Could you please compare in details your feature's API vs CustomPeriodicEvents.

Bibonaut:
@vfdev-5  Sorry for the short description of the feature. I thought the implemented test ([link](https://github.com/pytorch/ignite/pull/624/commits/dd8a659dc0679733f5ac87b689398b2469d07f09)) gives a good first understanding of the proposed feature. 
I have updated the PR description with a comparison of `CustomPeriodicEvent` and my proposal. Does it help? 
And I renamed the PR to `Periodic event handler execution` to make the similarity to `CustomPeriodicEvent` more clear. 
But instead of defining custom events that are fired every n-th epoch or n-th iteration with `CustomPeriodicEvent`, I want to add this functionality directly to `Engine`. That allows to define an *execution interval* or *period* for each handler that is added with `add_event_handler` or decorated with `@trainer.on(...)` without adding new custom events.

vfdev-5:
@Bibonaut thanks for updates ! Yes, I agree that `CustomPeriodicEvent` is a bit heavy to use:
- we need to setup the import 
- we need to setup its type

In the feature you propose it is more simple to setup the same behaviour. However, there could be a misunderstanding with `interval=1000` and **kwargs passed to the handler in the following:
```python
trainer.add_event_handler(Events.ITERATION_COMPLETED, on_every_1000_iterations, interval=1000)
```

I am not a fan of this solution. IMO we should keep the behaviour after handler function as previously: *args, **kwargs. The only place where we could provide the interval, IMO, is `Events.ITERATION_COMPLETED`. Maybe we could introduce something like
```
Events.Every(1000).ITERATION_COMPLETED
# or something similar and easy to setup.
```
and this could be automatically transformed to what you proposes.
What do you think ?

Bibonaut:
@vfdev-5 I agree that with the `interval` keyword, **kwargs passed to the handler are mixed up with keywords of `add_event_handler` which is not a nice and clean solution. 

I really like your idea using something like `Events.Every(1000).ITERATION_COMPLETED`. I had a short look at the features of the `Enum` class but I don't know, how to implement something like this. 

The most important question I have is, what should be returned if something like  `Events.Every(1000).ITERATION_COMPLETED` is called? 
If you have a hint or some references, I can work on this. 


[...] I played around a bit with the `Enum` class and the snippet below is what I have until now. I thought to overload the `__call__` method and add `every` as an attribute to the Events class. But this is a problem because `every` is added to the class and not to an individual `Events` object:
```python
class Events(Enum):
    """Events that are fired by the :class:`~ignite.engine.Engine` during execution."""
    EPOCH_STARTED = "epoch_started"
    EPOCH_COMPLETED = "epoch_completed"
    STARTED = "started"
    COMPLETED = "completed"
    ITERATION_STARTED = "iteration_started"
    ITERATION_COMPLETED = "iteration_completed"
    EXCEPTION_RAISED = "exception_raised"

    
    def __call__(cls, *args, every=1):
        setattr(cls, "every", every)
        return cls

event = Events.ITERATION_STARTED(every=10)

# We can then access the attribute of the event to get information about the interval:
print(event.every)  # --> outputs 10 

# The attribute `every` is a class attribute and is persistent 
event2 = Events.ITERATION_STARTED
print(event2 .every) # --> outputs 10 but sould output 1
```

Another way would be to output a tuple (Event, every) if the event is called but this changes the whole behaviour of `Events` class:

```python
class Events(Enum):
    """Events that are fired by the :class:`~ignite.engine.Engine` during execution."""
    EPOCH_STARTED = "epoch_started"
    EPOCH_COMPLETED = "epoch_completed"
    STARTED = "started"
    COMPLETED = "completed"
    ITERATION_STARTED = "iteration_started"
    ITERATION_COMPLETED = "iteration_completed"
    EXCEPTION_RAISED = "exception_raised"

    
    def __call__(cls, *args, every=1):
        return cls, every # returns a tuple

event = Events.ITERATION_STARTED(every=10)
print(type(event)) # tuple
print(event) # (<Events.ITERATION_STARTED: 'iteration_started'>, 10)

```

vfdev-5:
@Bibonaut well, for instance, I do not know how to implement this `Events.Every(1000).ITERATION_COMPLETED` :) 
Maybe, we can replace `Events(Enum)` by something else and where we could implement the previous idea (keeping previous options).

Most probably, `Events.Every(1000)` should be a static method to generate new enum class similarly to how it is done here : 
https://github.com/pytorch/ignite/blob/224076dabb7370d78be4b08294bcaec4f8880bc2/ignite/contrib/handlers/custom_events.py#L76-L78

vfdev-5:
Here is an implementation to discuss:
```python
from enum import Enum

_PERIODIC_EVENTS = {}

class Events(Enum):
    """Events that are fired by the :class:`~ignite.engine.Engine` during execution."""
    EPOCH_STARTED = (1, "epoch_started")
    EPOCH_COMPLETED = (1, "epoch_completed")
    STARTED = (1, "started")
    COMPLETED = (1, "completed")
    ITERATION_STARTED = (1, "iteration_started")
    ITERATION_COMPLETED = (1, "iteration_completed")
    EXCEPTION_RAISED = (1, "exception_raised")
    
    @classmethod
    def Every(cls, interval):        
        if interval in _PERIODIC_EVENTS:
            return _PERIODIC_EVENTS[interval]
        
        events = [
            ('EPOCH_STARTED', (interval, "epoch_started")),
            ('EPOCH_COMPLETED', (interval, "epoch_completed")),
            ('ITERATION_STARTED', (interval, "iteration_started")),
            ('ITERATION_COMPLETED', (interval, "iteration_completed")),
        ]
        _PERIODIC_EVENTS[interval] = Enum("PeriodicEvents", events)
        return _PERIODIC_EVENTS[interval]
```
with 
```
isinstance(Events.Every(10).EPOCH_STARTED, Enum), Events.Every(10).EPOCH_STARTED.value[0]
> (True, 10)
Events.Every(5).EPOCH_STARTED.value[0]
> 5
_PERIODIC_EVENTS
> {10: <enum 'PeriodicEvents'>, 5: <enum 'PeriodicEvents'>}
Events.EPOCH_STARTED.value
> (1, 'epoch_started')
```

Bibonaut:
@vfdev-5 I like your solution to introduce `PeriodicEvents` as a new `Events` class. It makes the difference between normal events and periodic events more clear.  I haven't thought about using a module-scoped variable like `_PERIODIC_EVENTS`. 
The only drawback I see is that introducing `PeriodicEvents`  requires some code changes, e.g. when we want to make use of  `State.event_to_attr[event_name]`.  Or do you already have a nice solution for this? 

In the meantime I also tried some things out and I am currently investigating another approach, where I create event objects from a new class called `Event`. `Event` objects should behave externally like `Events` enumeration to keep the required interface changes small.  

The following code contains two API versions:
* First version overloads `Events` `__call__` method so that the period can be defined with a keyword, eg. `Events.ITERATION_COMPLETED(every=10)`. 
* 2nd version adds method `Every` as you proposed. But here it is a static method and event can be created with `Events.ITERATION_COMPLETED.Every(10)`

```python

# Version 1: add __call__ to Events from which new Event objects are created. 
def _call(cls, every=1):
    return Event(cls, every=every)
setattr(Events, "__call__", _call)

# Version 2: add 'Every' method to Events from which new Event objects are created. 
def Every(cls, value):
    return Event(cls, every=value)
setattr(Events, "Every", Every) 

# Example of a new Event class, stores event_name and can store other properties
class Event:
    def __init__(self, event_name, every=1):
        self.event_name = event_name
        self.every = every

    def __hash__(self):
        return hash(self.event_name)

    def __eq__(self, other):
        if isinstance(other, Events):
            return self.event_name == other
        else:
            return super(Event, self).__eq__(other)
```

Example usage and behaviour of new `Event` objects:
```python
    e1 = Events.EPOCH_STARTED(every=10)  # API version 1
    e2 = Events.EPOCH_STARTED(every=11)  # API version 1
    e3 = Events.ITERATION_STARTED.Every(10)  # API version 2

    print(e1 == e2)  # -> False, object comparison works
    print(e1 == Events.EPOCH_STARTED)  # -> True, can compare `Events with `Event` object
   
    # test event_to_attr:
    from ignite.engine import State
    print(State.event_to_attr[e1])  # -> epoch
    print(State.event_to_attr[e3])  # -> iteration
```

What do you think? 
IMO, the main question at this point is, if it is likely that `Events` functionality will increase in the future and if we then need to store information other than one integer created from `Every`. 
E.g. another nice feature would be something like `Events.Once(15).EPOCH_COMPLETED` that calls the handler once at EPOCH_COMPLETED at epoch 15.

vfdev-5:
👍 for `Events.Once(15)` I needed few times such behaviour and I think this could be helpful.


> The only drawback I see is that introducing PeriodicEvents requires some code changes, e.g. when we want to make use of State.event_to_attr[event_name]. Or do you already have a nice solution for this?

If your question is about how to register new events, can we do this as in CustomPeriodicEvents:
https://github.com/pytorch/ignite/blob/224076dabb7370d78be4b08294bcaec4f8880bc2/ignite/contrib/handlers/custom_events.py#L80-L82
?

Your proposition with `class Event` looks interesting ! I have a question about the API, for example `Events.EPOCH_STARTED(every=10)` is fired every 10-th epoch at start. But there is no firing EPOCH_COMPLETED every 10th epochs... Maybe, this is OK too. If we compare with `Events.Every(10)` it creates events for epoch and iterations automatically....

Bibonaut:
> I have a question about the API, for example Events.EPOCH_STARTED(every=10) is fired every 10-th epoch at start. But there is no firing EPOCH_COMPLETED every 10th epochs... Maybe, this is OK too. If we compare with Events.Every(10) it creates events for epoch and iterations automatically....

Yes that's true. I do not intend to create  the whole set of iteration and epoch events with `Events.EPOCH_STARTED(every=10)`, because I only want to create this behaviour for a specific handler, when I call:
```python
trainer.add_event_handler(Events.EPOCH_STARTED(every=10), handler)
```
I thought, that you are creating the whole set of events, because you want to simply storing those events in a dict with `interval`/`period` as dict key.

-----------------------------

I want to come up with another more generic approach from which I think that it is more *ignite-style* than the discussed approaches.
 As we found out, there can more use cases than `Every`, e.g. `Once`. But for any good reason a fully customizable handler execution logic should be possible:
```
# Example 1
epoch no     | 1 2 3 4 5 6 7 8 ...
handler exec | x x     x x 

# Example 2
epoch no     | 1 2 3 4 5 6 7 8 ...
handler exec | x x x   x x x

# Or any other desired sequence
``` 

This shall be achieved by adding a customizable  *event gate* that determines, if the handler is executed when the event is fired. That means, with this approach I do not want do add any additional events. Instead I just want to suppress the handler execution if the *event gate* tells me to do so. 

To achieve this, we have to provide a function, that accepts the event attribute (e.g. epoch number, iteration number, custom) and outputs a boolean variable: 
* True, if the handler shall be executed
* else False 

E.g. for a periodic event (`Every`), executed every 10th event and a one-time event gate (`Once`): 
```python
every = 10
def periodic_event_gate(event_attribute):
    return (event_attribute-1) % every == 0

once = 10
def one_time_event_gate(event_attribute):
    return event_attribute == once
```

Then we have to change `Engine.add_event_handler` to pass the event gate function with the event:
```python
trainer.add_event_handler((Events.EPOCH_COMPLETED,periodic_event_gate), handler)
```
I propose to change the interface of `add_event_handler` that it allows to pass 
* an event as first argument (actual API)
* or a tuple of `(event, event_gate_fun)` (new API). 

In `Engine.add_event_handler`, we can check the signature of the event gate function and add a default gate function, if only the event is passed to the function. 
Note, that event gate function is called `event_execution_handler` in code.
```python
    def add_event_handler(self, event_name, handler, *args, **kwargs):
        if isinstance(event_name, tuple):
            event_execution_handler = event_name[1]
            # TODO: add signature check of event_execution_handler 
            event_name = event_name[0]
        else:
            def event_execution_handler(event_attribute):
                # default event gate never suppresses any events
                return True
        # [...] some argument checks
        # add event_execution_handler to the handler tuple
        self._event_handlers[event_name].append((handler, args, kwargs, event_execution_handler))
```

In `Engine._fire_event` we can add the query if event handler shall be executed (see line with comment `# event gate`):
```python
def _fire_event(self, event_name, *event_args, **event_kwargs):                             
    """Execute all the handlers associated with given event.                                
    [...]                                                                                       
    """                                                                                     
    if event_name in self._allowed_events:                                                  
        self._logger.debug("firing handlers for event %s ", event_name)                     
        event_attrib_value = self.state.get_event_attrib_value(event_name) \                
            if event_name in self.state.event_to_attr \                                     
            else 1                                                                          
        for func, args, kwargs, event_exec_handler in self._event_handlers[event_name]:     
            if event_exec_handler(event_attrib_value):  # event gate                                    
                kwargs.update(event_kwargs)                                                 
                func(self, *(event_args + args), **kwargs)                                  

```

I have created some example implementations for `Every` and `Once`. And I think it is very easy and straight forward, to implement any other desired logic with this scheme. 

```python
from ignite._six import with_metaclass


class BaseEventExecution(with_metaclass(ABCMeta, object)):
    """A handler that controls event execution for event handlers based on the event attribute"""
    
    @abstractmethod
    def __call__(self, event_attribute):
        """ Implements the logic if a event handler should be called on event

        Args:
            event_attribute (int): The event attribute obtained from :attr:`~ignite.engine.State.event_to_attr`
        Returns:
            Bool value, True if event handler shall be called else False
        """
        raise NotImplementedError


class Once(BaseEventExecution):
    """ Event execution handler that executes handler once

    Args:
        attr_value (int): Attribute value. Handler is executed when attr_value matches the event attribute.
    """

    def __init__(self, attr_value):
        super(Once, self).__init__()
        self.attr_value = attr_value

    def __call__(self, event_attribute):
        return event_attribute == self.attr_value


class Every(BaseEventExecution):
    """ Event execution handler that executes handler periodically every n-th event occurrence

    Args:
        n (int): Execute the event every n-th event occureence
        start_value (int, optional): Start value of the event attribute.
            E.g. 1 for iteration and epoch events. (default=1)
    """

    def __init__(self, n, start_value=1):
        super(Every, self).__init__()
        self.n = n
        self.start_value = start_value

    def __call__(self, event_attribute):
        return (event_attribute - self.start_value) % self.n == 0

```

@vfdev-5 What do you think?

vfdev-5:
@Bibonaut 👍 totally agree with the proposition. Maybe we can polish a bit the API, but provide such customizable handling would be awesome !
```
```

Bibonaut:
@vfdev-5 Thanks! I will update the PR. Then we can work on the API 👍

vfdev-5:
@Bibonaut I think your previous proposition makes much sense in this context:
```
Events.EPOCH_STARTED(event_transform=periodic_event_gate)
```

Bibonaut:
@vfdev-5 I am still uncertain, how the API of `Engine.add_event_handler` should look like. 
1. Do we want to pass a new type of event object that is created from `__call__` in Events class as previously propsed:
```python

# Version 1: add __call__ to Events from which new Event objects are created. 
def _call(cls, event_transform=None):
    return Event(cls, event_transform=event_transform)
setattr(Events, "__call__", _call)

# Example of a new Event class, stores event_name and can store other properties
class Event:
    def __init__(self, event_name, event_transform=None):
        self.event_name = event_name
        self.event_transform= event_transform if event_transform else lambda x: True

    def __hash__(self):
        return hash(self.event_name)

    def __eq__(self, other):
        if isinstance(other, Events):
            return self.event_name == other
        else:
            return super(Event, self).__eq__(other)
```

2. Or do we want to pass a tuple or something similar that is created from Events `__call__` method?
```python
class Events(Enum):
    """Events that are fired by the :class:`~ignite.engine.Engine` during execution."""
    EPOCH_STARTED = "epoch_started"
    EPOCH_COMPLETED = "epoch_completed"
    STARTED = "started"
    COMPLETED = "completed"
    ITERATION_STARTED = "iteration_started"
    ITERATION_COMPLETED = "iteration_completed"
    EXCEPTION_RAISED = "exception_raised"

    def __call__(self, event_transform=None):
        # add signature check here?
        if event_transform is not None:
            return self, event_transform
        else:
            return self
```

The problem here is, if we are using pure events, e.g. `Events.EPOCH_COMPLETED`, I can see no function call to any method of the `Events` class  (except `getattr`?).  In this case, always the pure event is returned instead of a new `Event` object (version 1) or  a tuple (version 2). 

I still have problems to understand and investigate the capabilities of the `Enum` class. Overloading methods mostly fail.

vfdev-5:
@Bibonaut I'll play with that as I have more time.

At first time I thought that keeping enum would be preferable, but as `Events.EPOCH_STARTED(events_filter=lambda e: True if e == 10 else False)` should not override all `Events.EPOCH_STARTED` but be an another event, either we can create a storage as `_CUSTOM_EVENTS` or findout a way to dynamically append to `Events`. Another solution, and maybe more simple, is to set all events as class instances... But in this case it would be preferable to keep previous simplicity for users who create custom events...

PS. I'm not sure now `event_transform` is an appropriate name...

Bibonaut:
> At first time I thought that keeping enum would be preferable, but as `Events.EPOCH_STARTED(events_filter=lambda e: True if e == 10 else False)` should not override all `Events.EPOCH_STARTED` but be an another event, either we can create a storage as `_CUSTOM_EVENTS` or findout a way to dynamically append to `Events`. 

@vfdev-5 From your comment I read out, that you prefer to dynamically add the `events_filter` as an attribute to the specific event which then would result in a unique event object? That would be something like version 2 from my previous comment but instead of returning a tuple, we are adding `events_filter` as an attribute:
```python
class Events(Enum):
    """Events that are fired by the :class:`~ignite.engine.Engine` during execution."""
    EPOCH_STARTED = "epoch_started"
    EPOCH_COMPLETED = "epoch_completed"
    STARTED = "started"
    COMPLETED = "completed"
    ITERATION_STARTED = "iteration_started"
    ITERATION_COMPLETED = "iteration_completed"
    EXCEPTION_RAISED = "exception_raised"

    def __call__(self, event_filterNone):
        # add signature check here?
        if event_filter is not None:
            setattr(self, "event_transform", event_transform)
        return self
```
Off course, using `Enum` base class, this would add `event_transform` to all events:
```python
a = Events.EPOCH_STARTED(event_filter=lambda x: True)
b = Events.EPOCH_STARTED
print(b.event_filter)  # --> lambda x: True
```
What we want is to add `event_filter` only to `a` but not to `b`, right? 
If we create classes for each event, we have to implement a comparison mechanisms so that `a==b` returns true even if `a` and `b` have different event filters? 

A the moment, I am still in favour of the *event gate function* version, since the implementation and mechanisms are very simple. 

But instead of controlling event exevution, we could also provide wrappers for the handlers instead of defining an event gate (not tested):
```python
class Once:
    def __init__(event_name, handler,attr_value):
        self.event_name = event_name
        self.handler = handler
        self.attr_value = attr_value

    def __call__(self,engine,*args,**kwargs):
       event_attrib_value = self.state.get_event_attrib_value(self.event_name) \
                if self.event_name in self.state.event_to_attr \
                else 1
            if event_attrib_value == self.attr_value:
                self.handler(engine,*args,**kwargs)
```
But there is a lot of redundant code here (if we create multiple wrappers) and we have to provide the `event_name` to get `event_attrib_value` which is not optimal in comparison to the *event gate function* approach where the mechanism to get `event_attrib_value` is implemented once in `Engine` class. 

We can reduce redundant code, if we use a base class (not tested):
```python
class BaseHandlerWrapper:
    def __init__(event_name, handler):
        self.event_name = event_name
        self.handler = handler
    
    def __call__(self, engine,*args,**kwargs):
        event_attrib_value = self.state.get_event_attrib_value(self.event_name) \
                if self.event_name in self.state.event_to_attr \
                else 1
        if self.exec_on_event(event_attrib_value):
            self.handler(engine, *args, **kwargs)

    def exec_on_event(event_attrib_value):
        """Implements the handler execution logic"""
        raise NotImplementedError

class Once(BaseHandlerWrapper):
    def __init__(event_name, handler,attr_value):
        super(BaseHandlerWrapper,self).__init__(event_name, handler)
        self.attr_value = attr_value

    def exec_on_event(event_attrib_value):
        return event_attrib_value == self.attrib_value
```
But I am not convinced of this solution. I just wanted to mention it.

P.S. I aggree that `event_transform` is not an appropriate name. 
I have no time left this week to work on it. But I can test some implementations next week if we have a suitable concept :)

vfdev-5:
@Bibonaut sorry for delay. Actually, you are right about `event gate function` if I understood correctly the idea. 

I'm just trying to simply recap the feature. As user would like to setup handlers with a flexbility: 
```python

def event_filter(engine):
     # some logic
     return True # of False

@trainer.on(Events.EPOCH_STARTED, *args, **kwargs)
def user_handler(engine, *args, **kwargs):
    if event_filter(engine):
        # run handler
```

We would like to hide inside `Engine` this `if event_filter(engine)` and provide predefined event filters as `once_filter`, `inside_filter` etc.

The only place where we can setup `event_filter` is the event:  `@trainer.on("Events.EPOCH_STARTED with event_filter", *args, **kwargs)`. I agree with you that we do not need to create new events etc. We just need to provide some "synthetic sugar" to pass event_filter into `Engine.add_event_handler` and wrap the handler.

So, we can play around this concept and decide on a solid API :)

vfdev-5:
@Bibonaut I close this PR in favor of #656 
Please, take a look at this PR. Thanks

PR #644, opened 2019-10-22
Enhancement of class `Metric` by adding attachment events to `__init__()` arguments

### LINKED ISSUE #643: [ignite.metrics.Metric](ignite/metrics/metric.py) for all use-cases: Expand Metric's arguments by attachment events
### Issue
[`ignite.metrics.Metric`](ignite/metrics/metric.py) currently only supports the calculation (and therefore also the visualization/tensorboard logging) of metrics at each completed epoch. In regular use-cases, e.g. to determine/visualize the optimum iteration for early stopping, it is required to return the loss metric after each iteration or at a customized event. This is currently always related to overriding `Metric.attach()`. So currently I experienced the number of use-cases to be very limited without overriding the source code all the time.

### Solution
Add all events used to attach the methods of Metric to the arguments with default values referring to the current attachment events. The default values of the added event arguments will guarantee the backward compatibility of existing code and keep boilerplate code at the same level for "epoch-metrics". For any other desired/required use-case, the event arguments will provide updating, calculating and outputting the metric to `engine.metrics` at any desired/required event.
This would of course require modifying all inheriting classes of `Metric` (e.g. `Loss`, `Accuracy`, `LambdaMetric` and their base classes)  but the modifications are simple and straight forward.

### Code suggestion
Here a possible implementation:
```python
from ignite._six import with_metaclass
import torch


class Metric(with_metaclass(ABCMeta, object)):
    """
    Base class for all Metrics.

    Args:
        output_transform (callable, optional): a callable that is used to transform the
            :class:`~ignite.engine.Engine`'s `process_function`'s output into the
        started_event (<enum Events>): event from which on the metric should be calculated
        iteration_completed_event (<enum Events>): event at which intermediate metric calculations (`self.update()`)
            are executed from current model outputs, e.g. for a mean loss metric this would refer to adding
            the current model loss output after each iteration to a summed loss and increasing to count of losses added.
        completed_event (<enum Events>): event at which the metric value us calculated and written to
            `engine.state.metrics[metric_name]`. E.g. for mean loss metric calculation the summed output losses
            of each iterations is devided by the number of iterations and outputed to `trainer.state.metrics['loss'].
    """

    def __init__(self, output_transform=lambda x: x, started_event=Events.EPOCH_STARTED,
                 iteration_completed_event=Events.ITERATION_COMPLETED, completed_event=Events.EPOCH_COMPLETED):
        self._output_transform = output_transform
        self.started_event = started_event
        self.iteration_completed_event = iteration_completed_event
        self.completed_event = completed_event
        self.reset()

    @abstractmethod
    def reset(self):
        """
        Resets the metric to it's initial state.

        This is called at the start of each epoch.
        """
        pass

    @abstractmethod
    def update(self, output):
        """
        Updates the metric's state using the passed batch output.

        This is called once for each batch.

        Args:
            output: the is the output from the engine's process function.
        """
        pass

    @abstractmethod
    def compute(self):
        """
        Computes the metric based on it's accumulated state.

        This is called at the end of each epoch.

        Returns:
            Any: the actual quantity of interest.

        Raises:
            NotComputableError: raised when the metric cannot be computed.
        """
        pass

    def started(self, engine):
        self.reset()

    @torch.no_grad()
    def iteration_completed(self, engine):
        output = self._output_transform(engine.state.output)
        self.update(output)

    def completed(self, engine, name):
        result = self.compute()
        if torch.is_tensor(result) and len(result.shape) == 0:
            result = result.item()
        engine.state.metrics[name] = result

        if not engine.has_event_handler(self.started, self.started_event):
            engine.add_event_handler(self.started_event, self.started)
        if not engine.has_event_handler(self.iteration_completed, self.iteration_completed_event):
            engine.add_event_handler(self.iteration_completed_event, self.iteration_completed)
        # `self.completed()` is always executed at the end, so it should be added to `engine` at the end
        engine.add_event_handler(self.completed_event, self.completed, name)

    def __add__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x + y, self, other)

    def __radd__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x + y, other, self)

    def __sub__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x - y, self, other)

    def __rsub__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x - y, other, self)

    def __mul__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x * y, self, other)

    def __rmul__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x * y, other, self)

    def __pow__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x ** y, self, other)

    def __rpow__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x ** y, other, self)

    def __mod__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x % y, self, other)

    def __div__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x.__div__(y), self, other)

    def __rdiv__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x.__div__(y), other, self)

    def __truediv__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x.__truediv__(y), self, other)

    def __rtruediv__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x.__truediv__(y), other, self)

    def __floordiv__(self, other):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x, y: x // y, self, other)

    def __getattr__(self, attr):
        from ignite.metrics import MetricsLambda

        def fn(x, *args, **kwargs):
            return getattr(x, attr)(*args, **kwargs)

        def wrapper(*args, **kwargs):
            return MetricsLambda(fn, self, *args, **kwargs)
        return wrapper

    def __getitem__(self, index):
        from ignite.metrics import MetricsLambda
        return MetricsLambda(lambda x: x[index], self)
```

DrStoop:
The original `Metric.__init__(self, output_transform=lambda x: x)` is enhanced to
`Metric.__init__(output_transform=lambda x: x,  started_event=Events.EPOCH_STARTED,
iteration_completed_event=Events.ITERATION_COMPLETED, completed_event=Events.EPOCH_COMPLETED)
to provide metric updating and computing at any individual event. The default values
guarantee backward compatibility.

Modified metrics inheriting from `Metric`s will follow up...

Fixes #643 

Description:

Check list:

* [x] `pytest`s were passed with constant number of failed tests, before/after = 14 of 341 / 14 of 341
* [ ] Test backward compatibility on few older examples (I only have 'fresh' code running)


vfdev-5 (commented):
Looks good ! Thanks @DrStoop 
Let me fix CI and I'll play a bit with new interface and if OK I'll merge it

DrStoop:
Great, I'm happy to see that merged :+1: 

... by the way despite some subclasses of `Metric`, I will also provide some more enhancements the next day, so we can check if they are appropriate, too.I'll try to issue them in a easy-to-merge order.

vfdev-5:
@DrStoop sorry, yes, you are right about other classes. Maybe we would need to get them all in this PR.

DrStoop:
no problem, wait with the merge. I will open separate issues for each topic I was working on. After deciding on the features to integrate, I'll tie the winners up and pull-request them at once...

vfdev-5:
@DrStoop I started to write some tests for this new feature and seems like with current events system it is rather difficult to setup something useful, e.g.
```python
# Would like to compute accuracy on the half of the loader
# the following should compute only on the first half
e1 = Events.ITERATION_STARTED(once=1),
e2 = Events.ITERATION_COMPLETED,
e3 = Events.ITERATION_COMPLETED(once=50)
m = Accuracy(..., started_event=e1, iteration_completed_event=e2, completed_event=e3)
```
The problem is with reset event. I added another issue #710 such that if we could pass a composite event we could `reset` at the begining and after each `compute`...

DrStoop:
Hi @vfdev-5,

yeah that's maybe because it was implemented before the event filter... but i'll come up with something possibly in a week or two... unfortunately won't make it earlier

PR #673, opened 2019-11-25
BC breaking change - Refactor ModelCheckpoint into Checkpoint + DiskSaver / ModelCheckpoint

### LINKED ISSUE #387: Custom configuration how/where to save the best model with ModelCheckpoint
In case when user would like to integrate `ModelCheckpoint` with a package for experiment tracking, e.g. mlflow, polyaxon, etc. 
In such case logging, model weights etc can be stored on a cloud storage, e.g.
```python
exp_tracking.log_artifact(filepath)
```
Be default `ModelCheckpoint` is saving the model to the provided path `dirname`.
Idea is to provide a flexibility to execute a custom code when model is saved and be eable to store everywhere we would like.

What do you think ?

cc @elanmart

### LINKED ISSUE #550: Improve ModelCheckpoint naming
The checkpointer currently appends the epoch/iteration to the filename. However this number is reset when resuming a training from a checkpoint.
This is really annoying when say for example the last checkpoint is named ```checkpoint_60.pth```, then training is stopped and later resumed. The next checkpoint will then be named ```checkpoint_1.pth```.

A solution might be to add a ```global_step_transform``` argument so that epoch/iteration from the saved checkpoint can be reused.

vfdev-5:
Fixes #550 
Fixes #387

Description:
Introduced classes Checkpoint and DiskSaver as basic bricks and rewritten ModelCheckpoint using them

Check list:

cc @TheCodez, @Bibonaut 

BC breaking change:

- Argument `save_as_state_dict` is deprecated and should not be used. It is considered as True.
        
- Argument `save_interval` is deprecated and should not be used. Please, use events filtering instead, e.g. :attr:`~ignite.engine.Events.ITERATION_STARTED(every=1000)`

- There is no more internal counter that has been used to indicate the number of save actions. User could see its value `step_number` in the filename, e.g. `{filename_prefix}_{name}_{step_number}.pth`. Actually, `step_number` is replaced by current engine's epoch if `score_function` is specified and current iteration otherwise.

- A single `pth` file is created instead of multiple files.

### DIFF
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -3,6 +3,10 @@ ignite.handlers
 
 .. currentmodule:: ignite.handlers
 
+.. autoclass:: Checkpoint
+
+.. autoclass:: DiskSaver
+
 .. autoclass:: ModelCheckpoint
 
 .. autoclass:: EarlyStopping
--- a/ignite/contrib/handlers/base_logger.py
+++ b/ignite/contrib/handlers/base_logger.py
@@ -105,7 +105,7 @@ class BaseOutputHandler(BaseHandler):
             if not isinstance(another_engine, Engine):
                 raise TypeError("Argument another_engine should be of type Engine, "
                                 "but given {}".format(type(another_engine)))
-            warnings.warn("Use of another_engine is deprecated and will be removed in 0.3.0. "
+            warnings.warn("Use of another_engine is deprecated and will be removed in 0.4.0. "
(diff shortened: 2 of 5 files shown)


eduardo4jesus:
What is the meaning of **BC** in `BC-breaking`?

alxlampe:
I think it's *backward compatibility*

vfdev-5:
Yes, BC - breaking change means backward compatibility breaking change.

PR #679, opened 2019-11-30
Added archived argument and fixed behaviour with requires_empty

vfdev-5:
Description: Added archived argument and fixed behaviour with requires_empty

Check list:

### DIFF
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -1,6 +1,17 @@
 ignite.handlers
 ===============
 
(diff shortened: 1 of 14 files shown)


vfdev-5:
Need to improve filename suffix:
- add `global_step_transform`, this will partially fix the problem with multiple files as `best_model_1_val_acc=1234.pth`, where 1 is evaluator's epoch.
- if a single object is stored, save directly it's state_dict instead of `{'object': state_dict}`

PR #685, opened 2019-12-11
Early stopping min delta

kagrze:
This PR introduces a new early stopping parameter that enables definition of minimum increase in the score to qualify as an improvement.

Check list:

### DIFF
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -10,7 +10,7 @@ class EarlyStopping(object):
         patience (int):
             Number of events to wait if no improvement and then stop the training.
         min_delta (float):
-            Minimum increase in the monitored quantity to qualify as an improvement,
+            Minimum increase in the score to qualify as an improvement,
             i.e. an increase of less than min_delta, will count as no improvement.
         score_function (callable):
             It should be a function taking a single argument, an :class:`~ignite.engine.Engine` object,
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -62,7 +62,7 @@ class EarlyStopping(object):
 
         if self.best_score is None:
             self.best_score = score
-        elif score - self.best_score <= self.min_delta:
+        elif score <= self.best_score + self.min_delta:
             if score > self.best_score:
                 self.best_score = score
             self.counter += 1
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -10,7 +10,7 @@ class EarlyStopping(object):
         patience (int):
             Number of events to wait if no improvement and then stop the training.
         min_delta (float):
-            Minimum increase in the score to qualify as an improvement,
+            Minimum increase in the score after one event to qualify as an improvement,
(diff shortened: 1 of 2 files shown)


vfdev-5:
@kagrze thanks for the PR ! Looks good, just some comments to clarify and we can merge.

ignite/handlers/early_stopping.py
```diff
@@ -55,7 +62,9 @@ def __call__(self, engine):
 
         if self.best_score is None:
             self.best_score = score
-        elif score <= self.best_score:
+        elif score - self.best_score <= self.min_delta:
```
vfdev-5:
```suggestion
        elif score <= self.best_score + self.min_delta:
```
Optional

ignite/handlers/early_stopping.py
```diff
@@ -55,7 +62,9 @@ def __call__(self, engine):
...
+            if score > self.best_score:
```
vfdev-5: Should we store new score as best one in this case ?
For example, `min_delta=0.1`, 
- the last best score is `3.0` and current score is now `3.05`  => counter += 1 and best score is `3.05`
- the last best score is `3.05` and current score is now `3.1`  => counter += 1 and best score is `3.1`
- the last best score is `3.1` and current score is now `3.15`  => counter += 2 and best score is `3.15`
And we see that in the 3rd evaluation the model has improved to 3.15 from 3.0 with min_delta=0.1 but counter still incremented.
What do you think ?

ignite/handlers/early_stopping.py
```diff
@@ -31,19 +34,23 @@ def score_function(engine):
...
+    def __init__(self, patience, score_function, trainer, min_delta=0.):
 
         if not callable(score_function):
             raise TypeError("Argument score_function should be a function.")
 
         if patience < 1:
             raise ValueError("Argument patience should be positive integer.")
 
+        if min_delta < 0.:
+            raise ValueError("Argument min_delta should not be a negative number.")
```
vfdev-5: Coverage patch complains about this line being not covered

ignite/handlers/early_stopping.py
kagrze: Changed

ignite/handlers/early_stopping.py
kagrze: Test added. Thanks.

vfdev-5:
@kagrze thanks for the updates, how about the last comment on "Should we store new score as the best one" ?

kagrze:
> @kagrze thanks for the updates, how about the last comment on "Should we store new score as the best one" ?

Not sure about it. Let me think about it.

ignite/handlers/early_stopping.py
kagrze: It depends whether `min_delta` should be a minimum increase in one step or a cumulative minimum increase since last patience reset. I assumed the former but it could be the latter as well. I've clarified a description in doc string.

vfdev-5: Maybe we could an additional flag to make both options possible ?

kagrze: I've added `cumulative_delta` flag and two more unit tests.

PR #695, opened 2019-12-31
Let the user specify tqdm parameters for log_message. BC breaking change

maxfrei750:
Description: Let the user specify tqdm parameters for log_message (e.g. to have the log message be written to a file).

Check list:
* [ ] New tests are added (if a new feature is added) => not required
* [ ] New doc strings: description and/or example code are in RST format => probably not required
* [ ] Documentation is updated (if required) => probably not required

### DIFF
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -39,6 +39,16 @@ class ProgressBar(BaseLogger):
             # Progress bar will looks like
             # Epoch [2/50]: [64/128]  50%|█████      [06:17<12:34]
 
+        Log output to a file instead of stderr (tqdm's default output)
+
+        .. code-block:: python
+
+            trainer = create_supervised_trainer(model, optimizer, loss)
+
+            log_file = open("output.log", "w")
+            pbar = ProgressBar(file=log_file)
+            pbar.attach(trainer)
+
         Attach metrics that already have been computed at :attr:`~ignite.engine.Events.ITERATION_COMPLETED`
         (such as :class:`~ignite.metrics.RunningAverage`)
 
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -127,8 +127,7 @@ class ProgressBar(BaseLogger):
         i2 = ProgressBar._events_order.index(event2)
         return i1 < i2
 
-    @staticmethod
-    def log_message(message, **tqdm_kwargs):
+    def log_message(self, message):
         """
         Logs a message, preserving the progress bar correct output format.
 
@@ -136,7 +135,7 @@ class ProgressBar(BaseLogger):
             message (str): string you wish to log.
         """
         from tqdm import tqdm
-        tqdm.write(message, **tqdm_kwargs)
+        tqdm.write(message, **self.tqdm_kwargs)
 
     def attach(self, engine, metric_names=None, output_transform=None,
                event_name=Events.ITERATION_COMPLETED,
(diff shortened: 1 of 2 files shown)


vfdev-5:
@maxfrei750 thanks for the PR. I wonder whether we should not improve the progress bar in more global way to write into file (not only `log_message` method) ?

maxfrei750:
You're very welcome. In turn, I'd like to thank you for the great piece of software that you are maintaining. 

> I wonder whether we should not improve the progress bar in more global way to write into file (not only log_message method) ?

The progress bar itself can already be logged to a file with the current code by setting the `file` parameter of tqdm when initializing the `ProgressBar` object (via `**tqdm_kwargs`). However, like that, there is no output to `stderr` (the default output of tqdm) any longer. However, for some people that may be fine. 

I also thought about introducing a `file` attribute for the `ProgressBar` class, which could be set during the initialization. This attribute could then be used in the `log_message` method. However, this would result in `log_message` no longer being a static method, which in turn might lead to problems for users, if they don't have the `ProgressBar` object available when calling it. Therefore, I refrained from doing that, also taking into consideration that the developers of tqdm might already have implemented it that way, if they had believed it to be a good idea. The current solution is IMHO the most flexible one.

Personally, I ultimately just ended up writing a wrapper that logs everything that is being written to `stderr` also to a file. Like that, everything is logged and I still have the output to `stderr`. However, this solution is independent of the functionality of the ProgressBar handler. It might however be a good idea for a new functionality, which the user can enable to simply log everything that is echoed via `stderr` or `stdout` to a file.

vfdev-5:
@maxfrei750 thanks for the explanation !

> I also thought about introducing a file attribute for the ProgressBar class, which could be set during the initialization. 

I think this is a good idea 

> log_message no longer being a static method

`ProgressBar` is in the contrib module which potentially have some BC breaking changes. I think if we could introduce this feature to log into a file in a simple and unified way, this could worth a bearking change.


PS:
However, ultimately utility of logging into a file depends on the context of execution. For example, just running a script in a simple terminal we need that. If it is done through ML experiments tracking systems which naturally logs stdout/stderr into their files, we probably wont use it...

maxfrei750:
I applied the change that we suggested. In the end, I just made use of the already existent `tqdm_kwargs` attribute and did not introduce a redundant `file` attribute.

vfdev-5:
@maxfrei750 sorry for delayed reply. We need to update tests: https://travis-ci.org/pytorch/ignite/jobs/632726465#L1688 
and check this example if it runs : https://github.com/pytorch/ignite/blob/master/examples/contrib/mnist/mnist_with_tqdm_logger.py
thanks

vfdev-5:
@maxfrei750 thanks for the update! And I thought that it would be good to add a test explicitly specifiying a file name where to log and checking if file exists.
Also it would be useful for other users if we update the docstring with an example saying how to log into a file. What do you think ?

maxfrei750:
I just checked if the MNIST example still works as expected and it does.

> ... it would be useful for other users if we update the docstring with an example saying how to log into a file. What do you think?

I just added a small example.

>  I thought that it would be good to add a test explicitly specifiying a file name where to log and checking if file exists.

I also thought about that. However, in the end this is rather a test of a tqdm functionality than of an ignite functionality. Especially, because the pass through of tqdm arguments is already covered by `test_pbar_with_tqdm_kwargs`. If the file attribute was tested specifically, one could reason that it would be consistent to check all other tqdm arguments, which is no good idea from my perspective.

vfdev-5:
@maxfrei750 thanks !

maxfrei750:
@vfdev-5  Nice! Thank you as well.

PR #723, opened 2020-01-24
Dynamic typing

### LINKED ISSUE #719: Dynamic Typing
Sometimes there are types assumed and hardcoded, where it is not necessary.

For example there is [this part](https://github.com/pytorch/ignite/blob/master/ignite/engine/engine.py#L971) where we should infer the previous type of the data loader and re-use this. 
Other examples include type specific operations. E.g. we often check for mappings in general but then we only create dicts after doing some operations without restoring the original type. 

I will go through the core package and change this.

justusschock:
Fixes #719 

Description:
Infers types and restores them were necessary.

Check list:
* [x] New doc strings: description and/or example code are in RST format[Not Necessary]
* [x] Documentation is updated (if required)[Not Necessary]


justusschock:
there is one test, that needs to be adapted, I will take care of this

PR #724, opened 2020-01-24
Splits engine into multiple parts

### LINKED ISSUE #717: Refactor engine.py
Idea is to refactor engine.py into several files:
- engine.py
- events.py
- ...

cc @justusschock

justusschock:
Fixes #717 

Description:

Check list:


vfdev-5:
I tried to fixed pep8 using github interface.

justusschock:
I fixed it now (there were still some wrong function calls, which I did not see in the first place)

vfdev-5:
@justusschock yeah I see, I'll merge #723 and then this one.

ignite/engine/utils.py
```diff
@@ -0,0 +1,68 @@
+import inspect
```
vfdev-5: @justusschock I wonder if this file should not be private? 
We do not really provide `ReproducibleBatchSampler` as lib's API...

justusschock: Well we could for sure make it private. But I thought it's fine, since we don't expose it in the `__init__.py`

Issue #687, opened 2019-12-16
TypeError: 'Events' object is not callable


2356852131:
please help,
when I run the code:
 @trainer.on(Events.ITERATION_COMPLETED(every=log_interval))
    def log_training_loss(engine):
        print("Epoch[{}] Iteration[{}/{}] Loss: {:.2f}"
              "".format(engine.state.epoch, iter, len(train_loader), engine.state.output))
        writer.add_scalar("training/loss", engine.state.output, engine.state.iteration)
arise the error:
Traceback (most recent call last):
  File "F:/keyan/code/cloudImagePredict/train.py", line 327, in <module>
    run(args.train_batch_size, args.val_batch_size, args.epochs, args.lr, args.log_interval, args.log_dir, args.checkpoint_dir )
  File "F:/keyan/code/cloudImagePredict/train.py", line 269, in run
    @trainer.on(Events.ITERATION_COMPLETED(every=log_interval))
TypeError: 'Events' object is not callable

vfdev-5:
@2356852131 please use latest nightly release to have it working:
https://github.com/pytorch/ignite#nightly-releases

vfdev-5:
I close the issue as solved, please feel free to reopen it if need more support on the issue.

avangelizer:
Hi, I've still got the error even though I installed the latest nihtly release. I'm currently working on google colab. Don't know if that makes any difference.

vfdev-5:
@avangelizer please take a look at this colab where it works : 
https://colab.research.google.com/drive/1nUkbM8WqatZJWYs9DfeNgJVljjk0NdBO

PR #726, opened 2020-01-24
Dict support

justusschock:
This PR adds additional support for more data-types by adding an additional argument to the `create_supervised_trainer` and `create_supervised_evaluator` functions and by letting the `apply_to_tensor` function do the necessary type dispatches rather than doing them explicitly in the `_prepare_batch`

Check list:

A minimal example would be:
```python
import torch
from torch.nn import functional as F


class CustomDset(torch.utils.data.Dataset):
    def __init__(self, length):
        super().__init__()
        self.length = length

    def __getitem__(self, item):
        return {'data': torch.rand(1, 28, 28), 'label': torch.rand(1)}

    def __len__(self):
        return self.length


def get_data_loaders(train_batch_size, val_batch_size):
    train_loader = torch.utils.data.DataLoader(CustomDset(100), batch_size=train_batch_size, shuffle=True)
    val_loader = torch.utils.data.DataLoader(CustomDset(100), batch_size=val_batch_size, shuffle=False)
    return train_loader, val_loader


from ignite.engine import Events, create_supervised_trainer, create_supervised_evaluator
from ignite.metrics import Accuracy, Loss


class Net(torch.nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.conv1 = torch.nn.Conv2d(1, 10, kernel_size=5)
        self.conv2 = torch.nn.Conv2d(10, 20, kernel_size=5)
        self.conv2_drop = torch.nn.Dropout2d()
        self.fc1 = torch.nn.Linear(320, 50)
        self.fc2 = torch.nn.Linear(50, 1)

    def forward(self, x):
        x = F.relu(F.max_pool2d(self.conv1(x), 2))
        x = F.relu(F.max_pool2d(self.conv2_drop(self.conv2(x)), 2))
        x = x.view(-1, 320)
        x = F.relu(self.fc1(x))
        x = self.fc2(x)
        return F.log_softmax(x, dim=-1)

def run(bs):
    train_loader, val_loader = get_data_loaders(bs, bs)
    model = Net()
    device = 'cpu'

    optimizer = torch.optim.SGD(model.parameters(), lr=1e-3)

    trainer = create_supervised_trainer(model, optimizer, torch.nn.L1Loss(), device=device, get_data=lambda batch: (batch['data'], batch['label']))
    evaluator = create_supervised_evaluator(model, metrics={'loss': Loss(torch.nn.L1Loss())}, device=device, get_data=lambda batch: (batch['data'], batch['label']))

        evaluator.run(train_loader)
        print(metrics['loss'])

    trainer.run(train_loader, max_epochs=10)

if __name__ == '__main__':
    run(10)

```

I'll add tests and update docs, once I got a general feedback on that.

Note: This PR should be merged after #687 as I accidentally based my branch on this branch.

### DIFF
--- a/ignite/engine/__init__.py
+++ b/ignite/engine/__init__.py
@@ -1,4 +1,5 @@
 from typing import Sequence, Union, Optional, Callable, Dict, Any, Tuple
+
 import torch
 
 from ignite.engine.engine import Engine
@@ -19,16 +20,15 @@ def _prepare_batch(batch: Sequence[torch.Tensor], device: Optional[Union[str, to
     """Prepare batch for training: pass to a device with options.
 
     """
-    x, y = batch
-    return (convert_tensor(x, device=device, non_blocking=non_blocking),
-            convert_tensor(y, device=device, non_blocking=non_blocking))
+    return convert_tensor(batch, device=device, non_blocking=non_blocking)
 
 
 def create_supervised_trainer(model: torch.nn.Module, optimizer: torch.optim.Optimizer,
                               loss_fn: Union[Callable, torch.nn.Module],
                               device: Optional[Union[str, torch.device]] = None, non_blocking: bool = False,
                               prepare_batch: Callable = _prepare_batch,
-                              output_transform: Callable = lambda x, y, y_pred, loss: loss.item()) -> Engine:
+                              output_transform: Callable = lambda x, y, y_pred, loss: loss.item(),
+                              get_data: Callable = lambda batch: (batch[0], batch[1])) -> Engine:
     """
     Factory function for creating a trainer for supervised models.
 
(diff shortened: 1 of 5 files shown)


ignite/engine/__init__.py
```diff
@@ -19,16 +20,15 @@ def _prepare_batch(batch: Sequence[torch.Tensor], device: Optional[Union[str, to
...
+    return convert_tensor(batch, device=device, non_blocking=non_blocking)
 
 
 def create_supervised_trainer(model: torch.nn.Module, optimizer: torch.optim.Optimizer,
                               loss_fn: Union[Callable, torch.nn.Module],
                               device: Optional[Union[str, torch.device]] = None, non_blocking: bool = False,
                               prepare_batch: Callable = _prepare_batch,
-                              output_transform: Callable = lambda x, y, y_pred, loss: loss.item()) -> Engine:
+                              output_transform: Callable = lambda x, y, y_pred, loss: loss.item(),
+                              get_data: Callable = lambda batch: (batch[0], batch[1])) -> Engine:
```
vfdev-5: What is the difference between `get_data` and old `prepare_batch` ?

justusschock: In the previous implementation `prepare_batch` was responsible for pushing it to the correct device and dtype and for ensuring it's a tuple. This has now changed.

Now the `prepare_batch` supports arbitrary types (and still tuples), since it has no longer an explicit constraint on tuples, but the `get_data` function defines how we actually build the tuple we need out of an arbitrary data source.

PR #727, opened 2020-01-24
Add __all__ for core part

### LINKED ISSUE #725: __all__  for each source file
I want to add the `__all__` attribute to each source file in the core package, which defines, what should be imported by `*` (e.g `from ignite.metrics import *`). Per default this are all objects whose name is not starting with '_'. But this includes all imports from other files and packages, which is not desired.

justusschock:
Fixes #725 

Description:
Adds `__all__` to each source file to determine what should be imported, when users do `from package.x import *`

Check list:
* [x] New doc strings: description and/or example code are in RST format  [Not Necessary]
* [x] Documentation is updated (if required)  [Not Necessary]


ignite/_utils.py
```diff
@@ -2,6 +2,10 @@
 # For compatibilty
 from ignite.utils import convert_tensor, apply_to_tensor, apply_to_type, to_onehot
 
+__all__ = [
```
vfdev-5: This one is not a public API, we can ignore it ?

justusschock: I'll remove it

ignite/engine/__init__.py
```diff
@@ -3,6 +3,13 @@
 from ignite.engine.engine import Engine, State, Events
 from ignite.utils import convert_tensor
 
+__all__ = [
```
vfdev-5: Do we need to add it into __init__  too ?

justusschock: We should, since sometimes we also define functions within the `__init__` and import packages, which are actually not part of the API, but required for some checks or something. So this will give us a clean definition, of what to import.

PR #604, opened 2019-09-03
Added Time Profiler

Patil2099:
Fixes #589

Description: Added basic time profiling handler

Check list:
* [ ] New tests are added (if a new feature is added)
* [ ] New doc strings: description and/or example code are in RST format
* [ ] Documentation is updated (if required)

@Patil2099, do not remove above text, please!

PR #729, opened 2020-01-26
Issue 589: Add basic time profiler to contrib.handlers

### LINKED ISSUE #589: Basic Time Profiling into contrib
Following https://github.com/pytorch/ignite/issues/63#issuecomment-523366321
we can provide some basic time profiling. 

cc @willprice what do you think about this profiler, any code, API updates to suggest ?

ykumards:
Fixes #589

Description:

Check list:


ykumards:
Hello @vfdev-5,

I picked off from PR #604

vfdev-5:
@ykumards thanks for the PR

I start with the second question:

> I found that there are two other events (Events.GET_BATCH_COMPLETED, Events.GET_BATCH_STARTED) that is not handled in the profiler code. Is this fine? I just added them to events_to_ignore for now.

These events were introduced very recently, much after `BasicTimeProfiler`, so yes we need to adapt the class for these events. I think even that they would simlify some of actual logic of `BasicTimeProfiler`.

> I am not sure how to come up with test cases for the data_flow_timers. If I understand the code correctly, they are timed between the Events.STARTED and the first Events.ITERATION_STARTED. Does this mean the delay has to be tested in the dataloader function? Am not really sure how to go about this.

Yes, idea is to provide some measures on dataflow preparation: dataflow times vs forward/backward passes times.
To test this you can setup a data iterator, something like this
```python
import time

def data_iterator(data):
    while True:
        for d in data:
            time.sleep(0.01)
            yield d
```

ignite/contrib/handlers/time_profilers.py
```diff
@@ -0,0 +1,302 @@
...
+
+    @staticmethod
+    def write_results(output_path):
+        try:
+            import pandas as pd
+        except ImportError:
+            print("Need pandas to write results as files")
+            return
+
+        raise NotImplementedError("")
```
vfdev-5: Maybe we can also implement this if this make sense log times in a file.

ignite/contrib/handlers/time_profilers.py
```diff
@@ -0,0 +1,302 @@
...
+            ('mean', torch.mean(data).item()),
+            ('std', torch.std(data).item()),
+            ('total', torch.sum(data).item())
+        ])
+
+    def get_results(self):
+        events_to_ignore = [
+            Events.EXCEPTION_RAISED,
+            Events.GET_BATCH_COMPLETED,
+            Events.GET_BATCH_STARTED
```
vfdev-5: We need to check if it is not more simple to replace current logic for measuring dataflow time using these new events.

tests/ignite/contrib/handlers/test_time_profilers.py
```diff
@@ -0,0 +1,271 @@
...
+    profiler.attach(dummy_trainer)
+    dummy_trainer.run(
+        dummy_data_loader(dummy_data),
+        max_epochs=true_max_epochs,
+        epoch_length=true_num_iters
+    )
+    results = profiler.get_results()
+    dataflow_results = results['dataflow_stats']
+
+    assert _equal(dataflow_results['min/index'][0], true_dataflow_time_per_ele)
```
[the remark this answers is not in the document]
ykumards: done.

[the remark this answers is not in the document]
vfdev-5: Thanks

ignite/contrib/handlers/time_profilers.py
ykumards: Hello,
Since the `get_results` method returns a nested OrderedDict, writing it into one csv file looks a bit tricky. We could either create separte files for `processing_stats`, `dataflow_stats`, etc or dump the results from the `print_results()` method into a text file (the output is already nicely formatted). Thoughts?

vfdev-5: Idea is to dump into csv non-aggregated times:

epoch | iteration | processing_stats | dataflow_stats | STARTED | EPOCH_STARTED | ITERATION_STARTED | ...
---|---|---|---|---|---|---|---
1 | 1 | 0.01 | 0.02 | 0.03 | 0.04 | 0.05
1 | 2 | 0.01 | 0.02 |  |  | 0.05
1 | 3 | 0.01 | 0.02 |  |  | 0.05

Something like that.

ignite/contrib/handlers/time_profilers.py
ykumards: Hello! Latest commit should fix this.

vfdev-5:
@ykumards thanks! So, it remains a dump to CSV and we are done :)

ignite/contrib/handlers/time_profilers.py
ykumards: Hi @vfdev-5

vfdev-5: Well, maybe we can do this in a single step (with a non-static method) without asking user to dump results. We can set into `__init__` the attributes
```
        self.dataflow_times = None
        self.processing_times = None
        self.event_handlers_times = None
```
such that user can reuse them freely. So, I would remove `dump_results` method.

Can you also post a screen shot of obtained csv as a table please.

ykumards: Yes, that sounds good. I pushed the changes.

Also, attaching the screenshot. Does this look good?
file_dump

ykumards: Oh sorry, you wanted the screenshot as a table. PFA
Screenshot 2020-02-02 at 17 08 37

vfdev-5: @ykumards table looks very cool :)

ignite/contrib/handlers/time_profilers.py
```diff
@@ -0,0 +1,348 @@
...
+class BasicTimeProfiler(object):
+    """
+    BasicTimeProfiler can be used to profile the handlers,
+    events, data loading and data processing times.
+
+    Args:
+        None
+
+    Attributes:
+        data_flow_times (torch.Tensor): tile elapsed during data loading
```
vfdev-5:
```suggestion
        data_flow_times (torch.Tensor): time elapsed during data loading
```

ignite/contrib/handlers/time_profilers.py
```diff
@@ -0,0 +1,348 @@
...
+        >>> profiler = BasicTimeProfiler()
+        >>> trainer = Engine(train_updater)
+        >>> profiler.attach(trainer)
+        >>> trainer.run(dataloader, max_epochs=3)
+
+        Once the engine is run, we can fetch the profiler stats using,
+
+        >>> results = profiler.get_results()
+        >>> profiler.print_results(results)
+
```
vfdev-5: @ykumards could you please insert a example of `print_results` output. Such that user has an idea the results.

ignite/contrib/handlers/time_profilers.py
```diff
@@ -0,0 +1,348 @@
...
+        Also optionally, you could choose to store the profiling results
+        into a file using,
+
+        >>> profiler.write_results('path_to_store/profiler_results.csv')
```
vfdev-5: @ykumards maybe same here, could you please insert an example of output csv.

vfdev-5:
And, for instance, let's do not pay attention to codecov/patch

ignite/contrib/handlers/time_profilers.py
ykumards: Latest commit should fix these...

ignite/contrib/handlers/time_profilers.py
```diff
@@ -0,0 +1,394 @@
...
+        >>> profiler.attach(trainer)
+        >>> trainer.run(dataloader, max_epochs=3)
+
+        Once the engine is run, we can fetch the profiler stats using,
+
+        >>> results = profiler.get_results()
+        >>> profiler.print_results(results)
+
+        The results would look as follows:
+        --------------------------------------------
```
vfdev-5: @ykumards I think we need to insert the output inside
```
    .. code-block:: text

        - Time profiling results:
        --------------------------------------------
        Processing function time stats (in seconds):
            min/index: (2.9754010029137135e-05, 0)
            max/index: (2.9754010029137135e-05, 0)
            mean: 2.9754010029137135e-05
            std: nan
            total: 2.9754010029137135e-05
```

ignite/contrib/handlers/time_profilers.py
```diff
@@ -0,0 +1,394 @@
...
+        Also optionally, you could choose to store the profiling results
+        into a file using,
+
+        >>> profiler.write_results('path_to_store/profiler_results.csv')
+
+        The CSV file contains the unaggregated profiler values for each
+        event type, processing stats and dataflow stats and may look as
+        follows:
+
+        --------------------------------------------
```
vfdev-5: Same here, please use
```
    .. code-block:: text

        --------------------------------------------
        epoch iteration processing_stats dataflow_stats Event_STARTED ...
```

vfdev-5:
@ykumards minor fixes still needed i think. I can check documentation rendering using:
```
cd docs && pip install -r requirements.txt && make html
cd build/html && python -m http.server 1234
```
In the browser open 0.0.0.0:1234 and it should be there.

ykumards:
Hello @vfdev-5 

Thanks! The docstring looks better formatted now. I moved the code and output examples to the corresponding function definitions, hoe that's okay.

Screenshots attached below:

Screen Shot 2020-02-04 at 10 48 53 AM
Screen Shot 2020-02-04 at 10 49 00 AM

ykumards:
> LGTM! Thanks Yogesh!

Just made one tiny change to fix these guys :D 

Screen Shot 2020-02-04 at 10 49 00 AM

PR #734, opened 2020-01-27
Fix checkpoint score documentation

### LINKED ISSUE #733: Checkpoint filename not using absolute score value
The [Checkpoint documentation](https://pytorch.org/ignite/handlers.html?highlight=checkpoint#ignite.handlers.Checkpoint) mentions the use of the absolute value of the score in the filename. 

> If score_function not None, it is possible to store its absolute value using score_name. See Notes for more details.

> If defined score_function and score_name, then the filename will be {filename_prefix}_{name}_{score_name}={abs(score)}.{ext}.

However, it seems the score (priority) is not used as an absolute value:
https://github.com/pytorch/ignite/blob/a1cae3148a92ddc82bfb1412859e04d4b51f0f25/ignite/handlers/checkpoint.py#L186

The `abs` call was removed in #679.

Joxis:
Fixes #733  

Description:
Update the Checkpoint documentation with the change to allow negative score values instead of the absolute value in the filename.

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -180,16 +180,17 @@ class Checkpoint:
 
         if self._check_lt_n_saved() or self._saved[0].priority < priority:
 
+            priority_str = '{:.4f}'.format(priority)
             if self._score_name is not None:
                 if len(suffix) > 0:
                     suffix += "_"
-                suffix = "{}{}={}".format(suffix, self._score_name, priority)
+                suffix = "{}{}={}".format(suffix, self._score_name, priority_str)
             elif self._score_function is not None:
                 if len(suffix) > 0:
                     suffix += "_"
-                suffix = "{}{}".format(suffix, priority)
+                suffix = "{}{}".format(suffix, priority_str)
             elif len(suffix) == 0:
-                suffix = "{}".format(priority)
+                suffix = "{}".format(priority_str)
 
             checkpoint = self._setup_checkpoint()
 
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -180,17 +180,16 @@ class Checkpoint:
 
         if self._check_lt_n_saved() or self._saved[0].priority < priority:
 
-            priority_str = '{:.4f}'.format(priority)
             if self._score_name is not None:
                 if len(suffix) > 0:
                     suffix += "_"
-                suffix = "{}{}={}".format(suffix, self._score_name, priority_str)
+                suffix = "{}{}={:.4f}".format(suffix, self._score_name, priority)
             elif self._score_function is not None:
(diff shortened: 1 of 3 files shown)


ignite/handlers/checkpoint.py
```diff
@@ -61,11 +61,11 @@ class Checkpoint:
         The filename will be `{filename_prefix}_{name}_{global_step}_{score}.pth`.
 
         If defined `score_function` and `score_name`, then the filename will
-        be `{filename_prefix}_{name}_{score_name}={abs(score)}.{ext}`. If `global_step_transform` is provided, then
-        the filename will be `{filename_prefix}_{name}_{global_step}_{score_name}={abs(score)}.{ext}`
+        be `{filename_prefix}_{name}_{score_name}={score}.{ext}`. If `global_step_transform` is provided, then
+        the filename will be `{filename_prefix}_{name}_{global_step}_{score_name}={score}.{ext}`
 
         For example, `score_name="val_loss"` and `score_function` that returns `-loss` (as objects with
```
vfdev-5: To better clarify:
```suggestion
        For example, `score_name="neg_val_loss"` and `score_function` that returns `-loss` (as objects with
```

ignite/handlers/checkpoint.py
```diff
@@ -61,11 +61,11 @@ class Checkpoint:
...
-        highest scores will be retained), then saved filename will be `{filename_prefix}_{name}_val_loss=0.1234.pth`.
+        highest scores will be retained), then saved filename will be `{filename_prefix}_{name}_val_loss=-0.1234.pth`.
```
```suggestion
        highest scores will be retained), then saved filename will be `{filename_prefix}_{name}_neg_val_loss=-0.1234.pth`.
```

vfdev-5:
Sorry, my changes broke flake8. We need split the line into two.

Joxis:
@vfdev-5 I'm on it. I will also add the extra formatting.

Joxis:
@vfdev-5 Would you like the decimal places formatting to be explicitly mentioned in the docs?

vfdev-5:
To have 100% user-friendly documentation, yes, please, we can mention that :)

Joxis:
@vfdev-5 The conflict is due to the fixed 4 decimal places, even when the original value has less decimal places. Is this desired or do you prefer rounding the value to enforce a maximum on the decimal places rather that having it always fixed to 4?

vfdev-5:
@Joxis we can simply adapt true values as we expected them:
```
# assert set(os.listdir(dirname)) == set(['best_model_8_val_acc=0.61.pth', 'best_model_9_val_acc=0.7.pth'])
assert set(os.listdir(dirname)) == set(['best_model_8_val_acc=0.6100.pth', 'best_model_9_val_acc=0.7000.pth'])
```

PR #736, opened 2020-01-27
Fixes #735, Replaced RuntimeWarning with EpochMetricWarning

### LINKED ISSUE #735: RuntimeWarning: Only one class present in y_true. ROC AUC score is not defined in that case..
Hi, I am using ROC_AUC metric to track the training performance. However, this warning frequently appears:

> RuntimeWarning: Probably, there can be a problem with `compute_fn`: Only one class present in y_true. ROC AUC score is not defined in that case.

It seems the ROC_AUC compute_fn will be called every time we update the (y, y_pred) accumulator. In my scenario, it's possible that one batch only contains negative samples (unbalanced classes). So this warning may be annoying. 

Is there any way to disable this warning? Or whether it's better to change the behavior of this sanity check for the AUC calculation? (e.g. check the label/pred type after collecting all the data) 

Appreciate your help!

vfdev-5:
Fixes #735  

Description: following the discussion in #735 
Replaced RuntimeWarning with EpochMetricWarning
- if compute_fn raises an error on batch data
such that it can be filtered if compute_fn fails only on batch data and not on complete data.

Check list:

cc @wangyirui

### DIFF
--- a/ignite/metrics/epoch_metric.py
+++ b/ignite/metrics/epoch_metric.py
@@ -80,7 +80,11 @@ class EpochMetric(Metric):
                 self.compute_fn(self._predictions, self._targets)
             except Exception as e:
                 warnings.warn("Probably, there can be a problem with `compute_fn`:\n {}.".format(e),
-                              RuntimeWarning)
+                              EpochMetricWarning)
 
     def compute(self):
         return self.compute_fn(self._predictions, self._targets)
+
+
+class EpochMetricWarning(UserWarning):
+    pass
--- a/tests/ignite/metrics/test_epoch_metric.py
+++ b/tests/ignite/metrics/test_epoch_metric.py
@@ -2,6 +2,7 @@ import os
 import torch
 
 from ignite.metrics import EpochMetric
+from ignite.metrics.epoch_metric import EpochMetricWarning
 
 import pytest
 
@@ -113,7 +114,7 @@ def test_bad_compute_fn():
 
     em.reset()
     output1 = (torch.rand(4, 3), torch.randint(0, 2, size=(4, 4), dtype=torch.long))
-    with pytest.warns(RuntimeWarning):
+    with pytest.warns(EpochMetricWarning, match=r"Probably, there can be a problem with `compute_fn`"):
         em.update(output1)
 
 


vfdev-5:
@justusschock can you please review. thanks

tests/ignite/metrics/test_epoch_metric.py
```diff
@@ -113,7 +114,7 @@ def compute_fn(y_preds, y_targets):
 
     em.reset()
     output1 = (torch.rand(4, 3), torch.randint(0, 2, size=(4, 4), dtype=torch.long))
-    with pytest.warns(RuntimeWarning):
+    with pytest.warns(EpochMetricWarning, match=r"Probably, there can be a problem with `compute_fn`"):
```
justusschock: Just a question: Is there a reason to use a raw string here but not in the actual implementation?

ignite/metrics/epoch_metric.py
```diff
@@ -80,7 +80,11 @@ def update(self, output):
...
             except Exception as e:
                 warnings.warn("Probably, there can be a problem with `compute_fn`:\n {}.".format(e),
-                              RuntimeWarning)
+                              EpochMetricWarning)
 
     def compute(self):
         return self.compute_fn(self._predictions, self._targets)
+
+
+class EpochMetricWarning(UserWarning):
```
justusschock: I think, this class should be moved to ignite/exceptions.py

tests/ignite/metrics/test_epoch_metric.py
vfdev-5: This can be helpful if we would like to use it with a regexp.

ignite/metrics/epoch_metric.py
vfdev-5: I thought about this. This warning is very specific to `EpochMetric` and it is not an exception...

justusschock: Currently it may be specific to `EpochMetric`. But what if we just rename it to e.g. `EpochWarning` and slightly change the text associated with this warning to make it more general (so that we could for example reuse it in the `Engine`?

(And btw: every warning is also an exception: https://docs.python.org/3/library/warnings.html#warning-categories)

justusschock: Currently it may be specific to `EpochMetric`. But what if we just rename it to e.g. `EpochWarning` and slightly change the text associated with this warning to make it more general (so that we could for example reuse it in the `Engine`?

(And btw: every warning is also an exception: https://docs.python.org/3/library/warnings.html#warning-categories)

vfdev-5: > And btw: every warning is also an exception: https://docs.python.org/3/library/warnings.html#warning-categories

I see :) Thanks 

Well, here it is introduced to be able to filter it as the issue states that. So reusing this warning can be later misleading, IMO.

justusschock: From our discussion somewhere else: 
> For EpochMetricWarning is just to ignore a sanity check as it is annoying in some cases as issues stated. Reusing is not a good idea I think

PR #738, opened 2020-01-28
[ignite/contrib/metrics/pr_auc.py] Area Under Precision Recall Curve

SamuelMarks:
Fixes # 

Description:
TensorFlow has one, so you should too! - https://www.tensorflow.org/api_docs/python/tf/keras/metrics/AUC

Check list:

TODO: tests!

Alternatively I can edit this PR to rename `ignite/contrib/metrics/roc_auc.py` to `ignite/contrib/metrics/auc.py` and accept a `curve` keyword argument, akin to TensorFlow's.

### DIFF
--- a/ignite/contrib/metrics/pr_auc.py
+++ b/ignite/contrib/metrics/pr_auc.py
@@ -1,0 +1,42 @@
+from ignite.metrics import EpochMetric
+
+
+def pr_auc_compute_fn(y_preds, y_targets):
+    try:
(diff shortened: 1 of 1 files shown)


vfdev-5:
@SamuelMarks thanks for the PR! 
Seems like in TF they can compute it in online fashion vs here using sklearn on whole data.

If you could add tests like for ROC_AUC we could merge this.

vfdev-5:
@SamuelMarks any updates on that ?

PR #739, opened 2020-01-28
Remove all occurences of `from __future__ import print_function`

### LINKED ISSUE #732: Remove from __future__ import print_function from all examples
We need to remove `from __future__ import print_function` from all examples, as it was forgotten while dropping python 2.7 support.

justusschock:
Fixes #732 

Description:
Removes all occurrences of `from __future__ import print_function`, which were necessary for Python 2.7 support

Check list:

### DIFF
--- a/examples/contrib/mnist/mnist_with_tensorboard_logger.py
+++ b/examples/contrib/mnist/mnist_with_tensorboard_logger.py
@@ -17,8 +17,6 @@
     python mnist_with_tensorboard_logger.py --log_dir=/tmp/tensorboard_logs
     ```
 """
-from __future__ import print_function
-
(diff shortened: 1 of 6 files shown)


CDitzel:
I recently got the error

>    from __future__ import annotations
    ^
SyntaxError: future feature annotations is not defined

what is this about?

vfdev-5:
@CDitzel it is related to typing (https://www.python.org/dev/peps/pep-0563/#enabling-the-future-behavior-in-python-3-7). And seems like it is a regression we introduced recently as this only work since 3.7...

see #752

CDitzel:
thank you!

PR #740, opened 2020-01-28
Support namedtuple for "convert_tensor"

hoangmit:
convert_tensor support Map, List, but not namedtuple.

This change is similar to: https://github.com/pytorch/pytorch/blob/master/torch/utils/data/_utils/pin_memory.py#L45

Fixes # 

Description:

Check list:


vfdev-5:
@hoangmit thanks for the PR ! Looks good.

Issue #710, opened 2020-01-21
Handler Events as flags


vfdev-5:
Idea is to provide the following flexibility:
```python

@trainer.on(Events.ITERATION_STARTED(once=1) | Events.ITERATION_STARTED(every=10) | Events.COMPLETED)
def execute_some_handler(_):
    pass
```
such that `execute_some_handler` is triggered at the first iteration and at 10th, 20th, ...etc and in the end of the training.

justusschock:
what we could most likely implement way easier is to provide something like 
```python
@trainer.on(Events.ITERATION_STARTED(once=1), Events.ITERATION_STARTED(every=10) , Events.COMPLETED)
def execute_some_handler(_):
    pass
```
where each argument given there is concatenated with `|`

vfdev-5:
@justusschock here is a subtle detail is that the signature of 
```
@trainer.on(<Event>, *args, **kwargs)
```
so, `Events.ITERATION_STARTED(once=1), Events.ITERATION_STARTED(every=10) , Events.COMPLETED` means `<Event>=Events.ITERATION_STARTED(once=1)` and others are `*args`...

justusschock:
Exactly, that's why we would just have to introduce additional checks there, which will wrap them to a function, but this is way easier to implement than having them connected by the user as a boolean via `|`. Because currently I see no way to implement flagging behaviour without having to previously evaluate them for each possible event value.

If you're okay on separating them with ',' instead of '|', I will start implementing this.

PR #744, opened 2020-01-29
Formalize event filters

justusschock:
This PR aims to formalise the event filters by introducing a separate class for them. This is a necessary preparation for #710 since we need to overload their magic methods.

Check list:

Docs and Docstrings will be updated once we all agree on the code part of this PR

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -13,7 +13,7 @@ from typing import Union, Optional, Callable, Iterable, Iterator, Any, \
 
 import torch
 
-from ignite.engine.events import Events, State, EventWithFilter, RemovableEventHandle
+from ignite.engine.events import Events, State, Filter, RemovableEventHandle, EventWithFilter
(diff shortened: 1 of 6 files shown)


ignite/engine/engine.py
```diff
@@ -210,8 +211,7 @@ class TBPTT_Events(CallableEvents, Enum):
     def _handler_wrapper(handler: Callable, event_name: str, event_filter: Callable) -> Callable:
```
justusschock: Here we could probably remove the event_name argument, but I haven't done this so far to avoid breaking changes

ignite/engine/events.py
```diff
@@ -63,9 +87,6 @@ def __call__(self, event_filter: Optional[Callable] = None,
             raise ValueError("Argument every should be integer and positive")
 
         if every is not None:
-            if every == 1:
```
vfdev-5: Why to remove this ?

PR #745, opened 2020-01-29
Checkpoint: fix returning worst model of the saved models.

chm90:
Fixes # 
returning the model with the lowest priority of the saved models
Description:
The `self._saved` list is always sorted in an ascending order(ensured by sorting it each time a new element is added). Hence, returning `self._saved[0]` in `last_checkpoint` actually yields the oldest checkpoint of the saved ones, or the worst if a score function is provided. Changing the above to`self._saved[-1]` yields the expected result, i.e, the model with the highest priority.

Check list:
* [X ] Documentation is updated (if required)

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -126,8 +126,8 @@ class Checkpoint:
     Item = namedtuple("Item", ["priority", "filename"])
 
     def __init__(self, to_save: dict, save_handler: Callable, filename_prefix: str = "",
-                 score_function: Optional[Callable] = None, score_name: Optional[str] = None, n_saved: int = 1,
-                 global_step_transform: Callable = None, archived: bool = False):
+                 score_function: Optional[Callable] = None, score_name: Optional[str] = None,
+                 n_saved: Optional[int] = 1, global_step_transform: Callable = None, archived: bool = False):
 
         if not isinstance(to_save, collections.Mapping):
             raise TypeError("Argument `to_save` should be a dictionary, but given {}".format(type(to_save)))
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -161,7 +161,7 @@ class Checkpoint:
     def last_checkpoint(self) -> str:
         if len(self._saved) < 1:
             return None
-        return self._saved[0].filename
+        return self._saved[-1].filename
 
     def _check_lt_n_saved(self, or_equal=False):
         if self._n_saved is None:
@@ -404,7 +404,7 @@ class ModelCheckpoint(Checkpoint):
     def last_checkpoint(self) -> Union[str, None]:
         if len(self._saved) < 1:
             return None
-        return os.path.join(self.save_handler.dirname, self._saved[0].filename)
+        return os.path.join(self.save_handler.dirname, self._saved[-1].filename)
 
     def __call__(self, engine: Engine, to_save: Mapping) -> None:
 
--- a/tests/ignite/handlers/test_checkpoint.py
+++ b/tests/ignite/handlers/test_checkpoint.py
@@ -273,6 +273,43 @@ def test_checkpoint_with_score_name_and_function_and_trainer_epoch():
     _test(to_save, model.state_dict(), 'model')
 
 
+def test_checkpoint_last_checkpoint():
+    save_handler = MagicMock()
+    save_handler.remove = MagicMock()
+    to_save = {'model': DummyModel()}
+
+    checkpointer = Checkpoint(to_save, save_handler=save_handler, n_saved=None)
+
+    trainer = Engine(lambda e, b: None)
+
+    for i in range(10):
+        trainer.state = State(epoch=1, iteration=i)
+        checkpointer(trainer)
+
+    assert save_handler.call_count == 10
+    assert checkpointer.last_checkpoint == "{}_9.pth".format('model')
+
+
+def test_checkpoint_last_checkpoint_on_score():
+    save_handler = MagicMock()
+    save_handler.remove = MagicMock()
+    to_save = {'model': DummyModel()}
+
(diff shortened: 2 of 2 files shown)


vfdev-5:
@chm90 thanks for the PR. I need to recheck that. I have a llttle doubt on that. 
Anyway, I think we should have some explicit tests on that.

vfdev-5:
@chm90 as it is a bug, I added tests by myself. So, thanks again for catching it !

chm90:
Thanks @vfdev-5, I didn't have time to implement tests yet. Thank's for implementing

chm90:
@vfdev-5, do I need to do anything else for you to be able to merge the pr?

Nic-Ma:
Hi @vfdev-5 ,

Have you added this bug fix to nightly build?
I faced this same issue before and manually get last_checkpoint in my project.
If you updated, I can update my code immediately.
Thanks.

vfdev-5:
Hi @Nic-Ma 

This commit will be normally added to the nighlty build today (30/01/2020). 
Cron job that built nightly release for 29/01/2020 is up to [Minor docstring fix (#743)](https://github.com/pytorch/ignite/pull/743) commit.
https://travis-ci.org/pytorch/ignite/builds/643449008

Nic-Ma:
Hi @vfdev-5 ,

Thanks for your information.
I already updated to your nightly build, but faced another issue:
` File "/usr/local/lib/python3.5/dist-packages/ignite/__init__.py", line 1, in <module>
    import ignite.engine
  File "/usr/local/lib/python3.5/dist-packages/ignite/engine/__init__.py", line 4, in <module>
    from ignite.engine.engine import Engine
  File "/usr/local/lib/python3.5/dist-packages/ignite/engine/engine.py", line 15, in <module>
    from ignite._utils import _to_hours_mins_secs
  File "/usr/local/lib/python3.5/dist-packages/ignite/_utils.py", line 4, in <module>
    from ignite.utils import convert_tensor, apply_to_tensor, apply_to_type, to_onehot
  File "/usr/local/lib/python3.5/dist-packages/ignite/utils.py", line 44, in <module>
    input_type: Union[Type, Tuple[Type[Any], Any]],
  File "/usr/lib/python3.5/typing.py", line 552, in __getitem__
    dict(self.__dict__), parameters, _root=True)
  File "/usr/lib/python3.5/typing.py", line 512, in __new__
    for t2 in all_params - {t1} if not isinstance(t2, TypeVar)):
  File "/usr/lib/python3.5/typing.py", line 512, in <genexpr>
    for t2 in all_params - {t1} if not isinstance(t2, TypeVar)):
  File "/usr/lib/python3.5/typing.py", line 1077, in __subclasscheck__
    if super().__subclasscheck__(cls):
  File "/usr/lib/python3.5/abc.py", line 225, in __subclasscheck__
    for scls in cls.__subclasses__():
TypeError: descriptor '__subclasses__' of 'type' object needs an argument`

Is it compatible with Python 3.5?
Thanks.

Nic-Ma:
@vfdev-5 ,

And BTW, I found some small bugs in current checkpoint handler:
1. The latest code added logic: **suffix = "{}{:.4f}"**, which forces the "priority" to be float number.
2. There is some mixed use of **" "** and **' ',** which causes the saved filename includes a ' '.

For example, the saved model filename for epoch 2 will be:
**'checkpoint_epoch=2.0000.pth'**

Could you please have a look?
Thanks.

vfdev-5:
@Nic-Ma thanks for the feedback, seems like it is a regression we introduced with typing. Sorry about that. We will fix it ASAP.

About the suffix, yes that's right, it should be related to another PR where such case was somehow out of scope. I'll take a look on that. Thanks again for reporting these things!

Nic-Ma:
@vfdev-5 ，

Cool, thanks for your quick fix.
I also faced other issues even when I updated to python 3.6:
If I extended torch.DataLoader and defined my own dataloader, it will crash when Ignite **replaces it** with torch.Dataloader for reproduction feature(_I already named all other memeber vars start with "\_"_).
The exception is: **__init__ error, unexpected keyname "dataset"** when re-construct torch.DataLoader object.

Anyway, we will still use **ignite 0.3.0** until the 0.4.0 version is stable and officially released.
Thanks.

vfdev-5:
> If I extended torch.DataLoader and defined my own dataloader, it will crash when Ignite replaces it with torch.Dataloader for reproduction feature(I already named all other memeber vars start with "_").
 
@Nic-Ma probably this should be related with an issue and its fix here: https://github.com/pytorch/ignite/blob/b5938178c745dc7d1479816a82e8dbf4e9831f83/ignite/engine/utils.py#L15
So, I would say that in python 3.7 you most probably should not have the issue. But, it would depend on what kind of extension you made in `torch.DataLoader`. 

Could you please open a separate issue and put a small snippet to reproduce the problem from our side. Or pointout an example if it is in your MONAI project. Thanks!

Nic-Ma:
Hi @vfdev-5 ,

Thanks for your tips.
We are using python 3.6 for MONAI project, and I encountered this issue when I tested locally.
We developed many arch design and haven't been in public repo yet.
It's not very easy to reproduce this issue because I ran in our NVIDIA public PyTorch docker env 
https://ngc.nvidia.com/catalog/containers/nvidia:pytorch and extended DataLoader.
I think 0.3.0 is enough for our current features, will continue to develop things until 0.4.0 released,
then evaluate new features to use.
Thanks.

vfdev-5:
@Nic-Ma I see, thanks. If you could however share some more details on your extended DataLoader (e.g. a snippet or just a description of what kind of extension you made) such that we could create some tests for that and ensure that ignite works for various cases. Thanks

> I think 0.3.0 is enough for our current features, will continue to develop things until 0.4.0 released, then evaluate new features to use.

Certainly, we will release several 0.3.X versions before 0.4.0

PR #751, opened 2020-01-30
Add missing type annotation

### LINKED ISSUE #747: correct the type hints of this function
https://github.com/pytorch/ignite/blob/ca738d8f3f106093aa04b6bce9506129a1059df8/ignite/engine/events.py#L81

this really should read

`    def wrapper(engine, event: int) -> bool:`

justusschock:
Fixes #747 

Description:
Adds missing type annotation for wrapper function.

Check list:


ignite/engine/events.py
```diff
@@ -78,7 +78,7 @@ def __call__(self, event_filter: Optional[Callable] = None,
 
     @staticmethod
     def every_event_filter(every: int) -> Callable:
-        def wrapper(engine, event: bool):
+        def wrapper(engine, event: bool) -> bool:
```
vfdev-5:
```suggestion
        def wrapper(engine, event: int) -> bool:
```

Issue #753, opened 2020-01-30
Setup github CI for python 3.5 and 3.6


vfdev-5:
Idea is to setup and run unit tests only with Github CI (.github/workflows/pythonapp.yml) for python 3.5 and 3.6 as they were removed from travis recently.

An example project using Github CI: [here](https://github.com/Project-MONAI/MONAI/blob/master/.github/workflows/pythonapp.yml).

vfdev-5:
Done by #762

PR #755, opened 2020-01-31
Update .travis.yml

justusschock:
Adds unittests for Python 3.5 and 3.6

Currently all stages should be run 3 times: Once for Python 3.5, Python 3.6 and Python 3.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,42 @@
 language: python
 
-python:
-  - "3.7"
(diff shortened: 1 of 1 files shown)


vfdev-5:
As discussed in slack, it would be better to address it following https://github.com/pytorch/ignite/issues/753

vfdev-5:
Closed in favor of #762

PR #762, opened 2020-02-01
Enable Github CI

justusschock:
Adds GitHub Action to Test on CPU for all Python versions

Fixes #658 
cc @vfdev-5 

Have a look at https://github.com/justusschock/ignite/actions/runs/33776965 for an example run

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -1,4 +1,4 @@
-on: [push]
+on: [pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


vfdev-5:
@justusschock looks great! 

What do you think about removing smoke tests of examples? I thinks we can have only unit tests.

justusschock:
I think, I would leave these examples, they are not consuming that much time (compared to the unit tests)

vfdev-5:
Yes, they take ~4.3min. Well, let's keep them.

vfdev-5:
I have no idea how to configure now Github Actions from yaml. Interface proposes only to create from scratch or prebuilt things, but no options from existing yaml.
Also there is no way to remove old removed Actions... :(

PR #768, opened 2020-02-06
Fixed ProgressBar.log_message.

### LINKED ISSUE #767: log_message() method fails `desc` is passed to contrib tqdm
## 🐛 Bug description

If I pass a `desc` argument when instantiating a `ignite.contrib.handlers.ProgressBar` the calls to its `log_message()` method fail with this exception:

```
TypeError: write() got an unexpected keyword argument 'desc'
```

## Environment

 - PyTorch Version: 1.3.1
 - Ignite Version: 0.3.0
 - OS: Linux
 - How you installed Ignite (`conda`, `pip`, source): Conda
 - Python version: 3.7

maxfrei750:
Fixes #767.

Check list:

### DIFF
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -146,12 +146,7 @@ class ProgressBar(BaseLogger):
         """
         from tqdm import tqdm
 
-        if "file" in self.tqdm_kwargs:
-            file = self.tqdm_kwargs["file"]
-        else:
-            file = None
-
-        tqdm.write(message, file=file)
+        tqdm.write(message, file=self.tqdm_kwargs.get("file", None))
 
     def attach(self, engine, metric_names=None, output_transform=None,
                event_name=Events.ITERATION_COMPLETED,
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -145,7 +145,13 @@ class ProgressBar(BaseLogger):
             message (str): string you wish to log.
         """
         from tqdm import tqdm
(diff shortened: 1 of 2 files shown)


maxfrei750:
I added a new test to fix the coverage. However, on my machine nothing was written to the test output file. I cannot see why not, so I wanted to give it a try on the test system.

maxfrei750:
@vfdev-5 Ok, I found the error. The test output file was just written after it had been closed at program termination.

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -145,7 +145,13 @@ def log_message(self, message):
             message (str): string you wish to log.
         """
         from tqdm import tqdm
-        tqdm.write(message, **self.tqdm_kwargs)
+
+        if "file" in self.tqdm_kwargs:
```
justusschock: can we replace this part with `tqdm.write(message, file=self.tqdm_kwargs.get('file', None))`?

maxfrei750: Sure! That's way more elegant.

vfdev-5:
Seems like tests on pytorch-nightly are not using nightly builds: https://travis-ci.org/pytorch/ignite/jobs/647336703#L514
This may fail CI etc...

maxfrei750:
@vfdev-5 Sorry, but I don't seem to understand. Is there anything that can be done or will the problem just take some time to resolve?

vfdev-5:
@maxfrei750 sorry, I just wanted to mention that there is a problem with CI and it does not depend on your PR. So, yes, probably it would take time to resolve it...

PR #769, opened 2020-02-07
Make all Events be CallableEventsWithFilter

justusschock:
Preparation for #710  

Description:
Ensures all events are of same type now.

Check list:


ignite/engine/engine.py
```diff
@@ -268,11 +270,11 @@ def print_epoch(engine):
 
         return RemovableEventHandle(event_name, handler, self)
 
-    @staticmethod
-    def _assert_non_callable_event(event_name: str):
-        if isinstance(event_name, EventWithFilter):
-            raise TypeError("Argument event_name should not be a callable event, "
-                            "please use event without any event filtering")
+    # @staticmethod
```
vfdev-5: Justus, could you please remove commented code

ignite/engine/events.py
```diff
@@ -4,50 +4,68 @@
...
+            boolean to indicate whether this event should be executed. Defaults to None, which will result to a
+            function that always returns `True`
 
-    def __str__(self) -> str:
-        return "<%s event=%s, filter=%r>" % (self.__class__.__name__, self.event, self.filter)
+    """
 
+    def __init__(self, value: str, event_filter: Optional[Callable] = None):
+        self._filter = lambda _, __: True
+        self.filter = event_filter
```
vfdev-5: This line and above both look a bit strange. Especially, that `filter` then becomes a property...

ignite/engine/events.py
```diff
@@ -4,50 +4,68 @@
...
+            self.value = value
 
-    .. code-block:: python
+    @property
+    def filter(self):
+        return self._filter
 
-        from ignite.engine.engine import CallableEvents
+    @filter.setter
+    def filter(self, new_event_filter: Optional[Callable]):
```
vfdev-5: Why do we need the setter ?

ignite/engine/events.py
```diff
@@ -93,44 +113,50 @@ def wrapper(engine, event: int) -> bool:
...
+        return "<event=%s, filter=%r>" % (self.__class__.__name__, self.filter)
+
 
-class Events(CallableEvents, Enum):
+#
+class Events(CallableEventWithFilter, Enum):
     """Events that are fired by the :class:`~ignite.engine.Engine` during execution.
 
-    Since v0.3.0, Events become more flexible and allow to pass an event filter to the Engine:
+     Since v0.3.0, Events become more flexible and allow to pass an event filter to the Engine:
```
vfdev-5: The spaces added by your IDE seem to be incorrect, no ? Lines should be aligned with above `"""`

ignite/engine/events.py
```diff
@@ -4,50 +4,68 @@
 from enum import Enum
 import numbers
 import weakref
+from copy import deepcopy
```
vfdev-5: This is not used ?

ignite/engine/events.py
```diff
@@ -186,9 +212,7 @@ def __init__(self, **kwargs):
             if not hasattr(self, value):
                 setattr(self, value, 0)
 
-    def get_event_attrib_value(self, event_name: Union[EventWithFilter, CallableEvents, Enum]) -> int:
-        if isinstance(event_name, EventWithFilter):
```
vfdev-5: Here I do not understand why we do not need such checking anymore ? Probably, I do not completely understand your modifications...

ignite/engine/events.py
justusschock: Basically the property wraps the `self._filter` attribute, where the setter handles the None argument and makes it a function returning always `True`. This also answers your question below.

ignite/engine/events.py
justusschock: Answered above

ignite/engine/events.py
justusschock: Correct. Forgot to remove this from a previous version

ignite/engine/events.py
vfdev-5: Yes, I see now. But creation of setter just for `self.filter = event_filter` is a bit overkill :) 
My below question is about whether we need to reset event filter dynamically ? I think we should not

ignite/engine/events.py
justusschock: Because currently all events must be of this type. So a type check is redundant and was thus removed

ignite/engine/events.py
justusschock: Otherwise we need to check for `None` all the time, so I found this to be more elegant. But I agree, that we might not want to change it dynamically, so I will just hardcode the check where needed.

vfdev-5:
Closed in favor of #788

PR #788, opened 2020-02-18
Callable events

justusschock:
This Pull Request is a clean rework of #769 to prepare #710 

Description:
To overload the necessary functions, we need to ensure, that all events are of a certain type. This is what this PR aims for.

Check list:

cc @vfdev-5 @ykumards


ignite/engine/events.py
```diff
@@ -1,53 +1,71 @@
 
 from typing import Callable, Optional, Union, Any
 
-from enum import Enum
+from enum import Enum, EnumMeta
```
vfdev-5:
```suggestion
from enum import Enum
```

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -172,19 +168,16 @@ def attach(self, engine, metric_names=None, output_transform=None,
         """
         desc = self.tqdm_kwargs.get("desc", "Epoch")
 
-        if not isinstance(event_name, (Events, EventWithFilter)):
-            raise ValueError("Logging event should be only `ignite.engine.Events`")
-
-        if isinstance(closing_event_name, EventWithFilter):
```
vfdev-5: This condition is necessary for closing events. We should not close progress bar with some event filtering, IMO...

ignite/engine/events.py
```diff
@@ -1,53 +1,71 @@
...
-        self.event = event
-        self.filter = filter
+class CallableEventWithFilter:
+    """Single Event containing a filter, specifying whether the event should
+    be run at the current event (if the event type is correct)
 
-    def __str__(self) -> str:
-        return "<%s event=%s, filter=%r>" % (self.__class__.__name__, self.event, self.filter)
+    Args:
+        value (str): The value (will only be set if not yet set by other parent class). Necessary for usage with
```
vfdev-5: This is a bit difficult to understand what user should pass here... please, rephrase it 
like "internal usage, do not touch"  ;)

ignite/engine/events.py
```diff
@@ -231,7 +273,7 @@ def print_epoch(engine):
         # print_epoch handler is now unregistered
     """
 
-    def __init__(self, event_name: Union[EventWithFilter, CallableEvents, Enum], handler: Callable, engine):
+    def __init__(self, event_name: Union[CallableEventWithFilter, Enum], handler: Callable, engine):
```
vfdev-5: Currently, I would say `event_name: Any`. We need to decide about it and fix all typings according to it.

tests/ignite/contrib/handlers/test_tqdm_logger.py
```diff
@@ -365,8 +365,7 @@ def test_pbar_wrong_events_order():
     with pytest.raises(ValueError, match="should be called before closing event"):
         pbar.attach(engine, event_name=Events.ITERATION_COMPLETED, closing_event_name=Events.ITERATION_STARTED)
 
-    with pytest.raises(ValueError, match="Closing event should not use any event filter"):
-        pbar.attach(engine, event_name=Events.ITERATION_STARTED, closing_event_name=Events.EPOCH_COMPLETED(every=10))
+    pbar.attach(engine, event_name=Events.ITERATION_STARTED, closing_event_name=Events.EPOCH_COMPLETED(every=10))
```
vfdev-5: I'm not sure whether this works correctly...
As commented previously, we should not set closing events as filtered events.

ignite/contrib/handlers/tqdm_logger.py
justusschock: Okay, then we can use a check for the type of filter. I'll add this

justusschock:
Have we already decided whether we want arbitrary types as callbacks? cc @vfdev-5 @ykumards

vfdev-5:
> Have we already decided whether we want arbitrary types as callbacks? 

@justusschock you mean event type ? So I would say an event can be `CallableEventWithFilter`, `Enum` and in addition a `str`...

ignite/engine/engine.py
```diff
@@ -252,8 +247,9 @@ def print_epoch(engine):
             See :class:`~ignite.engine.Events` for more details.
 
         """
-        if isinstance(event_name, EventWithFilter):
-            event_name, event_filter = event_name.event, event_name.filter
+        if (isinstance(event_name,
+                       CallableEventWithFilter) and event_name.filter != CallableEventWithFilter.default_event_filter):
```
vfdev-5: @justusschock I can be wrong, but can't we just set `event_name.filter = None` for default non filtered behaviour ?
I do not see where `CallableEventWithFilter.default_event_filter` is actually used except comparision...

vfdev-5:
@justusschock could you please resolve the conflicts

ignite/engine/engine.py
justusschock: We can't, since we explicitly call the filter without additional typechecks.

vfdev-5:
Thanks @justusschock  !

PR #790, opened 2020-02-18
Stores training checkpoints and best model as archived files

vfdev-5:
Description:
Idea is to reduce storage space by default, when using `setup_common_training_handlers` and `save_best_model_by_val_score` methods from `ignite.contrib.engines.common`.

Check list:

cc @justusschock @ykumards


ykumards:
Hi @vfdev-5, just a quick question: is there any downside to saving the models as archived `.pth` files? If no, am wondering why this argument is set to false by default in `ModelCheckpoint`?

vfdev-5:
@ykumards it is mainly for somewhat "compatilibity" reasons that `archived=False`. Previously we saved everything as `.pth`, even there is no code to change, user can be surprised by seeing now `.pth.tar` instead, I think.

justusschock:
I think this is good for default settings.

Just a side note: We should maybe change the extension in general from `.pth` to `.pt` since `.pth` is also the extension used internally in python for [package configurations](https://docs.python.org/3.6/library/site.html). But probably this should be done in a separate PR (just a reminder).

vfdev-5:
> We should maybe change the extension in general from .pth to .pt since .pth is also the extension used internally in python for package configurations.

@justusschock yes, you are right! Let's create another PR and merge it for 0.4.0, not 0.3.X

> I think this is good for default settings.

`.pth` or `.pth.tar` ?

justusschock:
Sorry that it was not clear. I would prefer `.pth.tar` because we should aim for an obvious difference for different file types.

vfdev-5:
Checked with recent pytorch and it seems like it is useless to add `.tar` in the end of the file as it is not archived.
```python
import torch
from torchvision.models.resnet import resnet152
m = resnet152()
torch.save(m.state_dict(), "m.pt")
torch.save(m.state_dict(), "m.pth.tar")
torch.save(m.state_dict(), "m.pt.tar")
!ls -allh m.*
> 231M Mar 15 21:35 m.pt
> 231M Mar 15 21:35 m.pt.tar
> 231M Mar 15 21:35 m.pth.tar
```

justusschock:
That's true, but we could manually create an archive (with `tarfile` or `zipfile` or something similar)

vfdev-5:
> That's true, but we could manually create an archive (with tarfile or zipfile or something similar)

Well, loading such checkpoint wont be evident for users, I think.

I close the PR.

justusschock:
Starting with PyTorch 1.4 there is a new argument for that in `torch.save`: `_use_new_zipfile_serialization=False`. We could probably check for the Pytorch version and use this argument, when the version is recent enough, I guess. 
Although, this argument is not yet part of the official docs.

vfdev-5:
@justusschock yes, I saw this argument too, I checked with True/False but output size is still the same.

justusschock:
This is strange, since I had a look at the source code and it seems they are actually writing it to a zip file. But I don't know, how well you can compress a binary dump as it comes from pickle...

PR #793, opened 2020-02-19
Remove 'module' key for parallel checkpointing

### LINKED ISSUE #750: Checkpoint and ModelCheckpoint automatically saves base model for DDP wrapper
Idea is to avoid saving [DDP wrapped](https://pytorch.org/docs/stable/nn.html#torch.nn.parallel.DistributedDataParallel) model with `Checkpoint`/`ModelCheckpoint` classes and save instead its base model. Such that state dictionary wont contain prefix `module.` in the keys.

justusschock:
Fixes #750 

Description:
Removes the 'module' prefix for checkpointing of parallel models

Check list:

cc @vfdev-5

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -205,6 +205,11 @@ class Checkpoint:
 
             checkpoint = self._setup_checkpoint()
 
+            # happens in distributed setting when not in first process.
+            # Avoids that all processes read (not necessary due to parameter syncing)
+            if checkpoint is None:
+                return
+
             name = "checkpoint"
             if len(checkpoint) == 1:
                 for k in checkpoint:
@@ -225,6 +230,9 @@ class Checkpoint:
         checkpoint = {}
         for k, obj in self.to_save.items():
             if isinstance(obj, torch.nn.parallel.DistributedDataParallel):
+                # only save in first process
+                if torch.distributed.get_rank() > 0:
+                    return
                 torch.distributed.barrier()
                 state_dict = obj.module.state_dict()
             elif isinstance(obj, torch.nn.parallel.DataParallel):
@@ -258,10 +266,13 @@ class Checkpoint:
             if isinstance(obj, torch.nn.parallel.DataParallel):
                 obj.module.load_state_dict(state_dict)
             elif isinstance(obj, torch.nn.parallel.DistributedDataParallel):
+                # only load in first process
+                if torch.distributed.get_rank() > 0:
+                    return
                 torch.distributed.barrier()
                 obj.module.load_state_dict(state_dict)
                 obj.require_forward_param_sync = True
-                obj.require_backward_grad_sync = True
+                obj.require_backward_grad_sync = torch.is_grad_enabled()
                 torch.distributed.barrier()
             else:
                 obj.load_state_dict(state_dict)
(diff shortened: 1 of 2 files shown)


justusschock:
@vfdev-5 any idea, why this is causing a timeout?

ignite/handlers/checkpoint.py
```diff
@@ -205,6 +205,11 @@ def __call__(self, engine: Engine) -> None:
 
             checkpoint = self._setup_checkpoint()
 
+            # happens in distributed setting when not in first process.
+            # Avoids that all processes read (not necessary due to parameter syncing)
+            if checkpoint is None:
```
vfdev-5: Sorry, I do not get your comments above why to do this.

ignite/handlers/checkpoint.py
```diff
@@ -224,7 +229,17 @@ def __call__(self, engine: Engine) -> None:
     def _setup_checkpoint(self) -> dict:
         checkpoint = {}
         for k, obj in self.to_save.items():
-            checkpoint[k] = obj.state_dict()
+            if isinstance(obj, torch.nn.parallel.DistributedDataParallel):
+                # only save in first process
+                if torch.distributed.get_rank() > 0:
```
vfdev-5: This point is very interesting to handle. Previously, in the examples, we show that usage of `Checkpoint` is conditioned by `torch.distributed.get_rank() > 0`. But we can say that all our classes can handle that internally. Similar to what is done here. But maybe it is need in `__call__` method...
So, for instance, I think we do not need this here.

ignite/handlers/checkpoint.py
```diff
@@ -246,18 +261,34 @@ def load_objects(to_load: Mapping, checkpoint: Mapping) -> None:
         Checkpoint._check_objects(to_load, "load_state_dict")
         if not isinstance(checkpoint, collections.Mapping):
             raise TypeError("Argument checkpoint should be a dictionary, but given {}".format(type(checkpoint)))
+
+        def load_state_dict(obj, state_dict):
+            if isinstance(obj, torch.nn.parallel.DataParallel):
+                obj.module.load_state_dict(state_dict)
+            elif isinstance(obj, torch.nn.parallel.DistributedDataParallel):
+                # only load in first process
+                if torch.distributed.get_rank() > 0:
```
vfdev-5: Same for this

ignite/handlers/checkpoint.py
```diff
@@ -246,18 +261,34 @@ def load_objects(to_load: Mapping, checkpoint: Mapping) -> None:
...
+                    return
+                torch.distributed.barrier()
+                obj.module.load_state_dict(state_dict)
+                obj.require_forward_param_sync = True
```
vfdev-5: Why do we need these 3 lines ?

justusschock: They are needed, because the state_dict should only be loaded by one process. In opposite to DP in DDP there exist permanent replicas on all devices. These lines tell it to synchronize the loaded parameters before doing another forward to all replicas

ignite/handlers/checkpoint.py
justusschock: This is to exit the function for all the processes not performing the actual saving

PR #796, opened 2020-02-19
add visdom_logger to common module

ykumards:
Fixes # 

Description: Add `visdom_logger` to common module. This is an exact copy of `setup_tb_logging` modified for visdom API

Check list:


vfdev-5:
@ykumards testing Visdom is a bit tricky :) Maybe with mocking we can avoid to run the server.

ykumards:
> @ykumards testing Visdom is a bit tricky :) Maybe with mocking we can avoid to run the server.

Ah! My bad. Visdom server was running in my machine when I ran local tests

vfdev-5:
Maybe this can help : https://github.com/pytorch/ignite/blob/master/tests/ignite/contrib/handlers/test_visdom_logger.py#L12

PR #820, opened 2020-03-02
Bugfix & Enhancement `GpuInfo` and `CpuInfo` to threaded implementation

DrStoop:
As far as I understand, the current `GpuInfo` is a `Metric` attached to `Events.ITERATION_COMPLETED`, meaning it logs the current GPU utilization during its downtime when the model in not  updated or inferring as `Engine().process_function()` is completed:
```python
class Engine():
...
   def _run_once_on_dataset(..):
        ...
        self._fire_event(Events.ITERATION_STARTED)
        self.state.output = self._process_function(self, self.state.batch)
        self._fire_event(Events.ITERATION_COMPLETED)
        ...
```
Therefore the measurement is not representative for the actual GPU usage during training. Or did I miss anything?

An alternative quick-fix to the suggestion below for the current `GpuInfo` - at least for the memory logging - would be to replace the `self.nvsmi.DeviceQuery("memory.used")` by [`torch.cuda.max_memory_allocated()`](https://pytorch.org/docs/stable/cuda.html#torch.cuda.max_memory_allocated) which would return the maximum used GPU-memory for each iteration (don't forget to also rest the method after logging).

Description:

The suggested `GpuInfo` and `CpuInfo` are both run on independent threads logging the hardware every user-defined time interval. This should basically lead to a random logging of GPU/CPU uptime and downtime usage which represents the actual use much better.

### DIFF
--- a/ignite/contrib/metrics/cpu_info.py
+++ b/ignite/contrib/metrics/cpu_info.py
@@ -21,6 +21,7 @@ class CpuInfo(Thread):
         per_cpu (bool): log overall CPU-utilization (`False`) by default or each CPU separately (`True`)
         unit (['KB', 'MB', 'GB']): logging unit defaults to `'GB'`
     """
+
     def __init__(self, logger_directory, logger_name='CPULogger', log_interval_seconds=0.5, per_cpu=False, unit='GB'):
         super().__init__(name=logger_name, daemon=True)
         # CAUTION: Always avoid more than one `SummaryWriter` logging to the same directory
@@ -69,7 +70,7 @@ class CpuInfo(Thread):
         cpu_utilization = {}
         if self._per_cpu:
             for idx_cpu, cpu_percentage in enumerate(cpu_percentages):
-                        cpu_utilization['CPU{}'.format(idx_cpu)] = cpu_percentage
+                cpu_utilization['CPU{}'.format(idx_cpu)] = cpu_percentage
         else:
             cpu_utilization['overall'] = cpu_percentages
         # log CPU utilization to tensorboard
--- a/ignite/contrib/metrics/cpu_info.py
+++ b/ignite/contrib/metrics/cpu_info.py
@@ -1,0 +1,86 @@
+from threading import Thread
+from time import sleep, time
+from psutil import cpu_count, cpu_percent, virtual_memory
+from torch.utils.tensorboard import SummaryWriter
+
+
+class CpuInfo(Thread):
+    """
+    Logger for the CPU resources CPU and RAM utilization run on a separate thread.
+
+    `CpuInfo` is implemented on a separate thread as any attachment to an event would would effectively measure
+    the GPU/CPU-utilization of the downtime, as all events are not fired during the `Engine().process()` where the
+    GPU/CPU is in use. Triggering the logging independently will randomize the measurements during
+    _up times_ (when `Engine().process()` is running) and _down times_.
+
+    Args:
+        logger_directory: directory for tensorboard logs
+        logger_name: name of logger
+        log_interval_seconds: logging interval in seconds. This interval practically cannot be reduced below 0.1sec \
+        as this is the minimum CPU measurement time of `psutil.cpu_percent` (see docu).
+        per_cpu (bool): log overall CPU-utilization (`False`) by default or each CPU separately (`True`)
+        unit (['KB', 'MB', 'GB']): logging unit defaults to `'GB'`
+    """
+    def __init__(self, logger_directory, logger_name='CPULogger', log_interval_seconds=0.5, per_cpu=False, unit='GB'):
+        super().__init__(name=logger_name, daemon=True)
+        # CAUTION: Always avoid more than one `SummaryWriter` logging to the same directory
+        # because this will lead to log file losses
+        self.logger_directory = logger_directory
+        self._log_interval_seconds = log_interval_seconds
+        self._per_cpu = per_cpu
+        self.cpu_count = cpu_count()
+        self._tb_logger = SummaryWriter(logdir=self.logger_directory)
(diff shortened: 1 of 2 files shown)


vfdev-5:
@DrStoop AFAIK `torch.cuda.max_memory_allocated()` does not show the same thing as `nvidia-smi`, so using pynmvl is IMO better in the way if we need to match what we see with `nvidia-smi`.

Concerning GPU utilization, yes probably it would tricky to catch it as a metric. Maybe there can be an option to request a mean value or something...

vfdev-5:
> The suggested GpuInfo and CpuInfo are both run on independent threads logging the hardware every user-defined time interval. This should basically lead to a random logging of GPU/CPU uptime and downtime usage which represents the actual use much better.

Thanks for the PR, I think we need to compare both implementations to see the diffs.

Do we care about CPU usage ?

DrStoop:
> Do we care about CPU usage ?

Depends, most OSs have multiple live CPU-logger integrated anyway, that are good enough for debugging.

Generally CPU become a bottleneck e.g. for data pre-processing. If you're working in a pipeline with "live'-pre-processing, this may become the relevant bottleneck...

If you're running batchwise pre-processing of data, this would probably be a necessary feature... but therefore one would have to first thing about a `DataPreprossorEngine` ;-)

Nevertheless, the `*Info` runs the thread independently of the rest of the Engine, so you could starte it even when only pre-processing (without a "live-pipe"), just needed to get the loop between the `GpuInfo`-timestamp & the engine iterations.

(Note: In the framework, it defaults to `n_samples_ref` of the first engine started in the `state`.)

vfdev-5:
>  Generally CPU become a bottleneck e.g. for data pre-processing or data handling. If you're working in a pipeline with "live'-pre-processing, this may become the relevant bottleneck, or shoveling data around in a (unwanted/unrecognized) inefficient way... The combination of both GPU & CPU can be helpful to understand why the GPU is not used to capacity while the CPU is running single-cored at 101%.

Yes, correct. Also I/O can be a bottleneck, such that GPU/CPU activities can be both ~0.

> Nevertheless, the *Info runs the thread independently of the rest of the Engine, so you could start it even when only pre-processing (without a "live-pipe"), just needed to get the loop between the GpuInfo-timestamp & the engine iterations.

I see your point. Let me take a look at your code and I'll comment out...

ignite/contrib/metrics/gpu_info.py
```diff
@@ -1,100 +1,91 @@
...
-            # Logging with TQDM
-            ProgressBar(persist=True).attach(trainer, metric_names=['gpu:0 mem(%)', 'gpu:0 util(%)'])
-            # Progress bar will looks like
-            # Epoch [2/10]: [12/24]  50%|█████      , gpu:0 mem(%)=79, gpu:0 util(%)=59 [00:17<1:23]
-
-            # Logging with Tensorboard
-            tb_logger.attach(trainer,
-                             log_handler=OutputHandler(tag="training", metric_names='all'),
-                             event_name=Events.ITERATION_COMPLETED)
+class GpuPynvmlLogger(Thread):
```
DrStoop: Already discovered the first bugs... the pytest wasn't running for other reasons:
`GpuPynvmlLogger` -> `GpuInfo`

ignite/contrib/metrics/gpu_info.py
```diff
@@ -1,100 +1,91 @@
...
-                    "GPU utilization information does not provide 'gpu_util' information in " "{}".format(util_report)
-                )
-                continue
-
-            engine.state.metrics[util_name] = int(util_report["gpu_util"])
-
-    def attach(self, engine, name="gpu", event_name=Events.ITERATION_COMPLETED):
-        engine.add_event_handler(event_name, self.completed, name)
+    def __init__(self, logger_directory, logger_name='GPULogger', log_interval_seconds=1, unit='GB'):
+        super(GpuPynvmlLogger, self).__init__(name=logger_name, daemon=True)
```
DrStoop: ` super(GpuPynvmlLogger, self)` -> `super()`

ignite/contrib/metrics/gpu_info.py
```diff
@@ -1,100 +1,91 @@
...
+                                    tag_scalar_dict=gpu_utilizations,
+                                    global_step=time() - self._start_time)
+
+    def close(self):
+        # Quit while-loop in `self.run()`
+        self._log_gpu = False
+        # Close tensorboard logger
+        self._tb_logger.close()
+        # Join thread
+        self.join()
```
DrStoop: Forgot the `attach` method, e.g.:
```python
    def attach(self, engine, name="gpu", event_name_started=Events.STARTED, event_name_completed=Events.COMPLETED):
        engine.add_event_handler(event_name_started, self.start, name)
        engine.add_event_handler(event_name_completed, self.close,name)
```

DrStoop (commented):
Sorry, first bugs came to my mind... the pytest wasn't running for other reasons.

ignite/contrib/metrics/cpu_info.py
```diff
@@ -0,0 +1,87 @@
...
+                                    tag_scalar_dict=cpu_utilization,
+                                    global_step=time() - self._start_time)
+
+    def close(self):
+        # Quit while-loop in `self.run()`
+        self._log_cpu = False
+        # Close tensorboard logger
+        self._tb_logger.close()
+        # Join thread
+        self.join()
```
DrStoop: Forgot the `attach` method, e.g.:
```python
    def attach(self, engine, name="gpu", event_name_started=Events.STARTED, event_name_completed=Events.COMPLETED):
        engine.add_event_handler(event_name_started, self.start, name)
        engine.add_event_handler(event_name_completed, self.close,name)
```

DrStoop (commented):
Same issue as with gpu_info.py...

PR #824, opened 2020-03-03
Bugfix #823 trainer common setup and transform output with numbers

### LINKED ISSUE #823: argument `metrics` in method `create_supervised_trainer` ?
## ❓ Questions/Help/Support

To define evaluators using helper `create_supervised_evaluator`, metrics can be added for attachment to engine :
```python
  metrics = {
    'accuracy': Accuracy(),
    'nll': Loss(loss_fn=loss_fn),
  }

  evaluator = create_supervised_evaluator(model=model, metrics=metrics)
```
Using helper `create_supervised_trainer`, metrics have to be added by hand :
```python
  trainer = create_supervised_trainer(model=model, optimizer=optimizer, loss_fn=loss_fn)

  RunningAverage(output_transform=lambda x: x).attach(trainer, 'loss')
```
What do you think about add argument `metrics` in method `create_supervised_trainer` ?

sdesrozis:
Fixes #823 

Description:

`create_supervised_trainer` with default `output_transform` is not compatible with `setup_common_training_handlers` and `output_names` :
```python
trainer = create_supervised_trainer(model=model,
                                    optimizer=optimizer,
                                    loss_fn=loss_fn)  

setup_common_training_handlers(trainer=trainer, output_names=['loss'])
```
is ko because default `output_transform` :
```python
output_transform: Callable = lambda x, y, y_pred, loss: loss.item()
```
Fix using `numbers` in `common.py` line 128
```python
def output_transform(x, index, name):
  if isinstance(x, Mapping):
    return x[name]
  elif isinstance(x, Sequence):
    return x[index]
  elif isinstance(x, (torch.Tensor, numbers.Number)):
    return x
  else:
    raise ValueError(
      "Unhandled type of update_function's output. "
      "It should either mapping or sequence, but given {}".format(type(x))
    )
``` 
NB: Some files were modfied due to black.

Check list:

### DIFF
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -1,5 +1,6 @@
 from functools import partial
 import warnings
+import numbers
 
 from collections.abc import Sequence, Mapping
 
@@ -129,7 +130,7 @@ def _setup_common_training_handlers(
                 return x[name]
             elif isinstance(x, Sequence):
                 return x[index]
-            elif isinstance(x, torch.Tensor):
+            elif isinstance(x, (torch.Tensor, numbers.Number)):
                 return x
             else:
                 raise ValueError(
--- a/ignite/contrib/handlers/custom_events.py
+++ b/ignite/contrib/handlers/custom_events.py
@@ -72,11 +72,11 @@ class CustomPeriodicEvent:
 
         self.custom_state_attr = "{}_{}".format(prefix, self.period)
         event_name = "{}_{}".format(prefix.upper(), self.period)
-        setattr(self, "Events",
-                EventEnum("Events",
-                          " ".join(["{}_STARTED".format(event_name),
-                                    "{}_COMPLETED".format(event_name)])
-                          ))
+        setattr(
+            self,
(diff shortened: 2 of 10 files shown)


vfdev-5:
We need to update the docs such that contrib.engines.common appears in the docs. @sdesrozis could you please send a PR wth this if you want to contribute more :) Otherwise no worries, we can handle it

sdesrozis:
> We need to update the docs such that contrib.engines.common appears in the docs. @sdesrozis could you please send a PR wth this if you want to contribute more :) Otherwise no worries, we can handle it

It would be a pleasure to contribute :) I just need some time to do it. Tho, I warn you, my english is not very good :(

Is there anything to know about writing the doc (more than what is noticed in the contribution guidelines) ?

vfdev-5:
Well idea is to add the missing module to https://github.com/pytorch/ignite/tree/master/docs/source/contrib
like this one : https://github.com/pytorch/ignite/blob/master/docs/source/contrib/engines.rst

Probably, nothing to write in english :)

PR #826, opened 2020-03-03
Add doc for module contrib.engines.common

sdesrozis:
Description:

Add doc for module `contrib.engines.common`. Is `helpers` a better name for this module ?  

Check list:


vfdev-5:
`helpers`, `common`, `tools`, `utils` they all IMO sound similarly. Let's keep the name.

docs/source/index.rst
```diff
@@ -86,6 +86,7 @@ of stable version as dependency):
    :caption: Contrib Package Reference
     
    contrib/engines
+   contrib/common
```
vfdev-5: contrib/engines/common

PR #834, opened 2020-03-09
Issue 802

### LINKED ISSUE #802: Accuracy computation with variable classes
## ❓ Questions/Help/Support

I've been using the `Accuracy` class (in `ignite.metrics`) successfully so far in a multi-class setting to compute test set accuracy.

Recently, I've hit a new scenario/dataset where the test set has different number of classes for each example. You can think of this as a multiple-choice selection task, where each example has a different number of candidates.

Now in this scenario, when I used `Accuracy`, I got the below error:
`ERROR:ignite.engine.engine.Engine:Current run is terminating due to exception: Input data number of classes has changed from 13 to 81.`

I dived into the [code](https://github.com/pytorch/ignite/blob/master/ignite/metrics/accuracy.py#L83L92) and I see that the base class for `Accuracy`, i.e., `_BaseClassification ` assumes that the number of classes/candidates is fixed across test set examples. And this is why I'm getting the above error at run-time.

However, for multi-class accuracy, as you can see in the `update()` method [here](https://github.com/pytorch/ignite/blob/master/ignite/metrics/accuracy.py#L154L156), we are simply computing the number of correct predictions via `argmax` along dimension-1 and measuring matches against `y`.

This computation of `correct` should be accurate even if the number of candidates changes across two examples from the test set, right?

So in effect, the computed accuracy would be correct even in such a case because the underlying variables are computed [correctly](https://github.com/pytorch/ignite/blob/master/ignite/metrics/accuracy.py#L165L166)?

What do you think, @vfdev-5? Is it possible to have a version of `Accuracy` for multi-class where there is no expectation that the number of classes is the same across all examples in the test set?

ykumards:
Fixes #802 

Description: Add flag to `Accuracy` to handle computation with varying number of classes in multi-class setting

Check list:

### DIFF
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -14,11 +14,11 @@ class _BaseClassification(Metric):
         self,
         output_transform: Callable = lambda x: x,
         is_multilabel: bool = False,
-        is_variable_multiclass: bool = False,
+        variable_classes: bool = False,
         device: Optional[Union[str, torch.device]] = None,
     ):
         self._is_multilabel = is_multilabel
-        self._is_variable_multiclass = is_variable_multiclass
+        self._variable_classes = variable_classes
         self._type = None
         self._num_classes = None
         super(_BaseClassification, self).__init__(output_transform=output_transform, device=device)
@@ -74,6 +74,8 @@ class _BaseClassification(Metric):
             if self._is_multilabel:
                 update_type = "multilabel"
                 num_classes = y_pred.shape[1]
+                if self._variable_classes:
+                    raise RuntimeError("variable_classes flag cannot be set to True for multi-label type")
             else:
                 update_type = "binary"
                 num_classes = 1
@@ -89,7 +91,7 @@ class _BaseClassification(Metric):
             if self._type != update_type:
                 raise RuntimeError("Input data type has changed from {} to {}.".format(self._type, update_type))
             if self._num_classes != num_classes:
-                if not self._is_variable_multiclass:
+                if not self._variable_classes:
                     raise ValueError(
                         "Input data number of classes has changed from {} to {}".format(self._num_classes, num_classes)
                     )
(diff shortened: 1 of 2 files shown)


ykumards:
Hi @vfdev-5 does this look like what we'd discussed about #802? Also, am not very confident about the name of the flag I've used here, do you have a better suggestion?

ignite/metrics/accuracy.py
```diff
@@ -14,9 +14,11 @@ def __init__(
         self,
         output_transform: Callable = lambda x: x,
         is_multilabel: bool = False,
+        is_variable_multiclass: bool = False,
```
justusschock: can't we rename this to something like `variable_classes`? We already have `is_multilabel` here and together they should be the same,  but maybe we can reuse `variable_classes` in some further way.

So `is_variable_multiclass` would be the same as `is_multilabel and variable_classes`. This won't change the functionality though but we don't have separate variables indicating the same thing (partially)

vfdev-5: `variable_classes` sounds good. However, we need emphasize that this is possible only for multiclass case and not multi-label.

ignite/metrics/accuracy.py
```diff
@@ -14,9 +14,11 @@ def __init__(
...
         device: Optional[Union[str, torch.device]] = None,
     ):
         self._is_multilabel = is_multilabel
+        self._is_variable_multiclass = is_variable_multiclass
```
vfdev-5: We need to check that this flag is False in case of multi-label

ignite/metrics/accuracy.py
ykumards: sounds good. I pushed the changes

ignite/metrics/accuracy.py
ykumards: done!

sdesrozis:
@ykumards @vfdev-5

Does this PR need more work or could we merge / close it?

vfdev-5:
@sdesrozis we wanted to refactor these classes in case of multi-label. So this code is a temporary solution to fix the issue.

vfdev-5:
Well, let's merge and we refactor thing later.

sdesrozis:
It seems that the branch of @ykumards is not longer reachable.

Moh-Yakoub:
@sdesrozis  @vfdev-5 It seems that this PR went inactive for a while, if we are still interested in that change I can prepare a similar PR on a more fresh master branch.

sdesrozis:
I let @vfdev-5 choose what to do. Maybe @g-karthic would tell us if this PR is still relevant.

vfdev-5:
@Moh-Yakoub @sdesrozis let's close this PR as it would only represent a temporary solution. Let's discuss more in the related issue.

PR #835, opened 2020-03-10
Create supverised fix

### LINKED ISSUE #830: model is moved too early in create_supervised_trainer/evaluator
Hi, first of all, thanks for the great library!
I have a general "bug" report or question on the specific implementation of `create_supervised_trainer` and `create_supervised_evaluator`.
As seen for example here:
https://github.com/pytorch/ignite/blob/master/ignite/engine/__init__.py#L64

The moment a supervised evaluator/trainer is created the model is moved to the specific device and 
*not* during the inference stage.

I agree, that this version doesn't have an impact in general, but I would argue that the model should be moved in the `_inference` function and not when I create the evaluator/trainer.
A weird example but similar to my case:
I have multiple evaluators and I would like to run one on the CPU and one on the GPU.
The reason being that a different library doesn't support GPUs yet and I would like to add it in an "ignite" style to keep the code similar.
Now, the run steps will not work, as the model is *not* moved again in the inference stage.

I can imagine that there is a reason why it is the way it is, but I am wondering if a future version couldn't do the `model.to(device)` step in the `_inference` function. :)

Thanks!

kai-tub:
Fixes #830 

Description:
As discussed, the model was moved too early in the `create_supervised` functions, which could lead to unexpected behavior in uncommon cases.
In my first commit I've added two similar test cases to each function.
The first test would be fixed by @vfdev-5's *lazy* solution.
I've implemented the fix @justusschock and I proposed, where the model will always be moved at each iteration. This gives the user more flexibility and is, in my opinion, closer to the documentation,
as the batches and labels are also processed at each iteration.
The `.to()` operation becomes a No-Op if the model is already on the correct device.

Check list:
* [Y] New tests are added (if a new feature is added)
* [N] New doc strings: description and/or example code are in RST format
* [N] Documentation is updated (if required)


vfdev-5:
Thanks for the PR @kai-tub !

Concerning No-Op, I'm not that sure, for example, consider the following profiling:
```python
import torch
from torchvision.models.resnet import resnet101

device = "cuda"
model = resnet101(pretrained=False).to(device)

%%timeit
for _ in range(1000):
    model.to(device)
    s = torch.sum(torch.rand(1000))

> 2.02 s ± 28.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
for _ in range(1000):
    s = torch.sum(torch.rand(1000))

> 9.9 ms ± 295 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

I think it would be more safe to set it inside a handler as suggested in the issue.

kai-tub:
I am a bit stunned.
I know that it is not the common case for the model to be moved over and over again, but I really thought they would've implemented an early exist. 
But it makes a bit of sense, as there is no easy way to check where the model lies.
On the other hand, I forgot that there are models that are split over multiple GPUs, so maybe it would even be *impossible* to add this feature...

Oh well, I will then delete the tests that move the model to the CPU in between and change it according to your solution. But I am going to change the documentation, as I still feel like it reads as if the model should be moved every time the data is moved. :)

vfdev-5:
Another point, probably also to see, however it is not related to initial issue.
What would happen if user gives a `model = DataParallel(model)` and sets a device to put batches on device. Probably, we need to check this cases and update the docs.

justusschock:
LGTM

justusschock:
> Another point, probably also to see, however it is not related to initial issue.
> What would happen if user gives a `model = DataParallel(model)` and sets a device to put batches on device. Probably, we need to check this cases and update the docs.

Maybe the easiest way to take this into account is to do this only, when the model is not a DP or DDP model

kai-tub:
> Another point, probably also to see, however it is not related to initial issue.
> What would happen if user gives a `model = DataParallel(model)` and sets a device to put batches on device. Probably, we need to check this cases and update the docs.

Sorry, but I have no experience with DP and DDP and have no access to multiple GPUs (usually not even to a single GPU). But what I understand from:
https://pytorch.org/docs/stable/nn.html#torch.nn.DataParallel
is that if the model is wrapped, the device id shouldn't be used to move the model.
But why should the user set `device` for the batches at all?

The example code from the documentation writes:
> input_var can be on any device, including CPU

Is setting a `device` here a wrong use from the user's side?

vfdev-5:
@kai-tub thanks for PR !

PR #839, opened 2020-03-15
FashionMNIST notebook, EarlyStopping correction

xen0f0n:
In the example there was a single evaluator and the EarlyStopping handler was attached to it. Using a single evaluator the handler takes into account the loss for both validation and training sets and updates its best_score and counter based on both losses. Since we want the score function to rely only on the val_loss we have to attach the handler to an evaluator that loops only over the validation set.


vfdev-5:
@xen0f0n thanks for the PR, yes, you are correct about that.

vfdev-5:
@xen0f0n could you please run it to make sure that it works. Thanks

xen0f0n:
@vfdev-5, I've ran it and it works.

vfdev-5:
@xen0f0n thanks for the PR ! Eventually, we can now rerun it in colab too :)

PR #847, opened 2020-03-22
Save only one checkpoint per priority

cqql:
Fixes `FileNotFoundError` during checkpointing

Description: The previous behavior would save two `Item`s for the same priority if `n_saved >= 2` but would store them in the same path. If you then store more checkpoints with higher priorities such that the first two of same priority drop out, the second one dropping out will raise a `FileNotFoundError` because its file has already been removed by the first `Item`.

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -217,6 +217,9 @@ class Checkpoint:
                 checkpoint = checkpoint[name]
             filename = "{}{}_{}{}".format(self._fname_prefix, name, suffix, self._ext)
 
+            if any(item.filename == filename for item in self._saved):
+                return
+
             self.save_handler(checkpoint, filename)
 
             self._saved.append(Checkpoint.Item(priority, filename))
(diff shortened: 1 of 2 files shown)


vfdev-5:
@cqql thanks for the PR ! I understand the issue, however I think we need to tackle it a bit differently. Instead of checking already existing priority, I would check the filename (as it the filename which is duplicated):
```python
filename = "{}{}_{}{}".format(self._fname_prefix, name, suffix, self._ext)
# check if filename exists in self._saved and return if exist
# otherwise go on
self.save_handler(checkpoint, filename)
```
IMO, checking file is better as we can have the same priority (as validation score) but if user provides also a `global_step_transform` such that filenames wont collide.
What do you think ?

cqql:
I think you are right. I did not think of any other sources "uniquifying" the name. Is the updated patch fine/better?

vfdev-5:
@cqql yes, it looks good

vfdev-5:
@cqql thanks again for the PR !

PR #850, opened 2020-03-26
Make torch.argmax optional : callable from logits to labels

### LINKED ISSUE #822: Make optional argmax for y_pred in Confusion Matrix
## 🚀 Feature

Today, the conditions on the input of the CM are the following:
```
    - `y_pred` must contain logits and has the following shape (batch_size, num_categories, ...)
    - `y` should have the following shape (batch_size, ...) and contains ground-truth class indices
        with or without the background class. During the computation, argmax of `y_pred` is taken to determine predicted classes.
```

Taking argmax on `y_pred` can be an option if we would like to determine winning class by some other rule. Let's keep argmax as default behaviour if `y_pred` is `(N, C, ...)` and do not apply it if `y_pred.shape == y.shape` and `(N, ...)`.

sdesrozis:
Fixes #822 

Description:

Make torch.argmax optional. I don't use output_transform but introduce a new callable dedicated to transform from logits to labels. Notice that to do binary, it's preferable to use this new feature rather than output_transform. (output_transform should be used to transform outputs of models to a conform input for metrics, i.e. [y_pred, y] in logits).
 
For review, tests neither docs are not yet done. 

Check list:

### DIFF
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -13,12 +13,14 @@ class _BaseClassification(Metric):
     def __init__(
         self,
         output_transform: Callable = lambda x: x,
+        logit_to_label_fn: Callable = lambda x: torch.argmax(x, dim=1),
         is_multilabel: bool = False,
         device: Optional[Union[str, torch.device]] = None,
     ):
         self._is_multilabel = is_multilabel
         self._type = None
         self._num_classes = None
+        self._logit_to_label_fn = logit_to_label_fn
         super(_BaseClassification, self).__init__(output_transform=output_transform, device=device)
 
     def reset(self) -> None:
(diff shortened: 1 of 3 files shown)


ignite/metrics/accuracy.py
```diff
@@ -130,12 +132,14 @@ def thresholded_output_transform(output):
     def __init__(
         self,
         output_transform: Callable = lambda x: x,
+        logits_to_label_fn: Callable = lambda x: torch.argmax(x, dim=1),
```
justusschock: Can't we use predefined functions here for each of the cases?

ignite/metrics/accuracy.py
```diff
@@ -152,7 +156,8 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
         if self._type == "binary":
             correct = torch.eq(y_pred.view(-1).to(y), y.view(-1))
         elif self._type == "multiclass":
-            indices = torch.argmax(y_pred, dim=1)
+            # indices = torch.argmax(y_pred, dim=1)
+            indices = self._logits_to_label_fn(y_pred)
```
justusschock: I would rather have such a function for all the cases (even if this is just `lambda x: x`) and apply them before the case switching to reduce the difference in behaviour among the cases

sdesrozis: IMHO, this is not a good idea. This reflects a design concern. I think that implementation of metrics should be splitted : (binary and multiclass) vs multilabel.
* binary is an specific case of multiclass. So we could apply the function in both cases.
* multilabel is different and should have a specific class.
Althought, a front-end class for user should be provided.

ignite/metrics/accuracy.py
sdesrozis: I don't understand what cases you're talking about.

justusschock: below there is a case switching for different cases (`self._type`) and for each of the cases I would define a logins_to_label_fn during init instead of having a general interface (kind of same discussion as below)

sdesrozis: Ok I see. Cases "binary" and "multilabel" don't need logits_to_label_fn call. I understand the idea to have an identity function for these cases and `apply logits_to_label_fn` in all cases. But I am skeptical about it. The problem that you point out sounds to me design concerns.

PR #858, opened 2020-03-29
fix issue 831 : tensorboardx no longer mandatory if pytorch>=1.2

### LINKED ISSUE #831: Update mnist example
## 🚀 Feature

Please update the mnist [example](https://github.com/pytorch/ignite/tree/master/examples/mnist) `mnist_with_tensorboardx.py`. Rather than depend on an external library (tensorboardX), it would be nice to have an example that utilizes the [built-in](https://pytorch.org/docs/stable/tensorboard.html) version of PyTorch.

sdesrozis:
Fixes #831 

Description:

Tensorboardx no longer mandatory if pytorch>=1.2

TODO Use case file should be renamed `mnist_with_tensorboard.py` ? Should impact CI/CD.

Check list:

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -64,8 +64,8 @@ jobs:
         sleep 10
         python examples/mnist/mnist_with_visdom.py --epochs=1
         kill %1
-        # 3) mnist_with_tensorboardx.py
-        python examples/mnist/mnist_with_tensorboardx.py --epochs=1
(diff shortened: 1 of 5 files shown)


vfdev-5:
@sdesrozis Looks good !

> TODO Use case file should be renamed mnist_with_tensorboard.py ? Should impact CI/CD.

Let's rename it. I think if you replace all the occurencies it should be fine.

PR #866, opened 2020-03-31
Fixes #795

### LINKED ISSUE #795: Dataflow randomness problem with infinite data iterators
## 🐛 Bug description

```python
import torch

def args_printing_wrapper(fn, prefix=""):    
    def wrapper(*args, **kwargs):
        print(prefix, *args, **kwargs)
        fn(*args, **kwargs)
    return wrapper

def random_train_data_generator():
    while True:
        yield torch.randint(0, 100, size=(4, ))

def random_val_data_generator():
    while True:
        yield torch.randint(0, 100, size=(4, )) + 100

def print_train_data(engine, batch):
    i = engine.state.iteration
    e = engine.state.epoch
    print("train", e, i, batch)

def print_val_data(engine, batch):
    i = engine.state.iteration
    print("val", i, batch)
    
trainer = Engine(print_train_data)
evaluator = Engine(print_val_data)

trainer._manual_seed = args_printing_wrapper(trainer._manual_seed, prefix="trainer")
evaluator._manual_seed = args_printing_wrapper(evaluator._manual_seed, prefix="evaluator")

def run_evaluation(_):
    evaluator.run(random_val_data_generator(), epoch_length=4)

trainer.run(random_train_data_generator(), max_epochs=4, epoch_length=6);
```
gives 
```
trainer 12 0
train 1 1 tensor([63,  3, 22, 77])
train 1 2 tensor([93, 86, 19, 83])
train 1 3 tensor([88, 20, 74, 93])
train 1 4 tensor([52, 25, 49, 41])
train 1 5 tensor([82, 87, 58, 18])
train 1 6 tensor([72,  9, 32, 64])
evaluator 12 0
val 1 tensor([163, 103, 122, 177])
val 2 tensor([193, 186, 119, 183])
val 3 tensor([188, 120, 174, 193])
val 4 tensor([152, 125, 149, 141])
train 2 7 tensor([82, 87, 58, 18])
train 2 8 tensor([72,  9, 32, 64])
train 2 9 tensor([77, 98, 57, 51])
train 2 10 tensor([58, 76, 39, 45])
train 2 11 tensor([82, 39, 27, 60])
train 2 12 tensor([58, 90,  2, 92])
evaluator 12 0
val 1 tensor([163, 103, 122, 177])
val 2 tensor([193, 186, 119, 183])
val 3 tensor([188, 120, 174, 193])
val 4 tensor([152, 125, 149, 141])
train 3 13 tensor([82, 87, 58, 18])
train 3 14 tensor([72,  9, 32, 64])
train 3 15 tensor([77, 98, 57, 51])
train 3 16 tensor([58, 76, 39, 45])
train 3 17 tensor([82, 39, 27, 60])
train 3 18 tensor([58, 90,  2, 92])
evaluator 12 0
val 1 tensor([163, 103, 122, 177])
val 2 tensor([193, 186, 119, 183])
val 3 tensor([188, 120, 174, 193])
val 4 tensor([152, 125, 149, 141])
train 4 19 tensor([82, 87, 58, 18])
train 4 20 tensor([72,  9, 32, 64])
train 4 21 tensor([77, 98, 57, 51])
train 4 22 tensor([58, 76, 39, 45])
train 4 23 tensor([82, 39, 27, 60])
train 4 24 tensor([58, 90,  2, 92])
evaluator 12 0
val 1 tensor([163, 103, 122, 177])
val 2 tensor([193, 186, 119, 183])
val 3 tensor([188, 120, 174, 193])
val 4 tensor([152, 125, 149, 141])
```
The problem is that `_manual_seed` is not called in this case and randomness of the data is configured by `evaluator`. Thus we have the same data all other epochs.

## Environment

 - Ignite Version (e.g., 0.3.0): 0.4.0.dev20200206
 - OS (e.g., Linux): Linux
 - Python version: 3.7

vfdev-5:
Fixes #795 

Description:
- added flag `deterministic` to `Engine.run`

Check list:
* [x] Documentation is updated (TODO)

* [x] Fix coverage : https://codecov.io/gh/pytorch/ignite/compare/44c2af1c9e9b0035f5fffb6458a967a3e0ad9359...52376ca0da029f25fdfbf533a5fc1b78e5334e50/src/ignite/engine/engine.py?before=ignite/engine/engine.py 

@y0ast, @sdesrozis, please take a look.

### DIFF
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -290,8 +290,151 @@ batch, this is how the user can use `output_transform` to get y_pred and y from 
           engine.state.new_attribute = 12345
 
 
-Resume training
----------------
+Deterministic training
+----------------------
+
+In general, it is rather difficult task to achieve deterministic and reproducible trainings as it relies on multiple
+aspects, e.g. data version, code version, software environment, hardware etc. According to `PyTorch documentation <https://pytorch.org/docs/stable/notes/randomness.html>`_:
+there are some steps to take in order to make computations deterministic on your specific problem on one specific
+platform and PyTorch release:
+
+- setup random state seed
+
+- set `cudnn to deterministic <https://pytorch.org/docs/stable/notes/randomness.html#cudnn>`_ if applicable
+
+In addition to that, Ignite's engine provides an option to control the dataflow by synchronizing random state on epoch.
+If data provider is `torch.utils.data.DataLoader`, batch data indices can be made completely deterministic. Thus,
+dataflow is _almost_ completely controlled. If model's computations are deterministic, then trainings can be reproducible.
+Moreover, resuming the training from a checkpoint can _almost_ repeat original training's behaviour.
+
+Random state synchronization
+````````````````````````````
+
+Engine's method :meth:`~ignite.engine.Engine.run` accepts `deterministic` as argument which enables or disables
+random state synchronization on every data provider's restart which corresponds in most of the cases to the epoch size.
+In this way, for a given iteration/epoch the dataflow can be the same for a given seed. More precisely it is something
+like
+
+.. code-block:: python
+
+    for e in range(num_epochs):
+        set_seed(seed + e)
+        do_single_epoch_iterations(dataloader)
+
+
+If input `dataloader` is `torch.utils.data.DataLoader`, its batch sampler is replaced by a batch sampler
+(:class:`~ignite.engine.engine.ReproducibleBatchSampler`) such that random sampling indices are reproducible
+by prefetching them before data iteration.
+
+Here is a trivial example of usage:
+
+.. code-block:: python
+
+    import torch
+    from torch.utils.data import DataLoader
+    from ignite.engine import Engine, Events
+
+
+    def random_train_data_loader(size):
+        data = torch.arange(0, size)
+        return DataLoader(data, batch_size=4, shuffle=True)
+
+
+    def print_train_data(engine, batch):
+        i = engine.state.iteration
+        e = engine.state.epoch
+        print("train", e, i, batch.tolist())
+
+
+    trainer = Engine(print_train_data)
+
+    print("Original Run")
+    trainer.run(random_train_data_loader(40), max_epochs=2, epoch_length=5, seed=7, deterministic=True)
+
+    print("Resumed Run")
+    trainer.load_state_dict({"epoch": 1, "seed": 7, "epoch_length": 5, "max_epochs": 2})
+    trainer.run(random_train_data_loader(40))
+
+.. code-block:: text
+
+    Original Run
+    train 1 1 [1, 18, 9, 28]
+    train 1 2 [39, 34, 30, 2]
+    train 1 3 [22, 7, 0, 12]
+    train 1 4 [14, 17, 38, 37]
+    train 1 5 [31, 5, 11, 29]
+    train 2 6 [35, 25, 10, 15]
+    train 2 7 [20, 32, 36, 24]
+    train 2 8 [16, 8, 26, 21]
+    train 2 9 [23, 3, 13, 33]
+    train 2 10 [27, 4, 6, 19]
+    Resumed Run
+    train 2 6 [35, 25, 10, 15]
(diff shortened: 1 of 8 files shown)


sdesrozis:
I have to check the Issue, sorry !

sdesrozis:
You did a great job, well done for documentation.

I wonder if `deterministic` should be a global parameter and not an argument for run. Thus, maybe the decorator @keep_random_state should be applied automatically ?

vfdev-5:
@sdesrozis thanks for the review and the comments !

> I wonder if deterministic should be a global parameter and not an argument for run. 

Good point ! I can't think of something infeasible if `deterministic` were a flag of Engine itself, instead of `run`. But I would keep it there as potentially an engine is reusable on multiple data providers... and maybe we would like to have different kind of runs ?

> Thus, maybe the decorator @keep_random_state should be applied automatically ?

I think for the loggers attached on iterations this could bring unwanted overhead...

sdesrozis:
> Good point ! I can't think of something infeasible if `deterministic` were a flag of Engine itself, instead of `run`. But I would keep it there as potentially an engine is reusable on multiple data providers... and maybe we would like to have different kind of runs ?

Ok but `deterministic` is related to a whole run, isn’t it ? IMO, it seems more global than `Engine` or `Engine.run()`. When `deterministic` is enabled, the user should have to point out what is not deterministic. Maybe I’m wrong.

> I think for the loggers attached on iterations this could bring unwanted overhead...

It should be measured. But the simplicity for user could be a very good point.

```python
ignite.deterministic = True
```

vfdev-5:
> > Good point ! I can't think of something infeasible if `deterministic` were a flag of Engine itself, instead of `run`. But I would keep it there as potentially an engine is reusable on multiple data providers... and maybe we would like to have different kind of runs ?
> 
> Ok but `deterministic` is related to a whole run, isn’t it ? IMO, it seems more global than `Engine` or `Engine.run()`. When `deterministic` is enabled, the user should have to point out what is not deterministic. Maybe I’m wrong.
> 
> > I think for the loggers attached on iterations this could bring unwanted overhead...
> 
> It should be measured. But the simplicity for user could be a very good point.
> 
> ```python
> ignite.deterministic = True
> ```

Yes, let me check that

vfdev-5:
> >    I think for the loggers attached on iterations this could bring unwanted overhead...

> It should be measured. But the simplicity for user could be a very good point.

In my tests, for a single handler called 500 times, the overhead is 0.3 seconds:
```python
import time
import torch

num_epochs = 5
epoch_length = 100

def update():
    time.sleep(0.05)
    
@keep_random_state
def decorated_update():
    update()


%%timeit -n 3 -r 1
for e in range(num_epochs):
    for i in range(epoch_length):
        update()
> 25 s ± 0 ns per loop (mean ± std. dev. of 1 run, 3 loops each)

%%timeit -n 3 -r 1
for e in range(num_epochs):
    for i in range(epoch_length):
        decorated_update()
> 25.3 s ± 0 ns per loop (mean ± std. dev. of 1 run, 3 loops each)
```
I think this decorator should be optional.

vfdev-5:
About `ignite.deterministic = True`, the implication of deterministic is only for Engine and its usage targets the training only. So, making it as a global parameter seems not to be an option due to its limited scope.

sdesrozis:
So overhead is 0.3 / 2500 = 0.1ms per call to have reproductible feature with brut force.

vfdev-5:
As discussed on the slack, we will opt for another way to rewrite Engine internals:

Main idea is "Synchronization on checkpoints only":
```python
set_seed(seed)
for e in range(n_epochs):
    # no more 
    # set_seed(e)
    for i, d in enumerate(data_loader):
        update(model, d)
        if need_to_checkpoint(i):
            checkpoint(i, to_save)
            set_seed(i)
```
When resume the run
```
load_checkpoint()
set_seed(start_iteration)
for e in range(start_epoch, n_epochs):
    for i, d in enumerate(data_loader):
        update(model, d)
        if need_to_checkpoint(i):
            checkpoint(i, to_save)
            set_seed(i)
```
Probably, this will simplify the code and make engine more transparent

PR #867, opened 2020-03-31
Improved README

vfdev-5:
Description:
- Better selling readme ;)

cc @ykumards @sdesrozis

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,13 @@ with the correct signature (we only require the first argument to be engine), e.
 function, class method etc. We do not require to inherit from an interface and override possibly 
 its abstract methods.
 
-#### Execute any number of functions whenever you wish
+### Execute any number of functions whenever you wish
+
+<details>
+<summary>
+Examples
+</summary>
+
 ```python
 trainer.add_event_handler(Events.STARTED, lambda engine: print("Start training"))
 
@@ -137,7 +143,14 @@ def log_something(engine):
     print(engine.state.output)
 ```
 
-#### Built-in events filtering
+</details>
+
+### Built-in events filtering
+
+<details>
+<summary>
+Examples
+</summary>
 
 ```python
 # run the validation every 5 epochs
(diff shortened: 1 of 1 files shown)


vfdev-5:
> we should notice that (thaks to python) data are easly accessible inside function
👍 

> the use case of custom event is not clear here. User should call function direclty.
Maybe I just put a link on TTBPT

> Note for later : we must do a tutorial to explain how develop metrics :)

this is done [here](https://pytorch.org/ignite/master/metrics.html#how-to-create-a-custom-metric), right ?

README.md
```diff
@@ -107,8 +107,84 @@ level of abstraction allows for a great deal more of flexibility, such
...
+        trainer.fire_event(CustomEvents.FOO_EVENT)
+```
+
+
+## Out-of-the-box metrics
+
+- Metrics for various tasks: Precision, Recall, Accuracy, Confusion Matrix, 
+IoU etc, ~20 regression metrics.
+
+- Users can compose its own metric by using basic arithmetical operations or torch methods:
```
ykumards: > Users can compose its own metric by using basic arithmetical operations or torch methods:

Shall we rephrase this as "Users can also compose their own metrics with ease from existing ones using arithmetic operations or torch methods" ?

README.md
```diff
@@ -107,8 +107,121 @@ level of abstraction allows for a great deal more of flexibility, such
...
 multiple losses and metrics in your training loop.
 
-Ignite also allows for multiple handlers to be attached to events, and a
-finer granularity of events in the engine loop.
+## Power of Events & Handlers
+
+The cool thing with handlers (vs “callback” interfaces) is that it can be any function 
+with the correct signature (we only require the first argument to be engine), e.g. lambda, simple 
+function, class method etc. We do not require to inherit from an interface and override possibly 
+its abstract methods.
```
ykumards: Suggested edit:
The cool thing with handlers is that they offer unparalleled flexibility (compared to say, callbacks). Handlers can be any function (e.g. lambda, simple function, class method etc.) with the correct signature, we only require that the first argument be `engine`. Thus, we do not require to inherit from an interface and override its abstract methods which could unnecessarily bulk up your code and its complexity.

vfdev-5: Thanks Yogesh! Will update

PR #868, opened 2020-03-31
issue #710 : add event list and or operator

### LINKED ISSUE #710: Handler Events as flags

sdesrozis:
Fixes #710  

Description:

Add an event list class to pack events which could be added with operator `|` on `Events` and `CallableEventWithFilter`

User code is
```
    def fn(e, b):
        pass

    trainer = Engine(fn)

    @trainer.on(Events.ITERATION_STARTED(once=1) | Events.ITERATION_STARTED(every=3) | Events.COMPLETED)
    def execute_some_handler(engine):
        print(engine.state.iteration)

    trainer.run(range(3), max_epochs=5)
```
TODO : improve doc

Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -9,7 +9,7 @@ from typing import Union, Optional, Callable, Iterable, Iterator, Any, Tuple
 
 import torch
 
-from ignite.engine.events import Events, State, CallableEventWithFilter, RemovableEventHandle, EventList
+from ignite.engine.events import Events, State, CallableEventWithFilter, RemovableEventHandle, EventsList
 from ignite.engine.utils import ReproducibleBatchSampler, _update_dataloader, _check_signature
 from ignite._utils import _to_hours_mins_secs
 
@@ -207,11 +207,41 @@ class Engine:
         wrapper._parent = weakref.ref(handler)
         return wrapper
 
+    def add_eventlist_handler(self, eventlist: EventsList, handler: Callable, *args, **kwargs):
+        """Add a collection of event handlers to be executed when the related events are fired.
+
+        See :meth:`~ignite.engine.Engine.add_event_handler`.
+
+        Args:
+            eventlist: A collection of events to attach the handler.
+            handler (callable): the callable event handler that should be invoked
+            *args: optional args to be passed to `handler`.
+            **kwargs: optional keyword args to be passed to `handler`.
+
+        Returns:
+            list of :class:`~ignite.engine.RemovableEventHandler`, which can be used to remove the handlers.
+
+        Example usage:
+
+        .. code-block:: python
+
+            engine = Engine(process_function)
+
+            def print_epoch(engine):
+                print("Epoch: {}".format(engine.state.epoch))
+
+            eventlist = Events.EPOCH_STARTED | Events.EPOCH_COMPLETED
+
+            engine.add_eventlist_handler(eventlist, print_epoch)
+
+        """
+        return [self.add_event_handler(e, handler, *args, **kwargs) for e in eventlist]
+
     def add_event_handler(self, event_name: str, handler: Callable, *args, **kwargs):
         """Add an event handler to be executed when the specified event is fired.
 
         Args:
-            event_name: An event to attach the handler to. Valid events are from :class:`~ignite.engine.Events`
+            event_name: An event to attach the handler. Valid events are from :class:`~ignite.engine.Events`
                 or any `event_name` added by :meth:`~ignite.engine.Engine.register_events`.
             handler (callable): the callable event handler that should be invoked
             *args: optional args to be passed to `handler`.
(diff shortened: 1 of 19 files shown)


ignite/engine/events.py
```diff
@@ -179,6 +182,33 @@ def call_once(engine):
...
+
+
+class EventList:
+
+    def __init__(self):
+        self._events = []
+
+    def append(self, event: Union[Events, CallableEventWithFilter]):
+        if not isinstance(event, (Events, CallableEventWithFilter)):
+            raise ValueError(f"argument event should be Events or CallableEventWithFilter, got: {type(event)}")
```
vfdev-5: This does not work in Py35 :(

tests/ignite/engine/test_custom_events.py
```diff
@@ -433,3 +433,36 @@ def test_distrib_gpu(distributed_context_single_node_nccl):
...
+
+    iterations = []
+
+    @engine.on(Events.ITERATION_STARTED(once=1) | Events.ITERATION_STARTED(every=3) | Events.COMPLETED)
+    def execute_some_handler(e):
+        iterations.append(e.state.iteration)
+
+    engine.run(range(3), max_epochs=5)
+
+    assert iterations == [1, 3, 6, 9, 12, 15, 15]
```
vfdev-5: Can we check that with mock and number of calls

ignite/engine/events.py
```diff
@@ -179,6 +182,33 @@ def call_once(engine):
     GET_BATCH_STARTED = "get_batch_started"
     GET_BATCH_COMPLETED = "get_batch_completed"
 
+    def __or__(self, other):
+        return EventList() | self | other
+
+
+class EventList:
```
vfdev-5: Shouldn't we inherit of something instead of introducing new type of events ?

sdesrozis: I didn't introduce a new type of event but a list of event with operator `__or__`.

ignite/engine/events.py
sdesrozis: Ouch you're right, I remove `fstring`.

ignite/engine/engine.py
```diff
@@ -334,7 +334,11 @@ def on(self, event_name, *args, **kwargs):
         """
 
         def decorator(f: Callable) -> Callable:
-            self.add_event_handler(event_name, f, *args, **kwargs)
+            if type(event_name) is EventList:
```
vfdev-5: Seems like EventList are only supported for decorator. It remains `engine.add_event_handler`

sdesrozis: Yep, I didn't realize that. I fix it.

ignite/engine/engine.py
```diff
@@ -244,6 +244,10 @@ def print_epoch(engine):
             See :class:`~ignite.engine.Events` for more details.
 
         """
+        if isinstance(event_name, EventList):
+            for e in event_name:
+                self.add_event_handler(e, handler, *args, **kwargs)
+            return
```
vfdev-5: Shouldn't we still return something here ?
Otherwise, the docstring is incomplete.

sdesrozis: Indeed, we should return something, I didn't catch that. Taking a closer look, I think `add_event_handler` shouldn't be the way to add a list of events. A specific function should be defined.
@vfdev-5 thoughts ?

sdesrozis: `add_eventlist_handler` which returns a list of `RemovableEventHandler` ?

sdesrozis: See new commit. Skeptical about `add_eventlist_handler` naming... `add_event_handlers` ?

vfdev-5:
@sdesrozis I'm not a fan of introducing another public method for that.
Idea of returning `RemovableEventHandler` is that it can be used as a context manager to remove attached handlers on exit of the context manager. I think we need to keep single method `add_event_handler` and return something such that it could be used as `RemovableEventHandler`... What do you think ?

sdesrozis:
Now `add_event_handler` returns `RemovableEventHandler` compatible with `EventsList`.

TODO doc

vfdev-5:
Sorry, my bad, the name should remain as previously as `RemovableEventHandle` and not `RemovableEventHandLER`.
Yes, now, it looks good for me. Please, add a test with `RemovableEventHandle` for event list.

sdesrozis:
Last missing point is 
* README.md in "Power of Events & Handlers"

sdesrozis:
@vfdev-5 you can comment :)

vfdev-5:
@sdesrozis looks good to me. For perfect fit, I would put added section just after "Built-in events filtering"...

vfdev-5:
Let me fix coverage relative decrease: https://codecov.io/gh/pytorch/ignite/pull/868/diff#diff-aWduaXRlL2VuZ2luZS9ldmVudHMucHk=
- https://codecov.io/gh/pytorch/ignite/pull/868/diff#D6-218
- https://codecov.io/gh/pytorch/ignite/pull/868/diff#D6-336

Issue #878, opened 2020-04-04
simple arg name mismatch for mnist tensorboard example


DhDeepLIT:
## 🐛 Bug description

Example for mnist with tensorboard does not work due to wrong arg name.

>git clone https://github.com/pytorch/ignite
>cd examples/mnist
>python3.6 mnist_with_tensorboard.py --log_dir=/tmp/tensorboard_logs
Traceback (most recent call last):
  File "mnist_with_tensorboard.py", line 161, in <module>
    run(args.batch_size, args.val_batch_size, args.epochs, args.lr, args.momentum, args.log_interval, args.log_dir)
  File "mnist_with_tensorboard.py", line 89, in run
    writer = create_summary_writer(model, train_loader, log_dir)
  File "mnist_with_tensorboard.py", line 76, in create_summary_writer
    writer = SummaryWriter(logdir=log_dir)
TypeError: __init__() got an unexpected keyword argument 'logdir'

Traceback (most recent call last):
  File "mnist_with_tensorboard.py", line 161, in <module>
    run(args.batch_size, args.val_batch_size, args.epochs, args.lr, args.momentum, args.log_interval, args.log_dir)
  File "mnist_with_tensorboard.py", line 89, in run
    writer = create_summary_writer(model, train_loader, log_dir)
  File "mnist_with_tensorboard.py", line 76, in create_summary_writer
    writer = SummaryWriter(logdir=log_dir)
TypeError: __init__() got an unexpected keyword argument 'logdir'

Just something that runs ....

## Environment
Ubuntu 18.06 LTS, dual 1080Ti, 32 Go RAM, intel i7-2600K 
>pip3.6 install torch
>pip3.6 install torchvision
>pip3.6 install tensorboard

>pip3.6 list | grep -E 'tensor'\|'torch'
pytorch-ignite            0.4.0              
tensorboard               2.2.0              
tensorboard-plugin-wit    1.6.0.post2        
torch                     1.4.0              
torchvision               0.5.0

vfdev-5:
@DhDeepLIT thanks for the report !
It looks like to be related to different signatures of SummayWriter between torch internal tensorboard and tensorboardX:
- https://pytorch.org/docs/stable/tensorboard.html#torch.utils.tensorboard.writer.SummaryWriter
vs
- https://github.com/lanpa/tensorboardX/blob/master/tensorboardX/writer.py#L210

vfdev-5:
[...]

PR #879, opened 2020-04-04
Update unittests.yml

vfdev-5:
Description:
This PR is to enforce our tests while using native torch tensorboard. It is related to the recent bug #878 . Once it is fixed we can merge this one.
- Removed tensorboardX to test with torch tensorboard

Check list:


sdesrozis:
We can’t do both ? With tensorboardX and without (using torch) ?

vfdev-5:
> We can’t do both ? With tensorboardX and without (using torch) ?

To do both, we need to uninstall tensorboardX and rerun all unit tests and smoke tests.
Idea is to keep Travis CI working with TbX and GithubCI with native torch tensorboard (as it will be probably the future).

PR #882, opened 2020-04-04
#864 decorator one_rank_only

### LINKED ISSUE #864: Add helper decorators to use event handler or method with specified rank
## 🚀 Feature

Idea is to simplify the following code:
```python
import torch.distributed as dist

if dist.get_rank() == 0:
     
    @trainer.on(...)
    def some_handler(_):
        ...

    tb_logger = TensorboardLogger(log_dir="experiments/tb_logs")        
```
into 

```python
from ignite.utils import one_rank_only

@trainer.on(...)
@one_rank_only   #  equivalent @one_rank_only(rank=0)
def some_handler(_):
    ...

# implicitly work only on rank 0
tb_logger = TensorboardLogger(log_dir="experiments/tb_logs") 
```

sdesrozis:
Fixes #864

Description:

add decorator `one_rank_only`. Use parameter `rank` to choose which process triggers the handler and `barrier` to add a sync over process.

```python
@engine.on(...)
@one_rank_only(rank=1, barrier=True)
def _(_):
    ...
```

Check list:


sdesrozis:
@vfdev-5 @ykumards any comment ?

Could you give me input to put a distributed test in CI if it matches the issue ? Thanks in advance :)

ignite/utils.py
```diff
@@ -125,3 +126,38 @@ def setup_logger(
         logger.addHandler(fh)
 
     return logger
+
+
+def one_rank_only(*args, **kwargs):
```
vfdev-5, line 131: Please, specify `rank=0` as default signature.
For example, like that (https://github.com/pytorch/ignite/blob/master/ignite/metrics/metric.py#L296):
```python
def one_rank_only(rank=0)
    def wrapper(func):
        def another_wrapper(*args, **kwargs):
            if dist.get_rank() == rank:
                return func(*args, **kwargs)
    return wrapper
```
?

Maybe, we also need to use `functools.wraps`.

vfdev-5:
@sdesrozis to test it it is possible at first to work with "gloo" backend.
For example, like here:
https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/test_accumulation.py#L366
So, idea is to check a value of a tensor. What do you think ?

ignite/utils.py
```diff
@@ -125,3 +126,38 @@ def setup_logger(
...
+        @one_rank_only # equivalent @one_rank_only(rank=0)
+        def some_handler(_):
+            ...
+
+        @engine.on(...)
+        @one_rank_only(rank=1)
+        def some_handler(_):
+            ...
+    """
+    def _one_rank_only(fun):
```
vfdev-5, line 147: Maybe, we need to enclose the code with barriers:
https://pytorch.org/docs/stable/distributed.html#torch.distributed.barrier

sdesrozis: It means every procs should fire barrier event. Although I’m not sure that it is necessary. The code will be synchronised to next synchronous collective or wait. 

Btw, no comm is allowed in the handler.

ignite/utils.py
sdesrozis: Ok but api will be
```
@one_rank_only()
````
not
```
@one_rank_only   # no ()
```
with `rank` passed 
```
@one_rank_only(tank=...)
```

My hand pattern allow to do not have `()` and respect what is designed in issue. But it is less readable and more complicated for few benefits.

vfdev-5: I do not mind to have `@one_rank_only()` and simplified code :)

ignite/utils.py
vfdev-5: > Btw, no comm is allowed in the handler.

What do you mean by this ?

sdesrozis: Only one rank runs the code so no comm otherwise dead lock.

ignite/utils.py
sdesrozis: Ok good ! 👍🏻

PR #884, opened 2020-04-04
WIP #872 metrics are now dillable (but not pickleable)

### LINKED ISSUE #872: Make Engine with Metrics pickleable
## 🚀 Feature

Idea is to make `Metric`class pickleable and fix #623
We need to be able to spawn Engine with metrics as suggested in #623 

cc @sdesrozis

sdesrozis:
Fixes #872 

Description:

Modify `metric` to allow `dill` serialization (but not `pickle`).

Add `mnist` test using `dill`.

Check list:


sdesrozis:
@vfdev-5 could you check the test added ?

ignite/metrics/metric.py
```diff
@@ -292,6 +292,13 @@ def __getitem__(self, index: Any):
 
         return MetricsLambda(lambda x: x[index], self)
 
+    def __getstate__(self):
+        state = self.__dict__.copy()
```
vfdev-5, line 296: Maybe we can simply return the dict without copying it ?

sdesrozis: I try that

PR #887, opened 2020-04-05
Fixed device conversion method

### LINKED ISSUE #886: utils.convert_tensor considers `device = 0` to be no device
## 🐛 Bug description
In utils.convert_tensor, this line appears:
`return tensor.to(device=device, non_blocking=non_blocking) if device else tensor`

This means that for `device = 0` (as returned from `torch.cuda.current_device`) no conversion is applied, which can be very confusing. I might add a PR for that tomorrow, unless people tell me there's a reason to leave that line as it is.

For reproduction:
```python
import torch

mytens = torch.zeros(2)
device = torch.cuda.current_device()
converted_tens = convert_tensor(mytens, device)
assert converted_tens.device == device
```

## Environment

 - Ignite Version (e.g., 0.3.0): 0.3
 - OS (e.g., Linux): Windows 10
 - Python version: 3.7.6

bendboaz:
Fixes #886

Description:
Fixed bug from #886 where device \#0 was considered "None" and no conversion was performed.
Three words added to the code overall, so no tests/docs added.

Check list:

### DIFF
--- a/ignite/utils.py
+++ b/ignite/utils.py
@@ -15,7 +15,7 @@ def convert_tensor(
     """Move tensors to relevant device."""
 
     def _func(tensor: torch.Tensor) -> torch.Tensor:
-        return tensor.to(device=device, non_blocking=non_blocking) if device else tensor
+        return tensor.to(device=device, non_blocking=non_blocking) if device is not None else tensor
 
     return apply_to_tensor(input_, _func)
 


bendboaz:
Definitely, and my fix for this (in my code, not in the PR) was actually moving to the correct definition of devices. Anyways, you're right on both accounts :)

PR #889, opened 2020-04-05
#869 engine optional in handlers

### LINKED ISSUE #869: Engine optional in signature of event function
## 🚀 Feature

The idea is to make optional engine argument in handler:
```python
trainer = ... 

@trainer.on(Events.COMPLETED)
def _(): # no need to pass engine
  print("Training is finished")
  print(f"number of iterations {trainer.state.iterations}") # use trainer directly
``` 

PS. Sylvain, I edited a bit the text

sdesrozis:
Fixes #869 

Description:

First attempt to discuss
- use of `inspect` to catch if `engine` is in signature

This is done in `_fire_events`. It should be better if `engine` was added in signature during registering. It implies a different order for arguments passed to `_fire_events` (basically `Exception`).

TODO : docs 

@vfdev-5 thoughts ?

Check list:


ignite/engine/engine.py
```diff
@@ -371,7 +372,12 @@ def _fire_event(self, event_name: Any, *event_args, **event_kwargs) -> None:
             self.last_event_name = event_name
             for func, args, kwargs in self._event_handlers[event_name]:
                 kwargs.update(event_kwargs)
-                func(self, *(event_args + args), **kwargs)
+                signature = inspect.signature(func)
+                try:
+                    signature.bind(self, *(event_args + args), **kwargs)
+                    func(self, *(event_args + args), **kwargs)
+                except TypeError:
+                    func(*(event_args + args), **kwargs)
```
vfdev-5, line 380: @sdesrozis can't we inspect and setup *args with or without engine at `add_event_handler` time and here just trigger the handler without dispatch ?

ignite/engine/utils.py
```diff
@@ -55,15 +55,31 @@ def __len__(self) -> int:
...
-        passed_params = [engine] + list(args) + list(kwargs)
+        passed_params = list(args) + list(kwargs)
+        raise ValueError(
+            "Error adding {} '{}': "
+            "takes parameters {} but will be called with {} "
+            "({}).".format(fn, fn_description, fn_params, passed_params, exception_msg)
+        )
+
+
+def _check_partial_signature(fn: Callable, fn_description: str, *args, **kwargs) -> None:
```
vfdev-5, line 74: Could you please explain the need of partial bind ? 
I thought we could check both in a single `_check_signature` like method ?

sdesrozis: First, I think `_check_signature` could be refactor to have engine in `*args` and should be generic.

I implemented the `_check_partial_signature` to check arguments of an handler without knowing if it should contains the `engine`.

I should have preferred to check the signature of the handler wrt `*args` and `**kwargs` to see if engine is needed or not. And thus, engine should have been added to `*args`. No check needed at `_fire_event`.

But `_fire_event` could add own argument... And `engine` is always the first argument of `*args`.

ignite/engine/engine.py
sdesrozis: See the other comment.

I would have preferred that. But exception arguments could be added by `_fire_event`. If we agree the order of argument (`engine` not always the first), we could do what you said (and I prefer that!)

vfdev-5: We can keep the order as it is currently: Engine should be the first argument.

sdesrozis: Consider this pseudo code
```
engine.add_event_handler(event, fn, *args, **kwargs)
```
2 alternatives :
- `fn` doesn’t have engine in signature, we keep `(fn,*args, **kwargs)` for `event`
- `fn` has engine in signature, we keep `(fn,(engine, *args), **kwargs)` for `event`
From now, we shouldn’t have no longer dealing with `engine`. 
But consider 
```
_fire_event(event, *event_args, **event_kwargs)
```
We have to merge arguments we have kept for `event` and the arguments mais passed to the function `_fire_event`.

ignite/engine/utils.py
sdesrozis: > I thought we could check both in a single `_check_signature` like method ?

We could do that 👍🏻

sdesrozis: Notice that with optional `engine`, the following snippet is correct
```python
def handler(arg):
    pass

engine.add_event_handler(Events.STARTED, handler)    # arg == engine 
engine.add_event_handler(Events.STARTED, handler, 1) # arg == 1
```

vfdev-5: Yes, it is correct. We need to put a similar example into docstring.

ignite/engine/engine.py
vfdev-5: @sdesrozis I think we still can store `(fn, *args, **kwargs)` at `add_event_handler`:
```python
class Engine:
    
    def __init__(self):
        self.hs = []


def add(engine, fn, *args, **kwargs):
    
    signature = inspect.signature(fn)
    
    event_args = ()
    event_kwargs = {}
        
    try:
        signature.bind(engine, *args, **kwargs)
        args = (engine, ) + event_args + args
        kwargs.update(event_kwargs)
    except TypeError as e:
        try:
            signature.bind(*args, **kwargs)
            args = event_args + args
            kwargs.update(event_kwargs)
        except TypeError as e:
            raise RuntimeError("Wrong signature...")            

    engine.hs.append((fn, args, kwargs))


def trigger(engine):
    for h in engine.hs:
        fn = h[0]
        fn(*h[1], **h[2])
```
What do you think ?

Anyway, we will need to provide extensive tests to cover possible cases.

sdesrozis: I exactly did that :)

ignite/engine/engine.py
```diff
@@ -5,6 +5,7 @@
 import weakref
 import random
 import warnings
+import inspect
```
vfdev-5, line 8: Seems like this import is not used

sdesrozis: Oups Yes !

PR #904, opened 2020-04-08
Update common.py

willfrey:
I think `_setup_common_training_handlers(...)` is missing its `return trainer` statement at the end of it. I've added it here.

If this is in error, please just reject and close this.

Fixes # 

Description:

Check list:

### DIFF
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -152,6 +152,8 @@ def _setup_common_training_handlers(
         ProgressBar(persist=True, bar_format="").attach(
             trainer, event_name=Events.EPOCH_STARTED, closing_event_name=Events.COMPLETED
         )
+        
+    return trainer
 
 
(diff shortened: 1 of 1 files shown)


vfdev-5:
@willfrey thanks for pointing out, actually all `common.setup_common*`  are working inplace on input trainer : https://github.com/pytorch/ignite/blob/master/examples/references/classification/imagenet/code/scripts/common_training.py#L81

So, I think we need to remove all `return trainer`. Could you please do this ?

willfrey:
I think that should do the trick. I removed the `return trainer` statements as well as the two `return _helper(...)` statements in the public `setup_common_training_handlers` function.

vfdev-5:
Thanks @willfrey ! I fixed flake8

willfrey:
Thanks! I made the edits directly on here, so the formatting was probably sloppy.

PR #910, opened 2020-04-09
#902-Patch avoid moving model

### LINKED ISSUE #902: Avoid moving model to device in create_supervised*
## 🚀 Feature

As discussed [here](https://github.com/pytorch/ignite/issues/830#issuecomment-611041703) and previous messages, according to pytorch docs on model and optimizer. User should not move model to another device after constructing an optimizer for the model: 
- https://pytorch.org/docs/stable/optim.html#constructing-it
- https://github.com/pytorch/pytorch/issues/7844#issuecomment-503713840

We need to handle this in the code:
- remove the code that moves the model to device.
- update the docstring and add a warning note on changed behaviour, cite pytorch docs.
- update all basic examples : mnist etc who currently use unwanted behaviour.

cc @kai-tub

kai-tub:
Fixes #902

Description:
- Remove the model move
- Update the docstring and include a warning note
- Updated the tests to make sure the behavior is as expected

Grepped the repo and made the required changes to keep everything
in-line with the new changes to `create_supervised`.
Hopefully, I didn't miss anything.
I did *not* rerun the FastaiLRFinder_MINST.ipynb.
The output would change and add unnecessary size to the repo.

Check list:

### DIFF
--- a/tests/ignite/engine/test_create_supervised.py
+++ b/tests/ignite/engine/test_create_supervised.py
@@ -81,11 +81,13 @@ def test_create_supervised_trainer_traced_with_cpu():
 
 @pytest.mark.skipif(not torch.cuda.is_available(), reason="Skip if no GPU")
 def test_create_supervised_trainer_on_cuda():
+    device = "cuda"
     model = Linear(1, 1)
+    model.to(device)
     model.weight.data.zero_()
     model.bias.data.zero_()
     optimizer = SGD(model.parameters(), 0.1)
-    trainer = create_supervised_trainer(model, optimizer, mse_loss, device="cuda")
+    trainer = create_supervised_trainer(model, optimizer, mse_loss, device=device)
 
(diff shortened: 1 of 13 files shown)


tests/ignite/engine/test_create_supervised.py
```diff
@@ -114,8 +117,9 @@ def test_create_supervised_trainer_on_cuda_with_model_on_cpu_after_init():
     y = torch.tensor([[3.0], [5.0]])
     data = [(x, y)]
 
-    model.to("cpu")
-    trainer.run(data)
+    with pytest.raises(RuntimeError) as runtime_exception:
```
vfdev-5, line 120: Can we use this to catch the exception:
```python
with pytest.raises(RuntimeError, match=r"device type"):
```

vfdev-5: Updated by myself to check

Issue #912, opened 2020-04-11
Higher API for training


Minyus:
## 🚀 Feature

Higher API for training will improve usability of Ignite as Keras did.

My implementation [here](https://github.com/Minyus/pipelinex/blob/master/src/pipelinex/ops/ignite_ops.py) could be used as the base code.

sdesrozis:
Thank you for this PR and your help !!

Let’s discuss and design now about this very nice feature 😊

sdesrozis:
I really enjoy the idea to have open (and clear) core tools (`engine`, `handlers`, etc.) and an api helper for compose app mixing these core tools  👍🏻

As we discussed in slack channel, we could think about very high level feature including graphical frontend. Although first step is focused on trainer 😊 

Could you use your api on some examples from `ignite/examples` ? Thank you again!

vfdev-5:
In my opinion, it would be interesting to create something like 
```python

class Trainer(Engine):

     def __init__(self, *agrs, **kwags):
         ...

     def train_step(self, ...):
          ...

     etc
```


cc @ericspod

ericspod:
I have simplified what I use in my code base which @vfdev-5 is referencing [here](https://gist.github.com/ericspod/f4da372d22cc8da420ee74b8968303cd). 

The idea is to extend the Engine class with a BaseEngine (a different name would be good) which adds methods for converting between numpy arrays and tensors, wrapper methods around doing forward passes on network and loss functions, and for simplifying inference with a network. The Trainer and Evaluator classes inherit from this and provide their own methods for implementing a training loop or using an external callable as in the base Engine class, evaluating the network on all the data from a dataset and returning the mean loss, and others. 

This is a much more opinionated set of classes than what is provided with Ignite's Engine. I think Ignite is correct in providing a base level of general purpose mechanisms rather than a more involved concept of what should be done. These classes for example work best with a training regime where the batch contains inputs to the network and to a loss function and the network produces values for the loss function as well. Not all training schemes would have these. Adding a high level API might be useful to many but so long as it's optional Ignite can stay generic. 

[...]

PR #924, opened 2020-04-15
Add declarative trainer

Minyus:
Fixes # 912

Description:
Higher API for training will improve usability of Ignite.

Check list:
* [x ] New doc strings: description and/or example code are in RST format

### DIFF
--- a/examples/contrib/mnist/mnist_with_declarative_trainer.py
+++ b/examples/contrib/mnist/mnist_with_declarative_trainer.py
@@ -1,0 +1,120 @@
+
+import torch
+from torch import nn
+from torch.optim import SGD
+import torch.nn.functional as F
+from torchvision.transforms import Compose, ToTensor, Normalize
+from torchvision.datasets import MNIST
+from ignite.metrics import Accuracy, Loss
+from ignite.contrib.handlers.param_scheduler import LinearCyclicalScheduler
+from ignite.contrib.trainers.declarative_trainer import NetworkTrain
+
+
+class Net(nn.Module):
+    def __init__(self):
+        super(Net, self).__init__()
+        self.conv1 = nn.Conv2d(1, 10, kernel_size=5)
+        self.conv2 = nn.Conv2d(10, 20, kernel_size=5)
+        self.conv2_drop = nn.Dropout2d()
+        self.fc1 = nn.Linear(320, 50)
+        self.fc2 = nn.Linear(50, 10)
+
+    def forward(self, x):
+        x = F.relu(F.max_pool2d(self.conv1(x), 2))
+        x = F.relu(F.max_pool2d(self.conv2_drop(self.conv2(x)), 2))
+        x = x.view(-1, 320)
+        x = F.relu(self.fc1(x))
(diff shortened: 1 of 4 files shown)


sdesrozis:
@Minyus Thank you very much for this PR !!

I make you my feedback on your proposal quickly :)

vfdev-5:
@Minyus thanks a lot for the PR ! I really appreciate your contribution, however it would be great to discuss more about the API. We already stated some general remarks on the API in the related issue #912 

In my opinion, class `NetworkTrain` has very specific integrations with MLflow which can be a limitation for some users. Another point is about exposing trainer in the higher API in case of adding more user-specific handlers. 

please, let's discuss more about the API in #912.

vfdev-5:
@Minyus can we work out this at first ? 
```python
class TimeLimit:
    def __init__(self, limit_sec=3600):
        self.limit_sec = limit_sec
        self.start_time = time.time()

    def __call__(self, engine):
        elapsed_time = time.time() - self.start_time
        if elapsed_time > self.limit_sec:
            log.warning("Reached the time limit: {} sec. Stop training".format(self.limit_sec))
            engine.terminate()
```

This looks good enough. Maybe, in addition, we can set `Checkpoint` or `ModelCheckpoint` to save the work before terminating ?

Minyus:
@vfdev-5 
OK, I'll move `TimeLimit` to `ignite/handlers/time_limit.py`. 
Let's add saving before terminating to ToDo list as it needs further consideration (e.g. what if model checkpoint path is not specified, etc.)

vfdev-5:
@Minyus 

> @vfdev-5
> OK, I'll move `TimeLimit` to `ignite/handlers/time_limit.py`.
> Let's add saving before terminating to ToDo list as it needs further consideration (e.g. what if model checkpoint path is not specified, etc.)

@Minyus thanks, but it would be much better to split this PR into several ones. One PR by feature.
So, please, send the code of `TimeLimit` into a new PR. It would be nice to follow  https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md 
Thanks

Minyus:
As discussed in Slack, this PR is suspended.

Meanwhile, please feel free to send PR (including adding documentation, test, example, etc.) to PipelineX so that I can send a complete PR to Ignite later on.

higher API for training
https://github.com/Minyus/pipelinex/blob/master/src/pipelinex/ops/ignite/declaratives/declarative_trainer.py

MNIST example to use higher API
https://github.com/Minyus/pipelinex/blob/master/examples/mnist/mnist_with_declarative_trainer.py

TimeLimit handler
https://github.com/Minyus/pipelinex/blob/master/src/pipelinex/ops/ignite/handlers/time_limit.py

Cohen Kappa Score metric
https://github.com/Minyus/pipelinex/blob/master/src/pipelinex/ops/ignite/metrics/cohen_kappa_score.py

Please also feel free to copy any code in PipelineX to use in PyTorch Ignite.
I'm happy to see Ignite to be enhanced in any way.

sdesrozis:
Thank you again and see you soon !!

PR #926, opened 2020-04-16
Wandb logger

### LINKED ISSUE #865: Add W&B logger
## 🚀 Feature

Add logger to W&B similarly to MLflow, Neptune...

fdlm:
Fixes #865 

Description:

Implements logging to Weights & Biases. 

Potential problems:
 * No integration testing, because you need a W&B login to initialize the logger.

Check list:


vfdev-5:
FYI, just found on their site : https://github.com/wandb/gitbook/blob/master/library/frameworks/pytorch/ignite.md

vfdev-5:
@fdlm thanks for the PR ! Could you please provide MNIST example of using WandB like https://github.com/pytorch/ignite/blob/master/examples/contrib/mnist/mnist_with_neptune_logger.py

vfdev-5:
@sdesrozis could you please check mnist example by running it locally. Thanks!

examples/contrib/mnist/mnist_with_wandb_logger.py
```diff
@@ -0,0 +1,191 @@
...
+    device = "cpu"
+
+    if torch.cuda.is_available():
+        device = "cuda"
+
+    optimizer = SGD(model.parameters(), lr=lr, momentum=momentum)
+    criterion = nn.CrossEntropyLoss()
+    trainer = create_supervised_trainer(model, optimizer, criterion, device=device)
+
+    if sys.version_info > (3,):
```
vfdev-5, line 81: Maybe we can remove GPU logging as W&D does it by itself.

examples/contrib/mnist/mnist_with_wandb_logger.py
```diff
@@ -0,0 +1,191 @@
...
+        config={
+            "train_batch_size": train_batch_size,
+            "val_batch_size": val_batch_size,
+            "epochs": epochs,
+            "lr": lr,
+            "momentum": momentum,
+        },
+    )
+
+    def iteration(engine):
```
vfdev-5, line 115: We can use `global_step_from_engine` instead of `iteration`:
`global_step_transform=global_step_from_engine(trainer)`

vfdev-5:
@fdlm thanks for the update. Remains Docs as here: https://github.com/pytorch/ignite/blob/master/docs/source/contrib/handlers.rst#neptune_logger

sdesrozis:
WB is just an awesome logger !! That works perfectly. Data in cloud on W&B are amazing. Very very nice job :) 

We can merge the PR when conversations will be resolved.

examples/contrib/mnist/mnist_with_wandb_logger.py
fdlm: Good point, will fix.

examples/contrib/mnist/mnist_with_wandb_logger.py
fdlm: The problem is that `global_step_from_engine` will use the **iteration** number for handlers called on `Events.ITERATION_COMPLETED`, but **epoch** number for handlers called on `Events.EPOCH_COMPLETED`. W&B does not allow to log events with a smaller step than previously logged, so I need to make sure to always use the iteration number.

vfdev-5: OK, I see what you mean. So, we can not log at all any epoch number if we log iterations ?

fdlm: Exactly, at least not directly. The workaround (as described in the W&B docs) is to log epochs as "metric", and then select it as X-axis in the W&B web interface. Not sure how to facilitate this in the Logger here.

Thanks for the simpler `global_step_transform`, will add it!

PR #895, opened 2020-04-07
Simplified engine - BC breaking change

vfdev-5:
This PR is split into multiple PRs to ease the review:
- #938 (merged)
- #939 (merged)
- #9XX

--- 
Fixes #795 

Related to https://github.com/pytorch/ignite/pull/866

Description: 
Now, `Engine` runs without dataflow sync and thus has no issue with random states (e.g. #795)

Example save-resume on MNIST (https://github.com/pytorch/ignite/blob/19447d07bbe3e25a80bfa634add6f3fe75f1f52b/examples/mnist/mnist_save_resume_engine.py) :

```bash
rm -rf logs && python3 mnist_save_resume_engine.py --log_dir=logs --epochs 1 --deterministic && python3 mnist_save_resume_engine.py --log_dir=logs --epochs 1 --resume_from=logs/checkpoint_550.pt --deterministic
```
Gives the logs : 
Screen Shot 2020-04-18 at 03 05 31

We still provide an option to keep BC with v0.3.0:
- added module deterministic to be able sync dataflow:
```python

    import torch
    from torch.utils.data import DataLoader
    from ignite.engine import DeterministicEngine, Events
    from ignite.utils import manual_seed


    def random_train_data_loader(size):
        data = torch.arange(0, size)
        return DataLoader(data, batch_size=4, shuffle=True)


    def print_train_data(engine, batch):
        i = engine.state.iteration
        e = engine.state.epoch
        print("train", e, i, batch.tolist())

    trainer = DeterministicEngine(print_train_data)

    print("Original Run")
    manual_seed(56)
    trainer.run(random_train_data_loader(40), max_epochs=2, epoch_length=5)

    print("Resumed Run")
    # Resume from 2nd epoch
    trainer.load_state_dict({"epoch": 1, "epoch_length": 5, "max_epochs": 2, "rng_states": None})
    manual_seed(56)
    trainer.run(random_train_data_loader(40))

```

- Added new engine events (used with deterministic methods)
    - DATALOADER_STOP_ITERATION : engine's specific event triggered when dataloader has no more data to provide

    - TERMINATE_SINGLE_EPOCH : triggered when the run is about to end the current epoch,
      after receiving :meth:`~ignite.engine.Engine.terminate_epoch()` call.

    - TERMINATE : triggered when the run is about to end completely,
      after receiving :meth:`~ignite.engine.Engine.terminate()` call.

- Updated docs

TODO: 
[...]

PR #938, opened 2020-04-18
Removed dataflow sync from Engine

### LINKED ISSUE #795: Dataflow randomness problem with infinite data iterators

### LINKED ISSUE #818: Strange program flow if using custom dataloader
If using a custom dataloader that implements the __len__ method, the current epoch is terminated upon this line https://github.com/pytorch/ignite/blob/90d4e1b78f280293116706d5c42cc9243212ea29/ignite/engine/engine.py#L455
which causes an exit of the while True loop in 
https://github.com/pytorch/ignite/blob/90d4e1b78f280293116706d5c42cc9243212ea29/ignite/engine/engine.py#L405.

However, at the start of the next epoch, the dataloader is exhausted and thus StopIteration is thrown only now, i.e. at the beginning of an actually new epoch. 

I find this behaviour rather strange and unnatural. Can someome explain me the reason for this, maybe I just dont understand the underlying purpose

### LINKED ISSUE #935: Improvements around ReproducibleBatchSampler.
## 🚀 Feature

IMHO an attempt should be made to not wrap or change the data pipeline objects. Despite the name, this `ReproducibleBatchSampler` seems to be about making datasets *resumable* from the middle of an epoch by skipping the first few examples, rather than being *reproducible* since it doesn't appear to set any seeds on the samplers.

(If this is not the case, I think the point is even stronger. I'd prefer not to have seeds and "reproducibility" introduced into my data pipeline in the background. They are all good things, but not when it happens without my knowledge, or without an option to disable it).

Silently wrapping objects or creating new instances is can introduce unexpected issues, eg:
- point brought up in #812, about side effects.
- `ReproducibleBatchSampler` seems to assume it's wrapping the pytorch `BatchSampler`. For instance, it assumes the `BatchSampler` has a `sampler` instance variable. This this the case for the pytorch class, but not required in general. `BatchSampler` is just a sampler where the `__iter__` returns a batch of indexes.
- `ReproducibleBatchSampler` samples all the indexes first. This isn't necessarily a trivial amount of time or memory to hold l those ints if the dataset is large.

First and foremost, I believe this new behavior should be *very* prominently noted in the documentation, change notes, a warning in the logs, etc. I realized after the fact this is a "note" in the `run` method engine,  but as a user who is upgrading from 0.2.1, I would have had no idea this was happening if it wasn't for my implementation of batch sampler trivially not being compatible. 

Perhaps the behavior should be changed to: if the data loader stack is not holding an instance of `ReproducibleBatchSampler`, then engine simply loads and does nothing with the batches that need to be skipped upon resume. Users who wish to have the `ReproducibleBatchSampler` option can explicitly use this class when constructing `DataLoader`.

Luckily, ignite is a small library and its code is very readable (awesome, very happy user, thanks!). Upon reading the code, I see that I can simply implement my own `ReproducibleBatchSampler` to bypass the three concerns I pointed out above. For the case where batches are simply sampled iid with replacement out of the data, it's sufficient to simply shorten the number of elements of the first call to `__iter__` when the dataset is resumed.

So this is more of a suggestion about default behavior, which I found to be somewhat unexpected.

Thanks for your work on this library!

vfdev-5:
- added deterministic methods
- Fixes #818
- Fixes #795
- Fixes #935 

Description:
PR (1/3) to that splits #895

- Removed dataflow sync from Engine

- Added new engine events (used with deterministic methods)
    - DATALOADER_STOP_ITERATION : engine's specific event triggered when dataloader has no more data to provide

    - TERMINATE_SINGLE_EPOCH : triggered when the run is about to end the current epoch,
      after receiving :meth:`~ignite.engine.Engine.terminate_epoch()` call.

    - TERMINATE : triggered when the run is about to end completely,
      after receiving :meth:`~ignite.engine.Engine.terminate()` call.

Check list:


ignite/engine/deterministic.py
```diff
@@ -0,0 +1,129 @@
...
+    torch.set_rng_state(rng_states[1])
+    try:
+        import numpy as np
+
+        np.random.set_state(rng_states[2])
+    except ImportError:
+        pass
+
+
+def _repr_rng_state(rng_states):
```
sdesrozis, line 107: Why do we need this function ?

vfdev-5: This can be helpful to debug random state

Issue #784, opened 2020-02-14
Improve Frequency


vfdev-5:
## 🚀 Feature

If we would like to log datapoints/second every 100 iterations, we most probably do like this 
```python
wps_metric = Frequency(output_transformer=lambda x: x['ntokens'])
wps_metric.attach(trainer, name='wps', event_name=Events.ITERATION_COMPLETED(every=100))
```
however, seems like this wont take into account all other iterations while computing the total number of tokens.
```python
class Frequency(Metric):
    ....
    def attach(self, engine, name, event_name=Events.ITERATION_COMPLETED):
        engine.add_event_handler(Events.EPOCH_STARTED, self.started)
        engine.add_event_handler(event_name, self.iteration_completed)
        engine.add_event_handler(event_name, self.completed, name)
```
IMO, should be 
```python
class Frequency(Metric):
    ....
    def attach(self, engine, name, event_name=Events.ITERATION_COMPLETED):
        engine.add_event_handler(Events.EPOCH_STARTED, self.started)
        engine.add_event_handler(Events.ITERATION_COMPLETED, self.iteration_completed)
        engine.add_event_handler(event_name, self.completed, name)
```

cc @erip

anmolsjoshi:
> however, seems like this wont take into account all other iterations while computing the totla number of tokens.

@vfdev-5 is this because the user can change the event?

Doesn't the above example work as proposed?

erip:
Not sure how I missed the cc here - sorry about that! Let me reacquaint myself quickly and I'll try to submit a PR.

erip:
Yes - this seems right. I'll write some tests to confirm this works as expected first.

erip:
I'm having some trouble executing the distributed tests on my Mac. 😬

I'm trying to execute 

```
py.test --cov ignite --cov-append --cov-report term-missing --dist=each --tx "2*python=python3.6" tests -m distributed -vvv
```
but I'm met with `ValueError: no gateway type found for 'python=python3.6'`.

Does this look familiar, @vfdev-5? Once I have this figured out I can start running some my tests locally and confirm your hypothesis that the frequency stats are being reset.

erip:
I'm actually having trouble getting the original tests to pass on my Mac. 😬 

When I run

```sh
$ WORLD_SIZE=2 py.test --dist=each --tx "2*popen//python=python3" tests/ignite/metrics/test_frequency.py -m distributed
```

[...]

PR #958, opened 2020-04-22
add time in state.

### LINKED ISSUE #916: Insert logged times into state
## 🚀 Feature

By default, Engine measures times for single epoch and total run. These times are logged with `engine.logger`:
https://github.com/pytorch/ignite/blob/efc45a79ee0a5bf8105c3caff194758439db2426/ignite/engine/engine.py#L793-L796
and
https://github.com/pytorch/ignite/blob/efc45a79ee0a5bf8105c3caff194758439db2426/ignite/engine/engine.py#L497-L500

 Idea is to put these times (at least `time_taken` for epoch and total time) into `engine.state` under a name (e.g. "times"). Such that we can also use this information without need to use `Timer` or setup logger.

erip:
Fixes #916

Description: Adds timing info for epoch completion and run completion to the engine's state.

Check list:

I'm feeling a bit sketchy about the tests, so I'd love another pair of eyes on them.

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -776,7 +776,7 @@ class Engine:
         le = self._dataloader_len if self._dataloader_len is not None else 1
         self._manual_seed(self.state.seed, self.state.iteration // le)
 
-    def _internal_run(self) -> State:        
+    def _internal_run(self) -> State:
         self.should_terminate = self.should_terminate_single_epoch = False
         self._init_timers(self.state)
         try:
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -12,7 +12,6 @@ import torch
 
 from ignite.engine.events import Events, State, CallableEventWithFilter, RemovableEventHandle, EventsList
 from ignite.engine.utils import ReproducibleBatchSampler, _update_dataloader, _check_signature
-from ignite._utils import _to_hours_mins_secs
 
 __all__ = ["Engine"]
 
@@ -433,7 +432,7 @@ class Engine:
         )
         self.should_terminate_single_epoch = True
 
-    def _run_once_on_dataset(self) -> Tuple[int, int, int]:
+    def _run_once_on_dataset(self) -> float:
         start_time = time.time()
 
         # We need to setup iter_counter > 0 if we resume from an iteration
@@ -497,10 +496,7 @@ class Engine:
             self.logger.error("Current run is terminating due to exception: %s.", str(e))
             self._handle_exception(e)
 
-        time_taken = time.time() - start_time
-        hours, mins, secs = _to_hours_mins_secs(time_taken)
-
-        return hours, mins, secs
+        return time.time() - start_time
 
     def _handle_exception(self, e: Exception) -> None:
         if Events.EXCEPTION_RAISED in self._event_handlers:
@@ -785,17 +781,15 @@ class Engine:
                 if self._dataloader_iter is None:
                     self._setup_engine()
 
-                hours, mins, secs = self._run_once_on_dataset()
-
-                self.logger.info("Epoch[%s] Complete. Time taken: %02d:%02d:%02d", self.state.epoch, hours, mins, secs)
+                time_taken = self._run_once_on_dataset()
+                self.state.times[Events.EPOCH_COMPLETED.name] = time_taken
                 if self.should_terminate:
                     break
                 self._fire_event(Events.EPOCH_COMPLETED)
 
-            self._fire_event(Events.COMPLETED)
             time_taken = time.time() - start_time
-            hours, mins, secs = _to_hours_mins_secs(time_taken)
-            self.logger.info("Engine run complete. Time taken %02d:%02d:%02d" % (hours, mins, secs))
+            self.state.times[Events.COMPLETED.name] = time_taken
+            self._fire_event(Events.COMPLETED)
 
         except BaseException as e:
             self._dataloader_iter = self._dataloader_len = None
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -616,7 +616,8 @@ class Engine:
 
         Engine has a state and the following logic is applied in this function:
 
-        - At the first call, new state is defined by `max_epochs`, `epoch_length`, `seed` if provided.
+        - At the first call, new state is defined by `max_epochs`, `epoch_length`, `seed` if provided. A timer for
+            total and per-epoch time is initialized when Events.STARTED is handled.
         - If state is already defined such that there are iterations to run until `max_epochs` and no input arguments
             provided, state is kept and used in the function.
         - If state is defined and engine is "done" (no iterations to run until `max_epochs`), a new state is defined.
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -685,7 +685,15 @@ class Engine:
             )
 
         self.state.dataloader = data
+        self.add_event_handler(Events.STARTED, self._init_timers)
         return self._internal_run()
+
+
+    @staticmethod
+    def _init_timers(engine: "Engine"):
+        engine.state.times[Events.EPOCH_COMPLETED.name] = 0.0
+        engine.state.times[Events.COMPLETED.name] = 0.0
+
 
     def _setup_engine(self) -> None:
 
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -686,13 +686,12 @@ class Engine:
             )
 
         self.state.dataloader = data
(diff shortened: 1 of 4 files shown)


tests/ignite/engine/test_engine.py
```diff
@@ -366,6 +367,40 @@ def test_state_get_event_attrib_value():
...
+
+        num_examples = len(data)
+
+        engine.run(data, max_epochs=max_epochs, epoch_length=epoch_length)
+
+        lower_bound_time = num_examples / epoch_length / max_epochs
+        upper_bound_time = 1.1 * lower_bound_time
+
+        def init(engine):
+            engine.state.times = {Events.EPOCH_COMPLETED.name: 0.0, Events.COMPLETED.name: 0.0}
```
vfdev-5, line 383: I think we should attach this automatically without explicitly require the user

vfdev-5:
@erip I think we should keep previous behavious about logging h,m,s times.

About the tests, I think we can just check the presence of values.

erip:
You want to log the h,m,s times in the engine, but store the timestamp?

vfdev-5:
> You want to log the h,m,s times in the engine, but store the timestamp?

Yes, I want to keep previous engine's behaviour and in addition to store the timestamp.

tests/ignite/engine/test_engine.py
erip: Excellent. This initialization happens in the `State` constructor, but I forgot to remove this from the test; that said, I think that's slightly different that wiring up the event handler.

ignite/engine/engine.py
```diff
@@ -433,7 +433,7 @@ def terminate_epoch(self) -> None:
         )
         self.should_terminate_single_epoch = True
 
-    def _run_once_on_dataset(self) -> Tuple[int, int, int]:
+    def _run_once_on_dataset(self) -> float:
```
vfdev-5, line 436: @erip could you please revert this

tests/ignite/engine/test_engine.py
erip: Is there some place where it makes sense to wire this or is `engine.run` good enough?

ignite/engine/engine.py
```diff
@@ -497,10 +497,7 @@ def _run_once_on_dataset(self) -> Tuple[int, int, int]:
             self.logger.error("Current run is terminating due to exception: %s.", str(e))
             self._handle_exception(e)
 
-        time_taken = time.time() - start_time
-        hours, mins, secs = _to_hours_mins_secs(time_taken)
-
-        return hours, mins, secs
+        return time.time() - start_time
```
vfdev-5, line 500: and here, please revert this too. And we are almost done.

ignite/engine/engine.py
erip: If this is reverted, we'll need to unparse the tuple to a timestamp unless we want to put the state-modifying code in the `_run_once_on_dataset`.

vfdev-5: Yes, you are right ! I didn't get it completely, sorry.

erip: Oh, so I should keep it as is with `-> float`? Since it's an internal method it shouldn't break anything (or anyone).

vfdev-5:
Looking into this code now:
```python
                time_taken = self._run_once_on_dataset()
                self.state.times[Events.EPOCH_COMPLETED.name] = time_taken
                self.logger.info("Epoch[%s] Complete. Time taken: %02d:%02d:%02d", self.state.epoch, hours, mins, secs)
                if self.should_terminate:
                    break
```
In case if `should_terminate` is True, we log and set the time of a completed epoch, but we do not trigger the event... I need to think if this is OK or we need to trigger the event...

vfdev-5:
> Excellent. This initialization happens in the State constructor, but I forgot to remove this from the test; that said, I think that's slightly different than wiring up the event handler.

> Is there some place where it makes sense to wire this or is engine.run good enough?

Good point, we need to check what happens if we run multiple times evaluator and whether times are still correct. Do we need to include that into state dict (maybe not) ?

vfdev-5:
And @erip I think it is déjàvu, are these values good for a single process and for DDP multi-process ?

erip:
>we need to check what happens if we run multiple times 

Ok, I'll add another test. :)

> Do we need to include that into state dict?

Currently no. It's unclear if these would be useful in the state info. If they were, I imagine it would just be simple enough to (de)serialize the values.

ignite/engine/events.py
```diff
@@ -270,6 +270,7 @@ def __init__(self, **kwargs):
         self.metrics = {}
         self.dataloader = None
         self.seed = None
+        self.times = {Events.EPOCH_COMPLETED.name: 0.0, Events.COMPLETED.name: 0.0}
```
vfdev-5, line 273: Maybe, we need to init values as `None`, such that if, for example, user request values of `Events.COMPLETED.name` it wont say that run took 0.0 seconds... Thoughts ?

tests/ignite/engine/test_engine.py
```diff
@@ -366,6 +367,34 @@ def test_state_get_event_attrib_value():
...
+        num_examples = len(data)
+
+        engine.run(data, max_epochs=max_epochs, epoch_length=epoch_length)
+
+        lower_bound_time = num_examples / epoch_length / max_epochs
+        upper_bound_time = 1.1 * lower_bound_time
+
+        def check_epoch_time(engine):
+            # TODO: can we make this test more meaningful?
+            assert engine.state.times[Events.EPOCH_COMPLETED.name] != 0
```
vfdev-5, line 384:
```suggestion
            assert engine.state.times[Events.EPOCH_COMPLETED.name] >= 0.01 * epoch_length 
```
something like that ?

tests/ignite/engine/test_engine.py
```diff
@@ -366,6 +367,34 @@ def test_state_get_event_attrib_value():
...
+def test_time_stored_in_state():
+    def _test(data, max_epochs, epoch_length):
+
+        engine = Engine(lambda e, b: time.sleep(0.01))
+
+        num_examples = len(data)
+
+        engine.run(data, max_epochs=max_epochs, epoch_length=epoch_length)
+
+        lower_bound_time = num_examples / epoch_length / max_epochs
```
vfdev-5, line 379: these two lines are unused

tests/ignite/engine/test_engine.py
```diff
@@ -366,6 +367,34 @@ def test_state_get_event_attrib_value():
...
+        upper_bound_time = 1.1 * lower_bound_time
+
+        def check_epoch_time(engine):
+            # TODO: can we make this test more meaningful?
+            assert engine.state.times[Events.EPOCH_COMPLETED.name] != 0
+
+        def check_completed_time(engine):
+            # TODO: can we make this test more meaningful?
+            assert engine.state.times[Events.COMPLETED.name] != 0
```
vfdev-5, line 388:
```suggestion
            assert engine.state.times[Events.COMPLETED.name] >= 0.01 * epoch_length * max_epochs
```

ignite/engine/events.py
erip: I agree. This makes the wiring more intentional, too.

tests/ignite/engine/test_engine.py
erip: They are? I thought the top line was driving the attached handlers which make assertions.

vfdev-5:
@erip thanks for the updates! Could you please update the docstring of `Engine` and show the usage in `examples/mnist/mnist.py`.

erip:
Absolutely

erip:
@vfdev-5 thanks for catching the missed TODO comments. Please let me know if there's anything else you want me to polish

vfdev-5:
@erip I thinks we are good with this one. Could you, please, take a look at this issue : https://github.com/pytorch/ignite/issues/784
 with `Frequency` you worked previously ?

erip:
Are the tests normally this flaky? The failure seems unrelated, but I've seen the same test fail a few times.

vfdev-5:
@erip OK, I understand now why tests are failing. In FastaiLRFiner tests we check that there is no more attached handlers, by default `Engine` has no internally attached handlers
```
    def test_with_attach(lr_finder, to_save, dummy_engine, dataloader):

        with lr_finder.attach(dummy_engine, to_save=to_save) as trainer_with_finder:

        assert lr_finder.get_results() is not None
        for event in dummy_engine._event_handlers:
>           assert len(dummy_engine._event_handlers[event]) == 0
```
and here we added `_init_timers` on started. Maybe we can call this method inside `run` and keep previous behaviour.

PR #967, opened 2020-04-22
Improve epoch_metric from O(N^2) --> O(N)

hoangmit:
The current method is O(N^2) because it create a new copy _predictions every append.

This improves it to O(N)


vfdev-5:
Thanks for the PR @hoangmit and for pointing out that ! Let me check about `O(N^2)`.

ignite/metrics/epoch_metric.py
```diff
@@ -68,21 +68,20 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
         if y.ndimension() == 2 and y.shape[1] == 1:
             y = y.squeeze(dim=-1)
 
-        y_pred = y_pred.to(self._predictions)
-        y = y.to(self._targets)
-
-        self._predictions = torch.cat([self._predictions, y_pred], dim=0)
-        self._targets = torch.cat([self._targets, y], dim=0)
+        self._predictions.append(y_pred.clone())
```
vfdev-5, line 71: At first, we need to pass to CPU for both `y_pred` and `y`, otherwise we can overflow GPU's memory.
I'm also now think about whether we need to call `y_pred.detach()` to clearly stop the graph etc...

hoangmit: > EDIT: it seems like CI is not happy with the changes.

From the error msg, it still take "self._predictions" as Tensor. That puzzled me.

sdesrozis: I don’t catch why it’s O(N^2) and why the implementation with append is better than cat. I would think that in both cases, memory should be allocated to receive the concat of existing array and new data. To me m, it’s more or less the same.

I must miss something but don’t know what.

hoangmit: @sdesrozis: The original algorithm is O(N^2) because:
torch.cat([A, B]) returns a new tensor instead of copying B to the end of A. Thus, it needs to allocate memory and iterate the contents of A every time. It is the same as the following loop:
```
for i in range(N):
    for j in range(i):
        # read data in A[j]
```
I noticed this because the running time of EpochMetric is multiple fold longer than the forward pass of the same data (even with dummy metric that only return length). And, it is worse when I reduce the batch_size.

For the new proposed approach, I believe torch.cat([...]) only allocates one time, and iterate the content of each tensor once.

sdesrozis: The advantage of tensors over list is to be in GPU memory.

hoangmit: Yes, it depends a lot on the implementation of torch.cat(). Let me find someone with the knowledge to review. In addition, we actually can benchmark this with big data.

sdesrozis: Related discussions on PyTorch forum : 
- https://discuss.pytorch.org/t/appending-in-pytorch/39313
- https://discuss.pytorch.org/t/appending-to-a-tensor/2665/3

Torch.cat is a good solution. To improve performance, we could preallocate as suggested. But I agree, we need to bench.

sdesrozis: Ok I misunderstood the implementation !! 

You do a list of tensors and use torch.cat only one time. Sorry for my reading...

ignite/metrics/epoch_metric.py
```diff
@@ -68,21 +68,20 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
         if y.ndimension() == 2 and y.shape[1] == 1:
             y = y.squeeze(dim=-1)
 
-        y_pred = y_pred.to(self._predictions)
-        y = y.to(self._targets)
-
-        self._predictions = torch.cat([self._predictions, y_pred], dim=0)
-        self._targets = torch.cat([self._targets, y], dim=0)
+        self._predictions.append(y_pred.detach())
```
albanD, line 71: Note that this changes the behavior.
The old code just cast the Tensors to the right dtype. While your versions does not and adds an extra `.detach()` that is not in the original code.

This is most likely the reason for the CI failures.

vfdev-5: @albanD thanks for the review !

ignite/metrics/epoch_metric.py
```diff
@@ -68,21 +68,20 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
         if y.ndimension() == 2 and y.shape[1] == 1:
             y = y.squeeze(dim=-1)
 
-        y_pred = y_pred.to(self._predictions)
-        y = y.to(self._targets)
-
-        self._predictions = torch.cat([self._predictions, y_pred], dim=0)
-        self._targets = torch.cat([self._targets, y], dim=0)
+        self._predictions.append(y_pred.detach().to(dtype=torch.float32, device="cpu"))
+        self._targets.append(y.detach().to(dtype=torch.long, device="cpu"))
```
vfdev-5, line 72: No need to detach for the targets:
```suggestion
        self._targets.append(y.to(dtype=torch.long, device="cpu"))
```

vfdev-5:
Seems like `EpochMetric` implementation and a base class for regression metrics can be unified: 
ignite/contrib/metrics/regression/_base.py

hoangmit:
Fix the tests:
* It turns out that regression metrics have their own version of EpochMetric
* The epoch metric's unit tests read the private variables for inspection

ignite/metrics/epoch_metric.py
hoangmit: Do we need to detach() the prediction? I am afraid some model may inplace update the tensor.

albanD: Note that .detach() does not give you new memory. If the original Tensor is modified, you will see the update.
.detach() only breaks the computational graph and prevent gradient from flowing.

If you want new memory to avoid side effects by someone modifying the other Tensor inplace, you need to use .clone().

vfdev-5: Yes, thanks @albanD, we need to apply `.detach().clone()` on predictions and `.clone()` on targets, such that we do not depend on any inplace updates on predictions or targets.

ignite/contrib/metrics/regression/_base.py
```diff
@@ -47,20 +47,18 @@ def __init__(self, compute_fn, output_transform=lambda x: x):
         EpochMetric.__init__(self, compute_fn=compute_fn, output_transform=output_transform)
 
     def reset(self):
-        self._predictions = torch.tensor([], dtype=torch.float32)
-        self._targets = torch.tensor([], dtype=torch.float32)
+        self._predictions = []
```
vfdev-5, line 50: @hoangmit now we can remove `reset()` method as it is exactly the same as in `EpochMetric`.

ignite/metrics/epoch_metric.py
```diff
@@ -46,8 +46,10 @@ def __init__(self, compute_fn: Callable, output_transform: Callable = lambda x:
         self.compute_fn = compute_fn
 
     def reset(self) -> None:
-        self._predictions = torch.tensor([], dtype=torch.float32)
-        self._targets = torch.tensor([], dtype=torch.long)
+        self._predictions = []
+        self._prediction_tensor = None
```
vfdev-5, line 50: I think we can avoid double storage of predictions and targets by removing `self._prediction_tensor ` and `self._target_tensor` and adapting impacted tests. What do you think ?

hoangmit:
Changes:
* Only keep one copy + update the test case
* Update regression + manual merge with @sdesrozis

vfdev-5:
> Update regression + manual merge with @sdesrozis branch

@hoangmit his branch is rather experimental and idea is also to relax dtype condition: float32 or long. With list of tensor we can even avoid casting to the type. 
If user mixes types, `torch.cat` will fail, maybe this should be supported in some way... (anyway this is another point).
So, let's keep for instance: preds as float32 and targets as long. And @sdesrozis 's PR will handle datatypes.

Issue #959, opened 2020-04-22
Transforming metric values


isolet:
## ❓ Questions/Help/Support

The [document](https://pytorch.org/ignite/metrics.html#ignite.metrics.Metric) of `Metircs` says that return value of `compute` can be `Any`. So I'm trying to do this in a single pass evaluation over the whole validation data loader.

```py


class SuperMetrics(Metric):

  def __init__(self, num_labels, output_transform=lambda x: x, device=None):
      self.num_labels = num_labels
      self._y = None
      self._y_pred = None
      self._num_drugs = None
      super(SuperMetrics, self).__init__(output_transform=output_transform,
                                         device=device)

  def compute_metrics(self, y, y_pred):  # pylint: disable=no-self-use
      y = y.toarray()
      y_pred = y_pred.toarray()
      y[y > 0] = 1
      y_pred[y_pred > 0] = 1

      hamming_loss = metrics.hamming_loss(y, y_pred)
      macro_f1 = metrics.f1_score(y, y_pred, average='macro')
      macro_precision = metrics.precision_score(y, y_pred, average='macro')
      macro_recall = metrics.recall_score(y, y_pred, average='macro')
      micro_f1 = metrics.f1_score(y, y_pred, average='micro')
      micro_precision = metrics.precision_score(y, y_pred, average='micro')
      micro_recall = metrics.recall_score(y, y_pred, average='micro')

      return {
          'hamming_loss': hamming_loss,
          'macro_f1': macro_f1,
          'macro_precision': macro_precision,
          'macro_recall': macro_recall,
          'micro_f1': micro_f1,
          'micro_precision': micro_precision,
          'micro_recall': micro_recall
      }

  @reinit__is_reduced
  def reset(self):
      self._y = []
      self._y_pred = []
      self._num_drugs = []
      super(SuperMetrics, self).reset()

  @reinit__is_reduced
  def update(self, output):
      y, y_pred, num_drugs = output
      self._y += y
      self._y_pred += y_pred
      self._num_drugs += num_drugs

  @sync_all_reduce('_y', '_y_pred', '_num_drugs')
  def compute(self):
      num_examples = len(self._num_drugs)

      rows = []
      y_columns, y_pred_columns = [], []
      for i, (y_sample, y_pred_sample, num_drug_sample) in enumerate(
              zip(self._y, self._y_pred, self._num_drugs)):
          rows += [i] * (num_drug_sample - 2)
[...]

PR #968, opened 2020-04-23
Add support for nested metric values. (#959)

isolet:
Description:

This commit add support for nested metric values. If `Metric.compute` returns a `dict`, its values will be flattened into `engine.state.metrics` in `Metric.completed`. 

Note that, in this case the value types of the returned `dict` is limited(i.e. `Dict[Any, non number type]` is not allowed), while the original allowed type of `Metric.compute` is `Any`. We can also allow other types of `dict` values, but this may be lead unclear behaviors to users when they try to return (different kinds of )`dict`s.

Check list:


ignite/metrics/metric.py
```diff
@@ -114,6 +116,19 @@ def _sync_all_reduce(self, tensor: Union[torch.Tensor, numbers.Number]) -> Union
             return tensor.item()
         return tensor
 
+    def ensure_metric_value(self, value: Any, raise_error=True) -> Union[numbers.Number, torch.Tensor]:
```
sdesrozis, line 119: I don’t know if an exception should be raised. Metrics are very generics, it can work with ˋAny`. Metrics associated to loggers should use restricted types. 

@vfdev-5 ?

isolet: I agree metrics are generics. Maybe remove type checking here and move to `Logger` later when necessary?

sdesrozis: Ok, let’s try this!

isolet: I just found one more issue, if `Any` type is allowed, shall we deep flatten a dict in this case? Or shall me only care about `dict` contains all number values?

vfdev-5: IMO, we should not check the output type and raise an error and just keep previous behaviour.

ignite/metrics/metric.py
```diff
@@ -137,9 +152,12 @@ def iteration_completed(self, engine: Engine) -> None:
 
     def completed(self, engine: Engine, name: str) -> None:
         result = self.compute()
-        if torch.is_tensor(result) and len(result.shape) == 0:
-            result = result.item()
-        engine.state.metrics[name] = result
+        if isinstance(result, dict):
```
vfdev-5, line 155:
```suggestion
        if isinstance(result, Mapping):
```

vfdev-5:
IMO, idea of the PR is to unnest one level of mappings (e.g. dict) as @isolet's use-case. Thoughts ?

ignite/metrics/metric.py
sdesrozis: I agree. At the moment, we should flat only the result of metrics. IMHO, what is embedded in dict is related to the metric itself.

We would notice in doc that generic result type ˋAny` is a generic purpose but interoperability with others handlers / features implies some restrictions : scalars (number, tensor) and dict of scalars.

What do you think ?

isolet:
My original thought is single enough just to flatten a level one dict of metrics. But as metrics is so generic, now I think maybe we should also not to make level one dict anything special? Before I opened the issue, I used the following snippet to get my goal.

```py


class MultipleMetricsOutputHandler(OutputHandler):

  def _setup_output_metrics(self, engine):
      metrics = {}
      if self.metric_names is not None:
          if (isinstance(self.metric_names, str)
                  and self.metric_names == 'all'):
              metrics = flatten_dict(engine.state.metrics)
          else:
              for name in self.metric_names:
                  if name not in engine.state.metrics:
                      warnings.warn('Provided metric name \'{}\' is missing '
                                    'in engine\'s state metrics: {}'.format(
                                        name,
                                        list(engine.state.metrics.keys())))
                      continue
                  metrics[name] = engine.state.metrics[name]

      if self.output_transform is not None:
          output_dict = self.output_transform(engine.state.output)

          if not isinstance(output_dict, dict):
              output_dict = {'output': output_dict}

          metrics.update(
              {name: value
               for name, value in output_dict.items()})

      return metrics
```

How about keeping anything of `Metric` unchanged? Is there a better place to achieve my goal? I'm not familiar enough with `ignite` so I can not get an answer now ...

vfdev-5:
@isolet I think one level flattening is a good level of simplifying things and keeping Metric rather generic. For other non-trivial cases, use should write its adapters.

ignite/metrics/metric.py
```diff
@@ -82,7 +82,9 @@ def compute(self) -> Any:
         This is called at the end of each epoch.
 
         Returns:
-            Any: the actual quantity of interest.
+            Any: the actual quantity of interest. However, if a :class:`dict` is returned, it should contain
+                 :class:`numbers.Number` or :class:`torch.tensor` values and will flattened into
```
vfdev-5, line 86: Let's remove those type limitations:
```
Any: the actual quantity of interest. However, if a :class:`dict` is returned, it will flattened into  `engine.state.metrics` when :func:`~ignite.metrics.Metric.completed` is called.
```

isolet: Done!

sdesrozis:
LGTM but CI quite unstable...

isolet:
Thanks!

PR #970, opened 2020-04-23
Merge of EpochMetric in _BaseRegressionEpoch

sdesrozis:
Related to #967 

Description:

Merge `EpochMetric` in `_BaseRegressionEpoch`.

Introduction of a parameter to define the usage of metric : "regression" or "classification"

Check list:

### DIFF
--- a/ignite/contrib/metrics/regression/_base.py
+++ b/ignite/contrib/metrics/regression/_base.py
@@ -46,21 +46,15 @@ class _BaseRegressionEpoch(_BaseRegression, EpochMetric):
     def __init__(self, compute_fn, output_transform=lambda x: x):
         EpochMetric.__init__(self, compute_fn=compute_fn, output_transform=output_transform)
 
-    def reset(self):
-        self._predictions = torch.tensor([], dtype=torch.float32)
-        self._targets = torch.tensor([], dtype=torch.float32)
-
     def _update(self, output):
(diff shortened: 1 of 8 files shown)


vfdev-5:
@sdesrozis could you please review ?

vfdev-5:
OK, `_BaseRegressionEpoch` inherits from `_BaseRegression` in order to benefit of `_check_shape` and `_check_type` methods (later I forgot to put in `_BaseRegression`). We may either define those methods externally and call them separately in each class, so that `_BaseRegressionEpoch` wont inherit from `_BaseRegression`. Another way is to define another base class for this two methods and make everything like
```
class _BaseBaseRegression:
    # define _check_shape, _check_type

class _BaseRegression(_BaseBaseRegression, Metric):
    # ...

class _BaseRegressionEpoch(_BaseBaseRegression, EpochMetric):
    # ...
```
and no diamonds.

sdesrozis:
Aggregation means use objects in class (« has a » concept) rather than inheritance (« is a » concept). Recall that I talk about code factorisation. Using external methods to share code is on the same line than aggregation.

I agree the 2 alternatives you listed, and I prefer to do not use inheritance here.

vfdev-5:
@sdesrozis please take a look

sdesrozis:
Perfect !!

PR #976, opened 2020-04-25
fix #784

### LINKED ISSUE #784: Improve Frequency

erip:
Fixes #784

Description: Explicitly wires `self.iteration_completed` as the handler for `Events.ITERATION_COMPLETED` instead of the user-provided one, which handles edge cases surrounding `Events`-specific kwargs.

~TODO: write some good tests. 😄~

Check list:


erip:
Hopefully I understood the issue correctly - the tests seem to be OK. If these tests look representative, I think they can (and should) be refactored to generalize the original frequency tests w/ the full engine setup.

tests/ignite/metrics/test_frequency.py
```diff
@@ -57,3 +57,45 @@ def test_frequency_with_engine():
...
+
+    estimated_wps = batch_size * workers / artificial_time
+
+    def update_fn(engine, batch):
+        time.sleep(artificial_time)
+        return {"ntokens": len(batch)}
+
+    engine = Engine(update_fn)
+    wps_metric = Frequency(output_transform=lambda x: x["ntokens"], device=device)
+    wps_metric.attach(engine, "wps")
```
vfdev-5, line 76: Shouldn't this be `wps_metric.attach(engine, "wps", event_name=Events.ITERATION_COMPLETED(every=every))` ?

vfdev-5:
Yes, you are right, we can refactor original and new tests into a single code. 

And there is a docstring update.

erip:
This looks good now. 😄 Which docstring should be updated? Just adding another example in the `Frequency` docstring?

tests/ignite/metrics/test_frequency.py
```diff
@@ -35,9 +35,10 @@ def update_fn(engine, batch):
 
     engine = Engine(update_fn)
     wps_metric = Frequency(output_transform=lambda x: x["ntokens"], device=device)
-    wps_metric.attach(engine, "wps")
+    event = Events.ITERATION_COMPLETED if every == 1 else Events.ITERATION_COMPLETED(every=every)
```
vfdev-5, line 38: Maybe we can just use `Events.ITERATION_COMPLETED(every=every)` ? It works now with `every=1`.

erip: Is it the case that `Events.ITERATION_COMPLETED(every=1)` is equivalent to `Events.ITERATION_COMPLETED`? For some reason I thought there was some nuanced difference... maybe not, though.

vfdev-5:
> Just adding another example in the Frequency docstring?

Yes, another example with the explanation of what it does :)

tests/ignite/metrics/test_frequency.py
erip: Ah - just checked the `CallableEventWithFilter.__call__`; it seems like they're forced to equivalent explicitly. Will update the test.

vfdev-5: Now `Events.ITERATION_COMPLETED(every=1) == Events.ITERATION_COMPLETED`.
https://github.com/pytorch/ignite/blob/master/ignite/engine/events.py#L78

vfdev-5:
Elijah, I was also wondering how's going integrations of training examples of fairseq with ignite you were working on previously ?

erip:
It's going slowly, unfortunately. Fairseq has been working on moving from a framework to a library and it's made development outside of adding new architecture slow. I have a handful of pending PRs that would hopefully move the needle a bit, but there hasn't been much movement.

vfdev-5:
I see, thanks for the info. Yes, maybe also they do not have a point to use such libs like ignite (and introduce it as a dependency)...

Are you at PyTorch slack where we could discuss in more fluent way ?

erip:
@vfdev-5 I just requested access. :-)

vfdev-5:
> @vfdev-5 I just requested access. :-)

Sounds good! If you have no answer from them within 1-2 days, please give me your email, I can try to make a request directly from the application (and it can be accepted more quickly :)

Issue #848, opened 2020-03-24
Set alpha value to 0 for RunningAverage


VinhLoiIT:
## ❓ Questions/Help/Support
When I look at the source code of `RunningAverage` class, I see that there is a factor variable, `alpha`, which is set to `0.98` as default. As far as I know, this is to smooth the value of a metric like this:
```
def compute(self) -> Union[torch.Tensor, float]:
    if self._value is None:
        self._value = self._get_src_value()
    else:
        self._value = self._value * self.alpha + (1.0 - self.alpha) * self._get_src_value()
    return self._value
```

The use case is that I want to log my metrics during validation time (using `contrib.metrics.ProgressBar`), these metrics should not be smoothed, however. One more case is that the smoothed value when plotting to Tensorboard is smoothed once again, which leads to quite unreliable the plot. By disabling the smooth factor, I set `alpha` to 0 and it raised an exception:
```
if not (0.0 < alpha <= 1.0):
    raise ValueError("Argument alpha should be a float between 0.0 and 1.0.")
```

I wonder why it is not `0.0 <= alpha <= 1.0` ? Is it a mistake or any idea behind it?

vfdev-5:
@VinhLoiIT `RunningAverage` is mostly used to average values during the training: batch loss, etc
When setting alpha to 0, there is no point to use `RunningAverage`, so just skip it.

> The use case is that I want to log my metrics during validation time (using contrib.metrics.ProgressBar)

Could you please detail your use case, such I could help better with the problem.



PS: 
> One more case is that the smoothed value when plotting to Tensorboard is smoothed once again

It is also possible to remove TB smoothing on the UI.

VinhLoiIT:
The use case is that I use `RunningAverage` while validating, because it will trigger my metric computation every iteration (which can then pass to the tqdm logger `ProgressBar` class), and it supports epoch bound. Therefore, by setting alpha to zero, I would have the raw value of my metric while iterating through the validation set but still have the correct average at the end of the phase

The problem with `EpochMetric` class is that it saves the whole output of the engine, or `Average` class does not handle the last batch right if `len(dataloader)` is not divisible to `batch_size`

vfdev-5:
> The use case is that I use RunningAverage while validating, because it will trigger my metric computation every iteration 

Normally, it should not trigger metric's computation.

OK, I think I understand your use-case, probably it would be better to do like that:
```
import time

import torch
from ignite.engine import Engine, Events
[...]

PR #979, opened 2020-04-25
Issue 874 : add custom events for Metrics

### LINKED ISSUE #874: Support custom events for metrics update
Currently for ignite metrics, `update` function is called per iteration. In scenarios like multitasking, there can be metrics for different tasks, and each iteration might only have outputs for a subset of tasks. In this situation, only the metrics associated with those tasks having outputs can/should be updated.

To make it work, I need to overwrite the `attach` function like this:
```
class CustomRnningAverage(RunningAverage):
    def attach(self, engine, name, custom_event=None):
        if custom_event is None:
            super().attach(engine, name)
        else:
            if self.epoch_bound:
                # restart average every epoch
            # compute metric
            engine.add_event_handler(custom_event, self.iteration_completed)
            # apply running average
            engine.add_event_handler(custom_event, self.completed, name)
```
This is doable, but would be nice to have it by default in ignite. Also, this manual overwrite doesn't work with DDP. The cause is related to `_internal_attach` in `metrics_lambda.py`.

I would like to propose
1.  Have `update` every iteration by default
2. Support overwrite when to `update` by accepting custom events
3. Make it work with DDP

Multitasking is one case where this can be very helpful. But I am sure there will be other use cases.

sdesrozis:
Fixes #874 

Description:

Allow customization for `Events` of `Metrics`.

Check list:


sdesrozis:
@vfdev-5 thoughts on this draft ?

vfdev-5:
We may have to take this comment into account for the usage : https://github.com/pytorch/ignite/pull/644#issuecomment-576557938

sdesrozis:
> We may have to take this comment into account for the usage : https://github.com/pytorch/ignite/pull/644#issuecomment-576557938

Same issue??

sdesrozis:
@vfdev-5 

Major refactoring
- Add interface `EngineMetric` which defines methods used by `Engine`
- Add functions `attach`, `detach` and `is_attached` external to `Metric` using `Engine`, `MetricUsage` and `EngineMetric`

The idea is to have `Metric` like an algorithm pluggable in any `Engine` wrt a `MetricUsage`. Today, `attach`, `detach` and `is_attached` are embedded in `Metric` but one could use external functions.

Thoughts ?

vfdev-5:
@sdesrozis please, show an example of usage. For instance, I do not understand the need of another interface... Thanks

sdesrozis:
The idea is to have `Metric` like an algorithm with API to be plugged to engine but no reason to have the plug in the metric. 

This is a separation of concerns.

vfdev-5:
@sdesrozis I'm not fan of "possible API", let's keep current API. 
If I get it right, `EngineMetric` interface is only for "possible API" ?

sdesrozis:
> @sdesrozis I'm not fan of "possible API", let's keep current API. 
> 
> If I get it right, `EngineMetric` interface is only for "possible API" ?
> 
> 

Yes, I can remove it, no problem 😊

sdesrozis:
Removed

vfdev-5:
@sdesrozis thanks ! What do you think about to enhance the following API by 
```python
engine1 = ...
engine2 = ...

m = MyMetric()

# current API
m.attach(engine1, "m")
m.attach(engine2, "m", usage="batch_wise")  # ~ BatchWise()
m.attach(engine2, "m", usage="epoch_wise")  # ~ EpochWise()
```
? 

No need to import those classes.

sdesrozis:
Note that `BatchFiltered` needs an argument.

sdesrozis:
Done !

Adding `MetricUsage` in `attach` rather than `__init__` implies that this feature is now usable by all metrics without any modification.

vfdev-5:
@snie2012 could you please take a look and give us your feedback if this PR covers your need in #874, thanks !

@VinhLoiIT could you also take a look on this PR if it covers your use-case to compute batch-wise metrics as mentioned in #848 ?

erip:
> Note that `BatchFiltered` needs an argument.

I'm not sure of a clean way to do this. There are three options for this:

1. A user provides a constructed `BatchFiltered` via `usage`.
1. A user provides `usage="batch_filter` and a kwarg predicate callable.
1. A user provides `usage=batch_filter` and not a kwarg predicate callable.

The logic for validating this is a bit hairy, I think. It would be something like...

```python
def attach(..., usage=..., filter: Optional[Callable] = None):
    if isinstance(usage, BatchFilter) and filter is not None:    
       raise ValueError("Can't provide a BatchFilter and a filter function.")
    elif usage == "batch_filter" and filter is None:
        raise ValueError("A filter function must be provided with \"batch_filter\".")
    elif usage == "batch_filter" and filter is not None:
        batch_filter = BatchFilter(filter)
    else:
        # handle metrics that don't require arguments here
        ... 
```

As more `MetricUsage`s are introduced that require arguments, this might become too messy. Maybe as a rule we say "if your `MetricUsage` requires arguments, its your responsibility to construct it and pass it in" -- WDYT @sdesrozis?

sdesrozis:
I prefer when things are simple 😊 I agree the last part, user is responsible if he does non standard usage (filtered or custom, because he can!) 😉

erip:
> I prefer when things are simple 😊 I agree the last part, user is responsible if he does non standard usage (filtered or custom, because he can!) 😉

I like that. It also doesn't open up `attach` more than necessary.

sdesrozis:
If needed, we can have `detach` and `is_attached` without usage by 1. testing usages (but what happen for custom usages) or removing from engine whatever the usage registered (carful for lambda...).

snie2012:
> @snie2012 could you please take a look and give us your feedback if this PR covers your need in #874, thanks !
> 
> @VinhLoiIT could you also take a look on this PR if it covers your use-case to compute batch-wise metrics as mentioned in #848 ?

Replied in the #874

VinhLoiIT:
Nice work!
I have two test cases mentioned in #848 to be tested:

1. The first one is that I would like to update my metric every batch so that it could be computed and logged with `ProgressBar` during evaluation. Your work solved peacefully by creating the `MetricUsage` instance:
```python
usage = MetricUsage(
    Events.EPOCH_STARTED,
    Events.ITERATION_COMPLETED,
    Events.ITERATION_COMPLETED,
)

MyMetric().attach(engine, 'MyMetric', usage=usage)
```

2. The second one is that I would like to reset my metric during training (similar to parameter `log_interval` in pure PyTorch print-style example. This case is useful in terms of calculating `Loss` (which is basically inherited from `Metric`) that it should be reset every `log_interval` while training to not be affected by averaging in the long run. I think it would better to use `RunningAverage` from ignite, but I have to add an event to reset by myself. To solve this by `MetricUsage`, it is equivalent to:
```python
usage = MetricUsage(
    Events.ITERATION_STARTED(every=log_interval), # started
    Events.ITERATION_COMPLETED,                   # completed
    Events.ITERATION_COMPLETED,                   # iteration_completed
)
```
However, it does not work as expected due to the exception:
```python
  ...
  File "/home/vinhloi/projects/fork/ignite/metrics/metric.py", line 278, in attach
    if not engine.has_event_handler(self.started, usage.STARTED):
  File "/home/vinhloi/projects/fork/ignite/engine/engine.py", line 303, in has_event_handler
    self._assert_non_filtered_event(event_name)
  File "/home/vinhloi/projects/fork/ignite/engine/engine.py", line 291, in _assert_non_filtered_event
    "Argument event_name should not be a filtered event, " "please use event without any event filtering"
TypeError: Argument event_name should not be a filtered event, please use event without any event filtering
```
I just wonder how to solve this problem. The `MetricUsage` as above logically true, I guess

sdesrozis:
@VinhLoiIT Thank for the report :)

For point 2. it's out of the scope of `MetricUsage` at the moment. There are too much scenarii of use and it's very easy to write bugs. The supported usages are `EpochWise`, `BatchWise` and `BatchFiltered`. We will warn about that and improve the document to be extremely clear about the scope. 

Btw, I will check your code to see if I can fix ;)

sdesrozis:
@vfdev-5 

filter can't apply to events of metrics.

See here https://github.com/pytorch/ignite/blob/c3c1ff22343d4f76979595ff7ac0707b49fc7c98/ignite/engine/engine.py#L285

sdesrozis:
@VinhLoiIT 

At the moment, we can't use filtered event with `MetricUsage`. However, it works with custom events. A workaround (ugly) could be
```python
# custom event for Events.ITERATION_COMPLETED(every=log_interval)
class MyEvent(EventEnum):
    ITERATION_STARTED = 'MyEvent_ITERATION_STARTED'

engine.register_events(*MyEvent)

usage = MetricUsage(
    started=MyEvent.ITERATION_STARTED, # use custom event
    completed=Events.ITERATION_COMPLETED,
    iteration_completed=Events.ITERATION_COMPLETED,
)

# to apply filter every log_interval
@engine.on(Events.ITERATION_COMPLETED(every=log_interval))
def _():
    engine.fire_event(MyEvent.IT_STARTED)
```

I can't do better :(

vfdev-5:
> @vfdev-5
> 
> filter can't apply to events of metrics.
> 
> See here
> 
> https://github.com/pytorch/ignite/blob/c3c1ff22343d4f76979595ff7ac0707b49fc7c98/ignite/engine/engine.py#L285

@sdesrozis can't we relax this condition ?

sdesrozis:
> @sdesrozis can't we relax this condition ?

We can do it but I don't have history about why this condition is enabled.

vfdev-5:
> > @sdesrozis can't we relax this condition ?
> 
> We can do it but I don't have history about why this condition is enabled.

This `_assert_non_filtered_event` was added before unification of `Events` into `EventsEnum` etc. Now, we must ensure that `has_event_handler` works correctly for filtered events, do not pick a wrong handlers instead etc and it can be enough.

sdesrozis:
I finish this PR this week-end. Sorry for the lag.

sdesrozis:
@vfdev-5 what do you think ?

ignite/engine/engine.py
```diff
@@ -300,7 +300,7 @@ def has_event_handler(self, handler: Callable, event_name: Optional[Any] = None)
                 to ``None`` to search all events.
         """
         if event_name is not None:
-            self._assert_non_filtered_event(event_name)
+            # self._assert_non_filtered_event(event_name)
```
vfdev-5, line 303: We can remove commented code. Have you checked the correct behaviour of `has_event_handler` on filtered events ?

ignite/engine/engine.py
```diff
@@ -327,7 +327,7 @@ def remove_event_handler(self, handler: Callable, event_name: Any):
             event_name: The event the handler attached to.
 
         """
-        self._assert_non_filtered_event(event_name)
+        # self._assert_non_filtered_event(event_name)
```
vfdev-5, line 330: Same here: remove commented code and if not yet done, we need a teston checking the correct behaviour with filtered events

ignite/engine/engine.py
sdesrozis: It seems correct. However I noticed that if a handler is registered with a filtered event, you can remove it using non filtered event (Ser test_custom_event.py)

sdesrozis:
Usage `epoch_wise` is well tested. Need use cases for others usages.

vfdev-5:
@sdesrozis code looks OK, documentation layout is not correct:
- missing docs for "MetricUsage", "EpochWise", "BatchWise", "BatchFiltered"
- missing indents renders badly docs, i'll comment out where.
Please, try to render the docs locally and check it.

ignite/metrics/metric.py
```diff
@@ -10,7 +10,86 @@
...
+    def ITERATION_COMPLETED(self):
+        return self.__iteration_completed
+
+
+class EpochWise(MetricUsage):
+    """
+    Epoch-wise usage of Metrics. It's the default and most common usage of metrics.
+
+    Metrics :
+    * starts to compute at Events.EPOCH_STARTED,
```
vfdev-5, line 46: I'm not sure if this would render correctly. To check.

ignite/metrics/metric.py
```diff
@@ -147,14 +232,28 @@ def completed(self, engine: Engine, name: str) -> None:
...
+    def attach(self, engine: Engine, name: str, usage: Union[str, MetricUsage] = EpochWise()) -> None:
         """
         Attaches current metric to provided engine. On the end of engine's run,
         `engine.state.metrics` dictionary will contain computed metric's value under provided name.
 
         Args:
             engine (Engine): the engine to which the metric must be attached
             name (str): the name of the metric to attach
+            usage (str or MetricUsage, optional): the usage of the metric. Valid string values should be
+            'epoch_wise' (default) or 'batch_wise'.
```
vfdev-5, line 256: Missing indent here

ignite/metrics/metric.py
```diff
@@ -182,6 +293,8 @@ def detach(self, engine: Engine) -> None:
 
         Args:
             engine (Engine): the engine from which the metric must be detached
+            usage (str or MetricUsage, optional): the usage of the metric. Valid string values should be
+            'epoch_wise' (default) or 'batch_wise'.
```
vfdev-5, line 297: Missing indent here

ignite/metrics/metric.py
```diff
@@ -194,23 +307,39 @@ def detach(self, engine: Engine) -> None:
...
+
+    def is_attached(self, engine: Engine, usage: Union[str, MetricUsage] = EpochWise()) -> bool:
         """
         Checks if current metric is attached to provided engine. If attached, metric's computed
         value is written to `engine.state.metrics` dictionary.
 
         Args:
             engine (Engine): the engine checked from which the metric should be attached
+            usage (str or MetricUsage, optional): the usage of the metric. Valid string values should be
+            'epoch_wise' (default) or 'batch_wise'.
```
vfdev-5, line 339: Missing indent here

sdesrozis:
@vfdev-5 thak you, I update soon :p

sdesrozis:
Doc done. Next, I would like to refactor a last thing after merge of fix for #1004 (if ok).

PR #984, opened 2020-04-26
Deprecate CustomPeriodicEvent

### LINKED ISSUE #983: Deprecate ignite.contrib.handlers.custom_events.CustomPeriodicEvent
## 🚀 Feature

Custom events `CustomPeriodicEvent` from contrib seem heavy and unusable. 

Idea is to 

- [ ] raise a warning about deprecation since v0.4.0 and removing since v0.5.0 and suggest to use filtered events.
- [ ] remove all docs about them

michiboo:
Fixes #983

Description:
add Deprecation warning for CustomPeriodicEvent class and remove doc related to it

Check list:

### DIFF
--- a/ignite/contrib/handlers/custom_events.py
+++ b/ignite/contrib/handlers/custom_events.py
@@ -51,10 +51,12 @@ class CustomPeriodicEvent:
     """
 
     def __init__(self, n_iterations=None, n_epochs=None):
-        
-        warnings.warn("CustomPeriodicEvent is deprecated since v0.4.0 and removing since v0.5.0"
-                      ", use filtered events instead", DeprecationWarning)
-        
+
+        warnings.warn(
+            "CustomPeriodicEvent is deprecated since v0.4.0 and removing since v0.5.0" ", use filtered events instead",
+            DeprecationWarning,
+        )
+
         if n_iterations is not None and (not isinstance(n_iterations, int) or n_iterations < 1):
             raise ValueError("Argument n_iterations should be positive integer number")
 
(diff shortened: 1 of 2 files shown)


vfdev-5:
@michiboo thanks for the PR. Please, fix flake and black issues: https://travis-ci.org/github/pytorch/ignite/jobs/679705447#L237
and we can merge it 👍 

To do that please see : https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md

vfdev-5:
@michiboo I think you need to run once `black .`, then add and commit modified file and it will be OK. Thanks !

erip:
Do we actually want to raise an exception? This will break a lot of code, I think. It might be better to _log_ a warning instead.

ignite/contrib/handlers/custom_events.py
```diff
@@ -51,8 +52,8 @@ def on_every_10_epochs(engine):
 
     def __init__(self, n_iterations=None, n_epochs=None):
         
-        raise DeprecationWarning("CustomPeriodicEvent is deprecated since v0.4.0 and"
-                                 "removing since v0.5.0, use filtered events instead")
+        warnings.warn("CustomPeriodicEvent is deprecated since v0.4.0 and removing since v0.5.0"
```
erip: Nit: `"CustomPeriodicEvent is deprecated since 0.4.0 and will be removed in 0.5.0. Use filtered events instead."`

erip:
This test seems hung. 🤔

vfdev-5:
Yes

> This test seems hung. 🤔

Yes, ddp tests on gloo hangs a lot since some time...

sdesrozis:
Arf barriers added don’t do the job 😔

PR #985, opened 2020-04-26
Issue 871

### LINKED ISSUE #871: Make epoch_length optional
## 🚀 Feature

Following discussions and the message: https://github.com/pytorch/ignite/issues/853#issuecomment-606732951 it could be useful to make `epoch_length` optional for finite iterators.

vfdev-5:
Fixes #871  

Description:
- updated FAQ 
- added FAQ's code into tests

Check list:

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -47,7 +47,7 @@ flexibility to the user to allow for this:
 
 .. code-block:: python
 
-    class BackpropEvents(Enum):
+    class BackpropEvents(EventEnum):
         """
         Events based on back propagation
         """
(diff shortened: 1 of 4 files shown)


snie2012:
Automatically determine the length of finite iterator of unknown length seems very neat! Thanks for adding it:)

docs/source/faq.rst
```diff
@@ -110,5 +110,239 @@ Based on `this blog article <https://medium.com/huggingface/training-larger-batc
...
+    def train_step(trainer, batch):
+        # ...
+        s = trainer.state
+        print(
+            "{}/{} : {} - {:.3f}".format(s.epoch, s.max_epochs, s.iteration, batch)
+        )
+
+    trainer = Engine(train_step)
+
+    @trainer.on(Events.ITERATION_COMPLETED(every=size))
```
sdesrozis, line 311: It could be the code of the previous section (using DATALOADER_STOP_ITERATION) ? 

Is it possible to restart iterator in backend using `epoch_lenght` for example ?

Btw, API is clear and simple.

vfdev-5: > It could be the code of the previous section (using DATALOADER_STOP_ITERATION) ?

yes, it can be, but here we show that as we know the lenght we can handle it explicitly.
I'll update the docs.

> Is it possible to restart iterator in backend using epoch_lenght for example ?

In general case, we do not know how to restart an iterator.

sdesrozis: I think it is not mandatory to update doc, cases are well splitted IMO :)

PR #988, opened 2020-04-27
[WIP] add TPU metrics reduction.

### LINKED ISSUE #965: Metrics reduction on distributed TPU setting
## 🚀 Feature

Ignite will support distributed training on TPU (e.g. #960). Currently, metric's computation is impacted in the same way as for DDP on GPUs.

Idea is to improve metric's computation and reduce internal values as it is done for DDP:
https://github.com/pytorch/ignite/blob/0b363974e93e2f79e468cf4a7851a149b96028a7/ignite/metrics/metric.py#L92

To check if we are running in distributed TPU, we can opt to
```
# global definition
try:
     on_xla_device = True
except ImportError:
     on_xla_device = False
```
and if we need to reduce:
```
xm.xrt_world_size() > 1
```

This issues depends on #963

erip:
Fixes #965 

Description: Adds TPU metrics all_reduce

~TODO: add tests~

Check list:


ignite/metrics/metric.py
```diff
@@ -91,14 +98,14 @@ def compute(self) -> Any:
...
-
+    def _do_reduction(
+        self,
+        tensor: Union[torch.Tensor, numbers.Number],
+        reduction: Callable[[Union[torch.Tensor, numbers.Number]], None],
+    ) -> Union[torch.Tensor, numbers.Number]:
         tensor_to_number = False
         if isinstance(tensor, numbers.Number):
-            tensor = torch.tensor(tensor, device=self._device)
+            tensor = torch.tensor(tensor, device=self._device, dtype=torch.float)
```
vfdev-5, line 108: Passing to `dtype=torch.float` is only necessary for TPU as they do not support `torch.long` and probably others.
Maybe we can wait until XLA fix it and remove casting

ignite/metrics/metric.py
```diff
@@ -108,14 +115,37 @@ def _sync_all_reduce(self, tensor: Union[torch.Tensor, numbers.Number]) -> Union
...
+        dist.all_reduce(tensor)
+
+    def _tpu_sync_all_reduce(self, tensor: Union[torch.Tensor, numbers.Number]) -> Union[torch.Tensor, numbers.Number]:
+        return self._do_reduction(tensor, self._tpu_reduce)
+
+    def _gpu_sync_all_reduce(self, tensor: Union[torch.Tensor, numbers.Number]) -> Union[torch.Tensor, numbers.Number]:
+        return self._do_reduction(tensor, self._gpu_reduce)
+
+    def _sync_all_reduce(self, tensor: Union[torch.Tensor, numbers.Number]) -> Union[torch.Tensor, numbers.Number]:
+        if on_xla_device:
```
vfdev-5, line 140: IMO, we need to test something equivalent to `dist.is_initialized()` which can be probably done by `xm.xrt_world_size() > 1`. This I checked on Colab.

ignite/metrics/metric.py
erip: In the interim I've added an optional dtype arg to the `_do_reduction` method which is only populated in the TPU case.

erip:
Should we wait for #992 to merge this? Since nothing here is user-facing we don't have to worry about breaking this, but it's probably best to hold off. WDYT @vfdev-5?

vfdev-5:
@erip we can refactor a bit current tests to accept `world_size` and `rank` as args and draft some TPU tests. In a PR for #992 we can rewrite the injection with new API.

tests/ignite/metrics/test_metric.py
```diff
@@ -602,37 +605,46 @@ def update(self, output):
     m.compute()
 
 
+@pytest.mark.tpu
+@pytest.mark.skipif(not on_xla_device or xm.xrt_world_size() < 1, reason="Skip if no TPU")
+def test_distrib_gpu():
```
vfdev-5, line 611:
```suggestion
def test_distrib_tpu():
```

erip: ugh, oops.

ignite/metrics/metric.py
```diff
@@ -108,14 +118,37 @@ def _sync_all_reduce(self, tensor: Union[torch.Tensor, numbers.Number]) -> Union
...
-        dist.all_reduce(tensor)
+        reduction(tensor)
 
         if tensor_to_number:
             return tensor.item()
         return tensor
 
+    @staticmethod
+    def _tpu_reduce(tensor: Union[torch.Tensor, numbers.Number]) -> None:
+        xm.all_reduce("sum", [tensor,], torch.float)
```
vfdev-5, line 129: I'm not sure about `torch.float` here. The API is a bit different
http://pytorch.org/xla/release/1.5/index.html#torch_xla.core.xla_model.all_reduce

erip: oops, this was a mistake when cleaning up the abstraction. that's supposed to happen in the number/tensor conversion.

erip:
It looks like the remaining tests are hanging (probably from normal flakiness). Does GitHub actions provide you with the option to rerun individual suites manually, @vfdev-5?

vfdev-5:
@erip I can only "cancel" and "re-run all jobs". It should be distributed tests...

ignite/metrics/metric.py
```diff
@@ -108,14 +118,37 @@ def _sync_all_reduce(self, tensor: Union[torch.Tensor, numbers.Number]) -> Union
...
+    def _gpu_sync_all_reduce(self, tensor: Union[torch.Tensor, numbers.Number]) -> Union[torch.Tensor, numbers.Number]:
+        return self._do_reduction(tensor, self._gpu_reduce)
+
+    def _sync_all_reduce(self, tensor: Union[torch.Tensor, numbers.Number]) -> Union[torch.Tensor, numbers.Number]:
+        if not (dist.is_available() and dist.is_initialized()) and not xrt_world_size > 0:
+            # Nothing to reduce
+            return tensor
+
+        if on_xla_device:
+            return self._tpu_sync_all_reduce(tensor)
```
vfdev-5, line 148: Can we directly call `self._do_reduction(tensor, self._gpu_reduce)` intead ?
The purpose of `_gpu_sync_all_reduce` and `_tpu_sync_all_reduce` is very limited for me...

vfdev-5:
@erip can add more tests with real metrics: Accuracy, Recall, Precision etc ?

erip:
Absolutely.

erip:
The distributed test fixtures are going to require some clever thought. [`xmp.spawn`](https://pytorch.org/xla/release/1.5/index.html#torch_xla.distributed.xla_multiprocessing.spawn) requires a function that is passed to all workers. In the case of our specs, that would be the actual test to be run by each worker and then joined. I'm trying to figure out how to defer that info until the tests run since it won't be available in the `conftest.py` where fixtures are set up.

vfdev-5:
@erip yes, let's execute tpu tests with a single worker which does `xmp.spawn` etc. In fixtures we can provide an executing method that requires test as input function.

tests/ignite/metrics/test_precision.py
```diff
@@ -730,7 +729,7 @@ def _test(average, n_epochs):
         n_classes = 7
 
         offset = n_iters * s
-        y_true = torch.randint(0, n_classes, size=(offset * dist.get_world_size(),)).to(device)
+        y_true = torch.randint(0, n_classes, size=(offset * world_size)).to(device)
```
vfdev-5, line 732:
```suggestion
        y_true = torch.randint(0, n_classes, size=(offset * world_size, )).to(device)
```

erip:
> @erip yes, let's execute tpu tests with a single worker which does `xmp.spawn` etc. In fixtures we can provide an executing method that requires test as input function.

So your thought is something like...

```python
@pytest.mark.tpu
@pytest.mark.distributed
@pytest.mark.skipif(..., reason="Skip if no TPU")
def test_distrib_gpu(local_rank, distributed_context_single_node_xrt_executor):
    device = xm.get_device()
    args = (device, xm.get_ordinal(), xm.get_world_size())
    distributed_context_single_node_xrt_executor(_test_distrib_itegration_multiclass, args)
    distributed_context_single_node_xrt_executor(_test_distrib_itegration_multilabel, args)
```

or did I misunderstand?

vfdev-5:
> > @erip yes, let's execute tpu tests with a single worker which does `xmp.spawn` etc. In fixtures we can provide an executing method that requires test as input function.
> 
> So your thought is something like...
> 
> ```python
> @pytest.mark.tpu
> @pytest.mark.distributed
> @pytest.mark.skipif(..., reason="Skip if no TPU")
> def test_distrib_tpu(local_rank, distributed_context_single_node_xrt_executor):
>     args = (xm.get_device(), xm.get_ordinal(), xm.get_world_size())
>     distributed_context_single_node_xrt_executor(_test_distrib_itegration_multiclass, args)
>     distributed_context_single_node_xrt_executor(_test_distrib_itegration_multilabel, args)
> ```
> 
> or did I misunderstand?

Yes, something like that :)

ignite/metrics/metric.py
```diff
@@ -108,14 +118,31 @@ def _sync_all_reduce(self, tensor: Union[torch.Tensor, numbers.Number]) -> Union
...
             return tensor.item()
         return tensor
 
+    @staticmethod
+    def _tpu_reduce(tensor: Union[torch.Tensor, numbers.Number]) -> None:
+        xm.all_reduce("sum", [tensor,])
+
+    @staticmethod
+    def _gpu_reduce(tensor: Union[torch.Tensor, numbers.Number]) -> None:
+        dist.barrier()
```
sdesrozis, line 133: I think this barrier is useless. The following all_reduce() is a sync collective call.

erip: Since it's there currently, I'd be a bit wary of changing the behavior in this PR. Since a lot of things are changing, I'd hate to have less information about potential breakages.

sdesrozis: I agree!! 😊

erip:
This is going to be blocked until https://github.com/pytorch/xla/pull/2023 is merged, I think.

vfdev-5:
On nightly the problem can be found here : https://pipelines.actions.githubusercontent.com/GjXvbhyRrTHRIrk9p03R8DrZ91xWSv51MZra799xsxhclrqZfD/_apis/pipelines/1/runs/934/signedlogcontent/4?urlExpires=2020-05-04T21%3A17%3A13.3601076Z&urlSigningMethod=HMACV1&urlSignature=PSs%2Bazok0xClsvtr8MlEGCgxKYN7mJQjgyXt60%2FLKOE%3D

A warning is not raised for multilabel case without average...

erip:
Building the nightly xla image now.

vfdev-5:
@erip the problem more precisely is here : https://github.com/pytorch/ignite/blob/ecf10c61f95baa3ae8421ddde187423ed9c251fc/ignite/metrics/precision.py#L22

erip:
Note for myself: need more tests. I believe many of them won't run on CI, but we can test on colab or kaggle.

ignite/metrics/metric.py
```diff
@@ -11,6 +11,15 @@
 
 from ignite.engine import Events, Engine
 
+try:
+    import torch_xla.core.xla_model as xm
+
+    on_xla_device = True
+    xrt_world_size = xm.xrt_world_size()
```
vfdev-5, line 18: By calling `xm.xrt_world_size()` outside `xm.spawn` we will always have 1.

vfdev-5:
@erip I changed the base branch from master to idist. As now in idist we have some common dist handling we can adapt metrics reduction tests to use idist instead of torch native dist

erip:
That's a good thought. Thanks - will try to roll in some of the new APIs

vfdev-5:
@erip I'll try to do a merge with idist branch

vfdev-5:
Closed in favor of #1042

PR #989, opened 2020-04-27
Fix README.md 

sdesrozis:
Description:

Minor fix for custom event, use `EventEnum` instead of `Enum`

Check list:


vfdev-5:
Please, check the same in FAQ.

PR #994, opened 2020-04-28
Fixes #993

### LINKED ISSUE #993: Engine restarts training from scratch unexpected
## 🐛 Bug description

Engine `_is_done` function only takes into account `max_epochs` parameter read from the state of the object. This leads to the followling unexpected behavior:

1. Instantiate a simple engine, train for 2 epochs (set max_epochs=2). Save checkpoints.
2. Train for 2 more epochs (set max_epochs=4) using last checkpoint.
3. This leads to the training restarting from scratch instead of continuing from the last checkpoint.

Why?

Engine.run functions has a `max_epochs` parameter that should have the precedence over the `max_epochs` stored in the checkpoint state in the [_is_done](https://github.com/pytorch/ignite/blob/master/ignite/engine/engine.py#L605). 
Currently, at runtime, to check if the training is done, we consider `max_epochs` from the state of the checkpoint, not the `max_epochs` parameter from the user. Therefore if the user specifies a different value (in the case he wants to train for more epochs), his input is not taken into account.

Here is a possible workaround to explain how to solve the problem:
```
    @staticmethod
    def _is_done(state, max_epochs=None):
        return state.iteration == state.epoch_length * (state.max_epochs if max_epochs is None else max_epochs)
```

## Environment

 - PyTorch Version (e.g., 1.4): 1.4.0
 - Ignite Version (e.g., 0.3.0): 0.4.0
 - OS (e.g., Linux): macOS
 - Python version: 3.6.10

vfdev-5:
Fixes #993

Description:
- Take into account run's argument before checking if run is done

Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -552,11 +552,12 @@ class Engine:
             data (Iterable): Collection of batches allowing repeated iteration (e.g., list or `DataLoader`).
             max_epochs (int, optional): Max epochs to run for (default: None).
                 If a new state should be created (first run or run again from ended engine), it's default value is 1.
-                This argument should be `None` if run is resuming from a state.
+                If run is resuming from a state, provided `max_epochs` will be taken into account and should be larger
+                than `engine.state.max_epochs`.
             epoch_length (int, optional): Number of iterations to count as one epoch. By default, it can be set as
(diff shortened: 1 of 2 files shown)


sdesrozis:
CI ko 😔

PR #1005, opened 2020-05-03
add TPU checkpointing to CPU.

### LINKED ISSUE #999: TPU and Checkpoint
## 🚀 Feature

In case when working with TPU, we need to go to cpu in order to save the model: http://pytorch.org/xla/release/1.5/index.html#saving-and-loading-xla-tensors
This case should be automatically covered by `Checkpoint`.

erip:
Fixes #999

Description: Automatically checkpoints to CPU when XLA is available.

Check list:


ignite/handlers/checkpoint.py
```diff
@@ -369,6 +378,12 @@ def __call__(self, checkpoint: Mapping, filename: str) -> None:
                 tmp.close()
                 os.rename(tmp.name, path)
 
+    def _save(self, checkpoint: Mapping, filename: str):
+        if on_xla_device:
+            xm.save(checkpoint, filename)
```
vfdev-5, line 383: @erip I think we need to do this manually by ourselves. 
Tested the following in the colab and seems like it gets stucked
```python
import torch
import torch.nn as nn
import torch_xla
import torch_xla.distributed.xla_multiprocessing as xmp

# "Map function": acquires a corresponding Cloud TPU core, creates a tensor on it,
# and prints its core
def save_fn(index, flags):
  # Sets a common random seed - both for initialization and ensuring graph is the same
  torch.manual_seed(1234)

  # Barrier to prevent master from exiting before workers connect.
  xm.rendezvous('init')

  # Acquires the (unique) Cloud TPU core corresponding to this process's index
  device = xm.xla_device()
  model = nn.Linear(10, 10).to(device)
  opt = torch.optim.SGD(model.parameters(), lr=0.01)

  model.train()
  opt.zero_grad()
  x = torch.rand(10).to(device)
  y = model(x)
  y.sum().backward()
  xm.optimizer_step(opt)

  to_save = {"model": model.state_dict(), "opt": opt.state_dict(), "other": [1, 2, 3]}
  
  if index == 0:
      xm.save(to_save, "test.pt")
      print(index, "Saved")

  print(index, "continue")
  x = torch.rand(10).to(device)


# Spawns eight of the map functions, one for each of the eight cores on
# the Cloud TPU
flags = {}
# Note: Colab only supports start_method='fork'
xmp.spawn(save_fn, args=(flags,), nprocs=8, start_method='fork')
```

erip: That's strange. Their [documentation](https://pytorch.org/xla/release/1.5/index.html#saving-and-loading-xla-tensors) says that `xm.save` should handle the bookkeeping of moving to CPU for you... I'll have a look at this soon. Thanks for flagging this.

erip: Oh, I see: 

> In case of multple devices, the above API will only save the data for the master device ordinal (0).

vfdev-5: It stucks at `converted_data = xm._maybe_convert_to_cpu(to_save, convert=True)` if I replace `xm.save` by that.

vfdev-5: > Oh, I see:
> 
> > In case of multple devices, the above API will only save the data for the master device ordinal (0).
> 
> This is something we'll need to think though...

I think this is not a problem. In DDP we also store only on rank 0.

erip: This is very surprising. I'd think `xm.save` would just work. Do you suspect it's a bug in XLA? It seems likelier that I've just misread the documentation.

erip: @vfdev-5 I think I found the issue in the above snippet. It seems like [xm.save already handles checking whether the saving process is the master ordinal](https://github.com/pytorch/xla/issues/1245#issuecomment-546708708). I'm testing on colab now...

erip: It works on colab as expected if you remove the `if`. You can verify this by changing the saved model file to `f"test-{index}.pt"` and you'll see the only artifact is `test-0.pt`.

vfdev-5:
@erip could you please integrate this PR into `idist` branch with tests: 1 proc and multi-procs  and save model, optimizer and trainer ?

erip:
> @erip could you please integrate this PR into `idist` branch with tests: 1 proc and multi-procs and save model, optimizer and trainer ?

Absolutely. We'll want to figure out the right way to handle the `rank == 0` situation: XLA handles this internally in `xm.save`, but `torch.save` does not. 😦 I will check to see if the new APIs have some flexibility for this.

vfdev-5:
> > @erip could you please integrate this PR into `idist` branch with tests: 1 proc and multi-procs and save model, optimizer and trainer ?
> 
> Absolutely. We'll want to figure out the right way to handle the `rank == 0` situation: XLA handles this internally in `xm.save`, but `torch.save` does not. 😦 I will check to see if the new APIs have some flexibility for this.

yes, something like that : 
```python
if idist.has_xla_support:
    xm.save(...)
elif idist.get_rank() == 0:
    torch.save(...)
```

ignite/handlers/checkpoint.py
```diff
@@ -29,7 +30,8 @@ def remove(self, filename: str) -> None:
 class Checkpoint:
     """Checkpoint handler can be used to periodically save and load objects which have attribute
     `state_dict`/`load_state_dict`. This class can use specific save handlers to store on the disk or a cloud
-    storage, etc.
+    storage, etc. The Checkpoint handler also handles automatically moving devices on TPU to CPU before writing
```
vfdev-5, line 33:
```suggestion
    storage, etc. The Checkpoint handler also handles automatically moving data on TPU to CPU before writing
```
?

ignite/handlers/checkpoint.py
```diff
@@ -30,7 +30,7 @@ def remove(self, filename: str) -> None:
 class Checkpoint:
     """Checkpoint handler can be used to periodically save and load objects which have attribute
     `state_dict`/`load_state_dict`. This class can use specific save handlers to store on the disk or a cloud
-    storage, etc. The Checkpoint handler also handles automatically moving devices on TPU to CPU before writing
+    storage, etc. The Checkpoint handler also handles automatically moving data on TPU to CPU before writing
```
vfdev-5, line 33: Actually, this is not done by Checkpoint but DiskSaver ?

vfdev-5:
Maybe it should out of the scope of this PR, but can we generalize this for other contrib handlers like NeptuneSaver or TrainsSaver ? 

- For TrainsSaver, seems like it is OK
- For NeptuneSaver probably we can address it in another PR

ignite/handlers/checkpoint.py
erip: Technically yes. My thought was that Since the `Checkpoint` handler defers to the `DiskSaver` and since I imagine most people work through `*Checkpoint`, including that info here would make most sense. I am happy to move it as needed.

erip:
> t of the scope of this PR, but can we generalize this for other contrib handlers like NeptuneSaver or TrainsSaver ?
> 
> * For TrainsSave

I was thinking about this, too.

PR #1006, opened 2020-05-03
Simplify BaseLogger attach APIs

### LINKED ISSUE #980: Simplified base logger's API
## 🚀 Feature

Idea is to simplify current logger's API by creating a shortcut methods:
Currently, to log output parameters and optmizer's params we do (example is for TB, but is generic for others too):
```python
    logger = TensorboardLogger(...)
    logger.attach(
        trainer,
        log_handler=OutputHandler(tag="training", metric_names="all"),
        event_name=Events.ITERATION_COMPLETED(every=log_every_iters),
    )
    logger.attach(
        trainer,
        log_handler=OptimizerParamsHandler(optimizer, param_name="lr", tag=k),
        event_name=Events.ITERATION_STARTED(every=log_every_iters),
    )
```
Proposed API:
```python
    logger = TensorboardLogger(...)
    logger.attach_output_handler(
        trainer,
        tag="training", metric_names="all",  # other args, kwargs of OutputHandler
        event_name=Events.ITERATION_COMPLETED(every=log_every_iters),
    )
    logger.attach_opt_params_handler(
        trainer,
        optimizer, param_name="lr", tag=k, # other args, kwargs of OutputHandler
        event_name=Events.ITERATION_STARTED(every=log_every_iters),
    )
```

In addition, this can fix "bad-practice" current implementation of `setup_any_logging` taking as input logger and logging_module:
https://github.com/pytorch/ignite/blob/d1aa5fd5e71676032c952aa7470a651c216e2dd9/ignite/contrib/engines/common.py#L209

erip:
Fixes #980

Description: Adds new interface, docstring, and implementations for simplified base_logger API.

Check list:

### DIFF
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -205,7 +205,7 @@ def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, lo
     raise DeprecationWarning("setup_any_logging is deprecated")
 
 
-def setup_logger(logger, trainer, optimizers, evaluators, log_every_iters):
+def _setup_logging(logger, trainer, optimizers, evaluators, log_every_iters):
     if optimizers is not None:
         from torch.optim.optimizer import Optimizer
 
@@ -265,7 +265,7 @@ def setup_tb_logging(output_path, trainer, optimizers=None, evaluators=None, log
         TensorboardLogger
     """
     logger = TensorboardLogger(log_dir=output_path)
-    setup_logger(logger, trainer, optimizers, evaluators, log_every_iters)
+    _setup_logging(logger, trainer, optimizers, evaluators, log_every_iters)
     return logger
 
 
@@ -289,7 +289,7 @@ def setup_visdom_logging(trainer, optimizers=None, evaluators=None, log_every_it
         VisdomLogger
     """
     logger = VisdomLogger(**kwargs)
-    setup_logger(logger, trainer, optimizers, evaluators, log_every_iters)
+    _setup_logging(logger, trainer, optimizers, evaluators, log_every_iters)
     return logger
 
 
@@ -312,7 +312,7 @@ def setup_mlflow_logging(trainer, optimizers=None, evaluators=None, log_every_it
         MLflowLogger
     """
     logger = MLflowLogger()
-    setup_logger(logger, trainer, optimizers, evaluators, log_every_iters)
+    _setup_logging(logger, trainer, optimizers, evaluators, log_every_iters)
     return logger
 
 
@@ -335,7 +335,7 @@ def setup_plx_logging(trainer, optimizers=None, evaluators=None, log_every_iters
         PolyaxonLogger
     """
     logger = PolyaxonLogger()
-    setup_logger(logger, trainer, optimizers, evaluators, log_every_iters)
+    _setup_logging(logger, trainer, optimizers, evaluators, log_every_iters)
     return logger
 
 
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -12,6 +12,7 @@ from ignite.contrib.handlers import ProgressBar
 from ignite.contrib.handlers import TensorboardLogger, global_step_from_engine
 from ignite.contrib.handlers import VisdomLogger
 from ignite.contrib.handlers import NeptuneLogger
+from ignite.contrib.handlers import WandBLogger
 from ignite.contrib.metrics import GpuInfo
 from ignite.engine import Engine, Events
 from ignite.handlers import TerminateOnNan, ModelCheckpoint, EarlyStopping
@@ -339,6 +340,29 @@ def setup_neptune_logging(trainer, optimizers=None, evaluators=None, log_every_i
         MLflowLogger
     """
     logger = NeptuneLogger()
+    _setup_logging(logger, trainer, optimizers, evaluators, log_every_iters)
+    return logger
+
+
+def setup_wandb_logging(trainer, optimizers=None, evaluators=None, log_every_iters=100):
+    """Method to setup WandB logging on trainer and a list of evaluators. Logged metrics are:
+        - Training metrics, e.g. running average loss values
+        - Learning rate(s)
+        - Evaluation metrics
+
+    Args:
+        trainer (Engine): trainer engine
+        optimizers (torch.optim.Optimizer or dict of torch.optim.Optimizer, optional): single or dictionary of
+            torch optimizers. If a dictionary, keys are used as tags arguments for logging.
+        evaluators (Engine or dict of Engine, optional): single or dictionary of evaluators. If a dictionary,
+            keys are used as tags arguments for logging.
+        log_every_iters (int, optional): interval for loggers attached to iteration events. To log every iteration,
+            value can be set to 1 or None.
+
+    Returns:
+        MLflowLogger
+    """
+    logger = WandBLogger()
     _setup_logging(logger, trainer, optimizers, evaluators, log_every_iters)
     return logger
 
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -11,6 +11,7 @@ from ignite.contrib.handlers import PolyaxonLogger
 from ignite.contrib.handlers import ProgressBar
 from ignite.contrib.handlers import TensorboardLogger, global_step_from_engine
 from ignite.contrib.handlers import VisdomLogger
+from ignite.contrib.handlers import NeptuneLogger
 from ignite.contrib.metrics import GpuInfo
 from ignite.engine import Engine, Events
 from ignite.handlers import TerminateOnNan, ModelCheckpoint, EarlyStopping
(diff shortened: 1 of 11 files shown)


erip:
I think the failure is unrelated. I still need to write some tests, though.

sdesrozis:
@erip Thanks :)

The file `ignite/contrib/engines/common.py` should be refactored to avoid import all loggers module (see `setup_any_logger` function).

erip:
Thanks for the comment @sdesrozis. I was thinking about the `setup_any_logger` code while doing this and I this it's possible to clean it up. I'm wondering if we should deprecate it or just completely remove it. It's technically part of the public API, but doesn't really serve any use outside of a private method... WDYT?

sdesrozis:
> Thanks for the comment @sdesrozis. I was thinking about the `setup_any_logger` code while doing this and I this it's possible to clean it up. I'm wondering if we should deprecate it or just completely remove it. It's technically part of the public API, but doesn't really serve any use outside of a private method... WDYT?

I suppose `setup_any_logger` is private. If it’s useless, we could remove it. We should just provide the setup of concrete loggers on public API. I don’t think it’s worth depreciating.

vfdev-5:
Technically `setup_any_logger` is in v0.3.0 as a public API method (but, yes I agree that it is more private than public). So, technically we should not just remove it without any suggestions on how to use new API. At least we can raise an exception with appropriate message.

ignite/contrib/handlers/mlflow_logger.py
```diff
@@ -258,3 +262,9 @@ def close(self):
         import mlflow
 
         mlflow.end_run()
+
+    def attach_output_handler(self, engine: Engine, event_name: str, *args: Any, **kwargs: Mapping):
```
vfdev-5, line 266: @sdesrozis @erip do you think we could refactor this part of code into base logger and avoid rewriting the same lines for all supported loggers ?

erip: I have been thinking about that. I don't know if it's possible, actually. 😦 My thought is that maybe there's some way to use inner classes to abstract over a specific implementation of `OutputHandler` and `OptimizerParamsHandler` ... but I don't think it's possible. See [here](https://stackoverflow.com/questions/61582840/is-there-some-mechanism-to-enforce-nested-interfaces-are-implemented-in-abstract).

erip: It may be possible with some incredible hackiness... maybe through registration or metaclass magic... we could define properties which are the `cls` constructor for the implementations specific to a given `BaseLogger` implementation. Not sure how to do this least invasively and in the least amount of hacks. 😆

vfdev-5:
@erip how about the following hack :)
```python
class BaseOutputHandler:
    
    def __call__(self, x):
        print("BaseOutputHandler", x)


class Engine:
    
    
    def __init__(self):
        self.handlers = []
    
        
    def __call__(self, x):
        for h in self.handlers:
            h(x)
        
        
class BaseLogger:
    
    output_handler_cls = BaseOutputHandler
    
    def attach_output(self, e):
        e.handlers.append(self.output_handler_cls())
        

        
class TbOutputHandler:
    
    def __call__(self, x):
        print("TbOutputHandler", x ** 2)

        
class TbLogger(BaseLogger):
    
    output_handler_cls = TbOutputHandler

engine = Engine()
logger = BaseLogger()
logger.attach_output(engine)

tb_logger = TbLogger()
tb_logger.attach_output(engine)

engine(2)
> BaseOutputHandler 2
> TbOutputHandler 4
```

erip:
I was just thinking that approach. I think it's reasonable as long as we mark the `Base...Handler`s as `ABCMeta` so force them to be abstract.

vfdev-5:
@erip please go ahead with that approach :)

ignite/contrib/handlers/mlflow_logger.py
sdesrozis: So let’s keep it simple ! 😉

sdesrozis:
Here I would prefer use interface to provide contract to implement for developers of logger. It's more or less the same but more constrained.

What do you think about that ?

```python


class BaseOutputHandler:

    def __call__(self, x):
        print("BaseOutputHandler", x)


class Engine:
    def __init__(self):
        self.handlers = []

    def __call__(self, x):
        for h in self.handlers:
            h(x)


class BaseLogger(metaclass=ABCMeta):

    # should not provide an implementation
    @abstractmethod
    def output_handler(self):
        pass

    def attach_output(self, e):
        e.handlers.append(self.output_handler())


class TbOutputHandler:
    def __call__(self, x):
        print("TbOutputHandler", x ** 2)


class TbLogger(BaseLogger):

    def output_handler(self):
        return TbOutputHandler()


engine = Engine()

# It don't work, BaseLogger is for inheritance
# BaseLogger().attach_output(engine)
TbLogger().attach_output(engine)

engine(2)

> python test.py
> TbOutputHandler 4
```

erip:
@sdesrozis agreed. Have a look at [this](https://github.com/pytorch/ignite/pull/1006/commits/16f19e1ca51dd3be24fadde6c3b909e9ef7e33e1#diff-13bd69f26d458ae6b93663a075081eeaR146-R194) - hopefully this fits your model.

ignite/contrib/handlers/neptune_logger.py
```diff
@@ -1,11 +1,14 @@
 import numbers
 import tempfile
-from typing import Mapping
+from typing import Mapping, Any
```
vfdev-5, line 3: Do we still need `Any` ?

ignite/contrib/handlers/neptune_logger.py
```diff
@@ -1,11 +1,14 @@
...
 import warnings
 
 import torch
 
 import ignite
+
+from ignite.engine import Engine
```
vfdev-5, line 10: Same for `Engine`, maybe useless ?

sdesrozis:
I should have defined `output_handler` and `output_opt_handler` as abstract function 

In `BaseLogger`
```python
@abstract
def output_handler(self) -> BaseOutputHandler
    pass
```
And in concrete logger
```python
class MyOutputHandler(BaseOutputHandler):
    ....

class MyLogger(BaseLogger):
    def output_handler(self) -> BaseOutputHandler
        return MyOutputHandler
```

I think my typehints is wrong : the good  type is “class of BaseOutputHandler “

vfdev-5:
@sdesrozis or even more proper solution would be to dispactch those creators
```python
class BaseLogger:
    @abstractmethod
    def create_output_handler(self, *args, **kwargs) -> BaseOutputHandler
        pass

    def attach_output_handler(self, engine: Engine, event_name: str, *args: Any, **kwargs: Mapping):
        engine.add_event_handler(event_name, self.create_output_handler(*args, **kwargs))

class MyOutputHandler(BaseOutputHandler):
    ....

class MyLogger(BaseLogger):
    def create_output_handler(self, ) -> MyOutputHandler
        return MyOutputHandler(*args, **kwargs)

```

sdesrozis:
That looks like the perfect pattern here!

sdesrozis:
Abstract method is a guideline for developer. I prefer that but @erip chooses at the end 😊 I talk and I do nothing 😅 (except edit messages 😱)

vfdev-5:
@erip I added the code with implementation of `_create_output_handler` and `_create_opt_param_handler`, added docstrings... Attaching the way it is done, it should fix failed tests. 
Remains helper methods `setup_*_logging` for other loggers.

erip:
I think I've got the `setup...` helpers wrapped up. Over to @sdesrozis for review. 🥳

ignite/contrib/engines/common.py
```diff
@@ -320,15 +316,59 @@ def setup_mlflow_logging(trainer, optimizers=None, evaluators=None, log_every_it
...
+            torch optimizers. If a dictionary, keys are used as tags arguments for logging.
+        evaluators (Engine or dict of Engine, optional): single or dictionary of evaluators. If a dictionary,
+            keys are used as tags arguments for logging.
+        log_every_iters (int, optional): interval for loggers attached to iteration events. To log every iteration,
+            value can be set to 1 or None.
+
+    Returns:
+        NeptuneLogger
+    """
+    logger = NeptuneLogger()
```
vfdev-5, line 342: I was thinking about those methods and loggers that need **kwargs to initialize. 
Either we need to pass those things or directly work on the instance of such logger...

vfdev-5:
Seems like I was late before it waas merged :(

vfdev-5:
@erip or @sdesrozis could you add that in another PR please

ignite/contrib/engines/common.py
erip: I was thinking about this, too. I think the tricky thing is that the `setup_X_logger` by convention passes in default kwargs. Adding `*logger_constructor_args` and `**logger_constructor_kwargs` isn't easy because of that. I think there are some clever things we could possibly do... will try to draft a gist quickly.

erip: See [here](https://gist.github.com/erip/42cb83e9a9b82e294cb338686f6563f6)

sdesrozis: Like here ?

https://github.com/pytorch/ignite/blob/662433798683ef72920f7ba4b248b16558df6adc/ignite/contrib/engines/common.py#L296

erip: I think that only works if it's a **kwargs-only constructor.

sdesrozis: > I think that only works if it's a **kwargs-only constructor.

Can’t test, I just have my phone 😅

erip:
```python
>>> class Foo:
...     def __init__(a, b, c=None):
...         pass
...
>>> Foo(**dict(a=1, b=2, c=3))
  File "<stdin>", line 1, in <module>
TypeError: __init__() got multiple values for argument 'a'
```

erip: Ah, I was wrong. As long as we can guarantee the user specifies all constructor args they can pass it in as a dict at the end like in `VisdomLogger` even for positional args.

sdesrozis: So let’s define defaults and that will be ok 👌🏻

erip: Have #1015 out :-)

Issue #966, opened 2020-04-22
Checkpoint stateful handlers and metrics


amatsukawa:
## 🚀 Feature

Things that are attached to the `Engine` might have state that would ideally be checkpointed and restored using as part of the `Engine`'s `state_dict`.

An example is a `Checkpoint` handler when it has a `score_function`. Currently, the priorities the `Checkpoint` class stores is not saved anywhere. It is not able to recover gracefully from failure without manual intervention to parse the checkpoint path names, and directly setting the internals of the class.

Handlers and Metrics should have `state_dict` and `load_state_dict` methods (empty by default), and I think it should be possible for these to automatically make it into/restored from the `Engine`'s `state_dict` when it's attached to an `Engine`.

vfdev-5:
@amatsukawa thanks for FR. Yes, it makes sense for handlers and metrics with internal state 👍

amatsukawa:
FWIW, I'm completely happy with #1156 and the way things work now. 

With complications I didn't think about, eg. some handlers needing to go on the valid engine and handlers needing to run in a specific order (`Checkpoint` needs to run last after all other stateful handlers) perhaps automatically adding things to the engine's checkpoints is not trivial and might make things harder to reason about.

H4dr1en:
Bringing here a specific use case that this FR could solve: 

Having the `RunningAverage` metric being able to restore the state would allow to not "forget" previous iterations scores when resuming an experiments. Otherwise the metric can show "peaks" when resuming an experiment, as can be shown in the figure below:
![newplot(3)](https://user-images.githubusercontent.com/16240134/123123264-fea5a580-d446-11eb-90c9-7ba8947a538c.png)

Issue #706, opened 2020-01-19
Make sure metric handlers are fired first


beyondyoni:
Hi,

I noticed that since event handlers are invoked at the order of their insertion, there's a potential bug- if a metric handler is inserted after a TensorBoard handler, the metric value will not be ready at the time of writing.

While this has only minor effects (one epoch shift in the results), I think it can be easily solved by one of two approaches:
1. separating handlers list to two: one for metrics, and the other handler types.
2. More generic and scalable- allowing users to decide the order of execution by adding priority argument to `add_event_handler` _(default would be that metric handlers are inserted with higher priority)_ and then sort the list when `run()` is invoked.

Let me know if you like any of the ideas, and I'd be happy to draft a PR

vfdev-5:
@beyondyoni thanks for the ideas. IMO, the order of handlers is up to user to respect. So, if user adds at first the loggers and after that handlers providing the info, it is his/her resposibility.

However, I think adding some argument to setup a sort of priority of handlers can be an interesting enhancement ! For instance, I do not know how can we provide such option. If you would like to draft some ideas, please do not hesitate to draft a PR and we can reiterate over.

beyondyoni:
I agree with your opinion of order of execution being the user's responsibility but I think the current situation isn't clear enough (I actually came across this bug because a co-worker of asked for my help to solve it).

I made a new push. Let's discuss it there

vfdev-5:
Okay, I see. Well, let's try to fix this behaviour :)

> I made a new push. Let's discuss it there

Could you please make a separate PR for this. Thanks

vfdev-5:
@justusschock what do you think about this ?

justusschock:
I think this is a general question. We should decide, whether we want priorities/sorting of handlers or not (but definitely not support both versions).

If we implement it, it should always be used, but this may also introduce unwanted behaviour if some handlers are for example added by an internal function with a different priority, this may also confuse the user, since he might expect some other behaviour. 

I think, I slightly prefer the way including priorities, but this has to be carefully implemented.
On the implementation side, I would prefer to have the priority option not in the function adding the handler but as an attribute of the handler (comparable to the logging level in python's internal logging module).

beyondyoni:
I think the current design implicitly supports priorities, and is, therefore, a source for (potential) confusion, so I'm glad to see you're leaning towards changing that.

[...]

PR #1008, opened 2020-05-04
Issue 966

vfdev-5:
Addresses #966 but not completely closes the FR

Description:
- Added internal state for Checkpoint
- Saving logic is changed

-- previously we 
1. save a checkpoint file
2. updated internal `_saved`
3. removed other files if more than `n_saved`
 
-- now we
1. updated internal `_saved`
2. removed other files if more than `n_saved`
3. save a checkpoint file

This is due to the fact that we need to insert `Checkpoint` itself into user's provided `to_save` and save it in the training checkpoint. 
When we setup the checkpoint, Checkpoint's state dict should be up to date as we add new item and remove optionally other items.

This does not work automatically for saving best models.

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -165,17 +165,8 @@ class Checkpoint(Serializable):
     ):
 
         if to_save is not None:  # for compatibility with ModelCheckpoint
-            if not isinstance(to_save, collections.MutableMapping):
-                raise TypeError("Argument `to_save` should be a dictionary, but given {}".format(type(to_save)))
-
-            if len(to_save) < 1:
-                raise ValueError("No objects to checkpoint.")
-
-            self._check_objects(to_save, "state_dict")
-
-            # Add this checkpoint if Engine is present in to_save:
-            if any([isinstance(o, Engine) for _, o in to_save.items()]):
-                to_save["checkpoint_{}".format(id(self))] = self
+            self._check_to_save(to_save)
+            to_save = self._setup_to_save(to_save)
 
         if not (callable(save_handler) or isinstance(save_handler, BaseSaveHandler)):
             raise TypeError("Argument `save_handler` should be callable or inherit from BaseSaveHandler")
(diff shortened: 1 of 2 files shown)


amatsukawa:
Hi @vfdev-5. I see that you've taken a crack at this but decided not to merge this in.

Related to this topic, can you confirm what guarantees, if any, ignite provides about the order in which handlers run? I believe right now, handlers for an event run in the order in which they were registered, because `_fire_events` just uses a for-loop.

This is relevant because eg. the `EarlyStopping` handler needs to have its state checkpointed to resume correctly. That means it needs to run before the `Checkpoint` handler.

vfdev-5:
Hi @amatsukawa , yes, missing I tried to implement and stopped as automatic injection of Checkpoint's state can work only with Checkpoints added to a trainer etc. Some parts of this PR were required by other FR. So, maybe we can just add `state_dict/load_state_dict` methods and no automatic stuff for instance to avoid complex behaviour...

> Related to this topic, can you confirm what guarantees, if any, ignite provides about the order in which handlers run? I believe right now, handlers for an event run in the order in which they were registered, because _fire_events just uses a for-loop.

Yes, that's correct. Handlers are triggered in the order they were registered. There is an issue about some internal priority reordering (https://github.com/pytorch/ignite/issues/706), but it is on stand by...

> This is relevant because eg. the EarlyStopping handler needs to have its state checkpointed to resume correctly. That means it needs to run before the Checkpoint handler.

`EarlyStopping` and `Checkpoint` for best models are attached to evaluators, so, probably, their states can not be automatically appended to training checkpoint instance. Yes, maybe, there should be a correct order to follow to achieve correct behaviour...

The reason why this PR was closed is related to the fact that auto stuff will not work everywhere and probably can be only done inside higher-api Trainers (which for instance are absent). On the other hand, explaning to user how to manually use this correctly to setup checkpointing of `model, optimizer, trainer`, checkpointer itself, checkpointer best model, early stopping and various metrics with state... it can become a nightmare :)

PR #1010, opened 2020-05-04
Updated workflows/tpu-tests.yaml

vfdev-5:
Description:
- Simplified the workflow without installing conda
- added matrix for xla version: 1.5 and nightly


erip:
This is great. Once this is merged, we should expect the TPU metrics test to pass on nightly, but not in stable.

PR #1016, opened 2020-05-05
pin pytorch version to 1.x.x in setup.py and in meta.yaml.

### LINKED ISSUE #897: PyTorch dependency is lacking version constraint
## 🐛 Bug description

PyTorch is a dependency of Ignite and, thus, is specified in `setup.py`

https://github.com/pytorch/ignite/blob/4b311cc82fe45b3082661125cd7ee54007283fb0/setup.py#L24-L26

and `conda.recipe/meta.yaml`:

https://github.com/pytorch/ignite/blob/4b311cc82fe45b3082661125cd7ee54007283fb0/conda.recipe/meta.yaml#L15-L23

The PyTorch dependency is lacking a version constraint which may work fine right now, but there is no guarantee that Ignite will be compatible with any future major PyTorch release (e.g. PyTorch v2.x).

I suggest to constrain the PyTorch version that Ignite is compatible with, e.g. `>=1.0,<2` or `<2` if any `0.x` and `1.x` version works. If PyTorch has a new major release, even previous Ignite versions can become compatible with the new major PyTorch release (especially if no changes to the code are necessary) by making new bug fix releases with relaxed version constraints to include the new PyTorch version.

In my opinion, it is highly preferable to be conservative about dependency version constraints through a [compatible release constraint](https://www.python.org/dev/peps/pep-0440/#compatible-release) in case the dependency conforms with semantic versioning. It is impossible to guarantee compatibility with a future major release of a dependency as its API can change arbitrarily.

erip:
Fixes #897 

Description: Pins the version to >=1.0, <2 in conda recipe and in setup.py.

Check list:


vfdev-5:
Shouldn't we run simple tests on those pytorch version: 1.0, 1.1, 1.2 etc ?

erip:
Yes, probably so

vfdev-5:
I do not want to load more our CI with complete matrix of Python and Pytorch versions.
Something in the middle can be to check for a single Python version, all previous pytorch version and run them cron-based once per week or so (another github action). No smoke or distrib tests there.

vfdev-5:
But maybe we can leave it for another PR...

erip:
I added an action to run on Sundays - let me know what you think, @vfdev-5

.github/workflows/pytorch-version-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+    - cron:  '0 0 * * 0'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 10
+      fail-fast: false
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
```
vfdev-5, line 13: I think 3.8 support was not available for <1.4
And we can remove pytorch 1.5 too

PR #1020, opened 2020-05-07
Add TrainsLogger

### LINKED ISSUE #892: Add Trains logger
## 🚀 Feature

Add logger to [Trains](https://github.com/allegroai/trains) similarly to Neptune, MLflow, ...

jkhenning:
Fixes #892 

Description:
- Adds TrainsLogger to support [Trains](https://github.com/allegroai/trains) Experiment Manager & Version Control, allowing you to log your experiment metadata to the [Trains Server](https://github.com/allegroai/trains-server) of your choice.
- Includes all handlers for Output, Grads(scalars), Optimizer Params, Model Weights (scalars) as well as logging model checkpoints to the server.

Check list:


vfdev-5:
@jkhenning thanks for the PR ! 
@erip could you please review. Thanks !

ignite/contrib/handlers/trains_logger.py
```diff
@@ -0,0 +1,633 @@
...
+                title="{}grads_{}".format(tag_prefix, title_name),
+                series=series_name,
+                step=global_step,
+                hist_data=p.grad.detach().cpu().numpy(),
+            )
+
+
+class TrainsLogger(BaseLogger):
+    """
+    Trains handler to log metrics, text, model/optimizer parameters, plots during the training and validation.
```
vfdev-5, line 386: Please add a link on your website :
```
`Trains <https://link>`_ handler
```

examples/contrib/mnist/mnist_with_trains_logger.py
```diff
@@ -0,0 +1,175 @@
...
+
+    model.to(device)  # Move model before creating optimizer
+    optimizer = SGD(model.parameters(), lr=lr, momentum=momentum)
+    criterion = nn.CrossEntropyLoss()
+    trainer = create_supervised_trainer(model, optimizer, criterion, device=device)
+
+    if sys.version_info > (3,):
+        from ignite.contrib.metrics.gpu_info import GpuInfo
+
+        try:
```
vfdev-5, line 82: @jkhenning do you have some auto-logging of GPU usage ? If yes, we can remove this part of using `GpuInfo`

vfdev-5:
@jkhenning there is also a redefinition of test flake error to fix : https://travis-ci.org/github/pytorch/ignite/jobs/684300841#L238
Thanks !

ignite/contrib/handlers/trains_logger.py
jkhenning: Done

examples/contrib/mnist/mnist_with_trains_logger.py
jkhenning: You're quite right :) removed.

erip:
Code looks clean and I didn't see any issues. One question for @jkhenning: why are you pinning to the release candidate in the `requirements-dev.txt`?

jkhenning:
@erip The reason for the release candidate is a small adjustment I had to make in the Trains binding code to make the integration cleaner. I wanted to make sure this PR was cleared so I can release the official Trains v0.14.3.
I'd appreciate it if you can wait with this PR's merge so I can release Trains v0.14.3 in the next day or two, change this dependency in the PR and it'll be ready to merge.

ignite/contrib/handlers/trains_logger.py
```diff
@@ -0,0 +1,578 @@
...
+            handler = Checkpoint(to_save, TrainsSaver(trains_logger), n_saved=1,
+                                 score_function=lambda e: 123, score_name="acc",
+                                 filename_prefix="best",
+                                 global_step_transform=global_step_from_engine(trainer))
+
+            validation_evaluator.add_event_handler(Events.EVENT_COMPLETED, handler)
+
+    """
+
+    def __init__(self, logger: TrainsLogger, output_uri=None, dirname=None, *args, **kwargs):
```
erip, line 576: Nit: would you mind removing the type for `logger` or adding types to the remaining args? It's not necessary for the whole class, but consistency within a single method would be nice. 😃

ignite/contrib/handlers/trains_logger.py
```diff
@@ -0,0 +1,578 @@
...
+        try:
+            from trains import Task
+        except ImportError:
+            raise RuntimeError(
+                "This contrib module requires trains to be installed. "
+                "You may install trains using: \n pip install trains \n"
+            )
+
+        if not dirname:
+            dirname = tempfile.mkdtemp(
```
erip, line 589: Does this temp dir get cleaned up somehow?

ignite/contrib/handlers/trains_logger.py
```diff
@@ -0,0 +1,578 @@
...
+        """
+        Get artifact local copy
+        :param filename: artifact name.
+        :return: a local path to a downloaded copy of the artifact
+        """
+        artifact = self.task.artifacts.get(filename)
+        if artifact:
+            return artifact.get_local_copy()
+        self.task.get_logger().report_text("Can not find artifact {}".format(filename))
+        return None
```
erip, line 637: This return isn't necessary.

erip:
Excellent. I don't have write access, so @vfdev-5 will have to make sure this doesn't get merged until the new trains release is cut.

ignite/contrib/handlers/trains_logger.py
jkhenning: Not explicitly, but it's in the system's temp folder, so just like any other system temporary files.
The thinking was that since we allow model checkpoint logging even if `dirname` was not provided, it would be nice for a user to be able to run her experiment (even for a quick try-out, debugging or whatever) and see model checkpoints logged, pointing to files that are actually stored on the training machine (even if she didn't bother providing a target folder yet).
The `output_uri` param in the saver (or in the logger, actually) allows the user to provide an upload target for model checkpoints, but an actual file is still required, at least until the checkpoint is uploaded (asynchronously).

ignite/contrib/handlers/trains_logger.py
jkhenning: Sure 😄

ignite/contrib/handlers/trains_logger.py
jkhenning: 👍

ignite/contrib/handlers/trains_logger.py
```diff
@@ -454,6 +454,8 @@ class TrainsLogger(BaseLogger):
 
     """
 
+    _bypass = None
```
erip, line 461: Can we clean this up somehow? I think class-level mutable attributes are asking for trouble. It _seems_ like it could be a constructor arg which would obviate the setters.

ignite/contrib/handlers/trains_logger.py
erip: Ah, I see. Makes sense. All good - just checking. 👍

ignite/contrib/handlers/trains_logger.py
```diff
@@ -0,0 +1,636 @@
...
+            self._task = trains.Task.init(
+                project_name=kwargs.get("project_name"),
+                task_name=kwargs.get("task_name"),
+                task_type=kwargs.get("task_type", trains.Task.TaskTypes.training),
+                **experiment_kwargs,
+            )
+
+        self.trains_logger = self._task.get_logger()
+
+        self.grad_helper = trains.binding.frameworks.tensorflow_bind.WeightsGradientHistHelper(
```
vfdev-5, line 512: I'm trying to find this class (`WeightsGradientHistHelper`) in your codebase: https://github.com/allegroai/trains/blob/master/trains/binding/frameworks/tensorflow_bind.py
@jkhenning is it related to your rc version ?

ignite/contrib/handlers/trains_logger.py
jkhenning: IMO a constructor argument will only confuse users as this is a hack for testing and should not be used in any normal use-case...
I absolutely agree and share the sentiment 😄 , and I think we can simply forego the class-scope definition and stick to using `getattr(cls, "_bypass", None)` and `setattr(cls, "_bypass")` - it won't show in the interface or class definition, and will serve it's purpose, limited only to `bypass_mode()` and `set_bypass_mode()`.
What do you think?

ignite/contrib/handlers/trains_logger.py
jkhenning: This is exactly the small refactoring in the RC version 😄

ignite/contrib/handlers/trains_logger.py
erip: If it's strictly for testing, can we mock it?

ignite/contrib/handlers/trains_logger.py
vfdev-5: Okay, thanks ! Just a bit funny to see tensorflow term in pytorch based code :)

ignite/contrib/handlers/trains_logger.py
jkhenning: I think that mocking `TrainsLogger` instead of using the bypass will mean that relevant tests (those where we actually use the bypass, such as `test_integration()` and `test_integration_as_context_manager()`) will not actually test anything... The bypass mode itself is actually a kind of internal mock (only for the internal Task representation and server-side communication), and is currently used since Trains does not yet support an offline mode (which would, when available, be a good constructor argument).
Since I think the tests are important, I would suggest keeping the bypass mode (as suggested, not using a class-scope member), and replacing it with an offline mode, once we introduce that into Trains.
What do you think?

vfdev-5: @jkhenning OK, I see. How do you perform test integrations for other libs, do you mock like that ? Otherwise we can log to your demo server ?

jkhenning: @vfdev-5 Yes, this is what we do in other integrations as well.
As for the demo server, we really try to keep it away from integration testing, mostly since we never want your integration testing dependent on a server that might be intermittently unavailable (due to load, maintenance, new version roll-outs etc.). Since the offline mode is on our to do list anyway, I really think this is the best solution going forward, and that the bypass is good enough for now 😄

ignite/contrib/handlers/trains_logger.py
```diff
@@ -0,0 +1,634 @@
...
+        project_name (str): The name of the project in which the experiment will be created. If the project
+            does not exist, it is created. If ``project_name`` is ``None``, the repository name is used. (Optional)
+        task_name (str): The name of Task (experiment). If ``task_name`` is ``None``, the Python experiment
+            script's file name is used. (Optional)
+        task_type (str): Optional. The task type. Valid values are:
+            - ``TaskTypes.training`` (Default)
+            - ``TaskTypes.train``
+            - ``TaskTypes.testing``
+            - ``TaskTypes.inference``
+        report_freq (int): Optional. Histogram processing frequency (handle hist values every X calls to the handler).
```
vfdev-5, line 406: @jkhenning I was checking documentation layout of `TrainsLogger` and seems like I missed those args:
- report_freq
- histogram_update_freq_multiplier

In ignite we can easily filter out how frequently call an event handler, so, does `report_freq` param mean to call `WeightsScalarHandler` every `report_freq` times :
```python
            trains_logger.attach(trainer,
                                 log_handler=WeightsScalarHandler(model),
                                 event_name=Events.ITERATION_COMPLETED(every=report_freq))
```
?

jkhenning: Hi @vfdev-5 ,

I think you're right, and it was my mistake to expose these values (they are related to the internal working of how we process and average the reports internally). I'll open a PR and remove them, if you'd like.

vfdev-5: Well, if it does almost the same as event filtering, so, yes please send a PR and remove them...

jkhenning: @vfdev-5 opened https://github.com/pytorch/ignite/pull/1036

PR #1021, opened 2020-05-07
Fix frequency tests

### LINKED ISSUE #1017: Fix test flakiness
## 🐛 Bug description

The tests are very flaky. We should identify the flaky tests are try to address them in some other way.

erip:
Fixes #1017

Description: because the sleeps in the frequency tests are so short, any issues that cause a delay become effectively extra sleeps. Since we compute expected times in terms of expected total sleep time, that makes it a very sensitive factor. This test increases the sleep times to make them slightly more robust.

Check list:

### DIFF
--- a/tests/ignite/metrics/test_frequency.py
+++ b/tests/ignite/metrics/test_frequency.py
@@ -21,9 +21,9 @@ def test_nondistributed_average():
     assert average_lower_bound < average < average_upper_bound
 
 
-def _test_frequency_with_engine(device, workers, every=1):
+def _test_frequency_with_engine(device, workers, lower_bound_factor=0.8, every=1):
 
-    artificial_time = 0.1 / workers  # seconds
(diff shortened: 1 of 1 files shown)


erip:
@vfdev-5 maybe it's worth rerunning this a few times to ensure it passes. I can do this by issuing dummy commits or you can by rerunning from GitHub. Do you have a preference?

vfdev-5:
As you wish :)

vfdev-5:
py3.5 : 1 failed on distrib
I restart github actions.

erip:
Specifically it failed on gloo. I found some relevant gloo issues, but I have enough information to say if it's a gloo bug or an ignite bug. 😄

erip:
More gloo errors. ☹️

vfdev-5:
Seems like that, let me restart it one more time.

PR #1014, opened 2020-05-05
[FR] Parallel helper tools

sdesrozis:
Description:

This PR aims to provide tools for parallel. It aims to handle tpu, gpu and cpu using backends from xla and dist (gloo for cpu and nccl for gpu).  

TODO:
- [x] Remove mnist distributed example
- [x] Revert unintended modifications
- [x] Add tests for `auto`
- [ ] Update new contrib/cifar10 example to have same features as the old one
  - [x] Resume from crash etc
     - [x] deterministic output path
  - [x] Automate everything
  - [ ] ~~replace python-fire by click ?~~
- [x] setup GPU CI to run cifar10
- [x] Remove old contrib/cifar10
- [x] Update docstrings
- [x] Update documentation

Check list:
* [x] New tests are added (if a new feature is added)
* [x] New doc strings: description and/or example code are in RST format
* [x] Documentation is updated (if required)

PR #1023, opened 2020-05-08
Improved parallel utils

### LINKED ISSUE #992: Ignite should expose the ability to run distributed jobs on TPUs
## 🚀 Feature

Because ignite relies on distribution to happen only when `dist.is_initialized()`, etc. are set and because xla has a different mechanism for distribution, distributed training won't work within ignite on TPUs. 

We'll need an abstraction within ignite to handle this for us. Things to consider:

1. How do we want the API to look?
1. How (if at all) do we handle mixed GPU/TPU training?
1. How do we let a user make the determination about devices to use for what operations?

vfdev-5:
PR to `idist` branch

Description:
- added `idist`
- computation models

Related to #992  #1014 and should enable #988 

TODO:
- documentation
- decide about `create_from_context` API
- decide on names: `ComputationModel`, `_SerialModel`, `_NativeDistModel`, `_XlaDistModel`

Check list:


vfdev-5:
@erip @sdesrozis please review

ignite/distributed/comp_models/native.py
```diff
@@ -0,0 +1,204 @@
...
+
+    name = "native-dist"
+    available_backends = tuple(
+        name
+        for name in [dist.Backend.NCCL, dist.Backend.GLOO, dist.Backend.MPI]
+        if getattr(dist, "is_{}_available".format(name))
+    )
+
+    @staticmethod
+    def create_from_context() -> Optional["_NativeDistModel"]:
```
sdesrozis, line 37: Returned type should be `Optional[ComputationModel]` to respect the `ComputationModel. create_from_context` type hint.

Same for `create_from_backend` below.

ignite/distributed/comp_models/native.py
```diff
@@ -0,0 +1,204 @@
...
+
+        os.environ["LOCAL_RANK"] = str(local_rank)
+        os.environ["RANK"] = str(node_rank * num_procs_per_node + local_rank)
+        os.environ["WORLD_SIZE"] = str(world_size)
+        os.environ["MASTER_ADDR"] = str(master_addr)
+        os.environ["MASTER_PORT"] = str(master_port)
+
+        model = _NativeDistModel.create_from_backend(backend)
+        _set_model(model)
+        fn(local_rank, *args)
```
sdesrozis, line 183: `local_rank` should not be mandatory in user function. `idiot.get_local_rank()` should the preferred way to get the local rank.

ignite/distributed/comp_models/xla.py
```diff
@@ -0,0 +1,116 @@
...
+    def finalize(self):
+        pass
+
+    @staticmethod
+    def _dist_worker_task_fn(local_rank, backend, fn, args):
+        from ignite.distributed.utils import _set_model
+
+        model = _XlaDistModel.create_from_backend(backend)
+        _set_model(model)
+        fn(local_rank, *args)
```
sdesrozis, line 100: Same remark than above about `local_rank` ;)

ignite/distributed/utils.py
```diff
@@ -0,0 +1,316 @@
...
+    - "nccl" or "gloo" or "mpi" for native torch distributed configuration
+    - "xla-tpu" for XLA distributed configuration
+
+    Returns:
+        str or None
+    """
+    return _model.backend()
+
+
+def available_backends() -> Tuple[str]:
```
sdesrozis, line 60: Here, "available" maybe be ambiguous : return a list of backends which could not be "available" in a sense of `dist`  (is_mpi_available for instance) or `xla`. Maybe it would be better to filter the list and return effectively
available backends ?

ignite/distributed/comp_models/native.py
```diff
@@ -0,0 +1,204 @@
...
+    - GLOO <-> CPU
+    - MPI  <-> CPU
+
+    """
+
+    name = "native-dist"
+    available_backends = tuple(
+        name
+        for name in [dist.Backend.NCCL, dist.Backend.GLOO, dist.Backend.MPI]
+        if getattr(dist, "is_{}_available".format(name))
```
sdesrozis, line 33: This returns a list of backends which have "is_{}_available" method even if `dist.is_{}_available() == False`. Missing `()` to call the method ?
```
if getattr(dist, "is_{}_available".format(name))()
```

ignite/distributed/comp_models/native.py
vfdev-5: As type hints they are more precise

ignite/distributed/comp_models/native.py
vfdev-5: We pass local rank to be compatible with torch multiprocessing: https://pytorch.org/docs/stable/multiprocessing.html#torch.multiprocessing.spawn

sdesrozis: Ok I understand even if I would have preferred that, by default, there was no constraint on the signature of the user method.

ignite/distributed/comp_models/native.py
sdesrozis: Too influenced by c++ 😅

sdesrozis:
I was thinking about mpi backend. With this backend, no local rank in env. The user will have to inform ignite in one way or another, and maybe he can’t 😔

Issue #996, opened 2020-04-30
Are there any ways to filter out or ignore a batch in engine?


CDWJustin:
## ❓ Questions/Help/Support

I'm really new to the ML area and I'm trying to train a network with a dataset that when sampling batches, it can create batches with really large sizes (the sampling method is just torch's weightedSampling). Though the batch size is fixed the data size in the batch will be really large sometimes. 
It seems pretty complicated to define a customized sampler, and the way I originally did was just ignore that batch when the data size in the batch is too large. Now, I'm trying to use ignite as the framework to train the network, I don't know how to do this.
I think it might relate to event_filter? But it seems to be triggered by event, so are there any ways to pass in a batch instead?

Thanks so much for your time!

sdesrozis:
@CDWJustin Thank you for your question !! 

I didn't catch 
> Though the batch size is fixed the data size in the batch will be really large sometimes.

So, I understand you want to ignore some batches during loop. I don't think we have this feature in engine. @vfdev-5 do you confirm ?

vfdev-5:
@CDWJustin currently, there is no way to skip easily an iteration in ignite.

With the current code base, we need to return the output anyway, so skipping a batch can be done as 
```python

def train_step(engine, batch):
     if not is_correct_batch(batch):
         # let's return the previous output => wont break other handlers using state.output
         # but metrics computation will be impacted... 
         return engine.state.output

     # otherwise, continue as if everyhing is OK

trainer = Engine(process_function)
```

In general, when skipping the batch you can output anything you would like: previous output (as in the example), zeros etc.


There is a similar request here : https://discuss.pytorch.org/t/how-to-make-ignite-trainer-handle-error-batch-instead-of-stop-the-program/75886

So, let's improve this part in ignite. I would suggest to have another method like `terminate_iteration()`, similar to existing `terminate_epoch()` or `terminate()`
```python
### Engine.__init__
        self.should_terminate = False
        self.should_terminate_single_epoch = False
        self.should_terminate_single_iteration = False

### Create new method in Engine
def terminate_iteration(self):
    self.should_terminate_single_iteration = True

### Engine._run_once_on_dataset
# https://github.com/pytorch/ignite/blob/master/ignite/engine/engine.py#L734
self._fire_event(Events.ITERATION_STARTED)
self.state.output = self._process_function(self, self.state.batch)
if self.should_terminate_single_iteration:
    self.should_terminate_single_iteration = False
    break
[...]

PR #1024, opened 2020-05-08
Ignore empty predictions in Accuracy calculation

oguzserbetci:
First of all, thanks for the package! I really enjoy using it. Since this was a pretty quick fix, I just post it as a PR.

Description:

I got errors coming from torch.argmax: 
`cannot perform reduction function argmax on a tensor with no elements because the operation does not have an identity`

Because I'm masking using indexing, training step sometime outputs empty predictions. I think it is more general to allow empty predictions in metric calculation.

Let me know if you want any tests covering the added code path. Furthermore, this idea can be applied to many other metrics. This is related to [pytorch issue #28380](https://github.com/pytorch/pytorch/issues/28380).


sdesrozis:
@oguzserbetci Thank you for your comment and this PR 😊

I agree you, tests are needed on that feature. Could you improve your PR ? And could you check others metrics ?

Thank you for your help 👍🏻

vfdev-5:
@oguzserbetci thanks for the PR ! 
I'm trying to understand your use-case: you compute accuracy metric on masked `y_pred` and there are evaluation steps such that `y_pred` becomes empty. In those cases, maybe we can try to skip out the iteration ? We have a FR on that : https://github.com/pytorch/ignite/issues/996
I think it would be better to work on this issue instead of making metrics work on empty data. 
@sdesrozis thoughts ?

oguzserbetci:
Good point. Skipping iteration was also something I wished for! However in my setup, I have many different outputs and that have different metrics each, so skipping iterations wouldn’t solve the problem. I actually started looking at pytorch code and will try to fix the problem there. Otherwise it will be a lot of code to add to metrics that will be rendered useless once it’s fixed in pytorch.

sdesrozis:
@vfdev-5 if this alternative matches the use case, of @oguzserbetci, I think it’s a good idea 👍🏻

vfdev-5:
@oguzserbetci okay I see, so your output of eval step is something like
```python
output = {
    "y_pred1": y_pred1[mask1],
    "y1": y1[mask1],
    "y_pred2" : y_pred2[mask2],
    "y2": y2[mask2],
}
```
and you compute various metrics with several output transforms
```python
def ot1(output):
    return output["y_pred1"], output["y1"]

def ot2(output):
    return output["y_pred2"], output["y2"]
```
and in some cases you can have `output["y_pred1"]` as an empty tensor ?

Maybe we can deal with that as following. We can make `Metric.iteration_completed`:
https://github.com/pytorch/ignite/blob/e3fc04e147db4c97a2ceb70597366d562e15c96f/ignite/metrics/metric.py#L123
skip `output=None`. So, in this case we skip a single output by Metric instead of complete iteration. In this case, you can check your output inside, for example
```python
def ot1(output):
    y_pred1, y1 = output["y_pred1"], output["y1"]
    if len(y_pred1) < 1:
        return None
   return y_pred1, y1
```

What do you think ?

oguzserbetci:
I like your approach, but this still seems pretty overkill for something that is actually to be fixed by pytorch. I will make a pull request there to have reduction over empty tensors behave similar to numpy. This would elavate the problem here as well...

However maybe skipping metrics is a cool thing to have in general anyways, in which case I like your solution!

vfdev-5:
@oguzserbetci sure that fix from pytorch side can be helpful. However, we should anyway make sure that thoses cases will correctly update internals of metrics like accuracy...

> However maybe skipping metrics is a cool thing to have in general anyways, in which case I like your solution!

Let's change this PR in this way :)

PR #1025, opened 2020-05-09
[WIP] trying to address flaky GLOO tests

erip:
Fixes #???

Description: the GLOO tests are quite flaky. This PR intends to address this.

Check list:

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,7 @@
 # Tests
 numpy
 pytest
+ifcfg
 codecov
 pytest-cov
 pytest-xdist
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -16,6 +16,7 @@ polyaxon-client
 mlflow
 neptune-client
 tensorboard
+pytest-rerunfailures
 pynvml; python_version > '3.5'
 trains
 # Examples dependencies
--- a/tests/ignite/conftest.py
+++ b/tests/ignite/conftest.py
@@ -5,12 +5,12 @@ import pytest
 import torch
 import torch.distributed as dist
 
-import pytest
-
 
 def get_ifname():
     import ifcfg
+
     return ifcfg.default_interface()["device"]
+
 
 @pytest.fixture()
 def dirname():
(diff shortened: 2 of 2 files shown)


erip:
Since this is very difficult to reproduce locally, this will mostly be experimental... no guarantee these commits will work.

sdesrozis:
Could it be related on zombi process remaining after stopping runs ? Sometimes it happens on my laptop...

erip:
It might be. The error is "connection refused" which might be related to a process trying to talk to a tcp port that is no longer alive

sdesrozis:
Maybe we don’t have a good stopping process. Missing a destroy process group call ? 🧐

vfdev-5:
I think maybe we can alter comm port to avoid the problems. Locally, I had some issue with fixed 2222 port. Changing it to random, seems help (locally).

@erip can you try this in conftest:

```python
def _find_free_port():
    # Taken from https://github.com/facebookresearch/detectron2/blob/master/detectron2/engine/launch.py
    import socket
    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    sock.bind(("", 0))
    port = sock.getsockname()[1]
    sock.close()
    return port


@pytest.fixture()
def distributed_context_single_node_nccl(local_rank, world_size):

    dist_info = {
        "backend": "nccl",
        "world_size": world_size,
        "rank": local_rank,
        "init_method": "tcp://localhost:{}".format(_find_free_port()),
    }

    dist.init_process_group(**dist_info)

    dist.barrier()

    torch.cuda.device(local_rank)

    yield {"local_rank": local_rank}

    dist.barrier()

    dist.destroy_process_group()


@pytest.fixture()
def distributed_context_single_node_gloo(local_rank, world_size):

    import os

    if "WORLD_SIZE" not in os.environ:
        os.environ["WORLD_SIZE"] = "1"

    dist_info = {
        "backend": "gloo",
        "world_size": world_size,
        "rank": local_rank,
        "init_method": "tcp://localhost:{}".format(_find_free_port()),
    }

    dist.init_process_group(**dist_info)

    dist.barrier()

    yield {"local_rank": local_rank}

    dist.barrier()

    dist.destroy_process_group()

```

PR #1032, opened 2020-05-12
[WIP] Create from context for XLA and Native

vfdev-5:
Description:
- added code to create comp model from context for XLA
  - pytorch xla provides local rank and thus we can setup everything
- for Native Dist model currently implemented fetching local rank from os env.
   - if not set => raise an issue. Probably, this should be reworked.

Idea is to cover the following case:
```python
# somewhere in the code

# check if in distrib setting
     pass

# 1) in background it tries to `_sync_model` if current model is `_SerialModel`
# 1.1) Check _NativeDistModel if we can create from current context
# 1.2) Check _XlaDistModel if we can create from current context
# 2) calls _model.get_world_size()
```

Check list:


vfdev-5:
Accidentally merged into origin/idist :(

PR #1043, opened 2020-05-16
Allow TrainsSaver to be initialized with no logger

### LINKED ISSUE #1039: TrainsSaver requires TrainsLogger but doesn't use it
## ❓ Questions/Help/Support

Hi,

I'm opening this issue as a question, because I'm not sure if it's an artifact or it has an intended purpose that I don't understand.

The new trains support introduced `TrainsSaver` together with `TrainsLogger`.  It seems that to initialize the `TrainsSaver` object, you're required to pass a `TrainsLogger` logger (https://github.com/pytorch/ignite/blob/master/ignite/contrib/handlers/trains_logger.py#L610).
However, this logger is not used anywhere. Why is it required?

I would expect to be able to just do 
```
Checkpoint(to_save, TrainsSaver(output_uri=output_uri), 'epoch', n_saved=n_saved,
                       global_step_transform=global_step_from_engine(trainer))
```

jkhenning:
Fixes #1039 

Description:
Allow creating a TrainsSaver instance with no `logger` argument, assuming the user has already initialized a Trains Task using `trains.Task.init()`. If a Trains Task was not initialized, show a runtime error.

Check list:


vfdev-5:
@jkhenning seems like `test_trains_disk_saver_integration` is broken now:
https://travis-ci.org/github/pytorch/ignite/jobs/687746212#L1952

PR #1045, opened 2020-05-16
[WIP] Merge idist into master

### LINKED ISSUE #965: Metrics reduction on distributed TPU setting

### LINKED ISSUE #991: RunningAverage should average output value instead of sum
## 🐛 Bug description

Current implementation of `RunningAverage` in DDP only sums the value of output quantity (e.g. loss), however it would be better to compute the average:

https://github.com/pytorch/ignite/blob/6ec32e36d9098dfa5f055bbc582911b7189766fc/ignite/metrics/running_average.py#L114

## Environment

 - OS (e.g., Linux):
 - Python version:

### LINKED ISSUE #992: Ignite should expose the ability to run distributed jobs on TPUs

### LINKED ISSUE #999: TPU and Checkpoint

### LINKED ISSUE #1011: Make setup_common_training_handlers less dependent on torch.dist DDP
## 🚀 Feature

Current implementation of https://github.com/pytorch/ignite/blob/bef668c152dc86334a8ab09e9ce9368c48e48102/ignite/contrib/engines/common.py#L25

makes a lot of assumption about distributed type (=torch.dist). In case of TPU dist, some of provided bricks are not applicable or lead to undesired exceptions.

What to improve:

- cover TPU dist config
https://github.com/pytorch/ignite/blob/bef668c152dc86334a8ab09e9ce9368c48e48102/ignite/contrib/engines/common.py#L79

- make this optional
https://github.com/pytorch/ignite/blob/bef668c152dc86334a8ab09e9ce9368c48e48102/ignite/contrib/engines/common.py#L107

- check this case if scheduler is [LRScheduler](https://github.com/pytorch/ignite/blob/master/ignite/contrib/handlers/param_scheduler.py#L555)
https://github.com/pytorch/ignite/blob/bef668c152dc86334a8ab09e9ce9368c48e48102/ignite/contrib/engines/common.py#L113

- this is not applicable for CPU or TPU dist
https://github.com/pytorch/ignite/blob/bef668c152dc86334a8ab09e9ce9368c48e48102/ignite/contrib/engines/common.py#L115

- replace this by `Checkpoint`
https://github.com/pytorch/ignite/blob/bef668c152dc86334a8ab09e9ce9368c48e48102/ignite/contrib/engines/common.py#L120

vfdev-5:
- Fixes #965
- Fixes #992
- Fixes #991
- Fixes #999
- Fixes #1011

Todo:
 
- [x] Ensure that no args were removed to keep BC.
- [x] Make `idist.device()` output uniform : `torch.device`, #1062
- [x] Add kwargs to spawned function: #1064
- [x] Improved device and setup_common_training_handlers: https://github.com/pytorch/ignite/pull/1066
- [x] Checkpoint and loggers should work on XLA, #1068 
- [ ] ~~tests with `parallel_api` on cifar10~~

Description:
Introduces distributed module to handle GPU, CPU, XLA distributed config.

Check list:


ignite/distributed/__init__.py
```diff
@@ -0,0 +1 @@
+from ignite.distributed.utils import *
```
sdesrozis, line 1: I did this during prototyping. I think it’s not a good idea to import all functions (and it’s not done in ignite).

ignite/distributed/comp_models/native.py
```diff
@@ -0,0 +1,290 @@
...
+        if any(all_env_vars_defined) and not all(all_env_vars_defined):
+            raise RuntimeError(
+                "PyTorch distributed configuration should define env variables '{}'".format(necessary_env_vars)
+            )
+
+        os.environ["RANK"] = os.environ.get("RANK", "0")
+        os.environ["LOCAL_RANK"] = os.environ.get("LOCAL_RANK", "0")
+        os.environ["WORLD_SIZE"] = os.environ.get("WORLD_SIZE", "1")
+        os.environ["MASTER_ADDR"] = os.environ.get("MASTER_ADDR", "127.0.0.1")
+        os.environ["MASTER_PORT"] = os.environ.get("MASTER_PORT", "15000")
```
sdesrozis, line 176: Should we check if the port is opened?

ignite/handlers/checkpoint.py
```diff
@@ -401,6 +403,14 @@ def __call__(self, checkpoint: Mapping, filename: str, metadata: Optional[Mappin
                 tmp.close()
                 os.rename(tmp.name, path)
 
+    def _save(self, checkpoint: Mapping, filename: str):
+        if idist.has_xla_support:
```
sdesrozis, line 407: It could be in models ?

ignite/distributed/__init__.py
vfdev-5: We have `__all__` inside `ignite.distributed.utils`. I think it is OK.

ignite/distributed/comp_models/native.py
vfdev-5: Honestly, I do not know. Maybe.

ignite/handlers/checkpoint.py
vfdev-5: Which models you mean ?

sdesrozis: CompModels. Like barrier, reduction. Maybe not a good idea.

ignite/distributed/comp_models/native.py
sdesrozis: Ok, I don’t know too, so let’s see 😉

ignite/distributed/__init__.py
sdesrozis: > Do you suggest to import manually what is helpful ?
Similar to others modules. But that is not really important. I thought it was not a good practice to import all.

ignite/handlers/checkpoint.py
vfdev-5: Well, I'm not sure that `idist` should replace all possible ops. IMO, specific code should be specific...

vfdev-5:
How about `idist.device()` returns "torch.device('cuda')" if non-dist conf and cuda device is available ?

ignite/contrib/engines/common.py
```diff
@@ -75,9 +81,11 @@ def setup_common_training_handlers(
         with_pbars=with_pbars,
         with_pbar_on_iters=with_pbar_on_iters,
         log_every_iters=log_every_iters,
-        device=device,
+        stop_on_nan=stop_on_nan,
     )
-    if dist.is_available() and dist.is_initialized():
+    from ignite.distributed.utils import _SerialModel
+
+    if idist.model_name() != _SerialModel.name:
```
vfdev-5, line 88: Maybe, this better should be : 
```suggestion
    if idist.get_world_size() > 1:
```

ignite/contrib/engines/common.py
```diff
@@ -75,9 +81,11 @@ def setup_common_training_handlers(
...
         _setup_common_distrib_training_handlers(trainer, train_sampler=train_sampler, **kwargs)
     else:
         if train_sampler is not None:
```
vfdev-5, line 91: Warn only if `train_sampler` has `set_epoch`

vfdev-5:
Something makes things stuck on TPUs in `parallel_api` PR. Need to investigate where is the problem, before merging this PR.

ignite/contrib/engines/common.py
```diff
@@ -63,8 +65,12 @@ def setup_common_training_handlers(
         with_pbar_on_iters (bool, optional): if True, a progress bar on iterations is attached to the trainer.
         log_every_iters (int, optional): logging interval for :class:`~ignite.contrib.metrics.handlers.GpuInfo` and for
             epoch-wise progress bar.
-        device (str of torch.device, optional): Optional device specification in case of distributed computation usage.
+        stop_on_nan (bool, optional): if True, :class:`~ignite.handlers.TerminateOnNan` handler is added to the trainer.
+        device (str of torch.device, optional): deprecated argument, it will be removed since v0.5.0.
     """
+    if device is not None:
+        warnings.warn("Argument device is unused and deprecated. It will be removed since v0.5.0")
```
erip, line 72: Nit: It will be removed in v0.5.0

ignite/contrib/engines/common.py
```diff
@@ -75,12 +81,13 @@ def setup_common_training_handlers(
...
-        device=device,
+        stop_on_nan=stop_on_nan,
     )
-    if dist.is_available() and dist.is_initialized():
+
+    if idist.get_world_size() > 1:
         _setup_common_distrib_training_handlers(trainer, train_sampler=train_sampler, **kwargs)
     else:
-        if train_sampler is not None:
+        if train_sampler is not None and hasattr(train_sampler, "set_epoch"):
```
erip, line 90: `hasattr`, `setattr`, and `getattr` make me a little nervous - is there a way to replace this? I haven't finished looking at the code, so maybe it's necessary...

ignite/contrib/engines/common.py
```diff
@@ -98,28 +105,34 @@ def _setup_common_training_handlers(
...
     if lr_scheduler is not None:
         if isinstance(lr_scheduler, torch.optim.lr_scheduler._LRScheduler):
             trainer.add_event_handler(Events.ITERATION_COMPLETED, lambda engine: lr_scheduler.step())
+        elif isinstance(lr_scheduler, LRScheduler):
+            trainer.add_event_handler(Events.ITERATION_COMPLETED, lr_scheduler)
         else:
             trainer.add_event_handler(Events.ITERATION_STARTED, lr_scheduler)
 
-    trainer.add_event_handler(Events.EPOCH_COMPLETED, empty_cuda_cache)
+    if torch.cuda.is_available():
```
erip, line 126: It seems like we should let a user opt into this. Clearing the cache can mean huge performance hits in some situations.

ignite/contrib/handlers/mlflow_logger.py
```diff
@@ -287,10 +283,7 @@ def __getattr__(self, attr):
 
         import mlflow
 
-        def wrapper(*args, **kwargs):
-            return getattr(mlflow, attr)(*args, **kwargs)
-
-        return wrapper
+        return getattr(mlflow, attr)
```
erip, line 286: Same comment as above - would be nice to replace this if we can.

ignite/contrib/handlers/neptune_logger.py
```diff
@@ -478,10 +479,7 @@ def __getattr__(self, attr):
 
         import neptune
 
-        def wrapper(*args, **kwargs):
-            return getattr(neptune, attr)(*args, **kwargs)
-
-        return wrapper
+        return getattr(neptune, attr)
```
erip, line 482: And here.

ignite/contrib/handlers/polyaxon_logger.py
```diff
@@ -271,10 +267,7 @@ def __init__(self, *args, **kwargs):
         self.experiment = Experiment(*args, **kwargs)
 
     def __getattr__(self, attr):
-        def wrapper(*args, **kwargs):
-            return getattr(self.experiment, attr)(*args, **kwargs)
-
-        return wrapper
+        return getattr(self.experiment, attr)
```
erip, line 270: And here.

erip:
My requested changes are fairly superficial and it looks like it's mostly for experiment loggers rather than in the "core" code; this is a very impressive PR and kudos to you and @sdesrozis for working so hard. 👏

PR #1062, opened 2020-05-22
Improved device() method

vfdev-5:
Description:
- Make idist.device() output uniform : torch.device

Check list:

### DIFF
--- a/ignite/distributed/comp_models/base.py
+++ b/ignite/distributed/comp_models/base.py
@@ -143,9 +143,6 @@ class _SerialModel(ComputationModel):
     def get_node_rank(self) -> int:
         return 0
 
-    def is_distributed(self) -> bool:
-        return False
-
     def device(self) -> torch.device:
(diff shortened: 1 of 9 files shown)


vfdev-5:
We can ignore codecov for a moment as we do not merge all reports.

ignite/distributed/comp_models/base.py
```diff
@@ -143,9 +143,6 @@ def get_num_nodes(self) -> int:
     def get_node_rank(self) -> int:
         return 0
 
-    def is_distributed(self) -> bool:
```
erip, line 146: Question: why was this removed? There may be a good reason, but I think now the contract enforced by the interface is more implicit.

vfdev-5: @erip it was discussed in a gdoc about idist API and we thought that it is sort of redundant and what to call `distributed`... So, now we check `get_world_size() > 1`.

PR #1064, opened 2020-05-23
Idist kwargs dict

vfdev-5:
Description:
- Added kwargs to spawned function: `func(lrank, *args, **kwargs)`
- updated usage of `_need_to_sync` with spawn

Check list:

### DIFF
--- a/ignite/distributed/utils.py
+++ b/ignite/distributed/utils.py
@@ -167,7 +167,14 @@ def hostname() -> str:
     return socket.gethostname()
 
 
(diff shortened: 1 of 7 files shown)


ignite/distributed/utils.py
```diff
@@ -167,7 +167,7 @@ def hostname() -> str:
     return socket.gethostname()
 
 
-def spawn(backend, fn, args, num_procs_per_node, **kwargs):
+def spawn(backend: str, fn: Callable, args: Tuple, kwargs_dict: Mapping = None, num_procs_per_node: int = 1, **kwargs):
```
sdesrozis, line 170: `kwargs_dict` type should be `Optional`.

PR #987, opened 2020-04-26
[WIP] add typing info to contrib.

erip:
Fixes #977

Description: Adds type hints and documentation corrections to contrib module.

Check list:
* [ ] New tests are added (if a new feature is added)
* [x] New doc strings: description and/or example code are in RST format
* [x] Documentation is updated (if required)

PR #1067, opened 2020-05-24
Improve typing (especially for **kwargs)

sdesrozis:
Description:

Improve typing and add `Any` for `*args` and `**kwargs`.

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -5,7 +5,7 @@ import tempfile
 import warnings
 from abc import ABCMeta, abstractmethod
 from collections import namedtuple
-from typing import Callable, Mapping, Optional, Union
+from typing import Callable, Mapping, Optional, Union, Any
 
 import torch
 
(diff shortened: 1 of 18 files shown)


vfdev-5:
Seems like overlaps with https://github.com/pytorch/ignite/pull/987

ignite/handlers/checkpoint.py
```diff
@@ -57,8 +57,8 @@ class Checkpoint:
     storage, etc.
 
     Args:
-        to_save (Mapping): Dictionary with the objects to save. Objects should have implemented `state_dict` and `
-            load_state_dict` methods.
+        to_save (Mapping, optional): Dictionary with the objects to save. Objects should have implemented `state_dict`
```
vfdev-5, line 60: It is not optional and should not be for public API, but for refactoring sake with `ModelCheckpoint`, it can accept None.

sdesrozis:
Ok so maybe we can close this PR and let @erip work on #987.

ignite/handlers/checkpoint.py
sdesrozis: Ok i thought it was a missing optional.

PR #1068, opened 2020-05-26
Fix #1055

### LINKED ISSUE #1055: Use decorator one_rank_only on Checkpoint and BaseLogger
## 🚀 Feature

Handlers may not be parallel and be triggered on the master proc (rank=0). The idea is to use the decorator `one_rank_only` (in branch `idist`) on `__call__` methods or specific working methods for `Checkpoint` and `BaseLogger`.

sdesrozis:
Fixes #1055 

Description:

Use decorator `@one_rank_only()` for `Checkpoint` and `Loggers`.
For `Loggers`, decoration is used on `attach` of base class, `__init__` and `close` of concrete loggers.
For `Checkpoint`, decoration is used on `__call__` and `__init__`. 

Check list:

### DIFF
--- a/ignite/contrib/handlers/base_logger.py
+++ b/ignite/contrib/handlers/base_logger.py
@@ -7,6 +7,7 @@ import torch
 
 from ignite.engine import Engine, State
 from ignite.handlers import global_step_from_engine
+from ignite.distributed import one_rank_only
 
 
 class BaseHandler(metaclass=ABCMeta):
@@ -136,6 +137,7 @@ class BaseLogger(metaclass=ABCMeta):
 
     """
 
+    @one_rank_only()
     def attach(self, engine, log_handler, event_name):
         """Attach the logger to the engine and execute `log_handler` function at `event_name` events.
 
@@ -155,6 +157,7 @@ class BaseLogger(metaclass=ABCMeta):
 
         return engine.add_event_handler(event_name, log_handler, self, name)
 
+    @one_rank_only()
     def attach_output_handler(self, engine: Engine, event_name: Any, *args: Any, **kwargs: Mapping):
         """Shortcut method to attach `OutputHandler` to the logger.
 
@@ -170,6 +173,7 @@ class BaseLogger(metaclass=ABCMeta):
         """
         return self.attach(engine, self._create_output_handler(*args, **kwargs), event_name=event_name)
 
+    @one_rank_only()
     def attach_opt_params_handler(self, engine: Engine, event_name: Any, *args: Any, **kwargs: Mapping):
         """Shortcut method to attach `OptimizerParamsHandler` to the logger.
 
--- a/ignite/contrib/handlers/base_logger.py
+++ b/ignite/contrib/handlers/base_logger.py
@@ -135,6 +135,11 @@ class BaseLogger(metaclass=ABCMeta):
     """
     Base logger handler. See implementations: TensorboardLogger, VisdomLogger, PolyaxonLogger, MLflowLogger, ...
 
+    Note:
+        This class is distributed configuration-friendly: it is not required to instantiate the class in rank 0 only
+        process. This class and its derived classes support automatically distributed configuration and perform logging
+        operations on rank 0 only.
+
     """
 
     @one_rank_only()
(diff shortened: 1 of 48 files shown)


sdesrozis:
Error from travis related to tpu... Help on this is welcome !

ignite/contrib/handlers/wandb_logger.py
```diff
@@ -301,6 +302,7 @@ def score_function(engine):
                 Events.COMPLETED, model_checkpoint, {'model': model})
     """
 
+    @one_rank_only()
```
vfdev-5, line 305: Maybe we need also to decorate 
```python
    def __getattr__(self, attr):
        return getattr(self._wandb, attr)
```
as it can be potentially used

ignite/contrib/handlers/mlflow_logger.py
```diff
@@ -267,6 +268,7 @@ class MLflowLogger(BaseLogger):
             )
     """
 
+    @one_rank_only()
```
vfdev-5, line 271:
```
def __getattr__(self, attr):
```
We need to decorate this method too

ignite/handlers/checkpoint.py
```diff
@@ -386,6 +388,7 @@ def __init__(
                     "".format(matched, dirname)
                 )
 
+    @one_rank_only()
```
vfdev-5, line 391: We need to rework `_save` method here:
According to the code of `xm.save`, it does barrier in the end of the method: https://github.com/pytorch/xla/blob/master/torch_xla/core/xla_model.py#L632
So, if we do one rank only, it will get stuck.

ignite/contrib/handlers/wandb_logger.py
sdesrozis: Done. The `wandb` version seems better.

ignite/handlers/checkpoint.py
sdesrozis: Ok so we can't decorate and we have to do conditional branches (already done)
```python
def _save(self, checkpoint: Mapping, filename: str):
    if idist.has_xla_support:

        xm.save(checkpoint, filename)
    elif idist.get_rank() == 0:
        torch.save(checkpoint, filename)
```
The `atomic` mode should be checked with multinode `xla` (work with a file per node? file per process?).

ignite/contrib/handlers/base_logger.py
```diff
@@ -155,6 +157,7 @@ def attach(self, engine, log_handler, event_name):
 
         return engine.add_event_handler(event_name, log_handler, self, name)
 
+    @one_rank_only()
```
vfdev-5, line 160: Is it necessary to decorate if it calls `attach` inside ?

ignite/contrib/handlers/base_logger.py
```diff
@@ -170,6 +173,7 @@ def attach_output_handler(self, engine: Engine, event_name: Any, *args: Any, **k
         """
         return self.attach(engine, self._create_output_handler(*args, **kwargs), event_name=event_name)
 
+    @one_rank_only()
```
vfdev-5, line 176: Is it necessary to decorate if it calls `attach` inside ?

vfdev-5:
There can be an issue with decorating logger's `__init__`:
```python
# distrib config:
tb_logger = TensorboardLogger(".")
tb_logger.writer.add_image(...)  # this will work for rank 0 only, others -> raise attribute error
```
for other classes who expose `__getattr__` we will have recursion error when calling logger with any member of corresponding logging system.

sdesrozis:
It’s not a good idea to use members directly. `getattr` allows control, it seems good. I would have preferred to do not use members directly by in smilar way than `wandb` logger.

sdesrozis:
@vfdev-5 Thank you to have done this work on `Checkpoint`. I can test integration of `Loggers` using new fixture 😊

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -742,7 +752,8 @@ def update_fn(engine, batch):
         assert key in loaded_optimizer_state_dict
         optim_value = optim_state_dict[key]
         loaded_optim_value = loaded_optimizer_state_dict[key]
-        assert optim_value == loaded_optim_value
+        if idist.get_rank() == 0:
```
vfdev-5, line 755: @sdesrozis why do you check for rank == 0 here ?
I have an impression that it works for all ranks ?
cc @sadra-barikbin

PR #1099, opened 2020-06-03
Fixed incorrect log position. (#1098)

### LINKED ISSUE #1098: Incorrect log position
## 🐛 Bug description

When `{bar}` is removed from `bar_format` of argument of `ProgressBar`, immediate logger output does not print in newline.

Screen Shot 2020-06-03 at 3 58 28 PM

Add 

```
import logging

logging.basicConfig(level=logging.INFO)
logger = logging.getLogger(__name__)
```

to beginning of [mnist_with_tqdm_logger.py](https://github.com/pytorch/ignite/blob/master/examples/contrib/mnist/mnist_with_tqdm_logger.py) and change this [line](https://github.com/pytorch/ignite/blob/master/examples/contrib/mnist/mnist_with_tqdm_logger.py#L70) to 

```
    pbar = ProgressBar(
        persist=True, bar_format="{desc}[{n_fmt}/{total_fmt}] {percentage:3.0f}%|{postfix} [{elapsed}<{remaining}]",
    )
```

This seems not happen to very small example

```py
import logging
import time

import tqdm

logging.basicConfig(level=logging.INFO)
logger = logging.getLogger(__name__)

logger.info("before")

for i in tqdm.tqdm(
    range(3),
    leave=True,
    bar_format="{desc}[{n_fmt}/{total_fmt}] {percentage:3.0f}%|{postfix} [{elapsed}<{remaining}]",
):
    time.sleep(0.5)

logger.info("after")
```

Screen Shot 2020-06-03 at 4 08 55 PM

Note that the  `leave` argument in `tqdm` corresponds to the `persist` in `ProgressBar`.

## Environment

 - PyTorch Version (e.g., 1.4): 1.5.0
 - Ignite Version (e.g., 0.3.0): `2d30d1d`
 - OS (e.g., Linux): macOS 10.15.5
 - How you installed Ignite (`conda`, `pip`, source): source
 - Python version: 3.7.5
 - Any other relevant information: None

Yevgnen:
Fixes #1098

Description:

Fire all events before logging time usages to prevent incorrect log position.

Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -698,11 +698,18 @@ class Engine(Serializable):
                 time_taken = self._run_once_on_dataset()
                 self.state.times[Events.EPOCH_COMPLETED.name] = time_taken
                 hours, mins, secs = _to_hours_mins_secs(time_taken)
+                elapsed_time_message = "Epoch[%s] Complete. Time taken: %02d:%02d:%02d" % (
+                    self.state.epoch,
+                    hours,
+                    mins,
+                    secs,
+                )
                 if self.should_terminate:
                     self._fire_event(Events.TERMINATE)
+                    self.logger.info(elapsed_time_message)
                     break
                 self._fire_event(Events.EPOCH_COMPLETED)
-                self.logger.info("Epoch[%s] Complete. Time taken: %02d:%02d:%02d", self.state.epoch, hours, mins, secs)
+                self.logger.info(elapsed_time_message)
 
             time_taken = time.time() - start_time
             hours, mins, secs = _to_hours_mins_secs(time_taken)
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -698,11 +698,11 @@ class Engine(Serializable):
                 time_taken = self._run_once_on_dataset()
                 self.state.times[Events.EPOCH_COMPLETED.name] = time_taken
                 hours, mins, secs = _to_hours_mins_secs(time_taken)
-                self.logger.info("Epoch[%s] Complete. Time taken: %02d:%02d:%02d", self.state.epoch, hours, mins, secs)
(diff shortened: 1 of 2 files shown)


ignite/engine/engine.py
```diff
@@ -698,11 +698,11 @@ def _internal_run(self) -> State:
                 time_taken = self._run_once_on_dataset()
                 self.state.times[Events.EPOCH_COMPLETED.name] = time_taken
                 hours, mins, secs = _to_hours_mins_secs(time_taken)
-                self.logger.info("Epoch[%s] Complete. Time taken: %02d:%02d:%02d", self.state.epoch, hours, mins, secs)
                 if self.should_terminate:
                     self._fire_event(Events.TERMINATE)
                     break
                 self._fire_event(Events.EPOCH_COMPLETED)
+                self.logger.info("Epoch[%s] Complete. Time taken: %02d:%02d:%02d", self.state.epoch, hours, mins, secs)
```
vfdev-5, line 705: Maybe, we also need to call it after `self._fire_event(Events.TERMINATE)` with message "Epoch[%s] Terminated. ..."

vfdev-5:
@Yevgnen thanks for the PR !
I think we need to readjust some tests...

Yevgnen:
@vfdev-5 Following the contributing guide, I try to re-run the tests, but got 

```
tests/run_cpu_tests.sh: line 5: CI_PYTHON_VERSION: unbound variable
```

vfdev-5:
Yes, this should be a forgotten to mention :) Thanks for pointing out !
Please, just set it as `export CI_PYTHON_VERSION=<your python version, e.g 3.7>`. Otherwise, you can simply run the failing test as : 
```
pytest tests/ignite/test_utils.py::test_setup_logger -vvv
```

PR #1101, opened 2020-06-03
Add missing colon to logging message

willfrey:
Noticed that the logs weren't aligned. Added a colon to fix that.


vfdev-5:
@willfrey thanks !

PR #1105, opened 2020-06-04
Fixes #1096 - bad XLA comp model configuration on single device

### LINKED ISSUE #1096: Training on TPU with threading
## ❓ Questions/Help/Support

Currently I am trying to train multiple folds of my dataset on each TPU core using threading but ignite has issues where the one thread either takes preference over others or causes my training to crash.

I created two simple versions to show an ignite version of MNIST on multiple folds vs a regular pytorch version as an example
- [Ignite TPU multi-fold]( https://colab.research.google.com/gist/ryanwongsa/1ce9618c5f900122513a24a5739ad02a/ignite_on_tpus.ipynb)
- [Pytorch TPU multi-fold](https://colab.research.google.com/gist/ryanwongsa/383c5daa16d3193c53b250b62395eb1b/pytorch.ipynb)

The pure Pytorch version doesn't seem to have any issues but ignite version seems inconsistent. I think it how the metrics work on multiple threads. So was wondering if there is any easy way to solve this issue?

vfdev-5:
Fixes #1096 

Description:
- handle properly single XLA device
- do not all reduce metrics if world size < 2
- minor improvements for comp models

Check list:

### DIFF
--- a/ignite/distributed/comp_models/base.py
+++ b/ignite/distributed/comp_models/base.py
@@ -21,6 +21,18 @@ class ComputationModel(metaclass=ABCMeta):
         self._nnodes = None
         self._node = None
 
+    def _setup_attrs(self):
+        if self._ntasks_per_node is None:
+            self._ntasks_per_node = self._compute_ntasks_per_node() if self.get_world_size() > 1 else 1
+        if self._nnodes is None:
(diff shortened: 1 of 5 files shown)


vfdev-5:
Seems like threading approach is not fully recommended one : https://github.com/pytorch/xla/issues/2171#issuecomment-639065832

PR #1116, opened 2020-06-09
[FR] Parallel helper tools (#1014)

vfdev-5:
Description:
- Merge #1014 into master 

- Checkout CIFAR10 example reworked with `auto` and `Parallel`

Check list:


vfdev-5:
@erip @sdesrozis please review ! We can not waste user's time :)

ignite/handlers/checkpoint.py
```diff
@@ -377,13 +385,17 @@ def load_objects(to_load: Mapping, checkpoint: Mapping, **kwargs) -> None:
             # single object and checkpoint is directly a state_dict
             key, obj = list(to_load.items())[0]
             if key not in checkpoint:
+                if isinstance(obj, (nn.DataParallel, nn.parallel.DistributedDataParallel)):
```
sdesrozis, line 388: That's just a comment but  is not useless if user puts `obj.module` in mapping. What happens if someone really wants to save DDP ? 
Note that `type` methods discussed on slack ignite channel should work ;)

ignite/distributed/auto.py
```diff
@@ -0,0 +1,276 @@
...
+            logger.info("Apply torch DistributedDataParallel on model, device id: {}".format(lrank))
+            model = torch.nn.parallel.DistributedDataParallel(model, device_ids=[lrank,])
+        elif idist.backend() == idist_native.GLOO:
+            logger.info("Apply torch DistributedDataParallel on model")
+            model = torch.nn.parallel.DistributedDataParallel(model)
+
+    # not distributed but multiple GPUs reachable so data parallel model
+    elif torch.cuda.device_count() > 1 and "cuda" in idist.device().type:
+        logger.info("Apply torch DataParallel on model")
+        model = torch.nn.parallel.DataParallel(model)
```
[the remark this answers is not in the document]
vfdev-5: It maybe also a wanted behaviour. If training step is complicated and it is more difficult to setup processing group with DDP etc

ignite/handlers/checkpoint.py
vfdev-5: Well, with actual behaviour it is not possible. But if we are using `auto_model` it is not garanteed that model is DDP, so user needs to check the type before calling `.module`.

> Note that type methods discussed on slack ignite channel should work ;)

I do not see how it could work. Anyway, using `type` to create new classes is also rather obscure way of patching things :)

ignite/distributed/launcher.py
```diff
@@ -0,0 +1,265 @@
...
+            # main.py
+
+            import ignite.distributed as idist
+
+            def training(local_rank, config, **kwargs):
+                # ...
+                print(idist.get_rank(), ": run with config:", config, "- backend=", idist.backend())
+                # ...
+
+            dist_config = {
```
sdesrozis, line 126: Why do not provide `torch.distributed.launch` API ? I mean `num_procs_per_node` => `nproc_per_node` etc.

vfdev-5: This is because of `idist.spawn` args. Maybe we could make it more uniform between `num_procs_per_node`, `nproc_per_node` and even `idist.get_ntasks_per_node()`

ignite/handlers/checkpoint.py
sdesrozis: > I do not see how it could work. Anyway, using `type` to create new classes is also rather obscure way of patching things :)

override `load_dict_state` to return module one

vfdev-5: Let's avoid such monkey-patching :)

Issue #1056, opened 2020-05-20
TrainsSaver doesn't respect Checkpoint's n_saved


achigeor:
## 🐛 Bug description

As the title says, it seems that `TrainsSaver` bypasses the `Checkpoint` `n_saved` parameter. That means that all models are saved and never updated / deleted.

Consider this simple example:
```
        task.phases['train'].add_event_handler(
            Events.EPOCH_COMPLETED(every=1),
            Checkpoint(to_save, TrainsSaver(output_uri=output_uri), 'epoch', n_saved=1,
                       global_step_transform=global_step_from_engine(task.phases['train'])))
```
The above saves every checkpoint. You end-up with
```
epoch_checkpoint_1.pt
epoch_checkpoint_2.pt
epoch_checkpoint_3.pt
...
```

Now if we do, the same with `DiskSaver`: 
```
        task.phases['train'].add_event_handler(
            Events.EPOCH_COMPLETED(every=1),
            Checkpoint(to_save, DiskSaver(dirname=dirname), 'epoch', n_saved=1,
                       global_step_transform=global_step_from_engine(task.phases['train'])))
```
We get only:
```
epoch_checkpoint_3.pt
```
as expected.

Same behaviour if we save only best models using `score_function`, i.e. `TrainsSaver` saves every best model.

## Environment

 - PyTorch Version: 1.3.1
 - Ignite Version: 0.4.0.dev20200519  (EDIT: update to latest nightly, issue still exists)
 - OS: Linux
 - How you installed Ignite: pip nightly
 - Python version: 3.6
 - Any other relevant information: trains version: 0.14.3

vfdev-5:
Thanks for the report @achigeor ! Let us reproduce the issue and have deeper look on where is the problem.

cc @jkhenning

vfdev-5:
Checked the number of temporary saved checkpoints and only one remains in temp folder. Probably, the problem is that we need to explicitly remove them from Trains' storage: override 
remove method
https://github.com/pytorch/ignite/blob/0452e4199ae0da627274f563a901493d9654b80d/ignite/handlers/checkpoint.py#L370-L372

@jkhenning do you have an API to remove stored artifact ?

jkhenning:
Hi @achigeor ,

When you say you see:
```
epoch_checkpoint_1.pt
epoch_checkpoint_2.pt
epoch_checkpoint_3.pt
...
```
[...]

PR #1117, opened 2020-06-10
reverse order of remove/save in Checkpoint handling

erip:
Description: Currently there can be a situation in which the number of saved checkpoints is technically `> n_saved` temporarily. This PR addresses this by forcing eviction before saving.

Related to #1056

Check list:


erip:
I think eventually it would be nice to move to an implementation with a priority queue structure implemented so these checks happen by virtue of properties of the data structure; for now, this is probably ok.

vfdev-5:
@erip PR looks OK. Maybe method name `_will_be_full`  is a bit surprising...
However, it would be also nice to switch another thing :
-    updated internal _saved before
-    save a checkpoint file

https://github.com/vfdev-5/ignite/blob/19f40aae1ca53167ba4bfed00477b104ceb6068b/ignite/handlers/checkpoint.py#L261

ignite/handlers/checkpoint.py
```diff
@@ -257,6 +257,11 @@ def _check_lt_n_saved(self, or_equal=False):
             return True
         return len(self._saved) < self._n_saved + int(or_equal)
 
+    def _will_be_full(self):
```
vfdev-5, line 260: Can't we reuse `_check_lt_n_saved` instead of `_will_be_full` ?

erip: No, I don't think so because we need to explicitly check equality. If the next element will overflow the queue (i.e., `if len(self._saved) == self._n_saved` before the update), then we need to pop (if necessary). Since `_check_lt_n_saved` only checks for lt or lte. This code will hopefully eventually be replaced by some `heapq` implementation, so inconsistency (in private API) is unfortunate but not the worst.

vfdev-5: Well, not exactly. We can append new item (line 319), then check if need to drop the worst checkpoint and finally save new checkpoint

erip: Isn't that exactly what this PR seeks to fix? 😄

vfdev-5: This PR seeks to fix the order of operations without introducing redundant functions :)

erip: The process you just described (append, check, drop) is exactly what is in master now and has been deemed problematic because it admits a case of `n+1` elements in `Checkpoint._saved`. I might be misunderstanding what you mean, but checking first if the `_saved` is going to overflow requires an equality check (not <=).

vfdev-5: Let me give a simplified example of the code from above link and what I think we would like

<details>

```python
from collections import namedtuple


_n_saved = 3
_saved = []


def _check_lt_n_saved(or_equal=False):
    if _n_saved is None:
        return True
    return len(_saved) < _n_saved + int(or_equal)


Item = namedtuple("Item", ["priority", "filename"])

for priority in [0, 1, 2, 1, 3, 5, 6, 7]:
    print("\n-", priority)

    if _check_lt_n_saved() or _saved[0].priority < priority:

        filename = "check_{}".format(priority)
        if any(item.filename == filename for item in _saved):
            print("Found the same filename")
            continue

        # When we setup the checkpoint, Checkpoint's state dict should be up to date as we add new item
        # and remove optionally other items.
        _saved.append(Item(priority, filename))        
        _saved.sort(key=lambda item: item[0])
        print("-- Appended", priority, ". Now _saved: ", [i.priority for i in _saved])

        if not _check_lt_n_saved(or_equal=True):
            item = _saved.pop(0)
            print("-- Removed ", item.priority, ". Now _saved", [i.priority for i in _saved])
#             if isinstance(self.save_handler, BaseSaveHandler):            
#                 self.save_handler.remove(item.filename)

#         checkpoint = self._setup_checkpoint()
#         self.save_handler(checkpoint, filename)
        print("-- Saved ", filename, ". Now _saved", [i.priority for i in _saved])
```

</details>

Output

<details>

```
- 0
-- Appended 0 . Now _saved:  [0]
-- Saved  check_0 . Now _saved [0]

- 1
-- Appended 1 . Now _saved:  [0, 1]
-- Saved  check_1 . Now _saved [0, 1]

- 2
-- Appended 2 . Now _saved:  [0, 1, 2]
-- Saved  check_2 . Now _saved [0, 1, 2]

- 1
Found the same filename

- 3
-- Appended 3 . Now _saved:  [0, 1, 2, 3]
-- Removed  0 . Now _saved [1, 2, 3]
-- Saved  check_3 . Now _saved [1, 2, 3]

- 5
-- Appended 5 . Now _saved:  [1, 2, 3, 5]
-- Removed  1 . Now _saved [2, 3, 5]
-- Saved  check_5 . Now _saved [2, 3, 5]

- 6
-- Appended 6 . Now _saved:  [2, 3, 5, 6]
-- Removed  2 . Now _saved [3, 5, 6]
-- Saved  check_6 . Now _saved [3, 5, 6]

- 7
-- Appended 7 . Now _saved:  [3, 5, 6, 7]
-- Removed  3 . Now _saved [5, 6, 7]
-- Saved  check_7 . Now _saved [5, 6, 7]
```

</details>

Make senses ?

erip: No, it doesn't make sense because 

> Now _saved:  [0, 1, 2, 3]

means you have `len(_saved) > n_saved` even if it's temporary. This is exactly what we have now.

vfdev-5: > Even if this 'checkpoint' isn't saved to disk, it still puts the Checkpoint object in an inconsistent state.

Inconsistent in sense of temporary `len(_saved) > n_saved` , right ?

vfdev-5: If it is the only remark, you can change the order of "append" and "remove":
```python
        # When we setup the checkpoint, Checkpoint's state dict should be up to date as we add new item
        # and remove optionally other items.
        if not _check_lt_n_saved(or_equal=False):
            item = _saved.pop(0)
            print("-- Removed ", item.priority, ". Now _saved", [i.priority for i in _saved])
#             if isinstance(self.save_handler, BaseSaveHandler):            
#                 self.save_handler.remove(item.filename)
    
        _saved.append(Item(priority, filename))        
        _saved.sort(key=lambda item: item[0])
        print("-- Appended", priority, ". Now _saved: ", [i.priority for i in _saved])
```
and it will give : 
```
- 3
-- Removed  0 . Now _saved [1, 2]
-- Appended 3 . Now _saved:  [1, 2, 3]
-- Saved  check_3 . Now _saved [1, 2, 3]

- 5
-- Removed  1 . Now _saved [2, 3]
-- Appended 5 . Now _saved:  [2, 3, 5]
-- Saved  check_5 . Now _saved [2, 3, 5]

- 6
-- Removed  2 . Now _saved [3, 5]
-- Appended 6 . Now _saved:  [3, 5, 6]
-- Saved  check_6 . Now _saved [3, 5, 6]

- 7
-- Removed  3 . Now _saved [5, 6]
-- Appended 7 . Now _saved:  [5, 6, 7]
-- Saved  check_7 . Now _saved [5, 6, 7]
```

PR #1127, opened 2020-06-13
Custom filename pattern for saving checkpoints

### LINKED ISSUE #1074: "=" in the checkpoint name may be incompatible with OSs
## ❓ Questions/Help/Support
Issue raised by Eric from MONAI team:
File names with operators in them may be valid names in Linux but that's not the case in other OSs, plus shells like bash really don't like them.

Joel-hanson:
Fixes #1074
 
This PR adds a feature to set custom filename by updating the class attribute `filename_pattern`.

Description:

There was an issue #1074 reported some days back where the "=" in the checkpoint name can cause certain issues in some operating systems.
 I have added a class attribute `filename_pattern` with the value `{filename_prefix}_{name}_{global_step}_{score_name}={score}.{ext}`, This file pattern is used when we save a checkpoint. If we update this attribute to a new pattern the checkpoint name will be updated to the new pattern which the user specifies.

Example:

Checkpoint: 
Screenshot 2020-06-13 at 7 33 38 PM

Output: 
Screenshot 2020-06-13 at 7 33 44 PM

Checklist:
* [x] New docstrings: description and/or example code are in RST format


vfdev-5:
@Joel-hanson thanks for the PR ! Let me check it and comment out things I do not understand...

ignite/handlers/checkpoint.py
```diff
@@ -200,6 +200,8 @@ def score_function(engine):
 
     """
 
+    filename_pattern = "{filename_prefix}_{name}_{global_step}_{score_name}={score}.{ext}"
+    _filename_pattern = "{filename_prefix}_{name}_{global_step}_{score_name}={score}.{ext}"
```
vfdev-5, line 204: `_filename_pattern` is not used, right ?

Joel-hanson: yes you are right will fix this.

Joel-hanson:
Looking into the test which are failed

Joel-hanson:
> Looking into the test which failed.

Fixed some of the tests which were related to the current implementation

@vfdev-5 @sdesrozis can you please help me fix the other test cases

vfdev-5:
@Joel-hanson our CI can fail on distributed tests, so no worries for failed tests...
 I restarted failed test on Travis

ignite/handlers/checkpoint.py
```diff
@@ -259,10 +262,14 @@ def _check_lt_n_saved(self, or_equal=False):
 
     def __call__(self, engine: Engine) -> None:
 
-        suffix = ""
+        filename_pattern_dict = {
```
vfdev-5, line 265: Can we initialize it empty, for example: `filename_pattern_dict = {}` ?

ignite/handlers/checkpoint.py
```diff
@@ -238,12 +240,13 @@ def __init__(
 
         self.to_save = to_save
         self._fname_prefix = filename_prefix + "_" if len(filename_prefix) > 0 else filename_prefix
+        self.filename_prefix = filename_prefix
```
vfdev-5, line 243: Normally, we should use internal attribute `_fname_prefix` as defined above, instead of introducing new attribute.
However, it is a good question whether we would like to open those init attributes like `filename_prefix`, `score_function`, `score_name`, `ext` etc... 

@sdesrozis thoughts ?

ignite/handlers/checkpoint.py
```diff
@@ -295,7 +298,17 @@ def __call__(self, engine: Engine) -> None:
                 for k in checkpoint:
                     name = k
                 checkpoint = checkpoint[name]
-            filename = "{}{}_{}{}".format(self._fname_prefix, name, suffix, self._ext)
+
+            filename_pattern_dict["name"] = name
+            filename_pattern_dict["filename_prefix"] = self.filename_prefix
+            filename_pattern_dict["ext"] = self._ext
+            if Checkpoint.filename_pattern != Checkpoint.default_filename_pattern:
```
vfdev-5, line 305: IMO, code L305-L311 looks a bit ugly. We can in all cases factorize `filename = Checkpoint.filename_pattern.format(**filename_pattern_dict)`. And this check is needed to apply removal of `_` if some of pattern keys were not used. 
Let's think if we can simplify this part somehow and also remove IMO redundant class attribute `default_filename_pattern` ...

vfdev-5:
@Joel-hanson thanks for working on it. I left several comments that may improve current implementation and open a small discussion about improvements of `Checkpoint` class itself.

ignite/handlers/checkpoint.py
Joel-hanson: You are right @vfdev-5 its a bit ugly. Let me try to change it a bit to remove the redundant class attribute

ignite/handlers/checkpoint.py
Joel-hanson:
```suggestion
        filename_pattern_dict = {
            "filename_prefix": self.filename_prefix,
            "ext": self._ext,
            "global_step": "",
            "score_name": "",
            "score": "",
        }
```
@vfdev-5 is this ok or should I remove the blank ones. I kept the blank key-value pairs to avoid the key error.

ignite/handlers/checkpoint.py
Joel-hanson:
```suggestion
            filename_pattern_dict["name"] = name

            filename = Checkpoint.filename_pattern.format(**filename_pattern_dict)

            if not self.filename_prefix:
                filename = filename.replace("_{}".format(name), name)

            if self.global_step_transform is None:
                filename = filename.replace("__", "_")

            if self._score_name is None:
                filename = filename.replace("_=", "_")

            if self._score_name is None and not filename_pattern_dict["score"]:
                filename = filename.replace("_.", ".")
```
Will this work?

ignite/handlers/checkpoint.py
```diff
@@ -278,15 +285,11 @@ def __call__(self, engine: Engine) -> None:
...
-            elif self._score_function is not None:
-                if len(suffix) > 0:
-                    suffix += "_"
-                suffix = "{}{}".format(suffix, priority_str)
-            elif len(suffix) == 0:
-                suffix = "{}".format(priority_str)
+                filename_pattern_dict["score_name"] = "{}".format(self._score_name)
+                filename_pattern_dict["score"] = "{}".format(priority_str)
+
+            elif self._score_function is not None or all(suffix == "" for suffix in filename_pattern_dict.values()):
```
sdesrozis, line 291: If `filename_pattern_dict` modified as suggested, the test `all(...)` seems wrong.

Joel-hanson: yes, you are right @sdesrozis. I can push the updated code if the above suggestions are ok.

sdesrozis: Feel free to push your modifications 😉 We will review the commit again.

Joel-hanson: Thanks, @sdesrozis. I have updated the PR with latest suggested changes.

ignite/handlers/checkpoint.py
sdesrozis: I agree, that could be interesting.

ignite/handlers/checkpoint.py
```diff
@@ -295,7 +299,21 @@ def __call__(self, engine: Engine) -> None:
                 for k in checkpoint:
                     name = k
                 checkpoint = checkpoint[name]
-            filename = "{}{}_{}{}".format(self._fname_prefix, name, suffix, self._ext)
+            filename_pattern_dict["name"] = name
+
+            filename = Checkpoint.filename_pattern.format(**filename_pattern_dict)
+
+            if not self.filename_prefix:
```
sdesrozis, line 306: I realized that in L306-316 post processing is done using the `filename_pattern`.

For instance, If a user defines a pattern `{filename_prefix}__{name}.{ext}` without `global_step_transform`, `__` will be replace and that is not wanted here.

Joel-hanson: yes @sdesrozis you are right, but that part was a bit ugly so changed the code to what @vfdev-5 has suggested.

I can change this back if you are ok. But when going back will introduce a new class attribute. 
https://github.com/pytorch/ignite/pull/1127/commits/1802709e86c8068c84c804b829806da03385abec#diff-18cd1a3c06de70b0e34b8fcf32940b30R203-R204

ignite/handlers/checkpoint.py
Joel-hanson: @sdesrozis So shall I also add others into the init attributes.

ignite/handlers/checkpoint.py
sdesrozis: I think that comment from @vfdev-5 was about introduction of public class members. We should have a public attribute (for user) and others should be private
```
filename_pattern  = "..." #public
_filename_pattern = filename_pattern # private
```
As you did in your code, we can compare.

ignite/handlers/checkpoint.py
vfdev-5: OK, let's make all user defined attributes public:
- `self._fname_prefix -> self.filename_prefix`
- `self._score_function -> self.score_function` 
- `self._score_name -> self.score_name`
- `self._n_saved -> self.n_saved`
- `self._ext -> self.ext`

vfdev-5:
@Joel-hanson the problem of using predefined pattern like `{filename_prefix}_{name}_{global_step}_{score_name}={score}.{ext}` forces additions and suppressions because actual default pattern is more like `{filename_prefix}{name}_{global_step}{score_name}{score}.{ext}` with special processing of fields.

Just tried to play around all of that, I think class attribute is not a good approach... For example, I would like to setup 2 savers: training checkpoint (to save trainer, model, optimizer etc) and best model. For the first one, I do not want to alter filename pattern and for the second I provide score name and score function but for the filename instead of `best_model_9_val_acc=0.77.pt` I would like to have `best_model_9.pt`...

Maybe, we can introduce it as an additional argument to `__init__` with default as `None`.

ignite/handlers/checkpoint.py
Joel-hanson: ok will that @vfdev-5 @sdesrozis

ignite/handlers/checkpoint.py
Joel-hanson: @sdesrozis I shall look into the implementation of introducing an additional argument to `__init__` as @vfdev-5 has mentioned it will cause issues when having  2 savers.

Joel-hanson:
Good catch @vfdev-5 

@vfdev-5 @sdesrozis 
Just to be clear for me I will note down what I have in mind for the new implementation introducing additional argument to `__init__`.

- I will remove the class attributes and will introduce a new additional argument called `filename_pattern`
- Will also make user-defined attributes as public
```
self._fname_prefix -> self.filename_prefix
self._score_function -> self.score_function
self._score_name -> self.score_name
self._n_saved -> self.n_saved
self._ext -> self.ext
```
- The string formatting is not used when there is no filename_pattern provided
- When a filename_pattern is provided, we will use the string formatting to replace its values.

vfdev-5:
@Joel-hanson yes I think your notes are correct. Two last points 
>  - The string formatting is not used when there is no filename_pattern provided
>  - When a filename_pattern is provided, we will use the string formatting to replace its values.

may be not clear how to implement. It could nice to have a single code for user pattern or our pattern. We can setup our default pattern `{filename_prefix}{name}_{global_step}{score_name}{score}.{ext}` at `__init__` if nothing provided. In `__call__` we can use something like 
```
    def _add_field(f, sep="_"):
        return "{}{}".format(f, sep) if len(f) > 0 else ""
```
for `"filename_prefix": add_field(kwargs.get("filename_prefix", ""))`, but certainly a specific processing of `{global_step}{score_name}{score}` is needed in order to have a good number of `_` etc.

Joel-hanson:
@vfdev-5 @sdesrozis I have updated the PR with the required changes.

There are 2 things to note
1. I have not set any default filename pattern but works the same way as before, What I have done is to take the pattern only when we have a custom filename pattern assigned as an argument. @vfdev-5 I can change this to what you have suggested in the above comment to have default pattern `{filename_prefix}{name}_{global_step}{score_name}{score}.{ext}` if the current implementation doesn't meet the expectations.
2. What should we do when a user gives a filename pattern which doesn't have the right keys.
Example `{filename}{scores}.{ext}` this pattern has the keys `filename` and `scores` which we are not string formatting, this can cause a key error.
Will we mention this in the docs?

sdesrozis:
@Joel-hanson keys should be described in docs and the list of handled keys must be fixed. 

Thank you again for your work!

sdesrozis:
I played with the existing rules of checkpoint naming.

First, documentation states the rules (slightly modified)
```python
"""
- `filename_prefix` is the argument passed to the constructor,
- `name` is the key in `to_save` if a single object is to store, otherwise `name` is "checkpoint".
- `suffix` is composed as following `{global_step}_{score_name}={score}`.
   
Above `global_step` defined by the output of `global_step_transform` and `score` defined 
by the output of `score_function`.

By default, none of `score_function`, `score_name`, `global_step_transform` is defined, 
then suffix is setup by attached engine's current iteration. The filename will be
`{filename_prefix}_{name}_{engine.state.iteration}.{ext}`.

If defined a `score_function`, but without `score_name`, then suffix is defined by provided score.
The filename will be `{filename_prefix}_{name}_{score}.{ext}`. If `global_step_transform` is 
provided, then the filename will be `{filename_prefix}_{name}_{global_step}_{score}.{ext}`.

If defined `score_function` and `score_name`, then the filename will
be `{filename_prefix}_{name}_{score_name}={score}.{ext}`. 
If `global_step_transform` is provided, then the filename will be 
`{filename_prefix}_{name}_{global_step}_{score_name}={score}.{ext}`

For example, `score_name="neg_val_loss"` and `score_function` that returns `-loss` 
(as objects with highest scores will be retained), then saved filename will be
`{filename_prefix}_{name}_neg_val_loss=-0.1234.pt`.
"""
```

I wrote a tiny test which could help
```python
from typing import Callable, Optional

def test_pattern(
    filename_prefix: str = "",
    score_function: Optional[Callable] = None,
    score_name: Optional[str] = None,
    global_step_transform: Callable = None,
    filename_pattern: Optional[str] = None,
):
    # ignite rules are applied
    if filename_pattern is None:

        # filename_prefix will be included in pattern at the end
        
        if score_function is None and score_name is None and global_step_transform is None:
            # policy 1 `{filename_prefix}_{name}_{engine.state.iteration}.{ext}`
            filename_pattern = "{name}_%s.{ext}" % 203  # fake engine.state.iteration
        elif score_function and score_name is None:
            if global_step_transform:
                # policy 2a `{filename_prefix}_{name}_{global_step}_{score}.{ext}`
                filename_pattern = "{name}_{global_step}_{score}.{ext}"
            else:
                # policy 2b `{filename_prefix}_{name}_{global_step}_{score}.{ext}`
                filename_pattern = "{name}_{score}.{ext}"
        elif score_function and score_name:
            if global_step_transform:
                # policy 3a `{filename_prefix}_{name}_{global_step}_{score_name}={score}.{ext}`
                filename_pattern = "{name}_{global_step}_{score_name}={score}.{ext}"
            else:
                # policy 3b `{filename_prefix}_{name}_{global_step}_{score}.{ext}`
                filename_pattern = "{name}_{score_name}={score}.{ext}"
        else:
            raise Exception("weird case score_name is not None but score_function is None")
        
        # prefix is added to pattern if needed
        if filename_prefix:
            filename_pattern = "{filename_prefix}_" + filename_pattern

    # filename_pattern contains user specific pattern or ignite pattern

    # keys should be computed
    # compute score : fake call
    score = score_function() if score_function else None
    # compute global_step : fake call
    global_step = global_step_transform() if global_step_transform else None

    # dict of keys
    kwargs = {
        "filename_prefix": filename_prefix,
        "ext": "pt",
        "name": "checkpoint",
        "score_name": score_name,
        "score": score,
        "global_step": global_step
    }

    # pattern is applied wrt keys
    return filename_pattern.format(**kwargs)


if __name__ == '__main__':
    print(test_pattern())
    print(test_pattern("best"))
    print(test_pattern(score_function=lambda: 0.9999))
    print(test_pattern(score_function=lambda: 0.9999, global_step_transform=lambda: 12))
    print(test_pattern(score_function=lambda: 0.9999, score_name="acc"))
    print(test_pattern(score_function=lambda: 0.9999, score_name="acc", global_step_transform=lambda: 12))
    print(test_pattern("best", score_function=lambda: 0.9999))
    print(test_pattern("best", score_function=lambda: 0.9999, global_step_transform=lambda: 12))
    print(test_pattern("best", score_function=lambda: 0.9999, score_name="acc"))
    print(test_pattern("best", score_function=lambda: 0.9999, score_name="acc", global_step_transform=lambda: 12))

    pattern = "SAVE:{name}-{score_name}--{score}.pth"
    print(test_pattern("best", score_function=lambda: 0.9999, score_name="acc", global_step_transform=lambda: 12, filename_pattern=pattern))
```

Results are
```python
checkpoint_203.pt
best_checkpoint_203.pt
checkpoint_0.9999.pt
checkpoint_12_0.9999.pt
checkpoint_acc=0.9999.pt
checkpoint_12_acc=0.9999.pt
best_checkpoint_0.9999.pt
best_checkpoint_12_0.9999.pt
best_checkpoint_acc=0.9999.pt
best_checkpoint_12_acc=0.9999.pt
SAVE:checkpoint-acc--0.9999.pth
```

Hope that helps and no bug inside :)

Joel-hanson:
@sdesrozis Thank you so much for the docs and the test. I shall add it to this PR.
I have some doubt

1. The docstring which you have shared didn't have the first part which is 

```
This class stores a single file as a dictionary of provided objects to save.
The filename has the following structure: `{filename_prefix}_{name}_{suffix}.{ext}` where
``` 
is this intentional if so I will remove this from the docstring.

---
2. I have read the docstring and found that this is the only part which is updated

**from**
```
If defined a `score_function`, but without `score_name`, then suffix is defined by provided score.
The filename will be `{filename_prefix}_{name}_{global_step}_{score}.pt`.
```
**to**
```
If defined a `score_function`, but without `score_name`, then the suffix is defined by the provided score.
The filename will be `{filename_prefix}_{name}_{score}.{ext}`. If `global_step_transform` is provided,
then the filename will be `{filename_prefix}_{name}_{global_step}_{score}.{ext}`.
```
correct, this is just to confirm.

---
3. The current code has the checkpoint extension as `self.ext = ".pt"`, so when someone gives a filename pattern like `custom_{name}_{global_step}_{score}.{ext}` he will get it like with 2 dots in front of extension.
`current_checkpoint_acc_0.9999..pt`. So I shall change the default extension value to pt, ie `self.ext = "pt"` this will solve this issue.

sdesrozis:
@Joel-hanson 

1. I extracted partially and I missed this part. The idea was to have rules in mind. Please, don't remove the missing part from docstring.
2. That is correct.
3. You are right. The dot should be removed. My idea was to have patterns written in the most understandable way. Maybe the current implementation fits perfectly the different cases but I was not sure... Feel free to use (and change) this snippet if it helps otherwise we could use it to check implementation.

Joel-hanson:
@sdesrozis @vfdev-5 I have updated the PR these latest changes
1. Updated the docstring as requested.
2. The dot in the extension is updated accordingly.
3. The tests for this feature is also added.
4. bug: I had missed the assignment of `name` in the filename_pattern, so I have fixed it.

The test function is a bit ugly and it's too long is there any suggestions or thoughts on this. Thanks, @sdesrozis I have used your test_pattern function for testing the default filenames of the checkpoints.

ignite/handlers/checkpoint.py
```diff
@@ -295,14 +312,21 @@ def __call__(self, engine: Engine) -> None:
                 for k in checkpoint:
                     name = k
                 checkpoint = checkpoint[name]
-            filename = "{}{}_{}{}".format(self._fname_prefix, name, suffix, self._ext)
+
+            filename_pattern_dict["name"] = name
+            filename_prefix = self.filename_prefix
+            if self.filename_pattern is None:
```
sdesrozis, line 318: Here we have 2 paths : in a case, the dict is useless, and for the other case, suffix is useless. Maybe we should refactor following the code I provided ? (See other comment). 

Btw, thank you again, we are close to a perfect score 😉

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -1023,3 +1023,191 @@ def _test_tpu_saves_to_cpu_nprocs(index, dirname):
...
+            filename_pattern=filename_pattern,
+        )
+
+        trainer = Engine(lambda e, b: None)
+        trainer.state = State(epoch=12, iteration=203, score=0.9999)
+
+        checkpointer(trainer)
+        return checkpointer.last_checkpoint
+
+    def _ignite_filename_pattern(
```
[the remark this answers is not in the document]
vfdev-5: Yes, this test helper function looks nice and I agree that we could refactor ignite's code using something like that for readibility.

vfdev-5:
Just to make you aware of this PR (https://github.com/pytorch/ignite/pull/1135/) and that here a filename is setup with a default pattern... Probably, that PR will be merged before this one. It would be nice to take into account it and modify `TrainsSaver` code too.

ignite/handlers/checkpoint.py
Joel-hanson: Will update accordingly @sdesrozis let's make this perfect 💪 .

Joel-hanson:
> Just to make you aware of this PR (#1135) and that here a filename is setup with a default pattern... Probably, that PR will be merged before this one. It would be nice to take into account it and modify `TrainsSaver` code too.

I Happy to help @vfdev-5. I will look into it that as well and make the required changes.

@sdesrozis @vfdev-5 
I would like to note down what I have understood and what I have in mind.
1. I have to update the part which makes the filename for checkpoints (current version) with the helper function, as it is much more readable.
2. Have to update the test to have the expected values instead of the helper function.
3. Have to look into the issue which @vfdev-5 had reported to update `TrainsSaver` (I have not read it completely will post an update after going through the whole issue thread.

sdesrozis:
Point 1. and 2. : perfect 👌🏻 

Point 3. : PR1135 will be merged first and use custom filename pattern. We must update wrt this pour 😉

Joel-hanson:
@sdesrozis @sdesrozis I have updated the code.

There are some doubts:
1. I have added the line `global_step = None` to have all the key values for formatting the filename_pattern is it ok to have None as its value when formatting. https://github.com/pytorch/ignite/pull/1127/files#diff-18cd1a3c06de70b0e34b8fcf32940b30R299 
2. @sdesrozis I have applied the following lines of code to the regular pattern feature as there is a situation where the global step function is present and no score name and no score function is present.
```
            elif global_step_transform:
                filename_pattern = "{name}_{global_step}.{ext}"
```
L287-L288 
3. I read some of the comments on PR # 1135. @vfdev-5 is the prerequisite that you listed is linked to the same implementation of a filename pattern that need to be implemented for 'TrainsSaver' is it? can you please elaborate on it.

ignite/handlers/checkpoint.py
```diff
@@ -253,19 +256,52 @@ def last_checkpoint(self) -> Optional[str]:
...
-        if self._n_saved is None:
+        if self.n_saved is None:
             return True
-        return len(self._saved) < self._n_saved + int(or_equal)
+        return len(self._saved) < self.n_saved + int(or_equal)
+
+    def _add_field(self, f, sep="_"):
+        return "{}{}".format(f, sep) if len(f) > 0 else f
+
+    def _get_filename_pattern(
```
sdesrozis, line 266: Here, arguments `Callable` are used as a boolean flag. Why do not introduce
```
def _get_filename_pattern(
    self,
    filename_prefix: str = "",
    has_score_fn: book = False,
    has_score: bool = False,
    ...
)
```
What do you think?

sdesrozis:
@Joel-hanson It looks good to me 😊 thank you!! Just one small comment to check.

I don’t have check PR # 1135 yet, I do it soon. @vfdev-5 might have some inputs.

ignite/handlers/checkpoint.py
Joel-hanson: @sdesrozis it seems like a good choice. Are you preferring to do the following when calling `_get_filename_pattern`?
```
filename_pattern = self._get_filename_pattern(
                filename_prefix=self.filename_prefix,
                has_score_fn=True if self.score_function is None else False,
                has_score_name=True if self.score_name is None else False,
                has_global_step_transform=True if self.global_step_transform is None else False,
                filename_pattern=self.filename_pattern,
            )
```

Joel-hanson: @sdesrozis was busy with some work. Is the above one ok shall I make this change?

Joel-hanson:
@vfdev-5 I can see that the PR is merged can you elaborate on what is needs to be done? what I understood is linked to the same implementation of a filename pattern that needs to be implemented for 'TrainsSaver' is it?

vfdev-5:
@Joel-hanson sorry for delay with this PR as we were occupied by current release.

ignite/handlers/checkpoint.py
vfdev-5: @Joel-hanson maybe it could make sense to make this function public and static... 
`_get_filename_pattern` -> `setup_filename_pattern`

ignite/handlers/checkpoint.py
```diff
@@ -253,19 +256,52 @@ def last_checkpoint(self) -> Optional[str]:
...
+        self,
+        filename_prefix: str = "",
+        score_function: Optional[Callable] = None,
+        score_name: Optional[str] = None,
+        global_step_transform: Callable = None,
+        filename_pattern: Optional[str] = None,
```
vfdev-5, line 272: This is strange as argument, probably, there is no purpose to pass it, right ?

ignite/handlers/checkpoint.py
vfdev-5: Even more simple: 
```
filename_pattern = Checkpoint.setup_filename_pattern(
                filename_prefix=self.filename_prefix,
                has_score_fn=self.score_function is None,
                has_score_name=self.score_name is None,
                has_global_step_transform=self.global_step_transform is None,
            )
```

vfdev-5:
> @vfdev-5 I can see that the PR is merged can you elaborate on what is needs to be done? what I understood is linked to the same implementation of a filename pattern that needs to be implemented for 'TrainsSaver' is it?

Actually, there is nothing to adapt in `TrainsSaver`, in the begining I though that filename is recomposed in the code, but actually it is not the case. We can keep it as is.

ignite/handlers/checkpoint.py
Joel-hanson: Will update accordingly @vfdev-5

ignite/handlers/checkpoint.py
Joel-hanson: You are right @vfdev-5 will update this as well.

ignite/handlers/checkpoint.py
Joel-hanson: @vfdev-5 will the naming be confusing when having `with_score=self.score_function is None` 
```
if with_score and with_score_name and with_global_step_transform:
    filename_pattern = "{name}_{score}.{ext}"
```
is it ok to have `with_score=self.score_function is not None` , this applies to others variables as well

vfdev-5: I think we can explain the meaning of `with_score` in docstring of the method

Joel-hanson:
@vfdev-5 @sdesrozis I have pushed the latest changes 🤞 .
- Have updated the function _get_filename_pattern to seetup_filename_pattern which is not public and static
- The arguments which were defined with typing were removed.
```
    @staticmethod
    def setup_filename_pattern(
        filename_prefix, with_score, with_score_name, with_global_step_transform, filename_pattern=None,
    ):
```
- Have updated the argument as with_score, with_score_name and with_global_step_transform to have the following values 
```
 with_score=self.score_function is None,
 with_score_name=self.score_name is None,
 with_global_step_transform=global_step is None,
```

ignite/handlers/checkpoint.py
```diff
@@ -219,6 +220,7 @@ def __init__(
         n_saved: Optional[int] = 1,
         global_step_transform: Callable = None,
         archived: bool = False,
+        filename_pattern: Optional[str] = None,
```
vfdev-5, line 223: We need to update docstring with `filename_pattern`

ignite/handlers/checkpoint.py
```diff
@@ -353,6 +362,33 @@ def _setup_checkpoint(self) -> dict:
             checkpoint[k] = obj.state_dict()
         return checkpoint
 
+    @staticmethod
+    def setup_filename_pattern(
+        filename_prefix, with_score, with_score_name, with_global_step_transform, filename_pattern=None,
+    ):
```
vfdev-5, line 368: We need to provide a docstring for this public method, remove `filename_pattern=None` from args and update the code accordingly. 
We also need annotate input args (`typing`)

ignite/handlers/checkpoint.py
```diff
@@ -353,6 +362,33 @@ def _setup_checkpoint(self) -> dict:
...
+                    filename_pattern = "{name}_{score}.{ext}"
+            elif not with_score and not with_score_name:
+                if not with_global_step_transform:
+                    filename_pattern = "{name}_{global_step}_{score_name}={score}.{ext}"
+                else:
+                    filename_pattern = "{name}_{score_name}={score}.{ext}"
+            elif not with_global_step_transform:
+                filename_pattern = "{name}_{global_step}.{ext}"
+            else:
+                raise Exception("weird case score_name is not None but score_function is None")
```
vfdev-5, line 385:
```suggestion
                raise ValueError("If score_name is provided, score_function can not be None")
```

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -1211,6 +1211,158 @@ def test_distrib_single_device_xla_nprocs(xmp_executor, dirname):
     xmp_executor(_test_tpu_saves_to_cpu_nprocs, args=(dirname,), nprocs=n)
 
 
+def test_checkpoint_filename_pattern():
```
vfdev-5, line 1214: It would be necessary to add tests for `setup_filename_pattern` as it is a public methods.

sdesrozis:
@Joel-hanson thank you! I don’t have any comment more than those from @vfdev-5 😊

Thank you again for your work !

ignite/handlers/checkpoint.py
Joel-hanson:
```
    @staticmethod
    def setup_filename_pattern(
        filename_prefix: str = "",
        with_score: bool = False,
        with_score_name: bool = False,
        with_global_step_transform: bool = False,
        filename_pattern: str = False,
    ) -> str:
```
Is this ok @vfdev-5

sdesrozis: The typehint for `filename_pattern` is wrong 
```
filename_pattern: Optional[str] = None
```

vfdev-5: @sdesrozis we do not need this arg at all. See my message above, please

ignite/handlers/checkpoint.py
Joel-hanson: **Args**
```
filename_pattern (str, optional): If ``filename_pattern`` is provided, this pattern will be used to render
checkpoint filenames. If the pattern is not defined, the default pattern would be used.
```
**Note**
```
If only ``global_step_transform`` is defined, then suffix is setup using its return value.
The filename will be `{filename_prefix}_{name}_{global_step}.{ext}`.

If defined ``score_function`` and ``score_name``, then the filename will
be `{filename_prefix}_{name}_{score_name}={score}.{ext}`. If ``global_step_transform`` is provided, then
the filename will be `{filename_prefix}_{name}_{global_step}_{score_name}={score}.{ext}`

If ``filename_pattern`` is not specified, the filename pattern must use the default pattern described above.
If ``filename_pattern`` is given, the filename template would be used to render the filenames.

For example, ``score_name="neg_val_loss"`` and ``score_function`` that returns `-loss` (as objects with highest
scores will be retained), then saved filename will be `{filename_prefix}_{name}_neg_val_loss=-0.1234.pt`. If a 
custom ``filename_pattern`` is given ``filename_pattern="SAVE:{filename_prefix}-{name}-{score_name}-{score}.{ext}`` 
then the saved filename will be `SAVE:{filename_prefix}-{name}-neg_val_loss-0.1234.{ext}`

```
Does this docstring look ok.

vfdev-5: @Joel-hanson yes, it looks good. Make sure to respect indentation when have multiline text for Args.

ignite/handlers/checkpoint.py
Joel-hanson: Sorry, @vfdev-5 I was a bit confused, 
> Checkpoint.setup_filename_pattern(
>                 filename_prefix=self.filename_prefix,
>                 with_score=self.score_function is None,
>                 with_score_name=self.score_name is None,
>                 with_global_step_transform=self.global_step_transform is None,
>             )

Are you suggesting that we don't require the filename_pattern to be passed but rather checked before calling `setup_filename_pattern`.
**def __call__**
```
if self.filename_pattern is None:
    filename_pattern = self.setup_filename_pattern(
        filename_prefix=self.filename_prefix,
        with_score=self.score_function is None,
        with_score_name=self.score_name is None,
        with_global_step_transform=global_step is None,
    )
else:
    filename_pattern = self.filename_pattern
```
**def setup_filename_pattern**
```
def setup_filename_pattern(
    filename_prefix: str = "",
    with_score: bool = False,
    with_score_name: bool = False,
    with_global_step_transform: bool = False,
) -> Union[str, None]:
```
is this what you are suggesting?

vfdev-5: Yes, something like that. Note that `setup_filename_pattern` returns `str` and not `Union`.
```python
def setup_filename_pattern(...) -> str
```

Joel-hanson:
@vfdev-5 I have updated the code accordingly which fixes the following.
1. The docstring for this new filename_pattern argument - https://github.com/pytorch/ignite/pull/1127#discussion_r449188777
2. The typing issue with the arguments and also the removal of filename_pattern from static function `setup_filename_pattern` https://github.com/pytorch/ignite/pull/1127#discussion_r446630655
3. The setup filename exception update https://github.com/pytorch/ignite/pull/1127#discussion_r446630877
4. The tests for the public function `setup_filename_pattern` https://github.com/pytorch/ignite/pull/1127#discussion_r446631082

ignite/handlers/checkpoint.py
Joel-hanson: you are right @vfdev-5 will update this as well.

ignite/handlers/checkpoint.py
```diff
@@ -353,6 +370,36 @@ def _setup_checkpoint(self) -> dict:
             checkpoint[k] = obj.state_dict()
         return checkpoint
 
+    @staticmethod
+    def setup_filename_pattern(
+        filename_prefix: str = "",
+        with_score: bool = False,
+        with_score_name: bool = False,
+        with_global_step_transform: bool = False,
+    ) -> str:
```
vfdev-5, line 379: @Joel-hanson a docstring for `setup_filename_pattern` is missing. Could you please add one. Thanks!

vfdev-5: Thinking more about the argument `filename_prefix` of `setup_filename_pattern`. Seems like it is a bit misleading as its value is not used. As we only check if it is defined or not I propose to replace it by `with_prefix`.

Joel-hanson: @vfdev-5 will this work?
```python
"""Helper method to get the default filename pattern for a checkpoint.
Args:
    with_prefix (bool): If True, the `filename_prefix` with an underscore (filename_prefix_) will be added to the filename pattern.
        By default the `filename_prefix` will not be appended to the filename pattern
    with_score (bool): If True, this indicates that the score function is not provided, then it will look for ``with_score_name``
        and ``with_global_step_transform`` to make the required filename pattern. If it is False then this indicates that the score
        function is provided.
    with_score_name (bool): If True, this indicates that the score name is not provided, then it will look for 
        ``with_global_step_transform`` to make the required filename pattern. If it is False then this indicates that the score
        name is provided.
    with_global_step_transform (bool): If True, this indicates that there is no global step, then it will make the
        required filename pattern accordingly. If it is False then this indicates that there is a global step.
Example:
    .. code-block:: python

        filename_pattern = Checkpoint.setup_filename_pattern("model", True, True, True)
        print(filename_pattern)
        > "{filename_prefix}_{name}_{score}.{ext}"
"""
```

vfdev-5: Yes, looks good. Nit: Make all default values to True and in the example `Checkpoint.setup_filename_pattern("model", True, True, True)` -> `Checkpoint.setup_filename_pattern()`

Joel-hanson: @vfdev-5 I have updated the requested changes.

ignite/handlers/checkpoint.py
```diff
@@ -353,6 +370,57 @@ def _setup_checkpoint(self) -> dict:
...
+        with_prefix: bool = True,
+        with_score: bool = True,
+        with_score_name: bool = True,
+        with_global_step_transform: bool = True,
+    ) -> str:
+        """Helper method to get the default filename pattern for a checkpoint.
+        Args:
+            with_prefix (bool): If True, the `filename_prefix` with an underscore (filename_prefix_) will be added to the filename pattern.
+                By default the `filename_prefix` will not be appended to the filename pattern
+            with_score (bool): If True, this indicates that the score function is not provided, then it will look for ``with_score_name``
```
vfdev-5, line 384: @Joel-hanson seems like I didn't read attentively the last time the doscrint. Now, the description IMO is a bit cumbersome. Argument `with_score`:
> If True, this indicates that the score function is not provided

it looks very misleading, IMO. More intuitive approach is if `with_score`, then template will have `{score}`. 
Am I missing something ?

ignite/handlers/checkpoint.py
```diff
@@ -353,6 +370,57 @@ def _setup_checkpoint(self) -> dict:
...
+
+                from ignite.handlers import Checkpoint
+
+                filename_pattern = Checkpoint.setup_filename_pattern()
+                print(filename_pattern)
+                > "{filename_prefix}_{name}_{score}.{ext}"
+        """
+        filename_pattern = None
+        if with_score and with_score_name and with_global_step_transform:
+            filename_pattern = "{name}_{score}.{ext}"
```
vfdev-5, line 403: is this contre-intuitive too or I do not understand something ?

"if with score and with score name and with global step transform"  -> I would assume that filepattern contains those field.

ignite/handlers/checkpoint.py
Joel-hanson: yes, @vfdev-5 its a bit confusing. I think I had mentioned this earlier https://github.com/pytorch/ignite/pull/1127#discussion_r446535788. Sorry for the confusion I can update this asap I should have communicated this properly.

ignite/handlers/checkpoint.py
Joel-hanson: @vfdev-5 we will have the pattern `{name}_{score}.{ext}` if the score function, score name, and global step transform function are not provided.

vfdev-5: The code is 
```
if with_score and with_score_name and with_global_step_transform:
            filename_pattern = "{name}_{score}.{ext}"
```
- `with_score` means filepattern will be with the value from score function 
- `with_score_name` means filepattern will be with score name
- `with_global_step_transform` means filepattern will be  with global step
so I expect to have `filename_pattern = "{name}_{global_step}_{score_name}={score}.{ext}"`

ignite/handlers/checkpoint.py
vfdev-5: Please, try to make it simple and intuitive as much as possible.

ignite/handlers/checkpoint.py
Joel-hanson: I have updated accordingly @vfdev-5.

vfdev-5:
@Joel-hanson I modified the code and docs from my side.

ignite/handlers/checkpoint.py
```diff
@@ -101,32 +104,52 @@ class Checkpoint(Serializable):
...
+    +----------------+------------+-----------------------+----------------------------------------------+
+    |      None      |   None     |         X             | ``{global_step}``                            |
+    +----------------+------------+-----------------------+----------------------------------------------+
+    |       X        |    X       |        None           | ``{score_name}={score}``                     |
+    +----------------+------------+-----------------------+----------------------------------------------+
+
+    Above `global_step` defined by the output of `global_step_transform` and `score` defined by the output
+    of `score_function`.
+
+    By default, none of ``score_function``, ``score_name``, ``global_step_transform`` is defined, then suffix is
```
sdesrozis, line 133: Double quoted ``score_function`` should be like single quoted `score_function`, shouldn’t it ? Same for others.

ignite/handlers/checkpoint.py
```diff
@@ -101,32 +104,52 @@ class Checkpoint(Serializable):
...
-        setup by attached engine's current iteration. The filename will be
-        `{filename_prefix}_{name}_{engine.state.iteration}.{ext}`.
-
-        If only ``global_step_transform`` is defined, then suffix is setup using its return value.
-        The filename will be `{filename_prefix}_{name}_{global_step}.{ext}`.
-
-        If defined a ``score_function``, but without ``score_name``, then suffix is defined by provided score.
-        The filename will be `{filename_prefix}_{name}_{global_step}_{score}.pt`.
+    Note:
+        If ``filename_pattern`` is given, it will be used to render the filenames. ``filename_pattern`` is a string
```
sdesrozis, line 141: Same here concerning quotes and below

sdesrozis:
I fixed double quotations. The PR is ready for merge when CI will be ok.

vfdev-5:
@sdesrozis actually, docs rendering with double quotes is better than with single quotes. I prefer to keep double quotes.

sdesrozis:
😱 I completely missed that! Very sorry, I revert!

sdesrozis:
@Joel-hanson thank you for this work !! A lot of iteration but the result is nice 👍🏻

Joel-hanson:
Thanks a lot, @vfdev-5 and @sdesrozis,

It was my biggest opportunity to contribute to a well-known library like ignite. There have been several iterations, and each iteration has brought me something different to experience. Thank you again for motivating me to get this PR as successful as possible. It would be fantastic if you could find any of the pieces where I should have improved that might have made fewer iterations and made it easy for you guys.

PR #1129, opened 2020-06-14
Auto pin_memory

### LINKED ISSUE #1128: auto_dataloader can auto set pin_memory if appropriate
## 🚀 Feature

Idea is to automatically set `pin_memory=True` if `idist.backend()` is `nccl` or default device is `cuda`.

In the code we can omit this param instead of 
https://github.com/pytorch/ignite/blob/6b88eb89cf40e3c4ebe05e0fe67f3c8013ecc5c2/examples/contrib/cifar10/main.py#L199

InCogNiTo124:
Fixes #1128 

Description:
Adds a new feature, if the `device` is `'cuda'`, set `pin_memory` to `True` automatically.

Check list:

### DIFF
--- a/examples/contrib/cifar10/main.py
+++ b/examples/contrib/cifar10/main.py
@@ -196,7 +196,6 @@ def get_dataflow(config):
         batch_size=config["batch_size"],
         num_workers=config["num_workers"],
         shuffle=True,
-        pin_memory="cuda" in idist.device().type,
         drop_last=True,
     )
 
--- a/examples/contrib/cifar10/main.py
+++ b/examples/contrib/cifar10/main.py
@@ -192,11 +192,7 @@ def get_dataflow(config):
 
     # Setup data loader also adapted to distributed config: nccl, gloo, xla-tpu
     train_loader = idist.auto_dataloader(
-        train_dataset,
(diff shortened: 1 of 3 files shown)


InCogNiTo124:
This ought to be the best commit I've ever pushed

vfdev-5:
@InCogNiTo124 it is almost perfect PR :)
Just to confirm if user set `pin_memory=False`, we rewrite it to True, right ?
I think it would be better to set it True only in case if user doesn't provide anything.

InCogNiTo124:
> Just to confirm if user set `pin_memory=False`, we rewrite it to True, right ?
I think that's already the case:

L101: `kwargs["pin_memory"] = kwargs.get("pin_memory", idist.device() == "cuda")`

What ever the user has provided for `pin_memory` (whether `True` or `False`) is passed directly with this. Only if the user hasn't provided anything, will the default parameter be returned, which is set to `True` if cuda is applicable, and `False` otherwise.

Please correct me if I am wrong though.

vfdev-5:
No, you are right, that's correct !

InCogNiTo124:
Not a problem :)

PR #1132, opened 2020-06-16
Accuracy MultiLabel Handling and Error Message

### LINKED ISSUE #1100: Misleading ValueError - Accuracy Metric Multilabel
## 🐛 Bug description

When using Accuracy.update() with both inputs having the second dimension 1, e.g. in my case `torch.Size([256,1])` the raised error message is misleading.

To reproduce 
```
from ignite.metrics import Accuracy
import torch
acc = Accuracy(is_multilabel=True)
acc.update((torch.zeros((256,1)), torch.zeros((256,1))))
```
`ValueError: y and y_pred must have same shape of (batch_size, num_categories, ...).`

In this case the y and y_pred do have the same shape but the issue is that it's not an accepted multilabel input (the `and y.shape[1] != 1` in the following code block from `_check_shape` in `_BaseClassification`). This should be indicated in the error message (or the if statement changed).

What is the argument to not allow a `y.shape[1]` of 1?

```
if self._is_multilabel and not (y.shape == y_pred.shape and y.ndimension() > 1 and y.shape[1] != 1):
            raise ValueError("y and y_pred must have same shape of (batch_size, num_categories, ...).")
```

## Environment

 - Ignite Version (e.g., 0.3.0): 0.3.0
 - OS (e.g., Linux): Linux
 - Python version:

anmolsjoshi:
Fixes #1100 

Description:
* Updated check for multilabel to handle cases where num_categories == 0, 1
* Updated error message with requirement num_categories > 1 for multilabel cases

Check list:


ignite/metrics/accuracy.py
```diff
@@ -98,7 +100,8 @@ class Accuracy(_BaseClassification):
     - `update` must receive output of the form `(y_pred, y)` or `{'y_pred': y_pred, 'y': y}`.
     - `y_pred` must be in the following shape (batch_size, num_categories, ...) or (batch_size, ...).
     - `y` must be in the following shape (batch_size, ...).
-    - `y` and `y_pred` must be in the following shape of (batch_size, num_categories, ...) for multilabel cases.
+    - `y` and `y_pred` must be in the following shape of (batch_size, num_categories, ...) and
+      num_categories must be greater than 1 for multilabel cases.
```
sdesrozis, line 104: I think it should produce wrong indentation. This piece of code should be merged with the previous sentence.

ignite/metrics/accuracy.py
```diff
@@ -43,8 +43,10 @@ def _check_shape(self, output: Sequence[torch.Tensor]) -> None:
         if not (y_shape == y_pred_shape):
             raise ValueError("y and y_pred must have compatible shapes.")
 
-        if self._is_multilabel and not (y.shape == y_pred.shape and y.ndimension() > 1 and y.shape[1] != 1):
-            raise ValueError("y and y_pred must have same shape of (batch_size, num_categories, ...).")
+        if self._is_multilabel and not (y.shape == y_pred.shape and y.ndimension() > 1 and y.shape[1] > 1):
+            raise ValueError(
+                "y and y_pred must have same shape of (batch_size, num_categories, ...) " "and num_categories > 1."
```
sdesrozis, line 48: Maybe merge two strings and let `black` do the correct format?

anmolsjoshi: Done!

ignite/metrics/accuracy.py
anmolsjoshi: I think this should work fine - please see attached screenshot.

image

sdesrozis: That is perfect ! Sorry for the extra work...

sdesrozis:
@vfdev-5 I'm not authorized to merge this pull request.

PR #1135, opened 2020-06-16
Fix TrainsSaver handling of Checkpoint's n_saved

### LINKED ISSUE #1056: TrainsSaver doesn't respect Checkpoint's n_saved

jkhenning:
Fixes #1056 

Description:
Fix Trains `TrainsSaver` to use Trains framework callbacks in concert with Ignite's revised `Checkpoint`'s delete/save behavior to make sure the `Checkpoint` `n_saved` parameter is respected not only with regard to checkpoints saved locally (as specified by `dirname`) but also for checkpoint entries (models) reported to the Trains platform (also includes checkpoints uploaded using Trains' remote storage support)

Check list:


jkhenning:
Hi @vfdev-5,

Please note that `requirements-dev.txt` was updated to require Trains 0.15.1rc0 that supports some new required changes. Once the PR is approved in principal, we'll release Trains 0.15.1 so this PR can be merged.

Thanks 😄

ignite/contrib/handlers/trains_logger.py
```diff
@@ -659,32 +661,84 @@ def _setup_check_trains(self, logger, output_uri):
...
-                if os.path.exists(path):
-                    trains.binding.frameworks.WeightsFileHandler.create_output_model(
-                        model=checkpoint,
-                        saved_path=path,
-                        framework=trains.model.Framework.pytorch,
-                        task=self._task,
-                        singlefile=True,
-                        model_name=os.path.basename(filename),
-                    )
+        try:
```
vfdev-5, line 709: Do we need to setup all this on all proceses of distributed configuration ?

bmartinn: I think that in a distributed scenario only a single instance (rank 0) will actually call the callbacks.

bmartinn:
@vfdev-5 let me add some missing background on the functionality of the callbacks.

The idea is to be able to affect the way Trains logs and uploads any stored model.
These callbacks will be called during from the Trains context, right after `torch.save` stored the mode.
This allows us to add metadata on the Model object created in Trains, change the uploaded file name (i.e. limit according to the n_saved argument), change the Model name in the Trains UI etc.

The `_CallbackContext` makes sure we allocate the correct "upload" file for the current save operation while respecting the `n_saved` limit. Basically it will find an empty "slot" (i.e. filename we can use overwrite) in the `pre_callback` setting the upload filename for Trains, then in the `post_callback` it will rename the Trains Model object with the original filename and add the metadata in the comment section (so it is later searchable)

>Can we test somehow (via mock at least) if we fix the issue ?

I think `examples/contrib/mnist/mnist_with_trains_logger.py` reproduces the issue, as `Checkpoint` is created with `n_saved=1`, or am I missing something ?

ignite/contrib/handlers/trains_logger.py
vfdev-5: OK, I see what you mean. Yes, technically in all distributed configurations, we call `torch.save` on a single process only.

vfdev-5:
> I think examples/contrib/mnist/mnist_with_trains_logger.py reproduces the issue, as Checkpoint is created with n_saved=1, or am I missing something ?

I wanted to say to unit test it in some way. For example, when remove is called, intended logic is executed ... Probably, I need to play a bit around it to better understand what happens.

ignite/contrib/handlers/trains_logger.py
```diff
@@ -659,32 +661,84 @@ def _setup_check_trains(self, logger, output_uri):
...
+        def __init__(
+            self, slots: List, checkpoint_key: str, filename: str, basename: str, metadata: Optional[Mapping] = None
+        ):
+            self._slots = slots
+            self._checkpoint_key = str(checkpoint_key)
+            self._filename = filename
+            self._basename = basename
+            self._metadata = metadata
+
+        def pre_callback(self, action: str, model_info: Any):
```
vfdev-5, line 674: @jkhenning what kind of object practically it is `model_info` ?
and as discussed with @bmartinn can we use an enum instead of `str` for checking `action == "save"` ?

ignite/contrib/handlers/trains_logger.py
```diff
@@ -659,32 +661,84 @@ def _setup_check_trains(self, logger, output_uri):
         if output_uri:
             self._task.output_uri = output_uri
 
+    class _CallbacksContext:
```
vfdev-5, line 664: @jkhenning as said previously (https://github.com/pytorch/ignite/pull/1135#issuecomment-645045035), we need to cover `_CallbacksContext.pre_callback` and `_CallbacksContext.post_callback` with basic unit tests to ensure the correct `model_info` as output.

ignite/contrib/handlers/trains_logger.py
jkhenning: @vfdev-5 `model_info`'s type is an internal type (`WeightsFileHandler.ModelInfo`).
As for the enum - we already added that to Trains after we released the RC - I'll update it soon.

ignite/contrib/handlers/trains_logger.py
jkhenning: @vfdev-5 Since the unit tests don't initialize a Trains Task, we can't really get a real callback from Trains and so can't test the `model_info` type returned by Trains. The instance returned to Trains is never instantiated in the Ignite code and we always return the same instance we received from Trains, so I don't see what can be tested there. Also, since that type is a Trains internal type, I'm not sure I understand what can be tested in the context of an Ignite unit test...
Be that as it may, I might be missing something :) - can you please explain what you meant?

vfdev-5: I see what you mean, however, what can be tested is (maybe a bit artificial) that 
`model_info.upload_filename` and `model_info.local_model_id` are correct as well as the content of `_slots`when returned from `pre_callback`. For example, first time model is saved, there is a `model_info` passed to `pre_callback` and we can check the fields of output `model_info`.
And something similar about other fields of `model_info` from `post_callback`.
What do you think ?

jkhenning: I see your point, however the thing is we currently don't have a way to get Trains to call these callbacks in a unit-test scenario, so we'll have to create a unit test that will:
* Mimic what `TrainsSaver.__call__()` usually does, i.e. create a `_CallbacksContext` instance
* Create a `model_info`-like object with the required attributes
* Pass it to `pre_callback()`
* Pass it to `post_callback()`

It's doable, but I have a feeling all we'll be testing is that the unit-test code is tailored exactly to the callbacks code to make sure they don't fail 😢 

However, I'll be happy to go with whatever you decide 😄

vfdev-5: 👍 

> It's doable, but I have a feeling all we'll be testing is that the unit-test code is tailored exactly to the callbacks code to make sure they don't fail 

Yes, it will be rather redundant test to ensure that returned structures are good, but it is necessary especially taking into account this PR #1127 where user would like to change filename pattern.

vfdev-5:
@jkhenning could you please add unit tests on `_CallbacksContext.pre_callback/post_callback` as we discussed above and update the version of trains as 0.15.1 is released :) Thanks 1

jkhenning:
@vfdev-5 I've added unit test, updated trains version and used the `CallbackType` enum instead of using plain strings - please let me know what you think.

vfdev-5:
@jkhenning looks good, thanks ! I'll try to add some more checks inside the test and I can merge then.

PR #1140, opened 2020-06-18
Added check_compute_fn argument to EpochMetric and related metrics

### LINKED ISSUE #1103: EpochMetric signature check caused error
## 🐛 Bug description
I tried to re-implement the `ROC_AUC` like the one in `contrib.metrics.roc_auc`, and passed some extra kwarg options to support for multiclass and multilabel classification. However, when I run my code, an Error "Number of classes in y_true not equal to the number of  columns in 'y_score' " was raised by `scikit-learn`. I checked the` y_true` of the whole epoch, and the number of unique values was equal to the number of columns of `y_score`. Next, I checked the source code of EpochMetric, and found that in `update` function, it tried to execute the `compute_fn` at the first iteration.  I supposed this was the cause: the` y_true` of a single batch might not contain all the classes of the dataset, therefore `scikit-learn` would throws the error mentioned above. The same problem occured for `AveragePrecision`.

## Environment

 - PyTorch Version (e.g., 1.5): 1.5
 - Ignite Version (e.g., 0.3.0): 0.3.0
 - OS (e.g., Linux): Ubuntu 18.04
 - Python version: 3.7.6
 - Any other relevant information: None

anmolsjoshi:
Fixes #1103  

Description:
- Added check_compute_fn argument to EpochMetric
- Updated related metrics in contrib section

Check list:


ignite/metrics/epoch_metric.py
```diff
@@ -34,16 +34,21 @@ class EpochMetric(Metric):
             :class:`~ignite.engine.Engine`'s `process_function`'s output into the
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
+        check_compute_fn (bool): if True, compute_fn is run on the first batch of data to ensure there are no
+            issues. If issues exist, user is warned that there might be an issue with the compute_fn.
```
vfdev-5, line 38:
```suggestion
            issues. If issues exist, user is warned that there might be an issue with the ``compute_fn``.
```

ignite/contrib/metrics/average_precision.py
```diff
@@ -22,6 +22,8 @@ class AveragePrecision(EpochMetric):
             :class:`~ignite.engine.Engine`'s `process_function`'s output into the
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
+        check_compute_fn (bool): if True, compute_fn is run on the first batch of data to ensure there are no
+            issues. If issues exist, user is warned that there might be an issue with the compute_fn.
```
vfdev-5, line 26: Maybe, it wont be clear about which `compute_fn` it says.

ignite/contrib/metrics/precision_recall_curve.py
```diff
@@ -23,6 +23,8 @@ class PrecisionRecallCurve(EpochMetric):
             :class:`~ignite.engine.Engine`'s `process_function`'s output into the
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
+        check_compute_fn (bool): if True, compute_fn is run on the first batch of data to ensure there are no
+            issues. If issues exist, user is warned that there might be an issue with the compute_fn.
```
vfdev-5, line 27: Same here: Maybe, it wont be clear about which `compute_fn` it says.

ignite/contrib/metrics/roc_auc.py
```diff
@@ -34,6 +34,8 @@ class ROC_AUC(EpochMetric):
             :class:`~ignite.engine.Engine`'s `process_function`'s output into the
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
+        check_compute_fn (bool): if True, compute_fn is run on the first batch of data to ensure there are no
+            issues. If issues exist, user is warned that there might be an issue with the compute_fn.
```
vfdev-5, line 38: Same here.

ignite/contrib/metrics/roc_auc.py
```diff
@@ -64,6 +68,8 @@ class RocCurve(EpochMetric):
...
```
vfdev-5, line 72: Same here.

ignite/contrib/metrics/average_precision.py
anmolsjoshi: Could you clarify Victor?

vfdev-5: Right, `AveragePrecision` has no `compute_fn` as argument and it executes `average_precision_score` internally. This argument `check_compute_fn` as described right now can be missleading.

I was also thinking that for those metrics like AP, ROC etc it would be more likely to see warnings as a single batch is not that representative... Another thought was to make it False by default or even always False...

anmolsjoshi: When implementing it - that's what I was thinking - although a sanity check is a good idea IMO. Keeping it True will keep the behavior consistent across versions.

anmolsjoshi: Do you mean keeping it False for just those metrics (AP, ROC, etc)?

vfdev-5: True, but somehow it causes this issue : #1103 and slowdown of computations (should be confirmed though) ... 
OK, let's keep it True and just improve docstring

vfdev-5: > Do you mean keeping it False for just those metrics (AP, ROC, etc)?

Yes, 
- False for only for AP, ROC etc 
- True for `EpochMetric`

anmolsjoshi: What are your thoughts on keeping it False and not giving the user an option to change it?

vfdev-5: > What are your thoughts on keeping it False and not giving the user an option to change it for these contrib metrics?

Let's keep it False and with option to change to True.

anmolsjoshi: Thoughts on the docstring below?

check_compute_fn (bool): Optional default False. If True, ``sklearn.metrics.average_precision_score`` is run on the first batch of data to ensure there are no issues. User will be warned in case there are any issues computing the function.

vfdev-5: Looks good ! Please, add `sklearn.metrics.average_precision_score` as link to sklearn docs.

anmolsjoshi: Done!

ignite/contrib/metrics/precision_recall_curve.py
anmolsjoshi: Done!

ignite/contrib/metrics/roc_auc.py
anmolsjoshi: Done!

ignite/contrib/metrics/roc_auc.py
anmolsjoshi: Done!

anmolsjoshi:
@vfdev-5 I added this functionality to _BaseRegressionEpoch as well.

ignite/contrib/metrics/average_precision.py
```diff
@@ -37,5 +41,7 @@ def activated_output_transform(output):
 
     """
 
-    def __init__(self, output_transform=lambda x: x):
-        super(AveragePrecision, self).__init__(average_precision_compute_fn, output_transform=output_transform)
+    def __init__(self, output_transform=lambda x: x, check_compute_fn=False):
```
sdesrozis, line 44: Typehint is missing on the new boolean parameter.

anmolsjoshi: Added!

PR #1145, opened 2020-06-20
Issue #1115 : pbar persists due to specific rule in tqdm (notebook) when n < total

### LINKED ISSUE #1115: ProgressBar does not close the bar on Colab if every is used
## 🐛 Bug description

Progress bar is set with `common.setup_common_training_handlers(... log_every_iters=15, ...)` and it does not close itself as being not terminated ...

Screen Shot 2020-06-09 at 16 34 53

## Environment

 - OS (e.g., Linux):
 - Python version:

sdesrozis:
pbar persists in notebook due to specific rules when n < total

Fixes #1115 

Description:

pbar persists due to specific rule in tqdm (notebook) when n < total.

Check list:


ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -199,6 +199,11 @@ def _reset(self, pbar_total):
 
     def _close(self, engine):
         if self.pbar is not None:
+            # https://github.com/tqdm/notebook.py#L240-L250
+            # issue #1115 : notebook backend of tqdm checks if n < total (error or KeyboardInterrupt)
+            # and the bar persists in 'danger' mode
+            if not self.pbar.leave:
```
vfdev-5, line 205: Why we check `if not self.pbar.leave` ?
In the code of `tqdm close` it remains in "danger" state if `self.n < self.total`:
```python
        if self.total and self.n < self.total:
            self.sp(bar_style='danger')
```
for any value of `self.pbar.leave`.

sdesrozis: Because if the pbar supposed to persist, we should not hack the values. I think.

vfdev-5: I think even if it should persist it would be better to keep it in "success" state instead of "danger" ?
If it is our call to close the pbar we can say the we update the state to the final step and then close it...

sdesrozis: If I understand the meaning, we must force the last step of the filtered event ? (I.e. every x batch iterations AND the last batch iteration).

vfdev-5: yes, in some sense it can be a solution. But I think if our closing event is triggered and we execute our code to close pbar we can simply advance tqdm's `n` to `total` without calling additional pbar updates on the last step.

sdesrozis: I didn’t catch the difference with what I did. Could you explain a bit more please ? Thanks 😊

vfdev-5: If you run the following
```python
pbar = ProgressBar(persist=True)

import time

e = Engine(lambda e, b: time.sleep(0.01))

pbar.attach(e, event_name=Events.ITERATION_COMPLETED(every=13))

e.run(range(40), max_epochs=3)
```
you will still have red bars in "danger" state instead of completed bars.

And if you replace the code as 
```python
    def _close(self, engine):
        if self.pbar is not None:
            # https://github.com/tqdm/notebook.py#L240-L250
            # issue #1115 : notebook backend of tqdm checks if n < total (error or KeyboardInterrupt)
            # and the bar persists in 'danger' mode
            if self.pbar.n < self.pbar.total:
                self.pbar.n = self.pbar.total
            self.pbar.close()
        self.pbar = None
```
Bars are closed = green and remain in completed state.

sdesrozis: Ok I see now, thank you 😊

vfdev-5:
But seems like it fails :)

vfdev-5:
@sdesrozis do you think we can test this somehow ? Otherwise, let's merge it like that.

sdesrozis:
I think it’s quite hard to test. Maybe we could add test on values of pbar.n and pbar.total but no way in notebook mode...

vfdev-5:
let's just merge it :)

PR #1146, opened 2020-06-20
Template codes

### LINKED ISSUE #1007: Add a template generic example script as a quick start with Ignite
## 🚀 Feature

Idea is to simplify user's learning curve and provide a template example that can be copied by the user and modified to his/her needs. Something like that but an executable working example:
```

def training():

    train_loader, val_loader = get_dataflow(fold_index=fold_index, **kwargs)
    model, optimizer, criterion, lr_scheduler = initialize_model(**kwargs)
    metrics = {
         "accuracy": Accuracy(),
    }

    trainer = create_supervised_trainer(model, optimizer, criterion)

    def update_lr_scheduler(_):
        lr_scheduler.step()

    evaluator = create_supervised_evaluator(model, metrics=metrics)

    @trainer.on(Events.ITERATION_COMPLETED(every=validate_every))
    def validate(trainer):
        evaluator.run(val_loader)
        print("After {} iterations, binary accuracy = {:.2f}"
              .format(trainer.state.iteration, metrics['accuracy']))

    trainer.run(train_loader, max_epochs=kwargs.get(max_epochs, 100))
    

```

vfdev-5:
Fixes #1007 

Description:
- Added templates folder
  - Vision template code

Check list:

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,7 +309,7 @@ Few pointers to get you started:
 
 - [Quick Start Guide: Essentials of getting a project up and running](https://pytorch.org/ignite/quickstart.html)
 - [Concepts of the library: Engine, Events & Handlers, State, Metrics](https://pytorch.org/ignite/concepts.html)
-- Full-featured template examples for common tasks (coming soon ...)
+- [Full-featured template examples for common tasks](examples/templates)
 
 
 <!-- ############################################################################################################### -->
--- a/examples/contrib/cifar10/main.py
+++ b/examples/contrib/cifar10/main.py
@@ -196,11 +196,7 @@ def get_dataflow(config):
     )
 
     test_loader = idist.auto_dataloader(
-        test_dataset,
-        batch_size=2 * config["batch_size"],
-        num_workers=config["num_workers"],
-        shuffle=False,
-        pin_memory="cuda" in idist.device().type,
+        test_dataset, batch_size=2 * config["batch_size"], num_workers=config["num_workers"], shuffle=False,
     )
     return train_loader, test_loader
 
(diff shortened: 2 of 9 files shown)


vfdev-5:
@erip @sdesrozis if you could review, please.

@amatsukawa may I also ask your opinion about this template code for vision task.


The entrypoint is here: 
- [examples/templates/README.md](https://github.com/pytorch/ignite/blob/0c254d48ffd992e2fd3372490261ec7acba0f83d/examples/templates/README.md)


Thanks !

amatsukawa:
I think it's generally great to have something like this, and it mostly looks good to me. A few minor:

- It seems to me like vision/NLP differences would mostly be hidden in the data loader, and maybe a different train_step if it's an RNN (the difference is really recurrent model vs not, not vision/nlp in that case).
- The code writes checkpoints, but does not try to load them AFAICT. Perhaps that should be included?

vfdev-5:
@amatsukawa @sdesrozis thanks for the review.


> The code writes checkpoints, but does not try to load them AFAICT. Perhaps that should be included?

Yes, that's a great idea to add resume part of code (forgot about that)

>   It seems to me like vision/NLP differences would mostly be hidden in the data loader, and maybe a different train_step if it's an RNN (the difference is really recurrent model vs not, not vision/nlp in that case).

Yes, that's true that difference of ignite's part of code between NLP vs Vision will be not that large. Let's see if it would better to add new example or generalize this example...

vfdev-5:
> It seems to me like vision/NLP differences would mostly be hidden in the data loader, and maybe a different train_step if it's an RNN (the difference is really recurrent model vs not, not vision/nlp in that case).

Thinking more about that. Actually, before implementing, the idea was to provide template codes structured by categories of training:
- supervised single model training (common NLP, Vision tasks)
- supervised multi-models training (GAN, Teacher-Student training etc)
- ...

What do you think would be more pertinent as template code ?

vfdev-5:
Closing this PR in favor of [code-generator app](https://code-generator.pytorch-ignite.ai/)

PR #1155, opened 2020-06-23
Use global_step as priority, if it exists

### LINKED ISSUE #1148: Getting iterations in Checkpoint is wrong for `global_step_transform`
There seems to be two bugs in `Checkpoint` related to `global_step_transform` when it's attached to the valid rather than train engine.

First, `global_step_transform` does a lookup based on the event fired. This causes issues when the handler is not attached to an `{EPOCH/ITERATION}_COMPLETED`, eg. when it's attached to `COMPLETED` on the valid engine as the docs suggest. 

Second, `global_step_transform` is intended to give the "true" count (iteration, epoch, whatever it may be). As such, it should not only be used in the filename, but also as the `priority`. Right now, priority is the iteration count of the engine it's attached to, which again does not work for valid engine.

A third point, which isn't really a bug but more usability: `Checkpoint` silently drops checkpoints if it has checkpointed the same filename before. I think such occurrences are likely user error (or in my case, framework error, since my iteration count of valid engine is always the same at `COMPLETED`). Perhaps a warning log is warranted. Alternatively, if the checkpoint is truly the same, writing it again is idempotent, so perhaps this check should be removed entirely.

amatsukawa:
Fixes #1148

Description:

Use `global_step` as the `priority` inside of `Checkpoint`, when it is available. Fall back to the iteration count. This synchronizes the filename of the checkpoint and its priority, when `global_step_transform` is provided without `score_*`.

Check list:
* [] New doc strings: description and/or example code are in RST format

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -111,7 +111,7 @@ class Checkpoint:
         setup by attached engine's current iteration. The filename will be
         `{filename_prefix}_{name}_{engine.state.iteration}.{ext}`.
 
-        If only ``global_step_transform`` is defined, then suffix is setup using its return value. 
+        If only ``global_step_transform`` is defined, then suffix is setup using its return value.
         The filename will be `{filename_prefix}_{name}_{global_step}.{ext}`.
 
         If defined a ``score_function``, but without ``score_name``, then suffix is defined by provided score.
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -111,6 +111,9 @@ class Checkpoint:
         setup by attached engine's current iteration. The filename will be
         `{filename_prefix}_{name}_{engine.state.iteration}.{ext}`.
 
+        If only ``global_step_transform`` is defined, then suffix is setup using its return value. 
+        The filename will be `{filename_prefix}_{name}_{global_step}.{ext}`.
+
         If defined a ``score_function``, but without ``score_name``, then suffix is defined by provided score.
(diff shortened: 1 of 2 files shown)


ignite/handlers/checkpoint.py
```diff
@@ -269,7 +272,10 @@ def __call__(self, engine: Engine) -> None:
             if not isinstance(priority, numbers.Number):
                 raise ValueError("Output of score_function should be a number")
         else:
-            priority = engine.state.get_event_attrib_value(Events.ITERATION_COMPLETED)
+            if self.global_step_transform is not None:
+                priority = global_step
```
vfdev-5, line 276: It looks generally good, but I have a minor correction to make here. It is not obvious that 
`global_step` is defined here.
It would be better to code it like that :
```python
        global_step = None
        if self.global_step_transform is not None:
            global_step = self.global_step_transform(engine, engine.last_event_name)
            suffix = "{}".format(global_step)
       # ....
        else:
            if global_step is not None:
                priority = global_step
            else:
                priority = engine.state.get_event_attrib_value(Events.ITERATION_COMPLETED)
```

What do you think ?

amatsukawa: Yes, I think that's more clear. Changed.

amatsukawa:
@vfdev-5 there are some CI failures, but they don't look related to this change. Can you confirm?

vfdev-5:
@amatsukawa yes our ci is currently a bit flaky. Yes, those failures are not related to the PR.

vfdev-5:
Thanks again for the PR @amatsukawa !

PR #1156, opened 2020-06-23
Stateful handlers

### LINKED ISSUE #966: Checkpoint stateful handlers and metrics

amatsukawa:
Related to #966

Description:

This introduces `state_dict` and `load_state_dict` into stateful handlers. It also adds a flag `include_self`, to include the checkpointer's state into the checkpoint.

With these changes, stateful handlers can be checkpointed as:

```python
train = ...
valid = ...

early_stop = EarlyStopping(...))
valid. add_event_handler(Events.COMPLETED, early_stop)

best_checkpoint = Checkpoint(..., score_function=..., score_name=...)
valid.add_event_handler(Events.COMPLETED, best_checkpoint)

# It's important that this comes last.
to_save = {"train": train, "early_stop": early_stop, "best_checkpoint": best_checkpoint}
checkpoint = Checkpoint(
        to_save, 
        ..., 
        include_self=True)
valid.add_event_handler(Events.COMPLETED, checkpoint)
```

They can be loaded by normally by `checkpoint.load_objects`, using the `to_save` dict above in addition to a `to_save["checkpointer"] = checkpoint` key.

Check list:
I have not updated the documentation, and indeed this is probably deserving of some because the order of setting it up matters, but I'm hoping to get this into 0.4.0 and add documentation later on when I have some more time :)


vfdev-5:
@amatsukawa thanks for this PR too !

I would change "Fixes #966" with related to #966 I think it wont completely fix it, right ?

I agree with updating `Checkpoint` and `EarlyStopping` to be derived from `Serializable`. However, automatic inclusion of `self` with `include_itself` should be carefully tested and I think it can be too short to have this in v0.4.0...
What do you think ?

amatsukawa:
I changed the description. This does support the most of the functionality I had in mind, except for the "do it automatically" part (which as we discussed elsewhere, may not be realistic after all).

I did add a test to make sure `include_self` at least seemed to work as intended. I followed a fairly simple rule that seems to me like would have a high likelihood of correctness, which is to add `self.state_dict` *right before* the call to the save handler (which is the last thing in the method). 

I'm happy to remove `include_self` from the PR for now. Its effect would be that `Checkpoint` object cannot respect the `n_saved` parameter *across resumes*.

vfdev-5:
@amatsukawa I need to play a bit with the code. Maybe, a complete integration test would be nice to ensure that saving and resume (`Checkpoint.load_objects`) work as intended...
Let's take a bit of time to check all this (during 24h) and than we take a decision either keep or remove as we plan to cut the release on June 25th. Does it sound good for you ?

amatsukawa:
Is there an existing integration test for things that this touches, without checkpointing? If so, I can try to just add to it or create a copy of it for checkpointing.

vfdev-5:
Maybe this one is the most integrating among the others : https://github.com/pytorch/ignite/blob/master/tests/ignite/handlers/test_checkpoint.py#L713

amatsukawa:
@vfdev-5 I made a version of that test you linked with new features, PTAL.

amatsukawa:
Oops, @vfdev-5 I'm sorry I didn't realized you had pushed some changes to this branch. I thought I had done something wrong in merging master to cause the conflict, so I force pushed.

Could you push your changes again?

vfdev-5:
Yep, sorry, I have this practice (push into existing PRs), it can be surprising :)
I'll try to merge your newer commits and mine...

vfdev-5:
I think if we are done with it, we can merge it like that. No need to update the docs as API is not that brilliant.
Ideally, it would be great to have a possibilty to add somehow those stateful components like `early_stop` and `best_checkpoint` to the trainer such that when we save the trainer, we save also appended components...
Something like that 
```python
trainer = Engine(lambda e, b: None)

to_save = {"trainer": trainer}
checkpoint = ModelCheckpoint(...)

evaluator = Engine(lambda e, b: None)
early_stopping = EarlyStopping(...)
evaluator.add_event_handler(Events.COMPLETED, early_stopping)

trainer.state.early_stopping = early_stopping
trainer.state_dict_user_keys.append("early_stopping")  # this wont work but an API like that maybe more interesting vs manual addition.
```

amatsukawa:
Thanks for the review!

I would vote for merging this in to support this functionality manually for 0.4, and improving usability from there. 

As future work, perhaps hook states can appear in engine states, and both engines can be checkpointed. Alternatively, #912 happens, it may be fine to hide it inside a Trainer how it is now.

vfdev-5:
Sounds good ! Let's go like that :)
Thanks again for the PR, Aki !

PR #1167, opened 2020-06-29
Livelossplot logger

### LINKED ISSUE #1069: Livelossplot logger
## 🚀 Feature - Livelossplot logger

The logger that is similar to Neptune, W&B, tensorboard loggers available in Ignite.
livelossplot:
[https://github.com/stared/livelossplot](url)
ignite plugin that is available in the livelossplot:[https://github.com/stared/livelossplot/blob/master/livelossplot/inputs/pytorch_ignite.py](url)
notebook example:
[https://github.com/stared/livelossplot/blob/master/examples/pytorch-ignite.ipynb](url)
The logger can be extended to log other things in addition to metrics (for ex. learning rate).

Bartolo1024:
Follow up discussion in #1069 

Fixes #1069 

Description:
Add livelossplot logger to ignite.contrib.handlers

Check list:
* [x] Logger
* [x] Example
* [ ] Tests


Bartolo1024:
What is the best way to call plot_losses.send() method only once per epoch? I had to use flag to prevent calling it from each logger but it is not the clearest way :(

vfdev-5:
Thanks for the PR @Bartolo1024 !

> What is the best way to call plot_losses.send() method only once per epoch? I had to use flag to prevent calling it from each logger but it is not the clearest way :(

I'd say `send()` can be called with `writer.update`, but probably I need to playaround it a bit to completely understand the problem you mention here.

vfdev-5:
@Bartolo1024 I updated a bit the code. I understand better now your question. I need to think about it how to implement in a clean way...
There are some blocks due to iteration-wise and epoch-wise logging, x label still displays "epochs" for iterations which is misleading. Opened an issue : https://github.com/stared/livelossplot/issues/115
Let's postpone this PR for sometime, until x label resolution. 
What do you think ?

Bartolo1024:
Ok, we can postpone it until we will add the functionality to livelossplot.

Bartolo1024:
Now xlabel can be changed in the livelossplot. Do you know how to implement calling `plot_losses.send()` once per epoch in a clean way?

vfdev-5:
@Bartolo1024 great ! 

> Do you know how to implement calling plot_losses.send() once per epoch in a clean way?

I was thinking about that and now I need to review again the code etc to remind a solution I had in mind. Let me get back to this PR in coming days, maybe the next week and we'll try to land it. Does it sound good for you ?

Bartolo1024:
@vfdev-5 Yes, sounds good!

vfdev-5:
@Bartolo1024 I'm trying to understand how to automatically setup those `step_names` and `group_pattern` such that user does not have to specify them at least for default configuration (examples/notebooks/LivelossplotLogger_MNIST.ipynb)

```python
livelossplot_logger = LivelossplotLogger()  
# here we create PlotLosses

livelossplot_logger.attach_output_handler(
    trainer,
    event_name=Events.ITERATION_COMPLETED(every=10),
    tag="training",
    output_transform=lambda loss: {"batch loss": loss},
)
# only here we know that we would like to log with `iteration` axis for "training batch loss"

for tag, evaluator in [("training", train_evaluator), ("validation", validation_evaluator)]:
    livelossplot_logger.attach_output_handler(
        evaluator,
        event_name=Events.EPOCH_COMPLETED,
        tag=tag,
        metric_names=["loss", "accuracy"],
    )

livelossplot_logger.attach_opt_params_handler(
    trainer, event_name=Events.ITERATION_COMPLETED(every=10), optimizer=optimizer
)
# same here, only here we know that we would like to log with `iteration` axis for "training batch loss"
```

Do you think we specify them after `PlotLosses` writer creation?

Bartolo1024:
Maybe it will be the best to set ignite-specific defaults in the longer constructor like that?:
```
class LivelossplotLogger(BaseLogger):
    def __init__(
            self,
            step_names: Optional[Union[str, Dict[str, str]]] = None,
            group_patterns: Optional[Dict[str, str]] = None,
            *args,
            **kwargs
    ):
        try:
            import livelossplot
        except ImportError:
            raise RuntimeError(
                "This contrib module requires livelossplot to be installed. "
                "You may install trains using: \n pip install livelossplot \n"
            )

        from livelossplot import PlotLosses
        
        if not group_patterns:
            group_patterns = [
                (r"^(training(\s))(.*)", "training"),
                (r"^(validation(\s))(.*)", "validation"),
                (r"^(batch(\s))(.*)", "batch"),
            ]

        if not step_names:
            step_names = {
                "training": "epoch",
                "validation": "epoch",
                "batch": "iteration",
            }

        self.writer = PlotLosses(step_names=step_names, group_patterns=group_patterns, *args, **kwargs)
```
Livelossplot has its own default setup, but now it has only epoch mode. Putting the default iteration mode to livelossplot defaults probably will decrease simplicity too much :(

vfdev-5:
@Bartolo1024 Yes, I understand that, but IMO `step_names` and `group_patterns` are not much user-friendly without knowing how `PlotLoss` works. I would like to have some intermediate interface to preconfigure internally those things while based on few info that user could provide...

PR #1171, opened 2020-07-01
Issue 1123 - Improve usage of contrib common methods with other save handlers 

### LINKED ISSUE #1123: Improve usage of contrib common methods with other save handlers
## 🚀 Feature

Currently, methods like `common.save_best_model_by_val_score` and `common.setup_common_training_handlers` work with local storage setup by `output_path`.

If we would like to use other save handlers than `DiskSaver`, it is almost impossible to use those helpers.

Idea is to be able user's setup of save handler for internal `Checkpoint` class. 
There can be several ways, I see, how to accomplish that:
- make `output_path` accept instances of `BaseSaveHandler`

```python
from ignite.contrib.engines import common

save_handler = TrainsSaver(dirname=output_path)

common.save_best_model_by_val_score(
        save_handler, evaluator, model=model, metric_name="accuracy", n_saved=3, trainer=trainer, tag="test"
)
```

- set module-wise default save handler as DiskSaver and allow user to replace it by another save hander:

```python
from ignite.contrib.engines import common

common.default_save_handler = TrainsSaver(dirname=output_path)

common.save_best_model_by_val_score(
        None, evaluator, model=model, metric_name="accuracy", n_saved=3, trainer=trainer, tag="test"
)
```

- add another argument `save_handler=None`

```python
save_handler = TrainsSaver(dirname=output_path)

common.save_best_model_by_val_score(
        None, evaluator, model=model, metric_name="accuracy", n_saved=3, trainer=trainer, tag="test", save_handler=save_handler
)
```

Any other ideas @sdesrozis @erip ?

vfdev-5:
Fixes #1123  

Description:
- added save_handler arg to setup_common_training_handlers
- added method delegated_save_best_models_by_val_score

Check list:

Personally, I'm not really fan of the name `delegated_save_best_models_by_val_score`, if you have other better ideas ...

### DIFF
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -42,7 +42,7 @@ def setup_common_training_handlers(
     stop_on_nan=True,
     clear_cuda_cache=True,
     save_handler=None,
-    **kwargs,
+    **kwargs
 ):
     """Helper method to setup trainer with common handlers (it also supports distributed configuration):
         - :class:`~ignite.handlers.TerminateOnNan`
@@ -132,7 +132,7 @@ def _setup_common_training_handlers(
     stop_on_nan=True,
     clear_cuda_cache=True,
     save_handler=None,
-    **kwargs,
+    **kwargs
 ):
     if output_path is not None and save_handler is not None:
         raise ValueError(
@@ -214,7 +214,7 @@ def _setup_common_distrib_training_handlers(
     stop_on_nan=True,
     clear_cuda_cache=True,
     save_handler=None,
-    **kwargs,
+    **kwargs
 ):
 
     _setup_common_training_handlers(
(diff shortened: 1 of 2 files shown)


vfdev-5:
@sdesrozis @erip any other better name for `delegated_save_best_models_by_val_score` ?

sdesrozis:
Possible to do in the same way than `setup_common_training_handlers` using mutual exclusion ?

No better idea atm

vfdev-5:
> Possible to do in the same way than `setup_common_training_handlers` using mutual exclusion ?
> 
> No better idea atm

No, due to `output_path` as first arg and not the last arg.

vfdev-5:
@sdesrozis how about:

- `any_save_best_models_by_val_score`  ?
- `save_handler_best_models_by_val_score` ?
- `store_save_best_models_by_val_score`  (as we also have `save_best_model_by_val_score` both maybe cumbersome)
- `gen_save_best_models_by_val_score` ?

instead of `delegated_save_best_models_by_val_score`

sdesrozis:
`save_handler_best_models_by_val_score` sounds good to me because it’s more explicit.

But I’m not a good audience for that, I never use such helpers. It encapsulates few code, I prefer write it.

vfdev-5:
@sdesrozis i renamed the method to `gen_save_best_models_by_val_score` which looks a bit better. 

Generally, with `save_best_models_by_val_score` you can gain 5-10 lines of code when doing explicitly the same thing...
Do you think we should make provide explicit code for saving best models in reproducible examples ?

sdesrozis:
It’s purely a matter of taste...

PR #1175, opened 2020-07-03
ValueError -> TypeError

WrRan:
Fixes # 

Description:
>`TypeError` may be raised when an operation or function is applied to an object of an inappropriate type, while `ValueError` may be raised when an operation or function receives an argument that has the right type but an inappropriate value.
>
> @ python at https://docs.python.org/3/library/exceptions.html

Thus, `TypeError` is preferred over `ValueError` here.

Check list:

### DIFF
--- a/ignite/engine/events.py
+++ b/ignite/engine/events.py
@@ -259,7 +259,7 @@ class EventsList:
 
     def _append(self, event: Union[Events, CallableEventWithFilter]):
         if not isinstance(event, (Events, CallableEventWithFilter)):
-            raise ValueError("Argument event should be Events or CallableEventWithFilter, got: {}".format(type(event)))
+            raise TypeError("Argument event should be Events or CallableEventWithFilter, got: {}".format(type(event)))
         self._events.append(event)
 
     def __getitem__(self, item):
--- a/tests/ignite/engine/test_event_handlers.py
+++ b/tests/ignite/engine/test_event_handlers.py
@@ -313,7 +313,7 @@ def test_events_list_removable_handle():
 
 def test_eventslist__append_raises():
     ev_list = EventsList()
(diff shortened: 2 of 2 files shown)


sdesrozis:
@WrRan Thank you ! Good catch !

Could you look a bit deeper to check if there are others misusing ? It would help a lot 😊 

Tests are ko because we catch the exception type and test it. I think that tests should be updated.

Thanks again 🙏🏻

vfdev-5:
@WrRan the following test is failing as it checks for `ValueError` instead of `TypeError`: https://travis-ci.org/github/pytorch/ignite/jobs/704562160#L2316
Could you please adapt it accordingly. Thanks!

WrRan:
Got it. I am working on it. :-)

WrRan:
@sean-adler 
I have gone through the codebase and found there is a wide spectrum of misuses of exceptions in lots of files (such as `NotImplemented` and `NotImplementedError` see #1178  and `RuntimeError` and `TypeError` see https://github.com/pytorch/ignite/blob/master/ignite/contrib/handlers/polyaxon_logger.py#L104 etc.)

To make matter worse, there are some situations we should divide one exception into more than one. For example, in https://github.com/pytorch/ignite/blob/master/ignite/contrib/handlers/param_scheduler.py#L860, the `ValueError` should be one `TypeError` and one `Value`.

This situation makes not only refactor code hard but also code review not easy.

So, I think it is a better idea to work on this problem one by one file through many PRs rather than all of them at one time. What do you think?

vfdev-5:
@WrRan thanks for pointing out those things. Fixing those incoherent with python best-practices exception is not a priority, but if you would like to make our codebase perfect, feel free to send PRs with fixed code and adapted test. Thanks !

PR #1178, opened 2020-07-04
[BC-breaking] NotImplementedError -> NotImplemented

WrRan:
Fixes # 

Description:
> **NotImplmented**:
> Special value which should be returned by the binary special methods to indicate that the operation is not implemented with respect to the other type.
> @ python from https://docs.python.org/3/library/constants.html#NotImplemented

while
> **exception.NotImplementedError**:
> In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.
> @ python from https://docs.python.org/3/library/exceptions.html#NotImplementedError

Note that they are not interchangeable.

Simply say, `NotImplemented` give the argument `other` a chance to support the binary operation (e.g. `__eq__(self, event: Event)` in this PR), while `NotImplementedError` not.
Thus, the former is preferred here. 

Check list:


vfdev-5:
Thanks for the PR! I read the docs and seems like the correct exception should be `NotImplmented`, but flake8 is not agree with that... Do you know why ?

WrRan:
I have made one mistake: `NotImplmented` is a value instead of an exception.
See https://docs.python.org/3/library/exceptions.html#exception-hierarchy and https://docs.python.org/3/library/exceptions.html#exception-hierarchy.

So I should return it rather than raise it.

vfdev-5:
Yes, seems like that. We need to return it instead of raising. I'm checking the implication of that and it changes current behaviour, namely, we restricted all custom events to be `EventEnum`, with this change it is no more required.

The following code is 
```python
class MyEvents(Enum):
    TEST_EVENT = "test_event"

engine.register_events(*MyEvents)
engine.add_event_handler(MyEvents.TEST_EVENT, lambda engine: None)
engine.fire_event(MyEvents.TEST_EVENT)
```

@sdesrozis do you think we still have to restrict event's type ? It means that with that change we can still register everything:
```
    engine = Engine(lambda engine, batch: 0)

    class A:
        pass

    for ev in [1, "a", A()]:
        engine.register_events(ev)
        engine.add_event_handler(ev, lambda engine: None)
        engine.fire_event(ev)
```

To restrict the type of possible events, it should be done now in `register_events`.

sdesrozis:
@vfdev-5 @WrRan 

It could be nice to register everything but events are a critical component of ignite. We should think about scenarios first because I’m afraid about collateral damage...

vfdev-5:
OK, let's make type restriction inside `register_events` by throwing a `TypeError`  if type is not `EventEnum` or `str`. @WrRan  could you please update the PR in this way ?

WrRan:
@vfdev-5 Please have a look.

ignite/engine/engine.py
```diff
@@ -191,7 +193,9 @@ class TBPTT_Events(EventEnum):
         if not (event_to_attr is None or isinstance(event_to_attr, dict)):
             raise ValueError("Expected event_to_attr to be dictionary. Got {}.".format(type(event_to_attr)))
 
-        for e in event_names:
+        for index, e in enumerate(event_names):
+            if not isinstance(e, (str, EventEnum)):
+                raise ValueError(f"Value at {index} of event_names should be a str or EventEnum, but given {e}")
```
vfdev-5, line 198: Shouldn't we raise TypeError instead of ValueError here ?

tests/ignite/engine/test_custom_events.py
```diff
@@ -48,6 +48,17 @@ def process_func(engine, batch):
     assert handle.called
 
 
+def test_custom_events_asserts():
```
vfdev-5, line 51: Could you please add the following tests too
```python
    engine = Engine(lambda engine, batch: 0)

    class A:
        pass

    for ev in [1, "a", A()]:
        with pytest.raises(ValueError, match=r"Value at \d of event_names should be a str or EventEnum"):
            engine.register_events(ev)
```

vfdev-5 (approved):

PR #1182, opened 2020-07-07
Fix Exception misuse in `ignite.contrib.engines.common.py`

WrRan:
Fixes # 

Description:
>`TypeError` may be raised when an operation or function is applied to an object of an inappropriate type, while `ValueError` may be raised when an operation or function receives an argument that has the right type but an inappropriate value.
>
> @ python at https://docs.python.org/3/library/exceptions.html

Thus, `TypeError` is preferred over `ValueError` here.

Check list:

### DIFF
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -9,7 +9,7 @@ jobs:
       with:
         python-version: '3.7'
     - run: |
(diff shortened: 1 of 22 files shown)


WrRan:
I do not know what's going wrong... I have already checked the test file `tests/ignite/contrib/engines/test_common.py`.

PR #1185, opened 2020-07-07
Refactored test_utils.py into 3 files

vfdev-5:
Description:
- Refactored test_utils.py into 3 files

Idea is that we can easily test new coming comp models (e.g. Horovod)

If everything is OK, we should have the same coverage etc...

Check list:


vfdev-5:
> This breaks the symmetry between test srcs and module srcs. Do you think we should refactor utils.py ?

@sdesrozis no, it is OK like that. We still keep a reminder of utils module in tests and split its tests into 3 files.

PR #1190, opened 2020-07-09
Update README.md

jspisak:
We are adding a disclaimer to all non-FB led repos in the PyTorch github org. Let me know if you have any concerns. Thanks!


jspisak:
updated per our offline chat. Thanks!

PR #1196, opened 2020-07-11
Minor optimization for idist.get_*

vfdev-5:
Description:
- Changes of dist model's synchronization to reduce the overhead of idist.get_*
- Test overhead on CircleCI with NCCL/Gloo backends
-  Removed `_sync_model_wrapper` to implicitly check if we need to sync model. This also reduces time of `idist.get_*` method calls vs native calls.

Check list:

### DIFF
--- a/ignite/distributed/utils.py
+++ b/ignite/distributed/utils.py
@@ -49,16 +49,20 @@ _need_to_sync = True
 def _sync_model_wrapper(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
-        if isinstance(_model, _SerialModel) and _need_to_sync:
-            sync()
+        if _need_to_sync and isinstance(_model, _SerialModel):
+            sync(temporary=True)
         return func(*args, **kwargs)
 
     return wrapper
 
 
-def sync():
+def sync(temporary=False):
     """Helper method to force this module to synchronize with current distributed context.
     This method should be used when distributed context is manually created or destroyed.
+
+    Args:
+        temporary (bool): If True, distributed model synchronization is done every call of ``idist.get_*`` methods.
+            This may have performance negative impact.
     """
     global _model
 
@@ -67,7 +71,7 @@ def sync():
             continue
         model = comp_model_cls.create_from_context()
         if model is not None:
-            _model = model
+            _set_model(model, temporary=temporary)
             return
 
     _model = _SerialModel()
@@ -356,11 +360,11 @@ def set_local_rank(index: int):
     ComputationModel._ext_local_rank = index
 
 
-def _set_model(model):
+def _set_model(model, temporary=False):
     global _model, _need_to_sync
     _model = model
     _need_to_sync = True
-    if not isinstance(_model, _SerialModel):
+    if not isinstance(_model, _SerialModel) and not temporary:
         _need_to_sync = False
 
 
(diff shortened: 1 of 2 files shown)


ignite/distributed/utils.py
```diff
@@ -46,19 +45,13 @@
 _need_to_sync = True
 
 
-def _sync_model_wrapper(func):
```
sdesrozis, line 49: Here, I don't understand why the decoration leads to the overhead.

ignite/distributed/utils.py
```diff
@@ -46,19 +45,13 @@
...
-    def wrapper(*args, **kwargs):
-        if isinstance(_model, _SerialModel) and _need_to_sync:
-            sync()
-        return func(*args, **kwargs)
-
-    return wrapper
-
-
-def sync():
+def sync(temporary=False):
```
sdesrozis, line 48: IMO `temporary` is obfuscating. Why something more explicit and related to the docstring, for instance `sync_every_call` ?

tests/ignite/distributed/utils/test_native.py
```diff
@@ -202,51 +202,63 @@ def test_idist_barrier_gloo(distributed_context_single_node_gloo):
...
+
+
 @pytest.mark.distributed
 @pytest.mark.skipif(not has_native_dist_support, reason="Skip if no native dist support")
+@pytest.mark.skipif(
+    not torch.cuda.is_available(), reason="Do not want to run this test on Github or Travis, but CircleCI"
+)
 def test_idist_methods_overhead_gloo(distributed_context_single_node_gloo):
-    import time
+    _test_idist_methods_overhead(2.5)
```
sdesrozis, line 238: Remark : overhead should depend on hardware. Maybe it could fail with some machines of users.

ignite/distributed/utils.py
vfdev-5: It is a small overhead (~1e-6 on my pc) but of the same order as the execution of wrapped function (e.g. `dist.get_world_size()`).

ignite/distributed/utils.py
vfdev-5: Maybe, public API should not expose `temporary`. We can have 2 functions: `sync` with implicit `temporary=False` and `_sync_once` with implicit `temporary=True` and used inside `idist.get_*`

tests/ignite/distributed/utils/test_native.py
vfdev-5: Well, it can depend on hardware. But we measure here execution time of our interfaces called before calling "worker" method:
```
idist.get_world_size() = need_sync() -> _model.get_world_size() -> dist.get_world_size()
```
vs 
```
dist.get_world_size()
```

Maybe we can mark this test as can fail or something ...

ignite/distributed/utils.py
sdesrozis: But more complicated. Maybe have a unique function with secure mode by default should be enough.

vfdev-5: > Maybe have a unique function with secure mode by default should be enough.

What do you mean by secure mode  ?

tests/ignite/distributed/utils/test_native.py
sdesrozis: It's a remark. Let's see if it happens.

ignite/distributed/utils.py
sdesrozis: Secure mode means it always works, so always sync.

vfdev-5: @sdesrozis so are we good with current API or you propose to change it or its behaviour ?

PR #1197, opened 2020-07-11
Docs update and auto_model change

### LINKED ISSUE #1174: Distributed model initialisation - CIFAR 10 example
## ❓ Questions/Help/Support

In the cifar10 example when a [model](https://github.com/pytorch/ignite/blob/master/examples/contrib/cifar10/main.py#L51) is defined inside the process, is the model updated and shared between the `nproc_per_node`?

In most of the TPU examples I have seen they define the model before the `Parallel` process.
- [tpu-training-super-fast-xlmroberta](https://www.kaggle.com/shonenkov/tpu-training-super-fast-xlmroberta)
- [i-like-clean-tpu-training-kernels-i-can-not-lie](https://www.kaggle.com/abhishek/i-like-clean-tpu-training-kernels-i-can-not-lie)

Does the way ignite handles the model ensure the same model is used throughout the training?

vfdev-5:
Fixes #1174 

Description:
- Updated docs
- Now `auto_model` puts params on device if they are not the device

Check list:

### DIFF
--- a/ignite/distributed/auto.py
+++ b/ignite/distributed/auto.py
@@ -139,6 +139,15 @@ def auto_model(model: nn.Module) -> nn.Module:
 
         model = idist.auto_model(model)
 
+    In addition with NVidia/Apex, it can be used in the following way:
+
+    .. code-block:: python
+
+        import ignite.distribted as idist
+
+        model, optimizer = amp.initialize(model, optimizer, opt_level="O1")
+        model = idist.auto_optim(model)
+
     Args:
         model (torch.nn.Module): model to adapt.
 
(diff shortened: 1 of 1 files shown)


ignite/distributed/auto.py
```diff
@@ -139,6 +139,15 @@ def auto_model(model: nn.Module) -> nn.Module:
...
         model = idist.auto_model(model)
 
+    In addition with NVidia/Apex, it can be used in the following way:
+
+    .. code-block:: python
+
+        import ignite.distribted as idist
+
+        model, optimizer = amp.initialize(model, optimizer, opt_level="O1")
+        model = idist.auto_optim(model)
```
sdesrozis, line 149: I think it should be ˋauto_model(model)` and. It ˋauto_optim`

PR #1211, opened 2020-07-21
REFACTOR: Early Return Pattern (if elif else -> if if return)

marload:
Fixes # 

Description:
I applied Early Return Pattern to the existing condition syntax.

Thank You!

Check list:

### DIFF
--- a/ignite/utils.py
+++ b/ignite/utils.py
@@ -38,16 +38,15 @@ def apply_to_type(
     """
     if isinstance(input_, input_type):
         return func(input_)
-    elif isinstance(input_, (str, bytes)):
+    if isinstance(input_, (str, bytes)):
         return input_
-    elif isinstance(input_, collections.Mapping):
+    if isinstance(input_, collections.Mapping):
         return type(input_)({k: apply_to_type(sample, input_type, func) for k, sample in input_.items()})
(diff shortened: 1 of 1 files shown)


sdesrozis:
@marload thank you !

This pattern is very subjective. I'm personally not fan of that. 

Here, you introduce this pattern in a small part of ignite. If we would promote this, we should use it as default to be consistent. 

Really I don't know.

vfdev-5:
@marload I agree with @sdesrozis and we should use this pattern everywhere and not in a small part of the code. Please, send another PR with fixes for all other cases where we have 
```python
if condition:
    return value
elif another_condition:
    return another_value
```

sdesrozis:
Example found [here](https://www.itamarweiss.com/personal/2018/02/28/return-early-pattern.html)

This piece of code 
```
def function foo():
  if <condition a>:
    if <condition a_1>:
      <code_block_a_1>
    else:
      <code_block_a_2>
  else:
    <code_block_b>
  return <result>
```
should be refactored like this
```
def function foo():
  if not <condition a>:
    <code_block_b>
    return <result>
  
  if <condition a_1>:
      <code_block_a_1>
      return <result>
    
  <code_block_a_2>
  return <result>
```

We can discuss about readability.

By the way, I think it's a hard work 😓

PR #1212, opened 2020-07-21
REFACTOR: Use Join method

marload:
Fixes # 

Description:
Concating strings using `+=` uses a lot of memory.

Check list:

### DIFF
--- a/ignite/engine/events.py
+++ b/ignite/engine/events.py
@@ -342,12 +342,12 @@ class State:
         return getattr(self, State.event_to_attr[event_name])
 
     def __repr__(self) -> str:
-        s = "State:\n"
+        s = ["State:"]
         for attr, value in self.__dict__.items():
             if not isinstance(value, (numbers.Number, str)):
                 value = type(value)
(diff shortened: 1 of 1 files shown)


sdesrozis:
@marload thank you for this.

There are several ways to concatenate strings in python and I'm not familiar about what you pointed out. Did you do that because you measured something ? I'm curious about that !

If we state that += should be replaced, we must check in whole code.

My point is that += is more simple and readable than join. But it's matter of taste 😊

vfdev-5:
@sdesrozis I think the point is about string concat performance impact which can be replaced by join as growing lists probably costs less...

vfdev-5:
@marload I agree with @sdesrozis that this is more a matter of code readibility and we prefer to keep current implementation.

I close this PR. Thanks again for your contribution. Please, see also "help_wanted" issues on which you can help us to move the project forward : https://github.com/pytorch/ignite/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22.

PR #1217, opened 2020-07-24
metrics: add SSIM

### LINKED ISSUE #1213: SSIM Metric
## 🚀 Feature

I would like to request to have SSIM metric in ignite. Since it has a few steps to implement on your own when you want to use in your project/paper implementation, I thought it's good to have OOTB in ignite and supported by the community.

I would like to implement this and if I am allowed, under which module/sub-module should my implementation go in?

Here's [my current implementation](https://gist.github.com/ydcjeff/ae94ef310cdf269353400aba7e65aae1).

ydcjeff:
Fixes #1213 

Description: added SSIM metric and tests

Check list:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
-.vscode
+
 # Distribution / packaging
 .Python
 build/
@@ -38,3 +38,7 @@ coverage.xml
 # Documentation
 /docs/src/
 /docs/build/
+
+# IDE hidden files
+.vscode
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
-
+.vscode
 # Distribution / packaging
 .Python
 build/
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,3 @@ coverage.xml
 # Documentation
 /docs/src/
 /docs/build/
-
-# IDE hidden files
-.vscode
-.idea
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -16,7 +16,7 @@ class SSIM(Metric):
     Args:
         data_range (int or float): Range of the image. Typically, ``1.0`` or ``255``.
         kernel_size (int or list or tuple of int): Size of the kernel. Default: (11, 11)
-        sigma (float or list or tuple of float): Standard deviation of the gaussian kernel. 
+        sigma (float or list or tuple of float): Standard deviation of the gaussian kernel.
             Argument is used if ``gaussian=True``. Default: (1.5, 1.5)
         k1 (float): Parameter of SSIM. Default: 0.01
         k2 (float): Parameter of SSIM. Default: 0.03
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -1,4 +1,4 @@
-from typing import Callable, Sequence
+from typing import Callable, Sequence, Union
 
 import torch
 import torch.nn.functional as F
@@ -15,22 +15,22 @@ class SSIM(Metric):
 
     Args:
         kernel_size (list or tuple of int): Size of the gaussian kernel. Default: (11, 11)
-        sigma: Standard deviation of the gaussian kernel. Default: (1.5, 1.5)
-        data_range: Range of the image. If ``None``, it is determined from the image (max - min)
-        k1: Parameter of SSIM. Default: 0.01
-        k2: Parameter of SSIM. Default: 0.03
+        sigma (list or tuple of float): Standard deviation of the gaussian kernel. Default: (1.5, 1.5)
+        data_range (int or float): Range of the image. Typically, ``1.0`` or ``255``.
+        k1 (float): Parameter of SSIM. Default: 0.01
+        k2 (float): Parameter of SSIM. Default: 0.03
         output_transform: A callable that is used to transform the
             :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
             form expected by the metric.
-
-    Returns:
-        A Tensor with SSIM
 
     Example:
 
     To use with ``Engine`` and ``process_function``, simply attach the metric instance to the engine.
     The output of the engine's ``process_function`` needs to be in the format of
     ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y, ...}``.
+
+    ``y_pred`` and ``y`` can be an un-normalized image tensor or a normalized image tensor.
+    Depending on that, the user might need to adjust ``data_range``.
 
     .. code-block:: python
 
@@ -65,34 +65,39 @@ class SSIM(Metric):
     .. code-block:: python
 
         >>> y_pred = torch.rand([16, 1, 16, 16])
-        >>> y = y_pred * 1.25
-        >>> ssim = SSIM()
+        >>> y = y_pred * 0.85
+        >>> ssim = SSIM(data_range=1.0)
         >>> ssim.update((y_pred, y))
         >>> ssim.compute()
-        tensor(0.9520)
+        tensor(0.9741)
     """
 
     def __init__(
         self,
-        kernel_size: Sequence[int] = (11, 11),
-        sigma: Sequence[float] = (1.5, 1.5),
-        data_range: float = None,
+        data_range: Union[int, float],
+        kernel_size: Union[int, Sequence[int]] = (11, 11),
+        sigma: Union[float, Sequence[float]] = (1.5, 1.5),
         k1: float = 0.01,
         k2: float = 0.03,
         output_transform: Callable = lambda x: x,
     ):
-        if len(kernel_size) != 2 or len(sigma) != 2:
-            raise ValueError(
-                "Expected `kernel_size` and `sigma` to have the length of two. Got kernel_size: {}, sigma: {}.".format(
-                    len(kernel_size), len(sigma)
-                )
-            )
+        if isinstance(kernel_size, int):
+            if kernel_size % 2 == 0 or kernel_size <= 0:
+                raise ValueError("Expected `kernel_size` to have odd positive number. Got {}.".format(kernel_size))
+        elif isinstance(kernel_size, Sequence):
+            if any(x % 2 == 0 or x <= 0 for x in kernel_size):
+                raise ValueError("Expected `kernel_size` to have odd positive number. Got {}.".format(kernel_size))
+        else:
+            raise ValueError("Invalid `kernel_size`.")
 
-        if any(x % 2 == 0 or x <= 0 for x in kernel_size):
-            raise ValueError("Expected `kernel_size` to have odd positive number. Got {}.".format(kernel_size))
-
-        if any(y <= 0 for y in sigma):
-            raise ValueError("Expected `sigma` to have positive number. Got {}.".format(sigma))
+        if isinstance(sigma, float):
+            if sigma <= 0:
+                raise ValueError("Expected `sigma` to have positive number. Got {}.".format(sigma))
+        elif isinstance(sigma, Sequence):
+            if any(y <= 0 for y in sigma):
+                raise ValueError("Expected `sigma` to have positive number. Got {}.".format(sigma))
+        else:
+            raise ValueError("Invalid `sigma`.")
 
         self.kernel_size = kernel_size
         self.sigma = sigma
@@ -114,8 +119,21 @@ class SSIM(Metric):
         return (gauss / gauss.sum()).unsqueeze(dim=0)  # (1, kernel_size)
 
     def _gaussian_kernel(self, channel, kernel_size, sigma, device):
-        gaussian_kernel_x = self._gaussian(kernel_size[0], sigma[0], device)
-        gaussian_kernel_y = self._gaussian(kernel_size[1], sigma[1], device)
+        if isinstance(kernel_size, int) and isinstance(sigma, float):
+            gaussian_kernel_x = self._gaussian(kernel_size, sigma, device)
+            gaussian_kernel_y = self._gaussian(kernel_size, sigma, device)
+        elif isinstance(kernel_size, Sequence) and isinstance(sigma, float):
+            gaussian_kernel_x = self._gaussian(kernel_size[0], sigma, device)
+            gaussian_kernel_y = self._gaussian(kernel_size[1], sigma, device)
+        elif isinstance(kernel_size, int) and isinstance(sigma, Sequence):
+            gaussian_kernel_x = self._gaussian(kernel_size, sigma[0], device)
+            gaussian_kernel_y = self._gaussian(kernel_size, sigma[1], device)
+        elif isinstance(kernel_size, Sequence) and isinstance(sigma, Sequence):
+            gaussian_kernel_x = self._gaussian(kernel_size[0], sigma[0], device)
+            gaussian_kernel_y = self._gaussian(kernel_size[1], sigma[1], device)
+        else:
+            raise TypeError("Invalid input type for `kernel_size` and `sigma`.")
+
         kernel = torch.matmul(gaussian_kernel_x.t(), gaussian_kernel_y)  # (kernel_size, 1) * (1, kernel_size)
 
         return kernel.expand(channel, 1, kernel_size[0], kernel_size[1])
@@ -143,9 +161,6 @@ class SSIM(Metric):
                     y_pred.shape, y.shape
                 )
             )
-
-        if self.data_range is None:
-            self.data_range = max(y_pred.max() - y_pred.min(), y.max() - y.min())
 
         C1 = pow(self.k1 * self.data_range, 2)
         C2 = pow(self.k2 * self.data_range, 2)
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -14,11 +14,12 @@ class SSIM(Metric):
     Computes Structual Similarity Index Measure
 
     Args:
+        data_range (int or float): Range of the image. Typically, ``1.0`` or ``255``.
         kernel_size (int or list or tuple of int): Size of the gaussian kernel. Default: (11, 11)
         sigma (float or list or tuple of float): Standard deviation of the gaussian kernel. Default: (1.5, 1.5)
-        data_range (int or float): Range of the image. Typically, ``1.0`` or ``255``.
         k1 (float): Parameter of SSIM. Default: 0.01
         k2 (float): Parameter of SSIM. Default: 0.03
+        gaussian (bool): ``True`` to use gaussian kernel, ``False`` to use uniform kernel
         output_transform (callable, optional): A callable that is used to transform the
             :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
             form expected by the metric.
@@ -49,6 +50,7 @@ class SSIM(Metric):
         sigma: Union[float, Sequence[float]] = (1.5, 1.5),
         k1: float = 0.01,
         k2: float = 0.03,
+        gaussian: bool = True,
         output_transform: Callable = lambda x: x,
     ):
         if isinstance(kernel_size, int):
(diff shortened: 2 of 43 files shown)


ydcjeff:
@vfdev-5 I have opened PR here. I might need some guidance for distributed tests as I have less knowledge about it.

.gitignore
```diff
@@ -3,7 +3,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
-
+.vscode
```
sdesrozis, line 6: I think that IDE hidden files should be tagged in specific part
```
# IDE hidden files
.vscode
.idea
etc.
```

ydcjeff: Okay, i will fix it.

sdesrozis: Thank you :)

You can look the link below for distributed configuration of tests 
tests https://github.com/pytorch/ignite/blob/6faa6ac1e3a46c79e0dfcfd976439b86329717b0/tests/ignite/metrics/test_accuracy.py#L781

Basically, you have to create full size tensors on every process 
https://github.com/pytorch/ignite/blob/6faa6ac1e3a46c79e0dfcfd976439b86329717b0/tests/ignite/metrics/test_accuracy.py#L749
Then you work with local bunch of tensors wrt to the rank
https://github.com/pytorch/ignite/blob/6faa6ac1e3a46c79e0dfcfd976439b86329717b0/tests/ignite/metrics/test_accuracy.py#L753
Finally you can compare to sequential results 
https://github.com/pytorch/ignite/blob/6faa6ac1e3a46c79e0dfcfd976439b86329717b0/tests/ignite/metrics/test_accuracy.py#L772

By the way, I think you should add first an integration example (meaning with `Engine`, `process_fn` and `data`).

Thank you again and feel free to ask if you need help.

ydcjeff: @sdesrozis Thank you, I will try it.

vfdev-5: @ydcjeff @sdesrozis unfortunately, we have to remove those entries in `.gitignore`. Let's keep the same way it was decided from the beginning to use locally `.git/info/exclude` to exclude files like those. Please, keep `.gitignore` unchanged.


@ydcjeff Btw, thanks for the PR !

ydcjeff: @vfdev-5 Okay, I will remove it then

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,171 @@
...
+
+__all__ = ["SSIM"]
+
+
+class SSIM(Metric):
+    """
+    Computes Structual Similarity Index Measure
+
+    Args:
+        kernel_size: Size of the gaussian kernel. Default: (11, 11)
```
vfdev-5, line 17: @ydcjeff please, add types for those args like that
```suggestion
        kernel_size (list or tuple of int): Size of the gaussian kernel. Default: (11, 11)
```

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,171 @@
...
+        sigma: Sequence[float] = (1.5, 1.5),
+        data_range: float = None,
+        k1: float = 0.01,
+        k2: float = 0.03,
+        output_transform: Callable = lambda x: x,
+    ):
+        if len(kernel_size) != 2 or len(sigma) != 2:
+            raise ValueError(
+                "Expected `kernel_size` and `sigma` to have the length of two."
+                f" Got kernel_size: {len(kernel_size)} and sigma: {len(sigma)}."
```
vfdev-5, line 87: Please, replace f-strings by `"".format` as we support python 3.5 too.

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,171 @@
...
+        self.k2 = k2
+        super(SSIM, self).__init__(output_transform=output_transform)
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._sum_of_batchwise_ssim = 0.0
+        self._num_examples = 0
+
+    def _gaussian_kernel(self, channel, kernel_size, sigma, device):
+        def gaussian(kernel_size, sigma, device):
```
vfdev-5, line 109: Maybe, it would be better to avoid function definition inside a function due to potential forking limitations...

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,171 @@
...
+                f"Expected `y_pred` and `y` to have the same shape. Got y_pred: {y_pred.shape} and y: {y.shape}."
+            )
+
+        if len(y_pred.shape) != 4 or len(y.shape) != 4:
+            raise ValueError(
+                f"Expected `y_pred` and `y` to have BxCxHxW shape. Got y_pred: {y_pred.shape} and y: {y.shape}."
+            )
+
+        if self.data_range is None:
+            self.data_range = max(y_pred.max() - y_pred.min(), y.max() - y.min())
```
vfdev-5, line 141: Does it make sense to setup `data_range` from a single batch ?

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,171 @@
...
+
+        sigma_pred_sq = F.conv2d(y_pred * y_pred, kernel, groups=channel) - mu_pred_sq
+        sigma_target_sq = F.conv2d(y * y, kernel, groups=channel) - mu_target_sq
+        sigma_pred_target = F.conv2d(y_pred * y, kernel, groups=channel) - mu_pred_target
+
+        UPPER = 2 * sigma_pred_target + C2
+        LOWER = sigma_pred_sq + sigma_target_sq + C2
+
+        ssim_idx = ((2 * mu_pred_target + C1) * UPPER) / ((mu_pred_sq + mu_target_sq + C1) * LOWER)
+        self._sum_of_batchwise_ssim += torch.mean(ssim_idx, (1, 2, 3))
```
vfdev-5, line 164: Maybe, we have to think about where to store this tensor (CPU or keep original its device)...

ignite/metrics/ssim.py
ydcjeff: Okay, I will change it.

ignite/metrics/ssim.py
ydcjeff: Okay, I will keep them separate.

ignite/metrics/ssim.py
ydcjeff: `data_range` is the maximum difference regardless of batch.
can you suggest more efficient way?

ignite/metrics/ssim.py
ydcjeff: I would like to keep its original device. What do you think?

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,178 @@
...
+__all__ = ["SSIM"]
+
+
+class SSIM(Metric):
+    """
+    Computes Structual Similarity Index Measure
+
+    Args:
+        kernel_size (list or tuple of int): Size of the gaussian kernel. Default: (11, 11)
+        sigma: Standard deviation of the gaussian kernel. Default: (1.5, 1.5)
```
vfdev-5, line 18: @ydcjeff I wanted to say that all arguments: sigma, data_range etc need to be annotated with the types like for `kernel_size`. Thanks !

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,178 @@
...
+    def __init__(
+        self,
+        kernel_size: Sequence[int] = (11, 11),
+        sigma: Sequence[float] = (1.5, 1.5),
+        data_range: float = None,
+        k1: float = 0.01,
+        k2: float = 0.03,
+        output_transform: Callable = lambda x: x,
+    ):
+        if len(kernel_size) != 2 or len(sigma) != 2:
```
vfdev-5, line 84: Maybe, we should also check the type of input args. For example, user can set `kernel_size` as a scalar (which can be also natural), so `len(kernel_size)` will fail with an unrelated error.

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,178 @@
...
+        kernel_size (list or tuple of int): Size of the gaussian kernel. Default: (11, 11)
+        sigma: Standard deviation of the gaussian kernel. Default: (1.5, 1.5)
+        data_range: Range of the image. If ``None``, it is determined from the image (max - min)
+        k1: Parameter of SSIM. Default: 0.01
+        k2: Parameter of SSIM. Default: 0.03
+        output_transform: A callable that is used to transform the
+            :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
+            form expected by the metric.
+
+    Returns:
```
vfdev-5, line 26: Technically, there is no `Returns` for the class. Let's remove it.

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,178 @@
...
+    .. code-block:: python
+
+        def process_function(engine, batch):
+            # ...
+            return y_pred, y
+        engine = Engine(process_function)
+        metric = SSIM()
+        metric.attach(engine, "ssim")
+
+    If the output of the engine is not in the format above, ``output_transform`` argument can be used to transform it.
```
vfdev-5, line 44: I think it is not that important to detail the usage of `output_transform` and `update/compute` API. But rather what kind of data and types are expected of `y_pred` and `y`. For example, `y` is a pure target 8-bit RGB image or a float normalized image etc ...

ignite/metrics/ssim.py
vfdev-5: Maybe, we should ask user to set it up and forbid to set it as `None` ? In some sense, type of `y` and `y_pred` could determine a default `data_range`.

ignite/metrics/ssim.py
vfdev-5: Yes, let's keep it on original device. Probably, there is no harm as we are not exposing it as a public attribute.

vfdev-5:
py3.5 CI complains about missing module:
- https://github.com/pytorch/ignite/pull/1217/checks?check_run_id=907660406#step:5:28

ignite/metrics/ssim.py
ydcjeff: I set `kernel_size` and `sigma` to be a list or tuple of length two to allow the users define different width or height of `kernel_size`. I know the user will want to set `kernel_size` as a scalar if the width and height are going to be same. For this, we might need another checking. That's why I let the users define in a concise way for the same or different width and height with minimal internal working code.

What do you think?

vfdev-5: Yes, I understand your choice. Ideally, it would be nice to support both types, like pytorch does for convolutions and pooling ops

ydcjeff: Alright, I will add them

edgarriba:
@vfdev-5 FYI, we have an implementation in kornia for this metric/loss

vfdev-5:
@edgarriba if you would like to promote your work, please provide an example of usage of your metrics/losses with ignite either in your repository and we can mention it in our readme, either here as a notebook tutorial. 

Anyway, seems like everyone now start to provide metrics what we did since the very beginning.

vfdev-5:
@ydcjeff thanks for the updates, I'll review it ASAP on Sunday.

edgarriba:
@vfdev-5 I'm not currently using ignite, so I'll have to learn first. I've been advocating for a `torchmetrics` package but seems every framework has its own details and interests :)

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,193 @@
...
+                    y_pred.shape, y.shape
+                )
+            )
+
+        C1 = pow(self.k1 * self.data_range, 2)
+        C2 = pow(self.k2 * self.data_range, 2)
+        device = y_pred.device
+
+        channel = y_pred.size(1)
+        kernel = self._gaussian_kernel(channel, self.kernel_size, self.sigma, device)
```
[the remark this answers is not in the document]
ydcjeff: using 1 channel is working too

tests/ignite/metrics/test_ssim.py
```diff
@@ -0,0 +1,148 @@
...
+def test_zero_div():
+    ssim = SSIM(data_range=1.0)
+    with pytest.raises(NotComputableError):
+        ssim.compute()
+
+
+def test_invalid_ssim():
+    y_pred = torch.rand(16, 1, 32, 32)
+    y = y_pred + 0.125
+    with pytest.raises(ValueError):
```
vfdev-5, line 22: Please, add `match=r"message"` to match exception. Thanks !

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,147 @@
...
+
+        if len(y_pred.shape) != 4 or len(y.shape) != 4:
+            raise ValueError(
+                "Expected y_pred and y to have BxCxHxW shape. Got y_pred: {} and y: {}.".format(y_pred.shape, y.shape)
+            )
+
+        C1 = pow(self.k1 * self.data_range, 2)
+        C2 = pow(self.k2 * self.data_range, 2)
+        channel = y_pred.size(1)
+        self.kernel = self.kernel.expand(channel, 1, -1, -1)
```
vfdev-5, line 119: This looks a bit weird as `update` is called iteratively. Maybe, we should perform this only the first time...

tests/ignite/metrics/test_ssim.py
```diff
@@ -0,0 +1,148 @@
...
+    engine = Engine(update)
+    SSIM(data_range=1.0).attach(engine, "ssim")
+
+    data = list(range(n_iters))
+    engine.run(data=data, max_epochs=1)
+
+    assert "ssim" in engine.state.metrics
+    res = engine.state.metrics["ssim"]
+
+    np_pred = np.random.rand(offset * idist.get_world_size(), 28, 28, 3)
```
vfdev-5, line 91: Shouldn't we use `y_pred` and `y` to setup `np_pred` and `np_true` ?

ignite/metrics/ssim.py
ydcjeff: Yea, I noticed that, but we need `channel` for that. I can't figure out how to do it only the first time.

tests/ignite/metrics/test_ssim.py
ydcjeff: Since `skimage` takes the last index as channels, I re-created it with numpy

vfdev-5: Technically, it is not the same data. Otherwise, you can transpose the tensor.

ydcjeff: Yes, it wouldn't be the same data.

But the difference between `y_pred` and `y` and that of `np_pred` and `np_true` is the same.
So, the computed ssim would be similar.

vfdev-5: Yes, I understood that from the code. Let's compare the same data.

ydcjeff: Okay
should I change to the same data for all tests? or only the distributed tests?

vfdev-5: Yes, it would be better for all the tests

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,147 @@
...
+        device = y_pred.device
+
+        if device is not self.kernel.device:
```
vfdev-5, line 122: I'm not sure whether this works correctly... Maybe, TPU tests failure is related to that.

ignite/metrics/ssim.py
vfdev-5: What's the shape of `self.kernel` before `expand` (after initialization) ? Maybe, we can compare the shape if it is as expected with channel or not.

tests/ignite/metrics/test_ssim.py
ydcjeff: Okay, will change that.

ignite/metrics/ssim.py
ydcjeff: After initialization, its shape is (3, 3)

vfdev-5:
@ydcjeff looks good, let me check locally this point about expanding the kernel if we can simplify this part. Also, I'd prefer rename all upper-case vars like `UPPER`, `C1` etc to be lower-cased...

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,144 @@
...
+                "Expected y_pred and y to have BxCxHxW shape. Got y_pred: {} and y: {}.".format(y_pred.shape, y.shape)
+            )
+
+        C1 = pow(self.k1 * self.data_range, 2)
+        C2 = pow(self.k2 * self.data_range, 2)
+        channel = y_pred.size(1)
+        device = y_pred.device
+        self.kernel = self.kernel.expand(channel, 1, -1, -1).to(device=device)
+
+        mu_pred = F.conv2d(y_pred, self.kernel, groups=channel)
```
vfdev-5, line 122: I'm also wondering whether it won't be more performance interesting to execute a single convolution on a stack of 
`[y_pred, y, y_pred * y_pred, y * y, y_pred * y]` instead of 5 convolutions. Typically, we can imagine that the input image is of size between 256 to 512.

vfdev-5:
@ydcjeff I'm comparing this implementation's result with scikit-image one on this example: https://scikit-image.org/docs/dev/auto_examples/transform/plot_ssim.html#sphx-glr-auto-examples-transform-plot-ssim-py
and the difference is about : 0.84172 (ski) vs 0.8366 (ignite). Any ideas why it is so ?

vfdev-5:
I updated the code fixing some minor bugs and incoherences and also raised error tolerance to 1e-6, such that probably tests wont pass while comparing with scikit-image. We need to investigate why the results are different. Seems like in scikit-image they use `trancate=3.5` in gaussian_filter (https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter.html#scipy.ndimage.gaussian_filter). Maybe, this can be in the origin of the difference.

ignite/metrics/ssim.py
ydcjeff: Ahh, yeah right, we can do that!

ydcjeff:
> @ydcjeff I'm comparing this implementation's result with scikit-image one on this example: https://scikit-image.org/docs/dev/auto_examples/transform/plot_ssim.html#sphx-glr-auto-examples-transform-plot-ssim-py
> and the difference is about : 0.84172 (ski) vs 0.8366 (ignite). Any ideas why it is so ?

@vfdev-5 I thought it could be due to `data_range`.

while we let users need to define the `data_range`, `skimage` calculates under the hood.
If we set `data_range` the same, the results are quite similar.

ignite/metrics/ssim.py
vfdev-5: I benchmarked that and there is no much gain using that. If you can benchmark that from your side, it would be nice

ydcjeff:
> I updated the code fixing some minor bugs and incoherences and also raised error tolerance to 1e-6, such that probably tests wont pass while comparing with scikit-image. We need to investigate why the results are different. Seems like in scikit-image they use `trancate=3.5` in gaussian_filter (https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter.html#scipy.ndimage.gaussian_filter). Maybe, this can be in the origin of the difference.

I think `trancate=3.5` is used to calculate `kernel_size` to be 11, so i think this would not be the origin of the difference.

vfdev-5:
We can also specify data range for skimage : https://scikit-image.org/docs/dev/api/skimage.metrics.html#skimage.metrics.structural_similarity

ignite/metrics/ssim.py
```diff
@@ -121,16 +121,16 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
         if len(self._kernel.shape) < 4:
             self._kernel = self._kernel.expand(channel, 1, -1, -1).to(device=y_pred.device)
 
-        mu_pred = F.conv2d(y_pred, self._kernel, groups=channel)
-        mu_target = F.conv2d(y, self._kernel, groups=channel)
+        input_list = [y_pred, y, y_pred * y_pred, y * y, y_pred * y]
+        output_list = [F.conv2d(x, self._kernel, groups=channel) for x in input_list]
```
vfdev-5, line 125: Here you still have 5 conv ops. IMO, I prefer the previous code as more clean. Idea was to have a single conv op on concatenated input...

ydcjeff: Ahhh,  I get it.

ydcjeff:
> We can also specify data range for skimage : https://scikit-image.org/docs/dev/api/skimage.metrics.html#skimage.metrics.structural_similarity
> 
> In my tests I specified the same data range and got those results. Difference is about 1%, this is large.

I got similar results here just like in the current tests.

ignite/metrics/ssim.py
ydcjeff: also we need to use `skimage.measure` for py3.5 since the only tagged version in req.txt has ssim metrics and support py3.5

vfdev-5:
> > We can also specify data range for skimage : https://scikit-image.org/docs/dev/api/skimage.metrics.html#skimage.metrics.structural_similarity
> > In my tests I specified the same data range and got those results. Difference is about 1%, this is large.
> 
> I got similar results here just like in the current tests.

Could you please put the code here with this test on cameraman image, such that I could reproduce it. Thanks. I edited my above message with the code I used to obtain the result I mentioned.

ignite/metrics/ssim.py
vfdev-5: Maybe, we can just add 
```python
try:
    from skimage.metrics import structural_similarity as ski_ssim
except ImportError:
    from skimage.measure ...
```

ydcjeff:
@vfdev-5 I have tested. [here is colab you can reproduce](https://colab.research.google.com/drive/1-ZCBWlSTbdURp6CMJ_6zAtVhZ5gysq3d?usp=sharing)

The main difference is we need to set `gaussian_weights=True` to use gaussian kernel.
skimage uses uniform kernel by default.

ignite/metrics/ssim.py
ydcjeff: ssim implementation of skimage between different python version is still the same.

vfdev-5:
@ydcjeff awesome, thanks a lot for digging the root of the difference !

ydcjeff:
> @ydcjeff awesome, thanks a lot for digging the root of the difference !

so now how much tolerance should we set?

ydcjeff:
> Do you think, it will be difficult to add uniform kernel as option in ignite ?

I think nope

vfdev-5:
> > @ydcjeff awesome, thanks a lot for digging the root of the difference !
> 
> so now how much tolerance should we set?

Maybe, like that 
```
pytest.approx(res.item(), abs=1e-4) == ssim_const
```

ydcjeff:
> I was thinking whether it would make sense to compute the last mean op with double precision instead of float32 ?

float64?
smth like this? `torch.sum(self._sum_of_batchwise_ssim / self._num_examples, dtype=torch.float64)`

vfdev-5:
> > I was thinking whether it would make sense to compute the last mean op with double precision instead of float32 ?
> 
> float64?
> smth like this? `torch.sum(self._sum_of_batchwise_ssim / self._num_examples, dtype=torch.float64)`

I mean let `self._sum_of_batchwise_ssim` be float64 and final division to be done as float64.

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,169 @@
...
+__all__ = ["SSIM"]
+
+
+class SSIM(Metric):
+    """
+    Computes Structual Similarity Index Measure
+
+    Args:
+        data_range (int or float): Range of the image. Typically, ``1.0`` or ``255``.
+        kernel_size (int or list or tuple of int): Size of the gaussian kernel. Default: (11, 11)
```
vfdev-5, line 18:
```suggestion
        kernel_size (int or list or tuple of int): Size of the kernel. Default: (11, 11)
```

ignite/metrics/ssim.py
```diff
@@ -0,0 +1,169 @@
...
+        sigma (float or list or tuple of float): Standard deviation of the gaussian kernel. Default: (1.5, 1.5)
```
vfdev-5, line 19:
```suggestion
        sigma (float or list or tuple of float): Standard deviation of the gaussian kernel. 
            Argument is used if ``gaussian=True``. Default: (1.5, 1.5)
```

vfdev-5:
@ydcjeff I recently merged PR with Horovod support, could you please send another PR after this one with a metric test for horovod backend like here : https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/test_accuracy.py#L804

ydcjeff:
@vfdev-5 I guess this can be merged now?

I will send another PR with horovod tests

vfdev-5:
Sorry, there was a trailing space in my docs suggestion and flake8 failed. I sent a fix and I'll merge it once the CI is done.

ydcjeff:
Okay 👌

Issue #309, opened 2018-11-01
Improve docstring of EpochMetric


vfdev-5:
As the current implementation stores all predictions and targets in RAM we need to explicitly detail such behaviour and note that this could lead to a memory error depending on the size of the input data.

vfdev-5:
Fixed by #314

ZhiliangWu:
Hi @vfdev-5, I used `EpochMetric()` as a utility to quickly save the entire output history so as they can be used for possible plotting functions (maybe the use case is a bit rare). Therefore, I defined the `compute_fn` as `lambda y_pred, y_true: torch.stack([y_pred, y_true], dim=1)`, where `y_pred.shape=(#samples, )`.  Things seem to work as expected. 

My point here is, maybe the [documentation for `EpochMetric()`](https://pytorch.org/ignite/metrics.html#ignite.metrics.EpochMetric)

> compute_fn (callable): a callable with the signature (`torch.tensor`, `torch.tensor`) takes as the input `predictions` and `targets` and **returns a scalar**. 

doesn't have to be restricted to output a `scalar`, though it's the most obvious use case. Thanks for your comments in advance.

vfdev-5:
Hi Zhiliang,

I think EpochMetric should remain a metric and compute a scalar or vector.
Maybe it would make more  sense to introduce some specific handler to store
all output history. What do you think ?

ZhiliangWu:
Hi @vfdev-5, yes, it would make more sense to have something like `entire_history` handler. Actually it took me a while to finally realize the `EpochMetric()` is able to do that, though it's designed for some more advanced computation usage.

vfdev-5:
Hi Zhiliang, if you would like to contribute a class, e.g.
ignite.handlers.stores.EpochOutputStore that appends engine's output into a
list during one epoch, it could help the library development :)

ZhiliangWu:
Hi @vfdev-5, I am happy to contribute if possible. Currently, I am a bit confused about whether it should be a handler or a metric. Maybe you can help me with it.
**From the implementation point of view it looks like a metric**. As the output values are accumulated in an online fashion from the `process_function` like any other metrics. It would be a subclass inherited from `EpochMetric()` and overwrites the `compute_fn` with some concatenation operation, which is quite straightforward to write. 
[...]

PR #1226, opened 2020-08-03
add the EpochOutputStore with tests

### LINKED ISSUE #1224: EpochOutputStorage
## 🚀 Feature

As discussed with @vfdev-5 in #309, it could be sometimes useful to provide a handler to store all output prediction history for visualization purposes. Following is my first try to implement it. 

```python
import torch

class EpochOutputStore(object):
    """EpochOutputStore handler to save output prediction and target history
    after every epoch, could be useful for e.g., visualization purposes.

    Note:
        This can potentially lead to a memory error if the output data is
    larger than available RAM.

    Args:
         output_transform (callable, optional): a callable that is used to
         transform the :class:`~ignite.engine.engine.Engine`'s
         ``process_function``'s output into the form `y_pred, y`, e.g.,
         lambda x, y, y_pred: y_pred, y

    Examples:
    .. code-block:: python
        import ...

        eos = EpochOutputStore()
        trainer = create_supervised_trainer(model, optimizer, loss)
        train_evaluator = create_supervised_evaluator(model, metrics={"acc": Accuracy()})
        eos.attach(train_evaluator)

            train_evaluator.run(train_loader)
            y_pred, y = eos.get_output()
            # plottings

    """
    def __init__(self, output_transform=lambda x: x):
        self.predictions = None
        self.targets = None
        self.output_transform = output_transform

    def reset(self):
        self.predictions = []
        self.targets = []

    def update(self, engine):
        y_pred, y = self.output_transform(engine.state.output)
        self.predictions.append(y_pred)
        self.targets.append(y)

    def attach(self, engine):
        engine.add_event_handler(Events.EPOCH_STARTED, self.reset)
        engine.add_event_handler(Events.ITERATION_COMPLETED, self.update)

    def get_output(self, to_numpy=False):
        prediction_tensor = torch.cat(self.predictions, dim=0)
        target_tensor = torch.cat(self.targets, dim=0)
        
        if to_numpy:
            prediction_tensor = prediction_tensor.cpu().detach().numpy()
            target_tensor = target_tensor.cpu().detach().numpy()
            
        return prediction_tensor, target_tensor
```

ZhiliangWu:
Fixes #1224 

Description:

add the handler `EpochOutputStore` as discussed in #1224 and #309. 

Check list:


ZhiliangWu:
@vfdev-5, I am not sure about the `from ignite.contrib.handlers import EpochOutputStore` in the test file. I referred to other test handlers and wrote it like that.

tests/ignite/handlers/test_stores.py
```diff
@@ -0,0 +1,55 @@
...
+
+    return dummy_evaluator
+
+
+@pytest.fixture
+def eos():
+    return EpochOutputStore()
+
+
+class TestEpochOutputStore:
```
vfdev-5, line 23: We do not use classes for tests, just `test_*` methods.

vfdev-5:
> @vfdev-5, I am not sure about the `from ignite.contrib.handlers import EpochOutputStore` in the test file. I referred to other test handlers and wrote it like that.

I think an import [here](https://github.com/pytorch/ignite/blob/master/ignite/contrib/handlers/__init__.py) is missing
```python
from ignite.contrib.handlers.stores import EpochOutputStore
```

ZhiliangWu:
@vfdev-5, it seems the master branch has been updated. Should I rebase the master of the branch accordingly? Or can I use this button?
image

vfdev-5:
Both are OK :)

vfdev-5:
@ZhiliangWu thanks for the updates, looks good ! I wonder whether we intend to use this also with distributed config ?
In this case we have to gather results across the processes. Let's leave it for another PR.

PR #1229, opened 2020-08-05
Added dist support for EpochMetric and other similar metrics

### LINKED ISSUE #978: Make EpochMetric work in DDP
## 🚀 Feature

Currently, metrics like `EpochMetric` which store complete history of predictions and targets, are not supporting DDP setting. 
Idea is to perform [gather](https://pytorch.org/docs/stable/distributed.html#torch.distributed.gather) on zero rank process in `compute` method to accumulate all predictions and targets across all the processes and then execute metric function...
We may also need to warn about high memory consumption of this approach.

cc @sdesrozis

vfdev-5:
Fixes #978 

Description:

- Added dist support for EpochMetric with tests
- Blocked by absence of `idist.broadcast`

Check list:


ignite/metrics/epoch_metric.py
```diff
@@ -113,6 +116,12 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
     def compute(self) -> None:
         _prediction_tensor = torch.cat(self._predictions, dim=0)
         _target_tensor = torch.cat(self._targets, dim=0)
+
+        if not self._is_reduced:
+            _prediction_tensor = idist.all_gather(_prediction_tensor)
+            _target_tensor = idist.all_gather(_target_tensor)
+            self._is_reduced = True
+
         return self.compute_fn(_prediction_tensor, _target_tensor)
```
vfdev-5, line 125: To reduce computations on CPU we can apply compute_fn only on a single rank and broadcast the result to all processes.

PR #1232, opened 2020-08-08
Make Metrics accumulate values on device specified by user

### LINKED ISSUE #1082: Metrics Implementation Question
Thanks for the great library, especially the metrics. I have a few questions to better understand the implementation:

During the update stage, why are the values converted to Python floats instead of keeping them as torch values (e.g. [here](https://github.com/pytorch/ignite/blob/master/ignite/metrics/mean_squared_error.py#L27))? This operation incurs a device->host transfer, so the operation is blocking, right? Wouldn't it be better to keep the metric values as torch values on the GPU so the `update` is async? Then, they can be converted to python floats in the `compute` method.

In the distributed case, the values are [put back in a tensor](https://github.com/pytorch/ignite/blob/master/ignite/metrics/metric.py#L202) before the all-reduce, so why not keep them as tensors to begin with?

n2cholas:
Fixes #1082

Description:
Metrics should use the `device` argument to control where `update` accumulates values. Currently, many metrics do not use this argument. This PR changes the metrics listed below so the accumulator is a tensor on the right device, and adds tests to ensure that they are indeed on the right device and that the tensors have been properly detached.

List of metrics affected:
- Accuracy
- Loss
- Mean Absolute Error
- Mean Pairwise Distance
- Mean Squared Error
- Top K Accuracy
- Precision
- Recall

Check list:

Please look at `recall.py` and `precision.py` closely. I changed the default device from `None` to `torch.device("cpu")` to maintain the old default behvaiour.


vfdev-5:
@n2cholas thanks for the PR, i'll review it ASAP today or tomorrow

vfdev-5:
@n2cholas PR looks OK from the first glance. I have to check it locally and see precisely Precision/Recall changes too.
Seems like TPU tests are not passing and it is related to the change, we have to investigate the problem...

There are accidental changes in `tests/run_cpu_tests.sh` and `tests/run_gpu_tests.sh`, can you revert them please.

n2cholas:
@vfdev-5 Thanks for the look-over. I've reverted `tests/run_cpu_tests.sh` and `tests/run_gpu_tests.sh`.

For the TPU issues, it seems like calling `tensor.to(self._device)` causes issues on XLA. I don't too much experience working with torch_xla, how can I reproduce this issue locally? Is it possible to run these tests on Colab's free TPU?

vfdev-5:
Thanks for the update ! 

> For the TPU issues, it seems like calling tensor.to(self._device) causes issues on XLA. I don't too much experience working with torch_xla, how can I reproduce this issue locally? Is it possible to run these tests on Colab's free TPU?

@n2cholas here the CI is running on CPU (not TPU) using pytorch/xla. Here is how you can locally setup it : https://github.com/pytorch/ignite/blob/master/.github/workflows/tpu-tests.yml#L20
I can help you a bit later with that after more detailed code review.

n2cholas:
Thanks for the tip, I was able to run the TPU tests locally. I'll take a look at the failing tests this weekend.

ignite/metrics/accumulation.py
```diff
@@ -38,7 +38,10 @@ class VariableAccumulation(Metric):
     _required_output_keys = None
 
     def __init__(
-        self, op: Callable, output_transform: Callable = lambda x: x, device: Optional[Union[str, torch.device]] = None
+        self,
+        op: Callable,
+        output_transform: Callable = lambda x: x,
+        device: Optional[Union[str, torch.device]] = torch.device("cpu"),
```
vfdev-5, line 44: I wonder if `Optional` is still makes sense here now. Device possible values now are "cpu", "cuda", ... and same with `torch.device`... If we put None it is the same as CPU. I think if we keep None as option, than default value should be None too. What do you think ?

ignite/metrics/accuracy.py
```diff
@@ -161,11 +161,11 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
             y = torch.transpose(y, 1, last_dim - 1).reshape(-1, num_classes)
             correct = torch.all(y == y_pred.type_as(y), dim=-1)
 
-        self._num_correct += torch.sum(correct).item()
+        self._num_correct += torch.sum(correct).to(self._device)
```
vfdev-5, line 164: Maybe, it would be better to put a comment here why we do not call detach on `correct`...

ignite/metrics/precision.py
```diff
@@ -39,13 +39,20 @@ def __init__(
...
+            self._positives = 0
+
         super(_BasePrecisionRecall, self).reset()
 
     def compute(self) -> Union[torch.Tensor, float]:
-        if not (isinstance(self._positives, torch.Tensor) or self._positives > 0):
+        is_scalar = not isinstance(self._positives, torch.Tensor) or self._positives.ndim == 0
+        if is_scalar and self._positives == 0:
             raise NotComputableError(
                 "{} must have at least one example before" " it can be computed.".format(self.__class__.__name__)
```
vfdev-5, line 57:
```suggestion
                "{} must have at least one example before it can be computed.".format(self.__class__.__name__)
```

ignite/metrics/top_k_categorical_accuracy.py
```diff
@@ -32,7 +35,7 @@ def update(self, output: Sequence) -> None:
         sorted_indices = torch.topk(y_pred, self._k, dim=1)[1]
         expanded_y = y.view(-1, 1).expand(-1, self._k)
         correct = torch.sum(torch.eq(sorted_indices, expanded_y), dim=1)
-        self._num_correct += torch.sum(correct).item()
+        self._num_correct += torch.sum(correct).to(self._device)
```
vfdev-5, line 38: Same here, please, add a comment why no `detach()`

vfdev-5:
Thanks for the update @n2cholas ! 
Actually, what was the problem with TPUs ?

ignite/metrics/accumulation.py
n2cholas: Makes total sense, thanks for catching this. I agree that this should just be `Union[str, torch.device]`. I'll make that change.

vfdev-5: @n2cholas actually, why we needed this change ? When None it is same as torch device CPU ?

n2cholas: @vfdev-5 Calling `tensor.to(None)` acts like the identity: it will keep the tensor on whatever device it was already on. I needed to make this change to fix the TPU tests. In some tests, the metric had no specified device but the inputs to `update` were on the TPU, so when we tried to add values to the accumulator we had a device mismatch.

n2cholas:
> Actually, what was the problem with TPUs ?

In a few of those tests, the metric device was not specified (it was `None`), so the accumulator tenors lived in the cpu. But, the input to update was on the TPU, so there was a device mismatch. For example, in accuracy

```python
self._num_correct += torch.sum(correct).to(self._device)
```

`self._num_correct` was on the CPU, `torch.sum(correct)` was on the TPU, and `self._device` was `None` so `torch.sum(correct).to(self._device)` was stayed on the TPU, which led to a device mismatch. In the other GPU tests, I believe the device for the metric was specified so we didn't have this issue. Setting the default device to the CPU for all the metrics made sure the `.to(self._device)` call always moved the tensor to the same device as the accumulator.

ignite/metrics/accumulation.py
vfdev-5: Okay, I see. Thanks ! Explicit device values is better then None reassigned afterward.

vfdev-5:
@n2cholas thanks for the explanation. Actually, GPU tests will be done once your branch will be merged to pytorch branch and I'll create a PR to merge to the master. Let me do this now.

PR #1238, opened 2020-08-16
[BC-breaking] Make Metrics accumulate values on device specified by user (#1232)

### LINKED ISSUE #1082: Metrics Implementation Question

vfdev-5:
* update accuracy to accumulate _num_correct in a tensor on the right device

* update loss metric to accumulate _sum in a tensor on the right device

* update mae metric to accumulate in a tensor on the right device

* update mpd metric to accumulate in a tensor on the right device

* update mse metric to accumulate in a tensor on the right device

* update top k accuracy  metric to accumulate in a tensor on the right device

* update precision and recall metrics to accumulate in tensors on the right device

* black formatting

* reverted run*.sh

* change all metrics default device to cpu except running_average

* Update ignite/metrics/precision.py

* remove Optional type from metric devices since default is cpu

* add comment explaining lack of detach in accuracy metrics

Fixes #1082 

Original PR #1232

Description:
- Merge PR to master

Check list:

cc @n2cholas


vfdev-5:
@n2cholas now I think we have to update the docstrings and main docs on metrics and how to use the device etc...
Please, send other PR to this branch from your fork such that we could update everything here and finally merge to master once we are done.

vfdev-5:
Seems like test on Horovod are failing: https://app.circleci.com/pipelines/github/pytorch/ignite/618/workflows/c74dff3a-5840-418e-837f-c43c4ac4b646/jobs/1308

n2cholas:
@vfdev-5 The new tests I wrote checking that the accumulators are on the right device are failing. It's very strange that the other two_gpu tests are passing and only horovod is failing. Is it possible to run horovod on CPU only (like it was for TPUs)? If so, I can take a stab at it.

vfdev-5:
We compile Horovod from source on the CI : https://github.com/pytorch/ignite/blob/master/.circleci/config.yml#L234
You can try either simply install it from pip, so probably it will be only basic CPU version. Either compile it as in our CI without Nccl support.

n2cholas:
@vfdev-5 I was able to set horovod up and run the distributed tests on the CPU but could not reproduce the error from CI (the tests pass for me). Would you mind taking a look, when you have a chance?

vfdev-5:
@n2cholas thanks for the info ! Yes, I'll try to take a look as soon as possible.

vfdev-5:
@n2cholas just checked one test `pytest tests/ignite/metrics/test_accuracy.py::test_distrib_hvd` on 1 GPU with Horovod installed with NCCL support and I could reproduce the error: 
```
pytest tests/ignite/metrics/test_accuracy.py::test_distrib_hvd
```
And I confirm that running the test on CPU is passing.

vfdev-5:
@n2cholas I checked the tests of accuracy and I proposed a way to improve them and also fix the issue:
https://github.com/pytorch/ignite/pull/1238/commits/b8a435553ffab6dcf78024ce32274aeee23789b8
Could you please adapt all other tests like that. Thanks !

n2cholas:
Thanks for the improved tests, I'll update the rest soon.

For the `RunningAverage` metric, I updated the metric to use `src`'s device when `src` is a Metric, and otherwise use the device parameter. However, this will only accumulate on the desired device when the input is a tensor. You will see what I mean if you take a look at [`90b5b85`](https://github.com/pytorch/ignite/pull/1239/commits/90b5b858a5862086f9783c8b082f90e0f0f0140d). Please let me know if you think the behaviour should be different.

Since I'm committing to my `metrics_impl` branch, this PR and the docs PR seem to be getting mixed up. Should I be using two different branches?

vfdev-5:
> For the RunningAverage metric, I updated the metric to use src's device when src is a Metric, and otherwise use the device parameter. However, this will only accumulate on the desired device when the input is a tensor. You will see what I mean if you take a look at 90b5b85. Please let me know if you think the behaviour should be different.

OK, I understand. Maybe we should check `RunningAverage.src.device` with input device and raise an error or override `RunningAverage.src.device`... What would be more helpful for the user ?

> Since I'm committing to my metrics_impl branch, this PR and the docs PR seem to be getting mixed up. Should I be using two different branches?

I will merge https://github.com/pytorch/ignite/pull/1239 once we addressed all visible issues and it will be in `pytorch/metrics_impl` branch. I know that this is not an ideal workflow but for instance it is an eco compromise between running gpu tests on all PRs and only on project's branches...

n2cholas:
> OK, I understand. Maybe we should check `RunningAverage.src.device` vs argument input device and raise an error if different or either override `RunningAverage.src.device`... What would be more helpful for the user ?

That makes sense to me. So, by default it will be None, and we will use `RunningAverage.src._device` in this case. If the user passes an actual device in we check to make sure it's the same as the device for `src`. I'll fix this and write the tests.

> I will merge #1239 once we addressed all visible issues and it will be in `pytorch/metrics_impl` branch. I know that this is not an ideal workflow but for instance it is an eco compromise between running gpu tests on all PRs and only on project's branches...

Got it, thanks!

vfdev-5:
> That makes sense to me. So, by default it will be None, and we will use RunningAverage.src._device in this case. If the user passes an actual device in we check to make sure it's the same as the device for src. I'll fix this and write the tests.

Maybe `None` wont work if we use output instead of metric as source... It would be better probably to keep the same init as for other metrics

n2cholas:
> Maybe `None` wont work if we use output instead of metric as source... It would be better probably to keep the same init as for other metrics

In the current logic, the init has `None` as the default, then if an output is used instead of a metric and the user doesn't specify a device, we use `torch.device("cpu")`.

My only concern with using CPU as the default argument is if the user passes in a metric on the GPU, they will have to specify the same device as an argument to RunningAverage, otherwise we throw an error. With the current implementation, we can check if `device` is `None` and then use the same device as the specified metric automatically. Then, we only throw an error when the user explicitly passes in the wrong device for running average.

Also, I would like to clarify one thing: we only use the specified device for the accumulator when the output/metric.compute produces a tensor (since when it is just a python scalar, we don't want to re-wrap it in a tensor). Is this behaviour acceptable? If so, I already made a note for this in the docstring. Should I clarify this anywhere else?

vfdev-5:
> In the current logic, the init has None as the default, then if an output is used instead of a metric and the user doesn't specify a device, we use torch.device("cpu").
My only concern with using CPU as the default argument is if the user passes in a metric on the GPU, they will have to specify the same device as an argument to RunningAverage, otherwise we throw an error. With the current implementation, we can check if device is None and then use the same device as the specified metric automatically. Then, we only throw an error when the user explicitly passes in the wrong device for running average.

Yes, you are right. Sounds good to me.


> Also, I would like to clarify one thing: we only use the specified device for the accumulator when the output/metric.compute produces a tensor (since when it is just a python scalar, we don't want to re-wrap it in a tensor). Is this behaviour acceptable? If so, I already made a note for this in the docstring. Should I clarify this anywhere else?

Do we have a cases when updated output produces a scalar and not tensor ? Maybe I didn't understand your point about  "output/metric.compute produces a tensor"...
I agree that it does not make much sense to wrap counters like `self._num_examples` etc.

> Should I clarify this anywhere else?

If this could help user to better understand what happens under the hood, I'd say yes. Well, let me see the complete doc and we can decide later if we need to add something specific...

n2cholas:
> Do we have a cases when updated output produces a scalar and not tensor ? Maybe I didn't understand your point about "output/metric.compute produces a tensor"...
> I agree that it does not make much sense to wrap counters like `self._num_examples` etc.

I believe most of the metrics's `compute` method produces a float instead of a tensor. For example, we can look at `MeanSquaredError.compute()`, which returns a float. So, if we use `RunningAverage(MeanSquaredError())`, every time we call `compute`, we have the following logic:

```python
    def compute(self) -> Union[torch.Tensor, float]:
        if self._value is None:
            self._value = self._get_src_value()
        else:
            self._value = self._value * self.alpha + (1.0 - self.alpha) * self._get_src_value()

        return self._value
```

We call MSE's `compute` via `RunningAverage._get_src_value()`, which returns a float, therefore `self._value` will be a float instead of being a tensor on the device the user specified.

I think this is fine, but I want to make sure it's what we want.

vfdev-5:
Ah OK, I understand now. Thanks for the explanation. Yes, me too, I think this is fine.

vfdev-5:
@n2cholas if you could add distritbuted tests that reproduce the problem from #1242 it would be awesome.

vfdev-5:
@fco-dev @sdesrozis please start to review

docs/source/metrics.rst
```diff
@@ -120,15 +120,15 @@ specific condition (e.g. ignore user-defined classes):
...
             self.ignored_class = ignored_class
             self._num_correct = None
             self._num_examples = None
-            super(CustomAccuracy, self).__init__(output_transform=output_transform)
+            super(CustomAccuracy, self).__init__(output_transform=output_transform, device=device)
 
         @reinit__is_reduced
         def reset(self):
-            self._num_correct = 0
+            self._num_correct = torch.tensor(0, device=self._device)
```
sdesrozis, line 131: No need to detach the tensor?

docs/source/metrics.rst
```diff
@@ -144,21 +144,28 @@ specific condition (e.g. ignore user-defined classes):
             indices = indices[mask]
             correct = torch.eq(indices, y).view(-1)
 
-            self._num_correct += torch.sum(correct).item()
+            # We must detach tensors before adding them to the internal variables. In this case, torch.eq is not
+            # differentiable, so the computation graph is detached implicitly
+            self._num_correct += torch.sum(correct).to(self._device)
```
sdesrozis, line 149: I don’t see any detach() call. Did I miss something ?

docs/source/metrics.rst
vfdev-5: Here tensor by default does not require grad, no need to detach

docs/source/metrics.rst
vfdev-5: See line 147

ignite/metrics/accuracy.py
```diff
@@ -161,11 +163,12 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
             y = torch.transpose(y, 1, last_dim - 1).reshape(-1, num_classes)
             correct = torch.all(y == y_pred.type_as(y), dim=-1)
 
-        self._num_correct += torch.sum(correct).item()
+        # Don't need to detach here because torch.eq is not differentiable, so the computation graph is detached anyway.
```
sdesrozis, line 166: It should be more readable to explicitly detach tensor.

vfdev-5: IMO, it is OK to add the comment only here

ignite/metrics/mean_absolute_error.py
```diff
@@ -17,18 +17,18 @@ class MeanAbsoluteError(Metric):
...
-        self._sum_of_absolute_errors = 0.0
+        self._sum_of_absolute_errors = torch.tensor(0.0, device=self._device)
         self._num_examples = 0
 
     @reinit__is_reduced
     def update(self, output: Sequence[torch.Tensor]) -> None:
         y_pred, y = output
         absolute_errors = torch.abs(y_pred - y.view_as(y_pred))
-        self._sum_of_absolute_errors += torch.sum(absolute_errors).item()
+        self._sum_of_absolute_errors += torch.sum(absolute_errors).detach().to(self._device)
```
sdesrozis, line 27: Does detach earlier avoid some storage in graph ? I mean detach `absolute_errors` rather than the sum.

ignite/metrics/mean_pairwise_distance.py
```diff
@@ -21,26 +21,26 @@ def __init__(
...
-        self._sum_of_distances = 0.0
+        self._sum_of_distances = torch.tensor(0.0, device=self._device)
         self._num_examples = 0
 
     @reinit__is_reduced
     def update(self, output: Sequence[torch.Tensor]) -> None:
         y_pred, y = output
         distances = pairwise_distance(y_pred, y, p=self._p, eps=self._eps)
-        self._sum_of_distances += torch.sum(distances).item()
+        self._sum_of_distances += torch.sum(distances).detach().to(self._device)
```
sdesrozis, line 39: Same question than above.

ignite/metrics/top_k_categorical_accuracy.py
```diff
@@ -32,7 +32,9 @@ def update(self, output: Sequence) -> None:
         sorted_indices = torch.topk(y_pred, self._k, dim=1)[1]
         expanded_y = y.view(-1, 1).expand(-1, self._k)
         correct = torch.sum(torch.eq(sorted_indices, expanded_y), dim=1)
-        self._num_correct += torch.sum(correct).item()
+
+        # Don't need to detach here because torch.eq is not differentiable, so the computation graph is detached anyway.
```
sdesrozis, line 36: Call explicitly detach is more readable.

ignite/metrics/mean_absolute_error.py
vfdev-5: Yes, good point. Detaching earlier we can avoid adding a node for backward op.

tests/ignite/metrics/test_mean_absolute_error.py
```diff
@@ -49,20 +49,57 @@ def update(engine, i):
             y_true[i * s + offset * rank : (i + 1) * s + offset * rank],
         )
 
-    engine = Engine(update)
+    def _test(metric_device):
+        engine = Engine(update)
 
-    m = MeanAbsoluteError()
-    m.attach(engine, "mae")
+        m = MeanAbsoluteError(device=metric_device)
```
sdesrozis, line 55: Tab used ?

tests/ignite/metrics/test_metrics_lambda.py
```diff
@@ -345,8 +345,8 @@ def update_fn(engine, i):
 
         evaluator = Engine(update_fn)
 
-        precision = Precision(average=False, device=device)
-        recall = Recall(average=False, device=device)
+        precision = Precision(average=False, device=metric_device)
```
sdesrozis, line 348: What happens if metrics are different in expression terms ?

ignite/metrics/mean_squared_error.py
```diff
@@ -17,18 +17,18 @@ class MeanSquaredError(Metric):
...
-        self._sum_of_squared_errors = 0.0
+        self._sum_of_squared_errors = torch.tensor(0.0, device=self._device)
         self._num_examples = 0
 
     @reinit__is_reduced
     def update(self, output: Sequence[torch.Tensor]) -> None:
         y_pred, y = output
         squared_errors = torch.pow(y_pred - y.view_as(y_pred), 2)
-        self._sum_of_squared_errors += torch.sum(squared_errors).item()
+        self._sum_of_squared_errors += torch.sum(squared_errors).detach().to(self._device)
```
sdesrozis, line 27: Again here.

tests/ignite/metrics/test_mean_squared_error.py
```diff
@@ -49,20 +49,59 @@ def update(engine, i):
             y_true[i * s + offset * rank : (i + 1) * s + offset * rank],
         )
 
-    engine = Engine(update)
+    def _test(metric_device):
+        engine = Engine(update)
 
-    m = MeanSquaredError()
-    m.attach(engine, "mse")
+        m = MeanSquaredError(device=metric_device)
```
sdesrozis, line 55: Tab used ?

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -46,25 +46,29 @@ def _test_distrib_integration(device, tol=1e-6):
     def update(engine, i):
         return y_preds[i * s : (i + 1) * s], y_true[i * s + offset * rank : (i + 1) * s + offset * rank]
 
-    engine = Engine(update)
+    def _test(metric_device):
+        engine = Engine(update)
 
-    m = RootMeanSquaredError()
-    m.attach(engine, "rmse")
+        m = RootMeanSquaredError(device=metric_device)
```
sdesrozis, line 52: Tab used ?

tests/ignite/metrics/test_running_average.py
```diff
@@ -305,61 +305,84 @@ def _test_distrib_on_metric(device):
...
+        data = list(range(n_iters))
+        np.random.seed(12)
+        all_y_true_batch_values = np.random.randint(
+            0, n_classes, size=(idist.get_world_size(), n_epochs * n_iters, batch_size)
+        )
+        all_y_pred_batch_values = np.random.rand(idist.get_world_size(), n_epochs * n_iters, batch_size, n_classes)
 
-    y_true_batch_values = iter(all_y_true_batch_values[rank, ...])
-    y_pred_batch_values = iter(all_y_pred_batch_values[rank, ...])
+        y_true_batch_values = iter(all_y_true_batch_values[rank, ...])
```
sdesrozis, line 316: Tab used ?

tests/ignite/metrics/test_metrics_lambda.py
vfdev-5: That's a good question, I think it does not matter as lambda works on the results of compute which are float numbers... But worth to check

sdesrozis:
Few comments / questions

tests/ignite/metrics/test_mean_squared_error.py
vfdev-5: It just shows the diff with previous code, current one is indented and lint passes.

ignite/metrics/mean_pairwise_distance.py
vfdev-5: Here maybe we can even detach `y_pred` before calling `pairwise_distance`...

docs/source/metrics.rst
sdesrozis: Ok I see. I definitely prefer explicit w/o comment rather than implicit w/ comments 😊

tests/ignite/metrics/test_mean_squared_error.py
sdesrozis: Ok, strange rendering for me. I thought it was due to tab 😓

vfdev-5:
@n2cholas could you please check time of TPU tests execution on Colab (for example test_precision or test_recall) of current branch and the master. Seems like XLA CPU tests time increased from 15 minutes to 47 minutes... If we have something similar on TPUs, it would be better not to use XLA as metric device and directly pass to CPU.

ignite/metrics/mean_pairwise_distance.py
n2cholas: I agree. If we detach `y_pred`, I think we should also detach `y`, since the user may backprop through it (e.g. for adversarial learning).

vfdev-5: @n2cholas I think when we do an adversarial learning the target is already "detached" as we do not want to update model's parameters that generated `y`... But it does not cost too much to apply it here though

vfdev-5:
@n2cholas this colab may help https://colab.research.google.com/drive/1yA0DJvYf9US_aheNovpB3iPVGQDHjVyU with TPU time testings...

vfdev-5:
I wonder what would be better to do:
- detach tensors in output of `update(output)` inside `iteration_completed` 
or 
- explicitly in each `update` function.

In the first case, as we do not restrict the API  and user can use either `attach` method or `update/compute` methods. If  `update/compute` methods are used, that it is up to user to detach things before calling `update` or use `torch.no_grad()` context manager. Is it acceptable as requirement ?

ignite/metrics/mean_absolute_error.py
n2cholas: Fixed this in 6edd30d.

ignite/metrics/mean_pairwise_distance.py
n2cholas: Fixed this in 6edd30d.

tests/ignite/metrics/test_metrics_lambda.py
n2cholas: I don't quite understand, what is the potential issue here?

ignite/metrics/accumulation.py
```diff
@@ -67,6 +72,9 @@ def update(self, output: Union[Any, torch.Tensor, numbers.Number]) -> None:
                 output = output.to(self._device)
 
         self.accumulator = self._op(self.accumulator, output)
+        if isinstance(self.accumulator, torch.Tensor):
```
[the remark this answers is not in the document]
n2cholas: Good point, removed the redundant call in 375f91e.

n2cholas:
> Seeing the colab, this corresponds to single TPU execution and not CPU. If `NUM_TPU_WORKERS=8` is set we run on 8 devices in distrib mode, otherwise it picks a single TPU and run on it.

Thanks for catching this!

> I propose to create a script to measure execution times, a sort of integration test with Engine and metrics:
> 
> * in a single test, let's generate random `y_pred`, `y` to compute metrics on for 100 iterations, 3 epochs.
> * reimplement metrics like Accuracy and Accumulation to keep current master implementation (let's call it "as number" implementation vs this PR's "as tensor" implementation)
> * run benchmarks per metric on both implementations
>   
>   * let's run 3-5 time each test for better time stats
>   * in various confs:
>     
>     * only on CPU, single device, no distributed
>     * only on GPU, single device, no distributed
>     * only on GPUs, two devices, distributed
>     * only on TPU, single device, no distributed
>     * only on TPUs, 8 devices, distributed
> 
> What do you think ?

This sounds good. I'll work on this soon.

> I wonder what would be better to do:
> 
> * detach tensors in output of `update(output)` inside `iteration_completed`
>   or
> * explicitly in each `update` function.
> 
> In the first case, as we do not restrict the API and user can use either `attach` method or `update/compute` methods. If `update/compute` methods are used, that it is up to user to detach things before calling `update` or use `torch.no_grad()` context manager. Is it acceptable as requirement ?

I would prefer to detach explicitly in each `update` function so the user doesn't have to think about it, as in the current master users don't need to worry about detaching. But I'm not too familiar with with how the metrics are used in the engine (I use ignite solely for the metrics), so I will defer to your judgement on this decision.


I addressed the comments in my branch (`n2cholas/ignite:metrics_impl`), and am not sure how to push those commits into this branch. Should I open a separate pull request...?

vfdev-5:
> I would prefer to detach explicitly in each update function so the user doesn't have to think about it, as in the current master users don't need to worry about detaching. But I'm not too familiar with with how the metrics are used in the engine (I use ignite solely for the metrics), so I will defer to your judgement on this decision.

Metrics reset-update-compute API can be introduced as 
```
acc = Accuracy()

acc.reset()

with torch.no_grad():
    acc.update((y_pred1, y1))

with torch.no_grad():
    acc.update((y_pred2, y2))

print(acc.compute())
```
In this case, `detach` is not needed in the `update`. Otherwise, we have ensure everywhere that we detached `y_pred` and `y`. This is just 2 additional lines of code, not a big deal though.


> But I'm not too familiar with with how the metrics are used in the engine (I use ignite solely for the metrics)

@n2cholas try it with the Engine :)

> I addressed the comments in my branch (n2cholas/ignite:metrics_impl), and am not sure how to push those commits into this branch. Should I open a separate pull request...?

Well, technically, yes, you have to send PRs to this branch. But if you plan to work more this week, exceptionally I can give you a temporary write permission to push directly to this branch.

n2cholas:
> Metrics reset-update-compute API can be introduced as
> 
> ```
> acc = Accuracy()
> 
> acc.reset()
> 
> with torch.no_grad():
>     acc.update((y_pred1, y1))
> 
> with torch.no_grad():
>     acc.update((y_pred2, y2))
> 
> print(acc.compute())
> ```
> 
> In this case, `detach` is not needed in the `update`. Otherwise, we have ensure everywhere that we detached `y_pred` and `y`. This is just 2 additional lines of code, not a big deal though.

I see! What about changing the `update` method so it has the `torch.no_grad()` decorator, as follows:

```python
@torch.no_grad()
@reinit__is_reduced
def update(self, output):
    ...
```

I'm just wondering if there's a good reason to push this overhead to the user. Is there any case where the user would not want the tensors to be deatched?

> @n2cholas try it with the Engine :)

Haha will do on my next PyTorch project!

> Well, technically, yes, you have to send PRs to this branch. But if you plan to work more this week, exceptionally I can give you a temporary write permission to push directly to this branch.

No worries, I was just curious. I will work on the benchmarks then create a PR as usual.

n2cholas:
I've run the integration benchmarks on CPU, GPU, and multi-GPU. [Here](https://gist.github.com/n2cholas/f674b1f655bb49cf473d9476df631f00) is a gist with the code and [here](https://docs.google.com/spreadsheets/d/1efBJ5psEq0bCb2PO26xOljzr1iDSVSvMz1Ae_xvQS6U/edit?usp=sharing) are the numbers I got (single gpu tests are on titan V, multi-gpu on titan V + titan RTX). The error represents the margin of error for a 95% confidence interval (you can see how it's generated in the gist). Overall, the results are what we expect: the new implementation is slightly slower for CPU workloads, but slightly faster for GPU and distributed workloads.

I was trying to see what was going on with XLA, and I think for now we should not allow users to keep metrics on an XLA device, and instead they should use the metrics on the normal CPU when training on the TPU. We can throw an error if they try to set the metric's device to an XLA device. I'm not sure how to fix the performance issues we were seeing. Does this sound good @vfdev-5?

vfdev-5:
@n2cholas thanks a lot for the benchmark ! It is very helpful to see complete picture of that !

> I was trying to see what was going on with XLA, and I think for now we should not allow users to keep metrics on an XLA device, and instead they should use the metrics on the normal CPU when training on the TPU. We can throw an error if they try to set the metric's device to an XLA device. I'm not sure how to fix the performance issues we were seeing. Does this sound good @vfdev-5?

I agree that we can use CPU instead of XLA for now, before finding out where the problem and if it can be solved somehow...
Probably, the issue is that XLA compiler should create a graph for all those updates and thus it takes time...

tests/ignite/metrics/test_metrics_lambda.py
vfdev-5: I think @sdesrozis wanted to know what would happen if we do (Sylvain, correct me I'm wrong)
```python
precision = Precision(average=False, device="cpu")
recall = Recall(average=False, device="gpu")
my_metric = precision * recall

my_metric.attach(engine, "my_metric")
```

n2cholas: I understand now. Currently, this will fail when you call `compute` since you will be trying to multiply tensors on different devices. Is this behaviour ok? 

Otherwise, we can fail early by adding the following check to all the arithmetic operations in `Metric`:

```python
    def __mul__(self, other):
        from ignite.metrics.metrics_lambda import MetricsLambda
        if self._device != self._device:
            raise ValueError("Metrics combined with arithmetic must be on the same device. "
                             "Got self on {} and other on {}".format(self._device, other._device))
        return MetricsLambda(lambda x, y: x * y, self, other
```

vfdev-5: Well, I thought in most of the cases of our metrics `Metric.compute()` returns a scalar on cpu. That's why I thought there is no issue. I wonder if we would like to keep metric on other devices than cpu ?

n2cholas: Yes, but sometimes (e.g. Precision and Recall when `average=False`) we return an entire tensor. I think it makes sense to keep the metrics on other devices sometimes (e.g. the GPU so the updates are quicker). 

In my opinion, the best solution is to move all the tensors to the CPU during the `compute` step for `MetricsLambda` (instead of throwing any errors). What do you think?

vfdev-5: I agree 👍

n2cholas: Done in bcc3cbb.

ignite/metrics/accuracy.py
vfdev-5, line 166: Let's remove this comment and call detach on the output as in other classes:
```python
# y_pred, y = output
y_pred, y = output[0].detach(), output[1].detach()
```

vfdev-5 (commented):
@n2cholas I left few comments. I still have the question about explicit detach calls. Now we have some classes where we call it directly on the output and other where we are not calling it at all... I think it would be better to make everything in the same way : call detach on the output. What do you think ?

ignite/metrics/accumulation.py
```diff
@@ -67,6 +72,7 @@ def update(self, output: Union[Any, torch.Tensor, numbers.Number]) -> None:
                 output = output.to(self._device)
 
         self.accumulator = self._op(self.accumulator, output)
+
```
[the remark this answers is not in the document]
n2cholas: Yes, we should, thanks for catching this.

ignite/metrics/accuracy.py
n2cholas: Sounds good!

n2cholas:
@vfdev-5 I agree, I'll ensure all the metrics consistently call detach on the outputs (and fix those horovod tests too).

tests/ignite/metrics/test_accumulation.py
```diff
@@ -198,105 +198,123 @@ def compute_mean_std(engine, batch):
...
-    y_true = y_true.to(device)
+        for y in y_true:
+            mean_var.update(y)
 
-    for y in y_true:
-        mean_var.update(y)
+        m = mean_var.compute()
+        log_y_true = torch.log(y_true)
+        log_y_true = idist.all_reduce(log_y_true)
+        assert m.item() == pytest.approx(torch.exp(log_y_true.mean(dim=0) / idist.get_world_size()).item())
```
vfdev-5, line 293: We have some precision issue with 8 TPUs and can be fixed
```suggestion
        np.testing.assert_almost_equal(
            m.item(), torch.exp(log_y_true.mean(dim=0) / idist.get_world_size()).item(), decimal=4
        )
```

tests/ignite/metrics/test_accumulation.py
```diff
@@ -198,105 +198,123 @@ def compute_mean_std(engine, batch):
...
+        for y in y_true:
+            mean_var.update(y)
 
-    for y in y_true:
-        mean_var.update(y)
+        m = mean_var.compute()
+        log_y_true = torch.log(y_true)
+        log_y_true = idist.all_reduce(log_y_true)
+        np.testing.assert_almost_equal(
+            m.cpu().numpy(), torch.exp(log_y_true.mean(dim=0) / idist.get_world_size()).cpu().numpy(), decimal=5
```
vfdev-5, line 306:
```suggestion
            m.cpu().numpy(), torch.exp(log_y_true.mean(dim=0) / idist.get_world_size()).cpu().numpy(), decimal=4
```
same here

vfdev-5 (commented):
@n2cholas thanks a lot for working on this huge PR !
I'll fix the CI failing on TPUs and we can merge it

vfdev-5:
@n2cholas we are preparing a v0.4.2 release and I think as being BC-breaking change this commit will be present since v0.5.0.

n2cholas:
@vfdev-5 Sounds good, thank you!

PR #1245, opened 2020-08-24
Allow passing keyword arguments to save function on checkpoint.

### LINKED ISSUE #1244: Backwards compability for torch.save in ModelCheckpoint
## 🚀 Feature

Pytorch made some changes to torch.save (https://pytorch.org/docs/stable/generated/torch.save.html)

> The 1.6 release of PyTorch switched torch.save to use a new zipfile-based file format. torch.load still retains the ability to load files in the old format. If for any reason you want torch.save to use the old format, pass the kwarg _use_new_zipfile_serialization=False.

It would be useful to have control over this flag when instantiating ModelCheckpoint.

Possible ways that I can think of:
- Have some *args, **kwargs passed on to the save function
- Be able to pass the save function itself to ModelCheckpoints

kilsenp:
Fixes #1244 

Description:
Changes are motivated to deal with backwards compability issue first raised in #1244. 

Now, additional keyword arguments can be passed to the constructor of `DiskSaver`, which are then during the call passed to the save functions to have more control there.
Both, the XLA and the standard torch.save function have keyword arguments.

https://pytorch.org/xla/release/1.5/index.html#torch_xla.core.xla_model.save
https://pytorch.org/docs/stable/generated/torch.save.html

ModelCheckpoint is also extended for the same reason.

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -526,7 +526,7 @@ class DiskSaver(BaseSaveHandler):
         create_dir (bool, optional): if True, will create directory ``dirname`` if it doesnt exist.
         require_empty (bool, optional): If True, will raise exception if there are any files in the
             directory ``dirname``.
-        **kwargs: Keyword arguments accepted for `torch.save` or `xm.save`.
+        **kwargs: Accepted keyword arguments for `torch.save` or `xm.save`.
     """
 
     def __init__(
@@ -655,7 +655,7 @@ class ModelCheckpoint(Checkpoint):
         archived (bool, optional): Deprecated argument as models saved by `torch.save` are already compressed.
         include_self (bool): Whether to include the `state_dict` of this object in the checkpoint. If `True`, then
             there must not be another object in ``to_save`` with key ``checkpointer``.
-        **kwargs: Keyword arguments accepted for `torch.save` or `xm.save` in `DiskSaver`.
+        **kwargs: Accepted keyword arguments for `torch.save` or `xm.save` in `DiskSaver`.
 
     Examples:
         >>> import os
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -526,14 +526,16 @@ class DiskSaver(BaseSaveHandler):
         create_dir (bool, optional): if True, will create directory ``dirname`` if it doesnt exist.
         require_empty (bool, optional): If True, will raise exception if there are any files in the
             directory ``dirname``.
+        **kwargs: Keyword arguments accepted for `torch.save` or `xm.save`.
     """
 
     def __init__(
-        self, dirname: str, atomic: bool = True, create_dir: bool = True, require_empty: bool = True,
+        self, dirname: str, atomic: bool = True, create_dir: bool = True, require_empty: bool = True, **kwargs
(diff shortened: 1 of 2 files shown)


ignite/handlers/checkpoint.py
```diff
@@ -526,14 +526,16 @@ class DiskSaver(BaseSaveHandler):
         create_dir (bool, optional): if True, will create directory ``dirname`` if it doesnt exist.
         require_empty (bool, optional): If True, will raise exception if there are any files in the
             directory ``dirname``.
+        **kwargs: Keyword arguments accepted for `torch.save` or `xm.save`.
```
vfdev-5, line 529:
```suggestion
        **kwargs: Accepted keyword arguments for `torch.save` or `xm.save`.
```

ignite/handlers/checkpoint.py
```diff
@@ -653,6 +655,7 @@ class ModelCheckpoint(Checkpoint):
         archived (bool, optional): Deprecated argument as models saved by `torch.save` are already compressed.
         include_self (bool): Whether to include the `state_dict` of this object in the checkpoint. If `True`, then
             there must not be another object in ``to_save`` with key ``checkpointer``.
+        **kwargs: Keyword arguments accepted for `torch.save` or `xm.save` in `DiskSaver`.
```
vfdev-5, line 658:
```suggestion
        **kwargs: Accepted keyword arguments for `torch.save` or `xm.save` in `DiskSaver`.
```

vfdev-5 (commented):
@kilsenp thanks for the PR ! Looks good. Could you please add a test with a usage of a kwargs (maybe also adding `@pytest.skipIf(torch version smaller 1.6.0)`) ?

kilsenp:
With the current approach it seems to me that it's a bit hard to test if the keyword is passed correctly to the save_fn.

I added therefore two tests
- One to check if this one specific keyword _use_new_zipfile_serialization keyword is accepted (introduced in torch 1.4.0)
- One to check if the keywords are passed correctly by passing an unknown keyword and expecting it to fail.

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -592,6 +593,31 @@ def _test_existance(atomic, _to_save, expected):
...
+
+    saver = DiskSaver(dirname, create_dir=False, _use_new_zipfile_serialization=False)
+    fname = "test.pt"
+    saver(to_save, fname)
+    fp = os.path.join(saver.dirname, fname)
+    assert os.path.exists(fp)
+    saver.remove(fname)
+
+
+@pytest.mark.xfail
```
vfdev-5, line 611: Can't we use
```python
with pytest.raises(TypeError, match=r"error message to match"):
    saver(to_save, fname)
```

vfdev-5 (commented):
@kilsenp thanks for the update !
Just left a nit to improve the test codebase.

tests/ignite/handlers/test_checkpoint.py
kilsenp: Sure, I already had the hunch there was an alternative to xfail being used..

vfdev-5 (approved):
LGTM. Thanks @kilsenp !

PR #1260, opened 2020-08-31
fix issue 1249

### LINKED ISSUE #1249: CycleGAN notebook is broken with updated ParamGroupScheduler
https://github.com/pytorch/ignite/blob/c2c4a368a8666070633242c478c3b4151a05d49d/ignite/contrib/handlers/param_scheduler.py#L960

@sdesrozis above code brakes our cycle gan notebook where we do
```python
gen_lr_scheduler = PiecewiseLinear(optimizer_D, param_name='lr', milestones_values=milestones_values)
desc_lr_scheduler = PiecewiseLinear(optimizer_G, param_name='lr', milestones_values=milestones_values)

lr_scheduler = ParamGroupScheduler([gen_lr_scheduler, desc_lr_scheduler], 
                                   names=['gen_lr_scheduler', 'desc_lr_scheduler'])
```

Can we simply remove that check ?

sdesrozis:
Fixes #1249

Description:

Update ParamScheduler (and ConcatScheduler) to handle schedulers built on different optimizers.

To do so, internal class `_ParamGroupsOptimizer` is introduced for packing parameters from optimizers and let `ParamScheduler` work as usual.

Check list:


sdesrozis:
@vfdev-5 notebook ok

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -468,9 +482,8 @@ def __init__(self, schedulers, durations, save_history=False):
         self.schedulers = schedulers
         self.durations = durations
 
-        self.optimizer = self.schedulers[0].optimizer
-        if not (all(id(s.optimizer) == id(self.optimizer) for s in self.schedulers)):
```
vfdev-5, line 472: I think the check here is valid as we unlikely want to concat lr schedulers working on different optimizers such that at first the first scheduler is applied on the first optimizer and then the second one on another optimizer.

sdesrozis: let me think about it

vfdev-5: Sounds good

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -402,6 +402,31 @@ def get_param(self):
...
+        self.param_groups = [s.optimizer_param_groups[0] for s in schedulers]
+
+    def state_dict(self):
+        return self.param_groups
+
+    def load_state_dict(self, state):
+        self.param_groups = state
+
+
+def _create_optimizer_list(schedulers):
```
vfdev-5, line 420: @sdesrozis i'm not sure to understand the purpose of this method as it is used only inside `ConcatScheduler` where we need a single optimizer only.

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -586,11 +612,10 @@ def simulate_values(cls, num_events, schedulers, durations, param_names=None, **
         # not perturb original scheduler.
         with tempfile.TemporaryDirectory() as tmpdirname:
             cache_filepath = Path(tmpdirname) / "ignite_lr_scheduler_cache.pt"
-            objs = {"lr_scheduler_{}".format(i): s.state_dict() for i, s in enumerate(schedulers)}
-            # all schedulers should be related to the same optimizer
-            objs["optimizer"] = schedulers[0].optimizer.state_dict()
+            scheduler_objs = {"lr_scheduler_{}".format(i): s.state_dict() for i, s in enumerate(schedulers)}
```
vfdev-5, line 615: Why can't we keep previous code here ?

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -980,10 +1005,7 @@ def __init__(self, schedulers: List[ParamScheduler], names: Optional[List[str]]
 
         self.schedulers = schedulers
         self.names = names
-
-        self.optimizer = self.schedulers[0].optimizer
-        if not (all(id(s.optimizer) == id(self.optimizer) for s in schedulers)):
-            raise ValueError("schedulers should be related to same optimizer")
+        self.optimizer = _ParamGroupsOptimizer(schedulers)
```
vfdev-5, line 1008: I think I missed that in your previous PR. But why `ParamGroupScheduler` is derived from `ParamScheduler` now ?
I think it does not match correctly the fields like `self.optimizer` or `self.optimizer_param_groups` etc as it can have multiple optimizers and groups at the same time...

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -402,6 +402,31 @@ def get_param(self):
         return self.start_value + ((self.end_value - self.start_value) / 2) * (1 - math.cos(math.pi * cycle_progress))
 
 
+class _ParamGroupsOptimizer:
+    """Internal tools to concat a list of parameters from different optimizers.
+    """
+
+    def __init__(self, schedulers):
+        self.schedulers = schedulers
+        self.param_groups = [s.optimizer_param_groups[0] for s in schedulers]
```
vfdev-5, line 411: Why only zero group `s.optimizer_param_groups[0]` ?

vfdev-5 (commented):
@sdesrozis thanks for working on the PR ! I really appreciate your effort despite the conditions !
I left some comments to discuss about the implementation and the choices made.

ignite/contrib/handlers/param_scheduler.py
sdesrozis: You are right, this method should be in class `ConcatScheduler`.

ignite/contrib/handlers/param_scheduler.py
sdesrozis: because `ParamGroupScheduler` could be in the list and as you mentionned above, it's a list of optimizers.

ignite/contrib/handlers/param_scheduler.py
sdesrozis: Yes, it is derived from. 
`self.optimizer` is now an object `_ParamGroupsOptimizer` encapsulating multiple optimizers. It works because API allows objects with `param_groups`. In fine, `ParamGroupScheduler` packs optimizers and relevant `param_groups` from them. I think that is correct.

ignite/contrib/handlers/param_scheduler.py
sdesrozis: Good catch. `optimizer_param_groups` is a list of parameters from legacy optimizer or `_ParamGroupsOptimizer`.

I think I have to think about all comments below to push a better solution...

ignite/contrib/handlers/param_scheduler.py
sdesrozis: I thought about the implementation. I feel you are right, the problem is related to `ParamGroupsScheduler` derived from `ParamScheduler`. It's easy to write buggy code.
The point that I have to solve is the implementation of `create_lr_scheduler_with_warmup` which mix `ParamGroupsScheduler` and `ConcatScheduler`. I have an idea about that. Let's see.

vfdev-5: Thanks @sdesrozis !

sdesrozis: Ok I think i got it. 

To have multi params group in warmup (issue 856), I modified `ParamGroupsScheduler` to work with `ConcatScheduler`. I think the idea was good but not the actual implementation... First, `ParamGroupsScheduler` (used with `ConcatScheduler`) should be related to the same optimizer. In that case, only one call of `scheduler()` should be done (maybe a bug here). Second, it should inherit from `ParamScheduler` and pass the optimizer.

In other hand, we need an implementation of `ParamGroupsScheduler` that packs schedulers related to different optimizers. In fact, we must be sure that schedulers with same optimizers are not embbeded to avoid mutliple calls in one call... (I will check that).

@vfdev-5 thoughts ? Sorry, this item is a bit tricky.

sdesrozis:
Close to submit a more concise PR

Issue #1242, opened 2020-08-17
Handling empty datasets in distributed metric computation


linhr:
## 🐛 Bug description

Metric computation does not work properly in distributed settings when some processes do not handle any batch in the dataset. It becomes a problem when small validation or test datasets are distributed to processes in an imbalanced manner.

### How to Reproduce

Create a Python script named `main.py` with the following content.

```python
import torch
import ignite.distributed as idist
from torch.utils.data import IterableDataset, DataLoader
from ignite.metrics import Loss
from ignite.engine.engine import Engine
from ignite.engine.events import Events

class SampleDataset(IterableDataset):
    def __iter__(self):
        if idist.get_rank() == 0:
            yield torch.zeros((2, 3)), torch.ones((2, 3))

def report_metrics(engine):
    print(engine.state.metrics)

def test(local_rank):
    data_loader = DataLoader(SampleDataset(), batch_size=None)
    engine = Engine(lambda _engine, batch: batch)
    Loss(torch.nn.BCELoss(reduction="mean")).attach(engine, "loss")
    engine.add_event_handler(Events.COMPLETED, report_metrics)
    engine.run(data_loader)

with idist.Parallel(backend="gloo") as parallel:
    parallel.run(test)
```

Run the following command inside a CPU Docker container with PyTorch and Ignite installed.

```bash
python -m torch.distributed.launch --nproc_per_node=2 --use_env main.py
```

### Problem 1

The command terminated with an error. Part of the output is shown below.

```text
terminate called after throwing an instance of 'gloo::EnforceNotMet'
  what():  [enforce fail at /opt/conda/conda-bld/pytorch_1595629403081/work/third_party/gloo/gloo/transport/tcp/pair.cc:490] op.preamble.length <= op.nbytes. 8 vs 4
```

It seems there is type inconsistency (`int` vs `float`) inside `idist.all_reduce()` when calling `compute()`, because not all processes have called `update()` at least once. A simple fix could be changing [this line](https://github.com/pytorch/ignite/blob/564b7450efd820526ff09b925f0e212e6bd00863/ignite/metrics/loss.py#L48) to `self._sum = 0.0`.

However this issue could affect other metrics as well. We probably need unit tests for such scenario for all metrics.

### Problem 2

In the above script, if we change `Loss(...)` to the precision or recall metric (e.g. `Precision()`), we get the following error message.

```text
Engine run is terminating due to exception: Precision must have at least one example before it can be computed..
```

[...]

PR #1266, opened 2020-09-07
Updates to metrics_impl

n2cholas:
Description:

Addresses comments in #1238 and prevents metrics from being accumulated on an XLA device.

Check list:


ignite/metrics/accumulation.py
```diff
@@ -51,6 +51,8 @@ def __init__(
         self.num_examples = None
         self._op = op
 
+        if torch.device(device).type == "xla":
```
vfdev-5, line 54: @n2cholas Maybe we can do this check once inside `Metric.__init__`, what do you think ?

vfdev-5 (commented):
@n2cholas thanks for the updating PR ! I left a comment and fixed isort issue.

ignite/metrics/accumulation.py
n2cholas: I initially had this line in `Metric.__init__`, but noticed that the `ConfusionMatrix` had tests on XLA, so I did it this way instead so I didn't break those tests. I would definitely prefer moving that line to `Metric.__init__`, if it's ok to break the existing `ConfusionMatrix` functionality. What do you think?

vfdev-5: Sorry for delayed answer. 

> if it's ok to break the existing ConfusionMatrix functionality

I think we can do that. You are talking about this test, right ? 
https://github.com/pytorch/ignite/blob/d192a8f43c63e7da922110474603389b1c0e030c/tests/ignite/metrics/test_confusion_matrix.py#L617

n2cholas: Yes, I was talking about that test. Sounds good, I'll work on moving the check to `Metric.__init__`!

ignite/metrics/precision.py
```diff
@@ -127,18 +127,6 @@ def thresholded_output_transform(output):
             default, CPU.
 
     """
-
-    def __init__(
```
vfdev-5, line 131: I think this was kept for proper docs rendering. Can you please check that ?

vfdev-5 (commented):
Looks good ! Thanks @n2cholas  !
Remains to fix black formatting and the question about Precision/Recall `__init__`.

ignite/metrics/precision.py
n2cholas: Reverted the change! The Precision and Recall docstrings render like before now.

n2cholas:
@vfdev-5 Done! Everything's black formatted and the Precision/Recall docstrings look like they did before.

vfdev-5:
@n2cholas thanks ! Let CI do its job and we can merge it to the branch for GPU checkings...
I think we are almost done with the FR, right ?

n2cholas:
@vfdev-5 Yes, I believe so! We discussed adding tests for #1242, but I think that should be a separate PR (since there's discussion on that issue about ways to fix the problem more generally).

vfdev-5:
I agree. Maybe, it would be good to add a test for `MetricLambda` if components has different devices, https://github.com/pytorch/ignite/pull/1238#discussion_r485717121

PR #1277, opened 2020-09-09
Fix TPU tests for metrics_impl branch

n2cholas:
Fixes broken tests in #1238 . `master` had a new `EpochMetric` test on XLA which caused the TPU tests to fail when we merged it into `metrics_impl`. This PR fixes that test.


vfdev-5:
@n2cholas thanks !

Issue #1263, opened 2020-09-04
ProgressBar ETA for an IterableDataset


g-karthik:
## ❓ Questions/Help/Support

We have a piece of code which uses `ignite` successfully and was originally written for a regular `TensorDataset` (adapted from [this](https://github.com/huggingface/transfer-learning-conv-ai/blob/master/train.py)).

But then we adapted it for an `IterableDataset` because we wanted to train on a much larger dataset (on the order of GBs). How does the `ProgressBar` estimate this ETA for an `IterableDataset`?

If there is any working sample code you can point me to on using `IterableDataset` instances successfully with ignite, that would be great!

@vfdev-5

vfdev-5:
> How does the ProgressBar estimate this ETA for an IterableDataset?

@g-karthik engine can estimate the epoch length using the first epoch and then tqdm total is set according to that : 
https://github.com/pytorch/ignite/blob/d8a54f6aa230f52b8f77162b8dcdbe9baa7f6c35/ignite/contrib/handlers/tqdm_logger.py#L48-L51

Here is an example of how it would work
```python
# ignite 0.4.1
import time

from torch.utils.data import DataLoader, IterableDataset
from ignite.engine import Engine
from ignite.contrib.handlers import ProgressBar


class MyIterableDataset(IterableDataset):
    def __init__(self):
        super(MyIterableDataset).__init__()
        self.start = 0
        self.end = 23

    def __iter__(self):
        return iter(range(self.start, self.end))


num_workers = 0
ds = MyIterableDataset()
data_loader = DataLoader(ds, num_workers=num_workers)

def foo(e, b):
    time.sleep(0.2)

engine = Engine(foo)

ProgressBar().attach(engine)
engine.run(data_loader, epoch_length=None, max_epochs=5)
```

g-karthik:
> engine can estimate the epoch length using the first epoch and then tqdm total is set according to that

@vfdev-5 but does this mean that the ETA reported for the first epoch is likely to be unrealistic? i.e., only from the second epoch the ETAs will be accurate?

vfdev-5:
@g-karthik normally, during the first epoch you shouldn't have any ETA reported but something like that : 
```
Epoch [1/5]: [8/?] 0%   [....] [00:04<00:00]
```
and only from the second epoch tqdm can estimate the time as it has now the epoch length:
```
Epoch [2/5]: [8/23] 35% [00:01<00:03]
```

g-karthik:
[...]

PR #1279, opened 2020-09-10
Rename Epoch to Iterations when using epoch_length with max_epochs=1

### LINKED ISSUE #1276: Rename Epoch to Iterations when using epoch_length with max_epochs=1
## 🚀 Feature

Per the discussion in https://github.com/pytorch/ignite/issues/1263, we'd like to ensure that the progress bar uses the name "Iterations" instead of "Epoch" when the `epoch_length` is passed to the trainer for `max_epochs=1`.

M3L6H:
Fixes #1276

Description:
Per the discussion in #1263, we'd like to ensure that the progress bar uses the name "Iterations" instead of "Epoch" when the epoch_length is passed to the trainer for max_epochs=1.

Check list:

### DIFF
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -49,7 +49,11 @@ class _OutputHandler(BaseOutputHandler):
         if logger.pbar is None:
             logger._reset(pbar_total=pbar_total)
 
-        desc = self.tag
+        state_dict = engine.state_dict()
+        max_epochs = state_dict["max_epochs"] if "max_epochs" in state_dict else None
+        default_desc = "Iterations" if max_epochs and max_epochs == 1 else "Epoch"
+
+        desc = self.tag or default_desc
         max_num_of_closing_events = self.get_max_number_events(self.closing_event_name, engine)
         if max_num_of_closing_events > 1:
             global_step = engine.state.get_event_attrib_value(self.closing_event_name)
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -49,9 +49,8 @@ class _OutputHandler(BaseOutputHandler):
         if logger.pbar is None:
             logger._reset(pbar_total=pbar_total)
 
-        state_dict = engine.state_dict()
-        max_epochs = state_dict["max_epochs"] if "max_epochs" in state_dict else None
-        default_desc = "Iterations" if max_epochs and max_epochs == 1 else "Epoch"
+        max_epochs = engine.state.max_epochs
+        default_desc = "Iterations" if max_epochs == 1 else "Epoch"
 
         desc = self.tag or default_desc
         max_num_of_closing_events = self.get_max_number_events(self.closing_event_name, engine)
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -98,8 +98,8 @@ class ProgressBar(BaseLogger):
             formatting, see `tqdm docs <https://tqdm.github.io/docs/tqdm/>`_.
         **tqdm_kwargs: kwargs passed to tqdm progress bar.
             By default, progress bar description displays "Epoch [5/10]" where 5 is the current epoch and 10 is the
-            number of epochs; however, if max_epochs are set to 1, the progress bar instead displays
-            "Iteration [5/10]". If tqdm_kwargs defines `desc`, e.g. "Predictions", than the description is
+            number of epochs; however, if ``max_epochs`` are set to 1, the progress bar instead displays
+            "Iteration: [5/10]". If tqdm_kwargs defines `desc`, e.g. "Predictions", than the description is
             "Predictions [5/10]" if number of epochs is more than one otherwise it is simply "Predictions".
 
     Examples:
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -50,7 +50,7 @@ class _OutputHandler(BaseOutputHandler):
             logger._reset(pbar_total=pbar_total)
 
         max_epochs = engine.state.max_epochs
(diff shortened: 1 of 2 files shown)


vfdev-5:
@M3L6H we have to set up desc when it is used with tqdm bar. Currently, we have [here](https://github.com/pytorch/ignite/blob/2778893db943e9869c2ce266c740c1310bc3b9de/ignite/contrib/handlers/tqdm_logger.py#L52)
```python 
desc = self.tag 
```
which is None if user does not set any desc option while creating pbar. We need to modify it to 
```python
if self.tag is not None:
    desc = self.tag
elif engine.state.max_epochs > 1:
    desc = "Epoch"
else:
    desc = "Iteration"
```
I let you update your code with that in a more compact way.

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -49,7 +49,11 @@ def __call__(self, engine, logger, event_name):
         if logger.pbar is None:
             logger._reset(pbar_total=pbar_total)
 
-        desc = self.tag
+        state_dict = engine.state_dict()
+        max_epochs = state_dict["max_epochs"] if "max_epochs" in state_dict else None
```
vfdev-5, line 53: Can you use `engine.state.max_epochs` instead of `engine.state_dict()`. Here, `engine.state.max_epochs` can not be `None`.

M3L6H:
@vfdev-5 This is the code I currently have implemented. It's slightly different from what you commented, because I did not see your comment until later, but it seems to be functioning fairly well

```python
state_dict = engine.state_dict()
max_epochs = state_dict["max_epochs"] if "max_epochs" in state_dict else None
default_desc = "Iterations" if max_epochs and max_epochs == 1 else "Epoch"

desc = self.tag or default_desc
```

I had to modify a few tests which set `max_epochs` to 1. However one test is failing, and I'm not sure if it is the test's fault or my code's fault. The test in question is this one

```python
capsys = <_pytest.capture.CaptureFixture object at 0x7f530cc55668>

    def test_tqdm_logger_epoch_length(capsys):
        loader = list(range(100))
        engine = Engine(update_fn)
        pbar = ProgressBar(persist=True)
        pbar.attach(engine)
        engine.run(loader, epoch_length=50)
    
        captured = capsys.readouterr()
        err = captured.err.split("\r")
        err = list(map(lambda x: x.strip(), err))
        err = list(filter(None, err))
        actual = err[-1]
        expected = "Epoch: [50/50] 100%|██████████ [00:00<00:00]"
>       assert actual == expected
E       AssertionError: assert 'Iterations: [50/50] 100%|██████████ [00:00<00:00]' == 'Epoch: [50/50] 100%|██████████ [00:00<00:00]'
E         - Epoch: [50/50] 100%|██████████ [00:00<00:00]
E         ? ^^ ^^
E         + Iterations: [50/50] 100%|██████████ [00:00<00:00]
E         ? ^^^^^^^ ^^

tests/ignite/contrib/handlers/test_tqdm_logger.py:454: AssertionError
```

My problem is that I don't see anywhere in this test where `max_epochs` is set to 1, yet my code seems to think it is (since "iterations" got printed). My question is whether there is some default behavior setting `max_epochs` to 1 when it isn't explicitly passed, or what am I missing?

ignite/contrib/handlers/tqdm_logger.py
M3L6H: Ok

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -49,7 +49,11 @@ def __call__(self, engine, logger, event_name):
...
+        default_desc = "Iterations" if max_epochs and max_epochs == 1 else "Epoch"
```
vfdev-5, line 54: `default_desc` this should be "Iteration" and not "Iteration**s**"

vfdev-5:
When you call `engine.run(loader, epoch_length=50)` then `max_epochs` is set by default as 1, see the API here : https://pytorch.org/ignite/engine.html#ignite.engine.engine.Engine.run

M3L6H:
Sounds good. I will update the test and push again

vfdev-5:
Could you also update the docstring of ProgressBar. This part : 
```
By default, progress bar description displays "Epoch [5/10]" where 5 is the current epoch and 10 is the
number of epochs. If tqdm_kwargs defines `desc`, e.g. "Predictions", than the description is number of epochs.
```
should be adapted and say when we display "Iteration" instead of "Epoch". Thanks !

M3L6H:
I have updated the docstring as requested.

vfdev-5 (approved):
Thanks @M3L6H, LGTM !

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -250,7 +254,7 @@ def attach(
         Note: accepted output value types are numbers, 0d and 1d torch tensors and strings
 
         """
-        desc = self.tqdm_kwargs.get("desc", "Epoch")
+        desc = self.tqdm_kwargs.get("desc", None)
```
g-karthik, line 257: Would `_OutputHandler` handle the case where `desc=None`?

g-karthik (commented):
Looks good to me, just have a small question about the `None` case!

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -49,7 +49,10 @@ def __call__(self, engine, logger, event_name):
         if logger.pbar is None:
             logger._reset(pbar_total=pbar_total)
 
-        desc = self.tag
+        max_epochs = engine.state.max_epochs
+        default_desc = "Iteration" if max_epochs == 1 else "Epoch"
+
+        desc = self.tag or default_desc
```
vfdev-5, line 55: @g-karthik None should be handled here. Are we missed something ?

g-karthik: @vfdev-5 oh gotcha, the `or` condition should handle it, makes sense!

g-karthik (commented):
Looks good! Thanks @M3L6H and @vfdev-5 !

PR #1280, opened 2020-09-11
metrics_impl fix 2 gpu hvd tests and ensure consistent detaching

n2cholas:
Description:
- Fixes 2 gpu hvd tests 
- ensures consistent output detaching for all metrics

Check list:


vfdev-5:
@n2cholas looks good, thanks ! 

Actually, what was the issue with hvd ?

vfdev-5 (approved):

n2cholas:
> Actually, what was the issue with hvd ?

It was similar to the horovod issue we had before. In the `test_distrib_hvd` functions for all the metrics tests, we have the following line: `device = torch.device("cpu" if not torch.cuda.is_available() else "cuda")` which becomes `torch.device("cuda")` for the 2 gpu tests. Then, in some of the actual tests, we check to make sure that the accumulating tensors are on the right device by checking if `device == metric.accumulater.device`. However, the accumulator's device would be `cuda:0` or `cuda:1` which would result in a mismatch and fail. To fix that, I replace the check with `idist.device()  == metric.accumulater.device` instead.

Issue #1278, opened 2020-09-10
Setup mypy type checking in CI


vfdev-5:
## 🚀 Feature

Idea is to setup type checking of the codebase with mypy with our Travis CI ([here](https://github.com/pytorch/ignite/blob/master/.travis.yml#L55)) :
```bash
mypy --config-file mypy.ini
```

- At the first iteration we can ignore most of the errors with `mypy.ini`. Example of [mypy.ini](https://github.com/pytorch/vision/blob/master/mypy.ini). => done #1296 

- For other iterations, let's set `ignore_errors = False` for a module (e.g. `[mypy-ignite.utils.*]`) (see [mypy.ini](https://github.com/pytorch/ignite/blob/master/mypy.ini)) and fix mypy errors.

For Hacktoberfest contributors, feel free to ask questions for details if any and say that you would like to tackle the issue.
Please, take a look at [CONTRIBUTING guide](https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md).

kenjihiraoka:
Hi guys, I'm interested to do this task! :smile:

vfdev-5:
Hi @kenjihiraoka thanks for your interest on working on this issue :)

I can assign it to you and you can start working on that. As the issue may be large, feel free to submit several PRs.


PS: Notes for Hacktoferfest

> September is Preptember – a full month for maintainers to groom your repositories and for contributors to learn about making quality pull requests.

> Contributions during September don't count toward Hacktoberfest. Only pull requests submitted between October 1st to 31st will count.

kenjihiraoka:
@vfdev-5 so the ideia is to add the mypy check inside of `Lint check stage`, right?
And by now, can I skip all errors?

vfdev-5:
> so the ideia is to add the mypy check inside of Lint check stage, right?

@kenjihiraoka that's right.

> And by now, can I skip all errors?

Let's see which files are directly passing and ignore for others.

sdesrozis:
@kenjihiraoka any updates on this issue from your side ?

sdesrozis:
@kenjihiraoka Can I close this issue ?

vfdev-5:
Currently, mypy ignores all errors for all modules. We have to rework our typing such that mypy checks the code.
We can close this issue and create another one to make mypy pass on some of ignite's core modules like utils etc

PR #1296, opened 2020-09-16
Setup Mypy check at CI step

kenjihiraoka:
In response to #1278 

Description: Added mypy check at CI deployment and skip the current errors.

Check list:


vfdev-5 (approved):
Looks good ! Thanks @kenjihiraoka !

vfdev-5:
@kenjihiraoka so currently we ignore all the files right ?

kenjihiraoka:
@vfdev-5 yes exactly, I setup the mypy.ini with each folder inside of ignite root folder, then we have flexibility to work the mypy error step by step

PR #1302, opened 2020-09-17
Add FID metric

### LINKED ISSUE #998: Metrics for GANs
## 🚀 Feature

It would be interesting to provide metrics for generated image evaluation like here: 
- https://github.com/clovaai/stargan-v2/blob/master/metrics/
- https://github.com/Tetratrio/stylegan2_pytorch/tree/master/stylegan2/metrics

kamahori:
Fixes #998

Description:
Added FID metric which is used for GANs.
Most codes are incorporated from [here](https://github.com/clovaai/stargan-v2/blob/master/metrics/fid.py), so further optimization might be needed.

Check list:


vfdev-5:
@kamahori thanks for working on this metric ! I agree that it is not a simplest metric the intergrate.
I have a few comments about the API I was thinking about. Idea is that user can attach the metric to evaluator engine:

```python

fid_metric = FID()
evaluator = ...
fid_metric.attach(evaluator, "fid")

evaluator.run(validation_dataloader)
```
See here for more details: https://pytorch.org/ignite/concepts.html

Such that we can avoid dataloader setup from path etc. ImageNet pretrained model like Inception can an optional argument:
```python
class FID(Metric):
    def __init__(self, ..., test_model=None):
        ...
        if test_model is None:
            try:
                 from torchvision ... import inception
                 test_model = ..
            except ImportError:
                 ...
``` 
Maybe something else... I'll comment later.

Other remarks, probable, we should not use numpy or tqdm in the code. We can replace numpy by torch and tqdm can be attached by user on the evaluator.

kamahori:
Thanks for reviewing @vfdev-5 
I fixed several points.

> Other remarks, probable, we should not use numpy or tqdm in the code. We can replace numpy by torch and tqdm can be attached by user on the evaluator.

How about scipy? Replacing `scipy.linalg.sqrtm` with torch seems a bit complicated.

vfdev-5:
> How about scipy? Replacing scipy.linalg.sqrtm with torch seems a bit complicated.

Good point, let me check if it can be possible or not at all...

kamahori:
Maybe we can use [this](https://github.com/steveli/pytorch-sqrtm) ?

vfdev-5:
Thanks for the link. Seems like it is a wrap over scipy and backprop enabled. Backprop is not a requirement for us. Can you check if there is a FR on pytorch for that ? I wonder if it is difficult to implement ...

kamahori:
I also found [this](https://github.com/pytorch/pytorch/issues/25481) and [this](https://github.com/pytorch/pytorch/issues/9983), but it seems they are not exactly what we are looking for...
And FYI [here](https://github.com/scipy/scipy/blob/v1.5.2/scipy/linalg/_matfuncs_sqrtm.py#L117-L194) is the original scipy implementation.

ignite/metrics/fid.py
```diff
@@ -0,0 +1,131 @@
...
+from scipy import linalg
+from torchvision import transforms
+
+from ignite.exceptions import NotComputableError
+from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
+
+
+__all__ = ["FID"]
+
+class DefaultDataset(data.Dataset):
```
vfdev-5, line 14: I think we should not deal with datasets, but have rely on batches provided by the evaluator.

ignite/metrics/fid.py
```diff
@@ -0,0 +1,131 @@
...
+
+class FID(Metric):
+    """Calculates FID metric
+
+    """
+
+    def __init__(
+        self, 
+        output_transform: Callable = lambda x: x, 
+        input_path: os.path,
```
vfdev-5, line 39: we need not to have input_path and output_path as argument

ignite/metrics/fid.py
```diff
@@ -0,0 +1,131 @@
...
+        output_path: os.path,
+        test_model = None,
+    ):
+        self._input_path = input_path
+        self._output_path = output_path
+
+        if test_model is None:
+            try:
+                from torchvision import models
+                test_model = models.inception_v3()
```
vfdev-5, line 49: Should not the model be pretrained=True ?

ignite/metrics/fid.py
```diff
@@ -0,0 +1,131 @@
...
+            cov.append(_cov(actvs, rowvar = False))
+        fid_value = self._frechet_distance(mu[0], cov[0], mu[1], cov[1])
+        return fid_value
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._value = None
+
+    @reinit__is_reduced
+    def update(self, output) -> None:
```
vfdev-5, line 125: We have to implement the metric such that it is computed in online manner:
- on update we accumulate some internal things (e.g. activations here)
- on compute we calculate the final metric using internal variables. (e.g compute frechet distance on mu and cov)

Let's define that the output is `y_pred` = batch of generated images = `generator(batch_x)` whatever the x is (scalar/vector for dcgans, image for cyclegans etc).
During the udpate we should think what to accumulate: activations average and covariance (not sure yet)

ignite/metrics/fid.py
```diff
@@ -0,0 +1,131 @@
...
+    """Calculates FID metric
+
+    """
+
+    def __init__(
+        self, 
+        output_transform: Callable = lambda x: x, 
+        input_path: os.path, 
+        output_path: os.path,
+        test_model = None,
```
vfdev-5, line 41: I like the implementation of fid here : https://github.com/Tetratrio/stylegan2_pytorch/blob/master/stylegan2/metrics/fid.py
and I think it would be better to call test_model as fid_model ?

vfdev-5 (commented):
@kamahori thanks for working on this metric !
I left some comments on how we need to reimplement the metric...
Honestly, it is not a simple metric to implement and hope we can find some nice solutions :)

vfdev-5:
@kamahori if there is no "simple" option to make it without scipy, let's put the metric into `ignite.contrib.metrics` and use scipy. In this module, there are other metrics using other deps like sklearn etc

kamahori:
@vfdev-5 
Thank you for your comments. I fixed codes again.
I moved FID metric to `ignite.contrib.metrics` because I could not find any simple solution to scipy's problem.

vfdev-5:
@kamahori thanks for the update ! It looks better now 👍  However, it remains a tough part about "online" fashion metric  computation, make it work with DDP and write tests

Let's handle them one by one.
1) Online fashion metric computation

Let me give you an example how it would work for a simple metric like accuracy. At first, how can we compute accuracy in general (without online) on predictions and targets:
```python

y_pred = ...   # tensor (N, K)  - predictions, N samples of K features
y = ...   # tensor (N, )  - targets, N samples, e.g. [0, 2, 1, 4, 5, 2, 3, 4] where K=6

pred_indices = torch.argmax(y_pred, dim=1)
num_correct = torch.eq(pred_indices, y).view(-1)  # compute the number of correctly predicted samples
num_samples = y.shape[0]

accuracy = num_correct / num_samples
```

In online manner, we have a bunch of pairs like `(y_pred, y)` and we would like to compute final accuracy. Naive way is to contatenate all `y_pred` into a single `total_y_pred` and same for `total_y` and then compute `num_correct`, `num_samples`. But this approach is not memory optimal as we have to store entire history of `(y_pred, y)`.
However, we can see that accuracy as metric can be computed as 
```
num_correct = num_correct_pair1 + num_correct_pair2 + ...num_correct = num_correct_pair1 + num_correct_pair2 + ...
accuracy = num_correct / num_samples
```

kamahori:
Thanks. How about this implementation?
Accumulate the features in `update` and calculate `mu`, `sigma`, FID in `compute`.

vfdev-5 (commented):
Thanks for the update ! I have some other comments.
I think it would be also good to write some basic tests.

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,97 @@
...
+                fid_model = models.inception_v3(pretrained=True)
+            except ImportError:
+                raise ValueError("Argument fid_model should be set")
+
+        self._device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        self._fid_model = fid_model.eval().to(self._devide)
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._value = None
```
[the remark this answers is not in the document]
kamahori: Yes, they were not used.

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self._features_real = []
+        self._features_fake = []
+
+    @reinit__is_reduced
+    def update(self, output) -> None:
+        y_pred, y = output[0].detach(), output[1].detach()
+
+        data_real, data_fake = y, y_pred
+        if isinstance(y_pred, (tuple, list)):
```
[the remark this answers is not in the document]
kamahori: Do we not need this conditioning? (line 52 - 54)

kamahori:
I implemented `update` to calculate covariance in an online manner (I think this would be fine) and added a FID test.

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,95 @@
...
+__all__ = ["FID"]
+
+
+class FID(Metric):
+    """Calculates FID metric
+
+    """
+
+    def __init__(
+        self, output_transform: Callable = lambda x: x, fid_model=None,
```
vfdev-5, line 25:
```suggestion
        self, output_transform: Callable = lambda x: x, fid_model: nn.Module = None,
```

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,95 @@
+import os
+from typing import Callable, Sequence, Union
+
+import torch
```
vfdev-5, line 4:
```suggestion
import torch
import torch.nn as nn
```

ignite/contrib/metrics/fid.py
vfdev-5: Recently, we reintroduced `device` argument to store internal vars as tensors: https://github.com/pytorch/ignite/blob/master/ignite/metrics/metric.py#L131, we have to add it here too. Please, take a look at how other metrics are implemetned, e.g. https://github.com/pytorch/ignite/blob/master/ignite/metrics/accuracy.py#L96

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,95 @@
...
+                raise ValueError("Argument fid_model should be set")
+
+        self._device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        self._fid_model = fid_model.eval().to(self._devide)
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._num_of_data = 0
+
+        self._features_sum_real = None
```
vfdev-5, line 42: Here, IMO, we can set all those things as `torch.tensor(0, device=self.device, dtype=torch.float32)` and we can simply use `+= batch_features_real` etc ...

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,95 @@
...
+    @reinit__is_reduced
+    def update(self, output) -> None:
+        y_pred, y = output[0].detach(), output[1].detach()
+
+        data_real, data_fake = y, y_pred
+        with torch.no_grad():
+            batch_features_real = self._fid_model(data_real)
+            batch_features_fake = self._fid_model(data_fake)
+
+        batch_features_real = batch_features_real.view(*batch_features_real.size()[:2], -1).mean(-1)
```
vfdev-5, line 57: shouldn't it be a sum instead of mean ?

vfdev-5 (commented):
@kamahori thanks ! This looks much better now ! Now we have to write some meaningful tests...

tests/ignite/contrib/metrics/test_fid.py
```diff
@@ -0,0 +1,37 @@
...
+
+from ignite.contrib.metrics import FID
+
+try:
+    from scipy import linalg
+except ImportError:
+    raise RuntimeError("This contrib test requires scipy to be installed.")
+
+
+def test_fid():
```
vfdev-5, line 12: @kamahori about meaningful tests, I'd like to do the following:
- let's use a real image from internet (<300px) and add it into `tests/ignite/contrib/metrics/assets`
- manually compute FID score for this image and its slightly noised version. Otherwise use one of those referenced implementation. Let's use computed score as true score.
- compute FID score with ignite's implementation and compare scores.

kamahori:
I fixed the test to use an image from ImageNet (`tests/ignite/contrib/metrics/assets/fid_sample.jpg`).

tests/ignite/contrib/metrics/test_fid.py
```diff
@@ -0,0 +1,46 @@
...
+
+    size = 10
+
+    y = torchvision.datasets.ImageFolder(
```
vfdev-5, line 19: I think we can read the image with Pillow and use torchvision to transform it to tensor without using `datasets.ImageFolder`.
```python
from pathlib import Path

import PIL
from torchvision.transforms.functional import to_tensor


@pytest.fixture()
def img_filepath():
    fp = Path(__file__).parent / "assets" / "fid_sample.jpg"
    assert fp.exists()
    yield fp.as_posix()


def test_fid(img_filepath):
    pil_img = PIL.Image.open(img_filepath)
    y = to_tensor(pil_img)
    y_pred = y + torch.rand_like(y) * 0.01

```

kamahori:
I fixed the implementation, but still have some errors in CI. Do you know what is the problem? @vfdev-5 
- <https://github.com/pytorch/ignite/pull/1302/checks?check_run_id=1182940285#step:5:899>
- <https://github.com/pytorch/ignite/pull/1302/checks?check_run_id=1182940300#step:5:16>

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,89 @@
...
+                self._num_of_data / (self._num_of_data + 1) ** 2
+            ) + self._cov_real * self._num_of_data / (self._num_of_data + 1)
+            self._cov_fake = torch.ger(X_fake, X_fake) * (
+                self._num_of_data / (self._num_of_data + 1) ** 2
+            ) + self._cov_fake * self._num_of_data / (self._num_of_data + 1)
+        else:
+            self._cov_real = torch.eye(len(y))
+            self._cov_fake = torch.eye(len(y_pred))
+
+    @sync_all_reduce
```
vfdev-5, line 76: @kamahori current fail is due to bad usage of `@sync_all_reduce` which should have attributes to reduce. See https://pytorch.org/ignite/metrics.html#metrics-and-distributed-computations

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,89 @@
...
+__all__ = ["FID"]
+
+
+class FID(Metric):
+    """Calculates FID metric
+
+    """
+
+    def __init__(
+        self, output_transform: Callable = lambda x: x, fid_model: nn.Module = None,
```
vfdev-5, line 26: Constructor should have device argument. See Accuracy as example: https://github.com/pytorch/ignite/blob/master/ignite/metrics/accuracy.py#L135

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,89 @@
...
+    ):
+        if fid_model is None:
+            try:
+                from torchvision import models
+
+                fid_model = models.inception_v3(pretrained=True, transform_input=True)
+            except ImportError:
+                raise ValueError("Argument fid_model should be set")
+
+        self._device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
vfdev-5, line 36: Device should be set with `super(FID, self).__init__(output_transform=output_transform, device=device)`

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,89 @@
...
+        self._fid_model = fid_model.eval().to(self._device)
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._num_of_data = 0
+
+        self._features_sum_real = torch.tensor(0, device=self._device, dtype=torch.float32)
+        self._features_sum_fake = torch.tensor(0, device=self._device, dtype=torch.float32)
+
+        self._cov_real = torch.tensor(0, device=self._device, dtype=torch.float32)
```
vfdev-5, line 46: Seems like initialization is not used as intended. Let's set them to None:
```python
self._cov_real = None
self._cov_fake = None
```

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,89 @@
...
+        with torch.no_grad():
+            batch_features_real = self._fid_model(y).sum(axis=0)
+            batch_features_fake = self._fid_model(y_pred).sum(axis=0)
+
+        self._num_of_data += 1
+
+        self._features_sum_real = self._features_sum_real + batch_features_real
+        self._features_sum_fake = self._features_sum_fake + batch_features_fake
+
+        if self._num_of_data > 1:
```
vfdev-5, line 62: Let's replace it like
```python
if self._cov_real is None:
    self._cov_real = torch.eye(len(y))
else:
     self._cov_real = self._compute_cov(batch_features_real, self._features_sum_real)
```
where we can refactor cov computation into `self._compute_cov` static function. Same for fake.

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,89 @@
...
+            self._cov_fake = torch.ger(X_fake, X_fake) * (
+                self._num_of_data / (self._num_of_data + 1) ** 2
+            ) + self._cov_fake * self._num_of_data / (self._num_of_data + 1)
+        else:
+            self._cov_real = torch.eye(len(y))
+            self._cov_fake = torch.eye(len(y_pred))
+
+    @sync_all_reduce
+    def compute(self) -> Union[torch.Tensor, float]:
+        mu_real = self._features_sum_real / self._num_of_data
```
vfdev-5, line 78: What are the shapes of `self._features_sum_real`, `self._cov_real`, `self._features_sum_fake` and `self._cov_fake` ?

ignite/contrib/metrics/fid.py
```diff
@@ -0,0 +1,89 @@
...
+        sigma_real = self._cov_real
+
+        mu_fake = self._features_sum_real / self._num_of_data
```
vfdev-5, line 81: Typo with `_features_sum_real` instead of `_features_sum_fake`

vfdev-5 (commented):
@kamahori thanks for the updates. I left several comments.

ignite/contrib/metrics/fid.py
kamahori: When inception v3 model is used, `features_sum` is 1000 dimension vector and `cov` is 1000 * 1000 matrix.

kamahori:
I found that calculating covariance in an online manner is not efficient. So I reverted `_cov` function and implemented to compute covariance only once.

kamahori:
I think most parts are completed.
Do you know why these errors are happening?:

- <https://github.com/pytorch/ignite/pull/1302/checks?check_run_id=1197935105#step:5:16>
- <https://github.com/pytorch/ignite/pull/1302/checks?check_run_id=1197935271#step:5:2463>

And is there any better way to extract real values to avoid [this](https://github.com/pytorch/ignite/pull/1302/checks?check_run_id=1197935107#step:5:968)?
@vfdev-5

ignite/contrib/metrics/fid.py
vfdev-5: @kamahori I think this is not correct as 1000 in our case corresponds to 1000 ImageNet classes and normally it should be Inception feature before classification as here: https://github.com/clovaai/stargan-v2/blob/master/metrics/fid.py#L53

vfdev-5:
@kamahori thanks for working on this PR which is not simple !

About XLA failing tests, let's see this a bit later as it looks like a bit unrelated to the PR.

First, I think we are incorrect about inception features. This is something to rework. Maybe, the simplest way is to replace final classification layer by identity, such that we do not need to do a lot of network surgery.

Second, I'm not that convinced about computing covariance like you changed. Now, we accumulate the sum and whole history of features... Let me also check it by myself (however, I'm a bit of lacking of time this weeks).

import-antigravity:
Since no one appears to have linked it, I believe this is considered to be the "reference implementation" for pytorch: https://github.com/mseitzer/pytorch-fid

sdesrozis:
@import-antigravity thank you for the link !

vfdev-5:
closed in favor of #2049

Issue #1305, opened 2020-09-20
Automatic docker build and publishing


vfdev-5:
## 🚀 Feature

Idea is to rebuild and publish [pytorch ignite images](https://github.com/pytorch/ignite/tree/master/docker) once a new tag/release is published or a push to the master inside docker folder.

How to publish docker image: 
- https://docs.github.com/en/actions/guides/publishing-docker-images#publishing-images-to-docker-hub-and-github-packages

TODO for maintainers:
- create secrets for username/tokens
- ensure that we can build apex, horovod etc on CPU-only machines

vfdev-5:
Current state of the issue:
- we provide scripts to build and push images: 
  - https://github.com/pytorch/ignite/blob/master/docker/hvd/build_all.sh
  - https://github.com/pytorch/ignite/blob/master/docker/main/build_all.sh
  - https://github.com/pytorch/ignite/blob/master/docker/msdp/build_all.sh
  - https://github.com/pytorch/ignite/blob/master/docker/push_all.sh

What remains to do:
- ensure that we can build apex, horovod etc on CPU-only machines
- create a github action which is triggered on "new tag/release is published" or "push to the master inside docker folder"
this action should call 3 build_all.sh in parallel and finally call push_all.sh

trsvchn:
@vfdev-5 I think I can help with this part:

> create a github action which is triggered on "new tag/release is published" or "push to the master inside docker folder"

But

> this action should call 3 build_all.sh in parallel and finally call push_all.sh

I'm not sure that it's possible with GH Actions: [Steps in parallel?](https://github.community/t/steps-in-parallel/16343): 

>  Not at this moment. Only jobs can run in parallel, but steps always run sequentially. I don’t know if there are any plans to ever support this.

PR #1306, opened 2020-09-21
Updated hvd images + scripts

vfdev-5:
Related to #1305 

Description:
- added scripts to auto build and push images

Github actions yaml in another PR. 

Check list:

### DIFF
--- a/docker/msdp/build_all.sh
+++ b/docker/msdp/build_all.sh
@@ -20,12 +20,12 @@ docker build -t pytorchignite/${image_name}:latest -f Dockerfile.${image_name} .
 image_tag=`docker run --rm -it pytorchignite/${image_name}:latest -c 'python -c "import torch; import ignite; print(torch.__version__ + \"-\" + ignite.__version__, end=\"\")"'`
 docker tag pytorchignite/${image_name}:latest pytorchignite/${image_name}:${image_tag}
 
-image_name="msdp-apex-vision"
-docker build -t pytorchignite/${image_name}:latest -f Dockerfile.${image_name} .
-docker tag pytorchignite/${image_name}:latest pytorchignite/${image_name}:${image_tag}
+for image_name in "msdp-apex-vision" "msdp-apex-nlp"
+do
 
-image_name="msdp-apex-nlp"
-docker build -t pytorchignite/${image_name}:latest -f Dockerfile.${image_name} .
-docker tag pytorchignite/${image_name}:latest pytorchignite/${image_name}:${image_tag}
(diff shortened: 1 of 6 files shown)


vfdev-5:
@fco-dev please review

docker/push_all.sh
```diff
@@ -0,0 +1,55 @@
...
+image_name="msdp-apex-base"
+image_tag=`docker run --rm -it pytorchignite/${image_name}:latest -c 'python -c "import torch; import ignite; print(torch.__version__ + \"-\" + ignite.__version__, end=\"\")"'`
+
+for image_name in "msdp-apex-base" "msdp-apex-vision" "msdp-apex-nlp"
+do
+
+    docker push pytorchignite/${image_name}:latest
+    docker push pytorchignite/${image_name}:${image_tag}
+
+done
```
fco-dv, line 55: Once pushed, the local repository would maybe need a `clean` step. It also seems that multi-stage builds generate, `<none>` images:
```bash
docker images | grep 'pytorchignite\|<none>' | awk '{print $3}' | xargs docker rmi -f
```
what do you think ?

docker/msdp/build_all.sh
```diff
@@ -0,0 +1,31 @@
...
+curr_dir=$PWD
+cd $curr_dir/msdp
+
+set -xeu
+
+image_name="msdp-apex-base"
+
+docker build -t pytorchignite/${image_name}:latest -f Dockerfile.${image_name} .
+image_tag=`docker run --rm -it pytorchignite/${image_name}:latest -c 'python -c "import torch; import ignite; print(torch.__version__ + \"-\" + ignite.__version__, end=\"\")"'`
+docker tag pytorchignite/${image_name}:latest pytorchignite/${image_name}:${image_tag}
```
fco-dv, line 21: Maybe put `image_tag` in a for loop like for `main` and `hvd` scripts.

fco-dv (commented):
@vfdev-5 great job! is there a need to separate the build/tag and push scripts ?

docker/push_all.sh
vfdev-5: Yes, this is important for a local system, on the CI it is less important I think.

docker/msdp/build_all.sh
vfdev-5: Yes, it was the first script I wrote. As there are 3 cases I though loop will not refactor a lot... 
OK, I can change that.

vfdev-5 (commented):
Well, the reason for splitting build and push is that we may potentially would like perform locally only the build step.

Thanks for the review @fco-dv

PR #1320, opened 2020-09-24
Improve ManhattanDistance metric for DDP 

### LINKED ISSUE #1319: Improve ManhattanDistance metric for DDP
## 🚀 Feature

Current formula of `ManhattanDistance` is not correct, it misses an absolute value. The idea is to use `sklearn` to ensure good computation. 

Current implementation of `ManhattanDistance` metric is not compatible with ddp. The idea is to improve this metric to work in parallel.

Related to #1284

sdesrozis:
Fixes #1319  

Description:

Fix ManhattanDistance formula and compare to sklearn.

Improvement of ManhattanDistance metric to work with ddp.

Check list:

### DIFF
--- a/ignite/contrib/metrics/regression/manhattan_distance.py
+++ b/ignite/contrib/metrics/regression/manhattan_distance.py
@@ -1,13 +1,16 @@
+from typing import Callable, Union
+
 import torch
 
 from ignite.contrib.metrics.regression._base import _BaseRegression
+from ignite.metrics.metric import reinit__is_reduced, sync_all_reduce
 
 
 class ManhattanDistance(_BaseRegression):
     r"""
     Calculates the Manhattan Distance:
 
-    :math:`\text{MD} = \sum_{j=1}^n (A_j - P_j)`,
+    :math:`\text{MD} = \sum_{j=1}^n |A_j - P_j|`,
 
     where :math:`A_j` is the ground truth and :math:`P_j` is the predicted value.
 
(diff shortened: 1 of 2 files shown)


ignite/contrib/metrics/regression/manhattan_distance.py
```diff
@@ -1,13 +1,16 @@
...
 from ignite.contrib.metrics.regression._base import _BaseRegression
+from ignite.metrics.metric import reinit__is_reduced, sync_all_reduce
 
 
 class ManhattanDistance(_BaseRegression):
     r"""
     Calculates the Manhattan Distance:
 
-    :math:`\text{MD} = \sum_{j=1}^n (A_j - P_j)`,
+    :math:`\text{MD} = \sum_{j=1}^n |A_j - P_j|`,
```
vfdev-5, line 13: I agree that Manhattan distance is computed with abs of error. However, in the paper it is clearly stated without abs. Let's use abs and replace the article citation with sklearn.

ignite/contrib/metrics/regression/manhattan_distance.py
```diff
@@ -1,32 +1,43 @@
...
 
     where :math:`A_j` is the ground truth and :math:`P_j` is the predicted value.
 
-    More details can be found in `Botchkarev 2018`__.
+    More details can be found in `sklearn`__.
```
vfdev-5, line 17:
```suggestion
    More details can be found in `scikit-learn distance metrics`__.
```

vfdev-5 (approved):
Thanks @sdesrozis !

PR #1328, opened 2020-09-26
Make SSIM accumulate on specified device

### LINKED ISSUE #1309: Update SSIM metric
## 🚀 Feature

Current implementation of [SSIM](https://github.com/pytorch/ignite/blob/master/ignite/metrics/ssim.py) does not use torch tensor as internal storage and hasn't device in the constructor. Following #1238 we expect all metrics to have a similar structure.

@n2cholas , could you please work on that ?

n2cholas:
Fixes #1309

Check list:

### DIFF
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -91,9 +91,8 @@ class SSIM(Metric):
 
     def _uniform(self, kernel_size):
         max, min = 2.5, -2.5
-        kernel = torch.arange(
-            start=(1 - kernel_size) / 2, end=(1 + kernel_size) / 2, step=1, dtype=torch.float32, device=self._device
-        )
+        ksize_half = (kernel_size - 1) * 0.5
+        kernel = torch.linspace(-ksize_half, ksize_half, steps=kernel_size, device=self._device)
         for i, j in enumerate(kernel):
             if min <= j <= max:
                 kernel[i] = 1 / (max - min)
@@ -103,10 +102,9 @@ class SSIM(Metric):
         return kernel.unsqueeze(dim=0)  # (1, kernel_size)
 
     def _gaussian(self, kernel_size, sigma):
-        kernel = torch.arange(
-            start=(1 - kernel_size) / 2, end=(1 + kernel_size) / 2, step=1, dtype=torch.float32, device=self._device
-        )
-        gauss = torch.exp(-kernel.pow(2) / (2 * pow(sigma, 2)))
+        ksize_half = (kernel_size - 1) * 0.5
+        kernel = torch.linspace(-ksize_half, ksize_half, steps=kernel_size, device=self._device)
+        gauss = torch.exp(-0.5 * (kernel / sigma).pow(2))
         return (gauss / gauss.sum()).unsqueeze(dim=0)  # (1, kernel_size)
 
     def _gaussian_or_uniform_kernel(self, kernel_size, sigma):
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -53,6 +53,7 @@ class SSIM(Metric):
         k2: float = 0.03,
         gaussian: bool = True,
         output_transform: Callable = lambda x: x,
+        device: Union[str, torch.device] = torch.device("cpu"),
     ):
         if isinstance(kernel_size, int):
             self.kernel_size = [kernel_size, kernel_size]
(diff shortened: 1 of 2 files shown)


ignite/metrics/ssim.py
```diff
@@ -74,23 +75,25 @@ def __init__(
...
         self.c2 = (k2 * data_range) ** 2
         self.pad_h = (self.kernel_size[0] - 1) // 2
         self.pad_w = (self.kernel_size[1] - 1) // 2
         self._kernel = self._gaussian_or_uniform_kernel(kernel_size=self.kernel_size, sigma=self.sigma)
-        super(SSIM, self).__init__(output_transform=output_transform)
 
     @reinit__is_reduced
     def reset(self) -> None:
-        self._sum_of_batchwise_ssim = 0.0
+        self._sum_of_batchwise_ssim = 0.0  # Not a tensor because batch size is not known in advance.
```
vfdev-5, line 88: Can't we use `torch.tensor(0, device=self._device)` ?

ignite/metrics/ssim.py
```diff
@@ -116,7 +121,7 @@ def _gaussian_or_uniform_kernel(self, kernel_size, sigma):
 
     @reinit__is_reduced
     def update(self, output: Sequence[torch.Tensor]) -> None:
-        y_pred, y = output[0].detach(), output[1].detach()
+        y_pred, y = output[0].detach().to(self._device), output[1].detach().to(self._device)
```
vfdev-5, line 124: I think we should keep user's device until we update `self._sum_of_batchwise_ssim`.

ignite/metrics/ssim.py
```diff
@@ -74,23 +75,25 @@ def __init__(
...
     def reset(self) -> None:
-        self._sum_of_batchwise_ssim = 0.0
+        self._sum_of_batchwise_ssim = 0.0  # Not a tensor because batch size is not known in advance.
         self._num_examples = 0
         self._kernel = self._gaussian_or_uniform_kernel(kernel_size=self.kernel_size, sigma=self.sigma)
 
     def _uniform(self, kernel_size):
         max, min = 2.5, -2.5
-        kernel = torch.arange(start=(1 - kernel_size) / 2, end=(1 + kernel_size) / 2, step=1, dtype=torch.float32)
+        kernel = torch.arange(
```
vfdev-5, line 94: Could you please improve the code here and for gaussian kernel:
- replace `torch.arange` by `torch.linspace`

ignite/metrics/ssim.py
```diff
@@ -100,7 +103,9 @@ def _uniform(self, kernel_size):
         return kernel.unsqueeze(dim=0)  # (1, kernel_size)
 
     def _gaussian(self, kernel_size, sigma):
-        kernel = torch.arange(start=(1 - kernel_size) / 2, end=(1 + kernel_size) / 2, step=1, dtype=torch.float32)
+        kernel = torch.arange(
```
vfdev-5, line 106: Here, i'd say:
```python
    ksize_half = (kernel_size - 1) * 0.5
    x = torch.linspace(-ksize_half, ksize_half, steps=kernel_size)
    pdf = torch.exp(-0.5 * (x / sigma).pow(2))
    kernel1d = pdf / pdf.sum()
```
may be a bit more efficient.

ignite/metrics/ssim.py
vfdev-5: Good question if we should put kernel on `self._device`. My reasoning is that, when we specifiy metric's device it is for storage and not for computing with that. So, i'd say let's kernel be of the same device as update's input. What do you think ?

vfdev-5 (commented):
Thank you for the PR @n2cholas !

ignite/metrics/ssim.py
n2cholas: No, since when we do `self._sum_of_batchwise_ssim += torch.mean(ssim_idx, (1, 2, 3), dtype=torch.float64)`, the shape of the `torch.mean(...)` is `(batch_size,)`, which we do not know in advance and cannot be added to `torch.tensor(0, device=self._device)`.

ignite/metrics/ssim.py
n2cholas: Sounds good, done!

ignite/metrics/ssim.py
n2cholas: Implemented the improved code!

I agree, the kernel should be on the same device as the update's input. I still think it makes sense to initially create the tensor on the device specified by the user, then move it during the first update step (if needed). It doesn't make a big difference either way, so I'm open to whichever one you prefer.

ignite/metrics/ssim.py
n2cholas: Done!

ignite/metrics/ssim.py
vfdev-5: Yes, you are right. Thanks !

ignite/metrics/ssim.py
```diff
@@ -74,23 +75,24 @@ def __init__(
...
-        super(SSIM, self).__init__(output_transform=output_transform)
 
     @reinit__is_reduced
     def reset(self) -> None:
-        self._sum_of_batchwise_ssim = 0.0
+        self._sum_of_batchwise_ssim = 0.0  # Not a tensor because batch size is not known in advance.
         self._num_examples = 0
         self._kernel = self._gaussian_or_uniform_kernel(kernel_size=self.kernel_size, sigma=self.sigma)
 
     def _uniform(self, kernel_size):
```
vfdev-5, line 92: This kernel's creation looks a bit weird. Maybe, we can rework it in another PR...

ignite/metrics/ssim.py
vfdev-5: OK, let's keep it like that.

vfdev-5 (approved):
LGTM !

ignite/metrics/ssim.py
n2cholas: Yes I agree, should be reworked in a separate PR.

PR #1333, opened 2020-09-28
Improve State and Engine docs #1259

### LINKED ISSUE #1259: Improve State and Engine docs
## 🚀 Feature

Currently, if we execute twice `trainer.run` we can fall into the following situation:
```python
# ...
trainer.run(train_loader, max_epochs=5)

# reset model weights etc and rerun
trainer.run(train_loader, max_epochs=2)
> ValueError: Argument max_epochs should be larger than the start epoch defined in the state: 2 vs 5
```

But if we would like to restart the training from the begining, we have to reset `trainer.state` as 
```python
trainer.run(train_loader, max_epochs=5)

# reset model weights etc and rerun
trainer.state.max_epochs = None
trainer.run(train_loader, max_epochs=2)
```

Let's improve two things:
1) let's provide `State.restart()` method that just sets `state.max_epochs = None`
2) Add a note into `Engine.run` docstring saying if we would like to restart the training from zero, we have to either call `trainer.state.restart()` or `trainer.state.max_epochs = None`.
3) Improve the message "ValueError: Argument max_epochs should be larger than the start epoch defined in the state: 2 vs 5" by adding something like "Please, call state.restart() before calling engine.run() if would like to restart from zero."

Please, take a look at [CONTRIBUTING guide](https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md).

fco-dv:
Fixes #1259 

Description:
- add State.restart() method
- add note in Engine.run() docstring / improve error message
- unit test for State.restart()

Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -639,6 +639,18 @@ class Engine(Serializable):
                 def switch_batch(engine):
                     engine.state.batch = preprocess_batch(engine.state.batch)
 
+            Restart the training from the beginning. User can reset `max_epochs = None` or either call
+            `trainer.state.restart()`:
+
+            .. code-block:: python
+
+                # ...
+                trainer.run(train_loader, max_epochs=5)
+
+                # Reset model weights etc. and restart the training
+                trainer.state.restart() # equivalent to trainer.state.max_epochs = None
+                trainer.run(train_loader, max_epochs=2)
+
         """
         if seed is not None:
             warnings.warn(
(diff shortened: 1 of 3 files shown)


vfdev-5 (approved):
Thanks @fco-dv !

sdesrozis (approved):
LGTM ! Ty @fco-dv !

vfdev-5:
Finally, I'm not sure about that as we discussed after that introducing `restart` method is error-prone...

PR #1335, opened 2020-10-02
warning if current device index is lower than current local rank

### LINKED ISSUE #1308: Show a warning if current device index is lower than current local rank
## 🚀 Feature

Following #1307, if user does not set `torch.cuda.set_device("cuda:lrank")`, ignite's code
https://github.com/pytorch/ignite/blob/0c417789599341d4b18a0b43ea35a9e7a7ea06b2/ignite/distributed/comp_models/native.py#L99-L102
will use the same device `cuda:0` for `all_reduce` op.

For older NCCL, it will setup itself such that i-th proc uses `cuda:0` device and thus following collective op will hang with other devices. For example
```python
import torch
import torch.distributed as dist

def main():

    # !!! We do not call torch.cuda.set_device("cuda:lrank")

    dist.init_process_group(backend="nccl", init_method="env://")
    import os
    local_rank = int(os.environ["LOCAL_RANK"])

    tensor = torch.tensor([local_rank + 1]).to("cuda")
    dist.all_reduce(tensor, op=dist.ReduceOp.MAX)
    print(tensor)

    tensor = torch.tensor([local_rank + 1]).to("cuda:{}".format(local_rank))
    # PROGRAM WILL HANG HERE >>>>
    dist.all_reduce(tensor, op=dist.ReduceOp.MAX)
    print(tensor)

    dist.destroy_process_group()

if __name__ == "__main__":
    main()
```

For newer NCCL, it raises the error as in #1307. 

Let's improve the code by raising a warning for native and horovod dist models when calling `idist.device()` if we encounter the situation where current cuda device index is smaller than the local rank.
PyTorch docs suggest that to use 1 proc per 1 cuda device => local rank should be equal to cuda device index.
However, it is also possible to have M procs with K devices / proc (e.g. 4 procs with 2 GPUs per proc) => local rank <= cuda device index.

HelioStrike:
Fixes #1308

Description:
- Display a warning if current device index is lower than local rank. (#1308 )

Check list:


ignite/distributed/comp_models/horovod.py
```diff
@@ -97,6 +98,8 @@ def get_node_rank(self) -> int:
         def device(self) -> torch.device:
             if torch.cuda.is_available():
                 index = torch.cuda.current_device()
+                if index < self.get_local_rank():
```
sdesrozis, line 101: @HelioStrike could you add a test for this condition ? Thank you !

ignite/distributed/comp_models/native.py
```diff
@@ -221,6 +221,8 @@ def get_node_rank(self) -> int:
         def device(self) -> torch.device:
             if self.backend() == dist.Backend.NCCL:
                 index = torch.cuda.current_device()
+                if index < self.get_local_rank():
```
sdesrozis, line 224: @HelioStrike could you add a test for this condition too ? Thank you !

sdesrozis:
@HelioStrike Here some comments.

Could you update the documentation with a note about this feature please ? Thank you again !!

vfdev-5:
@HelioStrike thanks for the PR. I'd say we have to add tests also for that...

HelioStrike:
@sdesrozis @vfdev-5 Sorry for asking such a silly question but, what should the tests be checking for, and what should it do? I had a look at `test_horovod.py` and `test_native.py`, but I'm still a little confused. :|

vfdev-5:
@HelioStrike it is good question (not that silly) !

I'd say we have to reproduce initial issue description and test that. Unfortunately, it is not explicitly said in the linked issues.
The problem is here : 
```python
import torch
import torch.distributed as dist

def main():

    # !!! We do not call torch.cuda.set_device("cuda:lrank")

    dist.init_process_group(backend="nccl", init_method="env://")
    import os
    local_rank = int(os.environ["LOCAL_RANK"])

    # if we call idist.get_world_size() the first time
    # it will try to setup Native dist comp model
    # and will call all reduce to get some parameter like number of nodes
    # effectively, it does below code:
    # tensor = torch.tensor([local_rank + 1]).to("cuda")
    # dist.all_reduce(tensor, op=dist.ReduceOp.MAX)
    # print(tensor)
      
    idist.get_world_size()

    dist.destroy_process_group()

if __name__ == "__main__":
    main()
```

So, the idea is to check the warning with `pytest.warns` that `idist.get_world_size()` raises a warning.
Let me know if you need more details.

HelioStrike:
@vfdev-5 Thanks, got most of the info I needed!

I added the tests, but the code hangs, so I commented it out for now. I think the expected behavior for now is for pytest to return true as soon as it sees the warning, instead of executing further. Would you happen to know how this issue can be fixed?

vfdev-5:
@HelioStrike thanks for the update! 

> I added the tests, but the code hangs, so I commented it out for now.

Let me give you some explanations of what happens in your test : 
```
    os.environ["RANK"] = "1"
    os.environ["WORLD_SIZE"] = "1"
    os.environ["MASTER_ADDR"] = "0.0.0.0"
    os.environ["MASTER_PORT"] = "2222"
    dist.init_process_group(backend="nccl", init_method="env://")
```
In this case it is indicated that total number of participating devices/processes (=world size) is 1. It means that distributed communication group will have processes indexed from 0 up to 1 non-included to communicate with.
And current process (=rank) is 1 instead of 0, so dist group waits for rank 0 to appear and thus hangs.

What we would like to check is the following: 

- backend: nccl => GPUs
- world size > 1
- user manually creates dist group as above: https://github.com/pytorch/ignite/pull/1335#issuecomment-702700856

The test setup should be something like here : 
https://github.com/pytorch/ignite/blob/d384dc664b0e9d7241b5f688ca0f92b600a3cb4d/tests/ignite/distributed/comp_models/test_native.py#L254

and add `@pytest.mark.skipif(torch.cuda.device_count() < 2, reason="Skip if less than 1 GPU")`


Please, let me know if you have no access to more than 1 GPU. Anyway, here in PRs our CI does not run tests on GPUs but will run once merged to the repository master or any other branch.

HelioStrike:
@vfdev-5 Thank you for the neat explanation! I'm new to the repo but I think I have a better understanding of things after reading it. I made the suggested changes but I don't have access to 2 GPUs to test it. :(

tests/ignite/distributed/comp_models/test_native.py
```diff
@@ -299,3 +299,27 @@ def test__native_dist_model_spawn_gloo():
 @pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
 def test__native_dist_model_spawn_nccl():
     _test__native_dist_model_spawn("nccl", num_workers_per_machine=torch.cuda.device_count(), device="cuda")
+
+
+@pytest.mark.distributed
+@pytest.mark.skipif(torch.cuda.device_count() < 2, reason="Skip if less than 2 GPUs")
+def test__warning_if_deviceindex_less_than_localrank():
```
vfdev-5, line 306: @HelioStrike please use `local_rank`, `world_size` pytest fixtures to setup the test

tests/ignite/distributed/comp_models/test_native.py
```diff
@@ -299,3 +299,27 @@ def test__native_dist_model_spawn_gloo():
...
+    import os
+    import torch.distributed as dist
+    import ignite.distributed as idist
+
+    os.environ["RANK"] = "1"
```
vfdev-5, line 311: WIth fixtures we do not need this os env setup for RANK and WORLD_SIZE.

tests/ignite/distributed/comp_models/test_native.py
```diff
@@ -299,3 +299,27 @@ def test__native_dist_model_spawn_gloo():
...
+    os.environ["WORLD_SIZE"] = "2"
+    os.environ["MASTER_ADDR"] = "0.0.0.0"
+    os.environ["MASTER_PORT"] = "2222"
+
+    dist.init_process_group(backend="nccl", init_method="env://")
+
+    pytest.warns("Current device index is less than current local rank.", idist.get_world_size)
```
vfdev-5, line 318: let's use `pytest.warns` as a context manager. See how it is done in other tests.

vfdev-5 (commented):
Thanks for the update

HelioStrike:
@vfdev-5 Updated! Only issue is that the `local_rank` fixture only sets up the `LOCAL_RANK` environment variable, and I still get errors saying that `RANK` is missing. Made all the other changes.

vfdev-5:
> Only issue is that the local_rank fixture only sets up the LOCAL_RANK environment variable, and I still get errors saying that RANK is missing

@HelioStrike yes, that's true. Actually I want to make the test like here https://github.com/pytorch/ignite/blob/d384dc664b0e9d7241b5f688ca0f92b600a3cb4d/tests/ignite/distributed/comp_models/test_native.py#L252-L254

which calls

https://github.com/pytorch/ignite/blob/d384dc664b0e9d7241b5f688ca0f92b600a3cb4d/tests/ignite/distributed/comp_models/test_native.py#L208

The problem with the way you configure ddp group is that the rank is always 1 but it has to be 0 and 1 for different processes.

HelioStrike:
@vfdev-5 I see now. I updated the code, but is the `true_device` argument correct? Also, like `_test__native_dist_model_create_from_context_dist`, are there any other cases we'd want to test?

vfdev-5:
@HelioStrike thanks for working on that. Sorry, I think I badly explained how we would like to test the warning. 
Previous links show the example of how to setup distributed configuration and test other things:
- `test__native_dist_model_create_dist_nccl` : executes tests
- `_test__native_dist_model_create_from_context_dist` : a subtest inside `test__native_dist_model_create_dist_nccl`

We would like to write another executable test where we setup distributed conf like above, but we test our particular thing : if `idist.get_world_size()` raises a warning.

If you'd like to understand all that in detail, you can try to run the test in Colab on a single GPU. It may be a bit tricky to run the test with 2 processes, but you can get a better understanding of how ddp works from user's perspective. 
If it is too complicated for you, I can update the test myself later...

HelioStrike:
@vfdev-5 I see.. I think I got it. I noticed that eventually `_test__native_dist_model_create_from_context_dist` calls `.get_world_size()`, so I put that in directly, but yes, I should probably create a new subtest.

vfdev-5:
@HelioStrike that's true, `_test__native_dist_model_create_from_context_dist` also calls get world size and probably raises warning too. So, we have to fix that test by adding `torch.cuda.set_device("cuda:lrank")`. Anyway, I have to check that from my side on 2 GPUs.

HelioStrike:
@vfdev-5 Sorry about this, I think I lack the knowledge to be able to work with this properly...

vfdev-5:
@HelioStrike no problems, it is a bit a tough subject and requires specific infrastructure to check. I'll finalize this PR later. 
Feel free to pick another hactoberfest issue which normally require less particular knowledge etc...

ignite/distributed/comp_models/native.py
```diff
@@ -97,7 +97,13 @@ def _init_from_context(self) -> None:
             self._setup_attrs()
 
         def _compute_nproc_per_node(self) -> int:
-            tensor = torch.tensor([self.get_local_rank() + 1]).to(self.device())
+            print("_compute_nproc_per_node")
```
sdesrozis, line 100: debug print to remove ?

sdesrozis:
Added a comment. Sorry for the delay...

ignite/distributed/comp_models/native.py
vfdev-5: Thanks !

PR #1338, opened 2020-10-02
Update documentation for terminate Events

### LINKED ISSUE #1332: Event description TERMINATE_SINGLE_EPOCH not matching its actual behavior
## 🐛 Bug description

From the documentation it seems that the event `TERMINATE_SINGLE_EPOCH` is only fired, if `engine.terminate_epoch` is called. 
I discovered, that the event is also fired, if I call `engine.terminate` inside of an epoch as can be seen in the code linked below.

Event description:
https://github.com/pytorch/ignite/blob/51d3e3e54ee042b5ef1b73a489aabe4b75968411/ignite/engine/events.py#L161-L162

Current implementation:
https://github.com/pytorch/ignite/blob/51d3e3e54ee042b5ef1b73a489aabe4b75968411/ignite/engine/engine.py#L817-L821

IMO, the implementation is fine and only the documentation should be updated. 

I think the other option to not fire `TERMINATE_SINGLE_EPOCH` if `enginge.terminate` is called would make handling different types of termination more complicated. 
Because:     
Assuming you have to do some post processing after each epoch and it doesn't matter if the epoch completes with `engine.terminate`, `engine.terminate_epoch` or by `StopIteration` from the data loader. 
You would have to attach the same processing function three times, for `EPOCH_COMPLETED`, `TERMINATE_SINGLE_EPOCH` and `TERMINATE`.  

At the moment one have to attach post processing function to the two events `EPOCH_COMPLETED` and `TERMINATE_SINGLE_EPOCH` to catch termination via command. 
I didn't expect that `EPOCH_COMPLETED` is not fired if I call any termination method. Is this documented somewhere? Haven't found it. 

Thanks in advance :) 

EDIT: Sorry, I missed that the signature of the function that is called on `EPOCH_COMPLETED` is different than for `TERMINATE_SINGLE_EPOCH`. 
EDIT2: Fixed event name `TERMINATE_SINGLE_EPOCH`

EDIT3: I think, now I got it:

- If `engine.terminate` is called, `TERMINATE_SINGLE_EPOCH` and `TERMINATE` are called but not `EPOCH_COMPLETED`.
- If `engine.terminate_epoch` is called, `TERMINATE_SINGLE_EPOCH` and `EPOCH_COMPLETED` are called but not `TERMINATE`. 
- If  epoch completes without termination, `EPOCH_COMPLETED` is called but not `TERMINATE_SINGLE_EPOCH` and `TERMINATE`. 

There is no common event that is called in all three cases and which could be used if all cases should be treated equally. 

## Environment

 - PyTorch Version (e.g., 1.4):  1.5.1
 - Ignite Version (e.g., 0.3.0): 0.4.2
 - OS (e.g., Linux): Windows
 - Python version: 3.7.7
 - Any other relevant information: -

shngt:
Fixes #1332 

Description: 

Adds additional comments and a table illustrating when the different terminate events are triggered.

Check list:

### DIFF
--- a/ignite/engine/events.py
+++ b/ignite/engine/events.py
@@ -166,20 +166,31 @@ class Events(EventEnum):
       after receiving :meth:`~ignite.engine.engine.Engine.terminate()` call.
 
     - EPOCH_COMPLETED : triggered when the epoch is ended. Note that this is triggered even
-      when :meth:`~ignite.engine.engine.Engine.terminate_epoch()`is called.
+      when :meth:`~ignite.engine.engine.Engine.terminate_epoch()` is called.
     - COMPLETED : triggered when engine's run is completed
 
     The table below illustrates which events are triggered when various termination methods are called.
 
-    +--------------------------------------------------------+-----------------+------------------------+-----------+
-    | Method \\ Event                                        |`EVENT_COMPLETED`|`TERMINATE_SINGLE_EPOCH`|`TERMINATE`|
-    +========================================================+=================+========================+===========+
-    | no termination                                         | ✔               |                        |           |
-    +--------------------------------------------------------+-----------------+------------------------+-----------+
-    | :meth:`~ignite.engine.engine.Engine.terminate_epoch()` | ✔               | ✔                      |           |
-    +--------------------------------------------------------+-----------------+------------------------+-----------+
-    | :meth:`~ignite.engine.engine.Engine.terminate()`       |                 | ✔                      | ✔         |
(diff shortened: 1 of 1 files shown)


vfdev-5:
@shngt thanks for the PR. Could you please make a screenshot of rendered docs with the table ?

shngt:
Screenshot of documentation changes
Here's how it looks on my system. Sorry for not doing this earlier, I haven't done this kind of thing before

vfdev-5 (approved):
@shngt Thanks a lot! Looks very nice :)

PR #1352, opened 2020-10-04
Activate mypy in ignite.utils

### LINKED ISSUE #1341: MyPy: improve ignite.utils module
## 🚀 Feature

Currently, mypy ignores all errors for all modules. We have to rework our typing such that mypy checks the code.
In this issue, let's improve https://github.com/pytorch/ignite/blob/master/ignite/utils.py module such that mypy passes on it.

Please, take a look at CONTRIBUTING guide.

gruebel:
Fixes #1341 

Description:
Since numpy doesn't offer any type hints or library stubs, I added the `ignore_missing_imports` flag for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,9 +38,11 @@ before_install: &before_install
   - pip install -r requirements-dev.txt
 
 install:
+  - pip install "mypy==0.782"
   - python setup.py install
 
 script:
+  - mypy --config-file mypy.ini
   - CI_PYTHON_VERSION="$TRAVIS_PYTHON_VERSION" sh tests/run_cpu_tests.sh
 
 after_success:
@@ -51,14 +53,11 @@ jobs:
     - stage: Lint check
       python: "3.7"
       before_install: # Nothing to do
-      install:
-        - python setup.py install
-        - pip install flake8 "black==19.10b0" "isort==4.3.21" "mypy==0.782"
+      install: pip install flake8 "black==19.10b0" "isort==4.3.21"
       script:
         - flake8 .
         - black --check .
         - isort -rc -c .
-        - mypy --config-file mypy.ini
       after_success: # Nothing to do
 
     # GitHub Pages Deployment: https://docs.travis-ci.com/user/deployment/pages/
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,9 @@ jobs:
     - stage: Lint check
       python: "3.7"
       before_install: # Nothing to do
-      install: pip install flake8 "black==19.10b0" "isort==4.3.21" "mypy==0.782"
+      install:
+        - python setup.py install
+        - pip install flake8 "black==19.10b0" "isort==4.3.21" "mypy==0.782"
       script:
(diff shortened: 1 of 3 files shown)


gruebel:
@vfdev-5 tests are failing, but it has nothing to do with my changes, any advise?

.travis.yml
```diff
@@ -51,7 +51,9 @@ jobs:
     - stage: Lint check
       python: "3.7"
       before_install: # Nothing to do
-      install: pip install flake8 "black==19.10b0" "isort==4.3.21" "mypy==0.782"
+      install:
+        - python setup.py install
```
vfdev-5, line 55: I wonder why we have to install the package at Lint stage ?

ignite/utils.py
```diff
@@ -41,11 +41,13 @@ def apply_to_type(
     if isinstance(input_, (str, bytes)):
         return input_
     if isinstance(input_, collections.Mapping):
-        return type(input_)({k: apply_to_type(sample, input_type, func) for k, sample in input_.items()})
+        return cast(Callable, type(input_))(
+            {k: apply_to_type(sample, input_type, func) for k, sample in input_.items()}
+        )
     if isinstance(input_, tuple) and hasattr(input_, "_fields"):  # namedtuple
-        return type(input_)(*(apply_to_type(sample, input_type, func) for sample in input_))
+        return cast(Callable, type(input_))(*(apply_to_type(sample, input_type, func) for sample in input_))
```
vfdev-5, line 48: I was wondering if this is the only way to do this and have mypy happy ?

vfdev-5 (commented):
Thanks for the PR @gruebel !
CI failure is unrelated and happens sometimes with distributed tests.
I left a question.

.travis.yml
gruebel: without it `mypy` can't find any type hints, otherwise I would need to add the `ignore_missing_imports` flag for the `torch` package or add `# type: ignore` whenever torch is imported

ignite/utils.py
gruebel: at least I couldn't find any better solution. Somehow `mypy` doesn't understand what is happening there. I have to admit, that I also didn't understand at the beginning, what that weird syntax is doing and then I figured it out 😄

.travis.yml
vfdev-5: Maybe, let's move mypy check to https://github.com/pytorch/ignite/blob/4408a32fb676adac619dd515c440038bc4554d13/.travis.yml#L40-L41
such that we have correct environment and deps installed and Lint stage remains quick.

ignite/utils.py
vfdev-5: Yeah, I agree that this looks weird but it is a way to recreate initial structure and apply func on its entries...

.travis.yml
gruebel: ok, no problem

vfdev-5 (approved):
LGTM ! Thanks @gruebel !

PR #1354, opened 2020-10-04
Add probability to Accuracy

### LINKED ISSUE #1089: Possible improvements for Accuracy
In full detail the feature request is described [here](https://github.com/pytorch/pytorch/issues/22439#issuecomment-636454497), below is a quick recap.

There are two inconveniences I experience with the current interface of Accuracy.

#### 1. Inconsistent input format for binary classification and multiclass problems
In the first case, Accuracy expects labels as input, whilst in the second case it expects probabilities/logits. I am a somewhat experienced Ignite user and I still get confused by this behavior.

#### 2. No shortcuts for saying "I want to pass logits/probabilities as input"
In practice, I have never used Accuracy in the following manner for binary classification:
```python
accuracy = Accuracy()
```
Instead, I always do one of the following:
```
accuracy = Accuracy(transform=lambda x: torch.round(torch.sigmoid(x)))
# either
accuracy = Accuracy(transform=lambda x: torch.round(x))
```

_Suggested solution for both problems_: let the user explicitly say in which form input will be passed:
```python
import enum
class Accuracy(...):
    class Mode(enum.Enum):
        LABELS = enum.auto()
        PROBABILITIES = enum.auto()
        LOGITS = enum.auto()

    def __init__(self, mode=Mode.LABELS, ...):
        ...
```

The suggested interface can be also extended to support custom thresholds by adding the `__call__` method to the Mode class.

vcarpani:
Fixes #1089

Description:
- This PR improves the API for `Accuracy` according to the suggestions from the issue.

Still WIP, to be done:
* [x] Support custom binarization threshold.
* [x] Support logits input.

Check list:

### DIFF
--- a/docs/source/metrics.rst
+++ b/docs/source/metrics.rst
@@ -17,6 +17,34 @@ value is then computed using the output of the engine's ``process_function``:
     metric = Accuracy()
     metric.attach(engine, "accuracy")
 
+If the engine's prediction output ``y_pred`` represents probability estimates, it can be binarized using the
+``Mode.PROBABILITIES``:
+.. code-block:: python
+
+    def process_function(engine, batch):
+        # ...
+        y = torch.from_numpy(np.array([0, 0, 1]))
+        y_pred = torch.from_numpy(np.array([0.1, 0.2, 0.7]))
+        return y_pred, y
+
+    engine = Engine(process_function)
+    metric = Accuracy(mode=Accuracy.Mode.PROBABILITIES)
+    metric.attach(engine, "accuracy")
+
+If the engine's prediction output ``y_pred`` represents logits, it can be binarized using the
+``Mode.LOGITS``:
+.. code-block:: python
+
+    def process_function(engine, batch):
+        # ...
+        y = torch.from_numpy(np.array([0, 0, 1]))
+        y_pred = torch.from_numpy(np.array([-2.1, 0.6, 1.7]))
+        return y_pred, y
+
+    engine = Engine(process_function)
+    metric = Accuracy(mode=Accuracy.Mode.LOGITS)
+    metric.attach(engine, "accuracy")
+
 If the engine's output is not in the format ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y, ...}``, the user can
 use the ``output_transform`` argument to transform it:
 
@@ -41,13 +69,13 @@ use the ``output_transform`` argument to transform it:
 .. warning::
 
     Please, be careful when using ``lambda`` functions to setup multiple ``output_transform`` for multiple metrics
-    
+
     .. code-block:: python
 
         # Wrong
         # metrics_group = [Accuracy(output_transform=lambda output: output[name]) for name in names]
         # As lambda can not store `name` and all `output_transform` will use the last `name`
-        
+
         # A correct way. For example, using functools.partial
         from functools import partial
 
@@ -55,7 +83,7 @@ use the ``output_transform`` argument to transform it:
             return output[name]
 
         metrics_group = [Accuracy(output_transform=partial(ot_func, name=name)) for name in names]
-    
+
     For more details, see `here <https://discuss.pytorch.org/t/evaluate-multiple-models-with-one-evaluator-results-weird-metrics/96695>`_
 
 .. Note ::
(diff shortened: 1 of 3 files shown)


docs/source/metrics.rst
```diff
@@ -17,6 +17,34 @@ value is then computed using the output of the engine's ``process_function``:
     metric = Accuracy()
     metric.attach(engine, "accuracy")
 
+If the engine's prediction output ``y_pred`` represents probability estimates, it can be binarized using the
+``Mode.PROBABILITIES``:
```
vfdev-5, line 21: Maybe, it would worth to say which threshold is used by default to transform probas to binary predictions.

ignite/metrics/accuracy.py
```diff
@@ -125,17 +117,30 @@ def thresholded_output_transform(output):
         device (str or torch.device): specifies which device updates are accumulated on. Setting the metric's
             device to be the same as your ``update`` arguments ensures the ``update`` method is non-blocking. By
             default, CPU.
+        mode (Mode): specifies in which form input will be passed. This can be useful to directly compute
+            accuracy on the output of a neural network, which ofter return probabilities. By default, LABELS.
+        binarization_threshold (float): threshold for binarization of the input, in case a Mode that uses
```
vfdev-5, line 122:
```suggestion
        threshold (float): threshold for binarization of the input, in case a Mode that uses
```

docs/source/metrics.rst
```diff
@@ -17,6 +17,34 @@ value is then computed using the output of the engine's ``process_function``:
...
+        y = torch.from_numpy(np.array([0, 0, 1]))
+        y_pred = torch.from_numpy(np.array([0.1, 0.2, 0.7]))
+        return y_pred, y
+
+    engine = Engine(process_function)
+    metric = Accuracy(mode=Accuracy.Mode.PROBABILITIES)
+    metric.attach(engine, "accuracy")
+
+If the engine's prediction output ``y_pred`` represents logits, it can be binarized using the
+``Mode.LOGITS``:
```
vfdev-5, line 35: Same here to mention that we can setup the threshold for this mode.

ignite/metrics/accuracy.py
```diff
@@ -125,17 +117,30 @@ def thresholded_output_transform(output):
...
+        LABELS = 0
+        PROBABILITIES = 1
+        LOGITS = 2
+
     def __init__(
         self,
         output_transform: Callable = lambda x: x,
         is_multilabel: bool = False,
         device: Union[str, torch.device] = torch.device("cpu"),
+        mode: Mode = Mode.LABELS,
```
vfdev-5, line 137: I'm not sure about correctness of name `LABELS` for multi-class case where we require to pass probas or logits and then take argmax.

vfdev-5 (commented):
@vcarpani thanks for the PR and sorry for the delay. I didn't see that it is not draft anymore.
I left few comments to improve it and we have to discuss a bit on probably misleading terms...

ignite/metrics/accuracy.py
vcarpani: What about `UNCHANGED` or `RAW_INPUT`?

vcarpani:
@vfdev-5 the last commit should solve your comments

ignite/metrics/accuracy.py
vfdev-5: Yes, something like `RAW_INPUT` could work. I'm thinking if we could not generalize this new options to all possible inputs and metric type: binary, multiclass, multilabel...

vfdev-5:
@vcarpani thanks for the update! I left a comment in a discussion thread. 
I'm thinking if it could be possible to replace argmax op for multiclass type using these new options. A similar proposition previously was for Confusion Matrix (https://github.com/pytorch/ignite/issues/822) but something was a blocker for that.

vcarpani:
I'll try to implement that behaviour somewhere next week ;)

vcarpani:
> I'll try to implement that behaviour somewhere next week ;)

@vfdev-5, just wanted to update you on this one, I've been quite busy recently, but I started working on it

vfdev-5:
@vcarpani thanks for the head-up :)

Yura52:
I would like to add that "logits" in the context of multiclass problems usually mean the input to the _softmax_ (not the _sigmoid_) function. I am not sure if there are any usa cases for the current `Mode.LOGITS` implementation for the multiclass setting. As for `Mode.PROBABILITIES`, it looks fine.

PR #1351, opened 2020-10-04
[1] [contrib/engines] setup typing in contrib part of the library 

RaviTezu:
Fixes #977

Description: Setup Typing in `contrib` part of the library

PR #1355, opened 2020-10-04
Activate mypy in ignite.distributed

### LINKED ISSUE #1344: MyPy: improve ignite.distributed module
## 🚀 Feature

Currently, mypy ignores all errors for all modules. We have to rework our typing such that mypy checks the code.
In this issue, let's improve https://github.com/pytorch/ignite/tree/master/ignite/distributed module such that mypy passes on it.

Please, take a look at CONTRIBUTING guide.

gruebel:
Fixes #1344 

Description:
I added the `# type: ignore` comments, where needed. Some stuff is really strange, like using `@overload` and the `disallow_untyped_defs` flag is not supported by `mypy` 🤯 
I tried my best to add the correct types to parameters and functions, so please have a look 🙏 

In general, after activating `mypy` for the whole project, we could try out the `--strict` flag, which will probably find addtional things.


ignite/distributed/auto.py
```diff
@@ -16,7 +17,42 @@
...
+    class _XLADistributedOptimizer(Optimizer):
+        def __init__(self, optimizer: Optimizer) -> None:
+            super(self.__class__, self).__init__(optimizer.param_groups, {})
+            self.wrapped_optimizer = optimizer
+
+        def step(self, closure: Optional[Callable] = None) -> None:
+            xm.optimizer_step(self.wrapped_optimizer, barrier=True)
+
+
+def auto_dataloader(dataset: Dataset, **kwargs: Any) -> Union[DataLoader, _MpDeviceLoader]:
```
vfdev-5, line 55: Can we use `Union[DataLoader, "_MpDeviceLoader"]` instead of moving all those private helper classes to the top ?

ignite/distributed/auto.py
```diff
@@ -73,6 +109,8 @@ def auto_dataloader(dataset, **kwargs):
             kwargs["num_workers"] = (kwargs["num_workers"] + nproc - 1) // nproc
 
         if "batch_sampler" not in kwargs:
+            sampler: Union[DistributedProxySampler, DistributedSampler]
```
vfdev-5, line 112: We support py3.5 and unfortunately inline type hints like that wont work.

ignite/distributed/auto.py
```diff
@@ -266,22 +304,30 @@ class DistributedProxySampler(DistributedSampler):
...
             raise TypeError("Argument sampler should have length")
 
-        super(DistributedProxySampler, self).__init__(sampler, num_replicas=num_replicas, rank=rank, shuffle=False)
+        super(DistributedProxySampler, self).__init__(
+            sampler, num_replicas=num_replicas, rank=rank, shuffle=False  # type: ignore[arg-type]
+        )
         self.sampler = sampler
 
-    def __iter__(self):
+    def __setattr__(self, name: str, value: Any) -> None:
```
vfdev-5, line 320: I wonder why we need override `__setattr__` and `__getattr__` here ?

ignite/distributed/comp_models/base.py
```diff
@@ -132,9 +134,20 @@ def _apply_op(self, tensor: torch.Tensor, device: torch.device, fn: Callable, *a
             return tensor.to(device=tensor_device)
         return tensor
 
+    @overload
```
vfdev-5, line 137: Maybe, we can simply ignore type check on `_collective_op` and `broadcast` now and come back later ?

vfdev-5 (commented):
Thanks a lot @gruebel for the PR !
Make existing codebase pass mypy is something really tough. 
I left few comments.

gruebel:
I just ran `mypy` on `ignite.metrics` ~100errors 😱  `ignite.engine` doesn't look so bad with ~30.

I already started to work on `ignite.contrib`. Should I create an issue for it?

ignite/distributed/comp_models/base.py
gruebel: Sure, I will try it out tomorrow.

ignite/distributed/auto.py
gruebel: Thanks, good to know! Will fix it.

vfdev-5:
Maybe, let's create an issue for metrics and engine.

> I already started to work on ignite.contrib. Should I create an issue for it?

@gruebel thanks, however I was thinking to do that in 2 steps: 1) add typing, 2) make mypy happy. 
There is already a PR on introducing typing to contrib module : https://github.com/pytorch/ignite/pull/1351

If you'd like to help with reviewing it and then if you wish, you can make mypy on contrib. What do you think ?

ignite/distributed/auto.py
gruebel: It's because of this part here
https://github.com/pytorch/ignite/blob/3f8daafa7ea119bcf2845c7d042d576f2b9ec122/ignite/distributed/auto.py#L110-L120
`_MpDeviceLoader` doesn't have a attribute `sampler`

gruebel:
> If you'd like to help with reviewing it (would be awesome) and then, if you wish, you can make mypy on contrib. What do you think ?
Yeah, sure. Didn't see there is an open PR for it.

ignite/distributed/auto.py
vfdev-5: As now the latest XLA includes `MpDeviceLoader`, I think we can rewrite this part of the code. 
Do you know if there is any mypy correct way to add an attribute to an instance ?

ignite/distributed/auto.py
gruebel: yeah, I already did it locally.

ignite/distributed/auto.py
gruebel: the only thing I found so far was to use `__setattr__` and  `__getattr__`

ignite/distributed/comp_models/native.py
```diff
@@ -245,7 +245,7 @@ def _dist_worker_task_fn(
             node_rank: int,
             master_addr: str,
             master_port: str,
-            **kw: Any,
+            kw: Any,
```
vfdev-5, line 248: Maybe we can keep kw as kwargs and just remove the comma ?

gruebel: you mean rename it? `kw` -> `kwargs`? I only reverted my change here, I falsely made `kw` to `**kw` and broke a few tests 🙈

vfdev-5: Ah, OK, sorry, I just saw the last modification. Let's keep as it was =>
```python
kw: Any,
```

ignite/distributed/comp_models/xla.py
```diff
@@ -50,17 +50,17 @@ def __init__(self, backend=None, **kwargs):
             else:
                 self._init_from_context()
 
-        def _create_from_backend(self, backend, **kwargs):
+        def _create_from_backend(self, backend: str, **_kwargs: Any) -> None:
```
vfdev-5, line 53: How about this rename, can we keep it as `kwargs` instead of `_kwargs` ?

gruebel: sure, I can change it back, I only renamed it, because the variable is not used and my IDE complains about it.

ignite/distributed/launcher.py
```diff
@@ -214,7 +214,14 @@ def __init__(
             self.logger.info("- Parameters to spawn processes: \n\t{}".format(msg))
 
     @staticmethod
-    def _setup_spawn_params(nproc_per_node, nnodes, node_rank, master_addr, master_port, **spawn_kwargs):
+    def _setup_spawn_params(
+        nproc_per_node: int,
+        nnodes: Optional[int],
```
vfdev-5, line 219: Actually, I'd say all these args are not `Optional`, right ?

ignite/distributed/utils.py
```diff
@@ -427,21 +432,21 @@ def run(local_rank, *args, **kwargs):
...
+def _set_model(model: Any, temporary: bool = False) -> None:
     global _model, _need_to_sync
     _model = model
     _need_to_sync = True
     if not isinstance(_model, _SerialModel) and not temporary:
         _need_to_sync = False
 
 
-def _assert_backend(backend):
+def _assert_backend(backend: Optional[str]) -> None:
```
vfdev-5, line 443: Same here for `Optional` ?

ignite/distributed/utils.py
```diff
@@ -522,7 +527,7 @@ def show_config():
     logger.info("node rank: {}".format(get_node_rank()))
 
 
-def one_rank_only(rank: int = 0, with_barrier: bool = False):
+def one_rank_only(rank: int = 0, with_barrier: bool = False) -> Optional[Callable]:
```
vfdev-5, line 530: Actually, again about Optional, why it is `Optional[Callable]` and not `Callable` ?

ignite/distributed/comp_models/base.py
```diff
@@ -256,18 +261,18 @@ def create_from_context() -> "_SerialModel":
...
-    def spawn(*args, **kwargs):
+    def spawn(*args: Any, **kwargs: Any) -> None:
         raise NotImplementedError("Serial computation model does not implement spawn method")
 
     def all_reduce(self, tensor: Union[torch.Tensor, Number], op: str = "sum") -> Union[torch.Tensor, Number]:
         return tensor
 
-    def all_gather(self, tensor: Union[torch.Tensor, Number]) -> Union[torch.Tensor, Number]:
-        return tensor
+    def all_gather(self, tensor: Union[torch.Tensor, Number, str]) -> Union[torch.Tensor, Number, List[str]]:
```
vfdev-5, line 274: Is it correct `Union[torch.Tensor, Number, str]` and `Union[torch.Tensor, Number, List[str]]` ?
Seems like they are not consistent with below `cast(Union[torch.Tensor, Number], tensor)`. Can we keep the typing without `str` ?

vfdev-5 (commented):
Thanks for the update, I left some more comments

ignite/distributed/comp_models/base.py
gruebel: if I do so, several issues will pop up
```
ignite/distributed/comp_models/base.py:274: error: Argument 1 of "all_gather" is incompatible with supertype "ComputationModel"; supertype defines the argument type as "Union[Tensor, Number, str]"  [override]
```
this I can fix with `# type: ignore`, but then I also get
```
ignite/distributed/utils.py:352: error: Argument 1 to "all_gather" of "_SerialModel" has incompatible type "Union[Tensor, Number, str]"; expected "Union[Tensor, Number]"  [arg-type]
```
because of the return type of the Function
https://github.com/pytorch/ignite/blob/221b6dedd95ef6187ba1989013dfc49049ebcb62/ignite/distributed/utils.py#L332-L347
but of course, I can also put an `# type: ignore` there 😄

ignite/distributed/launcher.py
gruebel: it is due to the `__init__` function, they are all `None` by default and therefore all can be `None` except `nproc_per_node`
https://github.com/pytorch/ignite/blob/51067381b457c098b8dd80972ecfb8ab113baef2/ignite/distributed/launcher.py#L176-L185
and there are also couple of `None` checks inside the function.

ignite/distributed/utils.py
gruebel: same as above, because in the `__init__` function of `Parallel` `backend` is `None` by default
https://github.com/pytorch/ignite/blob/51067381b457c098b8dd80972ecfb8ab113baef2/ignite/distributed/launcher.py#L176-L185
and in the function `run` it calls `utils.spwan`
https://github.com/pytorch/ignite/blob/51067381b457c098b8dd80972ecfb8ab113baef2/ignite/distributed/launcher.py#L271
and in the whole `Parallel` `self.backend` is not set with a new backend, therefore it could be `None` till this point.

ignite/distributed/launcher.py
vfdev-5: Thanks, it makes sense. Could you please add `= None` to make `Optional` more explicit.

ignite/distributed/utils.py
vfdev-5: I think there is some to fix in the code as here it shouldn't be optional and can't be None.
And `ignite/ignite/distributed/launcher.py` does not use `_assert_backend`. It is used only in
https://github.com/pytorch/ignite/blob/b977b746edcbec9c29b8b88f40157f7d7db2601f/ignite/distributed/utils.py#L449
where backend is not optional.

ignite/distributed/comp_models/base.py
vfdev-5: OK, it makes sense to implement it in the same way as distributed implementations. Let's do this in another PR.
Idea is to rework types for a) `utils.all_gather` and all its realizations: serial, native, hvd, xla. 
Thinking more about its type annotations, i'd say it should be 
```python
def all_gather(tensor: Union[torch.Tensor, Number, str]) -> Union[torch.Tensor, List[Number], List[str]]:
```
Such that here in _SerialModel it should be implemented as 
```python
def all_gather(self, tensor: Union[torch.Tensor, Number, str]) -> Union[torch.Tensor, List[Number], List[str]]:
    if isinstance(tensor, torch.Tensor):
        return tensor
    return cast(Union[List[Number], List[str]], tensor)
```

ignite/distributed/utils.py
```diff
@@ -190,8 +190,13 @@ def hostname() -> str:
 
 
 def spawn(
-    backend: str, fn: Callable, args: Tuple, kwargs_dict: Optional[Mapping] = None, nproc_per_node: int = 1, **kwargs
-):
+    backend: Optional[str],
```
vfdev-5, line 193: This is strange as backend can't be None at all here.

ignite/distributed/utils.py
gruebel: No, it is also used in the `spawn` function
https://github.com/pytorch/ignite/blob/b977b746edcbec9c29b8b88f40157f7d7db2601f/ignite/distributed/utils.py#L192-L199
https://github.com/pytorch/ignite/blob/b977b746edcbec9c29b8b88f40157f7d7db2601f/ignite/distributed/utils.py#L306
there is a very long function description, which I didn't want to add here 😄

ignite/distributed/comp_models/base.py
gruebel: ok, I will add the `# type: ignore` and it should be fixed via a different PR as you suggested 👍

ignite/distributed/utils.py
vfdev-5: Yes, true, that's why it was refactored into `_assert_backend`, but anyway in both `backend` can't be optional.

gruebel: ok, is it then ok to add here `... and self.backend is not None:`?
https://github.com/pytorch/ignite/blob/b977b746edcbec9c29b8b88f40157f7d7db2601f/ignite/distributed/launcher.py#L276-L278

gruebel: otherwise I can also put here `# type: ignore`

ignite/distributed/utils.py
gruebel: Because of the `wrapper` function, which can be `None` if `get_rank() != rank`
https://github.com/pytorch/ignite/blob/b977b746edcbec9c29b8b88f40157f7d7db2601f/ignite/distributed/utils.py#L552-L564

ignite/distributed/auto.py
gruebel: I will remove them and put `# type: ignore` next to it. This probably done via a separate PR.

ignite/distributed/utils.py
vfdev-5: Here `ret` is the output of the function, I think wrapper is always defined but `func` is called only on specified rank.

gruebel: yeah, you are totally right. no idea, why I misread it 🙈

gruebel:
@vfdev-5 any idea about the failing tests?

vfdev-5:
Good question. There are errors:
- mypy: https://travis-ci.org/github/pytorch/ignite/jobs/733127268#L1421
- test: https://github.com/pytorch/ignite/pull/1355/checks?check_run_id=1211799248#step:5:117

Let me restart XLA test. For mypy, maybe, we have investigate it more.

ignite/distributed/auto.py
```diff
@@ -101,7 +104,7 @@ def auto_dataloader(dataset, **kwargs):
         kwargs["pin_memory"] = kwargs.get("pin_memory", "cuda" in idist.device().type)
 
     logger.info("Use data loader kwargs for dataset '{}': \n\t{}".format(repr(dataset)[:20].strip(), kwargs))
-    dataloader = DataLoader(dataset, **kwargs)
+    dataloader = DataLoader(dataset, **kwargs)  # type: Union[DataLoader, "_MpDeviceLoader"]
```
vfdev-5, line 107: Here it is DataLoader type without _MpDeviceLoader

ignite/distributed/auto.py
vfdev-5: So, will you remove them here or in another PR ?

vfdev-5:
@gruebel could you please open a follow-up issue for https://github.com/pytorch/ignite/pull/1355#discussion_r499854850

ignite/distributed/auto.py
gruebel: at this specific line you are right, but a few lines later it can be reassigned with `_MpDeviceLoader` and results in following error
```
ignite/distributed/auto.py:122: error: Incompatible types in assignment (expression has type "_MpDeviceLoader", variable has type "DataLoader[Any]")  [assignment]
```

vfdev-5: OK, what if we do something like 
```python
dataloader = DataLoader(dataset, **kwargs)  # type: DataLoader

if idist.has_xla_support and idist.backend() == idist_xla.XLA_TPU and world_size > 1:
    sampler = dataloader.sampler  # type: ignore[union-attr]
    mp_dataloader = mp_device_loader_cls(dataloader, idist.device())  # type: "_MpDeviceLoader"
    mp_dataloader.sampler = sampler  # type: ignore[attr-defined]
    return mp_dataloader

return dataloader
```

gruebel: yea, should work. let me try it.

gruebel:
> @gruebel could you please open a follow-up issue for [#1355 (comment)](https://github.com/pytorch/ignite/pull/1355#discussion_r499854850)

done #1368!

gruebel:
@vfdev-5 everything is green, except the TPU test.

vfdev-5:
@gruebel thanks for the issue !

I'm trying to understand if XLA failing test is related to this PR or not, I restarted master XLA test which was green 6h ago : https://github.com/pytorch/ignite/runs/1214898148

Let's see and either we can merge this PR if it is unrelated or inspect and fix the issue...

gruebel:
I think I found it, it was from my change. I will fix it.

vfdev-5 (approved):
Thanks a lot @gruebel !

gruebel:
@vfdev-5 I'm just curious, was it really helpful for you or is it more work, than doing it on your own? Because you are spending a lot of time in reviewing what me and others are doing here.

Thanks for your effort and patience 💪

vfdev-5:
@gruebel that's a good question. For mypy it is something really helpful as we have no much experience with that and mypy reveals where the code is badly written. For other features sometimes it can be quickly to do it myself and what counts is if it matters to help the person to learn our package and other related tech.

PR #1372, opened 2020-10-07
Pin sphinx to 3.2.1 version

### LINKED ISSUE #1369: [DOCS] Rendered class and method signature overlaps with [source] after sphinx update
## 🐛 Bug description

I've noticed that after updating to `sphinx=3.1`, documentation has some problem with class header layout, if you go to the docs webpage, you can find that in some cases type hits overlap with [SOURCE] button, in some cases even definition overlaps as well. Zooming-in/out doesn't help.

Here is what I get:
docs1
docs2
doc3

After roll back to old version everything works fine, Also, I've noticed in [pytorch repo](https://github.com/pytorch/pytorch/blob/0da6730f029863b3c812e05ddb3cec8a316c7c1c/docs/requirements.txt#L1) that they're using `sphinx==2.4.4`, I've tried that version and it works well too.

## Environment

`sphinx==3.1`

trsvchn:
Related to #1369 

Description:

Downgrades `sphinx` to `2.4.4` which works well with `pytorch_sphinx_theme`

Check list:


vfdev-5:
Thanks for the PR @trsvchn !
Looks like the difference is with rendered links

2.4.4
Screen Shot 2020-10-07 at 09 46 54

3.1
Screen Shot 2020-10-07 at 09 46 16
 
str and ignite.engine.events.EventEnum are not link anymore and thus correctly rendered. I wonder if there is an option to deactivate that for 3.1 ?

trsvchn:
@vfdev-5 Looks like `sphinxcontrib-versioning` is not working, right? Since, we are expecting a lot doc builds, but here it just stuck.

I've found similar issue [Sphinx: building docs throws error](https://github.com/SiLab-Bonn/testbeam_analysis/issues/81) and the solution was to use previous `sphinx` version.

Ok, `sphinxcontrib-versioning` is yet another thing that need to be taken into consideration.

trsvchn:
But in our case it works with new one)

trsvchn:
> str, dict and ignite.engine.events.EventEnum are not links anymore and thus correctly rendered. I wonder if there is an option to disable that for 3.1 ?

I'll take a look

vfdev-5:
Thanks @trsvchn 

> Ok, sphinxcontrib-versioning is yet another thing that we need to take care about.

We use our fork of sphinxcontrib-versioning so it can be adapted if needed

trsvchn:
Ok, this is just a test to check that latest version works fine with `sphinxcontrib-versioning`

trsvchn:
Well, `3.1` and `3.2.1.` work fine with `sphinxcontrib-versioning`

trsvchn:
Well, `3.1` and `3.2.1.` work fine with `sphinxcontrib-versioning`

vfdev-5:
@trsvchn do you remember why we didn't keep 3.2.1 sphinx version, was there any blocker ?

trsvchn:
@vfdev-5 There was a problem with links in class/func signatures (that problem with [source] button) but we fixed it. So, I think there is no blocker anymore.

vfdev-5:
Thanks @trsvchn ! Could you please update it ?

trsvchn:
@vfdev-5 Sure!

vfdev-5 (approved):
Thanks @trsvchn !

PR #1374, opened 2020-10-07
Disable cross-ref links for type annotations

### LINKED ISSUE #1369: [DOCS] Rendered class and method signature overlaps with [source] after sphinx update

trsvchn:
Fixes #1369 

Description:

This will patch built-in type annotation parser and disable cross-ref links. This should fix issue signature rendering and [source]. At least for now, this patch works with `sphins` `3.1` and the latest one.

Check list:

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -192,3 +192,16 @@ intersphinx_mapping = {"https://docs.python.org/": None}
 
 # If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = True
+
+# -- Type hints configs ------------------------------------------------------
+
+autodoc_typehints = "signature"
+
+# -- A patch that turns-off cross refs for type annotations ------------------
+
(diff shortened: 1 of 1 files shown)


vfdev-5 (approved):
Thanks a lot @trsvchn !

PR #1379, opened 2020-10-10
Activate mypy in ignite.engine

gruebel:
Description:
I tried to add all missing types, not sure if I used the correct ones.

### DIFF
--- a/ignite/engine/deterministic.py
+++ b/ignite/engine/deterministic.py
@@ -2,7 +2,7 @@ import random
 import warnings
 from collections import OrderedDict
 from functools import wraps
-from typing import Callable, Generator, Iterator, Optional
+from typing import Any, Callable, Generator, Iterator, List, Optional, cast
 
 import torch
 from torch.utils.data import DataLoader
@@ -61,7 +61,7 @@ class ReproducibleBatchSampler(BatchSampler):
         if not isinstance(batch_sampler, BatchSampler):
             raise TypeError("Argument batch_sampler should be torch.utils.data.sampler.BatchSampler")
 
-        self.batch_indices = None
+        self.batch_indices = None  # type: Optional[List]
         self.batch_sampler = batch_sampler
         self.start_iteration = start_iteration
         self.sampler = self.batch_sampler.sampler
@@ -77,14 +77,14 @@ class ReproducibleBatchSampler(BatchSampler):
 
     def __iter__(self) -> Generator:
         self.setup_batch_indices()
-        for batch in self.batch_indices:
+        for batch in cast(List, self.batch_indices):
             yield batch
 
     def __len__(self) -> int:
         return len(self.batch_sampler)
 
 
-def _get_rng_states():
+def _get_rng_states() -> List[Any]:
     output = [random.getstate(), torch.get_rng_state()]
     try:
         import numpy as np
@@ -96,7 +96,7 @@ def _get_rng_states():
     return output
 
 
-def _set_rng_states(rng_states):
+def _set_rng_states(rng_states: List[Any]) -> None:
     random.setstate(rng_states[0])
     torch.set_rng_state(rng_states[1])
     try:
@@ -107,14 +107,14 @@ def _set_rng_states(rng_states):
         pass
 
 
-def _repr_rng_state(rng_states):
+def _repr_rng_state(rng_states: List[Any]) -> str:
     from hashlib import md5
 
     out = " ".join([md5(str(list(s)).encode("utf-8")).hexdigest() for s in rng_states])
     return out
 
 
-def keep_random_state(func: Callable):
+def keep_random_state(func: Callable) -> Callable:
     """Helper decorator to keep random state of torch, numpy and random intact
     while executing a function. For more details on usage, please see :ref:`Dataflow synchronization`.
 
@@ -123,7 +123,7 @@ def keep_random_state(func: Callable):
     """
 
     @wraps(func)
-    def wrapper(*args, **kwargs):
+    def wrapper(*args: Any, **kwargs: Any) -> None:
         rng_states = _get_rng_states()
         func(*args, **kwargs)
         _set_rng_states(rng_states)
@@ -182,9 +182,9 @@ class DeterministicEngine(Engine):
 
     def _init_run(self) -> None:
         seed = torch.randint(0, int(1e9), (1,)).item()
-        self.state.seed = seed
+        self.state.seed = cast(int, seed)
         if not hasattr(self.state, "rng_states"):
-            self.state.rng_states = None
+            self.state.rng_states = None  # type: ignore[attr-defined]
 
         if torch.cuda.is_available():
             torch.backends.cudnn.deterministic = True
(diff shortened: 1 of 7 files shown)


gruebel:
It seems like, that the missing attributes in the `DataLoader` interface were fixed with the newer `pytorch` version, that's why there is quite often a mismatch between the normal build and the nightly one regarding `mypy`.

vfdev-5:
@gruebel thanks for the PR ! I'll check it asap this coming week. 

> that's why there is quite often a mismatch between the normal build and the nightly one regarding mypy

Looks like CI still passes or I'm missing something ?

gruebel:
It is just a general information or you 😄 when the project uses a newer version of `pytorch`, we can probably remove a lot of the `# type: ignore` comments, because of the missing attributes in `DataLoader`

Can you probably restart the 3.6 and 3.7 nightly build, looks like there were many timeouts and mysterious errors.

ignite/engine/engine.py
```diff
@@ -706,23 +708,23 @@ def switch_batch(engine):
...
 
     @staticmethod
-    def _init_timers(state: State):
+    def _init_timers(state: State) -> None:
         state.times[Events.EPOCH_COMPLETED.name] = 0.0
         state.times[Events.COMPLETED.name] = 0.0
 
-    def _get_data_length(self, data):
-        data_length = None
+    def _get_data_length(self, data: Optional[Iterable] = None) -> Optional[int]:
```
vfdev-5, line 715: I'm not sure about `Optional[Iterable]` here, I think `data` can not be None here.

ignite/engine/engine.py
```diff
@@ -706,23 +708,23 @@ def switch_batch(engine):
...
         except TypeError:
             # _InfiniteConstantSampler can raise a TypeError on DataLoader length of a IterableDataset
             pass
-        return data_length
+        return None
 
     def _setup_engine(self) -> None:
         iteration = self.state.iteration
-        self._dataloader_iter = iter(self.state.dataloader)
+        if self.state.dataloader is not None:
```
vfdev-5, line 726: I suppose that this check is related to the fact that this function can be used with default `self.state = State()` and where dataloader is None. Let's raise an exception (as above) if it is None instead.

ignite/engine/deterministic.py
```diff
@@ -77,14 +77,14 @@ def setup_batch_indices(self) -> None:
 
     def __iter__(self) -> Generator:
         self.setup_batch_indices()
-        for batch in self.batch_indices:
+        for batch in cast(List, self.batch_indices):
```
vfdev-5, line 80: Can't we annotate `self.batch_indices = []  # type: List` instead of the cast here ?

ignite/engine/deterministic.py
```diff
@@ -228,28 +230,30 @@ def _setup_engine(self) -> None:
...
         in_the_middle = self.state.iteration % self._dataloader_len > 0 if self._dataloader_len is not None else False
         if (getattr(self.state, "rng_states", None) is not None) and in_the_middle:
-            _set_rng_states(self.state.rng_states)
-            self.state.rng_states = None
+            _set_rng_states(self.state.rng_states)  # type: ignore[attr-defined]
+            self.state.rng_states = None  # type: ignore[attr-defined]
 
     def _from_iteration(self, iteration: int) -> Iterator:
+        if self.state.dataloader is None:
+            raise TypeError("dataloader can not be None")
```
vfdev-5, line 238:
```suggestion
            raise RuntimeError(
                "Internal error, self.state.dataloader is None. Please, file an issue if encounter this error."
            )
```

ignite/engine/engine.py
```diff
@@ -791,7 +793,8 @@ def _run_once_on_dataset(self) -> float:
                     # Avoid Events.GET_BATCH_STARTED triggered twice when data iter is restarted
                     if self.last_event_name != Events.DATALOADER_STOP_ITERATION:
                         self._fire_event(Events.GET_BATCH_STARTED)
-                    self.state.batch = next(self._dataloader_iter)
+                    if self._dataloader_iter is not None:
```
vfdev-5, line 796: This condition seems a bit strange too. Let's check and raise an exception before entering while loop.

ignite/engine/engine.py
```diff
@@ -809,12 +812,15 @@ def _run_once_on_dataset(self) -> float:
...
                                 "Current iteration: {} vs Total iterations to run : {}".format(
-                                    self.state.iteration, self.state.epoch_length * self.state.max_epochs
+                                    self.state.iteration,
+                                    self.state.epoch_length * self.state.max_epochs,  # type: ignore[operator]
                                 )
                             )
                         break
 
                     self._fire_event(Events.DATALOADER_STOP_ITERATION)
+                    if self.state.dataloader is None:
```
vfdev-5, line 822: Can we do this and below "check->raise" once and before while loop ?

ignite/engine/events.py
```diff
@@ -437,7 +447,9 @@ def print_epoch(engine):
         # print_epoch handler is now unregistered
     """
 
-    def __init__(self, event_name: Union[CallableEventWithFilter, Enum, EventsList], handler: Callable, engine):
+    def __init__(
+        self, event_name: Union[CallableEventWithFilter, Enum, EventsList], handler: Callable, engine: Engine
```
vfdev-5, line 451: Do you think it would be better to add `EventType` as `Union[CallableEventWithFilter, Enum, EventsList, Events]` ?

vfdev-5 (commented):
Thanks a lot @gruebel !
Mypy definitely forces to rethink about the implementation and hopefully improve it. 
I left some comments to discuss.

ignite/engine/deterministic.py
```diff
@@ -182,9 +182,9 @@ def state_dict(self) -> OrderedDict:
 
     def _init_run(self) -> None:
         seed = torch.randint(0, int(1e9), (1,)).item()
-        self.state.seed = seed
+        self.state.seed = cast(int, seed)
```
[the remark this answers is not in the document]
gruebel: sure, no problem 👍

ignite/engine/engine.py
gruebel: Yeah, you are right. I retraced the usage of `_setup_engine()` and some hops back, there is an assignment of `dataloader` with an not `None` value. As it seems, it is too hard for `mypy` to figure that out. Then I will do a `None` check and `raise` an error.

ignite/engine/engine.py
gruebel: ok, gotcha 👍

ignite/engine/deterministic.py
gruebel: actually the problem is the `None` assignment in the `__init__` function
https://github.com/pytorch/ignite/blob/b2e1b33c2154483ea5204bc1322c8af33e51c0c1/ignite/engine/deterministic.py#L60-L67
therefore `mypy` thinks, it could be `None` at that point.

ignite/engine/events.py
```diff
@@ -3,10 +3,17 @@
...
 from types import DynamicClassAttribute
-from typing import Callable, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, Iterator, List, Optional, Union
+
+from torch.utils.data import DataLoader
 
 from ignite.engine.utils import _check_signature
 
+if TYPE_CHECKING:
+    from ignite.engine.engine import Engine
```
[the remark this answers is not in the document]
gruebel: yeah, correct! A forward reference only works with self referencing classes, BUT I can avoid the `else` part with the forward reference. Just checked the PEP484 documentation, always learning something new 💪

ignite/engine/events.py
gruebel: Do you mean to define a type alias?
```
EventType = TypeVar("EventType", CallableEventWithFilter, Enum, EventsList, Events)
```
or just add `Events` inside the Union block?

Yeah, it makes sense to add it regarding the purpose of the class. For `mypy` it doesn't matter with or without it 😄

gruebel: just a side note, it is definitely a bad idea to use a `TypeVar` because `mypy` doesn't understand the usage. directly complaining about it. There is also an open issue in the `mypy` project and it seems like, it works as designed, not a bug, it is a feature...

ignite/engine/deterministic.py
vfdev-5: I see, let's set `self.batch_indices = []` in the constructor. Shouldn't be that different though.

ignite/engine/events.py
vfdev-5: Yes, the idea was to create an alias instead of passing all the times Union of bunch of classes.
Well, if it is not possible, OK, but sad

vfdev-5 (approved):
LGTM ! Thanks @gruebel

vfdev-5:
@gruebel could you please send a follow-up PR for fixing this warning:
```
ignite/engine/__init__.py:31
  /ignite/ignite/engine/__init__.py:31: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    ) -> Tuple[Union[torch.Tensor, collections.Sequence, collections.Mapping, str, bytes], ...]:
```
Thanks !

PR #1381, opened 2020-10-13
Adding max_iters as an optional arg in Engine run

### LINKED ISSUE #1300: max_iters as optional arg in Engine run
## 🚀 Feature

Idea is to cover the following usage:
```python
trainer.run(data, max_iters=1234)
# here we still have epoch length as len(data) if defined
```
which can be roughly done as 
```python
max_iters = 1234
trainer.run(data, epoch_length=max_iters)
# here we run 1 epoch of size max_iters
```
or as something like
```python
max_iters = 1234
epoch_length = ...
max_epochs = max_iters // epoch_length + 1

@trainer.on(Events.ITERATION_COMPLETED(once=max_iters))
def stop():
    trainer.terminate()

trainer.run(data, max_epochs=max_epochs)
```

Argument `max_iters` is mutually exclusive with `max_epochs`.

thescripted:
Fixes #1300 

Description:
New tests have not yet been written, and I believe some descriptions/comments will need to be updated. I'm happy to do that, but I'm opening up this PR for feedback and general indications to see if I'm on the right track. 

I've assumed that reaching max iteration will fire a "Terminate" event rather than "Epoch Completed," but I don't know if either of those events are accurate for this condition.

FInally, adding a condition to this block:
https://github.com/pytorch/ignite/blob/a8a9c7cd878ff31640bd4056e056277b047c7ddf/ignite/engine/engine.py#L557-L558 

currently breaks some tests, so I have left that alone for the moment. 

Check list:


ignite/engine/engine.py
```diff
@@ -683,16 +686,28 @@ def switch_batch(engine):
...
         if self.state.max_epochs is None or self._is_done(self.state):
             # Create new state
-            if max_epochs is None:
-                max_epochs = 1
             if epoch_length is None:
                 epoch_length = self._get_data_length(data)
                 if epoch_length is not None and epoch_length < 1:
                     raise ValueError("Input data has zero size. Please provide non-empty data")
 
+            if epoch_length is not None:
```
sdesrozis, line 694: useless test

ignite/engine/engine.py
```diff
@@ -800,6 +818,8 @@ def _run_once_on_dataset(self) -> float:
                     if self.state.epoch_length is None:
                         # Define epoch length and stop the epoch
                         self.state.epoch_length = iter_counter
+                        if self.state.max_iters is None:
```
sdesrozis, line 821: I don't understand why `max_iters` is initialized if `None`

ignite/engine/engine.py
```diff
@@ -597,6 +598,7 @@ def run(
         self,
         data: Iterable,
         max_epochs: Optional[int] = None,
+        max_iters: Optional[int] = None,
```
sdesrozis, line 601: Need docstring as you said

sdesrozis:
@thescripted Thank you ! I left few comments.

ignite/engine/engine.py
thescripted: I'm unable to calculate max_iters if epoch_length is None before the engine is running (a few test throws an error):
https://github.com/pytorch/ignite/blob/e6ef77d7bbeb2bde5edee7282f34efb958a124c4/tests/ignite/engine/test_engine.py#L638-L659

In those case I let max_iters stay as None as default and defined it when StopIterations calculates the epoch_length

ignite/engine/engine.py
```diff
@@ -621,6 +623,9 @@ def run(
                 `len(data)`. If `data` is an iterator and `epoch_length` is not set, then it will be automatically
                 determined as the iteration on which data iterator raises `StopIteration`.
                 This argument should not change if run is resuming from a state.
+            max_iters (int, optional): Max Number of iterations to run for. By default, it will be calculated as
+                max_epochs * epoch_length. If `data` is an iterator and `max_iters` is not set, then it will be
+                automatically determined along with `epoch_length` on which data iterator raises `StopIteration`.
```
vfdev-5, line 628: Let's also mention that `max_iters` and `max_epochs` are mutually exclusive args and add this check inside `run` function.

ignite/engine/engine.py
```diff
@@ -673,6 +678,7 @@ def switch_batch(engine):
                         )
                     )
                 self.state.max_epochs = max_epochs
+                self.state.max_iters = max_epochs * self.state.epoch_length
```
vfdev-5, line 681: I think this can lead to an exception if user stopped and try to rerun the engine on an iterable without specifying `epoch_length`.

ignite/engine/engine.py
vfdev-5: @sdesrozis `epoch_length` can still be None here if we deal with unknown size iterators. The check is used.
However, i'd prefer to do something else. Let's do a sort of "lazy" assignment for `max_epochs`, `max_iters`.
User passes:
1) `max_epochs=N`, `max_iters=None` and whatever `epoch_length` => everything should work as previously for N epochs
2) `max_epochs=None`, `max_iters=None` and whatever `epoch_length` => everything should work as previously for 1 epoch
3) `max_epochs=N`, `max_iters=M` and whatever `epoch_length` => raise an exception
4) `max_epochs=None`, `max_iters=M` and `epoch_length=K` => we can setup directly `max_epochs=ceil(max_iters//epoch_length)` and stop on `max_iters` iteration.
4) `max_epochs=None`, `max_iters=M` and `epoch_length=None` => we keep `max_epochs=None` and iterate until:
a) `max_iters` is reached
b) `StopIterations` is raised and we can calculate the `epoch_length` -> `max_epochs=ceil(max_iters//epoch_length)`.

ignite/engine/engine.py
```diff
@@ -748,8 +766,11 @@ def _internal_run(self) -> State:
                 handlers_start_time = time.time()
                 if self.should_terminate:
                     self._fire_event(Events.TERMINATE)
-                else:
+                elif self.state.iteration % self.state.epoch_length == 0:
```
vfdev-5, line 769: This can lead to bug if `self.state.epoch_length` is None somehow here.
I would keep the previous code. Especially, as we set should terminate to True below
```python
                if self.state.max_iters is not None and self.state.iteration == self.state.max_iters:
                    self.should_terminate = True
                    break
```

ignite/engine/engine.py
vfdev-5: Let's do the assignement according to 5.b) case described above

vfdev-5 (commented):
@thescripted Thanks for working on the PR !
I left few comment on how to improve it

vfdev-5:
@thescripted could you also please resolve the conflict and update your branch to master.

thescripted:
On it. Sorry for the late reply I've put a bit on my plate with other projects. I'll go ahead and resolve these issues now.

vfdev-5 (commented):
Thanks @thescripted ! Looks good 

I still have few comments about the PR. For example, a special care is needed here
```python
while self.state.epoch < self.state.max_epochs
```
if `max_epochs` is None, right ?

thescripted:
Yes @vfdev-5, I didn't consider that. I believe a fix could be like so: 
```python 
while not self.should_terminate:  # type: ignore[operator]
    if self.state.max_epochs is not None and self.state.epoch >= self.state.max_epochs:
        break
``` 

I also think a condition to check the max_iters here is required.
https://github.com/pytorch/ignite/blob/acd2982dfc6ecbea347d1ff88d893f8acdd87a56/ignite/engine/engine.py#L559-L560

From what I can tell max_epochs will never be none, given where and how the method is used [Here](https://github.com/pytorch/ignite/blob/acd2982dfc6ecbea347d1ff88d893f8acdd87a56/ignite/engine/engine.py#L812-L821) and [Here](https://github.com/pytorch/ignite/blob/acd2982dfc6ecbea347d1ff88d893f8acdd87a56/ignite/engine/engine.py#L685
). I can just add another conditional checking if `state.iterations == state.max_iters` which may or may not be None.

If those are the right track (and any additional comments you have) I'll be happy to fix those up.

vfdev-5:
@thescripted good point about `_is_done`. So, I'm thinking that we can replace the code 
```python
while self.state.epoch < self.state.max_epochs and not self.should_terminate:
```
by
```python
def _is_done(state: State) -> bool: 
     # something like that
     c1 = state.max_iters and state.iterations == state.max_iters
     c2 = state.epoch_length and state.iteration == state.epoch_length * state.max_epochs     
     return c1 or c2

while not self._is_done(self.state) and not self.should_terminate:
```
What do you think ?

thescripted:
I believe the check for `epoch < max_epoch` is still necessary right? In the event that an epoch was terminated half-way through: 

https://github.com/pytorch/ignite/blob/acd2982dfc6ecbea347d1ff88d893f8acdd87a56/tests/ignite/engine/test_engine.py#L144-L159

vfdev-5:
Sorry for delay @thescripted and thanks for the update !

> I believe the check for epoch < max_epoch is still necessary right?

Checking iteration or epoch should be similar. However, maybe we have to improve `self._is_done` and check inequality instead of strict equality...

thescripted:
I appreciate all the time and effort you're providing! It's very helpful, thanks.

> Checking iteration or epoch should be similar. However, maybe we have to improve self._is_done and check inequality instead of strict equality...

I may have taken a naive approach here and simply changed the conditions to check for inequality in `self._is_done`. Is that what you had in mind, or am I missing something here?

ignite/engine/engine.py
```diff
@@ -670,16 +685,28 @@ def switch_batch(engine):
...
+                    max_epochs = 1
+            else:
+                if max_epochs is not None:
+                    raise ValueError(
+                        "Arguments max_iters and max_epochs are mutually exclusive."
+                        "Please provide only max_epochs or max_iters."
+                    )
+                else:
+                    if epoch_length is not None:
+                        max_epochs = math.ceil(max_iters / epoch_length)
```
vfdev-5, line 704:
```suggestion
                if epoch_length is not None:
                    max_epochs = math.ceil(max_iters / epoch_length)
```

tests/ignite/engine/test_engine.py
```diff
@@ -891,3 +891,42 @@ def switch_dataloader():
         trainer.set_data(data2)
 
     trainer.run(data1, max_epochs=10)
+
+
+def test_run_with_max_iters():
+    max_iters = 8
```
vfdev-5, line 897: Could you please add another test with max iters larger than 3 epochs for example, e.g. max_iters=73

vfdev-5 (commented):
@thescripted thanks a lot for the update ! Looks good. Just few nits and a minor test improvement and it will be good to land.

Roy-Kid:
@vfdev-5 Hi, why is this PR reverted so many times? The 0.5.2 still doesn't have `max_iter`, but master and nightly has it.

vfdev-5:
> @vfdev-5 Hi, why is this PR reverted so many times? The 0.5.2 still doesn't have `max_iter`, but master and nightly has it.

Mainly because of this : https://github.com/pytorch/ignite/issues/1521

Roy-Kid:
> > @vfdev-5 Hi, why is this PR reverted so many times? The 0.5.2 still doesn't have `max_iter`, but master and nightly has it.
> 
> Mainly because of this : #1521

lol, that's funny. I have implemented it with a handler but just found it in master. Thx!

vfdev-5:
Yes, it works with a handler, but it wont work properly if we try save/load and rerun the engine...

PR #1391, opened 2020-10-18
Activate mypy in ignite.metrics

gruebel:
Description:
That was some work, I hope I made everything correct. Sometimes I had to rearrange the code, so `mypy` understood the typing better.

### DIFF
--- a/ignite/metrics/accumulation.py
+++ b/ignite/metrics/accumulation.py
@@ -1,5 +1,5 @@
 import numbers
-from typing import Any, Callable, Optional, Union, cast
+from typing import Any, Callable, Tuple, Union, cast
 
 import torch
 
@@ -55,7 +55,7 @@ class VariableAccumulation(Metric):
     @reinit__is_reduced
     def reset(self) -> None:
         self.accumulator = torch.tensor(0.0, dtype=torch.float64, device=self._device)
-        self.num_examples = torch.tensor(0, dtype=torch.long, device=self._device)
+        self.num_examples = 0
 
     def _check_output_type(self, output: Union[Any, torch.Tensor, numbers.Number]) -> None:
         if not (isinstance(output, numbers.Number) or isinstance(output, torch.Tensor)):
@@ -78,8 +78,8 @@ class VariableAccumulation(Metric):
             self.num_examples += 1
 
     @sync_all_reduce("accumulator", "num_examples")
-    def compute(self) -> list:
-        return [self.accumulator, self.num_examples]
+    def compute(self) -> Tuple[torch.Tensor, int]:
+        return self.accumulator, self.num_examples
 
 
 class Average(VariableAccumulation):
@@ -132,10 +132,10 @@ class Average(VariableAccumulation):
         super(Average, self).__init__(op=_mean_op, output_transform=output_transform, device=device)
 
     @sync_all_reduce("accumulator", "num_examples")
-    def compute(self) -> Union[Any, torch.Tensor, numbers.Number]:
+    def compute(self) -> Union[torch.Tensor, numbers.Number]:
         if self.num_examples < 1:
             raise NotComputableError(
-                "{} must have at least one example before" " it can be computed.".format(self.__class__.__name__)
+                "{} must have at least one example before it can be computed.".format(self.__class__.__name__)
             )
 
         return self.accumulator / self.num_examples
@@ -172,7 +172,7 @@ class GeometricAverage(VariableAccumulation):
     def __init__(
         self, output_transform: Callable = lambda x: x, device: Union[str, torch.device] = torch.device("cpu")
     ):
-        def _geom_op(a: torch.Tensor, x: Union[Any, numbers.Number, torch.Tensor]) -> torch.Tensor:
+        def _geom_op(a: torch.Tensor, x: Union[numbers.Number, torch.Tensor]) -> torch.Tensor:
             if not isinstance(x, torch.Tensor):
                 x = torch.tensor(x)
             x = torch.log(x)
@@ -183,10 +183,15 @@ class GeometricAverage(VariableAccumulation):
         super(GeometricAverage, self).__init__(op=_geom_op, output_transform=output_transform, device=device)
 
     @sync_all_reduce("accumulator", "num_examples")
-    def compute(self) -> torch.Tensor:
+    def compute(self) -> Union[torch.Tensor, numbers.Number]:
         if self.num_examples < 1:
             raise NotComputableError(
-                "{} must have at least one example before" " it can be computed.".format(self.__class__.__name__)
+                "{} must have at least one example before it can be computed.".format(self.__class__.__name__)
             )
 
-        return torch.exp(self.accumulator / self.num_examples)
+        tensor = torch.exp(self.accumulator / self.num_examples)
+
+        if tensor.numel() == 1:
+            return cast(numbers.Number, tensor.item())
+
+        return tensor
--- a/ignite/metrics/accumulation.py
+++ b/ignite/metrics/accumulation.py
@@ -1,5 +1,5 @@
 import numbers
-from typing import Any, Callable, Optional, Union
+from typing import Any, Callable, Optional, Union, cast
 
 import torch
 
@@ -47,8 +47,8 @@ class VariableAccumulation(Metric):
     ):
         if not callable(op):
             raise TypeError("Argument op should be a callable, but given {}".format(type(op)))
-        self.accumulator = None  # type: Optional[torch.Tensor]
-        self.num_examples = None  # type: Optional[torch.Tensor]
+        self.accumulator = 0.0  # type: Union[float, torch.Tensor]
+        self.num_examples = 0
         self._op = op
 
         super(VariableAccumulation, self).__init__(output_transform=output_transform, device=device)
@@ -56,7 +56,7 @@ class VariableAccumulation(Metric):
     @reinit__is_reduced
     def reset(self) -> None:
         self.accumulator = torch.tensor(0.0, dtype=torch.float64, device=self._device)
-        self.num_examples = torch.tensor(0, dtype=torch.long, device=self._device)
+        self.num_examples = 0
 
     def _check_output_type(self, output: Union[Any, torch.Tensor, numbers.Number]) -> None:
         if not (isinstance(output, numbers.Number) or isinstance(output, torch.Tensor)):
@@ -65,11 +65,6 @@ class VariableAccumulation(Metric):
     @reinit__is_reduced
     def update(self, output: Union[Any, torch.Tensor, numbers.Number]) -> None:
         self._check_output_type(output)
-
-        if self.num_examples is None:
-            raise RuntimeError(
-                "Internal error, self.num_examples is None. Please, file an issue if you encounter this error."
-            )
 
         if isinstance(output, torch.Tensor):
             output = output.detach()
@@ -139,11 +134,6 @@ class Average(VariableAccumulation):
 
     @sync_all_reduce("accumulator", "num_examples")
     def compute(self) -> Union[Any, torch.Tensor, numbers.Number]:
-        if self.num_examples is None:
-            raise RuntimeError(
-                "Internal error, self.num_examples is None. Please, file an issue if you encounter this error."
-            )
-
         if self.num_examples < 1:
             raise NotComputableError(
                 "{} must have at least one example before" " it can be computed.".format(self.__class__.__name__)
@@ -195,14 +185,9 @@ class GeometricAverage(VariableAccumulation):
 
     @sync_all_reduce("accumulator", "num_examples")
     def compute(self) -> torch.Tensor:
-        if self.num_examples is None:
-            raise RuntimeError(
-                "Internal error, self.num_examples is None. Please, file an issue if you encounter this error."
-            )
-
         if self.num_examples < 1:
             raise NotComputableError(
                 "{} must have at least one example before" " it can be computed.".format(self.__class__.__name__)
             )
 
-        return torch.exp(self.accumulator / self.num_examples)
+        return torch.exp(cast(torch.Tensor, self.accumulator) / self.num_examples)
--- a/ignite/metrics/accumulation.py
+++ b/ignite/metrics/accumulation.py
@@ -47,8 +47,7 @@ class VariableAccumulation(Metric):
     ):
         if not callable(op):
             raise TypeError("Argument op should be a callable, but given {}".format(type(op)))
-        self.accumulator = 0.0  # type: Union[float, torch.Tensor]
-        self.num_examples = 0
+
         self._op = op
 
         super(VariableAccumulation, self).__init__(output_transform=output_transform, device=device)
@@ -56,7 +55,7 @@ class VariableAccumulation(Metric):
     @reinit__is_reduced
     def reset(self) -> None:
         self.accumulator = torch.tensor(0.0, dtype=torch.float64, device=self._device)
-        self.num_examples = 0
+        self.num_examples = torch.tensor(0, dtype=torch.long, device=self._device)
 
     def _check_output_type(self, output: Union[Any, torch.Tensor, numbers.Number]) -> None:
         if not (isinstance(output, numbers.Number) or isinstance(output, torch.Tensor)):
@@ -190,4 +189,4 @@ class GeometricAverage(VariableAccumulation):
                 "{} must have at least one example before" " it can be computed.".format(self.__class__.__name__)
             )
 
-        return torch.exp(cast(torch.Tensor, self.accumulator) / self.num_examples)
+        return torch.exp(self.accumulator / self.num_examples)
--- a/ignite/metrics/accumulation.py
+++ b/ignite/metrics/accumulation.py
@@ -1,5 +1,5 @@
 import numbers
-from typing import Any, Callable, Union
+from typing import Any, Callable, Optional, Union
 
 import torch
 
@@ -47,8 +47,8 @@ class VariableAccumulation(Metric):
     ):
         if not callable(op):
             raise TypeError("Argument op should be a callable, but given {}".format(type(op)))
-        self.accumulator = None
-        self.num_examples = None
+        self.accumulator = None  # type: Optional[torch.Tensor]
+        self.num_examples = None  # type: Optional[torch.Tensor]
         self._op = op
 
         super(VariableAccumulation, self).__init__(output_transform=output_transform, device=device)
@@ -66,6 +66,11 @@ class VariableAccumulation(Metric):
     def update(self, output: Union[Any, torch.Tensor, numbers.Number]) -> None:
         self._check_output_type(output)
 
+        if self.num_examples is None:
+            raise RuntimeError(
+                "Internal error, self.num_examples is None. Please, file an issue if you encounter this error."
+            )
+
         if isinstance(output, torch.Tensor):
             output = output.detach()
             if output.device != self._device:
@@ -73,7 +78,7 @@ class VariableAccumulation(Metric):
 
         self.accumulator = self._op(self.accumulator, output)
 
-        if hasattr(output, "shape"):
+        if isinstance(output, torch.Tensor):
             self.num_examples += output.shape[0] if len(output.shape) > 1 else 1
         else:
             self.num_examples += 1
@@ -125,8 +130,8 @@ class Average(VariableAccumulation):
     def __init__(
         self, output_transform: Callable = lambda x: x, device: Union[str, torch.device] = torch.device("cpu")
     ):
-        def _mean_op(a, x):
-            if isinstance(x, torch.Tensor) and x.ndim > 1:
+        def _mean_op(a: Union[float, torch.Tensor], x: Union[float, torch.Tensor]) -> Union[float, torch.Tensor]:
+            if isinstance(x, torch.Tensor) and x.ndim > 1:  # type: ignore[attr-defined]
                 x = x.sum(dim=0)
             return a + x
 
@@ -134,6 +139,11 @@ class Average(VariableAccumulation):
 
     @sync_all_reduce("accumulator", "num_examples")
     def compute(self) -> Union[Any, torch.Tensor, numbers.Number]:
+        if self.num_examples is None:
+            raise RuntimeError(
+                "Internal error, self.num_examples is None. Please, file an issue if you encounter this error."
+            )
+
         if self.num_examples < 1:
             raise NotComputableError(
                 "{} must have at least one example before" " it can be computed.".format(self.__class__.__name__)
@@ -177,7 +187,7 @@ class GeometricAverage(VariableAccumulation):
             if not isinstance(x, torch.Tensor):
                 x = torch.tensor(x)
             x = torch.log(x)
-            if x.ndim > 1:
+            if x.ndim > 1:  # type: ignore[attr-defined]
                 x = x.sum(dim=0)
             return a + x
 
@@ -185,6 +195,11 @@ class GeometricAverage(VariableAccumulation):
 
     @sync_all_reduce("accumulator", "num_examples")
     def compute(self) -> torch.Tensor:
+        if self.num_examples is None:
+            raise RuntimeError(
+                "Internal error, self.num_examples is None. Please, file an issue if you encounter this error."
+            )
+
         if self.num_examples < 1:
             raise NotComputableError(
                 "{} must have at least one example before" " it can be computed.".format(self.__class__.__name__)
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -1,4 +1,4 @@
-from typing import Callable, Optional, Sequence, Tuple, Union, cast
+from typing import Callable, Optional, Sequence, Tuple, Union
 
 import torch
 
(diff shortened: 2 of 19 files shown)


ignite/metrics/accumulation.py
```diff
@@ -66,14 +66,19 @@ def _check_output_type(self, output: Union[Any, torch.Tensor, numbers.Number]) -
     def update(self, output: Union[Any, torch.Tensor, numbers.Number]) -> None:
         self._check_output_type(output)
 
+        if self.num_examples is None:
```
vfdev-5, line 69: I understand that here we bump in limits of the mypy to understand that reset is called in the constructor by the base class. However, I'd expect the checks for both vars: `num_examples` and `accumulator`. Looks a bit strange.

ignite/metrics/accumulation.py
```diff
@@ -125,15 +130,20 @@ class Average(VariableAccumulation):
...
+        def _mean_op(a: Union[float, torch.Tensor], x: Union[float, torch.Tensor]) -> Union[float, torch.Tensor]:
+            if isinstance(x, torch.Tensor) and x.ndim > 1:  # type: ignore[attr-defined]
                 x = x.sum(dim=0)
             return a + x
 
         super(Average, self).__init__(op=_mean_op, output_transform=output_transform, device=device)
 
     @sync_all_reduce("accumulator", "num_examples")
     def compute(self) -> Union[Any, torch.Tensor, numbers.Number]:
+        if self.num_examples is None:
```
vfdev-5, line 142: I'd love to find a way to remove those mypy-happy checks :)

EDIT: Let's do the same trick as for Accuracy. Let's set `num_examples` as a scalar = 0, instead of torch tensor.

ignite/metrics/confusion_matrix.py
```diff
@@ -124,13 +128,12 @@ def compute(self) -> torch.Tensor:
...
     def normalize(matrix: torch.Tensor, average: str) -> torch.Tensor:
-        if average not in ("recall", "precision"):
-            raise ValueError("Argument average one of 'samples', 'recall', 'precision'")
-
         if average == "recall":
             return matrix / (matrix.sum(dim=1).unsqueeze(1) + 1e-15)
         elif average == "precision":
             return matrix / (matrix.sum(dim=0) + 1e-15)
+        else:
+            raise ValueError("Argument average one of 'samples', 'recall', 'precision'")
```
vfdev-5, line 136:
```suggestion
            raise ValueError("Argument average should be one of 'samples', 'recall', 'precision'")
```

ignite/metrics/metric.py
```diff
@@ -28,21 +31,21 @@ class MetricUsage:
             :meth:`~ignite.metrics.Metric.iteration_completed`.
     """
 
-    def __init__(self, started, completed, iteration_completed):
+    def __init__(self, started: Events, completed: Events, iteration_completed: CallableEventWithFilter) -> None:
```
vfdev-5, line 34: I'm a bit surprised with `iteration_completed: CallableEventWithFilter`, why it can't be `iteration_completed: Events`  ?

vfdev-5 (commented):
Thanks a lot @gruebel ! Great work on the typing and mypy.
I left few comments. I have to understand why for mypy we have to define arguments in the constructor (which calls `reset`) as inside `reset` to avoid temporary `None` assignement.

ignite/metrics/accumulation.py
gruebel: the reason, why `self.num_examples` has to be checked here and not `self.accumulator` is because of those lines
https://github.com/pytorch/ignite/blob/c8ad7dc0dbfc6b62251336b57fc504b648410ea7/ignite/metrics/accumulation.py#L81-L84
`mypy` complains, that `None` and `int` have an supported opperand for `+`.

ignite/metrics/metric.py
gruebel: It's actually because of this lines here
https://github.com/pytorch/ignite/blob/c8ad7dc0dbfc6b62251336b57fc504b648410ea7/ignite/metrics/metric.py#L117-L122
This usage suggests that the used parameter for `iteration_completed` is not of type `Events` but of `CallableEventWithFilter `.

ignite/metrics/accumulation.py
gruebel: Interesting, after I assigned scalar 0 to `num_examples`, I got a lot of `mypy` findings for `self.accumulator`, which also makes sense, but I don't understand, why they didn't pop up earlier...

vfdev-5: Yeah, the wisdom of `mypy` is unattainable for us humans :)

ignite/metrics/metric.py
vfdev-5: OK, I see, true that `Events.ITERATION_COMPLETED(*args, **kwargs)` returns an instance of `CallableEventWithFilter`:
https://github.com/pytorch/ignite/blob/11f6d6212cba56e6f7e20098a6e029421c7ac937/ignite/engine/events.py#L96

I wonder if this should not be replaced by something like 
```python
cast(Callable, type(self))(self.value, event_filter, self.name)
```
as `self` can be a derived class. On the other hand, maybe this does not worth a change...

gruebel: This results in an error... it looks like that the constructor of `Events` calls the `__call__` method of `Enum`, because `EventEnum(CallableEventWithFilter, Enum)` multi inherits. But somehow I don't get, why `Enum.__call__` is called and not `CallableEventWithFilter.__call__`?!
```
>       return type(self)(self.value, event_filter, self.name)
E       TypeError: __call__() takes from 2 to 3 positional arguments but 4 were given
```

vfdev-5: This is a bit strange. Let's keep it as you did.

ignite/metrics/accumulation.py
```diff
@@ -47,16 +47,16 @@ def __init__(
     ):
         if not callable(op):
             raise TypeError("Argument op should be a callable, but given {}".format(type(op)))
-        self.accumulator = None
-        self.num_examples = None
+        self.accumulator = 0.0  # type: Union[float, torch.Tensor]
```
vfdev-5, line 50: @gruebel could you please confirm if it is the case for you too, if we remove those initialization lines, mypy is still happy ?

ignite/metrics/accuracy.py
```diff
@@ -167,7 +167,7 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
         self._num_examples += correct.shape[0]
 
     @sync_all_reduce("_num_examples", "_num_correct")
-    def compute(self) -> torch.Tensor:
+    def compute(self) -> float:
         if self._num_examples == 0:
             raise NotComputableError("Accuracy must have at least one example before it can be computed.")
-        return self._num_correct.item() / self._num_examples
+        return cast(torch.Tensor, self._num_correct).item() / self._num_examples
```
vfdev-5, line 173: Looks like here it can be done in the same way. We can remove initializations inside the constructor and even remove this cast.

gruebel: Sorry, didn't see your comment, but now I added your suggestion 😄

ignite/metrics/frequency.py
```diff
@@ -70,10 +70,10 @@ def compute(self):
         # Returns the average processed objects per second across all workers
         return self._n / self._elapsed * time_divisor
 
-    def completed(self, engine, name):
+    def completed(self, engine: Engine, name: str) -> None:
         engine.state.metrics[name] = int(self.compute())
 
-    def attach(self, engine, name, event_name=Events.ITERATION_COMPLETED):
+    def attach(self, engine: Engine, name: str, event_name: Any = Events.ITERATION_COMPLETED) -> None:
```
vfdev-5, line 76:
```suggestion
    def attach(self, engine: Engine, name: str, event_name: Events = Events.ITERATION_COMPLETED) -> None:
```

ignite/metrics/metric.py
```diff
@@ -191,12 +194,14 @@ def compute(self):
...
     # public class attribute
-    required_output_keys = ("y_pred", "y")
+    required_output_keys = ("y_pred", "y")  # type: Optional[Tuple]
     # for backward compatibility
     _required_output_keys = required_output_keys
 
     def __init__(
-        self, output_transform: Callable = lambda x: x, device: Union[str, torch.device] = torch.device("cpu"),
+        self,
+        output_transform: Optional[Callable] = lambda x: x,
```
vfdev-5, line 203: Maybe, better to have 
```suggestion
        output_transform: Callable = lambda x: x,
```
?

ignite/metrics/metric.py
```diff
@@ -275,6 +280,11 @@ def iteration_completed(self, engine: Engine) -> None:
             engine (Engine): the engine to which the metric must be attached
         """
 
+        if self._output_transform is None:
```
vfdev-5, line 283: As `self._output_transform` is not optional, this check is unnecessary ?

ignite/metrics/precision.py
```diff
@@ -30,8 +30,8 @@ def __init__(
                 )
 
         self._average = average
-        self._true_positives = None
-        self._positives = None
+        self._true_positives = 0  # type: Union[int, torch.Tensor]
```
vfdev-5, line 33: Let's remove these lines as for others

ignite/metrics/precision.py
```diff
@@ -41,32 +41,33 @@ def __init__(
...
         if is_scalar and self._positives == 0:
             raise NotComputableError(
                 "{} must have at least one example before it can be computed.".format(self.__class__.__name__)
             )
 
         if not (self._type == "multilabel" and not self._average):
-            if not self._is_reduced:
-                self._true_positives = idist.all_reduce(self._true_positives)
-                self._positives = idist.all_reduce(self._positives)
+            if not self._is_reduced:  # type: ignore[has-type] # mypy cann't determine type
```
vfdev-5, line 62: I had that this error from time to time. Let's remove it too and see if we can not do otherwise.

ignite/metrics/precision.py
```diff
@@ -41,32 +41,33 @@ def __init__(
...
+                self._true_positives = idist.all_reduce(self._true_positives)  # type: ignore[arg-type, assignment]
```
vfdev-5, line 63: Maybe, mypy is right here that we should not reuse the same attribute for the output as we can just define a local vars initialized as `self._positives` or `self._true_positives` ?

vfdev-5 (commented):
@gruebel thanks a lot for working on this PR !
Reviewing the PR, I see that we previously mixed up with `compute` output type. I think the main point is to have it as float everywhere it is possible and torch.Tensor otherwise or other structures if intended.
I left some comments in the code to update the current code. I didn't manage to review all 17 files. Review others later.

ignite/metrics/accumulation.py
```diff
@@ -73,7 +72,7 @@ def update(self, output: Union[Any, torch.Tensor, numbers.Number]) -> None:
 
         self.accumulator = self._op(self.accumulator, output)
 
-        if hasattr(output, "shape"):
+        if isinstance(output, torch.Tensor):
             self.num_examples += output.shape[0] if len(output.shape) > 1 else 1
         else:
             self.num_examples += 1
```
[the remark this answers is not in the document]
gruebel: Sure, I changed it to
```python
    def compute(self) -> Tuple[torch.Tensor, int]:
        return (self.accumulator, self.num_examples)
```
because it makes more sense here.

ignite/metrics/accumulation.py
```diff
@@ -177,7 +176,7 @@ def _geom_op(a: torch.Tensor, x: Union[Any, numbers.Number, torch.Tensor]) -> to
             if not isinstance(x, torch.Tensor):
```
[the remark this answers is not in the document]
gruebel: Sure, I changed it to `Union[torch.Tensor, numbers.Number]`, but code wise it can only be `torch.Tensor` at he moment.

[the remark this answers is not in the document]
vfdev-5: Yes, that's correct. We can just check it like that : 
```python
if tensor.numel() == 1:
    return cast(Number, tensor.item())
```

ignite/metrics/confusion_matrix.py
```diff
@@ -171,13 +169,13 @@ def IoU(cm: ConfusionMatrix, ignore_index: Optional[int] = None) -> MetricsLambd
     iou = cm.diag() / (cm.sum(dim=1) + cm.sum(dim=0) - cm.diag() + 1e-15)
     if ignore_index is not None:
 
-        def ignore_index_fn(iou_vector):
-            if ignore_index >= len(iou_vector):
+        def ignore_index_fn(iou_vector: torch.Tensor) -> torch.Tensor:
```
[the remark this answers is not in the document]
gruebel: sadly this doesn't work, I tried it already before :(
how do you like this variation?
```python
    if ignore_index is not None:
        ignore_idx = ignore_index  # used due to typing issues

        def ignore_index_fn(iou_vector: torch.Tensor) -> torch.Tensor:
            if ignore_idx >= len(iou_vector):
                raise ValueError(
                    "ignore_index {} is larger than the length of IoU vector {}".format(ignore_idx, len(iou_vector))
                )
            indices = list(range(len(iou_vector)))
            indices.remove(ignore_idx)
            return iou_vector[indices]

        return MetricsLambda(ignore_index_fn, iou)
```

ignite/metrics/confusion_matrix.py
```diff
@@ -284,12 +282,12 @@ def DiceCoefficient(cm: ConfusionMatrix, ignore_index: Optional[int] = None) ->
     if ignore_index is not None:
 
         def ignore_index_fn(dice_vector: torch.Tensor) -> torch.Tensor:
-            if ignore_index >= len(dice_vector):
+            if cast(int, ignore_index) >= len(dice_vector):
```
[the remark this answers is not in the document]
gruebel: I can do the same as above.

ignite/metrics/frequency.py
gruebel: this is actually a bit more complicated, because `Frequency` inherits `Metric` the `attach` method should have following typing
```python
def attach(self, engine: Engine, name: str, usage: Union[str, MetricUsage] = EpochWise()) -> None:
``` 
when I change it `Events`, then I have to ignore the typing here.

vfdev-5: Oh, I see. Seems like `Frequency.attach` signature was not updated when `MetricUsage` was introduced. Thanks for catching !

ignite/metrics/confusion_matrix.py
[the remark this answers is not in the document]
vfdev-5: I'm trying to understand why mypy still think after `if ignore_index is not None:` that `ignore_index` can be None. Very strange...

ignite/metrics/metric.py
gruebel: The reason behind the `Optional` here is because of the `RunningAverage` class, where the `output_transform` can be None
https://github.com/pytorch/ignite/blob/bcac5c52f5a54aca9f874ab663ed217cfe107a65/ignite/metrics/running_average.py#L49-L56

vfdev-5: That's true :) Oh, what a badly designed library :)

vfdev-5: I think the best we can do is to put ignores what those bad designs inside Frequency.attach and RunningAverage.__init__ and create an issue to fix them. And we annotate correctly the args where we intend the correct behaviour like `Metric.__init__(output_transform: Callable = lambda x: x)`

vfdev-5: Thanks a lot for checking all that @gruebel !

ignite/metrics/precision.py
gruebel: actually it is a problem related to the `numbers.Number` used in `all_reduce`
```python
def all_reduce(tensor: Union[torch.Tensor, Number], op: str = "SUM") -> Union[torch.Tensor, Number]:
```
there is still an open issue in `mypy` https://github.com/python/mypy/issues/3186 which suggests the usage of `float` because it would allow both `int` and `float`.
What do you think about replacing the `Number` type with `float` in the whole project? I could create a new issue to tackle that problem.

ignite/metrics/confusion_matrix.py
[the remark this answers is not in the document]
vfdev-5: @gruebel seems like there is no other way to tell mypy that `ignore_index` is not None anymore here. Let's do as you suggest with `ignore_idx = ignore_index  #  type: int`

ignite/metrics/confusion_matrix.py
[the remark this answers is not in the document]
vfdev-5: Yes, please

ignite/metrics/precision.py
vfdev-5: @gruebel sounds good about replacing `Number` with `float` for type annotations (if I understood correctly). Runtime code can still keep Number, right? 
So, yes, please open an issue for that.

ignite/metrics/confusion_matrix.py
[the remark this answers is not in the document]
gruebel: Yeah, I already spent some time to try figuring it out, but this is out of my ability of understanding `mypy` 😄

ignite/metrics/metric.py
gruebel: Sure thing 💪 for me it is just hard to decide, if the code is intended to be like this or it is really a bug.

ignite/metrics/accumulation.py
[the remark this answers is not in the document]
gruebel: I hope, I changed the code, how you expected it.

ignite/metrics/precision.py
gruebel: Yeah correct it is just for type hinting 😄  and I created a new issue for it.

ignite/metrics/accumulation.py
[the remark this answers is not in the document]
vfdev-5: Thanks, looks good !

ignite/metrics/confusion_matrix.py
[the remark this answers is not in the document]
vfdev-5: I also asked my collegues on that and there are various opinions from bug in mypy to "all the assignment or usage of a variable need to be the same type is the general rule." Let's keep it as you did. Another way to cheat mypy is to set `ignore_index` as arg into `ignore_index_fn`:
```python
if ignore_index is not None:

        def ignore_index_fn(iou_vector: torch.Tensor, ignore_index: int = ignore_index) -> torch.Tensor:
```

ignite/metrics/confusion_matrix.py
```diff
@@ -170,14 +168,15 @@ def IoU(cm: ConfusionMatrix, ignore_index: Optional[int] = None) -> MetricsLambd
     cm = cm.type(torch.DoubleTensor)
     iou = cm.diag() / (cm.sum(dim=1) + cm.sum(dim=0) - cm.diag() + 1e-15)
     if ignore_index is not None:
+        ignore_idx = ignore_index  # type: int
```
vfdev-5, line 171: nit: I'd add a  omment to make it really obvious, saying that this is required to satisfy mypy

ignite/metrics/frequency.py
```diff
@@ -70,10 +66,11 @@ def compute(self):
         # Returns the average processed objects per second across all workers
         return self._n / self._elapsed * time_divisor
 
-    def completed(self, engine, name):
+    def completed(self, engine: Engine, name: str) -> None:
         engine.state.metrics[name] = int(self.compute())
 
-    def attach(self, engine, name, event_name=Events.ITERATION_COMPLETED):
+    # TODO: see issue https://github.com/pytorch/ignite/issues/1405
+    def attach(self, engine: Engine, name: str, usage: Union[str, MetricUsage] = EpochWise()) -> None:
```
vfdev-5, line 73: I was thinking to keep the signature as it was : `attach(self, engine, name, event_name=Events.ITERATION_COMPLETED)` and an ignore mark for mypy.
Otherwise, `engine.add_event_handler(usage, self.completed, name)` looks wrong.

ignite/metrics/precision.py
```diff
@@ -30,43 +30,43 @@ def __init__(
                 )
 
         self._average = average
-        self._true_positives = None
-        self._positives = None
         self.eps = 1e-20
+        self._is_reduced = False
```
vfdev-5, line 34: Normally, `self._is_reduced = False` is called in `Metric.__init__`, so let's remove it from here.

ignite/metrics/precision.py
vfdev-5: Perfect, thanks!

ignite/metrics/recall.py
```diff
@@ -115,8 +115,14 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
 
         if self._type == "multilabel":
             if not self._average:
-                self._true_positives = torch.cat([self._true_positives, true_positives], dim=0)
-                self._positives = torch.cat([self._positives, actual_positives], dim=0)
+                self._true_positives = torch.cat(
```
vfdev-5, line 118: Here it should be almost the same as for precision (L180-L181), right ?
```python
                self._true_positives = torch.cat([cast(torch.Tensor, self._true_positives), true_positives], dim=0)
                self._positives = torch.cat([cast(torch.Tensor, self._positives), all_positives], dim=0)
```

ignite/metrics/ssim.py
```diff
@@ -85,11 +85,13 @@ def __init__(
 
     @reinit__is_reduced
     def reset(self) -> None:
-        self._sum_of_batchwise_ssim = 0.0  # Not a tensor because batch size is not known in advance.
+        self._sum_of_batchwise_ssim = (
```
vfdev-5, line 88: Now, `self._sum_of_batchwise_ssim` is defined as a tuple, right ? Can't we keep it as float ?

vfdev-5 (commented):
Thanks a lot the update Anton ! I think we are almost on the finish line. I left some comments to tweak few things and it is good to go.

ignite/metrics/frequency.py
gruebel: ok, gotcha 👍

ignite/metrics/precision.py
gruebel: if I remove it, then I get a `mypy` issue and have to ignore it like I did before
```python
if not self._is_reduced:  # type: ignore[has-type] # mypy can't determine type
```

ignite/metrics/recall.py
gruebel: yeah, you are right, but it is so strange. My IDE is totally fine with using `cast`, but `mypy` still complains with
```
error: Cannot determine type of '_true_positives'  [has-type]
error: Cannot determine type of '_positives'  [has-type]
```
I even tried the newer `mypy` version, still same error 😢

ignite/metrics/precision.py
vfdev-5: Yeah, I do not understand why it looses the type. I checked with `reveal_type(self._is_reduced)` inside `_BaseClassification` and `Metric` and it is bool. But in `Precision` it becomes `Any`...

ignite/metrics/ssim.py
gruebel: yeah, it is still float, but `black` just formatted it. To be a single value tuple it needs to be written with a comma
```python
self._sum_of_batchwise_ssim = (0.0, )
```
I will move the comment just above the attribute assignment

vfdev-5: I mean, previously it was a plain float, not tuple and now it becomes a tuple with a single float:
```diff
- self._sum_of_batchwise_ssim = 0.0
+ self._sum_of_batchwise_ssim = (0.0, )
```
It is not intended, right ?

gruebel: sorry, what I meant
`0.0 = (0.0) != (0.0, )`
So it is still a float and not a tuple 😄

ignite/metrics/precision.py
gruebel: what do you prefer? Define it in the `__init__` or add the `ignore`?

ignite/metrics/ssim.py
vfdev-5: Thanks ! Never got that `0.0 = (0.0)` :)

ignite/metrics/precision.py
vfdev-5: OK, let's ignore it for this PR and fix if possible after.

ignite/metrics/ssim.py
gruebel: I have seen it many times, when I started to use `black`. I usually used that syntax for multiline strings.

ignite/metrics/precision.py
vfdev-5: I think I found how to fix it:
```
            if not self._is_reduced:
                self._true_positives = idist.all_reduce(self._true_positives)  # type: ignore[arg-type, assignment]
                self._positives = idist.all_reduce(self._positives)  # type: ignore[arg-type, assignment]
                self._is_reduced = True  # type: bool

```

gruebel: you are right and it works, but I have no idea, why this fixes this problem?!

vfdev-5 (approved):
Thanks a lot, Anton !
Really huge work was done here ! LGTM

PR #1398, opened 2020-10-23
Added new time profiler `HandlersTimeProfiler` which allows per handler time profiling

### LINKED ISSUE #1346: Improve or New time profiling utility
## 🚀 Feature

Currently, ignite provides only a basic time profiling tool : https://pytorch.org/ignite/master/contrib/handlers.html#ignite.contrib.handlers.time_profilers.BasicTimeProfiler

Which can print something like 
```
----------------------------------------------------
| Time profiling stats (in seconds):                 |
 ----------------------------------------------------
total  |  min/index  |  max/index  |  mean  |  std

Processing function:
157.46292 | 0.01452/1501 | 0.26905/0 | 0.07730 | 0.01258

Dataflow:
6.11384 | 0.00008/1935 | 0.28461/1551 | 0.00300 | 0.02693

Event handlers:
2.82721

- Events.STARTED: []
0.00000

- Events.EPOCH_STARTED: []
0.00006 | 0.00000/0 | 0.00000/17 | 0.00000 | 0.00000

- Events.ITERATION_STARTED: ['PiecewiseLinear']
0.03482 | 0.00001/188 | 0.00018/679 | 0.00002 | 0.00001

- Events.ITERATION_COMPLETED: ['TerminateOnNan']
0.20037 | 0.00006/866 | 0.00089/1943 | 0.00010 | 0.00003

- Events.EPOCH_COMPLETED: ['empty_cuda_cache', 'training.<locals>.log_elapsed_time', ]
2.57860 | 0.11529/0 | 0.14977/13 | 0.12893 | 0.00790

- Events.COMPLETED: []
not yet triggered
```

What would be nice to have:
- more compact representation (maybe inspire from https://pytorch.org/docs/stable/autograd.html#torch.autograd.profiler.profile)
- profile handlers and not events. As we are looking for bottlenecks it would be helpful to know which functions take time and not on which event.
- better representation of data prep time vs training time

harsh8398:
Fixes #1346 

Description:

- Added HandlersTimeProfiler which allows per handler time profiling
- Added test cases and updated docs

Output of the `print_results()` is as follows:
```
-----------------------------------------  -----------------------  --------------  --------------  --------------  --------------  --------------
Handler                                    Event Name                     Total(s)      Min(s)/IDX      Max(s)/IDX         Mean(s)          Std(s)
-----------------------------------------  -----------------------  --------------  --------------  --------------  --------------  --------------
run.<locals>.log_training_results          EPOCH_COMPLETED                 9.69906       None/None       None/None            None            None
run.<locals>.log_validation_results        EPOCH_COMPLETED                 1.27098       None/None       None/None            None            None
run.<locals>.log_time                      EPOCH_COMPLETED                 0.00023       None/None       None/None            None            None
run.<locals>.log_intermediate_results      EPOCH_COMPLETED           not triggered       None/None       None/None            None            None
run.<locals>.log_training_loss             ITERATION_COMPLETED             0.02946       1e-05/874     0.00028/519           3e-05           7e-05
run.<locals>.log_time                      COMPLETED                 not triggered       None/None       None/None            None            None
-----------------------------------------  -----------------------  --------------  --------------  --------------  --------------  --------------
Total                                                                     10.99972
-----------------------------------------  -----------------------  --------------  --------------  --------------  --------------  --------------
Processing took total 5.67398s [min/index: 0.00443s/937, max/index: 0.00784s/0, mean: 0.00605s, std: 0.00035s]
Dataflow took total 8.15311s [min/index: 0.00538s/936, max/index: 0.01123s/937, mean: 0.00869s, std: 0.00112s]
```

Check list:

I have trimmed the output in the docstrings and doc for the table to make it not break lines and look pretty.


ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,395 @@ def odict_to_str(d: Mapping):
...
+    def get_results(self):
+        """
+        Method to fetch the aggregated profiler results after the engine is run
+
+        .. code-block:: python
+
+            results = profiler.get_results()
+
+        """
+        total_eh_time = round(
```
vfdev-5, line 642: This black formatting is ugly :) could you please split it into several expressions.

ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,395 @@ def odict_to_str(d: Mapping):
...
+            Events.EPOCH_COMPLETED,
+            Events.ITERATION_STARTED,
+            Events.ITERATION_COMPLETED,
+            Events.GET_BATCH_STARTED,
+            Events.GET_BATCH_COMPLETED,
+            Events.COMPLETED,
+        ]
+
+        # Events which are triggered only once
+        self._global_events = [Events.STARTED, Events.COMPLETED]
```
vfdev-5, line 492: I'm not sure to understand why we need to split by type of event like that : global, epoch, iteration. Can't we treat all events in the same way as we just need to measure the time passed inside a handler.

ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,395 @@ def odict_to_str(d: Mapping):
...
+
+    def _timeit_epoch_events(self, handler: Callable, engine: Engine, event: Events, *args, **kwargs):
+        # wrapper for profiling handlers attached to epoch events
+        self._event_handlers_timer.reset()
+        func, args, kwargs = handler
+        func(*args, **kwargs)
+        e = engine.state.epoch - 1
+        t = self._event_handlers_timer.value()
+        hname = hname = self._get_callable_name(func)
+        self.event_handlers_times[event][hname][e] = t
```
vfdev-5, line 525: Maybe, instead of adding to a slot we can append to a list. There is another point, a bit unrelated to this handler but which may impact the behaviour. For example, if user adds a handler attached to a filtered event like `Events.EPOCH_COMPLETED(every=3)`. Currently, this `every=3` filtering is translated to the handler: decorated handler is executed, falls into condition that we should or not execute user's original handler. So, for example, my handler takes in average 1s to execute, the times we'll have then: 
- epoch 1, T ~ 0s (time to filter)
- epoch 2, T ~ 0s (time to filter)
- epoch 3, T ~ 1s (time to filter + original handler)
- epoch 4, T ~ 0s (time to filter)
...

This is something to check and think about. Maybe, we have to add a threshold and do not append times smaller then this threshold.

ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,395 @@ def odict_to_str(d: Mapping):
...
+                h.__qualname__ if hasattr(h, "__qualname__") else h.__class__.__name__
+                for (h, _, _) in engine._event_handlers[e]
+                if not self._is_internal_handler(h)
+            ]
+            for e in self._events
+        }
+
+        self._reset(self.max_epochs, self.total_num_iters)
+
+        self._orig_event_handlers = engine._event_handlers.copy()
```
vfdev-5, line 595: I'm afraid of copying any type objects like that and not sure if it would work as expected. As event handlers can contain function and its args/kwargs where user can pass everything he/she wants.

vfdev-5 (commented):
Thanks for the PR @harsh8398 ! 
This is not as simple as I thought, but I'm sure we can make it work after several iterations.

ignite/contrib/handlers/time_profilers.py
harsh8398: I did that mostly because we don't know which handler will get triggered how many times. So this split allows me to statically initialize torch.tensor for every event measurements. We can use list and remove this split by type of event. But I was worried about python's list performance and memory because handlers can get triggered many times depending on number of iterations or trigger conditions.

Anyway, currently I am removing this splitting and making it more general to support custom events as well. But I believe we would still like to use tensor for storing these measurements. Let me know if you have any idea or suggestion on that.

ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,395 @@ def odict_to_str(d: Mapping):
...
+
+    def _detach_profiler_handlers(self, engine):
+        # reverts handlers to original handlers
+        engine._event_handlers = self._orig_event_handlers
+
+    def _as_first_started(self, engine):
+        if hasattr(engine.state.dataloader, "__len__"):
+            num_iters_per_epoch = len(engine.state.dataloader)
+        else:
+            num_iters_per_epoch = engine.state.epoch_length
```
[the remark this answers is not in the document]
harsh8398: I think we can remove this part if we make the storing of measurements dynamic instead of statically initializing tensors for storing measurements later.

ignite/contrib/handlers/time_profilers.py
harsh8398: Okay. Let me see if I can fine tune the threshold for "every" and "once" event filters.

harsh8398: For custom filters I am not sure how to handle them.

vfdev-5:
@harsh8398 thanks for the update and pinging me. I'll try to checkout your code and play around to understand better how it works and if everything works as intended etc.

Meanwhile, if you could also test the profiler on a concrete example, e.g. examples/contrib/cifar10. For example, you can add this handler in the existing code and run it (eventually, you can reduce the number of epochs to 7 or 10) and see if everything works as intended. It would be very helpful :+1:

harsh8398:
@vfdev-5 Thanks, Okay I'll test it on cifar10 example. I did try MNIST before. Also FYI, there are some things left to do like:

- threshold for event filters
- export results to csv

There are some tests added but I think I can write some better tests for this. I'll update/add the tests at the end once the profiler code is finalized.

vfdev-5:
> Okay I'll test it on cifar10 example. I did try MNIST before

Oh, that's great ! Maybe, cifar10 checking is optional then. However, what I was thinking and which can be done in another follow-up PR, is providing a notebook with a basic training and showing usage of this class... Such that users could have a working example...

harsh8398:
> Oh, that's great ! Maybe, cifar10 checking is optional then. However, what I was thinking and which can be done in another follow-up PR, is providing a notebook with a basic training and showing usage of this class... Such that users could have a working example...

Should I open a follow up issue on that?

vfdev-5:
> Should I open a follow up issue on that? I can work on that after this.

Yes, please.

vfdev-5:
@harsh8398 I checked the code and I have few questions and comments:
- code already works with custom events, that's pretty cool.

<summary>
my code
<details>

```python
import time
from ignite.engine import EventEnum


class BackpropEvents(EventEnum):
    OPTIM_STEP_COMPLETED = 'optim_step_completed'



def train_step(e, b):
    time.sleep(0.1)
    engine.fire_event(BackpropEvents.OPTIM_STEP_COMPLETED)


engine = Engine(train_step)

engine.register_events("a", "b")
engine.register_events(*BackpropEvents)


@engine.on(BackpropEvents.OPTIM_STEP_COMPLETED)
def function_before_backprop(engine):
    time.sleep(0.0333)


@engine.on(Events.EPOCH_STARTED)
def ep_start():
    time.sleep(0.15)

    
@engine.on(Events.EPOCH_COMPLETED)
def ep_end():
    time.sleep(0.2)

a_ev_args1 = [12, 34]

@engine.on(Events.EPOCH_COMPLETED(every=5), a_ev_args1)
def trigger_a_event(args1):
    engine.fire_event("a")

a_ev_args2 = [122, 324]
    
@engine.on("a", a_ev_args2)
def on_a_event(args2):
    print("triggered 'a' event", args2)
    time.sleep(1.0)


profiler = HandlersTimeProfiler()
profiler.attach(engine)


@engine.on(Events.EPOCH_COMPLETED(every=3) | Events.EPOCH_COMPLETED(once=1))
def print_profiling():
    profiler.print_results(profiler.get_results())

def get_data(n_iters):
    for i in range(n_iters):
        time.sleep(0.05)
        yield i

        
data = get_data(5 * 10)

engine.run(data, max_epochs=10, epoch_length=5)
```

</details>

</summary>


- I understand that below condition helps to avoid std to be Nan:
```python
            min_index = ("None", "None")
            max_index = ("None", "None")
            mean = "None"
            std = "None"
            if len(data) > 1:
```
on the other hand, when we see None for triggered events, it is not obvious to understand why:
```
ep_start                      EPOCH_STARTED                    0.15022       None/None       None/None            None            None  
ep_end                        EPOCH_COMPLETED                  0.20028       None/None       None/None            None            None  
trigger_a_event               EPOCH_COMPLETED                    3e-05       None/None       None/None            None            None  
```
How about doing like that 
```python
            min_index = ("None", "None")
            max_index = ("None", "None")
            mean = "None"
            std = "None"
            if len(data) > 0:
                min_index = (round(torch.min(data).item(), 5), torch.argmin(data).item())
                max_index = (round(torch.max(data).item(), 5), torch.argmax(data).item())
                mean = round(torch.mean(data).item(), 5)
                if len(data) > 1:
                    std = round(torch.std(data).item(), 5)
```

I'll comment out also in the code.

ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,299 @@ def odict_to_str(d: Mapping):
...
+        self._event_handlers_timer = Timer()
+
+        self.dataflow_times = None
+        self.processing_times = None
+        self.event_handlers_times = None
+
+    @staticmethod
+    def _get_callable_name(handler: Callable):
+        # get name of the callable handler
+        return handler.__qualname__ if hasattr(handler, "__qualname__") else handler.__class__.__name__
```
vfdev-5, line 482:
```suggestion
        return getattr(handler, "__qualname__", handler.__class__.__name__)
```

ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,299 @@ def odict_to_str(d: Mapping):
...
+        for e in engine._event_handlers:
+            for i, h in enumerate(engine._event_handlers[e]):
+                if hasattr(h[0], "_profiler_original"):
+                    engine._event_handlers[e][i] = h[0]._profiler_original
+
+    def _as_first_started(self, engine):
+
+        self.event_handlers_names = {
+            e: [
+                h.__qualname__ if hasattr(h, "__qualname__") else h.__class__.__name__
```
vfdev-5, line 544:
```suggestion
                getattr(handler, "__qualname__", handler.__class__.__name__)
```

ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,299 @@ def odict_to_str(d: Mapping):
...
+            std = "None"
+            if len(data) > 1:
+                min_index = (round(torch.min(data).item(), 5), torch.argmin(data).item())
+                max_index = (round(torch.max(data).item(), 5), torch.argmax(data).item())
+                mean = round(torch.mean(data).item(), 5)
+                std = round(torch.std(data).item(), 5)
+            return [total, min_index, max_index, mean, std]
+
+        event_handler_stats = [
+            [h, e.name, *compute_basic_stats(torch.tensor(self.event_handlers_times[e][h]))]
```
vfdev-5, line 610:
```suggestion
            [h, getattr(e, "name", str(e)), *compute_basic_stats(torch.tensor(self.event_handlers_times[e][h]))]
```
For custom events as `str` as they do not have `name` attribute.

ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,299 @@ def odict_to_str(d: Mapping):
...
+            # ):
+            self._event_handlers_timer.reset()
+            func, args, kwargs = handler
+            func(*args, **kwargs)
+            t = self._event_handlers_timer.value()
+            hname = self._get_callable_name(func)
+            self.event_handlers_times[event][hname].append(t)
+
+        # required to revert back to original handler after profiling
+        _timeit_handler._profiler_original = handler
```
vfdev-5, line 499:
```suggestion
        setattr(_timeit_handler, "_profiler_original", handler)
```
Anticipate later work for mypy.

ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,299 @@ def odict_to_str(d: Mapping):
...
+            ]
+            for e in engine._allowed_events
+        }
+
+        self._reset(self.event_handlers_names)
+
+        for e in engine._allowed_events:
+            for i, h in enumerate(engine._event_handlers[e]):
+                if not self._is_internal_handler(h):
+                    engine._event_handlers[e][i] = (self._timeit_handler_wrapper(h), (e,), {})
```
vfdev-5, line 556: I wonder if we could only wrap handler function instead of replacing whole bundle as you do. And it would be good to use `functools.wraps` to decorate `_timeit_handler_wrapper` such that if user inspects handlers, he/she still sees its own handlers.

ignite/contrib/handlers/time_profilers.py
harsh8398: I thought python Enum types will always have name attribute. Is it not?

harsh8398: oh got it. Didn't know we can use strings for registering custom events. 😃

vfdev-5: Yes, event can be `Events` or a str if registered.

ignite/contrib/handlers/time_profilers.py
harsh8398: Wrapper just needs event name but I think we can use `Engine.last_event_name` for that purpose as we get engine as first argument anyway. So the wrapper will look as follows with both the changes:
```python
def _timeit_handler_wrapper(self, handler: Callable):
    @functools.wraps(handler)
    def _timeit_handler(engine: Engine, *args: Any, **kwargs: Any):
        self._event_handlers_timer.reset()
        handler(engine, *args, **kwargs)
        t = self._event_handlers_timer.value()
        hname = self._get_callable_name(handler)
        self.event_handlers_times[engine.last_event_name][hname].append(t)

    # required to revert back to original handler after profiling
    setattr(_timeit_handler, "_profiler_original", handler)
    return _timeit_handler
```

harsh8398: Oh sorry, this won't work. We have handler bundle as type tuple here so we won't be able to replace the item in that. I guess we will have to replace whole handler tuple.

vfdev-5: Yes, that's right, we have handler bundle as `(handler, args, kwargs)`. I was thinking to make the following replacement:
```
(handler, args, kwargs) -> (wrapped_handler, args, kwargs)
```
with `wrapped_handler = create_wrapped_handler(handler, event_name)`.

I'm not sure if it would work as you did but would be more simple to read the code ... Let's see.

harsh8398: I just pushed the code to only change/wrap function inside bundle/tuple instead of replacing whole bundle. It still requires recreating bundle with same args and kwargs because of how Tuple works. It does seem more clean now though. Let me know if this looks good.

vfdev-5: @harsh8398 thanks ! Looks good now. We can even reduce it a bit like that:
```python
        for e in engine._allowed_events:
            for i, (func, args, kwargs) in enumerate(engine._event_handlers[e]):
                if not self._is_internal_handler(func):
                    engine._event_handlers[e][i] = (self._create_wrapped_handler(func, e), args, kwargs)
```

harsh8398:
@vfdev-5 I'll start with mypy, test cases and maybe some doc changes. Please let me know if there are any more improvements needed in profiler code.

ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,317 @@ def odict_to_str(d: Mapping):
...
+        # reset the variables used for profiling
+        self.dataflow_times = []
+        self.processing_times = []
+
+        self.event_handlers_times = {e: {h: [] for h in event_handlers_names[e]} for e in event_handlers_names}
+
+    @staticmethod
+    def _is_internal_handler(handler: Callable):
+        # checks whether the handler is internal
+        return "HandlersTimeProfiler." in repr(handler)
```
vfdev-5, line 521:
```suggestion
        return any(n in repr(handler) for n in ["HandlersTimeProfiler.", "Timer."])
```
We have to ignore timers too, I think

vfdev-5:
@harsh8398 Looks awesome now ! 
A little improvement for filtering timers and good to go.

For instance we'll keep this behaviour with thresholding of filtered handlers. In future, I think, it would be better to replace the way we wrap handlers to filter the execution, such that we could easily replace/wrap original handler inside filtering logic. Such that we could add timeit wrapper to the original handler instead of filtered. Anyway, this is something to think about.

vfdev-5 (approved):
Thanks @harsh8398 !

vfdev-5:
@harsh8398 could you please take a look how to fix the rendering in another PR ?

Screen Shot 2020-10-31 at 14 46 43

harsh8398:
@vfdev-5 I'll have to trim it down to fix it. From my monitor:
image

Does sphinx allow scrollbar?

vfdev-5:
@harsh8398 looks good with the trimming. 

> Does sphinx allow scrollbar?

For formula or code block I'm note sure...

harsh8398:
> @harsh8398 looks good with the trimming.

Its already trimmed actually. I think you are looking docs at lower resolution than my monitor.

vfdev-5:
> Its already trimmed actually. I think you are looking docs at lower resolution than my monitor.

That's possible. However when I zoom in/out with browser, layout remins the same. What is your screen resolution, actually ?

Btw, Python 3.5 CI is failing: https://github.com/pytorch/ignite/pull/1398/checks?check_run_id=1336007242#step:6:1091

trsvchn:
> @harsh8398 looks good with the trimming.
> 
> > Does sphinx allow scrollbar?
> 
> For formula or code block I'm note sure...

The same, I'm not sure...

trsvchn:
@vfdev-5 on my `  dimensions:    1366x768 pixels (361x203 millimeters)` and Firefox it looks as follows

With 100% zoom (no zoom)


But with 120% zoom it's okay


I'll check what we can do

vfdev-5:
@trsvchn that's true, nice catch with 120% ! I tried that and also see the correct rendering. Maybe, we can try to make central window a bit larger and thus reduce the size of the left sidebar.

ignite/contrib/handlers/time_profilers.py
```diff
@@ -437,3 +438,321 @@ def odict_to_str(d: Mapping):
...
+            return [total, min_index, max_index, mean, std]
+
+        event_handler_stats = [
+            [h, getattr(e, "name", str(e)), *compute_basic_stats(torch.FloatTensor(self.event_handlers_times[e][h]))]
+            for e in self.event_handlers_times
+            for h in self.event_handlers_times[e]
+        ]
+        event_handler_stats.append(["Total", "", total_eh_time, "", "", "", ""])
+        event_handler_stats.append(
+            ["Processing", "None", *compute_basic_stats(torch.FloatTensor(self.processing_times))]
```
vfdev-5, line 609: Let's keep the type of `compute_basic_stats` input as sequence and if we need to create a float tensor, we do it inside the function using `torch.tensor(.., dtype=torch.float32)`

ignite/contrib/handlers/time_profilers.py
```diff
@@ -3,6 +3,7 @@
 from typing import Any, Callable, Dict, Iterable, List, Mapping, Sequence, Union
 
 import torch
+from torch._C import dtype
```
vfdev-5, line 6: This is unnecessary, please remove this import :)

ignite/contrib/handlers/time_profilers.py
```diff
@@ -600,15 +601,21 @@ def compute_basic_stats(data: Sequence) -> List[Union[str, float]]:
             return [total, min_index, max_index, mean, std]
 
         event_handler_stats = [
-            [h, getattr(e, "name", str(e)), *compute_basic_stats(torch.FloatTensor(self.event_handlers_times[e][h]))]
+            [
+                h,
+                getattr(e, "name", str(e)),
+                *compute_basic_stats(torch.tensor(self.event_handlers_times[e][h], dtype=torch.float32)),
```
vfdev-5, line 607: As `def compute_basic_stats(data: Sequence)`, let's create tensor inside `compute_basic_stats`.

ignite/contrib/handlers/time_profilers.py
harsh8398: 😄 Result of PyLance in VSCode. I think I will get back to old one.

trsvchn:
@vfdev-5 No worries, we can fix that with css tweaks, by changing font size of a code block for that class.

trsvchn:
Here is a test on master branch
Screenshot from 2020-10-31 17-48-00
Screenshot from 2020-10-31 17-48-04

vfdev-5:
Looks neat ! Thanks @trsvchn

trsvchn:
@vfdev-5 I'm working  with updated master and I'm having this:
```bash
/home/trs/dev/hacktoberfest-2020/ignite/docs/source/faq.rst:389: WARNING: Title underline too short.

Basic time profiling
``````````````````
.../ignite/docs/source/faq.rst:389: WARNING: Title underline too short.

Basic time profiling
``````````````````
.../ignite/docs/source/faq.rst:450: WARNING: Title underline too short.

Event handlers time profiling
``````````````````
...ignite/docs/source/faq.rst:450: WARNING: Title underline too short.

Event handlers time profiling
``````````````````
```

PR #1404, opened 2020-10-25
Gitignore jetbrains products

BanzaiTokyo:
Fixes #{issue number} 

Description:

Check list:


vfdev-5:
Thanks for the PR @BanzaiTokyo. Unfortunately, I have to close it as it was decided since the beginning of the project to use local `.git/info/exclude` to exclude custom files unrelated to the project, like: `.idea`, `.vscode` etc.

Issue #1406, opened 2020-10-26
MyPy: improve ignite.contrib module


gruebel:
## 🚀 Feature

The `ignite.contrib` module is the last one, which is ignored by the `mypy` config 💪 For anyone, who wants to do it, I would recommend to split it in 3-4 PRs

- `ignite.contrib.engines`
- `ignite.contrib.handlers`
- `ignite.contrib.metrics`
- (Optional separate PR) `ignite.contrib.metrics.regression`

After that we could also enable more stricter `mypy` configuration to further improve the typing of the project.

So, feel free to ask any questions here or in your PRs, I will try my best to give proper guidance and @vfdev-5 will do the same for sure, like he did with my previous PRs 👍

zhxxn:
Hi, I would like to give it a try!

vfdev-5:
@gruebel thanks a lot for opening the issue ! 
@zhxxn let me assign it to you. 
As this issue is large, we can assign more than one person to the issue.

zhxxn:
Hi! I would then start with `ignite.contrib.metrics` module 😄

zhxxn:
I just finished my work on `ignite.contrib.metrics`. 
Next I would like to try `ignite.contrib.handlers` 😄

vfdev-5:
> Next I would like to try `ignite.contrib.handlers`

@zhxxn how's going your attempt on `ignite.contrib.handlers` ?

zhxxn:
Hi, I am still working on this and it will take some time.. 😅
Is it urgent for you?

vfdev-5:
Hi @zhxxn, no problems, take your time. I just wanted to check if it is not abandoned :)

zhxxn:
@gruebel Hi, would you be interested in taking over my part `ignite.contrib.handlers`? 
Sadly, timewise I cannot finish it 😥😢

gruebel:
@zhxxn sure, no problem 👍 
@vfdev-5 can you assign me to this issue? I will start working on it this weekend 💪

vfdev-5:
@gruebel thanks !

PR #1416, opened 2020-10-28
Activate MyPy in ignite.contrib.engines

rzats:
Fixes _a subset of_ #1406

Description:

Makes the `ignite.contrib.engines` submodule MyPy-compliant.

Check list:

### DIFF
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -216,7 +216,9 @@ def _setup_common_training_handlers(
     if with_pbars:
         if with_pbar_on_iters:
             ProgressBar(persist=False).attach(
-                trainer, metric_names="all", event_name=cast(Events, Events.ITERATION_COMPLETED(every=log_every_iters))
+                trainer,
+                metric_names="all",
+                event_name=Events.ITERATION_COMPLETED(every=log_every_iters),  # type: ignore[arg-type]
             )
 
         ProgressBar(persist=True, bar_format="").attach(
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -47,7 +47,7 @@ def setup_common_training_handlers(
     clear_cuda_cache: bool = True,
     save_handler: Optional[Union[Callable, BaseSaveHandler]] = None,
     **kwargs: Any
-):
+) -> None:
     """Helper method to setup trainer with common handlers (it also supports distributed configuration):
 
         - :class:`~ignite.handlers.TerminateOnNan`
@@ -150,7 +150,7 @@ def _setup_common_training_handlers(
     clear_cuda_cache: bool = True,
     save_handler: Optional[Union[Callable, BaseSaveHandler]] = None,
     **kwargs: Any
-):
+) -> None:
     if output_path is not None and save_handler is not None:
         raise ValueError(
             "Arguments output_path and save_handler are mutually exclusive. Please, define only one of them"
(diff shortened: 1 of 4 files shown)


vfdev-5:
@gruebel could you please review this PR ?

gruebel:
Sure, I will take a look today.

vfdev-5:
@rzats thanks for the PR ! Currently, we have to recover mypy testing runs, now on GitHub Actions, due to partial out of service by Travis. Once it is done we could rerun your PR.

ignite/contrib/engines/common.py
```diff
@@ -262,8 +284,8 @@ def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, lo
 def _setup_logging(
     logger: BaseLogger,
     trainer: Engine,
-    optimizers: Union[Optimizer, Dict[str, Optimizer]],
-    evaluators: Union[Engine, Dict[str, Engine]],
+    optimizers: Union[Optimizer, Dict[str, Optimizer], None],
+    evaluators: Union[Engine, Dict[str, Engine], None],
```
gruebel, line 288: we usually use `Optional[Union[...]]` instead of adding `None` inside the `Union` part. Please see my comment below for the corrected signature.

ignite/contrib/engines/common.py
```diff
@@ -575,7 +598,7 @@ def gen_save_best_models_by_val_score(
         to_save = {"model": models}
 
     best_model_handler = Checkpoint(
-        to_save,
+        cast(Dict[str, torch.nn.Module], to_save),
```
gruebel, line 601: No need to use `cast` here, you can adjust the code
https://github.com/pytorch/ignite/blob/92455594401681bbd3e675a664072b905401274b/ignite/contrib/engines/common.py#L596-L598
to
```python
    if isinstance(models, nn.Module):
        to_save = {"model": models}  # type: Dict[str, nn.Module]
    else:
        to_save = models
```

gruebel (changes requested):

gruebel:
@rzats  Could you also add missing return types in `common.py` and `tbptt.py`? Would be a great help for making `mypy` checking more strict in the near future 💪

rzats:
@gruebel thanks for your comments! I've pushed the requested fixes.

vfdev-5 (commented):
@rzats Thanks for the update ! Looks good to me.
@gruebel what do you think ?

gruebel:
@rzats thanks, looks good!
@vfdev-5 the only thing I would need an opinion from you is about the `ProgressBar.attach()` method. Should we just ignore the incompatible type
https://github.com/pytorch/ignite/blob/1926f2ec999c4e8337d5eec89370be31857093cc/ignite/contrib/engines/common.py#L218-L222
or extend it?
https://github.com/pytorch/ignite/blob/1926f2ec999c4e8337d5eec89370be31857093cc/ignite/contrib/handlers/tqdm_logger.py#L157-L164
`event_name: Events = Events.ITERATION_COMPLETED` -> `event_name: Union[CallableEventWithFilter, Events] = Events.ITERATION_COMPLETED`

vfdev-5:
> the only thing I would need an opinion from you is about the ProgressBar.attach() method. Should we just ignore the incompatible type

@gruebel you are right, I'd say we have to extend `ProgressBar.attach` to what you suggest:

> event_name: Events = Events.ITERATION_COMPLETED -> event_name: Union[CallableEventWithFilter, Events] = Events.ITERATION_COMPLETED

gruebel:
@rzats can you extend the type hint in `ignite/ignite/contrib/handlers/tqdm_logger.py` as mentioned above? After that this PR is good to go!

vfdev-5:
@rzats could you please update the PR

rzats:
@vfdev-5 I've updated this PR (and fixed some additional strict errors that were enabled recently)

vfdev-5 (approved):
Thanks @rzats ! LGTM

PR #1420, opened 2020-10-29
Add balanced accuracy metric

### LINKED ISSUE #1409: Add balanced accuracy metric
## 🚀 Feature

https://scikit-learn.org/stable/modules/generated/sklearn.metrics.balanced_accuracy_score.html#sklearn.metrics.balanced_accuracy_score

https://scikit-learn.org/stable/modules/model_evaluation.html#balanced-accuracy-score

rzats:
Fixes #1409

Description:

Adds sklearn's [balanced_accuracy_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.balanced_accuracy_score.html#sklearn.metrics.balanced_accuracy_score) to the `contrib.metrics` module.

Check list:


vfdev-5:
@rzats thanks for the PR, currently, it is not clearly stated in the issue, but I was thinking about online way to compute it without using sklearn. But let's see if online way to compute it much more complicated vs using sklearn (and RAM limitations etc).

rzats:
@vfdev-5 thanks for the comment! I've actually taken a look at the existing online implementation of precision/recall and it seems that this metric might already be implemented in a way? sklearn defines `balanced_accuracy_score` as "the average of recall on each class" which can be [calculated](https://github.com/pytorch/ignite/blob/master/ignite/metrics/precision.py#L67-L68) by `Recall(average=True)` AFAICT.

vfdev-5:
@rzats yes, that's looks like that. I'm not sure to understand now the implementation and the definition from here : https://scikit-learn.org/stable/modules/model_evaluation.html#balanced-accuracy-score
```
BalAcc = 0.5 * (TP / (TP + FN)  + TN / (TN + FP) ) = 0.5 * (Recall + TN / (TN + FP))
```

They seem to compute average of recall per class from the confusion matrix: https://github.com/scikit-learn/scikit-learn/blob/0fb307bf39bbdacd6ed713c00724f8f871d60370/sklearn/metrics/_classification.py#L1810
and I do not recognize where the term `TN / (TN + FP)`.

Anyway, we can also recompute it using confusion matrix like it is done for IoU.

rzats:
Got it! Would it then make sense to turn this metric into `BalancedAccuracy(output_transform, is_multilabel, device)` which is just a wrapper for `Recall(output_transform, average=True, is_multilabel, device)`? (And then write some tests to confirm that this is indeed the correct calculation)

vfdev-5:
For instance I'm not sure about its implementation. Could you please check for few examples if the result in sklearn of balanced accuracy is `0.5 * (TP / (TP + FN)  + TN / (TN + FP) )` (multiclass and binary cases) and compare it with pure recall results. According to that, we can decide either to implement it as a function derived from existing ConfusionMatrix (like our IoU) or a function using Recall. What do you think ?

rzats:
@vfdev-5 I've made a quick Colab to test this: https://colab.research.google.com/drive/15SHZXA9DbKR-xCa4CWqrXj7G2L0MlVv_?usp=sharing

In brief, the "average of recalls" interpretation seems to be correct, but the formula doesn't extend well to the multiclass case.

vfdev-5:
@rzats yes, you are right about the formula. It is for binary case and as sklearn docs states that there is no single multiclass definition. I think I catched up now your point:

> "the average of recall on each class"

It is what sklearn docstring states. In this case I think there is no benefit of adding a shortcut for ignite's `Recall(average=True)`

However, if we consider below definition:

> Class balanced accuracy as described in [Mosley2013]: the minimum between the precision and the recall for each class is computed. Those values are then averaged over the total number of classes to get the balanced accuracy.

This can be interesting to implement as [metrics arithmetics](https://pytorch.org/ignite/metrics.html#metric-arithmetics)...

rzats:
@vfdev-5 I've updated the PR with a metric based on arithmetic with precision/recall - is that similar to what we want? (If so, I'll get to updating the tests!)

ignite/metrics/balanced_accuracy.py
```diff
@@ -0,0 +1,62 @@
...
+    if recall is None:
+        recall = Recall(
+            output_transform=(lambda x: x) if output_transform is None else output_transform,
+            average=False,
+            device=device,
+        )
+    elif recall._average:
+        raise ValueError("Input recall metric should have average=False")
+
+    min_pr = [min(value) for value in zip(precision, recall)]
```
vfdev-5, line 59: I wonder if it works actually like that with metric arithmetics ? :)

ignite/metrics/balanced_accuracy.py
```diff
@@ -0,0 +1,62 @@
...
+            :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
+            form expected by the metric. It is used only if precision or recall are not provided.
+        device (str or torch.device): specifies which device updates are accumulated on. Setting the metric's
+            device to be the same as your ``update`` arguments ensures the ``update`` method is non-blocking. By
+            default, CPU.
+
+    Returns:
+        MetricsLambda, balanced accuracy metric
+    """
+    if precision is not None and output_transform is not None:
```
vfdev-5, line 35: Same check for `device` is necessary.

vfdev-5 (commented):
@rzats thanks for the update ! Does it work the implementation ?

ignite/metrics/balanced_accuracy.py
rzats: Good point! I'll add tests for this early next week. (Actually, even if this works, would it need a check for whether precision and recall have the same number of classes?)

vfdev-5: I think a colab check first can be interesting before implementing all tests as it is done for other metrics.
Yes, eventually, checking number of classes can be important.

PR #1423, opened 2020-10-30
Save model with same filename

### LINKED ISSUE #1421: How to save model with same filename?
## 🚀 Feature

Hi @vfdev-5 ,

I found that `CheckpointHandler` can't use the fixed filename to save latest model:
https://github.com/pytorch/ignite/blob/v0.4.2/ignite/handlers/checkpoint.py#L149
But this is a very important requirement on our side...How can I achieve that?
For example, I want to always save the best model as `model.pt` after validations during training.

Thanks.

sdesrozis:
Fixes #1421 

Description:

This solves the problem to do not save model with an existing filename. Now, filename is overwritten.

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -356,13 +356,12 @@ class Checkpoint(Serializable):
                 "priority": priority,
             }
 
-            filename_already_exists = any(item.filename == filename for item in self._saved)
+            saved = [item for item in self._saved if item.filename != filename]
 
-            if filename_already_exists:
+            if self._saved != saved:
                 if isinstance(self.save_handler, BaseSaveHandler):
                     self.save_handler.remove(filename)
-                # list is purged
-                self._saved = [item for item in self._saved if item.filename != filename]
+                self._saved = saved
             elif not self._check_lt_n_saved():
                 item = self._saved.pop(0)
                 if isinstance(self.save_handler, BaseSaveHandler):
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -363,11 +363,10 @@ class Checkpoint(Serializable):
                     self.save_handler.remove(filename)
                 # list is purged
                 self._saved = [item for item in self._saved if item.filename != filename]
-            else:
-                if not self._check_lt_n_saved():
-                    item = self._saved.pop(0)
-                    if isinstance(self.save_handler, BaseSaveHandler):
-                        self.save_handler.remove(item.filename)
+            elif not self._check_lt_n_saved():
+                item = self._saved.pop(0)
+                if isinstance(self.save_handler, BaseSaveHandler):
+                    self.save_handler.remove(item.filename)
 
             self._saved.append(Checkpoint.Item(priority, filename))
             self._saved.sort(key=lambda item: item[0])
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -148,7 +148,7 @@ class Checkpoint(Serializable):
 
         **Warning:** Please, keep in mind that if filename collide with already used one to saved a checkpoint,
         new checkpoint will replace the older one. This means that filename like ``checkpoint.pt`` will be saved
-        every call and will always be overwrittDiskSaveren by newer checkpoints.
+        every call and will always be overwritten by newer checkpoints.
 
     Note:
         To get the last stored filename, handler exposes attribute ``last_checkpoint``:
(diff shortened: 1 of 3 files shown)


ignite/handlers/checkpoint.py
```diff
@@ -147,8 +147,8 @@ class Checkpoint(Serializable):
         ``30000-checkpoint-94.pt``
 
         **Warning:** Please, keep in mind that if filename collide with already used one to saved a checkpoint,
-        new checkpoint will not be stored. This means that filename like ``checkpoint.pt`` will be saved only once
-        and will not be overwritten by newer checkpoints.
+        new checkpoint will replace the older one. This means that filename like ``checkpoint.pt`` will be saved
+        every call and will always be overwrittDiskSaveren by newer checkpoints.
```
vfdev-5, line 151:
```suggestion
        every call and will always be overwritten by newer checkpoints.
```

sdesrozis: Sorry 😅

ignite/handlers/checkpoint.py
```diff
@@ -350,19 +350,24 @@ def __call__(self, engine: Engine) -> None:
...
-                item = self._saved.pop(0)
+            filename_already_exists = any(item.filename == filename for item in self._saved)
+
+            if filename_already_exists:
                 if isinstance(self.save_handler, BaseSaveHandler):
-                    self.save_handler.remove(item.filename)
+                    self.save_handler.remove(filename)
+                # list is purged
+                self._saved = [item for item in self._saved if item.filename != filename]
+            else:
```
vfdev-5, line 366: We can use `elif` instead of `else -> if`

ignite/handlers/checkpoint.py
```diff
@@ -350,19 +350,24 @@ def __call__(self, engine: Engine) -> None:
...
-            if not self._check_lt_n_saved():
-                item = self._saved.pop(0)
+            filename_already_exists = any(item.filename == filename for item in self._saved)
+
+            if filename_already_exists:
                 if isinstance(self.save_handler, BaseSaveHandler):
-                    self.save_handler.remove(item.filename)
+                    self.save_handler.remove(filename)
+                # list is purged
+                self._saved = [item for item in self._saved if item.filename != filename]
```
vfdev-5, line 365: I'd like to simplify `any(item.filename == filename for item in self._saved)` and `self._saved = [item for item in self._saved if item.filename != filename]` somehow.

sdesrozis: Ok I think about it.

ignite/handlers/checkpoint.py
```diff
@@ -356,13 +356,12 @@ def __call__(self, engine: Engine) -> None:
...
-            filename_already_exists = any(item.filename == filename for item in self._saved)
+            saved = [item for item in self._saved if item.filename != filename]
 
-            if filename_already_exists:
+            if self._saved != saved:
                 if isinstance(self.save_handler, BaseSaveHandler):
                     self.save_handler.remove(filename)
-                # list is purged
-                self._saved = [item for item in self._saved if item.filename != filename]
+                self._saved = saved
```
vfdev-5, line 364: Thanks Sylvain! I'm a bit bothered by the fact that we redefine `self._saved` every time if `filename_already_exists`. I was thinking if we could get the Item that has existing filename and then call remove on the list of `self._saved` ? What do you think ?

sdesrozis: I agree with you.

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -1446,3 +1449,23 @@ def test_checkpoint_load_state_dict():
     sd = {"saved": [(0, "model_0.pt"), (10, "model_10.pt"), (20, "model_20.pt")]}
     checkpointer.load_state_dict(sd)
     assert checkpointer._saved == true_checkpointer._saved
+
+
+def test_checkpoint_fixed_filename():
+    save_handler = MagicMock(spec=BaseSaveHandler)
+    model = DummyModel()
+    to_save = {"model": model}
+    checkpointer = Checkpoint(to_save, save_handler=save_handler, n_saved=None, filename_pattern="{name}.{ext}")
```
vfdev-5, line 1458: Could you please test with different values of `n_saved`: None and 1 or 2
and call `checkpointer(trainer)` at least 5-10 times.

vfdev-5 (approved):
Thanks a lot, Sylvain ! I like your last solution 👍 
I left a nit comment to improve a bit the test.

PR #1425, opened 2020-10-31
Replace Number type with float

### LINKED ISSUE #1407: Replace type hint numbers.Number with float
## 🚀 Feature

Due to limitations in `mypy` or to be precise in `typeshed`, `int` and `float` are no `numbers.Number`. (For some extra details https://github.com/pytorch/ignite/pull/1391#discussion_r511656836) 

Therefore it would be helpful to change the typing from `number.Number` to `float` and remove several `# type: ignore`, where applicable.

gruebel:
Fixes #1407 

Description:
I also removed unneeded type ignores, due to the improved typing in pytorch 1.7.0

### DIFF
--- a/ignite/distributed/auto.py
+++ b/ignite/distributed/auto.py
@@ -284,19 +284,19 @@ class DistributedProxySampler(DistributedSampler):
 
     def __iter__(self) -> Iterator:
         # deterministically shuffle based on epoch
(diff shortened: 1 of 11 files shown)


vfdev-5 (approved):
Thanks a lot, Anton !

Issue #1405, opened 2020-10-25
Fix incorrect design for RunningAverage, Frequency, ModelCheckpoint, ...


vfdev-5:
## 🐛 Bug description

Thanks to PRs on mypy annotation by @gruebel we discovered few classes where we have to redesign methods in order to keep consistency with their base classes.

- Frequency: https://github.com/pytorch/ignite/pull/1391#discussion_r511651133
- RunningAverage: https://github.com/pytorch/ignite/pull/1391#discussion_r511651876
- Checkpoint: `to_save` should not be `Optional` but it is due to `ModelCheckpoint` and current implementation.
https://github.com/pytorch/ignite/blob/ddcce01eb7f64efa274ce25703d8f28245715319/ignite/handlers/checkpoint.py#L241

@gruebel please comment here if you think about other classes/methods we did a type relaxation.

cc @sdesrozis @fco-dv

gruebel:
Similar to `Frequency` also `GpuInfo` inherits from `Metric`, so the `attach` method should have the same signature
https://github.com/pytorch/ignite/blob/ddcce01eb7f64efa274ce25703d8f28245715319/ignite/contrib/metrics/gpu_info.py#L107-L108

vfdev-5:
Closing as solved

PR #1433, opened 2020-11-03
enable extra flags for stricter type checking

gruebel:
@vfdev-5 Sorry, was to lazy to create an issue for it 😄 

Description:
I enabled as much extra `mypy` flags as possible and fixed related issues. For each flag, which `mypy` would normally set to `True`, if the strict mode is used, I added a comment, why I disabled it.

### DIFF
--- a/ignite/distributed/utils.py
+++ b/ignite/distributed/utils.py
@@ -496,7 +496,7 @@ def initialize(backend: str, **kwargs: Any) -> None:
     for comp_model_cls in registered_computation_models:
         if backend not in comp_model_cls.available_backends:
             continue
-        _set_model(comp_model_cls(backend, **kwargs))
+        _set_model(comp_model_cls(backend, **kwargs))  # type: ignore[arg-type]
 
 
 def finalize() -> None:
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -6,7 +6,7 @@ import warnings
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict, namedtuple
 from tempfile import _TemporaryFileWrapper  # type: ignore[attr-defined]
-from typing import Callable, Mapping, Optional, Union
+from typing import Any, Callable, Dict, List, Mapping, NamedTuple, Optional, Tuple, Union
 
 import torch
 import torch.nn as nn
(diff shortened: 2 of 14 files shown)


ignite/handlers/checkpoint.py
```diff
@@ -6,7 +6,7 @@
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict, namedtuple
```
vfdev-5, line 7: Can we remove `namedtuple` now ?

ignite/distributed/utils.py
```diff
@@ -496,7 +496,7 @@ def train_fn(local_rank, a, b, c):
     for comp_model_cls in registered_computation_models:
         if backend not in comp_model_cls.available_backends:
             continue
-        _set_model(comp_model_cls(backend, **kwargs))
+        _set_model(comp_model_cls(backend, **kwargs))  # type: ignore[arg-type]
```
gruebel, line 499: @vfdev-5 I had to set it to ignore here, because otherwise I get this error
```
error: Argument 1 to "_HorovodDistModel" has incompatible type "str"; expected "bool"
```
it also makes sense, because the `__init__` method is not defined as in the other models.

mypy.ini
```diff
@@ -3,6 +3,27 @@ files = ignite
 pretty = True
 show_error_codes = True
 
+check_untyped_defs = True
+; a lot of work needed to fix issues
```
vfdev-5, line 7: Could you please detail this point ? Is it only for `disallow_any_generics = False` to make it `True`, right ?

vfdev-5 (commented):
Thanks for the PR @gruebel !
I left few comments

ignite/distributed/utils.py
vfdev-5: Yes, this is weird actually in the sense how it worked previously with `_HorovodDistModel` :)

gruebel: Should I add it to the pile of shame? #1405 😄

ignite/handlers/checkpoint.py
gruebel: good catch!

mypy.ini
gruebel: Yeah, if I set `disallow_any_generics = True` a lot of "errors" will pop up, like all the `Callable` types have to be more precise.

ignite/distributed/utils.py
gruebel: Never mind, I see, you created already a new issue for it 👍

PR #1447, opened 2020-11-10
fix ReduceOp typing issue

### LINKED ISSUE #1446: Fix mypy issue with pytorch nightly
## 🐛 Bug description

Currently, there is an issue with running mypy on ignite when using pytorch-nightly : 
```
 ignite/distributed/comp_models/native.py:334: error: Incompatible types in
assignment (expression has type "ReduceOp", variable has type "str") 
[assignment]
                op = self._reduce_op_map[op]
                     ^
Found 1 error in 1 file (checked 84 source files)
```

https://github.com/pytorch/ignite/runs/1373059337#step:5:17

@gruebel could you please take a look if we can fix it ? Thanks

gruebel:
Fixes #1446 

Description:
Simple and easy fix 😄


vfdev-5 (approved):
Thanks @gruebel ! Simple and clean fix as always :)

Issue #1442, opened 2020-11-06
Migrate from Travis CI to GitHub Actions


vfdev-5:
## 🚀 Feature

The idea is to migrate 3 following stages from Travis CI to GitHub Actions:
  - [x] Docs: https://github.com/pytorch/ignite/blob/master/.travis.yml#L64-L95
  - [x] Deploy: https://github.com/pytorch/ignite/blob/master/.travis.yml#L97-L145
  - [x] Nightly: https://github.com/pytorch/ignite/blob/master/.travis.yml#L147-L175 (done by https://github.com/pytorch/ignite/pull/1450)

Each stage can be a separate action: e.g. Docs -> `.github/workflows/docs.yml`, Deploy -> `.github/workflows/binaries-release.yml`, Nightly -> `.github/workflows/binaries-nightly-release.yml`

For PyDataGlobal contributors, feel free to ask questions for details if any and say that you would like to tackle the issue.
Please, take a look at CONTRIBUTING guide.

ydcjeff:
@vfdev-5 docs and deploy left to done?
can i take this?

vfdev-5:
@ydcjeff yes, docs and deploy are still open. 
You can take a look at how to migrate docs, if you like. This task is also overlaps a little with this https://github.com/pytorch/ignite/pull/1350 in the sense that I'd like to have ["Nice to have" part](https://github.com/pytorch/ignite/issues/1270#issuecomment-703180980) for this issue.

ydcjeff:
Okay, I will start working if that PR get closed.

vfdev-5:
@ydcjeff #1350 does not solve the same problem. It will be closed once we migrated from Travis.

ydcjeff:
Alright then I will draft a PR

vfdev-5:
Thanks @ydcjeff !

vfdev-5:
Done via #1458 #1457 and #1450

Issue #1270, opened 2020-09-08
Preview of built docs in the PR


vfdev-5:
## 🚀 Feature

Idea is to be able to preview built docs on the CI to simplify PR review.
This issue may be a bit tricky to solve, here some guidelines:
- we build our docs with travis ci : https://github.com/pytorch/ignite/blob/master/.travis.yml#L62
- find out how we could keep built docs as artifacts and make them accessible after the build
- update .travis.yml for that
- update https://github.com/pytorch/ignite/blob/master/.github/PULL_REQUEST_TEMPLATE.md with some details on how to fetch the link of built docs.

For Hacktoberfest contributors, feel free to ask questions for details if any and say that you would like to tackle the issue.
Please, take a look at [CONTRIBUTING guide](https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md).

sroy8091:
@vfdev-5 currently are we using github pages for deploying docs?

vfdev-5:
@sroy8091 yes, docs are built and deployed as a Travis CI job on master push: https://github.com/pytorch/ignite/blob/master/.travis.yml#L62

sroy8091:
We can upload the artifact on build success to s3, including the build number as s3 bucket path. Next step would be to include that link in the pr.

sroy8091:
Any better approach @vfdev-5

vfdev-5:
@sroy8091 I'm eager to do it with netlify like here : https://github.com/beeware/beeware.github.io/pull/426
Let me check it first and conifgure it on netlify side. You can then provide additional info...

vfdev-5:
Here is netlify configured: 
- a test PR : https://github.com/pytorch/ignite/pull/1348
and docs preview for the PR : https://deploy-preview-1348--pytorch-ignite-preview.netlify.app/

Currently, it builds the docs as described in CONTRIBUTING and not as in travis (maybe OK).
What remains is to do:
- add condition to travis yaml to run Docs stage on master only
- update CONTRIBUTING guide saying that there "netlify/pytorch-ignite-preview/deploy-preview" check available to visualize the docs.

sroy8091:
awesome. will push the travis changes

vfdev-5:
Close this as done

PR #1457, opened 2020-11-11
[ci] migrate docs build to github actions

ydcjeff:
Part of #1442 

Description: migrate docs build to github actions

Check list:


ydcjeff:
There are some warnings in the logs. logo is missing

ydcjeff:
@vfdev-5 does it look good to you?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,40 @@
...
+          pip install torch torchvision -f https://download.pytorch.org/whl/cpu/torch_stable.html -U
+          pip install -r requirements-dev.txt
+          pip install -r docs/requirements.txt
+          pip install git+https://github.com/vfdev-5/sphinxcontrib-versioning.git
+
+      - name: build docs
+        shell: bash -l {0}
+        run: sphinx-versioning --use-master-conf --use-master-templates build --greatest-tag --whitelist-branches master docs/source docs/build/html
+
+      # - name: deploy docs
```
vfdev-5, line 33: Deploy part should be uncommented and have if condition on pushed to master or on tag, right ?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.7"
+      - name: install docs reqs
+        shell: bash -l {0}
+        run: |
+          pip install -U pip wheel
```
vfdev-5, line 23: I was also thinking to replace installation part with the one we do it with conda and cache... Your cache PR is still open, let me comment out there what have to be done...

vfdev-5 (commented):
Thanks for the PR @ydcjeff

.github/workflows/docs.yml
ydcjeff: There is `on: push - branches: [master]` and `tag: ["v*"]`. Yes, it should be uncommented, but no need for if condition, I think. Can you verify `tag` will work as I haven't used `tag` before?

.github/workflows/docs.yml
ydcjeff: I would suggest using `actions/setup-python` as it's easier for caching. But it's up to you, conda or pip

vfdev-5: I agree with you and OK let's keep it simple here with `actions/setup-python`.

.github/workflows/docs.yml
vfdev-5: Yes, you are right about push and tag. Probably, should work like that (to verfiy in the playground). My point is about pull request. Actually, I forgot that we wanted to build docs with netlify, so we can remove `pull_request:
branches: [master]` from here, I think. Thus, we build and deploy on push-master and on tag. PR's preview is done on netlify and it was also a nice to have from https://github.com/pytorch/ignite/issues/1270#issuecomment-703180980 and the idea is to refactor deps installation and docs building into 2 scripts that could be used here and in netlify. Could you please create those 2 scripts inside `.github/workflows` and reuse them here ?

ydcjeff: I see!
That line should be uncommented, just want to test via PR if docs can be built and yes `pull_request` needs to be removed.

ydcjeff: @vfdev-5 when I tested on my fork, `tags` works if you do with `git push --tags` and make a tag that starts with `v` from github UI. `release` only works if you make a release/tag from github UI.

vfdev-5: Both are OK, I think

.github/workflows/docs.yml
```diff
@@ -0,0 +1,33 @@
...
+        shell: bash -l {0}
+        run: bash .github/workflows/install_docs_deps.sh
+
+      - name: Build docs
+        shell: bash -l {0}
+        run: bash .github/workflows/build_docs.sh
+
+      - name: deploy docs
+        uses: peaceiris/actions-gh-pages@v3
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
vfdev-5, line 28: Indent is wrong here

vfdev-5 (approved):
@ydcjeff thanks for the PR !

PR #1463, opened 2020-11-14
[docs] add missing function in handlers docs

ydcjeff:
Description: add missing function in handlers docs

Check list:


vfdev-5 (approved):
Thanks for the PR, @ydcjeff !

PR #1468, opened 2020-11-15
setup toml, yaml, prettier in pre-commit

ydcjeff:
Description: setup toml, yaml, prettier in pre-commit

Check list:


vfdev-5:
@ydcjeff thanks for the PR and the suggestion ! I was wondering if `.pre-commit-config.yaml` is the only file that should be prettified ?

ydcjeff:
> I was wondering if .pre-commit-config.yaml is the only file that should be prettified ?

Can you explain? pre-commit hooks format the changed files only.

vfdev-5:
Yes, true. So, OK, once modified it will be prettified.

vfdev-5 (approved):
Thanks @ydcjeff for the PR !

ydcjeff:
@vfdev-5 thank you so now .yaml, .toml will be modified once it changed.

PR #1470, opened 2020-11-16
[docs] make GIF look good on mobile

ydcjeff:
Description: make GIF on the front page look good on mobile
It's not needed but nice to have

Check list:


vfdev-5:
Thanks for the PR @ydcjeff ! Yes, that makes sense to make the front page look good on mobile. However, on my phone it is too small with 60% width. Let's try 75 or 80% ?

ydcjeff:
made it 75%, 80 would be large on desktop.

vfdev-5:
Yes, I agree that 80% is too big on desktop and even 75% is rather large, IMO. It would be nice to make a sort of `if-mobile` than 75%, `if-desktop` then either width in pixel or 60%... @trsvchn what do you think, any ideas for that ?

trsvchn:
@vfdev-5 Yes, I like idea with a kind `mobile/desktop` switch.

vfdev-5:
@trsvchn thanks for the feedback! 

Actually, I'm just wondering if how many people consulting our docs on mobile phones and see the GIF ?

Maybe, we can just set it to 67% such that it is rendered OK on desktop and never mind if it is too small for mobile... If we would like to improve it. This can be done later.

ydcjeff:
We can use the directive `.. raw:: html` to make GIF looks good on mobile and desktop for now as a workaround.

trsvchn:
In [How to Make Your Images Mobile-Friendly (Responsive Design)](https://www.thesitewizard.com/css/responsive-images-for-mobile-friendly-website.shtml) I've found
> The simplest and most painless way to quickly get all your existing foreground pictures to fit a mobile screen is to resize them with CSS. This can easily be done with the following rules:

```css
img {
  width: auto ;
  max-width: 100% ;
  height: auto ;
} 
```

vfdev-5 (approved):
Thanks !

PR #1475, opened 2020-11-24
Docker : remove Entrypoints / update Readme and build_all.sh scripts

fco-dv:
Fixes #{issue number} 

Description:
Remove Dockerfile entrypoints , update READMEs accordingly. Add a quotes fix in build_all.sh => needs a test in the CI

Check list:


vfdev-5:
@fco-dv could you please udpate horovod version to v0.21.0 ? Maybe it worth to make a global arg for that instead of the changing all related files ?

vfdev-5 (approved):
Thanks @fco-dv

PR #1478, opened 2020-11-25
Add dict in metrics results

### LINKED ISSUE #1476: Breaking change in metrics behaviour
## 🐛 Bug description

Following https://github.com/pytorch/ignite/pull/968 (and associated issue ) metric's output is flatten if it returns a mapping/dict. In details, with ignite v0.4.2, for a 
custom metric's we have the following:
```python
class PerformanceIndicators(Metric):

    def compute(self):
        # ...
        return {
            'a': 12,
            'b': 23
        }

PerformanceIndicators().attach(evaluator, name="indicators")

assert "a" in evaluator.state.metrics
assert "b" in evaluator.state.metrics

# This is a breaking change introduced by the PR
assert not "indicators" in evaluator.state.metrics

print(evaluator.state.metrics)
> {'a': 12, 'b': 23}
```

The questions we would like to address in this issue:
- Should we fix the breaking change by readding the dict into `evaluator.state.metrics` such that we'll have:
```
print(evaluator.state.metrics)
> {'a': 12, 'b': 23, 'indicators': {'a': 12, 'b': 23}}
```
 
- `name` parameter (e.g. "indicators") is never used. Should we append it to the flatten names (e.g. `indicators/a`, `indicators/b`) or we can accept that it is never used ?

Thanks @lidq92 for reporting this.

## Environment

 - PyTorch Version (e.g., 1.4): 1.6.0
 - Ignite Version (e.g., 0.3.0): 0.4.2
 - OS (e.g., Linux): 
 - Python version:

cc @sdesrozis @liebkne

sdesrozis:
Fixes #1476 

Description:

As discussed, in case of metrics returning dict, in addition to flattened values, dict is added.

Check list:

### DIFF
--- a/ignite/metrics/metric.py
+++ b/ignite/metrics/metric.py
@@ -307,7 +307,7 @@ class Metric(metaclass=ABCMeta):
             if name in result.keys():
                 raise ValueError(
                     "Argument name '{}' is conflicting with mapping keys: {}".format(name, list(result.keys()))
-                 )
+                )
 
             for key, value in result.items():
                 engine.state.metrics[key] = value
--- a/ignite/metrics/metric.py
+++ b/ignite/metrics/metric.py
@@ -305,7 +305,9 @@ class Metric(metaclass=ABCMeta):
         result = self.compute()
         if isinstance(result, Mapping):
             if name in result.keys():
-                raise ValueError("name of metric is conflicting with mapping keys")
+                raise ValueError(
+                    "Argument name '{}' is conflicting with mapping keys: {}".format(name, list(result.keys()))
+                 )
 
             for key, value in result.items():
                 engine.state.metrics[key] = value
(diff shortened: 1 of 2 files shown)


ignite/metrics/metric.py
```diff
@@ -300,11 +300,16 @@ def completed(self, engine: Engine, name: str) -> None:
 
         Args:
             engine (Engine): the engine to which the metric must be attached
+            name (str): the name of the metric used as key in dict `engine.state.metrics`
         """
         result = self.compute()
         if isinstance(result, Mapping):
+            if name in result.keys():
+                raise ValueError("name of metric is conflicting with mapping keys")
```
vfdev-5, line 308:
```suggestion
                raise ValueError(
                    "Argument name '{}' is conflicting with mapping keys: {}".format(name, list(result.keys()))
                 )
```

vfdev-5 (commented):
Thanks for the PR @sdesrozis 
I left a nit comment

vfdev-5:
@sdesrozis tests are failing: https://github.com/pytorch/ignite/runs/1455446251
could you please fix them ?

vfdev-5 (approved):
LGTM. Thanks @sdesrozis !

PR #1480, opened 2020-11-28
[tests] py.test -> pytest

ydcjeff:
Description: change `py.test` to `pytest`. `pytest` is recommended over `py.test`, just to be future-proof.

> The issue with the 'other tool' has been resolved for some time now, so from pytest 3.0 we will supported and recommend use of pytest as the main command instead of py.test. It's possible that in future we will deprecate py.test and potentially even remove it.

https://github.com/pytest-dev/pytest/issues/1629#issue-161422224

@vfdev-5


vfdev-5 (approved):
Thanks @ydcjeff !

PR #1483, opened 2020-12-02
Use python setup.py install and format all yaml files

ydcjeff:
Description: Use python setup.py install everywhere (change in .circleci/config.yaml) and format all yaml files

**Formatting is done by pre-commit hook**

Check list:

@vfdev-5


vfdev-5:
@ydcjeff thanks for the PR. Why do you want to change `python setup.py install` with `pip install .` ?
Personally, I'd prefer to keep `python setup.py` if there is no a good reason of using pip install ...
As for formatting with precommit, I'm hesitating with reformatting GA yaml files without breaking them as GA may be sensible for the indent ... However, seems like all tests passed...

ydcjeff:
> @ydcjeff thanks for the PR. Why do you want to change `python setup.py install` with `pip install .` ?
> Personally, I'd prefer to keep `python setup.py` if there is no a good reason of using pip install ...
> As for formatting with precommit, I'm hesitating with reformatting GA yaml files without breaking them as GA may be sensible for the indent ... However, seems like all tests passed...

I read from the stackoverflow and https://discuss.python.org, core devs discouraged using of `python setup.py develop` and `python setup.py test`, so I thought the same would go for `install` as well
Also, I noticed in circleci tests, we are using `pip install .`, so I changed the rest

I use pre-commit for the GA yaml daily, I can say they are valid yaml for GA

vfdev-5:
@ydcjeff thanks for the info ! I read [a topic](https://discuss.python.org/t/the-difference-between-python-setup-py-build-ext-i-and-pip-install-e/3716) on the python forum and saw a mentions about certain deprecations of `python setup.py`. 
Let me think about that and collect more feedback on that.

vfdev-5:
@ydcjeff I propose to do the following here. Let's use everywhere `python setup.py install` instead of `pip install .` until I gathered a comprehensive feedback on why using pip is better (personnally, I think this deprecation does not make much sense and a lot of projects rely on `python setup.py install/develop`). And let's update all yaml with the precommit suggestion. What do you think ?

ydcjeff:
@vfdev-5 Sounds good, will change back to `python setup.py`

.circleci/config.yml
```diff
@@ -85,14 +80,13 @@ install_dependencies: &install_dependencies
       name: Install dependencies
       command: |
         docker exec -it pthd pip install -r requirements-dev.txt
-        export install_ignite_cmd='pip install .'
+        export install_ignite_cmd='python setup.py install'
```
ydcjeff, line 83: The change is here, the rest should be formatting by pre-commit
I have checked this is the only change, if you find other else where, let me know.

vfdev-5: Yes, I also think this is the only change.

ydcjeff:
btw, why doesn't circleci run even I edit it?

vfdev-5:
> btw, why doesn't circleci run even I edit it?

It's disabled on PR for let's say ecological reasons :)

vfdev-5 (approved):
Thanks @ydcjeff !

PR #1484, opened 2020-12-02
Activate mypy for contrib.handlers

### LINKED ISSUE #1406: MyPy: improve ignite.contrib module

gruebel:
Fixes #1406

Description:
That was brutal. Especially the `param_scheduler.py` is such a mess 😄 A lot of the time I had to guess and debug to find out what types are used/returned. 
Since the private class `torch.optim.lr_scheduler._LRScheduler` has no types defined, I had to add `type: ignore` quite often 😢

### DIFF
--- a/ignite/contrib/handlers/time_profilers.py
+++ b/ignite/contrib/handlers/time_profilers.py
@@ -102,7 +102,7 @@ class BasicTimeProfiler:
                     "As epoch_length is not set, we can not use BasicTimeProfiler in this case."
                     "Please, set trainer.run(..., epoch_length=epoch_length) in order to fix this."
                 )
-            num_iters_per_epoch = cast(int, engine.state.epoch_length)
+            num_iters_per_epoch = engine.state.epoch_length
 
         self.max_epochs = cast(int, engine.state.max_epochs)
         self.total_num_iters = self.max_epochs * num_iters_per_epoch
--- a/ignite/contrib/handlers/time_profilers.py
+++ b/ignite/contrib/handlers/time_profilers.py
@@ -588,7 +588,7 @@ class HandlersTimeProfiler:
         def compute_basic_stats(
             times: Union[Sequence, torch.Tensor]
         ) -> List[Union[str, float, Tuple[Union[str, float], Union[str, float]]]]:
-            data = torch.as_tensor(data, dtype=torch.float32)
+            data = torch.as_tensor(times, dtype=torch.float32)
             # compute on non-zero data:
             data = data[data > 0]
             total = round(torch.sum(data).item(), 5) if len(data) > 0 else "not triggered"  # type: Union[str, float]
--- a/ignite/contrib/handlers/time_profilers.py
+++ b/ignite/contrib/handlers/time_profilers.py
@@ -101,7 +101,7 @@ class BasicTimeProfiler:
                 raise ValueError(
                     "As epoch_length is not set, we can not use BasicTimeProfiler in this case."
                     "Please, set trainer.run(..., epoch_length=epoch_length) in order to fix this."
-                )            
+                )
             num_iters_per_epoch = cast(int, engine.state.epoch_length)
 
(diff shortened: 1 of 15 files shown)


vfdev-5:
@gruebel thanks a lot for the PR ! Maybe, it would be more simple to split the PR into multiple ones (simplier reviewing) and either we can try to rework `param_scheduler.py` before passing mypy inside ?

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -232,25 +233,25 @@ class _OutputHandler(BaseOutputHandler):
...
             metric_names = []
         super(_OutputHandler, self).__init__(description, metric_names, output_transform, global_step_transform=None)
         self.closing_event_name = closing_event_name
 
     @staticmethod
-    def get_max_number_events(event_name: Union[CallableEventWithFilter, Enum], engine: Engine):
+    def get_max_number_events(event_name: Union[str, Events], engine: Engine) -> int:
         if event_name in (Events.ITERATION_STARTED, Events.ITERATION_COMPLETED):
-            return engine.state.epoch_length
+            return cast(int, engine.state.epoch_length)
```
vfdev-5, line 249: I understand here mypy's reasoning as `epoch_length` and `max_epochs` can be both None (for unknown-size iterators and when user sets max_iters instead of max_epochs). Probably, we can improve it by explicitly saying that the output can be also None...

vfdev-5 (commented):
@gruebel I reviewed all files except ignite/contrib/handlers/param_scheduler.py , so let's split this PR into 2: 1 = everything without ignite/contrib/handlers/param_scheduler.py and 2 = only ignite/contrib/handlers/param_scheduler.py

I left a comment to improve a bit tqdm handler. Otherwise, looks good to me.

ignite/contrib/handlers/tqdm_logger.py
gruebel: ok, let me check, what happens 😄

gruebel:
Sure, no problem. I removed the changes on `param_scheduler.py` and will create a new PR after this one is done 😄

ignite/contrib/handlers/time_profilers.py
```diff
@@ -93,13 +93,13 @@ def _reset(self, num_epochs: int, total_num_iters: int):
...
         }
 
-    def _as_first_started(self, engine: Engine):
+    def _as_first_started(self, engine: Engine) -> None:
         if hasattr(engine.state.dataloader, "__len__"):
-            num_iters_per_epoch = len(engine.state.dataloader)
+            num_iters_per_epoch = len(engine.state.dataloader)  # type: ignore[arg-type]
         else:
-            num_iters_per_epoch = engine.state.epoch_length
+            num_iters_per_epoch = cast(int, engine.state.epoch_length)
```
vfdev-5, line 100: Let's add here a check/raise for `epoch_length` like:
```python
            if engine.state.epoch_length is None:
                raise ValueError(
                    "As epoch_length is not set, we can not use BasicTimeProfiler in this case."
                    "Please, set trainer.run(..., epoch_length=epoch_length) in order to fix this."
                )
            num_iters_per_epoch = cast(int, engine.state.epoch_length)
```

vfdev-5 (approved):
Thanks @gruebel !

PR #1485, opened 2020-12-04
remove `shell` in cache step

ydcjeff:
Description: there should be no `shell` in using actions/cache@v2
See: https://github.com/pytorch/ignite/actions/runs/399592195
@vfdev-5 I'm sorry about that.

Check list:


vfdev-5 (approved):
@ydcjeff thanks for the fix!

PR #1488, opened 2020-12-05
Activate mypy for contrib.handlers.param_scheduler

gruebel:
Relates to #1406 

Description:
Now the whole project is typed, where possible. And we made `mypy` happy 😄 🎉

### DIFF
--- a/ignite/contrib/handlers/param_scheduler.py
+++ b/ignite/contrib/handlers/param_scheduler.py
@@ -6,7 +6,7 @@ from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
 from copy import copy
 from pathlib import Path
-from typing import Any, List, Mapping, Optional, Sequence, Tuple, Union
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple, Type, Union, cast
 
 import torch
 from torch.optim.lr_scheduler import _LRScheduler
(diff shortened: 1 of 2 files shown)


ignite/contrib/handlers/param_scheduler.py
```diff
@@ -54,11 +54,11 @@ def __init__(
         self.optimizer = optimizer
         self.param_group_index = param_group_index
         self.param_name = param_name
-        self.save_history = save_history
         self.event_index = 0
+        self._save_history = save_history
         self._state_attrs = ["event_index", "param_name", "save_history", "param_group_index"]
 
-    def __call__(self, engine: Engine, name: Optional[str] = None):
+    def __call__(self, engine: Optional[Engine], name: Optional[str] = None) -> None:
```
vfdev-5, line 61: I wonder what would be better: either `engine: Optional[Engine] = None` or remove `Optional` and set up a fake engine in `simulate_values`. What do you think ?

vfdev-5 (commented):
@gruebel thanks !

ignite/contrib/handlers/param_scheduler.py
gruebel: I think the `Optional` part is totally fine here. When we setup a fake Engine, then the bottom part will just store the parameter-value pairs in the history of the engine state.

vfdev-5 (approved):
Thanks a lot @gruebel !

PR #1509, opened 2020-12-19
[docker] Pillow -> Pillow-SIMD

### LINKED ISSUE #1489: Install Pillow-SIMD to docker images
## 🚀 Feature

Currently, we install Pillow to pytorch ignite docker images. Let's install Pillow-SIMD which could accelerate image processing if torchvision is used for that:
```bash
pip uninstall pillow
CC="cc -mavx2" pip install -U --force-reinstall pillow-simd
```

cc @fco-dv

ydcjeff:
Fixes #1489 

See: https://github.com/ydcjeff/play-gh-actions/actions/runs/432578220
https://github.com/ydcjeff/play-gh-actions/blob/ydcjeff-patch-1/Dockerfile

Description: Replace Pillow with Pillow-SIMD


vfdev-5:
Thanks for the PR @ydcjeff . First, we have to fix our automatic docker image build check. It is WIP.
Then about the PR, I think a bit more solution is to install pillow-simd and remove all additional packages. Another solution is to build it like apex in a builder. I'll come back a bit later on this PR...
Another thing I was also thinking is that main/Dockerfile.base also ships torchvision from pytorch docker image and it would make sense to replace pillow-simd too.

vfdev-5:
> First, we have to fix our automatic docker image build check

@ydcjeff , This is done now. So we can proceed with this PR.

Let's try the following solution:
- measure the size of docker image before changing Pillow -> Pillow-SIMD
- as the last docker layer: install g++ and only requirements for building Pillow (not build-essential)
- build Pillow-SIMD as usual
- remove all installed debian packages: g++ etc and run `apt autoremove`
- measure the enw size of docker image after changing Pillow -> Pillow-SIMD
- report here
What do you think ?

ydcjeff:
@vfdev-5 since `pillow` is already installed in base docker files, I replaced `pillow` with `pillow-simd` in those docker files.
It seems I don't have permission to push to circleci.

vfdev-5:
@ydcjeff sorry for delay on this PR, if you plan to continue on that, could you please do the following steps:

-  measure the size of docker image before changing Pillow -> Pillow-SIMD

-  edit dockerfile by adding g++ and building Pillow-SIMD as you did +
    remove all installed debian packages: g++ etc and run apt autoremove (currently missing)

- measure the new size of docker image after changing Pillow -> Pillow-SIMD and report it here

ydcjeff:
@vfdev-5 
- old docker image size (pytorchignite/base): 9.97 GB [[link]](https://app.circleci.com/pipelines/github/pytorch/ignite/1315/workflows/abec0e17-9f3c-45f4-8448-caf4716640e3/jobs/3730/parallel-runs/0/steps/0-105)
- new docker image size (pytorchignite/base): 6.27 GB [[link]](https://github.com/ydcjeff/play-gh-actions/runs/1893776414?check_suite_focus=true#step:4:4) 

[[Dockerfile]](https://github.com/ydcjeff/play-gh-actions/blob/pillow-simd/Dockerfile)

vfdev-5:
Thanks a lot @ydcjeff ! Just realized that pytorch finally released 1.7.1 docker image.

I checked that from my side and seems like the temporary hack to install pytorch 1.7.1 on 1.7.0 image costs space. Removing it and using now available 1.7.1 docker image then base image is 
```
pytorchignite/base             latest                          d59f896e06f0        About a minute ago   6.12GB
```

I checked also if autoremove removes something
```
    apt-get -y install --no-install-recommends git && \
    apt-get autoremove -y && \    # <------ this one
    rm -rf /var/lib/apt/lists/*
```
and image size is still 6.12 GB.

To handle properly installing and removing of g++ we have to do that in the same docker layer:
```
# replace pillow with pillow-simd
RUN apt-get update && apt-get -y install --no-install-recommends g++ && \
    pip uninstall -y pillow && \
    CC="cc -mavx2" pip install --upgrade --no-cache-dir --force-reinstall pillow-simd && \
    apt-get remove -y g++ && \
    apt-get autoremove -y && \
    rm -rf /var/lib/apt/lists/*
```
Doing that we'll have 
```
pytorchignite/base              latest                          bc9b095457ed        43 seconds ago      6.14GB
```

vfdev-5:
@ydcjeff this way your docker image size will be still as it has g++. You should have 6.24Gb for base image. Please, do it I suggested above. Thanks

vfdev-5 (approved):
Thanks @ydcjeff !

PR #1516, opened 2020-12-21
Docker image update to 1.7.0-cuda11.0-cudnn8

vfdev-5:
* Fixed Apex installation

@fco-dv please update pytorch version

Check list:

### DIFF
--- a/docker/hvd/build_all.sh
+++ b/docker/hvd/build_all.sh
@@ -34,11 +34,15 @@ image_tag=""
 if [[ -z "${PTH_VERSION}" ]]; then
     echo "PTH_VERSION is not set"
     exit 1
-else
-    pth_version=${PTH_VERSION}
 fi
 
-hvd_version=${HVD_VERSION:-v0.21.0}
+if [[ -z "${HVD_VERSION}" ]]; then
+    echo "HVD_VERSION is not set"
+    exit 1
+fi
+
+pth_version=${PTH_VERSION}
+hvd_version=${HVD_VERSION}
 
 for image_name in "hvd-base" "hvd-vision" "hvd-nlp" "hvd-apex" "hvd-apex-vision" "hvd-apex-nlp"
 do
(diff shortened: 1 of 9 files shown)


vfdev-5:
@fco-dv let's remove all those default values as it leads to a lot of confusion: from bash and in dockerfiles.

To set up pytorch version it is here (single entry):
- https://github.com/pytorch/ignite/blob/3affd95f9480b4f6a9070d800d8505bc3e80e0dc/.circleci/config.yml#L23 
Let's also print an error in bash if env var is not defined.

docker/hvd/build_all.sh
```diff
@@ -30,7 +30,14 @@ cd $curr_dir/hvd
 set -eu
 
 image_tag=""
-pth_version=${PTH_VERSION:-1.6.0-cuda10.1-cudnn7}
+
+if [[ -z "${PTH_VERSION}" ]]; then
+    echo "PTH_VERSION is not set"
+    exit 1
+else
```
vfdev-5, line 37: We can define `pth_version=${PTH_VERSION}` without else for better understanding.

docker/hvd/build_all.sh
```diff
@@ -30,7 +30,14 @@ cd $curr_dir/hvd
...
+    pth_version=${PTH_VERSION}
+fi
+
 hvd_version=${HVD_VERSION:-v0.21.0}
```
vfdev-5, line 41: Let's remove default values for HVD_VERSION and MSDP_VERSION too

docker/hvd/build_all.sh
fco-dv: yes of course

docker/hvd/build_all.sh
fco-dv: yes indeed

vfdev-5:
@fco-dv I asked your review as this PR is initialized by myself and I can not approve my own PR :)

fco-dv (approved):
LGTM

PR #1519, opened 2020-12-22
Add None check for max_epochs

### LINKED ISSUE #1513: Improve typing in engine
## 🚀 Feature

The following code may be fail if `max_epochs is None`:

https://github.com/pytorch/ignite/blob/a756f9ca956ee0ee72bb61401a87a6eaf249dbba/ignite/engine/engine.py#L839-L842

https://github.com/pytorch/ignite/blob/a756f9ca956ee0ee72bb61401a87a6eaf249dbba/ignite/engine/engine.py#L562-L566

Maybe, we have to add additional check of `max_epochs is not None` and remove `# type: ignore[operator]`.

@gruebel what do you think ?

gruebel:
Fixes #1513 

I hope it fits your expectations 😄 

The `param_scheduler` change was due to flake8 warning of 120+ chars

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -561,7 +561,8 @@ class Engine(Serializable):
         is_done_iters = state.max_iters is not None and state.iteration >= state.max_iters
         is_done_count = (
             state.epoch_length is not None
-            and state.iteration >= state.epoch_length * state.max_epochs  # type: ignore[operator]
+            and state.max_epochs is not None
+            and state.iteration >= state.epoch_length * state.max_epochs
         )
         is_done_epochs = state.max_epochs is not None and state.epoch >= state.max_epochs
         return is_done_iters or is_done_count or is_done_epochs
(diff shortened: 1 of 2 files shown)


ignite/engine/engine.py
```diff
@@ -833,12 +834,17 @@ def _run_once_on_dataset(self) -> float:
                     # Should exit while loop if we can not iterate
                     if should_exit:
                         if not self._is_done(self.state):
+                            total_iters = (
+                                self.state.epoch_length * self.state.max_epochs
+                                if self.state.max_epochs is not None
+                                else None
```
vfdev-5, line 840: Normally, either `state.max_epochs` or `state.max_iters` is defined. We can set `else self.state.max_iters` instead of None.

vfdev-5 (commented):
Thanks for the PR @gruebel !
I left a comment about how to define `total_iters`.

Issue #1496, opened 2020-12-14
Replace string .format with f-string


vfdev-5:
## 🚀 Feature

Once #1495 is merged we can replace all `.format` with f-strings.

If interested in helping with the issue, please comment here and send a PR according to our contributing guide.
Feel free to ask questions for details if needed.

PR #1523, opened 2020-12-22
Format small

uribgp:
Related to #1496

Description: Breaking up large pull

Check list:


vfdev-5 (approved):
Thanks a lot for the PRs @uribgp ! 
Another chunk of 15 files is done. It should remain ~2 chunks more...

PR #1532, opened 2020-12-25
[WIP] Static events filter

vfdev-5:
Description:

Will enable the following:
```python
events = Events.STARTED(args.start_validation) | Events.COMPLETED | Events.EPOCH_COMPLETED(every=3)

trainer = ...
@trainer.on(events)
def run_validation(trainer):
    # do something
```

To simplify this implementation in a real use-case :

https://github.com/huggingface/transfer-learning-conv-ai/blob/d4c760731e8f062abd3d61ccdc8fa29ccb5e3485/train.py#L212-L215

```python
trainer.add_event_handler(Events.EPOCH_COMPLETED, lambda _: evaluator.run(val_loader))
if args.n_epochs < 1:
    trainer.add_event_handler(Events.COMPLETED, lambda _: evaluator.run(val_loader))
if args.eval_before_start:
    trainer.add_event_handler(Events.STARTED, lambda _: evaluator.run(val_loader))
```
which can be 
```python
e = Events.EPOCH_COMPLETED
if args.n_epochs < 1:
    e |= Events.COMPLETED
if args.eval_before_start:
    e |= Events.STARTED
trainer.add_event_handler(e, lambda _: evaluator.run(val_loader))
```

and with the PR will be 
```python
e = Events.EPOCH_COMPLETED | Events.COMPLETED(args.n_epochs < 1) | Events.STARTED(args.eval_before_start)
@trainer.on(e):
    evaluator.run(val_loader)
```

TODO:
- Maybe, implementation can be done differently without a custom event_filter.
- How about `Events.EPOCH_COMPLETED(args.condition, every=3)` ?

Check list:


vfdev-5:
@sdesrozis any comments on the idea ? At first I was thinking that this is good. Now, the implementation (not yet committed) would require some `_SkippedEvent` class and we can potentially do
```python
Events.EPOCH_COMPLETED(args.condition)  -> if True: Events.EPOCH_COMPLETED else _SkippedEvent (which will be ignored at `Engine.add_event_handler`)
Events.EPOCH_COMPLETED(args.condition, every=3)  -> if True: Events.EPOCH_COMPLETED(every=3) else _SkippedEvent
```
and I'm hesitating about that ...

sdesrozis:
I think that is a good PR and having a way to skip events is interesting and surely mandatory.

Btw the hard issue is to avoid handler to be triggered multiple times at the same event. Maybe this should help...

PR #1544, opened 2021-01-05
Use events list for logger handlers

### LINKED ISSUE #1543: Allow to attach output handler using event list
## 🚀 Feature

Attaching output handlers using event lists is not possible. The idea is to make it work.

See comments https://github.com/pytorch/ignite/issues/1541#issuecomment-753923598

sdesrozis:
Fixes #1543 

Description:

Allow `EventsList` in `BaseLogger` attachment. The key point is to attach each event independently as it is done in `Engine`.

Check list:

### DIFF
--- a/ignite/contrib/handlers/base_logger.py
+++ b/ignite/contrib/handlers/base_logger.py
@@ -155,8 +155,8 @@ class BaseLogger(metaclass=ABCMeta):
             engine (Engine): engine object.
             log_handler (callable): a logging handler to execute
             event_name: event to attach the logging handler to. Valid events are from
-                :class:`~ignite.engine.events.Events` or any `event_name` added by
-                :meth:`~ignite.engine.engine.Engine.register_events`.
+                :class:`~ignite.engine.events.Events` or class:`~ignite.engine.events.EventsList` or any `event_name`
+                added by :meth:`~ignite.engine.engine.Engine.register_events`.
 
         Returns:
             :class:`~ignite.engine.RemovableEventHandle`, which can be used to remove the handler.
--- a/ignite/contrib/handlers/base_logger.py
+++ b/ignite/contrib/handlers/base_logger.py
@@ -7,7 +7,7 @@ import torch
 import torch.nn as nn
 from torch.optim import Optimizer
 
-from ignite.engine import Engine, Events, State
+from ignite.engine import Engine, Events, EventsList, State
 from ignite.engine.events import CallableEventWithFilter, RemovableEventHandle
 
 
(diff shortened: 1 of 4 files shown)


ignite/contrib/handlers/base_logger.py
```diff
@@ -161,12 +161,20 @@ def attach(
         Returns:
             :class:`~ignite.engine.RemovableEventHandle`, which can be used to remove the handler.
         """
-        name = event_name
+        if isinstance(event_name, EventsList):
+            for name in event_name:
+                if name not in State.event_to_attr:
+                    raise RuntimeError(f"Unknown event name '{name}'")
```
ydcjeff, line 167: this line is not covered by tests

ydcjeff (approved):
LGTM, Thank you!

ignite/contrib/handlers/base_logger.py
sdesrozis: You are right even if it implies hacking `EventsList` in order to raise this exception. Btw python so permissive that I add it !!

That you for this catch !

ignite/contrib/handlers/base_logger.py
```diff
@@ -147,26 +147,34 @@ class BaseLogger(metaclass=ABCMeta):
...
-                :class:`~ignite.engine.events.Events` or any `event_name` added by
-                :meth:`~ignite.engine.engine.Engine.register_events`.
+                :class:`~ignite.engine.events.Events` or class:`~ignite.engine.events.EventsList` or any `event_name`
+                added by :meth:`~ignite.engine.engine.Engine.register_events`.
 
         Returns:
             :class:`~ignite.engine.RemovableEventHandle`, which can be used to remove the handler.
         """
-        name = event_name
+        if isinstance(event_name, EventsList):
```
vfdev-5, line 164: I think it would be better to use `return engine.add_event_handler(event_name, log_handler, self, event_name)` for any type of events. Below for loop if considered without check `if name not in State.event_to_attr` is exactly what is done by `Engine.add_event_handler`.
I think we can simply generalize event checking like
```python
event_to_check = event_name if isinstance(event_name, EventsList) else [event_name, ]
for name in event_to_check:
    if name not in State.event_to_attr:
        raise RuntimeError(f"Unknown event name '{name}'")

return engine.add_event_handler(event_name, log_handler, self, event_name)
```

sdesrozis: That does not work.

The last argument of the method `add_event_handler` is a list of events. Internally in engine, there is a loop where the handler is attached at every event in the list. But the list is forwarded to each handler while just each event should be.

If there was no argument to the method, it would work. 

Thoughts ?

vfdev-5: Yes, you are right. In this case, my suggestion is not valid. OK, let's do it as you proposed.

vfdev-5 (approved):
Thanks for the PR @sdesrozis !

PR #1551, opened 2021-01-10
remove old configs leftover from removal of py3.5/py2

ydcjeff:
Description: remove old configs leftover from removal of py3.5/py2

- removed `.travis.yaml`
- Changed/remove leftover python3.5
- removed comments that's no longer true
- regular import of `skimage` after py3.5 removal
- remove python version from requirement-dev.txt
- `universal` no longer needed since ignite only supports >= py3.6


vfdev-5:
> universal no longer needed since ignite only supports >= py3.6

Could you please give a reference about why we can omit universal for >= py3.6 ?

ydcjeff:
> > universal no longer needed since ignite only supports >= py3.6
> 
> Could you please give a reference about why we can omit universal for >= py3.6 ?

https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels

vfdev-5 (approved):
Thanks @ydcjeff

PR #1560, opened 2021-01-19
ClearMLLogger: add older Trains SDK backwards compatibility

jkhenning:
Description:
Add backwards-compatibility so that users who still use older Trains SDK packages can still use the ClearMLLogger (formerly TrainsLogger).
Tested using `mnist_with_clearml_logger.py`.
Obviously, any test explicitly importing `clearml` will neither be affected nor will support backwards compatibility.
Runtime errors were kept as-is since there's no reason to instruct users to install anything but the latest ClearML SDK in case it is required.

Check list:

### DIFF
--- a/examples/contrib/cifar10/main.py
+++ b/examples/contrib/cifar10/main.py
@@ -48,7 +48,11 @@ def training(local_rank, config):
             config["cuda device name"] = torch.cuda.get_device_name(local_rank)
 
         if config["with_clearml"]:
-            from clearml import Task
+            try:
+                from clearml import Task
+            except ImportError:
+                # Backwards-compatibility for legacy Trains SDK
+                from trains import Task
 
             task = Task.init("CIFAR10-Training", task_name=output_path.stem)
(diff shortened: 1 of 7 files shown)


vfdev-5 (commented):
Thanks for the PR @jkhenning !
Could you please add the following to mypy.ini
```
[mypy-trains.*]
ignore_missing_imports = True
```
as it was previously (https://github.com/pytorch/ignite/commit/2485fd42c6ef4d3e97fd606a52f8c6e5d940357e#diff-6f2d4ba9ca9a357d31014946667b7bed1bfdbc6d2530afc77778fa0a36bee457). Otherwise mypy is failing.

jkhenning:
Sure @vfdev-5 , my bad for missing that 🙂

vfdev-5 (approved):
Thanks @jkhenning ! Looks good !

ydcjeff (approved):
Thank you @jkhenning LGTM too

PR #1565, opened 2021-01-20
Improved max_iters handling

### LINKED ISSUE #1521: Possible issues with max_iters when loading/saving engine's state
Currently Engine implicitly assumes : 
https://github.com/pytorch/ignite/blob/e3ef192c94d0d793a9303bec915fb846aaa3161f/ignite/engine/engine.py#L122

but we recently introduced possibility to trigger a run with `max_iters`. If epoch length is unknown, saving and reloading engine's state probably wont work. We have to ensure that everything still works as expected.

@thescripted would you be interested in looking into this ?

vfdev-5:
Fixes #1521

Description:
- bug fixes
- state_dict/load_state_dict refactor
- added helper `Engine.debug()` method
- more tests

Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -5,7 +5,7 @@ import warnings
 import weakref
 from collections import OrderedDict, defaultdict
 from collections.abc import Mapping
-from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Tuple, Union, Sized
 
 from torch.utils.data import DataLoader
 
@@ -117,10 +117,11 @@ class Engine(Serializable):
 
     """
 
-    _state_dict_all_req_keys = ("epoch_length", "max_epochs")
-    _state_dict_one_of_opt_keys = ("iteration", "epoch")
+    _state_dict_all_req_keys = ("epoch_length", )
+    _state_dict_one_of_opt_keys = (("iteration", "epoch"), ("max_epochs", "max_iters"))
 
     def __init__(self, process_function: Callable):
+        super(Engine, self).__init__()
         self._event_handlers = defaultdict(list)  # type: Dict[Any, List]
         self.logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
         self._process_function = process_function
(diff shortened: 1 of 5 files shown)


ignite/engine/engine.py
```diff
@@ -118,18 +117,18 @@ def compute_mean_std(engine, batch):
 
     """
 
-    _state_dict_all_req_keys = ("epoch_length", "max_epochs")
-    _state_dict_one_of_opt_keys = ("iteration", "epoch")
+    _state_dict_all_req_keys = ("epoch_length",)
```
sadra-barikbin, line 120: But if we interrupted the engine during the first epoch, we would not have `epoch_length`.

ignite/engine/engine.py
```diff
@@ -118,18 +117,18 @@ def compute_mean_std(engine, batch):
...
+    _state_dict_one_of_opt_keys = (("iteration", "epoch",), ("max_epochs", "max_iters",))
```
sadra-barikbin, line 121: We can relax the constraint that the both `max_epochs` and `max_iters` could not have values.

ignite/engine/engine.py
```diff
@@ -545,7 +541,7 @@ def load_state_dict(self, state_dict: Mapping) -> None:
             self.state.epoch = 0
             if self.state.epoch_length is not None:
                 self.state.epoch = self.state.iteration // self.state.epoch_length
-        elif "epoch" in state_dict:
+        else:
```
sadra-barikbin, line 544: According to `state_dict` this should never happen.

ignite/engine/engine.py
```diff
@@ -728,6 +713,53 @@ def _get_data_length(self, data: Iterable) -> Optional[int]:
...
+                raise ValueError(
+                    "Argument max_iters should be larger than the current iteration "
+                    f"defined in the state: {max_iters} vs {self.state.iteration}. "
+                    "Please, set engine.state.max_iters = None "
+                    "before calling engine.run() in order to restart the training from the beginning."
+                )
+            self.state.max_iters = max_iters
+
+    def _check_and_set_epoch_length(self, data: Iterable, epoch_length: Optional[int] = None) -> None:
+        # Can't we accept a redefinition ?
```
sadra-barikbin, line 743: I think we can't because the relationship between epoch and iteration becomes invalid then. By the way in the case `epoch_length` is changed, we could reinitialize the state (with epoch and iteration zero along with a message to the user) as well.

ignite/engine/engine.py
```diff
@@ -873,3 +903,14 @@ def _run_once_on_dataset(self) -> float:
             self._handle_exception(e)
 
         return time.time() - start_time
+
+    def debug(self, enabled: bool = True) -> None:
```
sadra-barikbin, line 907: Do you still want to merge this method into the master?

PR #1567, opened 2021-01-22
Fix incorrect design of handlers (Frequency, RunningAverage, etc.)

### LINKED ISSUE #1405: Fix incorrect design for RunningAverage, Frequency, ModelCheckpoint, ...

sdesrozis:
Fixes #1405 

Description:

* `Frequency` : introduction of a specific usage `FrequencyWise` to fit the `Metric.attach()` api. 

Check list:

### DIFF
--- a/ignite/metrics/frequency.py
+++ b/ignite/metrics/frequency.py
@@ -3,9 +3,19 @@ from typing import Any, Callable, Union
 import torch
 
 import ignite.distributed as idist
-from ignite.engine import Engine, Events
+from ignite.engine import Engine, Events, CallableEventWithFilter
 from ignite.handlers.timing import Timer
-from ignite.metrics.metric import EpochWise, Metric, MetricUsage, reinit__is_reduced, sync_all_reduce
+from ignite.metrics.metric import Metric, MetricUsage, reinit__is_reduced, sync_all_reduce
+
+
+class FrequencyWise(MetricUsage):
+
+    def __init__(self, event: CallableEventWithFilter = Events.ITERATION_COMPLETED) -> None:
+        super(FrequencyWise, self).__init__(
+            started=Events.EPOCH_STARTED,
+            completed=event,
+            iteration_completed=Events.ITERATION_COMPLETED,
+        )
 
 
 class Frequency(Metric):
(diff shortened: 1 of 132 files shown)


ignite/metrics/frequency.py
```diff
@@ -58,21 +68,9 @@ def update(self, output: int) -> None:
...
-
-    # TODO: see issue https://github.com/pytorch/ignite/issues/1405
-    def attach(  # type: ignore
-        self, engine: Engine, name: str, event_name: Events = Events.ITERATION_COMPLETED
-    ) -> None:
-        engine.add_event_handler(Events.EPOCH_STARTED, self.started)
-        engine.add_event_handler(Events.ITERATION_COMPLETED, self.iteration_completed)
-        engine.add_event_handler(event_name, self.completed, name)
+    # override the method attach() of Metrics to define a different default value for usage
+    def attach(self, engine: Engine, name: str, usage: Union[str, MetricUsage] = FrequencyWise()) -> None:
```
vfdev-5, line 75: Maybe, we can try to use [typing overload](https://python.readthedocs.io/en/stable/library/typing.html#typing.overload) to keep BC with the previous signature ?

vfdev-5 (commented):
Looks good, thanks @sdesrozis !

ignite/metrics/frequency.py
sdesrozis: I took a look but it doesn't seem simple to use typing overload because argument names are different (to fit the `Metrics.attach()` method).

vfdev-5:
@sdesrozis thanks for the update.I was thinking as Checkpoint and ModelCheckpoint fix is the most simple and we do not break other things, it would be better to do it in a separate PR. For others, I'm not yet sure about the fix to adopt.

sdesrozis:
Ok so if I understand correctly, I do a new PR with last commits about ModelCheckpoint fix, right ? And let's see for others ?

vfdev-5:
Exactly!

vfdev-5:
To be honest currently I'm not that sure about introducing things in this way. Let's try to find a better design that you already told previously about... I'd prefer to keep things as is instead of introducing a half-good solution. What do you think ?

sdesrozis:
Not relevant at the moment. I prefer close this PR.

PR #1570, opened 2021-01-23
[metrics] add PSNR

### LINKED ISSUE #1568: [metrics] add PSNR
## 🚀 Feature
I would like to add PSNR metric which is used in super resolution applications.
Implementation will be based from:

https://github.com/pytorch/examples/blob/792d336019a28a679e29cf174e10cee80ead8722/super_resolution/main.py#L69-L70

Though it's relatively easy to implement on our own, having in `ignite` will benefit everyone else.

ydcjeff:
Fixes #1568 

Description: add PSNR metric

Check list:
https://deploy-preview-1570--pytorch-ignite-preview.netlify.app/metrics.html#ignite.metrics.PSNR

### DIFF
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -57,7 +57,7 @@ class PSNR(Metric):
 
     @reinit__is_reduced
     def reset(self) -> None:
-        self._sum_of_batchwise_psnr = 0
+        self._sum_of_batchwise_psnr = 0  # type: Union[float, torch.Tensor]
         self._num_examples = 0
 
     @reinit__is_reduced
@@ -101,4 +101,4 @@ class PSNR(Metric):
     def compute(self) -> torch.Tensor:
         if self._num_examples == 0:
             raise NotComputableError("PSNR must have at least one example before it can be computed.")
-        return torch.sum(self._sum_of_batchwise_psnr / self._num_examples)
+        return torch.sum(self._sum_of_batchwise_psnr / self._num_examples)  # type: ignore[arg-type]
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -16,8 +16,10 @@ class PSNR(Metric):
 
     where :math:`\text{MSE}` is `mean squared error <https://en.wikipedia.org/wiki/Mean_squared_error>`_.
 
+    - `y_pred` and `y` must have (batch_size, ...) shape.
+    - `y_pred` and `y` must have same dtype and same shape.
+
     Note:
-        `y_pred` and `y` must have same dtype and same shape.
         If `y_pred` and `y` have almost identical values, the result will be infinity.
 
     Args:
@@ -58,7 +60,7 @@ class PSNR(Metric):
         device: Union[str, torch.device] = torch.device("cpu"),
     ):
         super().__init__(output_transform=output_transform, device=device)
-        self.data_range = self.__data_range = data_range
+        self.data_range = data_range
 
     def _check_shape_dtype(self, output: Sequence[torch.Tensor]) -> None:
         y_pred, y = output
@@ -74,15 +76,16 @@ class PSNR(Metric):
 
     @reinit__is_reduced
     def reset(self) -> None:
-        self._sum_of_squared_error = 0  # type: Union[int, torch.Tensor]
-        self._num_elements = 0
+        self._sum_of_batchwise_psnr = 0  # type: Union[int, torch.Tensor]
+        self._num_examples = 0
 
     @reinit__is_reduced
     def update(self, output: Sequence[torch.Tensor]) -> None:
         self._check_shape_dtype(output)
         y_pred, y = output[0].detach(), output[1].detach()
+        data_range = self.data_range
 
-        if self.__data_range is None:
+        if data_range is None:
             dmin, dmax = _dtype_range[y.dtype]
             true_min, true_max = y.min(), y.max()
             if true_max > dmax or true_min < dmin:
@@ -92,22 +95,19 @@ class PSNR(Metric):
                 )
             if true_min >= 0:
                 # most common case (255 for uint8, 1 for float)
-                self.data_range = dmax
+                data_range = dmax
             else:
-                self.data_range = dmax - dmin
+                data_range = dmax - dmin
 
-        self._sum_of_squared_error += (
-            torch.pow(y_pred.double() - y.view_as(y_pred).double(), 2).sum().to(device=self._device)
-        )
-        self._num_elements += y.numel()
+        rmse_error = torch.pow(y_pred.double() - y.view_as(y_pred).double(), 2).mean(dim=list(range(1, y.ndim))).sqrt()
+        self._sum_of_batchwise_psnr += 20.0 * torch.log10(data_range / rmse_error).to(self._device)
+        self._num_examples += y.shape[0]
 
-    @sync_all_reduce("_sum_of_squared_error", "_num_elements")
-    def compute(self) -> Union[float, torch.Tensor]:
-        if self._num_elements == 0:
+    @sync_all_reduce("_sum_of_batchwise_psnr", "_num_examples")
+    def compute(self) -> torch.Tensor:
+        if self._num_examples == 0:
             raise NotComputableError("PSNR must have at least one example before it can be computed.")
-        return 10.0 * torch.log10(
-            self.data_range ** 2 / (self._sum_of_squared_error / self._num_elements)  # type: ignore[operator, arg-type]
-        )
+        return torch.sum(self._sum_of_batchwise_psnr / self._num_examples)  # type: ignore[arg-type]
 
 
 _dtype_range = {
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -24,7 +24,7 @@ class PSNR(Metric):
         - If `y_pred` and `y` have almost identical values, the result will be infinity.
 
     Args:
-        data_range (int or float): The data range of the target image (distance between minimum
+        data_range (int or float, optional): The data range of the target image (distance between minimum
             and maximum possible values). If not provided, it will be estimated from the input
             data type: ``1.0`` for float/double tensor or ``255`` for unsigned 8-bit tensor.
             For other data types, please set the data range, otherwise an exception will be raised.
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -94,7 +94,9 @@ class PSNR(Metric):
                 self.data_range = dmax - dmin
 
         mse_error = F.mse_loss(y_pred, y)
-        self._sum_of_batchwise_psnr += 10.0 * torch.log10((self.data_range ** 2) / mse_error)
+        self._sum_of_batchwise_psnr += 10.0 * torch.log10((self.data_range ** 2) / mse_error).to(
+            dtype=torch.float64, device=self._device
+        )
         self._num_examples += y.shape[0]
 
     @sync_all_reduce("_sum_of_batchwise_psnr", "_num_examples")
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -64,7 +64,7 @@ class PSNR(Metric):
         super().__init__(output_transform=output_transform, device=device)
         self.data_range = data_range
 
-    def _check_shape_dtype(self, output: Sequence[torch.Tensor]) -> None:
+    def _check_shape_dtype_drange(self, output: Sequence[torch.Tensor], data_range: Union[int, float]) -> None:
         y_pred, y = output
         if y_pred.dtype != y.dtype:
             raise TypeError(
@@ -75,17 +75,6 @@ class PSNR(Metric):
             raise ValueError(
                 f"Expected y_pred and y to have the same shape. Got y_pred: {y_pred.shape} and y: {y.shape}."
             )
-
-    @reinit__is_reduced
-    def reset(self) -> None:
-        self._sum_of_batchwise_psnr = torch.tensor(0.0, dtype=torch.float64, device=self._device)
-        self._num_examples = 0
-
-    @reinit__is_reduced
-    def update(self, output: Sequence[torch.Tensor]) -> None:
-        self._check_shape_dtype(output)
-        y_pred, y = output[0].detach(), output[1].detach()
-        data_range = self.data_range
 
         if data_range is None:
             try:
@@ -105,6 +94,18 @@ class PSNR(Metric):
                 data_range = dmax
             else:
                 data_range = dmax - dmin
+        return data_range
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._sum_of_batchwise_psnr = torch.tensor(0.0, dtype=torch.float64, device=self._device)
+        self._num_examples = 0
+
+    @reinit__is_reduced
+    def update(self, output: Sequence[torch.Tensor]) -> None:
+        data_range = self.data_range
+        data_range = self._check_shape_dtype_drange(output, data_range)
+        y_pred, y = output[0].detach(), output[1].detach()
 
         dim = tuple(range(1, y.ndim))
         mse_error = torch.pow(y_pred.double() - y.view_as(y_pred).double(), 2).mean(dim=dim)
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -65,7 +65,7 @@ class PSNR(Metric):
         self.data_range = data_range
 
     def _check_shape_dtype_drange(
-        self, output: Sequence[torch.Tensor], data_range: Union[int, float]
+        self, output: Sequence[torch.Tensor], data_range: Union[int, float, None]
     ) -> Union[int, float]:
         y_pred, y = output
         if y_pred.dtype != y.dtype:
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -1,4 +1,4 @@
-from typing import Union, Callable, Sequence, Optional
+from typing import Callable, Optional, Sequence, Union
 
 import torch
 import torch.nn.functional as F
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -54,7 +54,6 @@ class PSNR(Metric):
     ):
         super().__init__(output_transform=output_transform, device=device)
         self.data_range = data_range
-        self.__data_range = data_range
 
     def _check_shape_dtype(self, output: Sequence[torch.Tensor]) -> None:
         y_pred, y = output
@@ -77,9 +76,10 @@ class PSNR(Metric):
     def update(self, output: Sequence[torch.Tensor]) -> None:
         self._check_shape_dtype(output)
         y_pred, y = output[0].detach(), output[1].detach()
+        data_range = self.data_range
 
-        if self.__data_range is None:
-            dmin, dmax = dtype_range[y.dtype]
+        if data_range is None:
+            dmin, dmax = _dtype_range[y.dtype]
             true_min, true_max = y.min(), y.max()
             if true_max > dmax or true_min < dmin:
                 raise ValueError(
@@ -88,12 +88,12 @@ class PSNR(Metric):
                 )
             if true_min >= 0:
                 # most common case (255 for uint8, 1 for float)
-                self.data_range = dmax
+                data_range = dmax
             else:
-                self.data_range = dmax - dmin
+                data_range = dmax - dmin
 
         mse_error = torch.pow(y_pred.double() - y.view_as(y_pred).double(), 2).mean()
-        data_range_sqrt = self.data_range ** 2  # type: ignore[operator]
+        data_range_sqrt = data_range ** 2  # type: ignore[operator]
         self._sum_of_batchwise_psnr += 10.0 * torch.log10(data_range_sqrt / mse_error).to(device=self._device)
         self._num_examples += y.shape[0]
 
@@ -108,9 +108,9 @@ _int_types = (torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64)
 
 _int_ranges = {k: (torch.iinfo(k).min, torch.iinfo(k).max) for k in _int_types}
 
-dtype_range = {
+_dtype_range = {
     torch.float16: (-1, 1),
     torch.float32: (-1, 1),
     torch.float64: (-1, 1),
 }
-dtype_range.update(_int_ranges)
+_dtype_range.update(_int_ranges)
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -1,7 +1,6 @@
 from typing import Callable, Optional, Sequence, Union
 
 import torch
-import torch.nn.functional as F
 
 from ignite.exceptions import NotComputableError
 from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
@@ -10,11 +9,17 @@ __all__ = ["PSNR"]
 
 
 class PSNR(Metric):
-    """Computes `Peak signal-to-noise ratio (PSNR) <https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio>`_.
+    r"""Computes average `Peak signal-to-noise ratio (PSNR) <https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio>`_.
+
+    .. math::
+        \text{PSNR}(i, j) = 10 * \log_{10}\left(\frac{ MAX_{i}^2 }{ \text{ MSE } }\right)
+
+    - `y_pred` and `y` must be in the following shape (batch_size, ...).
+    - `y_pred` and `y` must have same dtype and same shape.
 
     Args:
-        data_range (float): Range of the image. If not provided, it will be determined
-            from the image type. Typically, ``1.0``.
+        data_range (int or float): Range of the image. If not provided, it will be determined
+            from the image type. Typically, ``1.0`` or ``255``.
         output_transform (callable, optional): A callable that is used to transform the Engine’s
             process_function’s output into the form expected by the metric.
         device (str or torch.device): specifies which device updates are accumulated on.
@@ -27,10 +32,6 @@ class PSNR(Metric):
     The output of the engine's ``process_function`` needs to be in format of
     ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y, ...}``.
 
-    Note:
-        ``y_pred`` and ``y`` have to be normalized image tensors, should have the same shape / dtype,
-        should have BxCxHxW shape.
-
     .. code-block:: python
 
         def process_function(engine, batch):
@@ -39,31 +40,22 @@ class PSNR(Metric):
         engine = Engine(process_function)
         psnr = PSNR()
         psnr.attach(engine, "psnr")
+
+    .. versionadded:: 0.5.0
     """
 
     def __init__(
         self,
-        data_range: Optional[float] = None,
+        data_range: Optional[Union[int, float]] = None,
         output_transform: Callable = lambda x: x,
         device: Union[str, torch.device] = torch.device("cpu"),
     ):
         super().__init__(output_transform=output_transform, device=device)
         self.data_range = data_range
-        self.dtype_range = {
-            torch.float16: (-1, 1),
-            torch.float32: (-1, 1),
-            torch.float64: (-1, 1),
-        }
+        self.__data_range = data_range
 
-    @reinit__is_reduced
-    def reset(self) -> None:
-        self._sum_of_batchwise_psnr = 0  # type: Union[float, torch.Tensor]
-        self._num_examples = 0
-
-    @reinit__is_reduced
-    def update(self, output: Sequence[torch.Tensor]) -> None:
-        y_pred, y = output[0].detach(), output[1].detach()
-
+    def _check_shape_dtype(self, output: Sequence[torch.Tensor]) -> None:
+        y_pred, y = output
         if y_pred.dtype != y.dtype:
             raise TypeError(
                 f"Expected y_pred and y to have the same data type. Got y_pred: {y_pred.dtype} and y: {y.dtype}."
@@ -74,13 +66,18 @@ class PSNR(Metric):
                 f"Expected y_pred and y to have the same shape. Got y_pred: {y_pred.shape} and y: {y.shape}."
             )
 
-        if y_pred.ndim != 4 or y.ndim != 4:
-            raise ValueError(
-                f"Expected y_pred and y to have BxCxHxW shape. Got y_pred: {y_pred.shape} and y: {y.shape}."
-            )
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._sum_of_batchwise_psnr = 0  # type: Union[int, torch.Tensor]
+        self._num_examples = 0
 
-        if self.data_range is None:
-            dmin, dmax = self.dtype_range[y.dtype]
+    @reinit__is_reduced
+    def update(self, output: Sequence[torch.Tensor]) -> None:
+        self._check_shape_dtype(output)
+        y_pred, y = output[0].detach(), output[1].detach()
+
+        if self.__data_range is None:
+            dmin, dmax = dtype_range[y.dtype]
             true_min, true_max = y.min(), y.max()
             if true_max > dmax or true_min < dmin:
                 raise ValueError(
@@ -93,14 +90,25 @@ class PSNR(Metric):
             else:
                 self.data_range = dmax - dmin
 
-        mse_error = F.mse_loss(y_pred, y)
-        self._sum_of_batchwise_psnr += 10.0 * torch.log10((self.data_range ** 2) / mse_error).to(
-            dtype=torch.float64, device=self._device
-        )
+        mse_error = torch.pow(y_pred.double() - y.view_as(y_pred).double(), 2).mean()
+        data_range_sqrt = self.data_range ** 2  # type: ignore[operator]
+        self._sum_of_batchwise_psnr += 10.0 * torch.log10(data_range_sqrt / mse_error).to(device=self._device)
         self._num_examples += y.shape[0]
 
     @sync_all_reduce("_sum_of_batchwise_psnr", "_num_examples")
-    def compute(self) -> torch.Tensor:
+    def compute(self) -> Union[float, torch.Tensor]:
         if self._num_examples == 0:
             raise NotComputableError("PSNR must have at least one example before it can be computed.")
-        return torch.sum(self._sum_of_batchwise_psnr / self._num_examples)  # type: ignore[arg-type]
+        return self._sum_of_batchwise_psnr / self._num_examples
+
+
+_int_types = (torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64)
+
+_int_ranges = {k: (torch.iinfo(k).min, torch.iinfo(k).max) for k in _int_types}
+
+dtype_range = {
+    torch.float16: (-1, 1),
+    torch.float32: (-1, 1),
+    torch.float64: (-1, 1),
+}
+dtype_range.update(_int_ranges)
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -42,6 +42,9 @@ class PSNR(Metric):
         engine = Engine(process_function)
         psnr = PSNR()
         psnr.attach(engine, "psnr")
+        # ...
+        state = engine.run(data)
+        print(f"PSNR: {state.metrics['psnr']}")
 
     .. versionadded:: 0.5.0
     """
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -64,7 +64,9 @@ class PSNR(Metric):
         super().__init__(output_transform=output_transform, device=device)
         self.data_range = data_range
 
-    def _check_shape_dtype_drange(self, output: Sequence[torch.Tensor], data_range: Union[int, float]) -> Union[int, float]:
+    def _check_shape_dtype_drange(
+        self, output: Sequence[torch.Tensor], data_range: Union[int, float]
+    ) -> Union[int, float]:
         y_pred, y = output
         if y_pred.dtype != y.dtype:
             raise TypeError(
--- a/ignite/metrics/psnr.py
+++ b/ignite/metrics/psnr.py
@@ -16,12 +16,14 @@ class PSNR(Metric):
 
     where :math:`\text{MSE}` is `mean squared error <https://en.wikipedia.org/wiki/Mean_squared_error>`_.
 
-    - `y_pred` and `y` must be in the following shape (batch_size, ...).
-    - `y_pred` and `y` must have same dtype and same shape.
+    Note:
+        `y_pred` and `y` must have same dtype and same shape.
+        If `y_pred` and `y` have almost identical values, the result will be infinity.
 
     Args:
-        data_range (int or float): Range of the image. If not provided, it will be determined
-            from the image type. Typically, ``1.0`` or ``255``.
+        data_range (int or float): The data range of the target image (distance between minimum
+            and maximum possible values). If not provided, it will be estimated from the input
+            data type. Typically, ``1.0`` for float tensor or ``255`` for unsigned 8-bit tensor.
         output_transform (callable, optional): A callable that is used to transform the Engine’s
             process_function’s output into the form expected by the metric.
         device (str or torch.device): specifies which device updates are accumulated on.
@@ -56,7 +58,7 @@ class PSNR(Metric):
         device: Union[str, torch.device] = torch.device("cpu"),
     ):
         super().__init__(output_transform=output_transform, device=device)
-        self.data_range = data_range
+        self.data_range = self.__data_range = data_range
 
     def _check_shape_dtype(self, output: Sequence[torch.Tensor]) -> None:
         y_pred, y = output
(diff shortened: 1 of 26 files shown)


ignite/metrics/psnr.py
```diff
@@ -0,0 +1,104 @@
...
+        output_transform: Callable = lambda x: x,
+        device: Union[str, torch.device] = torch.device("cpu"),
+    ):
+        super().__init__(output_transform=output_transform, device=device)
+        self.data_range = data_range
+        self.dtype_range = {
+            torch.float16: (-1, 1),
+            torch.float32: (-1, 1),
+            torch.float64: (-1, 1),
+        }
```
ydcjeff, line 56: should we have `dtype_range` similar to
https://github.com/scikit-image/scikit-image/blob/master/skimage/util/dtype.py#L18-L33

Currently we only need `dtype_range` for `PSNR` and `SSIM` metrics AFAIK.

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,104 @@
...
+                    "y has intensity values outside the range expected "
+                    "for its data type. Please manually specify the `data_range`."
+                )
+            if true_min >= 0:
+                # most common case (255 for uint8, 1 for float)
+                self.data_range = dmax
+            else:
+                self.data_range = dmax - dmin
+
+        mse_error = F.mse_loss(y_pred, y)
```
ydcjeff, line 96: `mse_loss` doesn't accept `ByteTensor` so input tensors must be normalized tensors.
So should we allow both un-normalized and normalized tensors?

ydcjeff (commented):
@vfdev-5  I have drafted a PR. There are few things we need to clear about.

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,106 @@
...
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._sum_of_batchwise_psnr = 0  # type: Union[float, torch.Tensor]
+        self._num_examples = 0
+
+    @reinit__is_reduced
+    def update(self, output: Sequence[torch.Tensor]) -> None:
+        y_pred, y = output[0].detach(), output[1].detach()
+
+        if y_pred.dtype != y.dtype:
```
vfdev-5, line 67: Let's put all check into a separate private method.

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,106 @@
...
+            raise TypeError(
+                f"Expected y_pred and y to have the same data type. Got y_pred: {y_pred.dtype} and y: {y.dtype}."
+            )
+
+        if y_pred.shape != y.shape:
+            raise ValueError(
+                f"Expected y_pred and y to have the same shape. Got y_pred: {y_pred.shape} and y: {y.shape}."
+            )
+
+        if y_pred.ndim != 4 or y.ndim != 4:
```
vfdev-5, line 77: Is it a hard requirement ? We may accept probably, 2d, 3d inputs ?

ignite/metrics/psnr.py
sdesrozis: Is this really useful considering that all min/max values are similar ?

In `slearn`, it makes sense because différent `dtype` are handled. Maybe I'm missing something?

tests/ignite/metrics/test_psnr.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def test_invalid_psnr():
+    y_pred = torch.rand(1, 3, 8, 8)
+    y = torch.rand(1, 3, 8, 8)
+
+    with pytest.raises(TypeError, match=r"Expected y_pred and y to have the same data type."):
+        psnr = PSNR()
+        psnr.update((y_pred, y.double()))
+        psnr.compute()
```
vfdev-5, line 29: I think this command is unnecessary as exception will be caught on `update`

vfdev-5:
Thanks for the PR @ydcjeff !

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,106 @@
...
+                    "for its data type. Please manually specify the `data_range`."
+                )
+            if true_min >= 0:
+                # most common case (255 for uint8, 1 for float)
+                self.data_range = dmax
+            else:
+                self.data_range = dmax - dmin
+
+        mse_error = F.mse_loss(y_pred, y)
+        self._sum_of_batchwise_psnr += 10.0 * torch.log10((self.data_range ** 2) / mse_error).to(
```
vfdev-5, line 97: we maybe avoid power operation with multiplying by 20 : 
```suggestion
        self._sum_of_batchwise_psnr += 20.0 * torch.log10(self.data_range / mse_error).to(
```

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,106 @@
...
+import torch.nn.functional as F
+
+from ignite.exceptions import NotComputableError
+from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
+
+__all__ = ["PSNR"]
+
+
+class PSNR(Metric):
+    """Computes `Peak signal-to-noise ratio (PSNR) <https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio>`_.
```
vfdev-5, line 13:
```suggestion
    """Computes average `Peak signal-to-noise ratio (PSNR) <https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio>`_.
```

sdesrozis:
Hey @ydcjeff ! Thank you, I left some comments.

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Args:
+        data_range (float): Range of the image. If not provided, it will be determined
```
vfdev-5, line 16: Let me get some info about how to deal with n-channels images and when channel do not have the same range (for example, HSV color space).

ignite/metrics/psnr.py
vfdev-5: Maybe, we can check and suppose that `y_pred` and `y` are float. Un-normalized case is more or less covered with data_range, right ?

vfdev-5:
@lidq92 could you please take a look this PR ?

ignite/metrics/psnr.py
ydcjeff: if the difference between the min and max values of the image (float type) is low (0.5/6/7), the final value is different from `sklearn` (starting from 2/3 decimal places.)

ignite/metrics/psnr.py
ydcjeff: As `F.mse_loss` doesn't allow `ByteTensor`, should we copy the code from `ignite.metrics.MeanSquaredError`? It should be able to handle `ByteTensor`.

ignite/metrics/psnr.py
ydcjeff: Ahh yes, we can accept 2/3/4 dim inputs

ignite/metrics/psnr.py
ydcjeff: If we do `20`, we have to find `sqrt` of `mse_error` as per https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio

vfdev-5: That's true, my bad. Let's keep square here and 10.

ignite/metrics/psnr.py
vfdev-5: I think @sdesrozis 's comment is about as the same data range: `(-1, 1)` for all floating types, we can directly put it without using dictionary structure: `dmin, dmax = -1.0, 1.0` (line 83). That, there is still a valid question if input dtype is not float ...

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,106 @@
...
+            raise ValueError(
+                f"Expected y_pred and y to have the same shape. Got y_pred: {y_pred.shape} and y: {y.shape}."
+            )
+
+        if y_pred.ndim != 4 or y.ndim != 4:
+            raise ValueError(
+                f"Expected y_pred and y to have BxCxHxW shape. Got y_pred: {y_pred.shape} and y: {y.shape}."
+            )
+
+        if self.data_range is None:
```
sdesrozis, line 82: `data_range` value is computed only during the first call, therefore it is frozen, right ? So it is strongly related to the first batch. It seems a bit weird to me.

sdesrozis (commented):
@ydcjeff some comments 😉

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,106 @@
...
+        self.data_range = data_range
+        self.dtype_range = {
+            torch.float16: (-1, 1),
+            torch.float32: (-1, 1),
+            torch.float64: (-1, 1),
+        }
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._sum_of_batchwise_psnr = 0  # type: Union[float, torch.Tensor]
```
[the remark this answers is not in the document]
ydcjeff: I think it doesn't need to create tensor for `_sum_of_batchwise_psnr`

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,114 @@
...
+    .. code-block:: python
+
+        def process_function(engine, batch):
+            # ...
+            return y_pred, y
+        engine = Engine(process_function)
+        psnr = PSNR()
+        psnr.attach(engine, "psnr")
+
+    .. versionadded:: 0.5.0
```
ydcjeff, line 44: should we use this kind of directive? [[link](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded)]

to clearly indicate which features is added at which version, so that we won't need to change the docs from version to version

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __init__(
+        self,
+        data_range: Optional[Union[int, float]] = None,
+        output_transform: Callable = lambda x: x,
+        device: Union[str, torch.device] = torch.device("cpu"),
+    ):
+        super().__init__(output_transform=output_transform, device=device)
+        self.data_range = data_range
+        self.__data_range = data_range
```
ydcjeff, line 55: @vfdev-5 @sdesrozis Is there any other way to avoid this? What we need to check is
- if `data_range` is given, then we use given `data_range`
- if not, we need to determine `data_range` every batch

ignite/metrics/psnr.py
vfdev-5: Sounds good

ignite/metrics/psnr.py
vfdev-5: you can use `self.data_range` as None if not defined by the user and in the update you can use a local var instead of `self.data_range`:
```python
data_range = self.data_range
if data_range is None:
     # compute it for the batch

data_range_sqrt = data_range ** 2 
```

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,114 @@
...
+_int_types = (torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64)
+
+_int_ranges = {k: (torch.iinfo(k).min, torch.iinfo(k).max) for k in _int_types}
+
+dtype_range = {
+    torch.float16: (-1, 1),
+    torch.float32: (-1, 1),
+    torch.float64: (-1, 1),
+}
+dtype_range.update(_int_ranges)
```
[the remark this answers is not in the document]
vfdev-5: I have no special opinion about that. Maybe, we can set `dtype_range` as `_dtype_range` (private) and keep it here. When needed, we can move it later...

vfdev-5 (commented):
Thanks for the update @ydcjeff !
I'm a bit short of time this week, but I'll try to check the code and tests in 
details a bit later (sorry for that)

ignite/metrics/psnr.py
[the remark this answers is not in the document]
ydcjeff: I found out that that is currently used by PSNR and SSIM (in the future, will need to update this)
that's why shall we put to `ignite.metrics.__init__.py`?
Yes, you can take time to review.

ignite/metrics/psnr.py
ydcjeff: Fixed!

sdesrozis:
@ydcjeff I review as soon as possible

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,115 @@
...
+    ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y, ...}``.
+
+    .. code-block:: python
+
+        def process_function(engine, batch):
+            # ...
+            return y_pred, y
+        engine = Engine(process_function)
+        psnr = PSNR()
+        psnr.attach(engine, "psnr")
```
vfdev-5, line 44:
```suggestion
        psnr.attach(engine, "psnr")
    # ...
    state = engine.run(data)
    print(f"PSNR: {state.metrics['psnr']}")
```
I was thinking that providing a complete example could be more helpful. Same in this PR : https://github.com/pytorch/ignite/pull/1583/files

ignite/metrics/psnr.py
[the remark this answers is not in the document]
vfdev-5: > I found out that that is currently used by PSNR and SSIM (in the future, will need to update this)
that's why shall we put to ignite.metrics.__init__.py?

Sounds good. Let's do that in a follow-up PR with SSIM update.

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,118 @@
...
+        self._num_examples += y.shape[0]
+
+    @sync_all_reduce("_sum_of_batchwise_psnr", "_num_examples")
+    def compute(self) -> Union[float, torch.Tensor]:
+        if self._num_examples == 0:
+            raise NotComputableError("PSNR must have at least one example before it can be computed.")
+        return self._sum_of_batchwise_psnr / self._num_examples
+
+
+_int_types = (torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64)
```
vfdev-5, line 109: I think we should either support only uint8 and floats as dtype or do as for SSIM where data_range is required parameter.

For example, I'm trying that:
```python
torch.manual_seed(1)
y = torch.randint(0, 256, size=(32, 3, 32, 32))
y_pred = torch.randint(0, 256, size=(32, 3, 32, 32))

p.reset()
p.update((y_pred, y))
p.compute()
```
and see the error: `RuntimeError: Overflow when unpacking long`. As error message is not indicating the source of the error, it is almost impossible to understand what happens.

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,118 @@
...
+                    "for its data type. Please manually specify the `data_range`."
+                )
+            if true_min >= 0:
+                # most common case (255 for uint8, 1 for float)
+                data_range = dmax
+            else:
+                data_range = dmax - dmin
+
+        mse_error = torch.pow(y_pred.double() - y.view_as(y_pred).double(), 2).mean()
+        self._sum_of_batchwise_psnr += 10.0 * torch.log10(data_range ** 2 / mse_error).to(device=self._device)
```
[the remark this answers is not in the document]
vfdev-5: Another thing about computing PSNR this way, actually computes PSNR per fixed batch and not average of image PSNRs. It means that the metric wont be the same for different batch_size which is probably wrong way to do that.
How about SSIM for this case, does it also depend on batch_size ?

vfdev-5 (commented):
@ydcjeff I have few comments about the implementation. Let me know what do you think. Thanks

ignite/metrics/psnr.py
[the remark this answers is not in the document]
ydcjeff: > We have to be careful to properly say "inf" if y_pred == y. => seems like it is OK

Should we mention in the docs with Note section?

> Another thing about computing PSNR this way, actually computes PSNR per fixed batch and not average of image PSNRs. It means that the metric wont be the same for different batch_size which is probably wrong way to do that.

You are absolutely right.

>  How about SSIM for this case, does it also depend on batch_size ?

not sure about that, results are slightly different when varying batch_size, will catch up on that.

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,118 @@
...
+    .. math::
+        \text{PSNR}(i, j) = 10 * \log_{10}\left(\frac{ MAX_{i}^2 }{ \text{ MSE } }\right)
+
+    where :math:`\text{MSE}` is `mean squared error <https://en.wikipedia.org/wiki/Mean_squared_error>`_.
+
+    - `y_pred` and `y` must be in the following shape (batch_size, ...).
+    - `y_pred` and `y` must have same dtype and same shape.
+
+    Args:
+        data_range (int or float): Range of the image. If not provided, it will be determined
```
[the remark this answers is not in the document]
ydcjeff: Added locally.

ignite/metrics/psnr.py
ydcjeff: So, shall we hardcode with `(-1, 1)` and `(0, 255)`?

ignite/metrics/psnr.py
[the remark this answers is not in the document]
ydcjeff: @vfdev-5 The main issue is we accumulate the mean squared error and divide with batch_size when call `compute()`.
That's why results vary with batch_size.
There are 2 ways to do it as far as I find out.

```py
mse_error = torch.pow(y_pred.double() - y.view_as(y_pred).double(), 2).mean()
self._sum_of_batchwise_psnr += 10.0 * torch.log10(data_range ** 2 / mse_error).to(device=self._device)

def compute():
    return self._sum_of_batchwise_psnr
```
```py
sum_se_error = torch.pow(y_pred.double() - y.view_as(y_pred).double(), 2).sum()
self._sum_of_batchwise_se += mse_error
self._num_examples += y.numel()

def compute():
    mse_error = self._sum_of_batchwise_psnr / self._num_examples
    return 10.0 * torch.log10(self.data_range ** 2 / mse_error)
```
We need `data_range` to have values to compute in 2nd way, but no need in 1st way.
How shall we proceed?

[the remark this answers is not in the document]
vfdev-5: I think the 1st approach still suffers the same issue, right? 
IMO, PSNR metric should be an average of image-wise psnr's. Maybe, 2nd approach is the only correct one.
To check that we have to ensure that : 
```python
p.reset()
p.update((y_preds_batch, y_batch))
res1 = p.compute()

p.reset()
for y_pred, y in zip(y_preds_batch, y_batch):
    p.update((y_preds, y))
res2 = p.compute()

assert res1 == res2
```


> Should we mention in the docs with Note section?

Yes, please.

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,116 @@
...
+    def reset(self) -> None:
+        self._sum_of_squared_error = 0  # type: Union[int, torch.Tensor]
+        self._num_elements = 0
+
+    @reinit__is_reduced
+    def update(self, output: Sequence[torch.Tensor]) -> None:
+        self._check_shape_dtype(output)
+        y_pred, y = output[0].detach(), output[1].detach()
+
+        if self.__data_range is None:
```
vfdev-5, line 85: Should we obligatory reintroduce `__data_range` , previous trick wont work anymore ?

tests/ignite/metrics/test_psnr.py
```diff
@@ -0,0 +1,236 @@
...
+        psnr.update((y_pred, y.squeeze(dim=0)))
+
+    with pytest.raises(ValueError, match=r"y has intensity values outside the range expected"):
+        psnr = PSNR()
+        psnr.update((y_pred, y))
+        # to catch ValueError for this batch
+        psnr.update((y_pred, y + 1.0))
+
+
+def _test_psnr(y_pred, y, data_range, device):
```
vfdev-5, line 39: Can we add a consistency test batch result vs image-wise computation ?

tests/ignite/metrics/test_psnr.py
```diff
@@ -0,0 +1,236 @@
...
+    psnr.update((y_pred, y))
+    psnr_compute = psnr.compute()
+
+    np_y_pred = y_pred.cpu().numpy()
+    np_y = y.cpu().numpy()
+
+    assert isinstance(psnr_compute, torch.Tensor)
+    assert psnr_compute.dtype == torch.float64
+    assert psnr_compute.device == torch.device(device)
+    assert np.allclose(psnr_compute.numpy(), ski_psnr(np_y, np_y_pred, data_range=data_range))
```
vfdev-5, line 50: Skimage can also compute on PSNR on batch of images ? (Haven't checked their doc yet)

vfdev-5 (commented):
Thanks for the update @ydcjeff !

ignite/metrics/psnr.py
ydcjeff: Since `self.data_range` is needed in `compute()`, I can't think of no other way
Let me know if there is.

tests/ignite/metrics/test_psnr.py
ydcjeff: On it!

tests/ignite/metrics/test_psnr.py
ydcjeff: Yes, it does

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,118 @@
...
+            torch.pow(y_pred.double() - y.view_as(y_pred).double(), 2).sum().to(device=self._device)
+        )
+        self._num_elements += y.numel()
+
+    @sync_all_reduce("_sum_of_squared_error", "_num_elements")
+    def compute(self) -> Union[float, torch.Tensor]:
+        if self._num_elements == 0:
+            raise NotComputableError("PSNR must have at least one example before it can be computed.")
+        return 10.0 * torch.log10(
+            self.data_range ** 2 / (self._sum_of_squared_error / self._num_elements)  # type: ignore[operator, arg-type]
```
vfdev-5, line 109: I'm still not very certain about this implementation. Ultimately, we would like to compute average PSNR as 
```python
avg_psnr = sum_on_images(psnr[image]) =

= 10 * log10 ( prod_on_images( data_range[image] / mse[image] ) ) = 
= 10 * log10 ( prod_on_batch( prod_images_in_batch( data_range[image] / mse[image] ) ) = 
= 10 * log10 ( prod_on_batch( accumulated_pow10_pnsr[batch] ) ) 
```

What do you think ?

tests/ignite/metrics/test_psnr.py
vfdev-5: No, actually, it considers batch of images as N-dim image and compute psnr on batch images considering it as a single image and we would like to have average of B psnr's.

ignite/metrics/psnr.py
ydcjeff: The thing is we allow any shape for the metrics, so we can't assume the first dim is batch_size and cannot compute mean for the rest of dims.

vfdev-5: Well, for the images I've never seen data with shapes like `(..., B, ...)`. Otherwise, we can put this as a requirement which is rather natural.

ydcjeff: I don't have an example, I usually have NCHW images. Since skimage allows any shape, I try to adopt that.
[TensorFlow](https://www.tensorflow.org/api_docs/python/tf/image/psnr) has some restrictions tho.
Should we allow only 4 dim use case?
Is there 2/3 dim use case I am not aware of?
Also, this metrics can apply to audio domain as well (I am completely new to that domain).

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,118 @@
...
+                    "for its data type. Please manually specify the data_range."
+                )
+            if true_min >= 0:
+                # most common case (255 for uint8, 1 for float)
+                data_range = dmax
+            else:
+                data_range = dmax - dmin
+
+        rmse_error = torch.pow(y_pred.double() - y.view_as(y_pred).double(), 2).mean(dim=list(range(1, y.ndim))).sqrt()
+        self._sum_of_batchwise_psnr += 20.0 * torch.log10(data_range / rmse_error).to(self._device)
```
ydcjeff, line 103: Should we call `torch.sum()` here or in the `compute()`?
I saw the other metrics calling `torch.sum()` here. How about the memory consumption?

vfdev-5: Sorry, i'll bother you again with if we are certain now that this implementation is correct one now :) ?

ydcjeff: I think I am certain about it.
Right now I am computing the MSE (or RMSE) per image of a batch and psnr from that, resulting in psnr shape of [batch_size] with a requirement of dim 1 being batch. The main issue is I am computing MSE per batch, not per image at the start.

Do you have any other way in mind?

vfdev-5: > The main issue is I am computing MSE per batch, not per image at the start.

which part of the code correspond to that (at start) ?
Anyway, let me take sometime to check it from my side and we'll see.

ignite/metrics/psnr.py
vfdev-5: Well, I think a support for case starting from 2D would be great ! Let's see what is possible to do

ignite/metrics/psnr.py
ydcjeff: This one
```py
mse_error = torch.pow(y_pred.double() - y.view_as(y_pred).double(), 2).mean()
```
That was around the start commit of the PR

vfdev-5: ah, OK I see what you wanted to say.

ignite/metrics/psnr.py
ydcjeff: Supported now

ignite/metrics/psnr.py
vfdev-5: About your question on where to call the sum. Let's call it in update as in other metrics. Such that we could also remove `# type: ignore[arg-type]` from compute.


Another point, IMO, it is faster to compute square of data_range than rmse. Let's do 10 * log10(max / mse).

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,120 @@
...
+            )
+
+        if y_pred.shape != y.shape:
+            raise ValueError(
+                f"Expected y_pred and y to have the same shape. Got y_pred: {y_pred.shape} and y: {y.shape}."
+            )
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._sum_of_batchwise_psnr = 0  # type: Union[int, torch.Tensor]
```
vfdev-5, line 80:
```suggestion
        self._sum_of_batchwise_psnr = torch.tensor(0, device=self._device)
```

vfdev-5 (commented):
Thanks for the update @ydcjeff ! Looks almost good now.
I have few other comments on docs and the code.

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,120 @@
...
+        self._num_examples = 0
+
+    @reinit__is_reduced
+    def update(self, output: Sequence[torch.Tensor]) -> None:
+        self._check_shape_dtype(output)
+        y_pred, y = output[0].detach(), output[1].detach()
+        data_range = self.data_range
+
+        if data_range is None:
+            dmin, dmax = _dtype_range[y.dtype]
```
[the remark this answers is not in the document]
ydcjeff: ahhh, yes, I forgot about that.

tests/ignite/metrics/test_psnr.py
```diff
@@ -0,0 +1,288 @@
...
+    psnr = PSNR(data_range=data_range, device=device)
+    psnr.update((y_pred, y))
+
+    np_y_pred = y_pred.cpu().numpy()
+    np_y = y.cpu().numpy()
+    np_psnr = []
+    for np_y_pred_, np_y_ in zip(np_y_pred, np_y):
+        np_psnr.append(ski_psnr(np_y_, np_y_pred_, data_range=data_range))
+
+    assert np.allclose(psnr._sum_of_batchwise_psnr.cpu().numpy(), np_psnr)
```
ydcjeff, line 49: @vfdev-5 do you think this test is still required after we move `torch.sum()` to `update()`?
We are comparing a list of batch results vs a list of batch results.
But after we move `torch.sum()` to `update()`, we are comparing sum values which are being covered by the below tests

ignite/metrics/psnr.py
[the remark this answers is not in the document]
vfdev-5: I think this is still uncovered, right ? Maybe, also it would make sense to put the check and raise into `_check_shape_dtype`...

tests/ignite/metrics/test_psnr.py
vfdev-5: Yes, let's remove it.

ignite/metrics/psnr.py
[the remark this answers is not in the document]
ydcjeff: It's already added locally, just waiting your reply below test `_test_psnr_per_image` to be removed

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,129 @@
...
+    def __init__(
+        self,
+        data_range: Optional[Union[int, float]] = None,
+        output_transform: Callable = lambda x: x,
+        device: Union[str, torch.device] = torch.device("cpu"),
+    ):
+        super().__init__(output_transform=output_transform, device=device)
+        self.data_range = data_range
+
+    def _check_shape_dtype_drange(self, output: Sequence[torch.Tensor], data_range: Union[int, float]) -> None:
```
ydcjeff, line 67:
```suggestion
    def _check_shape_dtype_drange(self, output: Sequence[torch.Tensor], data_range: Union[int, float]) -> Union[int, float]:
```

ignite/metrics/psnr.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Note:
+        A few things to note:
+
+        - `y_pred` and `y` **must** have (batch_size, ...) shape.
+        - `y_pred` and `y` **must** have same dtype and same shape.
+        - If `y_pred` and `y` have almost identical values, the result will be infinity.
+
+    Args:
+        data_range (int or float): The data range of the target image (distance between minimum
```
vfdev-5, line 27:
```suggestion
        data_range (int or float, optional): The data range of the target image (distance between minimum
```

vfdev-5 (approved):
LGTM! Thanks a lot for the PR @ydcjeff !

ydcjeff:
Thanks a lot for reviewing the PR thoroughly. @vfdev-5

lidq92:
> @lidq92 could you please take a look this PR ?

@vfdev-5  Sorry for the late reply.

I have reviewed the PSNR implementation. 
I have also read some relevant references (e.g., Tensorflow, PyTorch examples, Lightening, kornia, etc.),  but at the time I write the following comments, I did not take a look at the PSNR implementation in `skimage`.

For **psnr.py**, 
- Line 15: i->I, j->J (for better notations of images)
- About `_dtype_range`: _Why the default range of float16/float32/float64 is (-1.0, 1.0)?_ And I think we can just set it to (0.0, 1.0).
Then we may make some changes to check data_range (Users should specify the `data_range` if the data range is uncommon or other dtypes are considered) . 
```python
if data_range is None:
    try:
        dmin, dmax = _dtype_range[y.dtype]
    except KeyError:
        raise ValueError(
            "Range for this dtype cannot be automatically estimated. Please manually specify the data_range."
        )
    true_min, true_max = y.min(), y.max()
    if true_max > dmax or true_min < dmin:
        raise ValueError(
            "y has intensity values outside the range expected "
            "for its data type. Please manually specify the data_range."
        )
    data_range = dmax - dmin
return data_range
```
- Situation when `mse_error=0`: Do we need to consider a small positive constant EPS (e.g., 1e-8) to avoid division by zero?
- We may provide an option for users to decide the way to compute PSNR: convert the image to Y channel only before computing PSNR, compute PSNR based on average MSE over all channels, or return a scale value per channel (which will allow different data ranges among the channels), etc.

For **test_psnr.py**, 
- I did not understand why different data ranges are set for the same input data since `data_range` should have its meaning.

===========
===========
I finally realized that current implementation is heavily based on the implementation in `skimage` (I missed the conversations, and sorry for that).

But I keep my comments.
For an image in PyTorch framework, if its dtype is float, (common) default range is (0.0, 1.0).

By the way, for the meaning of `data_range`, besides the maximal data range of a certain `dtype`, there is another understanding, i.e., the actual range of the "**signal**" (`y` in the code).
The alternative understanding may be more elegant, since the range of PSNR is [0, infinity) under this definition.
But the minimum of PSNR is larger than zero if the actual range of the signal is smaller than the maximal data range of a certain `dtype` under the first (and **common**) definition.
This can be happened, for example, considering the U, V channels of the YUV space.
Anyway, we may continue to use the first definition, since it is **commonly** used in the image processing community.

vfdev-5:
Thanks a lot for the review @lidq92 !

> Line 15: i->I, j->J (for better notations of images)

That's true. We have to fix that in a follow-up PR

> For an image in PyTorch framework, if its dtype is float, (common) default range is (0.0, 1.0).

Actually, I'd say in Pytorch, float range is something like (-2.12, 2.64) for ImageNet normalized imagery:
```python
import torch
from torchvision.transforms import ToTensor, ToPILImage, Normalize, Compose, RandomErasing
x = torch.arange(0, 3 * 32 * 32, dtype=torch.uint8).reshape(3, 32, 32)
t = Compose([ToPILImage(), ToTensor(), Normalize((0.485, 0.456, 0.406), (0.229, 0.224, 0.225))])
float_image = t(x)
float_image.min().item(), float_image.max().item()
```
But I agree depending on how it is normalized. In ML i'd say it is sort of recommended (probably before BN) to normalize the data to have zero mean. On the other hand, in OpenCV float image range is [0.0, 1.0] ([link](https://docs.opencv.org/master/d8/d01/group__imgproc__color__conversions.html#ga397ae87e1288a81d2363b61574eb8cab)).
For us, maybe, the main point is to have data_range to be configurable.

> Situation when mse_error=0: Do we need to consider a small positive constant EPS (e.g., 1e-8) to avoid division by zero?

This is a good point and currently if we compute psnr on the same tensor (e.g. `psnr.update((y, y))`), we have `tensor(inf, dtype=torch.float64)`. I think this is acceptable value. What do you think ?

> We may provide an option for users to decide the way to compute PSNR: convert the image to Y channel only before computing PSNR, compute PSNR based on average MSE over all channels, or return a scale value per channel (which will allow different data ranges among the channels), etc.

This is also a good point, I think we can use `output_transform` for that. For example
```python
def ot_get_y_channel(output):
    y_pred, y = output
    # y is (B, 3, H, W) and is YUV
    # let's select Y channel
    return y_pred[:, 0, ...], y[:, 0, ...]

psnr = PSNR(data_range=(0.0, 1.0), output_transform=ot_get_y_channel)
```
Maybe, it worth mention that in the docs.

> I did not understand why different data ranges are set for the same input data since data_range should have its meaning.

Good question. @ydcjeff what do you think ?


> The alternative understanding may be more elegant, since the range of PSNR is [0, infinity) under this definition.
But the minimum of PSNR is larger than zero if the actual range of the signal is smaller than the maximal data range of a certain dtype under the first (and common) definition.
This can be happened, for example, considering the U, V channels of the YUV space.

Yes, PSNR can be 0 in very particular situation, right ? It is 0 if `log10(1.0)` or `data_range**2 == MSE`. Yes, I agree as MSE is bounded by actual signal range and it can be a situation if data_range is badly estimated then PSNR result may be incosistent. It feels like it is very important to have coherent and same data_range...

Actually, what is exactly this alternative definition of PSNR ?

lidq92:
> > For an image in PyTorch framework, if its dtype is float, (common) default range is (0.0, 1.0).
> 
> Actually, I'd say in Pytorch, float range is something like (-2.12, 2.64) for ImageNet normalized imagery
> But I agree depending on how it is normalized. In ML i'd say it is sort of recommended (probably before BN) to normalize the data to have zero mean. On the other hand, in OpenCV float image range is [0.0, 1.0] ([link](https://docs.opencv.org/master/d8/d01/group__imgproc__color__conversions.html#ga397ae87e1288a81d2363b61574eb8cab)).
> For us, maybe, the main point is to have data_range to be configurable.
I see. Yes, considering these transformations, data range can be quite different. And PSNR is invariant to such normalization with global constants since it is a dimensionless metric.
In this situation, setting value of float16/float32/float64 in `_dtype_range` to (-1.0, 1.0) may cause an inexact estimation of `data_range`. For example, `I` is a float image (after some transformations) with an actual range (-0.5, 0.5), and `dmin<=true_min<0` leads to `data_range=2.0`, which overestimates the true PSNR value.
This also involves the calculation/estimation of the actual range of the signal, i.e., MAX_I. I would leave additional comments on it at the end.

> > Situation when mse_error=0: Do we need to consider a small positive constant EPS (e.g., 1e-8) to avoid division by zero?
> 
> This is a good point and currently if we compute psnr on the same tensor (e.g. `psnr.update((y, y))`), we have `tensor(inf, dtype=torch.float64)`. I think this is acceptable value. What do you think ?
Although inf is computable in PyTorch, it may lead to some unexpected results. 
For example, if this metric is used to compare different Image Processing Algorithms (debarring/denoising/super-resolution, etc.), the average PSNR (in a batch) would be inf given that one of the PSNR values is inf.
That is, if one algorithm does perfectly in processing one image, but it performs terribly on other images in the batch, it would still get the best PSNR (i.e., inf).
This would leads to an inappropriate comparison (And this can also be true when one algorithm does very well on a specific image but performance poorly on other images). 
It is not a problem of division by zero, and I would say it is the issue of using average PSNR in a batch for overall performance comparison, which may be out of scope in this PSNR implementation.

> > We may provide an option for users to decide the way to compute PSNR: convert the image to Y channel only before computing PSNR, compute PSNR based on average MSE over all channels, or return a scale value per channel (which will allow different data ranges among the channels), etc.
> 
> This is also a good point, I think we can use `output_transform` for that. For example
> 
> ```python
> def ot_get_y_channel(output):
>     y_pred, y = output
>     # y is (B, 3, H, W) and is YUV
>     # let's select Y channel
>     return y_pred[:, 0, ...], y[:, 0, ...]
> 
> psnr = PSNR(data_range=(0.0, 1.0), output_transform=ot_get_y_channel)
> ```
> 
> Maybe, it worth mention that in the docs.
> 
I agree to do so. 
> > I did not understand why different data ranges are set for the same input data since data_range should have its meaning.
> 
> Good question. @ydcjeff what do you think ?
> 

> > The alternative understanding may be more elegant, since the range of PSNR is [0, infinity) under this definition.
> > But the minimum of PSNR is larger than zero if the actual range of the signal is smaller than the maximal data range of a certain dtype under the first (and common) definition.
> > This can be happened, for example, considering the U, V channels of the YUV space.
> 
> Yes, PSNR can be 0 in very particular situation, right ? It is 0 if `log10(1.0)` or `data_range**2 == MSE`. Yes, I agree as MSE is bounded by actual signal range and it can be a situation if data_range is badly estimated then PSNR result may be incosistent. It feels like it is very important to have coherent and same data_range...
> 
> Actually, what is exactly this alternative definition of PSNR ?
The alternative definition of PSNR is still the same as illustrated in the comments of the code, but with an **alternative** calculation of MAX_I: not the maximal difference of a `dtype`, e.g., 255-0 for any uint8 image channel, but the true range on an image channel, e.g., given a uint8 Y'CbCr image, 235-16 for Y', 240-16 for Cb and Cr.
In current implementation, if `data_range=None`, then the `data_range` will be set to 255 for a uint8 Y'CbCr image, which will overestimate PSNR.
To ensure users be aware of this actual range, I would recommend the `data_range` be a required parameter without a default value (though it may be less convenient).

vfdev-5:
Thanks for the details, @lidq92 ! Yes, I understand your point about `data_range` and agree that we should put it as required arg instead auto-estimation which may lead to under/over-estimated PSNR.

And yes, I see your point about inf handling. Let's introduce `eps=1e-10` to avoid "inf".

ydcjeff:
Thank you very much for reviewing @lidq92 

> Situation when mse_error=0: Do we need to consider a small positive constant EPS (e.g., 1e-8) to avoid division by zero?

> Although inf is computable in PyTorch, it may lead to some unexpected results.
For example, if this metric is used to compare different Image Processing Algorithms (debarring/denoising/super-resolution, etc.), the average PSNR (in a batch) would be inf given that one of the PSNR values is inf.
That is, if one algorithm does perfectly in processing one image, but it performs terribly on other images in the batch, it would still get the best PSNR (i.e., inf).
This would leads to an inappropriate comparison (And this can also be true when one algorithm does very well on a specific image but performance poorly on other images).
It is not a problem of division by zero, and I would say it is the issue of using average PSNR in a batch for overall performance comparison, which may be out of scope in this PSNR implementation.

Yes, I miss the point that PSNR of one image can be inf. In `skimage`, the intended way is to pass per image which can result in `inf` which is ok for per image calculation.

> We may provide an option for users to decide the way to compute PSNR: convert the image to Y channel only before computing PSNR, compute PSNR based on average MSE over all channels, or return a scale value per channel (which will allow different data ranges among the channels), etc.

Currently I assume the input image is 3 dimensional RGB image, I found out Y channel is only used to compute PSNR for YCbCr and YUV images, but I am not sure the sources I found are reliable. That's why I didn't add them.

> compute PSNR based on average MSE over all channels

Currently we are computing with this option, but this is only true for RGB images, right? @lidq92 

> return a scale value per channel (which will allow different data ranges among the channels)

We currently didn't support this option. @vfdev-5 thoughts?

How to compute PSNR for HSV images? @lidq92 

> I did not understand why different data ranges are set for the same input data since data_range should have its meaning.

I passed different data_range values to check if the output is still the same regardless of data_range meaning.
But, I think now I am computing wrongly in the tests with over/under given data_range based on the review.

>  To ensure users be aware of this actual range, I would recommend the data_range be a required parameter without a default value (though it may be less convenient).

TIL that `data_range` has an another meaning.

So to recap,
I will come up with a follow up PR making
- making `data_range` a required argument
- mentioning in the docs that Y channel is only need for YCbCr and YUV images
- add small eps to avoid inf in a batch
- more option for computing per channel or average of all channels?

lidq92:
@ydcjeff Thanks for your efforts to improve the ignite.
> > We may provide an option for users to decide the way to compute PSNR: convert the image to Y channel only before computing PSNR, compute PSNR based on average MSE over all channels, or return a scale value per channel (which will allow different data ranges among the channels), etc.
> 
> Currently I assume the input image is 3 dimensional RGB image, I found out Y channel is only used to compute PSNR for YCbCr and YUV images, but I am not sure the sources I found are reliable. That's why I didn't add them.
> 
> > compute PSNR based on average MSE over all channels
> 
> Currently we are computing with this option, but this is only true for RGB images, right? @lidq92
> 
For multi-channel images, the way to calculate PSNR is not officially defined.
There are some literatures combining PSNR values across all channels to obtain an overall measure of image quality.
The R, G, B channels are in the equal position, so computing PSNR based on average MSE over all channels is acceptable.
For images in other color spaces, the same calculation procedure can be performed, but averaging MSE over all channels may be unreasonable.
A more complicated and perceptually-relevant combination of PSNR or MSE values over all channels can be studied, which is out of scope. 
> > return a scale value per channel (which will allow different data ranges among the channels)
> 
> We currently didn't support this option. @vfdev-5 thoughts?
> 
> How to compute PSNR for HSV images? @lidq92
> 
See above. 
One can compute PSNR/MSE for each channel of the HSV image.
But the combination of PSNR_H/MSE_H, PSNR_S/MSE_S, and PSNR_V/MSE_V can be different. 
One unified way is that transform the colored images (no matter which color space is used) to gray-scale images, and calculate PSNR of one-channel image.

> > I did not understand why different data ranges are set for the same input data since data_range should have its meaning.
> 
> I passed different data_range values to check if the output is still the same regardless of data_range meaning.
> But, I think now I am computing wrongly in the tests with over/under given data_range based on the review.
> 
> > To ensure users be aware of this actual range, I would recommend the data_range be a required parameter without a default value (though it may be less convenient).
> 
> TIL that `data_range` has an another meaning.
> 
Today I learned the meaning of TIL😸
> So to recap,
> I will come up with a follow up PR making
> 
> * making `data_range` a required argument
> * mentioning in the docs that Y channel is only need for YCbCr and YUV images
> * add small eps to avoid inf in a batch
> * more option for computing per channel or average of all channels?
>
Sounds good~

vfdev-5:
> So to recap,
I will come up with a follow up PR making
> -    making data_range a required argument
> -    mentioning in the docs that Y channel is only need for YCbCr and YUV images
> -    add small eps to avoid inf in a batch
> -    more option for computing per channel or average of all channels?

Yes, sounds good for : 
- making data_range a required argument
- add small eps to avoid inf in a batch

> mentioning in the docs that Y channel is only need for YCbCr and YUV images

Yes and adding an example code
```python
def ot_get_y_channel(output):
    y_pred, y = output
    # y is (B, 3, H, W) and is YUV
    # let's select Y channel
    return y_pred[:, 0, ...], y[:, 0, ...]

psnr = PSNR(data_range=(0.0, 1.0), output_transform=ot_get_y_channel)
``` 

> more option for computing per channel or average of all channels?

I think we can keep for now the behaviour where we compute the average. If user would like compute for each channel then we can advise to use specific output_transform to compute 3 metrics for each channel (HSV or YUV etc)...

sdesrozis:
> To ensure users be aware of this actual range, I would recommend the data_range be a required parameter without a default value (though it may be less convenient).

I agree !

Issue #1575, opened 2021-01-26
[docs] use `versionadded`, `versionchanged` and `deprecated` directive


ydcjeff:
## 🚀 Feature
Use `.. versionadded::`, `.. versionchanged::` and `.. deprecated::` directive, so that user knows which features are added / changed / deprecated in which version and they can navigate the docs easily without changing the docs from version to version.

Ref: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded

vfdev-5:
Thanks for the idea @ydcjeff ! It would be very nice to have that.

ydcjeff:
@vfdev-5 we can start adding for features to be added in 0.5.0.

Where should we begin?

vfdev-5:
Yes, starting from v0.5.0

vfdev-5:
@ydcjeff could you please mention those `versionadded`, `versionchanged` in the CONTRIBUTING in the docs section. Thanks !

ydcjeff:
This is already in use, so closing

PR #1578, opened 2021-01-26
[docs] added `versionadded` and `versionchanged` for v0.4.2

ydcjeff:
Part of #1575 

Description: added `versionadded` and `versionchanged` for [v0.4.2](https://github.com/pytorch/ignite/releases/tag/v0.4.2)

[1](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/distributed.html#ignite.distributed.auto.auto_model), [2](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/distributed.html#ignite.distributed.auto.auto_optim), [3](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/distributed.html#ignite.distributed.launcher.Parallel.run), [4](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/distributed.html#ignite.distributed.utils.device), [5](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/distributed.html#ignite.distributed.utils.backend), [6](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/distributed.html#ignite.distributed.utils.model_name), [7](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/distributed.html#ignite.distributed.utils.broadcast),  [8](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/distributed.html#ignite.distributed.utils.sync),  [9](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/distributed.html#ignite.distributed.utils.initialize),  [10](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/handlers.html#ignite.handlers.ModelCheckpoint),  [11](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/handlers.html#ignite.handlers.DiskSaver),  [12](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/metrics.html#ignite.metrics.Metric.attach),  [13](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/metrics.html#ignite.metrics.SSIM), [14](https://deploy-preview-1578--pytorch-ignite-preview.netlify.app/contrib/handlers.html#stores)

Check list:

### DIFF
--- a/docs/source/contrib/handlers.rst
+++ b/docs/source/contrib/handlers.rst
@@ -43,6 +43,17 @@ time_profilers
 .. automodule:: ignite.contrib.handlers.time_profilers
    :members:
 
+stores
+-------
+
+.. currentmodule:: ignite.contrib.handlers.stores
+
+.. autosummary:
+    :nosignatures:
+    :autolist:
+
+.. automodule: ignite.contrib.handlers.stores
+    :members:
 
 tensorboard_logger
 ------------------
--- a/docs/source/contrib/handlers.rst
+++ b/docs/source/contrib/handlers.rst
@@ -48,11 +48,11 @@ stores
 
 .. currentmodule:: ignite.contrib.handlers.stores
 
-.. autosummary:
+.. autosummary::
     :nosignatures:
     :autolist:
 
-.. automodule: ignite.contrib.handlers.stores
+.. automodule:: ignite.contrib.handlers.stores
     :members:
 
 tensorboard_logger
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -570,6 +570,9 @@ class DiskSaver(BaseSaveHandler):
         require_empty (bool, optional): If True, will raise exception if there are any files in the
             directory ``dirname``.
         **kwargs: Accepted keyword arguments for `torch.save` or `xm.save`.
+
+    .. versionchanged:: 0.4.2
+        Accept ``kwargs`` for `torch.save` or `xm.save`.
     """
 
     def __init__(
@@ -693,6 +696,9 @@ class ModelCheckpoint(Checkpoint):
         include_self (bool): Whether to include the `state_dict` of this object in the checkpoint. If `True`, then
             there must not be another object in ``to_save`` with key ``checkpointer``.
         **kwargs: Accepted keyword arguments for `torch.save` or `xm.save` in `DiskSaver`.
+
+    .. versionchanged:: 0.4.2
+        Accept ``kwargs`` for torch.save or xm.save
 
     Examples:
         >>> import os
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -698,7 +698,7 @@ class ModelCheckpoint(Checkpoint):
         **kwargs: Accepted keyword arguments for `torch.save` or `xm.save` in `DiskSaver`.
 
     .. versionchanged:: 0.4.2
-        Accept ``kwargs`` for torch.save or xm.save
+        Accept ``kwargs`` for `torch.save` or `xm.save`
 
     Examples:
         >>> import os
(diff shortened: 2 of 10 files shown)


ydcjeff:
We need to have some conventions about that. For now, I added every change at the end of the docstring as I find it easier to read and find changelog. But, let me know what you think about that.

Make some css changes – bold and italic to be able to find quickly.

ignite/handlers/checkpoint.py
```diff
@@ -694,6 +697,9 @@ class ModelCheckpoint(Checkpoint):
             there must not be another object in ``to_save`` with key ``checkpointer``.
         **kwargs: Accepted keyword arguments for `torch.save` or `xm.save` in `DiskSaver`.
 
+    .. versionchanged:: 0.4.2
+        Accept ``kwargs`` for torch.save or xm.save
```
ydcjeff, line 701:
```suggestion
        Accept ``kwargs`` for `torch.save` or `xm.save`
```

vfdev-5:
Thanks a lot for the PR @ydcjeff ! 
Looks very pretty now. I'm wondering about two things however:
- what's the best practice of doing things with such versionadded things
- should we keep multiple notes like that:
```
Changed in version 0.4.2: Added another thing

Changed in version 0.4.1: Added something
```
or keep the latest one ?
- Can we aggregate message into a single one for the same version:
```
Changed in version 0.4.2: 
  - Added Horovod distributed framework. 
  - Added sync_bn argument.
```

What do you think ?

Devanshu24:
Sorry to jump in, I feel this could help the discussion
While researching a bit on how to implement deprecation messages in libraries I found [this](https://github.com/tantale/deprecated/blob/master/deprecated/sphinx.py). It seems relevant to what you are discussing here
Maybe I could add something similar when adding the `@deprecated` decorator

Please ignore if it is not helpful :)

vfdev-5:
@Devanshu24 thanks for joining the discussion and there is no restrictions about that, so no worries. Thanks for the link, let me check that too. The only thing I'm a bit worried is that it is a new dependency that we usually tend to avoid if really necessary...

Devanshu24:
> @Devanshu24 thanks for joining the discussion and there is no restrictions about that, so no worries. Thanks for the link, let me check that too. The only thing I'm a bit worried is that it is a new dependency that we usually tend to avoid if really necessary...

No, I am not suggesting to include the full dependency. It does a lot of error checking and that increases it's size, I could try to write a stripped-down version of it, basically, just something that can do the version management using decorators and also be compatible with sphinx

ydcjeff:
@vfdev-5 I noticed this a few days ago while reading python docs. So, I think we should follow what python ecosystem used to.

> - what's the best practice of doing things with such versionadded things

I noticed python docs adds those at the end of the docstring for some and in the middle (before example) for some. But, I find that is harder to read in ignite docs (like `Checkpoint`) where there are many notes in between. So, I would like to add at the end of docstring as a convention as I find it easier to read and find both on the docs (we describe this convention somewhere in the docs, users will become used to it later) and in the IDE. WDYT?

> - should we keep multiple notes like that:

I think we don't need all of them, may be start from 0.4.0? as the changes will not be reflected to previous version docs.

> ```
> Changed in version 0.4.2: 
>  - Added Horovod distributed framework. 
>  - Added sync_bn argument.
> ```
I haven't tried this one tho, will try.

vfdev-5:
@Devanshu24 sounds good !

> So, I would like to add at the end of docstring as a convention as I find it easier to read and find both on the docs (we describe this convention somewhere in the docs, users will become used to it later) and in the IDE. WDYT?

@ydcjeff I agree !

docs/source/contrib/handlers.rst
```diff
@@ -43,6 +43,17 @@ time_profilers
 .. automodule:: ignite.contrib.handlers.time_profilers
    :members:
 
+stores
```
vfdev-5, line 46: Thanks for catching that !

vfdev-5 (approved):
Thanks a lot for the PR @ydcjeff ! 
Looks good to me. Something else do you think we need to add here ?

ydcjeff:
I think there's any left to add.

PR #1585, opened 2021-01-27
Managing Deprecation using decorators

### LINKED ISSUE #1479: Improve deprecation messages
## 🚀 Feature

House keeping request to improve the codebase. 
The idea is to 1) improve deprecation messages inspired from scikit-image: https://github.com/scikit-image/scikit-image/blob/87a8806cca7fb5366b6e5ddbe5e46364b44f90fe/skimage/_shared/utils.py#L15-L214
```
            f"{self.arg_name} argument is deprecated and will be removed "
            f"in version {self.changed_version}. To avoid this warning, "
            f"please do not use the {self.arg_name} argument. Please "
            f"see {func.__name__} documentation for more details.")
```
and 2) introduce similar decorators in our codebase in a minimalistic way.

Devanshu24:
This is a *very* stripped down version of the code, I have not written any tests yet. This is primarily for me to check if I am going in the correct direction. So please do tell everything that needs improvement/changing.

Fixes #1479 

Description:
Implemented till now
- Make functions deprecated using the `@deprecated` decorator
- Add arguments to the @deprecated decorator to customize it for each function
- The deprecation messages also reflect in the documentation (written in Sphinx compatible format)

Check list:

Thank you to @ydcjeff for giving the idea to add the version update information :)

### DIFF
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -276,12 +276,12 @@ def empty_cuda_cache(_: Engine) -> None:
 
 
 @deprecated(
-    "0.4.0",  # type: ignore
+    "0.4.0",
     "0.6.0",
     ["Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc."],
-    raiseException=True,
+    raise_exception=True,
 )
-def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, log_every_iters) -> None:
+def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, log_every_iters) -> None:  # type: ignore
     pass
 
 
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -190,7 +190,7 @@ def _setup_common_training_handlers(
 
     if with_gpu_stats:
         GpuInfo().attach(
-            trainer, name="gpu", event_name=Events.ITERATION_COMPLETED(every=log_every_iters)  # type: ignore
+            trainer, name="gpu", event_name=Events.ITERATION_COMPLETED(every=log_every_iters)  # type: ignore[arg-type]
         )
 
     if output_names is not None:
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -30,6 +30,7 @@ from ignite.engine import Engine, Events
 from ignite.handlers import Checkpoint, DiskSaver, EarlyStopping, TerminateOnNan
 from ignite.handlers.checkpoint import BaseSaveHandler
 from ignite.metrics import RunningAverage
+from ignite.utils import deprecated
 
 
 def setup_common_training_handlers(
@@ -389,7 +390,7 @@ def setup_visdom_logging(
     _setup_logging(logger, trainer, optimizers, evaluators, log_every_iters)
     return logger
 
-
+@deprecated("0.4.2", "0.6.0", ["This is reason 1", "This is reason 2"])
 def setup_mlflow_logging(
     trainer: Engine,
     optimizers: Optional[Union[Optimizer, Dict[str, Optimizer]]] = None,
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -190,7 +190,7 @@ def _setup_common_training_handlers(
 
     if with_gpu_stats:
         GpuInfo().attach(
-            trainer, name="gpu", event_name=Events.ITERATION_COMPLETED(every=log_every_iters)  # type: ignore[arg-type]
+            trainer, name="gpu", event_name=Events.ITERATION_COMPLETED(every=log_every_iters)  # type: ignore
         )
 
     if output_names is not None:
@@ -280,8 +280,8 @@ def empty_cuda_cache(_: Engine) -> None:
     "0.6.0",
     ["Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc."],
     raiseException=True,
-)
-def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, log_every_iters) -> None:  # type: ignore
+)  # type: ignore
+def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, log_every_iters) -> None:
     pass
 
 
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -276,11 +276,11 @@ def empty_cuda_cache(_: Engine) -> None:
 
 
 @deprecated(
-    "0.4.0",
+    "0.4.0",  # type: ignore
     "0.6.0",
     ["Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc."],
     raiseException=True,
-)  # type: ignore
+)
 def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, log_every_iters) -> None:
     pass
 
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -275,11 +275,14 @@ def empty_cuda_cache(_: Engine) -> None:
     gc.collect()
 
 
+@deprecated(
+    "0.4.0",
+    "0.6.0",
+    ["Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc."],
+    raiseException=True,
+)
 def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, log_every_iters) -> None:  # type: ignore
-    raise DeprecationWarning(
-        "ignite.contrib.engines.common.setup_any_logging is deprecated since 0.4.0. and will be remove in 0.6.0. "
-        "Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc."
-    )
+    pass
 
 
 def _setup_logging(
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -278,10 +278,17 @@ def empty_cuda_cache(_: Engine) -> None:
 @deprecated(
     "0.4.0",
     "0.6.0",
-    ["Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc."],
+    ("Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc.",),
     raise_exception=True,
 )
-def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, log_every_iters) -> None:  # type: ignore
+def setup_any_logging(
+    logger: BaseLogger,
+    logger_module: Any,
+    trainer: Engine,
+    optimizers: Optional[Union[Optimizer, Dict[str, Optimizer], Dict[None, Optimizer]]],
+    evaluators: Optional[Union[Engine, Dict[str, Engine]]],
+    log_every_iters: int,
+) -> None:
     pass
 
 
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -390,7 +390,7 @@ def setup_visdom_logging(
     _setup_logging(logger, trainer, optimizers, evaluators, log_every_iters)
     return logger
 
-@deprecated("0.4.2", "0.6.0", ["This is reason 1", "This is reason 2"])
+
 def setup_mlflow_logging(
     trainer: Engine,
     optimizers: Optional[Union[Optimizer, Dict[str, Optimizer]]] = None,
--- a/ignite/utils.py
+++ b/ignite/utils.py
@@ -2,6 +2,7 @@ import collections.abc as collections
 import logging
 import random
 from typing import Any, Callable, Optional, Tuple, Type, Union, cast
+import functools
 
 import torch
 
@@ -160,3 +161,24 @@ def manual_seed(seed: int) -> None:
         np.random.seed(seed)
     except ImportError:
         pass
+
+
+def deprecated(deprecated_in, removed_in=None, reasons=[], raiseWarning=False):
+
+    def decorator(func):
+        func_doc = func.__doc__
+
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            if raiseWarning:
+                raise DeprecationWarning()
+            return func(*args, **kwargs)
+
+        appended_doc = f'.. deprecated:: {deprecated_in}' + ('\n\n\t' if len(reasons) else '')
+
+        for reason in reasons:
+            appended_doc += '\n\t- ' + reason
+        wrapper.__doc__ = '**Deprecated function**.' + '\n\n    ' + func_doc + appended_doc
+        return wrapper
+
+    return decorator
--- a/ignite/utils.py
+++ b/ignite/utils.py
@@ -2,8 +2,9 @@ import collections.abc as collections
 import functools
 import logging
 import random
+import sys
 import warnings
-from typing import Any, Callable, Optional, Tuple, Type, TypeVar, Union, cast
+from typing import Any, Callable, Optional, TextIO, Tuple, Type, TypeVar, Union, cast
 
 import torch
 
@@ -57,14 +58,19 @@ def to_onehot(indices: torch.Tensor, num_classes: int) -> torch.Tensor:
     """Convert a tensor of indices of any shape `(N, ...)` to a
     tensor of one-hot indicators of shape `(N, num_classes, ...) and of type uint8. Output's device is equal to the
     input's device`.
+
+    .. versionchanged:: 0.4.3
+        This functions is now torchscriptable.
     """
-    onehot = torch.zeros(indices.shape[0], num_classes, *indices.shape[1:], dtype=torch.uint8, device=indices.device)
+    new_shape = (indices.shape[0], num_classes) + indices.shape[1:]
+    onehot = torch.zeros(new_shape, dtype=torch.uint8, device=indices.device)
     return onehot.scatter_(1, indices.unsqueeze(1), 1)
 
 
 def setup_logger(
     name: Optional[str] = None,
     level: int = logging.INFO,
+    stream: Optional[TextIO] = None,
     format: str = "%(asctime)s %(name)s %(levelname)s: %(message)s",
     filepath: Optional[str] = None,
     distributed_rank: Optional[int] = None,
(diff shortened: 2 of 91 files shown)


sdesrozis:
@Devanshu24 Thank you, it looks very good to me!

Devanshu24:
Thank you for your review @sdesrozis!
So since no one suggested any changes I hope this is a good way to proceed.
I plan to add the following decorators (and their tests):
- [ ] `@deprecated` (the one currently shown)
- [ ] `@version_added`
- [ ] `@version_changed`
- [ ] `@change_default_value` changes the default value of an argument
- [ ] `@remove_arg`
- [ ] `@deprecate_kwarg`

I found these from the link @vfdev-5 gave in #1479 ([this one to be precise](https://github.com/scikit-image/scikit-image/blob/87a8806cca7fb5366b6e5ddbe5e46364b44f90fe/skimage/_shared/utils.py#L15-L214)) and little googling for similar stuff. However, I am not sure which of these would actually be helpful. I am more than happy to do all of them but I felt that if there is even one of them which will never be used it will just be some dead code.

So I am looking for your suggestions :)

vfdev-5:
Thanks for the suggestions @Devanshu24 !
Let's start with `@deprecated`, as for the others let's see a bit later. Personnally, I do not have any strong opnion about other decorators.

sdesrozis:
I agree @vfdev-5 about others decorators. Let's see first using `@deprecated` :) and we keep in mind other decorators !

Devanshu24:
I have made some minimal changes to the `@deprecated` function to cover some edge cases. I have also added tests to test most of the use-cases (I hope nothing important has been left out :) )

If there are any changes that need to be made please let me know :)

PS: I have not removed the `DeprecationWarning` from the [`setup_any_logging`](https://github.com/pytorch/ignite/blob/bebb82148e575d0acb1047920d6f60bbd31bcb93/ignite/contrib/engines/common.py#L279) function as I was not sure what to enter in the body of the function.

sdesrozis (approved):
It looks good to me !! @Devanshu24 Thank you !

vfdev-5:
@sdesrozis should we also use in the code this decorator ?

Devanshu24:
@vfdev-5 Any chages/additions needed?

Also if you could guide me on this :)
> PS: I have not removed the `DeprecationWarning` from the [`setup_any_logging`](https://github.com/pytorch/ignite/blob/bebb82148e575d0acb1047920d6f60bbd31bcb93/ignite/contrib/engines/common.py#L279) function as I was not sure what to enter in the body of the function.

vfdev-5:
@Devanshu24 I think we should update `setup_any_logging` and use your decorator to show an example of usage.

Devanshu24:
> @Devanshu24 I think we should update `setup_any_logging` and use your decorator to show an example of usage.

Okay, and in the function body should I just add a `pass` statement?

vfdev-5:
@Devanshu24 maybe we can configure the decorator such that it could either raise `DeprecationWarning` or use `warning.warn` ?

Devanshu24:
> @Devanshu24 maybe we can configure the decorator such that it could either raise `DeprecationWarning` or use `warning.warn` ?

Oh okay, I thought that as long as the function is still in the deprecated state (i.e. not yet removed) we wouldn't want the code to cause an exception and interrupt the program.
If you say so I could also add an option that raises an Exception?
So the options would be, 
- No Exception, No Warning
- Just Warning, No Exception
- Raise Exception

PS: I learnt that `raise DeprecationWarning()` causes an exception from [here](https://stackoverflow.com/a/29378596/10457734)

vfdev-5:
Well, you are right that usually we do not raise exceptions on deprecated functions. It actually may depend to security criticity of the function (for example, see `yaml.load`: https://github.com/yaml/pyyaml/issues/292). In our case, usage of `setup_any_logging` is not that critical, we just removed the implementation, so showing a warning and doing nothing is not an option for our code neither...

A bit details for your table, if function is decorated with `deprecated`:
    
    Deprecation in docs, Just Warning, No Exception
    Deprecation in docs, Raise Exception
    ~No Exception, No Warning~ (= no decorator, right?)

Devanshu24:
> Well, you are right that usually we do not raise exceptions on deprecated functions. It actually may depend to security criticity of the function (for example, see `yaml.load`: [yaml/pyyaml#292](https://github.com/yaml/pyyaml/issues/292)). In our case, usage of `setup_any_logging` is not that critical, we just removed the implementation, so showing a warning and doing nothing is not an option for our code neither...
> 

Oh okay, got to learn something new, thanks! :D
> 
> A bit details for your table, if function is decorated with `deprecated`:
> 
> * Deprecation in docs, Just Warning, No Exception
> * Deprecation in docs, Raise Exception
> * ~No Exception, No Warning~ (= no decorator, right?)

Yeah right that makes sense, I thought that maybe we would want to have a case in which we just say it is deprecated in the docs, but not raise any warnings or exceptions. But your point makes more sense

I'll make the edits, and now we will just two options:
 * Deprecation in docs, Just Warning, No Exception
 * Deprecation in docs, Raise Exception

Devanshu24:
I am getting the following error in the unit tests and during deploying the docs
```bash
NameError: name 'TextIO' is not defined
```
I think the addition was made in #1601, to be precise in the line https://github.com/pytorch/ignite/blob/c25ce51dac5a0d519362449f6b6af4a6f82e9bba/ignite/utils.py#L71

I haven't understood the error yet, thought I would ask if one of you know what it is, if not I'll start looking :)

vfdev-5:
Oh, this looks like a bug. @sdesrozis any ideas ?

Devanshu24:
Okay I figured it. It was a small thing, `TextIO` wasn't imported from `typing`

vfdev-5:
> Okay I figured it. It was a small thing, `TextIO` wasn't imported from `typing`

Normally, it should be imported here : https://github.com/pytorch/ignite/blob/c25ce51dac5a0d519362449f6b6af4a6f82e9bba/ignite/utils.py#L5

Devanshu24:
> @Devanshu24 at which moment exactly do you see that, on which command ?
```bash
pytest tests/ignite/test_utils.py -s -vvv -k test_deprecated
```
and inside the docs folder with the following command
```bash
make html
```
_____
Although after importing it both the commands work as expected :)

vfdev-5:
> Although after importing it both the commands work as expected :)

What do you mean exactly by importing it ?


Btw, netlify is also broken: https://app.netlify.com/sites/pytorch-ignite-preview/deploys/601ac2d3f9ee5000099c99a4

Devanshu24:
Also a small question, is it okay if I add the import in this PR itself or should I create a separate PR so that the tests start passing immediately.
I am okay with both the options :)

vfdev-5:
> Also a small question, is it okay if I add the import in this PR itself or should I create a separate PR so that the tests start passing immediately.
> I am okay with both the options :)

It depends on how things are broken. Currently, I do not understand if master is also broken or it is related to your code only ...

Devanshu24:
> > Also a small question, is it okay if I add the import in this PR itself or should I create a separate PR so that the tests start passing immediately.
> > I am okay with both the options :)
> 
> It depends on how things are broken. Currently, I do not understand if master is also broken or it is related to your code only ...

Oh yes yes, I am sorry there is some inconsistency on my local branch it seems. It is not the same as the master, give me 5 minutes I will concretely write what is and is not working.
Really sorry for the confusion I caused.

ignite/contrib/engines/common.py
```diff
@@ -190,7 +190,7 @@ def _setup_common_training_handlers(
 
     if with_gpu_stats:
         GpuInfo().attach(
-            trainer, name="gpu", event_name=Events.ITERATION_COMPLETED(every=log_every_iters)  # type: ignore[arg-type]
+            trainer, name="gpu", event_name=Events.ITERATION_COMPLETED(every=log_every_iters)  # type: ignore
```
Devanshu24, line 193: Changed this as it was giving an error in `flake8`
The exact error was
```
ignite/contrib/engines/common.py:193:96: F821 undefined name 'ignore'
ignite/contrib/engines/common.py:193:96: F821 undefined name 'arg'
```
The possible solutions I found were
- Change `ignore[arg-type]` to a general `ignore` (easiest to implement, but may be prone to errors)
- Similar [Issue](https://github.com/PyCQA/pyflakes/issues/475) , [solution](https://github.com/PyCQA/pyflakes/pull/455)  (Seems okay)
- Only use the allowed argument type (not sure if it is feasible)

Devanshu24:
I am not sure why the `mypy` test keeps failing, as they are working locally for me.
Trying to debug what is happening
___
Output of running it locally
```bash
$ mypy --config-file mypy.ini
Success: no issues found in 86 source files
```

vfdev-5:
@Devanshu24 are using the same version ?

Devanshu24:
> @Devanshu24 are using the same version ?

It is the same as the one on the CI i.e. `mypy 0.800`
Although https://github.com/pytorch/ignite/pull/1585/commits/2d49a48d9031290c6c091aae4d6246eaa4702985 fixes the mypy errors, the changes made in that commit pass tests on CI and also does not interfere with the local tests (which were already passing)

Devanshu24:
I think all tests are now passing (except `codecov`, but the uncovered line will be uncovered)
Please have a look at your convenience :)

ignite/contrib/engines/common.py
```diff
@@ -274,11 +275,14 @@ def empty_cuda_cache(_: Engine) -> None:
...
-def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, log_every_iters) -> None:  # type: ignore
-    raise DeprecationWarning(
-        "ignite.contrib.engines.common.setup_any_logging is deprecated since 0.4.0. and will be remove in 0.6.0. "
-        "Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc."
-    )
+@deprecated(
+    "0.4.0",  # type: ignore
+    "0.6.0",
+    ["Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc."],
+    raiseException=True,
```
vfdev-5, line 282:
```suggestion
    raise_exception=True,
```
More pythonic style

ignite/contrib/engines/common.py
```diff
@@ -274,11 +275,14 @@ def empty_cuda_cache(_: Engine) -> None:
     gc.collect()
 
 
-def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, log_every_iters) -> None:  # type: ignore
-    raise DeprecationWarning(
-        "ignite.contrib.engines.common.setup_any_logging is deprecated since 0.4.0. and will be remove in 0.6.0. "
-        "Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc."
-    )
+@deprecated(
+    "0.4.0",  # type: ignore
```
vfdev-5, line 279: `# type: ignore` is not correctly placed. Let's remove it.

vfdev-5 (commented):
Thanks for the updates @Devanshu24 !
Few other comments to address, please

ignite/contrib/engines/common.py
vfdev-5: I'm not sure to understand what happens here and why it starts to fail in the unrelated place ?
If there is an issue, then master CI should fail as well, right ?

ignite/contrib/engines/common.py
Devanshu24: Actually, this is the problem I was facing with `mypy` I originally had not placed it there, but `mypy` was passing on CI only after I placed it there.
I will again try and remove it once

ignite/contrib/engines/common.py
Devanshu24: Oh yes, thanks!

ignite/contrib/engines/common.py
Devanshu24: I get the following error on removing the line
```bash
$  mypy --config-file mypy.ini
ignite/contrib/engines/common.py:284: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
    def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, log_every_iters) -> None:
    ^
Found 1 error in 1 file (checked 86 source files)
```
I probably think it is because the decorator takes place for the actual function, hence the line numbers are a bit weird. 
If I place the `# type : ignore` in the line mentioned in the above error i.e. L284, the mypy tests pass locally but fail on CI
That is what I was referring to [here](https://github.com/pytorch/ignite/pull/1585#issuecomment-772734896)

ignite/contrib/engines/common.py
Devanshu24: Yes I was thinking the same thing. However, if I do keep it there, `pre-commit` gives me a `flake8` error mentioned above.

Devanshu24:
@vfdev-5 `mypy` is again failing on CI. It still passes locally though

ignite/contrib/engines/common.py
vfdev-5: Do you use the same flake8 version as on the CI ?

vfdev-5:
@Devanshu24 I have to check that locally from my side.

@gruebel do you have an idea about why mypy is failing ?

Devanshu24:
Okayy I think I get it, the `Python3.8` CI is still passing, which is my local pyhton version too
There is some issue with `3.7` and `3.6`

ignite/contrib/engines/common.py
Devanshu24: Yes I think so
```bash
$ flak8 --version
3.8.4 (mccabe: 0.6.1, pycodestyle: 2.6.0, pyflakes: 2.2.0) CPython 3.8.5 on Linux
```
And by looking at the logs, the CI is also using the same version for all of them

gruebel:
@vfdev-5 I will take a look today, I also left some comments on the typing.

ignite/utils.py
```diff
@@ -169,3 +171,32 @@ def manual_seed(seed: int) -> None:
         np.random.seed(seed)
     except ImportError:
         pass
+
+
+def deprecated(deprecated_in: str, removed_in: str = "", reasons: list = [], raise_exception: bool = False) -> Callable:
```
gruebel, line 176: Please use the appropriate generic Type for `list`, then you can make it more precise `List[str]` 

And is it intentional to use a mutable object as a default for `reasons`?

vfdev-5:
Thanks a lot @gruebel !

ignite/utils.py
```diff
@@ -169,3 +171,32 @@ def manual_seed(seed: int) -> None:
...
+    def decorator(func: F) -> F:
+        func_doc = func.__doc__ if func.__doc__ else ""
+        deprecation_warning = (
+            f"This function has been deprecated since version `{deprecated_in}`"
+            + (f" and will be removed in version `{removed_in}`" if removed_in else "")
+            + ".\n Please refer to the documentation for more details."
+        )
+
+        @functools.wraps(func)
+        def wrapper(*args: int, **kwargs: float) -> Callable:
```
gruebel, line 189: shouldn't it be `def wrapper(*args: Any, **kwargs: Any) -> Callable:` ?

ignite/utils.py
```diff
@@ -169,3 +171,32 @@ def manual_seed(seed: int) -> None:
...
+            if raise_exception:
+                raise DeprecationWarning(deprecation_warning)
+            warnings.warn(deprecation_warning, DeprecationWarning, stacklevel=2)
+            return func(*args, **kwargs)
+
+        appended_doc = f".. deprecated:: {deprecated_in}" + ("\n\n\t" if len(reasons) else "")
+
+        for reason in reasons:
+            appended_doc += "\n\t- " + reason
+        wrapper.__doc__ = "**Deprecated function**." + "\n\n    " + func_doc + appended_doc
```
gruebel, line 199: how about using f-strings here?

ignite/utils.py
Devanshu24: Thanks @gruebel !

> Please use the appropriate generic Type for list, then you can make it more precise List[str]

Sure!

> And is it intentional to use a mutable object as a default for reasons?

Do you mean that we could use something like tuples, I don't think there is any issue with using tuples here. Should we shift to tuples, I did not know if one was better than the other

ignite/utils.py
Devanshu24: I can't find the `mypy` docs link for this, but this [link](https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values) is related. (As in I referred to this while deciding the types)

ignite/utils.py
Devanshu24: Sure, not a problem!

ignite/utils.py
vfdev-5: @Devanshu24 previously we annotated args and kwargs as Any as function's arg and kwargs can be `Tuple[Any]` and `Dict[str, Any]`. Let's keep it with Any. 

In the mypy docs they give an example where they have ints for args and float for kwargs.

Devanshu24: Okay, I wasn't able to get that. I will change it to `Any`. And also read up a little more on type annotations to get a better understanding.

ignite/utils.py
gruebel: if you use a mutable object like `list` here, then some funny surprises can happen. 

Here a good link for extra infos https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

ignite/utils.py
gruebel: we used mainly `Any` and `Any` for both, but I also like `Dict[str, Any]` for kwargs.

gruebel:
So, I digged into the problem it seems like to be an issue with `mypy` and how it handles multiline functions/decorators. The decorator belongs to the function, so `mypy` gets messed up and blames the first line. Interestingly it was fixed for Python 3.8, but not lower https://github.com/python/mypy/issues/3251.

So, to make `mypy` happy it needs to be like @Devanshu24 implemented it before
```
@deprecated(
    "0.4.0",  # type: ignore
    "0.6.0",
    ["Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc."],
    raise_exception=True,
)
def setup_any_logging(logger, logger_module, trainer, optimizers, evaluators, log_every_iters) -> None:
    pass
```

or just add some type hints and then you can remove the `# type: ignore` part 😄 
```
@deprecated(
    "0.4.0",
    "0.6.0",
    ["Please use instead: setup_tb_logging, setup_visdom_logging or setup_mlflow_logging etc."],
    raise_exception=True,
)
def setup_any_logging(
    logger: BaseLogger,
    logger_module: Any,
    trainer: Engine,
    optimizers: Optional[Union[Optimizer, Dict[str, Optimizer], Dict[None, Optimizer]]],
    evaluators: Optional[Union[Engine, Dict[str, Engine]]],
    log_every_iters: int,
) -> None:
    pass
```

I would prefer just to add the types, which approach do you prefer @vfdev-5 ?

vfdev-5:
Thank a lot @gruebel for investigating the issue 💪 !
Yes, I also prefer the second solution with type hints as you suggested.
@Devanshu24 could you please apply those changes .

Devanshu24:
Thanks @gruebel !
Sure @vfdev-5 will get them done today!

Devanshu24:
Made the suggested changes, please have a look at your convenience :)

ignite/contrib/engines/common.py
vfdev-5: I'll revert this change and see if it fails here.

vfdev-5 (approved):
Great work @Devanshu24 ! Thanks !

PR #1589, opened 2021-01-29
[create_supervised_trainer] add automatic mixed precision

### LINKED ISSUE #1235: Add more options to create_supervised_trainer
## 🚀 Feature

Idea is to add more options to `create_supervised_trainer` helper:
- native AMP option similar to https://github.com/lidq92/LinearityIQA/blob/master/modified_ignite_engine.py 
  - we may either use nvidia/apex if installed or torch native amp if v1.6.0. Otherwise, raise an error.
- grads cliping as proposed #419

For Hacktoberfest/PyDataGlobal contributors, feel free to ask questions for details if any and say that you would like to tackle the issue.
Please, take a look at [CONTRIBUTING guide](https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md).

ydcjeff:
Fixes #1235 

Description: Add automatic mixed precision using `torch.cuda.amp` and `apex`.

Usage:
```py
# using autocast only
trainer = create_supervised_trainer(amp_mode='amp')

# using autocast and default scaler
trainer = create_supervised_trainer(amp_mode='amp', scaler=True)
# trainer state will have attribute scaler
print(trainer.state.scaler)
<torch.cuda.amp.grad_scaler.GradScaler object at 0x7f8e0dac7b80>

# using autocast and custom scaler
# trainer state will not have attribute scaler if scaler instance is passed
scaler = GradScaler(2**10)
trainer = create_supervised_trainer(amp_mode='amp', scaler=scaler)

# using apex
trainer = create_supervised_trainer(amp_mode='apex')

# scaler will be ignored, warning will show up
trainer = create_supervised_trainer(amp_mode='apex', scaler=True)
```

Check list:

https://deploy-preview-1589--pytorch-ignite-preview.netlify.app/engine.html#


ignite/engine/__init__.py
```diff
@@ -91,20 +101,54 @@ def create_supervised_trainer(
 
     device_type = device.type if isinstance(device, torch.device) else device
     on_tpu = "xla" in device_type if device_type is not None else False
+    has_native_amp, has_apex_amp = False, False
 
     if on_tpu and not idist.has_xla_support:
         raise RuntimeError("In order to run on TPU, please install PyTorch XLA")
 
+    if amp and LooseVersion(torch.__version__) >= LooseVersion("1.6.0"):
```
vfdev-5, line 109: I think we can not do amp with TPU ? Maybe, we should explicitly raise an error if on_tpu is True and user passes amp and scaler...

ignite/engine/__init__.py
```diff
@@ -91,20 +101,54 @@ def create_supervised_trainer(
...
+        from torch.cuda.amp import autocast
+
+        has_native_amp = True
+    elif amp:
+        try:
+            from apex import amp as apex_amp
+
+            has_apex_amp = True
```
vfdev-5, line 117: maybe we can even initialize model and optimizer here with amp ?

ignite/engine/__init__.py
```diff
@@ -91,20 +101,54 @@ def create_supervised_trainer(
...
+            with autocast(enabled=amp):
+                y_pred = model(x)
+                loss = loss_fn(y_pred, y)
+        else:
+            y_pred = model(x)
+            loss = loss_fn(y_pred, y)
+
+        if has_native_amp:
+            scaler.scale(loss).backward()
+            scaler.unscale_(optimizer)
```
vfdev-5, line 135: Let's perform unscale below before the line 145.

vfdev-5 (commented):
Thanks a lot for the PR @ydcjeff !

ignite/engine/__init__.py
ydcjeff: As per apex docs, we can't do that.
> amp.initialize should be called **after** you have finished constructing your model(s) and optimizer(s), but **before** you send your model through any DistributedDataParallel wrapper. See Distributed training in the Imagenet example.

`amp.initialize` should be called **after** model construction, but **before** setting DDP or DP, so we have no way to do that if user pass output of `idist.auto_model(model)`.

And I think, given user to initialize model is the cleanest way in terms of code and usability.

ignite/engine/__init__.py
ydcjeff: Any reason why? We need another `if` check.

ignite/engine/__init__.py
ydcjeff: I totally forgot again :\

ignite/engine/__init__.py
vfdev-5: Normally, we do not need to unscale the gradients. So if user do not clip, it is an unnecessary op, right?
However, if user provides grad_norm_kwargs we will unscale and perform the clipping.

ignite/engine/__init__.py
```diff
@@ -91,20 +101,54 @@ def create_supervised_trainer(
...
     on_tpu = "xla" in device_type if device_type is not None else False
+    has_native_amp, has_apex_amp = False, False
 
     if on_tpu and not idist.has_xla_support:
         raise RuntimeError("In order to run on TPU, please install PyTorch XLA")
 
+    if amp and LooseVersion(torch.__version__) >= LooseVersion("1.6.0"):
+        from torch.cuda.amp import autocast
+
+        has_native_amp = True
```
vfdev-5, line 112: As for the scaler, we can create a default one if user do not provide one in the args.

ignite/engine/__init__.py
vfdev-5: Yes, make sense. Thanks!

ignite/engine/__init__.py
ydcjeff: The docs say even if we don't do `unscale_`, it performs internally when we called ,`step`. So it doesn't relate to grad clip.

vfdev-5: OK, I see. However, I think it would be more simple to understand the logic as 
```python
scaler.unscale_(optimizer)
torch.nn.utils.clip_grad_norm_(model.parameters(), **grad_norm_kwargs)
```
and avoid user's questions where we just point out to the official docs as you did right now ;)

ignite/engine/__init__.py
ydcjeff: For the scaler, I let users create by themselves if they want to save `state_dict` since engine instance is the only return.

vfdev-5: Thanks, that's a great point to think about ! We can add `apex.amp` or `scaler` to `engine.state` as attributes and also add to `state_dict_user_keys` to checkpoint if trainer is checkpointed: https://pytorch.org/ignite/engine.html#ignite.engine.engine.Engine.state_dict

ydcjeff: I see! Great
So should we still have `scaler` arg?

vfdev-5: Good question! Maybe, it is unnecessary, let's see...

ydcjeff: For now, I will push current progress as I want to run GPU test on colab.

vfdev-5: I think I can enable GPU tests for PRs now as now we can filter cancel workflow if code was not updated...

vfdev-5:
Currently, launched it manually : https://app.circleci.com/pipelines/github/pytorch/ignite/1195/workflows/27d5b840-72bb-41e1-8d1c-84640f1f623c but I think either next commit or new PR will run auto on GPUs

ydcjeff:
> Currently, launched it manually : https://app.circleci.com/pipelines/github/pytorch/ignite/1195/workflows/27d5b840-72bb-41e1-8d1c-84640f1f623c but I think either next commit or new PR will run auto on GPUs 

Thank you!

ignite/engine/__init__.py
ydcjeff: @vfdev-5 we can't create scaler internally. `GradScaler` is used for gradients that are near to zero after autocast.
If the gradients are not near zero, we don't need that. It's user responsibility to pass scaler depending on their model's gradient.
https://pytorch.org/tutorials/recipes/recipes/amp_recipe.html#adding-autocast

And also, if `amp` and `scaler` are passed but no `grad_norm_kwargs`, `scaler.step()` will do `unscale_()`
so it's ok to call only `unscale_()` if `grad_norm_kwargs` are passed.

ydcjeff:
I need help with the tests specifically with `apex` and `GradScaler`.

ignite/engine/__init__.py
vfdev-5: I think what we could do is the following. We follow the guidelines from amp recipe "All together: Automatic Mixed Precision" no matter values of the gradient. GradScaler is created internally as discussed above and we keep scaler arg which could be interpreted as following: 
- None -> create internal one, 
- instance of GradScaler -> use provided one
- False -> disable internal one

```python
def create_supervised_trainer(
    model: torch.nn.Module,
    optimizer: torch.optim.Optimizer,
    output_transform: Callable = lambda x, y, y_pred, loss: loss.item(),
    deterministic: bool = False,
    amp: bool = False,
    scaler: Optional[Union[bool, "torch.cuda.amp.GradScaler"]] = None,  # type: ignore[assignment]
    **grad_norm_kwargs: Any,
) -> Engine:
```
What do you think ?

vfdev-5: > And also, if amp and scaler are passed but no grad_norm_kwargs, scaler.step() will do unscale_()
so it's ok to call only unscale_() if grad_norm_kwargs are passed.

This is what we do now, right ?

vfdev-5:
> I need help with the tests specifically with `apex` and `GradScaler`.

I'll try to implement something from my side and we'll see.

ignite/engine/__init__.py
ydcjeff: > - None -> create internal one,
> - instance of GradScaler -> use provided one
> - False -> disable internal one

I think `bool` and instance of GradScaler is enough.
- True -> create internal one,
- instance of GradScaler -> use provided one
- False -> disable internal one

vfdev-5:
We discussed this PR and related issue with the team and we think that we should explore a bit different approach. Helper method `create_supervised_trainer` is roughly made of 2 things : `update` function definition and `Engine` setup.

Probably, it would more helpful to provide public methods like:
- `supervised_training_step`
- `supervised_training_step_tpu`
- `supervised_training_step_apex`
- `supervised_training_step_amp`

and inside `create_supervised_trainer` we could setup the trainer according to provided options without lots of if/else. Maybe, we can skip for instance grad norm.

Basically, the idea is something like that : 
```python
```


cc @sdesrozis

sdesrozis:
That would be great for users to have these functions, helpful to check under the hood. 

My thoughts on this topic is about the update function. The dream would be to pass a generic function and have an automatic (or close) tools to adapt to features like amp, tpu, etc.

ydcjeff:
Shall we also accept `scaler` argument or create only internally?

vfdev-5 (commented):
Thanks for the update!

vfdev-5:
@sdesrozis can you review the PR please

ignite/engine/__init__.py
```diff
@@ -42,16 +45,217 @@ def _prepare_batch(
...
+            tuple of tensors `(batch_x, batch_y)`.
+        output_transform (callable): function that receives 'x', 'y', 'y_pred', 'loss' and returns value
+            to be assigned to engine's state.output after each iteration. Default is returning `loss.item()`.
+
+    Returns:
+        Callable: update function.
+    """
+
+    def update(engine: Engine, batch: Sequence[torch.Tensor]) -> Union[Any, Tuple[torch.Tensor]]:
+        x, y = _train_zero_grad_prepare_batch(model, optimizer, device, non_blocking, prepare_batch, batch)
```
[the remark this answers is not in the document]
ydcjeff: Already changed locally.

ignite/engine/__init__.py
```diff
@@ -118,68 +320,31 @@ def create_supervised_trainer(
...
+    if amp_mode == "amp":
+        _update = supervised_trainer_step_amp(
+            model, optimizer, loss_fn, device, non_blocking, prepare_batch, output_transform, scaler
+        )
+    elif amp_mode == "apex":
+        _update = supervised_trainer_step_apex(
+            model, optimizer, loss_fn, device, non_blocking, prepare_batch, output_transform
+        )
+
+    if on_tpu:
```
sdesrozis, line 340:
```suggestion
    elif on_tpu:
```
Function `_check_arg` ensures that all parameters is coherent if I understand correctly. So 
```
flag = _check_arg(...)

# no need to understand the rules in _check_arg
# apply the relevant mode
if flag == "amp":
   ...
elif flag == "apex":
   ...
elif ...
```

ydcjeff: Thanks @sdesrozis

tests/ignite/engine/test_create_supervised.py
```diff
@@ -105,12 +119,59 @@ def test_create_supervised_trainer_traced_with_cpu():
...
 
+def test_create_supervised_trainer_apex_error():
+    with pytest.raises(
+        ModuleNotFoundError, match="Please install apex from https://github.com/nvidia/apex to use amp_mode='apex'."
+    ):
+        _test_create_supervised_trainer(amp_mode="apex")
+
+
+@pytest.mark.skipif(LooseVersion(torch.__version__) > LooseVersion("1.6.0"), reason="Skip if > 1.6.0.")
+def test_create_supervised_trainer_amp_error():
```
ydcjeff, line 130: Can we make this test work in torch > 1.6?
I don't know how to make it work if torch > 1.6

ignite/engine/__init__.py
```diff
@@ -1,7 +1,9 @@
 from collections.abc import Mapping
 from typing import Any, Callable, Dict, Optional, Sequence, Tuple, Union
+from warnings import warn
 
 import torch
+from torch.cuda import amp
```
sdesrozis, line 6: Remove this line

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+    return "tpu" if on_tpu else amp_mode
+
+
+def supervised_trainer_step(
+    model: torch.nn.Module,
+    optimizer: torch.optim.Optimizer,
+    loss_fn: Union[Callable, torch.nn.Module],
+    device: Optional[Union[str, torch.device]],
+    non_blocking: bool,
+    prepare_batch: Callable,
```
sdesrozis, line 70:
```suggestion
    prepare_batch: Callable = _prepare_batch,
```

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+    return update
+
+
+def supervised_trainer_step_amp(
+    model: torch.nn.Module,
+    optimizer: torch.optim.Optimizer,
+    loss_fn: Union[Callable, torch.nn.Module],
+    device: Optional[Union[str, torch.device]],
+    non_blocking: bool,
+    prepare_batch: Callable,
```
sdesrozis, line 112:
```suggestion
    prepare_batch: Callable = _prepare_batch,
```

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+    return update
+
+
+def supervised_trainer_step_apex(
+    model: torch.nn.Module,
+    optimizer: torch.optim.Optimizer,
+    loss_fn: Union[Callable, torch.nn.Module],
+    device: Optional[Union[str, torch.device]],
+    non_blocking: bool,
+    prepare_batch: Callable,
```
sdesrozis, line 168:
```suggestion
    prepare_batch: Callable = _prepare_batch,
```

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+
+
+def supervised_trainer_step(
+    model: torch.nn.Module,
+    optimizer: torch.optim.Optimizer,
+    loss_fn: Union[Callable, torch.nn.Module],
+    device: Optional[Union[str, torch.device]],
+    non_blocking: bool,
+    prepare_batch: Callable,
+    output_transform: Callable,
```
sdesrozis, line 71:
```suggestion
    output_transform: Callable = lambda x, y, y_pred, loss: loss.item(),
```

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+    Args:
+        model (torch.nn.Module): the model to train.
+        optimizer (torch.optim.Optimizer): the optimizer to use.
+        loss_fn (torch.nn loss function): the loss function to use.
+        device (str): device type specification (default: None).
+            Applies to batches after starting the engine. Model *will not* be moved.
+            Device can be CPU, GPU or TPU.
+        non_blocking (bool): if True and this copy is between CPU and GPU, the copy may occur asynchronously
+            with respect to the host. For other cases, this argument has no effect.
+        prepare_batch (callable): function that receives `batch`, `device`, `non_blocking` and outputs
```
sdesrozis, line 84:
```suggestion
        prepare_batch (callable, optional): function that receives `batch`, `device`, `non_blocking` and outputs
```

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+            tuple of tensors `(batch_x, batch_y)`.
+        output_transform (callable): function that receives 'x', 'y', 'y_pred', 'loss' and returns value
```
sdesrozis, line 86:
```suggestion
        output_transform (callable, optional): function that receives 'x', 'y', 'y_pred', 'loss' and returns value
```

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+
+    return "tpu" if on_tpu else amp_mode
+
+
+def supervised_trainer_step(
+    model: torch.nn.Module,
+    optimizer: torch.optim.Optimizer,
+    loss_fn: Union[Callable, torch.nn.Module],
+    device: Optional[Union[str, torch.device]],
+    non_blocking: bool,
```
sdesrozis, line 69:
```suggestion
    non_blocking: bool = False,
```

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+    if amp_mode and on_tpu:
+        raise ValueError("amp_mode cannot be used with xla device. Consider using amp_mode=None or device='cuda'.")
+
+    if scaler is not None and not amp_mode:
+        warn("scaler argument is provided, but amp_mode is not. Consider using amp_mode='amp'.")
+
+    return "tpu" if on_tpu else amp_mode
+
+
+def supervised_trainer_step(
```
sdesrozis, line 64: Please, add similar default values than `create_supervised_trainer`. It should help users who want to use directly these helpers.

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+    model: torch.nn.Module,
+    optimizer: torch.optim.Optimizer,
+    loss_fn: Union[Callable, torch.nn.Module],
+    device: Optional[Union[str, torch.device]],
+    non_blocking: bool,
+    prepare_batch: Callable,
+    output_transform: Callable,
+) -> Callable:
+    """Helper function defined the training step.
```
sdesrozis, line 73: Update the docstring wrt default values.

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+        y_pred = model(x)
+        loss = loss_fn(y_pred, y)
+        loss.backward()
+        optimizer.step()
+        return output_transform(x, y, y_pred, loss)
+
+    return update
+
+
+def supervised_trainer_step_amp(
```
sdesrozis, line 106: Update default values

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+    model: torch.nn.Module,
+    optimizer: torch.optim.Optimizer,
+    loss_fn: Union[Callable, torch.nn.Module],
+    device: Optional[Union[str, torch.device]],
+    non_blocking: bool,
+    prepare_batch: Callable,
+    output_transform: Callable,
+    scaler: Optional["torch.cuda.amp.GradScaler"],
+) -> Callable:
+    """Helper function defined the training step for torch.cuda.amp.
```
sdesrozis, line 116: Update the docstring wrt default values.

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+def supervised_trainer_step_apex(
+    model: torch.nn.Module,
+    optimizer: torch.optim.Optimizer,
+    loss_fn: Union[Callable, torch.nn.Module],
+    device: Optional[Union[str, torch.device]],
+    non_blocking: bool,
+    prepare_batch: Callable,
+    output_transform: Callable,
+) -> Callable:
+    """Helper function defined the training step for apex.
```
sdesrozis, line 171: Update the docstring wrt default values.

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+            scaler.update()
+        else:
+            loss.backward()
+            optimizer.step()
+        return output_transform(x, y, y_pred, loss)
+
+    return update
+
+
+def supervised_trainer_step_apex(
```
sdesrozis, line 162: Update default values

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+        loss = loss_fn(y_pred, y)
+        with apex_amp.scale_loss(loss, optimizer) as scaled_loss:
+            scaled_loss.backward()
+        optimizer.step()
+        return output_transform(x, y, y_pred, loss)
+
+    return update
+
+
+def supervised_trainer_step_tpu(
```
sdesrozis, line 210: Update default values

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+    model: torch.nn.Module,
+    optimizer: torch.optim.Optimizer,
+    loss_fn: Union[Callable, torch.nn.Module],
+    device: Optional[Union[str, torch.device]],
+    non_blocking: bool,
+    prepare_batch: Callable,
+    output_transform: Callable,
+) -> Callable:
+    """Helper function defined the training step for tpu.
```
sdesrozis, line 219: Update the docstring wrt default values.

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+        output_transform (callable): function that receives 'x', 'y', 'y_pred', 'loss' and returns value
+            to be assigned to engine's state.output after each iteration. Default is returning `loss.item()`.
+        scaler (torch.cuda.amp.GradScaler, optional): GradScaler instance for gradient scaling if `torch>=1.6.0`
+            and ``amp_mode`` is ``amp``. If ``amp_mode`` is ``apex``, this argument will be ignored.
+
+    Returns:
+        Callable: update function.
+    """
+
+    if hasattr(torch.cuda, "amp") and hasattr(torch.cuda.amp, "autocast"):
```
sdesrozis, line 138: Why not try/except ?
```
    try:
        from torch.cuda.amp import autocast
    except ModuleNotFoundError:
        raise ModuleNotFoundError("autocast cannot be imported, please install torch>=1.6.0.")
```

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+        prepare_batch (callable): function that receives `batch`, `device`, `non_blocking` and outputs
+            tuple of tensors `(batch_x, batch_y)`.
+        output_transform (callable): function that receives 'x', 'y', 'y_pred', 'loss' and returns value
+            to be assigned to engine's state.output after each iteration. Default is returning `loss.item()`.
+
+    Returns:
+        Callable: update function.
+    """
+
+    def update(engine: Engine, batch: Sequence[torch.Tensor]) -> Union[Any, Tuple[torch.Tensor]]:
```
sdesrozis, line 239: `xm` should be tested here (as `amp` and `apex`). This is important to raise an explicit message rather than have an import error (line 246).

Please remove line 14-15
```
if idist.has_xla_support:
```

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
     )
 
 
+def _check_arg(on_tpu: bool, amp_mode: Optional[str], scaler: Optional["torch.cuda.amp.GradScaler"]) -> Optional[str]:
```
sdesrozis, line 50: Move before `create_supervised_trainer` ?

ignite/engine/__init__.py
```diff
@@ -41,6 +47,208 @@ def _prepare_batch(
...
+        raise AttributeError("autocast cannot be imported, please install torch>=1.6.0.")
+
+    def update(engine: Engine, batch: Sequence[torch.Tensor]) -> Union[Any, Tuple[torch.Tensor]]:
+        model.train()
+        optimizer.zero_grad()
+        x, y = prepare_batch(batch, device=device, non_blocking=non_blocking)
+        with autocast(enabled=True):
+            y_pred = model(x)
+            loss = loss_fn(y_pred, y)
+        if scaler:
```
sdesrozis, line 150: Does this respect the comment https://github.com/pytorch/ignite/pull/1589#discussion_r567254484 ?

A `scaler` should be created internally if `scaler is None`.

sdesrozis:
@ydcjeff I added some comments. 

Question, should we use amp / apex / tpu for inference ?

tests/ignite/engine/test_create_supervised.py
```diff
@@ -42,7 +54,9 @@ def _test_create_supervised_trainer(
     if model_device == trainer_device or ((model_device == "cpu") ^ (trainer_device == "cpu")):
         state = trainer.run(data)
 
-        assert state.output == approx(17.0)
+        if amp_mode:
```
sdesrozis, line 57: Why do not test else case ? I wonder if this test is really relevant...

tests/ignite/engine/test_create_supervised.py
```diff
@@ -42,7 +54,9 @@ def _test_create_supervised_trainer(
...
+            assert state.output[0].dtype is torch.float16
```
sdesrozis, line 58:
```suggestion
            assert state.output[0].dtype is torch.half
```

tests/ignite/engine/test_create_supervised.py
sdesrozis: `amp` is available since PyTorch 1.6 so it can't have an error raised with torch > 1.6.

I don't understand what you mean...

sdesrozis:
@ydcjeff  I added some comments.

Question, should we use amp for inference ? (I mean autocast context).

ignite/engine/__init__.py
ydcjeff: Yes, I mistakenly imported it.

tests/ignite/engine/test_create_supervised.py
ydcjeff: Any difference?

ignite/engine/__init__.py
ydcjeff: Creating `scaler` internally has not been added.

ignite/engine/__init__.py
ydcjeff: For readability?
I defined after `_prepare_batch` just for placing internal functions in one place.

tests/ignite/engine/test_create_supervised.py
sdesrozis: No, comprehensibility for reader.

ignite/engine/__init__.py
sdesrozis: This function is defined line 50 but used line 250. The idea is to reduce the gap for readability, yes.

ydcjeff: Should `_prepare_batch` move too?

ydcjeff:
> @ydcjeff  I added some comments.
> 
> Question, should we use amp for inference ? (I mean autocast context).

For validation? This PR is only for training for now.
Validation will come in later PR.

tests/ignite/engine/test_create_supervised.py
ydcjeff: That test is for checking `cuda.amp` can be imported or not.
Since tests are running with latest pytorch version, this test is skipped in the unit tests.
And I want to make this test run for coverage

vfdev-5: You can use mock.patch with a context manager to temporarily patch `cuda.amp` module and thus be able to raise the tested error. Take a look here for example: https://github.com/pytorch/ignite/blob/master/tests/ignite/contrib/metrics/test_gpu_info.py#L91

ignite/engine/__init__.py
sdesrozis: The docstring must be updated, see docstring of `create_supervised_trainer`. Thanks!

ignite/engine/__init__.py
sdesrozis: Same, docstring not updated

ignite/engine/__init__.py
sdesrozis: Ok ! Please, do not resolve the conversation until it is done. Thanks !

ignite/engine/__init__.py
sdesrozis: Same about docstring

ignite/engine/__init__.py
sdesrozis: Same about docstring

ignite/engine/__init__.py
```diff
@@ -41,6 +42,221 @@ def _prepare_batch(
...
+            (default: False)
+
+    Returns:
+        Tuple[Callable, Union[bool, torch.cuda.amp.GradScaler]]: update function and scaler
+    """
+
+    try:
+        from torch.cuda.amp import GradScaler, autocast
+    except ModuleNotFoundError:
+        raise ModuleNotFoundError("autocast cannot be imported, please install torch>=1.6.0.")
```
sdesrozis, line 123:
```suggestion
        raise ModuleNotFoundError("Please install torch>=1.6.0 to use amp_mode='amp'.")
```
what do you think ?

ignite/engine/__init__.py
ydcjeff: `scaler` options for `True`, `False` and GradScaler instance has been added.

ignite/engine/__init__.py
ydcjeff: Thanks! added!

ydcjeff:
Apex has to be installed to run the tests on gpu @vfdev-5

ignite/engine/__init__.py
```diff
@@ -41,6 +42,221 @@ def _prepare_batch(
...
+    prepare_batch: Callable = _prepare_batch,
+    output_transform: Callable = lambda x, y, y_pred, loss: loss.item(),
+) -> Callable:
+    """Factory function for supervised training.
+
+    Args:
+        model (torch.nn.Module): the model to train.
+        optimizer (torch.optim.Optimizer): the optimizer to use.
+        loss_fn (torch.nn loss function): the loss function to use.
+        device (str): device type specification (default: None).
```
sdesrozis, line 60:
```suggestion
        device (str, optional): device type specification (default: None).
```
Please, have a look line 278-283 and update the docstring according to `create_supervised_trainer`. It misses `optional` to `device`, `non_blocking`, `prepare_batch` and `output_transform`.

This should be done for `supervised_training_step` and
* `supervised_training_step_amp` line 87
* `supervised_training_step_apex` line 147
* `supervised_training_step_tpu` line 195

Thank you !

ignite/engine/__init__.py
```diff
@@ -41,6 +42,221 @@ def _prepare_batch(
...
+def supervised_training_step_amp(
+    model: torch.nn.Module,
+    optimizer: torch.optim.Optimizer,
+    loss_fn: Union[Callable, torch.nn.Module],
+    device: Optional[Union[str, torch.device]] = None,
+    non_blocking: bool = False,
+    prepare_batch: Callable = _prepare_batch,
+    output_transform: Callable = lambda x, y, y_pred, loss: loss.item(),
+    scaler: Union[bool, "torch.cuda.amp.GradScaler"] = False,
+) -> Tuple[Callable, Union[bool, "torch.cuda.amp.GradScaler"]]:
```
sdesrozis, line 96:
```suggestion
) -> Tuple[Callable, Optional["torch.cuda.amp.GradScaler"]]:
```
The idea is to return a valid scaler (created internally or given by user) or `None` (meaning `scaler` bool is `False`.

ignite/engine/__init__.py
```diff
@@ -69,48 +287,71 @@ def create_supervised_trainer(
...
+        _update = supervised_training_step_tpu(
+            model, optimizer, loss_fn, device, non_blocking, prepare_batch, output_transform
+        )
+    else:
+        _update = supervised_training_step(
+            model, optimizer, loss_fn, device, non_blocking, prepare_batch, output_transform
+        )
 
     trainer = Engine(_update) if not deterministic else DeterministicEngine(_update)
+    if scaler and isinstance(scaler, bool) and mode == "amp":
```
sdesrozis, line 353:
```suggestion
    if _scaler:
```

ignite/engine/__init__.py
```diff
@@ -69,48 +287,71 @@ def create_supervised_trainer(
...
-        if on_tpu:
-            xm.optimizer_step(optimizer, barrier=True)
-        else:
-            optimizer.step()
-
-        return output_transform(x, y, y_pred, loss)
+    mode = _check_arg(on_tpu, amp_mode, scaler)
+
+    if mode == "amp":
+        _update, scaler_ = supervised_training_step_amp(
```
sdesrozis, line 336:
```suggestion
        _update, _scaler = supervised_training_step_amp(
```
coding style should be consistent.

ignite/engine/__init__.py
ydcjeff: I am afraid that it would result in variable undefined in not scaler cases.

ydcjeff:
Thank you for the review @sdesrozis , I have addresses your review.

ignite/engine/__init__.py
sdesrozis: of course, my mistake !

sdesrozis: Maybe we should include rules in `_check_arg` 
```
mode, _scaler = _check_arg(on_tpu, amp_mode, scaler) # build scaler here or None

if mode == "amp":
        _update = supervised_training_step_amp(
            model, optimizer, loss_fn, device, non_blocking, prepare_batch, output_transform, _scaler
        )
elif ...

trainer = Engine(_update) if not deterministic else DeterministicEngine(_update)

if _scaler:
    trainer.state.scaler = _scaler
```

What do you think ? I insist because line 358 is related to a rule used inside `supervised_training_step_amp`. Therefore, change this rule means modify 2 places in code...

ignite/engine/__init__.py
```diff
@@ -41,6 +42,226 @@ def _prepare_batch(
...
+        non_blocking (bool, optional): if True and this copy is between CPU and GPU, the copy may occur asynchronously
+            with respect to the host. For other cases, this argument has no effect.
+        prepare_batch (callable, optional): function that receives `batch`, `device`, `non_blocking` and outputs
+            tuple of tensors `(batch_x, batch_y)`.
+        output_transform (callable, optional): function that receives 'x', 'y', 'y_pred', 'loss' and returns value
+            to be assigned to engine's state.output after each iteration. Default is returning `loss.item()`.
+
+    Returns:
+        Callable: update function.
+    """
```
vfdev-5, line 72: @ydcjeff let's also add `versionadded` tags with 0.5.0 for all of those docstrings

ignite/engine/__init__.py
ydcjeff: Yes, I agree, it's better API
Thank you!

ignite/engine/__init__.py
```diff
@@ -335,11 +345,11 @@ def create_supervised_trainer(
 
     device_type = device.type if isinstance(device, torch.device) else device
     on_tpu = "xla" in device_type if device_type is not None else False
-    mode = _check_arg(on_tpu, amp_mode, scaler)
+    mode, _scaler = _check_arg(on_tpu, amp_mode, scaler)
 
     if mode == "amp":
         _update, _scaler = supervised_training_step_amp(
```
sdesrozis, line 351:
```suggestion
        _update = supervised_training_step_amp(
```

No need to return the scaler now, right ?

ignite/engine/__init__.py
```diff
@@ -109,44 +111,36 @@ def supervised_training_step_amp(
...
         else:
             loss.backward()
             optimizer.step()
         return output_transform(x, y, y_pred, loss)
 
-    if scaler:
-        return update, scaler_
-    else:
-        return update, None
+    return update, scaler
```
sdesrozis, line 143:
```suggestion
    return update
```

Return scaler is useless now

sdesrozis:
@ydcjeff Thank you for this PR! I left my last comments, sorry for that !! 

Thank you again !

sdesrozis:
@ydcjeff It looks good to me. 

The last thing is to handle test using mock for amp and it could be merged.

ydcjeff:
Can we run gpu tests?
I have updated the tests so that gpu tests can pass.
Also, can someone help me about mocking amp test? I tried as per [this test](https://github.com/pytorch/ignite/blob/e81767b594ec0609a2e0899923610c7e61f0327e/tests/ignite/contrib/metrics/test_gpu_info.py#L91), but it didn't help.
It mocked to give error on calling, but here this case is for importing module, I tried google it, not much info found.
Thanks!

vfdev-5:
@ydcjeff here is an example on how to sort of remove `torch.cuda.amp` module with patch
```python
from unittest.mock import patch


def test_no_torch_cuda_amp_package():

    new_modules = {
        "torch.cuda.amp": None, 
        "torch.cuda.amp.grad_scaler": None,
        "torch.cuda.amp.autocast_mode": None        
    }
    
    with patch.dict("sys.modules", new_modules):
        import torch.cuda.amp

        
test_no_torch_cuda_amp_package()
> ModuleNotFoundError: import of torch.cuda.amp halted; None in sys.modules

```

ydcjeff:
It works! Thank you @vfdev-5

ignite/engine/__init__.py
```diff
@@ -69,48 +302,71 @@ def create_supervised_trainer(
...
-        loss = loss_fn(y_pred, y)
-        loss.backward()
-
-        if on_tpu:
-            xm.optimizer_step(optimizer, barrier=True)
-        else:
-            optimizer.step()
-
-        return output_transform(x, y, y_pred, loss)
+    mode, scaler_ = _check_arg(on_tpu, amp_mode, scaler)
```
sdesrozis, line 348:
```suggestion
    mode, _scaler = _check_arg(on_tpu, amp_mode, scaler)
```
Please prefix local variables / function by `_` rather than suffix.

sdesrozis (approved):
@ydcjeff Thank you! It looks good to me !! Sorry for the multiple iterations and comments!

ydcjeff:
Thank you @sdesrozis for your review.
if gpu tests pass, we are ready to merge.

ignite/engine/__init__.py
```diff
@@ -41,6 +42,236 @@ def _prepare_batch(
...
+) -> Callable:
+    """Factory function for supervised training using ``torch.cuda.amp``.
+
+    Args:
+        model (torch.nn.Module): the model to train.
+        optimizer (torch.optim.Optimizer): the optimizer to use.
+        loss_fn (torch.nn loss function): the loss function to use.
+        device (str, optional): device type specification (default: None).
+            Applies to batches after starting the engine. Model *will not* be moved.
+            Device can be CPU, GPU or TPU.
```
vfdev-5, line 107:
```suggestion
            Device can be CPU, GPU.
```
Can't be TPU right ?

ignite/engine/__init__.py
```diff
@@ -41,6 +42,236 @@ def _prepare_batch(
...
+) -> Callable:
+    """Factory function for supervised training using apex.
+
+    Args:
+        model (torch.nn.Module): the model to train.
+        optimizer (torch.optim.Optimizer): the optimizer to use.
+        loss_fn (torch.nn loss function): the loss function to use.
+        device (str, optional): device type specification (default: None).
+            Applies to batches after starting the engine. Model *will not* be moved.
+            Device can be CPU, GPU or TPU.
```
vfdev-5, line 163:
```suggestion
            Device can be CPU, GPU.
```

vfdev-5 (approved):
Looks good to me as well ! Thanks a lot @ydcjeff !
I left few nit comments about removing TPU mentions where it is inappropriate.
The comment about the warning and adding usage examples of these features could be done in a follow-up PR...

sdesrozis:
Could we add tests to decrease codecov warnings ?

vfdev-5:
> Could we add tests to decrease codecov warnings ?

Let's do that all in a follow-up PR :)

ignite/engine/__init__.py
```diff
@@ -41,6 +42,236 @@ def _prepare_batch(
...
+    """Checking tpu, amp and GradScaler instance combinations."""
+    if on_tpu and not idist.has_xla_support:
+        raise RuntimeError("In order to run on TPU, please install PyTorch XLA")
+
+    if amp_mode and on_tpu:
+        raise ValueError("amp_mode cannot be used with xla device. Consider using amp_mode=None or device='cuda'.")
+
+    if scaler:
+        if amp_mode != "amp":
+            warn(
```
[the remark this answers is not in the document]
ydcjeff: I also thought about raising an error here. But, the training can still run even if there is scaler argument and not amp mode.
What do you think? Should we raise an error about invalid combination of `scaler` and `amp_mode`?

ydcjeff:
> Could we add tests to decrease codecov warnings ?

I think those warnings are from `one_gpu_tests` failing to upload coverage.

>  The comment about the warning and adding usage examples of these features could be done in a follow-up PR...

will do that.

ydcjeff:
Found out that `amp` module is available in torch 1.5 which doesn't have `autocast` yet.
Changed to `ImportError` to handle all torch version.

ignite/engine/__init__.py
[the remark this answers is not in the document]
sdesrozis: Let's raise an error. This combination could work but there is no guarantee regarding to the next versions of PyTorch.

sdesrozis:
> I think those warnings are from `one_gpu_tests` failing to upload coverage.

Do we know why it does not work ?

ydcjeff:
> > I think those warnings are from `one_gpu_tests` failing to upload coverage.
> 
> Do we know why it does not work ?

I don't know very well, but it can be codecov failed to upload to its server.

ignite/engine/__init__.py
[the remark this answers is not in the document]
ydcjeff: Added in https://github.com/pytorch/ignite/pull/1589/commits/e4681d8acd89c9b483f3c14090b7ab84e59a1933

vfdev-5:
> > I think those warnings are from `one_gpu_tests` failing to upload coverage.
> 
> Do we know why it does not work ?

Asked here: https://github.com/codecov/codecov-bash/issues/411

Anyway, if there is no way to fix the uploading we can remove `- Z` option and silently ignore uploading issue.

vfdev-5:
@sdesrozis can you please merge this PR once the ci is OK.

ydcjeff:
Thank you for your help and reviews.

PR #1596, opened 2021-01-31
[metrics] add math equations in ignite.metrics

### LINKED ISSUE #1577: [docs] add equations in `ignite.metrics`
## 🚀 Feature
Add equations in `ignite.metrics` like PyTorch does on loss functions and also we can revisit the equation later if needed :]

ydcjeff:
Fixes #1577 

Description: add math equations in `ignite.metrics`

Check list:

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -193,7 +193,10 @@ texinfo_documents = [
 # -- Options for intersphinx extension ---------------------------------------
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {"https://docs.python.org/3/": None}
+intersphinx_mapping = {
+    "python": ("https://docs.python.org/3", None),
+    "torch": ("https://pytorch.org/docs/stable/", None),
+}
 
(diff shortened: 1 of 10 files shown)


docs/source/conf.py
```diff
@@ -193,7 +193,10 @@
 # -- Options for intersphinx extension ---------------------------------------
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {"https://docs.python.org/3/": None}
+intersphinx_mapping = {
+    "python": ("https://docs.python.org/3", None),
```
vfdev-5, line 197: I think I've already seen such change in another PR :)

vfdev-5 (commented):
Thanks for the PR @ydcjeff !

vfdev-5 (approved):

docs/source/conf.py
ydcjeff: Yes, this is needed to render properly. I made the change exactly the same, so there won't be conflicts.

PR #1597, opened 2021-02-01
1586 Add greater_or_equal option to Checkpoint handler

### LINKED ISSUE #1586: Support options to compare in CheckpointHander
## 🚀 Feature

Hi @vfdev-5 ,

In current ignite CheckpointHander, it compares `priority`(may be metrics) > `previous priority` and save models:
https://github.com/pytorch/ignite/blob/master/ignite/handlers/checkpoint.py#L357
But we received several requests to customize this compare logic, is it possible to make options for it?

Thanks.

Nic-Ma:
Signed-off-by: Nic Ma <nma@nvidia.com>

Fixes #1586

Description:
This PR added an option `greater_or_equal` to the Checkpoint handler, whether to save checkpoint if new priority equals to _saved[0].

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -343,6 +343,12 @@ class Checkpoint(Serializable):
             return True
         return len(self._saved) < self.n_saved + int(or_equal)
 
+    def _compare_fn(self, new: Union[int, float]) -> bool:
+        if self.greater_or_equal:
+            return new >= self._saved[0].priority
+        else:
+            return new > self._saved[0].priority
+
     def __call__(self, engine: Engine) -> None:
 
         global_step = None
@@ -358,13 +364,7 @@ class Checkpoint(Serializable):
                 global_step = engine.state.get_event_attrib_value(Events.ITERATION_COMPLETED)
             priority = global_step
 
-        def _compare_fn(old: Union[int, float], new: Union[int, float]) -> bool:
-            if self.greater_or_equal:
-                return new >= old
-            else:
-                return new > old
-
-        if self._check_lt_n_saved() or _compare_fn(self._saved[0].priority, priority):
+        if self._check_lt_n_saved() or self._compare_fn(priority):
 
             priority_str = f"{priority}" if isinstance(priority, numbers.Integral) else f"{priority:.4f}"
 
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -98,8 +98,8 @@ class Checkpoint(Serializable):
             details.
         include_self (bool): Whether to include the `state_dict` of this object in the checkpoint. If `True`, then
             there must not be another object in ``to_save`` with key ``checkpointer``.
-        greater_or_equal (bool): when comparing priority, whether to save if new priority equals to _saved[0],
-            default to `False` to only save when new priority is greater.
+        greater_or_equal (bool): if `True`, the latest equally scored model is stored. Otherwise, the first model.
+            Default, `False`.
 
     .. _DistributedDataParallel: https://pytorch.org/docs/stable/generated/
         torch.nn.parallel.DistributedDataParallel.html
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -358,7 +358,7 @@ class Checkpoint(Serializable):
                 global_step = engine.state.get_event_attrib_value(Events.ITERATION_COMPLETED)
             priority = global_step
 
-        def _compare_fn(old, new):
+        def _compare_fn(old: Union[int, float], new: Union[int, float]) -> bool:
             if self.greater_or_equal:
                 return new >= old
             else:
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -98,6 +98,8 @@ class Checkpoint(Serializable):
             details.
         include_self (bool): Whether to include the `state_dict` of this object in the checkpoint. If `True`, then
             there must not be another object in ``to_save`` with key ``checkpointer``.
+        greater_or_equal (bool): when comparing priority, whether to save if new priority equals to _saved[0],
+            default to `False` to only save when new priority is greater.
 
     .. _DistributedDataParallel: https://pytorch.org/docs/stable/generated/
         torch.nn.parallel.DistributedDataParallel.html
@@ -261,6 +263,7 @@ class Checkpoint(Serializable):
         global_step_transform: Optional[Callable] = None,
         filename_pattern: Optional[str] = None,
         include_self: bool = False,
+        greater_or_equal: bool = False,
     ) -> None:
 
         if to_save is not None:  # for compatibility with ModelCheckpoint
@@ -301,6 +304,7 @@ class Checkpoint(Serializable):
         self.filename_pattern = filename_pattern
         self._saved = []  # type: List["Checkpoint.Item"]
         self.include_self = include_self
+        self.greater_or_equal = greater_or_equal
 
     def reset(self) -> None:
         """Method to reset saved checkpoint names.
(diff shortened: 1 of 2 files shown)


vfdev-5:
Hi @Nic-Ma , thanks a lot for the PR ! 
I see your point about score_function and this option : https://github.com/pytorch/ignite/issues/1586#issuecomment-770737387

Seems like `greater_or_equal` will be only used with `score_function`, right ? As if we use Checkpoint that defines its priority by the iteration, than `greater_or_equal=False` wont be helpful...

Nic-Ma:
Hi @vfdev-5 ,

Yes, I didn't change the default behavior, just added an option for the case that `score` equals to the previous `_save[0]` and you want to save the latest checkpoint even with the same score.
If using `iteration` as the priority, it will not have the case that priority equals to the previous one, right?

Thanks.

vfdev-5:
@Nic-Ma I think I misunderstood the PR. 

> Yes, I didn't change the default behavior, just added an option for the case that score equals to the previous _save[0] and you want to save the latest checkpoint even with the same score.
If using iteration as the priority, it will not have the case that priority equals to the previous one, right?

I think it was asked previously and we added that such that latest equally scored model is stored... Let me check that

vfdev-5:
OK, catched up, what we did was to be able to save the latest model with the same filename.

ignite/handlers/checkpoint.py
```diff
@@ -98,6 +98,8 @@ class Checkpoint(Serializable):
             details.
         include_self (bool): Whether to include the `state_dict` of this object in the checkpoint. If `True`, then
             there must not be another object in ``to_save`` with key ``checkpointer``.
+        greater_or_equal (bool): when comparing priority, whether to save if new priority equals to _saved[0],
+            default to `False` to only save when new priority is greater.
```
vfdev-5, line 102:
```suggestion
        greater_or_equal (bool): if True, the latest equally scored model is stored. Otherwise, the first model.
            Default, False.
```

ignite/handlers/checkpoint.py
```diff
@@ -354,7 +358,13 @@ def __call__(self, engine: Engine) -> None:
                 global_step = engine.state.get_event_attrib_value(Events.ITERATION_COMPLETED)
             priority = global_step
 
-        if self._check_lt_n_saved() or self._saved[0].priority < priority:
+        def _compare_fn(old, new) -> bool:
```
vfdev-5, line 361:
```suggestion
        @staticmethod
        def _compare_fn(old: float, new: float) -> bool:
```

ignite/handlers/checkpoint.py
```diff
@@ -354,7 +358,13 @@ def __call__(self, engine: Engine) -> None:
...
+            if self.greater_or_equal:
+                return new >= old
+            else:
+                return new > old
+
+        if self._check_lt_n_saved() or _compare_fn(self._saved[0].priority, priority):
```
vfdev-5, line 367:
```suggestion
        if self._check_lt_n_saved() or self._compare_fn(self._saved[0].priority, priority):
```

Nic-Ma:
Thanks for your confirmation, and this is a really useful feature, especially in FL, for example:
During FL training, we keep the training run for a long time and add more and more data to train the model day by day, the later model should be better even has the same metrics.

Thanks.

vfdev-5 (commented):
Just some comments to fix the implementation

vfdev-5:
Thanks for the explanation! Yes, true that this makes sense. I wonder if we should not set that as a default behaviour now ? And for users who'd like to have BC, they could use `greater_or_equal=False`...

Nic-Ma:
> Thanks for the explanation! Yes, true that this makes a lot of sense. I wonder if we should not set that as a default behaviour now ? And for users who'd like to have BC, they could use `greater_or_equal=False`...

But I think for regular training on a fixed dataset, maybe the earlier model with the same metrics is better, because the later model may be overfitting? That's why we usually use `early-stopping`?

Thanks.

vfdev-5:
yes, it could also prevent saving overfitted model. Let's keep it False by default, I agree.

ignite/handlers/checkpoint.py
Nic-Ma: Updated in the latest commit.

ignite/handlers/checkpoint.py
Nic-Ma: May I know why you want to change this internal function to `static function`? And as it uses `self.greater_or_equal`, users can't call it directly.

Thanks.

ignite/handlers/checkpoint.py
```diff
@@ -354,7 +358,13 @@ def __call__(self, engine: Engine) -> None:
                 global_step = engine.state.get_event_attrib_value(Events.ITERATION_COMPLETED)
             priority = global_step
 
-        if self._check_lt_n_saved() or self._saved[0].priority < priority:
+        def _compare_fn(old: Union[int, float], new: Union[int, float]) -> bool:
```
vfdev-5, line 361:
```suggestion
        def _compare_fn(self, old: Union[int, float], new: Union[int, float]) -> bool:
```

ignite/handlers/checkpoint.py
```diff
@@ -354,7 +358,13 @@ def __call__(self, engine: Engine) -> None:
...
+            if self.greater_or_equal:
+                return new >= old
+            else:
+                return new > old
```
vfdev-5, line 365:
```suggestion
            if self.greater_or_equal:
                return new >= self._saved[0].priority
            else:
                return new > self._saved[0].priority
```

ignite/handlers/checkpoint.py
Nic-Ma: I see your point.
Updated in the latest commit.

Thanks.

vfdev-5: Thanks! At first I was thinking that the method could be static, then realized that it has `self.greator_or_equal`, so I think it could be interesting to fetch `self._saved[0].priority` inside this function...

ignite/handlers/checkpoint.py
Nic-Ma: updated.
Thanks.

vfdev-5 (approved):
LGTM! Thanks a lot @Nic-Ma !

Nic-Ma:
I can't see the CI errors, could you please help me figure it out?

Thanks.

vfdev-5:
> I can't see the CI errors, could you please help me figure it out?
> 
> Thanks.

@Nic-Ma it is not an error, I just updated the PR to the latest master and thus cancelled Circle CI tests, but Github interpret this as a failure with a red cross.

vfdev-5:
Oh, I forgot about adding [`versionadd`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded) in the docs.
In the end of the docs, we can add 
```
.. versionadded:: 0.4.3
    Added ``greater_or_equal`` parameter.
```

@Nic-Ma could you please send a follow-up PR with that ?

ydcjeff:
Most project uses `versionadd` for new class, funtion, etc and uses `versionchanged` for adding new args / behaviours and bug fixes.
Example: https://docs.python.org/3/library/os.html#os.environb

vfdev-5:
@ydcjeff thanks for the details ! In the provided link, actually I couldn't find any new args added, but I think it may seem reasonable to use `versionchanged` for that.

Nic-Ma:
Hi @vfdev-5 @ydcjeff ,

Sure, I submitted PR https://github.com/pytorch/ignite/pull/1600 and changed to `versionchanged`.
Could you please help review it?

Thanks.

ydcjeff:
@vfdev-5 Yea, the example may be belongs to changing of internal behaviour.

Nic-Ma:
Hi @vfdev-5 ,

This is a missing feature for our project, could you please help add this MR to your 0.4.4 bug fix release?

Thanks.

vfdev-5:
Hi @Nic-Ma, sure !

vfdev-5:
Hi @Nic-Ma , actually, just checked but this PR is already present in v0.4.3: https://pytorch.org/ignite/handlers.html#ignite.handlers.Checkpoint

Nic-Ma:
Cool, thanks!!

PR #1600, opened 2021-02-01
1586 update docs of Checkpoint handler

Nic-Ma:
Fixes #{1586} 

Description:
This PR updated the docs of Checkpoint handler.

Check list:
* [ ] New tests are added (if a new feature is added)
* [x] New doc strings: description and/or example code are in RST format
* [ ] Documentation is updated (if required)

PR #1601, opened 2021-02-01
Use standard stream for loggers

sdesrozis:
Description:

By default, loggers stream on `sys.stderr`. This PR allows to define the stream and the default stream is now `sys.stdout`. This is useful when `ignite` is used with workload manager like `slurm`.

EDIT : replace default `sys.stderr` by `std.stdout` does not work in tests (check_config idist). It could be an issue capsys / pytest

EDIT 2 : It seems that `https://pypi.org/project/pytest-catchlog/` provides fixture that helps to catch stream from logs.

Check list:

### DIFF
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -22,7 +22,7 @@ jobs:
   code-style:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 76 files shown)


vfdev-5 (approved):
Looks good, thanks @sdesrozis !

vfdev-5:
@sdesrozis please, add a mention here about added arg, like here : https://github.com/pytorch/ignite/pull/1600/files

vfdev-5 (commented):
Thanks Sylvain!

PR #1598, opened 2021-02-01
Reduced the list of ignored errors by flake8

vfdev-5:
Description:
- Reduced the list of ignored errors by flake8
- fixed errors

Check list:
* [ ] New tests are added (if a new feature is added)
* [ ] New doc strings: description and/or example code are in RST format
* [ ] Documentation is updated (if required)

PR #1602, opened 2021-02-01
Removed ignored errors and warnings

vfdev-5:
Copy of failing https://github.com/pytorch/ignite/pull/1598

Check list:

### DIFF
--- a/examples/contrib/cifar10/main.py
+++ b/examples/contrib/cifar10/main.py
@@ -1,23 +1,20 @@
-from pathlib import Path
 from datetime import datetime
-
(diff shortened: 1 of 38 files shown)


trsvchn:
@vfdev-5 What is happening?) why do we have two active workflows? and one is failing

vfdev-5:
> @vfdev-5 What is happening?) why do we have two active workflows? and one is failing

I have no idea :) Well, it is the same commit I pushed here (my fork) and in the branch of pytorch. Probably, that's why.

Issue #1609, opened 2021-02-04
Support for MultiLabel MultiClass classification confusion matrix computation.


touqir14:
## ❓ Questions/Help/Support

Does Ignite support multilabel multiclass classification confusion matrix computation? I was looking for something analogous to this tensorflow feature : https://www.tensorflow.org/addons/api_docs/python/tfa/metrics/MultiLabelConfusionMatrix .

vfdev-5:
Thanks for asking! Currently, we do not provide such metric but it would make sense to add a feature request for this metric. Let's see if we could implement it...

touqir14:
I once sent a PR to Pytorch's tnt project implementing this functionality : https://github.com/pytorch/tnt/pull/137 . Tnt's last stable release was back in 2019, so I am guessing they won't be releasing any new versions. Will a similar implementation work for ignite?

vfdev-5:
Sounds great ! Yes, I think we could adapt this code to ignite's codebase keeping in mind the differences in the API and 
- computations using torch instead of numpy
- adaptation for distributed computation

Anyway, please send a draft PR and we can iterate over to finalize it. 
To get started, please check:
- [CONTRIBUTING](https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md)
- [How metrics work in Ignite](https://pytorch.org/ignite/metrics.html#how-to-create-a-custom-metric)

touqir14:
Sure, I will send a draft PR within a day or two.

PR #1613, opened 2021-02-04
Adding the MultiLabelConfusionMatrix for computing multi label confusion matrices.

touqir14:
Fixes [1609](https://github.com/pytorch/ignite/issues/1609) 

Description:
This is an in-progress PR that adds functionalities for computing confusion matrices for multi-label multi-class classification problems. I will add tests and doc strings once I get an initial feedback to accomodate any necessary changes (if any). The [test_confusion_matrix.py](https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/test_confusion_matrix.py) file tests the ConfusionMatrix class quite thoroughly. I was wondering how thorough the tests for MultiLabelConfusionMatrix needs to be.

The current version only works with prediction and ground truth tensors of shape ```[batch_size, num_classes]```. The tensor values need to be binary. The example below illustrates its use.
```python
import torch
from ignite.metrics import MultiLabelConfusionMatrix

'''
Both inputs below must be a tensor of shape N x K, where N is the number of samples 
and K is the number of classes. The values of the tensors must be binary where a 
1 in the i'th row and j'th column marks the j'th label for the i'th sample.
'''

predicted = torch.Tensor([[1, 1, 0], [0, 1, 0], [1, 0, 1], [1, 1, 1]])
ground_truth = torch.Tensor([[0, 1, 0], [0, 0, 1], [1, 0, 1], [1, 0, 1]])
mtr_ignite = MultiLabelConfusionMatrix(3, normalized=False)
mtr_ignite.update([predicted, ground_truth])
conf_matrix = mtr_ignite.compute()


'''
conf_matrix[i, 0, 0] corresponds to count/rate of true negatives of class i,
conf_matrix[i, 0, 1] corresponds to count/rate of false positives of class i,
conf_matrix[i, 1, 0] corresponds to count/rate of false negatives of class i,
conf_matrix[i, 1, 1] corresponds to count/rate of true positives of class i.

With normalization : meter.MultiLabelConfusionMeter(k=3, normalized=True), for all i:
conf_matrix[i, 0, 0] + conf_matrix[i, 0, 1] + conf_matrix[i, 1, 0] + conf_matrix[i, 1, 1] = 1
'''
```
ccing @vfdev-5 

Check list:


vfdev-5:
@touqir14 thanks a lot for a quick and nice PR! Overall code looks good, later I can try to check the implementation vs tfa. Maybe, we could add a comment with credits to the original implementation in the code if it is just ported tfa implementation. I also wonder for the tests if we could compare the results with something like sklearn ?

touqir14:
This was not ported from tfa but since you mentioned it, a simple transpose operation will ensure that its output matches that of tfa's implementation. I think it is reasonable to test against scikit learn's implementation.

touqir14:
Btw, should we aim to keep the output consistent with tfa or just keep this as it is, @vfdev-5 ?

vfdev-5:
Let's keep it consistent to sklearn and mention about the difference vs tfa and what to do to make it consistent with tfa...

touqir14:
Added some tests, let me know what you think.

tests/ignite/metrics/test_confusion_matrix.py
```diff
@@ -647,6 +654,132 @@ def _test_distrib_accumulator_device(device):
...
+    output = torch.Tensor([[1, 1, 0], [0, 1, 0], [1, 0, 1], [1, 1, 1]])
+    target = torch.Tensor([[0, 1, 0], [0, 0, 1], [1, 0, 1], [1, 0, 1]])
+
+    mtr = MultiLabelConfusionMatrix(3, normalized=False)
+    mtr.update([output, target])
+
+    conf_mtrx = mtr.compute()
+    correct_conf_mtrx = torch.tensor([[[1, 1], [0, 2]], [[1, 2], [0, 1]], [[1, 0], [1, 2]]])
+
+    TestCase.assertTrue(
```
vfdev-5, line 668: in general we use pytest and do simply `assert something, err message`, so let's in the same way here.

About the tests, I'm thinking if it won't be interesting to put them into a separate file : `test_multilabel_confusion_matrix.py`. Maybe, same for the code of `MultiLabelConfusionMatrix`.

In the tests, you can simply generate inputs, compute the result and compare it vs sklearn result as it is done for ConfusionMatrix.
Do you think that it would make sense to ensure that MultiLabelConfusionMatrix works and then test it on the input of shape like `(batch_size, classes, dim1, ...)` for predictions/targets ?

Another point to address a bit later is to ensure that it works with DDP. In this case, we have to replicate what is done here : https://github.com/pytorch/ignite/blob/f1cc9fbb335273f4303d6a68fc7ddbb2d0610b87/tests/ignite/metrics/test_confusion_matrix.py#L783-L790

touqir14: Regarding the input format of ```(batch_size, classes, dim1, ...) ``` , is there any special use case? The tfa implementation and sklearn's function take 2d matrices.

vfdev-5: We can imagine, for example, image segmentation with overlapping classes...

touqir14: I see. So we can think of this as each class having ```num_batches x dim_1 x ... x dim_n``` configurations in the general case where in the 2d case, each class had ```num_batches``` number of configurations to compare between the predictions and the ground truth.

vfdev-5: Yes, it can be seen like that. In ConfusionMatrix implementation we however do in some sense this reshape internally and can accept inputs like `(B, C, H, W)`.

touqir14: So in the test cases we can simply compare the output of ignite's implementation with sklearn's implementation having feeding the latter with predictions and ground truth arrays after being reshaped as ```[num_batches x dim_1 x ... x dim_n , num_classes]```

vfdev-5: Yes, exactly ! Transposed and reshape

touqir14:
@vfdev-5 , I have been a little busy, so this commit got a bit delayed. Let me know what you think.

vfdev-5:
@touqir14 thanks for the updates and no worries about the delay ! I'll take a look later the code and comment out.

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,112 @@
...
+from ignite.exceptions import NotComputableError
+from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
+from ignite.metrics.metrics_lambda import MetricsLambda
+
+__all__ = [
+    "MultiLabelConfusionMatrix",
+]
+
+
+class MultiLabelConfusionMatrix(Metric):
```
vfdev-5, line 15: Let's add a docstring here. You can try to copy and adapt the one from ConfusionMatrix.

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,112 @@
...
+        y_reshaped = y.swapaxes(0, 1).reshape(self.num_classes, -1)
+        y_pred_reshaped = y_pred.swapaxes(0, 1).reshape(self.num_classes, -1)
+        y_total = torch.unsqueeze(torch.count_nonzero(y_reshaped, axis=1), 0)
+        y_pred_total = torch.unsqueeze(torch.count_nonzero(y_pred_reshaped, axis=1), 0)
+        tp = torch.unsqueeze(torch.count_nonzero(torch.mul(y_reshaped, y_pred_reshaped), axis=1), 0)
+        fp = y_pred_total - tp
+        fn = y_total - tp
+        tn = y_reshaped.shape[1] - tp - fp - fn
+
+        self.confusion_matrix += torch.cat([tn, fp, fn, tp], dim=0).T.reshape(-1, 2, 2)
```
vfdev-5, line 51: Can we directly create the tensor in the good shape and device (`self._device`) ? For the shape I'm not yet sure, but device will be necessary.

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @sync_all_reduce("confusion_matrix", "_num_examples")
+    def compute(self) -> torch.Tensor:
+        if self._num_examples == 0:
+            raise NotComputableError("Confusion matrix must have at least one example before it can be computed.")
+
+        if self.normalized:
+            conf = self.confusion_matrix.type(torch.float64)
+            sums = conf.sum(axis=2).sum(axis=1)
+            return (conf.reshape(conf.shape[0], 4) / sums[:, None]).reshape(conf.shape)
```
vfdev-5, line 61: Shouldn't we add eps=1e-15 to sums if it is zero or we can admit `inf` ?

tests/ignite/metrics/test_confusion_matrix.py
```diff
@@ -1,4 +1,5 @@
 import os
+from unittest import TestCase
```
vfdev-5, line 2: Please, remove this import

tests/ignite/metrics/test_multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,438 @@
+import os
+from unittest import TestCase
```
vfdev-5, line 2: Same here, please

vfdev-5 (commented):
Thanks for the update @touqir14 ! 
I haven't yet seen the tests in details. There are some comments to address for the PR and seems like that code-formatting is not passing neither...

ignite/metrics/multilabel_confusion_matrix.py
touqir14: You mean running the concatenation operation on self._device?

vfdev-5: There are two points:
- can we do `torch.cat([tn, fp, fn, tp], dim=0).T.reshape(-1, 2, 2)` in less ops ?
- I think there is a device mismatch if `self.confusion_matrix` is on CPU and tp, tn, ... are on GPU.

ignite/metrics/multilabel_confusion_matrix.py
touqir14: By definition, all the entries of ```sums``` will be equal to the number of configurations for each class. For instance in the case of 2D ```y_pred``` and ```y```, the entries of ```sums``` will all be equal to the batch size.

ignite/metrics/multilabel_confusion_matrix.py
touqir14: First point : We should not be able to do it any faster. I think we are good here since this should not be slower than: 
```
tp = torch.unsqueeze(torch.count_nonzero(torch.mul(y_reshaped, y_pred_reshaped), axis=1), 0)
fp = y_pred_total - tp
fn = y_total - tp
tn = y_reshaped.shape[1] - tp - fp - fn
```
Second point : In case the original ```y``` and ```y_pred``` are not on the same device as ```self._device```, should we use ```.to(self._device)``` on both of them? That should ensure no device mismatch.

vfdev-5: Second point: the idea of `self._device` is to have the option where to store internal things: CPU or GPU. If user provides y_preds and y on X device, let's perform all the compurations on this X device and only the final op will convert the result to the required `self._device` (to avoid device mismatch).

First point: sounds good. We can come back to this point later if there are any other ideas on how to improve the performances.

touqir14:
@vfdev-5 , feel free to review the tests. I can push updates addressing your current suggestions and any future suggestions together.

vfdev-5:
> @vfdev-5 , feel free to review the tests. I can push updates addressing your current suggestions and any future suggestions together.

@touqir14 let me do that in the coming days. Thanks for pinging

ignite/metrics/multilabel_confusion_matrix.py
vfdev-5, line 51: The code can be simplified a bit more like that:
```python
y_reshaped = y.transpose(0, 1).reshape(num_classes, -1)
y_pred_reshaped = y_pred.transpose(0, 1).reshape(num_classes, -1)

y_total = y_reshaped.sum(dim=1)
y_pred_total = y_pred_reshaped.sum(dim=1)

tp = (y_reshaped * y_pred_reshaped).sum(dim=1)
fp = y_pred_total - tp
fn = y_total - tp
tn = y_reshaped.shape[1] - tp - fp - fn

self.confusion_matrix += torch.stack([tn, fp, fn, tp], dim=1).reshape(-1, 2, 2).to(self._device)
```

- Seems like `sum` is also faster then `count_nonzero`.
- `axis` arg may not be supported by older versions -> let's use `dim`

Maybe, we could also find a single op for `tp = (y_reshaped * y_pred_reshaped).sum(dim=1)`, but haven't yet found...

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,112 @@
...
+
+        self.confusion_matrix += torch.cat([tn, fp, fn, tp], dim=0).T.reshape(-1, 2, 2)
+
+    @sync_all_reduce("confusion_matrix", "_num_examples")
+    def compute(self) -> torch.Tensor:
+        if self._num_examples == 0:
+            raise NotComputableError("Confusion matrix must have at least one example before it can be computed.")
+
+        if self.normalized:
+            conf = self.confusion_matrix.type(torch.float64)
```
vfdev-5, line 59:
```suggestion
            conf = self.confusion_matrix.to(dtype=torch.float64)
```

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,112 @@
...
+            sums = conf.sum(axis=2).sum(axis=1)
```
vfdev-5, line 60:
```suggestion
            sums = conf.sum(dim=(1, 2))
```

ignite/metrics/multilabel_confusion_matrix.py
vfdev-5, line 61:
```suggestion
            return conf / sums[:, None, None]
```

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,112 @@
...
+            return (conf.reshape(conf.shape[0], 4) / sums[:, None]).reshape(conf.shape)
+        else:
+            return self.confusion_matrix
```
vfdev-5, line 63:
```suggestion        
        return self.confusion_matrix
```

tests/ignite/metrics/test_multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,438 @@
...
+    for _ in range(num_iters):
+        target = np.random.randint(0, 2, size=(num_samples, num_classes))
+        prediction = np.random.randint(0, 2, size=(num_samples, num_classes))
+        sklearn_CM = multilabel_confusion_matrix(target, prediction)
+        mlcm = MultiLabelConfusionMatrix(num_classes, normalized=False)
+        mlcm.update([torch.tensor(prediction), torch.tensor(target)])
+        ignite_CM = mlcm.compute().numpy()
+        assert np.all(sklearn_CM.astype(np.int64) == ignite_CM.astype(np.int64))
+
+    return
```
vfdev-5, line 235:
```suggestion
```

tests/ignite/metrics/test_multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,438 @@
...
+
+    assert np.all(ignite_CM == sklearn_CM)
+
+
+def _test_distrib_multiclass_images(device):
+    def _test(metric_device):
+        num_classes = 3
+        cm = MultiLabelConfusionMatrix(num_classes=num_classes, device=metric_device)
+
+        y_true, y_pred = get_y_true_y_pred()
```
vfdev-5, line 150: We have to generate data a bit differently depending on distributed rank.

tests/ignite/metrics/test_multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,438 @@
...
+
+def test_simple_ND_input():
+
+    # 3D tests
+    num_iters = 100
+    num_samples = 100
+    num_classes = 10
+    size_3d = 4
+
+    for _ in range(num_iters):  # 3D tests
```
vfdev-5, line 246: Why we would like to run 100 (num_iters) the same (random data) test ? 2-3 times may be enough ?

tests/ignite/metrics/test_multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,438 @@
...
+        target_reshaped = target.transpose(0, 2, 3, 4, 1).reshape(
+            size_3d * size_4d * size_5d * num_samples, num_classes
+        )
+        prediction_reshaped = prediction.transpose(0, 2, 3, 4, 1).reshape(
+            size_3d * size_4d * size_5d * num_samples, num_classes
+        )
+        sklearn_CM = multilabel_confusion_matrix(target_reshaped, prediction_reshaped)
+        assert np.all(sklearn_CM.astype(np.int64) == ignite_CM.astype(np.int64))
+
+    return
```
vfdev-5, line 285:
```suggestion
```

tests/ignite/metrics/test_multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,438 @@
...
+def test_simple_batched():
+
+    num_iters = 100
+    num_samples = 100
+    num_classes = 10
+    batch_size = 1
+
+    for _ in range(num_iters):
+        mlcm = MultiLabelConfusionMatrix(num_classes, normalized=False)
+        target = np.random.randint(0, 2, size=(1, num_classes))
```
vfdev-5, line 297: Nit: maybe, we can simply generate n batches and iterate over that data instead of concat:
```python
targets = np.random.randint(0, 2, size=(num_iters, batch_size, num_classes))
predictions = np.random.randint(0, 2, size=(num_iters, batch_size, num_classes))
for target, prediction in zip(targets, predictions):
    ...
```

vfdev-5 (commented):
@touqir14 thanks again for the PR ! 
I have few comments on the implementation and tests. Let me know what do you think.

ignite/metrics/confusion_matrix.py
```diff
@@ -7,7 +7,15 @@
 from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
 from ignite.metrics.metrics_lambda import MetricsLambda
 
-__all__ = ["ConfusionMatrix", "mIoU", "IoU", "DiceCoefficient", "cmAccuracy", "cmPrecision", "cmRecall"]
+__all__ = [
```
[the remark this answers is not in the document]
touqir14: I think this was caused by black. I ran black, flake, isort on those files that I touched upon.

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,112 @@
...
+        if (
+            not isinstance(output, Sequence)
+            or len(output) < 2
+            or not isinstance(output[0], torch.Tensor)
+            or not isinstance(output[1], torch.Tensor)
+        ):
+            raise ValueError(
+                (r"Argument must consist of a Python Sequence of two tensors such that the first is the predicted"
+                 r" tensor and the second is the ground-truth tensor")
+            )
```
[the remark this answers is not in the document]
touqir14: Or maybe, we could include this check in other metrics as we see fit? Either one is fine by me.

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,112 @@
...
+            raise ValueError("y and y_pred shapes must match.")
+
+        valid_types = (torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64)
+        if y_pred.dtype not in valid_types:
+            raise ValueError(f"y_pred must be of any type: {valid_types}")
+
+        if y.dtype not in valid_types:
+            raise ValueError(f"y must be of any type: {valid_types}")
+
+        if y_pred.numel() != ((y_pred == 0).sum() + (y_pred == 1).sum()).item():
```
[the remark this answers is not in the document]
touqir14: Thats a clever optimization! I will replace that.

ignite/metrics/confusion_matrix.py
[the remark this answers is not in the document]
vfdev-5: There is additional comma in the end to remove

ignite/metrics/multilabel_confusion_matrix.py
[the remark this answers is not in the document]
vfdev-5: Let's remove it here and maybe we could add this check in Metric class in another PR

tests/ignite/metrics/test_multilabel_confusion_matrix.py
touqir14: I thought the more number of times we run the better we are off since we might be covering edge cases there. Btw, I am not setting the numpy random number generator seed here, so each run should produce different data.

vfdev-5: Well, covering edge cases in practice should be done explicitly than using a random input. That's a good point about numpy input, let's generate data with torch as we do it in other tests and we also set seed here for torch generator.

ignite/metrics/confusion_matrix.py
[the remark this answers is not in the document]
touqir14: Is the additional comma really an issue? Its fairly common to use comma's at the end of lists created statically - it gets easier to remove or add new elements statically.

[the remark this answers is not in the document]
vfdev-5: In practice if a PR introduces a feature in separate files etc it should fix other things in other files. Especially here it is really a matter of taste.

tests/ignite/metrics/test_multilabel_confusion_matrix.py
touqir14: So, instead of concatenating I am now generating matrices once in every iteration of the outer loop.

tests/ignite/metrics/test_multilabel_confusion_matrix.py
touqir14: Changed the matrix generation part from numpy to torch.

tests/ignite/metrics/test_multilabel_confusion_matrix.py
touqir14: Do you have any suggestions?

ignite/metrics/multilabel_confusion_matrix.py
touqir14: Added a docstring.

touqir14:
There seems to be a test failure in the ```test_distrib_cpu``` function. I am not sure what is causing that. In the non distributed setting the tests ran fine.

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @sync_all_reduce("confusion_matrix", "_num_examples")
+    def compute(self) -> torch.Tensor:
+        if self._num_examples == 0:
+            raise NotComputableError("Confusion matrix must have at least one example before it can be computed.")
+
+        if self.normalized:
+            conf = self.confusion_matrix.to(dtype=torch.float64)
+            sums = conf.sum(dim=(1, 2))
+            return conf / sums[:, None, None]
```
vfdev-5, line 92: Seems like there is no test for normalized CM ?

tests/ignite/metrics/test_multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,422 @@
...
+        sklearn_CM = multilabel_confusion_matrix(target.numpy(), prediction.numpy())
+        mlcm = MultiLabelConfusionMatrix(num_classes, normalized=False)
+        mlcm.update([prediction, target])
+        ignite_CM = mlcm.compute().numpy()
+        assert np.all(sklearn_CM.astype(np.int64) == ignite_CM.astype(np.int64))
+
+
+def test_simple_ND_input():
+
+    num_iters = 100
```
vfdev-5, line 231: Let's change it to 5
```suggestion
    num_iters = 5
```

tests/ignite/metrics/test_multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,422 @@
...
+        prediction_reshaped = prediction.permute(0, 2, 3, 4, 1).reshape(
+            size_3d * size_4d * size_5d * num_samples, num_classes
+        )
+        sklearn_CM = multilabel_confusion_matrix(target_reshaped.numpy(), prediction_reshaped.numpy())
+        assert np.all(sklearn_CM.astype(np.int64) == ignite_CM.astype(np.int64))
+
+
+def test_simple_batched():
+
+    num_iters = 100
```
vfdev-5, line 277: Same here
```suggestion
    num_iters = 5
```

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,132 @@
...
+    same dimensions.
+    - The confusion matrix 'M' is of dimension (num_classes, 2, 2).
+        M[i, 0, 0] corresponds to count/rate of true negatives of class i,
+        M[i, 0, 1] corresponds to count/rate of false positives of class i,
+        M[i, 1, 0] corresponds to count/rate of false negatives of class i,
+        M[i, 1, 1] corresponds to count/rate of true positives of class i.
+    - The classes present in M are indexed as 0, ..., num_classes-1 as can be inferred from above.
+
+    Args:
+        num_classes (int): Number of classes, should be > 1. See notes for more details.
```
vfdev-5, line 34:
```suggestion
        num_classes (int): Number of classes, should be > 1.
```

tests/ignite/metrics/test_multilabel_confusion_matrix.py
vfdev-5: Oh, I see, our tests for confusion matrix have the same remark. I think we should rework them. Let me propose a better tests for our confusion matrix which could be adapted for this PR as well. We can inspire from here : https://github.com/pytorch/ignite/blob/e17acc740d38909ddbbed6648423750abe593e58/tests/ignite/metrics/test_recall.py#L784-L785

ignite/metrics/multilabel_confusion_matrix.py
touqir14: I can add that in the next commit along with the new distributed settings test. Since the normalization feature is independent from the shapes of y and y_pred, I think it suffices to add similar tests as in ```test_simple_2D_input``` using ```normalized=True```

vfdev-5:
@touqir14 let's do like that for this PR: let's comment out distributed tests and fix all remaining nits asked below. And we can merge it. After that in a follow-up PR we can rework both distrib tests for CM and MLCM. What do you think ?

ignite/metrics/multilabel_confusion_matrix.py
vfdev-5: >  I think it suffices to add similar tests as in test_simple_2D_input using normalized=True

Sounds good ! We can even refactor existing test_simple_2D_input as 
```python

def test_simple_2D_input():

    def _test(normalized):
         # all previous code with adaptation for normalized

    _test(True)
    _test(False)

```

touqir14:
That sounds good. I will push a commit next addressing the remaining issues.

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,132 @@
...
+        output_transform (callable, optional): a callable that is used to transform the
+            :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
+            form expected by the metric. This can be useful if, for example, you have a multi-output model and
+            you want to compute the metric with respect to one of the outputs.
+        device (str or torch.device): specifies which device updates are accumulated on. Setting the metric's
+            device to be the same as your ``update`` arguments ensures the ``update`` method is non-blocking. By
+            default, CPU.
+
```
vfdev-5, line 42: I also almost forgot about another tag we are now adding in the docstring. Please, add
```python
    .. versionadded:: 0.5.0
```
for a tag like here: https://pytorch.org/ignite/v0.4.3/metrics.html#ignite.metrics.PSNR

touqir14:
Just pushed a commit. I think all the issues have been addressed. Let me know if I missed something

vfdev-5 (approved):
Looks good @touqir14 ! Thanks !
Just fix code formatting issue and it is good to go

vfdev-5:
Another point from the checklist is not yet done : 

We have to add an entry here as well : https://github.com/pytorch/ignite/blob/master/docs/source/metrics.rst

Could you please also merge current master into your branch as "This branch is out-of-date with the base branch". Thanks

touqir14:
For the docs part here : https://github.com/pytorch/ignite/blob/master/docs/source/metrics.rst , I should just add ```.. autoclass:: MultiLabelConfusionMatrix ``` in the complete list of metrics right?

vfdev-5:
Yes, just `.. autoclass:: MultiLabelConfusionMatrix` in alphabetical order and should be good

ignite/metrics/multilabel_confusion_matrix.py
```diff
@@ -0,0 +1,134 @@
...
+    according to the ground truth.
+    - both 'y' and 'y_pred' must be torch Tensors having any of the following types:
+    {torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64}. They must have the
+    same dimensions.
+    - The confusion matrix 'M' is of dimension (num_classes, 2, 2).
+        M[i, 0, 0] corresponds to count/rate of true negatives of class i,
+        M[i, 0, 1] corresponds to count/rate of false positives of class i,
+        M[i, 1, 0] corresponds to count/rate of false negatives of class i,
+        M[i, 1, 1] corresponds to count/rate of true positives of class i.
+    - The classes present in M are indexed as 0, ..., num_classes-1 as can be inferred from above.
```
vfdev-5, line 31: This part is not correctly rendered, see https://deploy-preview-1613--pytorch-ignite-preview.netlify.app/metrics.html#ignite.metrics.MultiLabelConfusionMatrix
The issue is with missing indent for the new line. Maybe, 2 spaces can help to fix that.

touqir14:
Reformatted the docstring, let me know how it looks.

vfdev-5:
> Reformatted the docstring, let me know how it looks.

Looks better :) https://deploy-preview-1613--pytorch-ignite-preview.netlify.app/metrics.html#ignite.metrics.MultiLabelConfusionMatrix

Description of M[i, 0, 0] values is inlined but it can go...

touqir14:
Its better to put the ```M[]``` parts in a bulleted format I think. Will adding an extra "-" before each line put them in their own bulleted list? Like:
```
- The confusion matrix 'M' is of dimension (num_classes, 2, 2).
      - M[i, 0, 0] corresponds to count/rate of true negatives of class i,
      - M[i, 0, 1] corresponds to count/rate of false positives of class i,
      - M[i, 1, 0] corresponds to count/rate of false negatives of class i,
      - M[i, 1, 1] corresponds to count/rate of true positives of class i.
```

vfdev-5:
Yes, I agree. I'm not sure if rst/sphinx would accept it simply like that, anyway It's better to try that locally :)
Here is info on how to generate docs locally: https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md#local-documentation-building-and-deploying

touqir14:
Looks good to me.

vfdev-5:
@touqir14 thanks again for the PR! If you are interested in finalize the things with distributed we can discuss about that in the issue opened for that : https://github.com/pytorch/ignite/issues/1657

touqir14:
Thanks @vfdev-5 for your valuable feedbacks! I will soon have a look at #1657

PR #1622, opened 2021-02-07
Update pre-commit hooks and CONTRIBUTING.md

### LINKED ISSUE #1618: Update CONTRIBUTING and pre-commit file
Seems like these lines should be updated in order to reflect the latest state of things:
- https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md#formatting-without-pre-commit
```diff
- isort -rc .
- black .
+ black .
+ isort .
```
and
https://github.com/pytorch/ignite/blob/29a436fac250bbec2ead8857cdd38b40ae5b8b41/.pre-commit-config.yaml#L23
should be updated to `5.7.0` release (to check the exact change).

cc @Devanshu24 I remember you are using pre-commit, could you please take a look at this issue. Thanks!

Devanshu24:
Fixes #1618  

Description:
- Change pre-commit config and CONTRIBUTING.md
  - Update hook versions
  - Remove seed-isort-config (details [here](https://pycqa.github.io/isort/docs/configuration/pre-commit/))
  - Add black profile to isort (details [here](https://pycqa.github.io/isort/docs/configuration/black_compatibility/)) 
-  Fix files based on you pre-commit config

Check list:
* [ ] New tests are added (if a new feature is added) (N.A.)
* [ ] New doc strings: description and/or example code are in RST format (N.A.)


vfdev-5:
Thanks @Devanshu24 ! Modifications of md, yml files looks reasonable, but svg images and notebooks shouldn't be modified, I think. Let's revert that.

Devanshu24:
> Thanks @Devanshu24 ! Modifications of md, yml files looks reasonable, but svg images and notebooks shouldn't be modified, I think. Let's revert that.

Okay, even I was doubtful on those. Thanks!

Devanshu24:
Done, also added `.rst` and `.gitignore`. If you want can revert them :)

Devanshu24:
Out of curiosity, why did we not update the version for `black` ?

vfdev-5:
> Out of curiosity, why did we not update the version for `black` ?

It produces more garbage in the code for me.

Devanshu24:
> > Out of curiosity, why did we not update the version for `black` ?
> 
> It produces more garbage in the code for me.

Ohh okay, makes sense! Thanks :)

.pre-commit-config.yaml
```diff
@@ -1,37 +1,32 @@
...
       - id: trailing-whitespace
 
-  - repo: https://github.com/prettier/pre-commit
-    rev: main
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.2.1
     hooks:
       - id: prettier
 
-  - repo: https://github.com/asottile/seed-isort-config
```
vfdev-5, line 16: Actually, why we removed that ?
The purpose of this was to correctly distinguish 3rd party libs etc Does it still work like that in this PR ?

README.md
```diff
@@ -7,11 +7,11 @@
...
-| ![image](https://img.shields.io/badge/-Supported_PyTorch/Python_versions:-black?style=flat-square) [![link](https://img.shields.io/badge/-check_here-blue)](https://github.com/pytorch/ignite/actions?query=workflow%3A%22PyTorch+version+tests%22) |
+| :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
```
vfdev-5, line 10: This is ugly :(

examples/references/classification/imagenet/README.md
```diff
@@ -79,9 +79,9 @@ uses distributed sampling, scales batch size etc).
 
 ### Results
 
-Model | Training Top-1 Accuracy | Training Top-5 Accuracy | Test Top-1 Accuracy | Test Top-5 Accuracy
----|---|---|---|---
-ResNet-50 | 78% | 92% | 77% | 94%
+| Model     | Training Top-1 Accuracy | Training Top-5 Accuracy | Test Top-1 Accuracy | Test Top-5 Accuracy |
```
vfdev-5, line 82: I'm not a fan of such "improvement" ...

examples/references/segmentation/pascal_voc2012/README.md
```diff
@@ -8,9 +8,9 @@ Features:
 - Distributed training with mixed precision by [nvidia/apex](https://github.com/NVIDIA/apex/)
 - Experiments tracking with [MLflow](https://mlflow.org/) or [Polyaxon](https://polyaxon.com/) or [ClearML](https://github.com/allegroai/clearml)
 
- Tensorboard | MLflow 
----|---
- ![tb_dashboard](assets/tb_dashboard.png) | ![mlflow_dashboard](assets/mlflow_dashboard.png) 
+| Tensorboard                              | MLflow                                           |
```
vfdev-5, line 11: Same here

examples/references/segmentation/pascal_voc2012/README.md
```diff
@@ -84,9 +84,9 @@ uses distributed sampling, scales batch size etc).
 
 ### Results
 
-Model | with SBD | Training mIoU+BG | Test mIoU+BG
----|---|---|---
-DeepLabV3 ResNet-101 | X | 86% | 68%
+| Model                | with SBD | Training mIoU+BG | Test mIoU+BG |
```
vfdev-5, line 87: And here

.pre-commit-config.yaml
Devanshu24: I removed it after reading [this](https://pycqa.github.io/isort/docs/configuration/pre-commit/) page in the documentation. Apparently, they have incorporated the facilities internally.
And the tests are still passing in pre-commit

README.md
Devanshu24: I agree! :D
Will change it right away

vfdev-5:
These tools produce rather opinionated changes saying that this is "nicer" or "prettier" which is really a matter of taste in some cases... That's why i'd like to keep changes minimal as possible....

examples/references/classification/imagenet/README.md
Devanshu24: Will revert it. 
I agree readability is better in the original one

Devanshu24:
Prettier won't let me change tables in the markdown files (gives an error during commit)
We could do 2 things:
- Find some older prettier version that allows it
- It is not looking so bad on full screen so maybe we could keep it (image below for reference)
Screenshot from 2021-02-07 16-08-18

Which would you prefer?

CONTRIBUTING.md
```diff
@@ -104,8 +104,8 @@ black manually to format files and commit them.
 
 ```bash
 # This should autoformat the files
-isort -rc .
 black .
+isort --profile black .
```
vfdev-5, line 108: In addition to this change, we have to modify also : 
- https://github.com/pytorch/ignite/blob/f51320a33571217818a3b429551b4bcdd97835f4/.github/workflows/code-style.yml#L33
- https://github.com/pytorch/ignite/blob/f51320a33571217818a3b429551b4bcdd97835f4/.github/workflows/unit-tests.yml#L85

vfdev-5:
> Prettier won't let me change tables in the markdown files (gives an error during commit)
> We could do 2 things:
> 
>     * Find some older prettier version that allows it
> 
>     * It is not looking so bad on full screen so maybe we could keep it (image below for reference)
>       Screenshot from 2021-02-07 16-08-18
> 
> 
> Which would you prefer?

I'm hesitating between removing this prettier and just having it optional without changing all files...

CONTRIBUTING.md
Devanshu24: Oh right, the Actions workflow. Will do that, thanks!

Devanshu24:
> > Prettier won't let me change tables in the markdown files (gives an error during commit)
> > We could do 2 things:
> > ```
> > * Find some older prettier version that allows it
> > 
> > * It is not looking so bad on full screen so maybe we could keep it (image below for reference)
> >   Screenshot from 2021-02-07 16-08-18
> > ```
> > 
> > 
> > Which would you prefer?
> 
> I'm hesitating between removing this prettier and just having it optional without changing all files...

One option may be to selectively keep it? Make it ignore markdown files, and only run on python files.
Not sure how good of an option it is, just came to my mind :)

vfdev-5:
@Devanshu24 I agree, let's configure it to run only on text files like .md, .rst, .yml and not .py, .ipynb, .sh, .gitignore. 
Then, those table formatting if we could avoid at least the one in README.md with a enormous line of dashes, would be great...

Devanshu24:
> @Devanshu24 I agree, let's configure it to run only on text files like .md, .rst, .yml and not .py, .ipynb, .sh, .gitignore.
> Then, OK for those table formatting if we could avoid at least the one in README.md with a enormous line of dashes, would be great...

Sure, I think that'll be doable!
I'll get to it :)

Devanshu24:
Made the changes, please have a look at your convenience :)

Devanshu24:
Made the changes, please have a look at your convenience :)

vfdev-5 (approved):
Thanks @Devanshu24 !

PR #1634, opened 2021-02-12
Basic Rouge Score Implementation

gucifer:
Added a simple Implementation for the ROUGE Metric.

Description:
This is a rather simple implementation and I do not think it supports DDP just yet.
Tagging @vfdev-5, can we discuss on how to add DDP support?
Note: Comments Not added, what is the mathematical notation used?
Check list:

### DIFF
--- a/ignite/metrics/rouge_n.py
+++ b/ignite/metrics/rouge_n.py
@@ -1,0 +1,85 @@
+from ignite.metrics import Metric
+from ignite.exceptions import NotComputableError
+from collections import defaultdict
+import torch
+
+# These decorators helps with distributed settings
+from ignite.metrics.metric import sync_all_reduce, reinit__is_reduced
+
+
+class Rouge(Metric):
+
+    def __init__(self, alpha, n, output_transform=lambda x: x, device="cpu"):
+        self._rougelist = None
+        self._num_examples = 0
+        self.alpha = alpha
+        self.n = n
+        super(Rouge, self).__init__(output_transform=output_transform, device=device)
+
+    def _ngramify(self, text, n):
+        ngram_dict = defaultdict(int)
+        start = 0
+        end = n
+        ngram = ''
+        for i in range(start, end):
+            ngram += text[i]
+            ngram += ' '
+        while end < len(text):
+            ngram_dict[ngram] += 1
+            ngram = ngram[len(text[start]) + 1:]
+            ngram += text[end]
+            ngram += ' '
+            start += 1
+            end += 1
+        ngram_dict[ngram] += 1
+        return ngram_dict
+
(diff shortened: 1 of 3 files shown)


vfdev-5:
Thanks for the PR @gucifer !

I few questions:
- module name is `rouge_n`, can we call it simply `rouge` ?
- can we think about a reference implementation for tests, for example in sklearn if there is one ?
- what are `peer, models`, data type, shape if any etc? And in terms of `y_pred` (model prediction) and `y` (target) ?

I'll add more questions in the code once I understood better.

> Comments Not added, what is the mathematical notation used?

We can do as here : 
https://github.com/pytorch/ignite/blob/964ae647df7a47e8088fe6a2b0d5162898eed3a7/ignite/metrics/accuracy.py#L97

gucifer:
- rouge_n because there are two types of rouge implementations, Firstly based on ngrams which I have implemented, second are based on Longest common sub-sequence which is called rouge_l.
- Test is very simple as both are bodies of texts so we will have a few paragraphs and compare rouge score. [Wikipedia definition](https://en.wikipedia.org/wiki/ROUGE_(metric))
- peer refer to the text corpus generated by the system. model refer to a model text corpus which are mostly written by humans. These terms come from language naming scheme (for eg model essays refers to essays which are really good whereas peer essays are written by your peer) and is not a Machine Learning naming scheme per se.

ignite/metrics/rouge_n.py
```diff
@@ -0,0 +1,85 @@
...
+            recall_total += max(len(model) - n + 1, 0)
+            for ngram in peer_dict:
+                if model_dict[ngram]:
+                    matches += peer_dict[ngram]
+        precision_total = len(models) * max((len(peer) - n + 1), 0)
+        return self._f1_score(matches, recall_total, precision_total)
+
+    @reinit__is_reduced
+    def reset(self):
+        self._rougelist = torch.tensor([], device=self._device)
```
vfdev-5, line 70: I think we can compute it a bit differently without storing as list but as running sum:
```
def reset():
    self._rouge_sum = torch.tensor(0, device=self._device)
    self._num_examples = 0

def update(self, output):
    ...
    # output of self.rouge_n should be a tensor
    self._rouge_sum += self.rouge_n(peer, models).to(self._device)
    self._num_examples += 1

def compute(self) -> float:
     ...
     return self._rouge_sum.item() / self._num_examples
```

gucifer: True but rouge score of every training sample is important cause it can help find if for a particular type of text generation the model is performing badly.

vfdev-5: Yes, this makes sense but we should handle it a bit differently.

PR #1646, opened 2021-02-16
ddp precision recall

### LINKED ISSUE #1283: Fix Recall/Precision multi_label, not averaged for DDP
## 🚀 Feature

Idea is to fix that 
https://github.com/pytorch/ignite/blob/002b595daa8a8345286c5e096c33e278948686a7/ignite/metrics/precision.py#L23-L30

fco-dv:
Fixes #1283

Description:
- Enable the case for Recall/Precision in multi_label, not averaged configuration for DDP
- Remove associated warnings, tests updated by scaling output size with `idist.get_world_size()`
- usage of `metric_device` should be checked in https://github.com/pytorch/ignite/blob/002b595daa8a8345286c5e096c33e278948686a7/tests/ignite/metrics/test_precision.py#L771
Check list:


sdesrozis:
@fco-dv well done !

Please could you update docstring ?
https://github.com/pytorch/ignite/blob/63a6f1be7d15b98183730ed6b0438a537396d91a/ignite/metrics/recall.py#L53

tests/ignite/metrics/test_precision.py
```diff
@@ -825,19 +825,13 @@ def update(engine, i):
...
 
     if idist.get_world_size() > 1:
-        with pytest.warns(
-            RuntimeWarning,
-            match="Precision/Recall metrics do not work in distributed setting when "
-            "average=False and is_multilabel=True",
-        ):
-            pr = Precision(average=False, is_multilabel=True)
-
+        pr = Precision(average=False, is_multilabel=True)
```
vfdev-5, line 828: First of all, could you please fix the issue we spotted during the call with unused `metric_device` that should go to `Precision` definition (line 795).

Ultimately, we would like to enable the tests in the same way as `test_multilabel_input_NCHW` where we run `_test(average=True, ...)` and `_test(average=False, ...)`. In case of average False, there is no equivalent option in sklearn (as probably it does not make much sense neither). So, please, take a look how we test things in `test_multilabel_input_NCHW` and do same thing here. Then we can simply remove this dummy test below.


PS.: average=False and is_multilabel=True gives as a result for precision = multilabel precision per sample
```python
import torch

pr = Precision(average=False, is_multilabel=True)

pr.update((torch.tensor(y_pred[:5, :]), torch.tensor(y_true[:5, :])))
pr.update((torch.tensor(y_pred[5:, :]), torch.tensor(y_true[5:, :])))
print(pr._true_positives, pr._positives, pr.compute(), pr.compute().mean().item())
> (tensor([2., 2., 2., 2., 2., 2., 1., 0.], dtype=torch.float64),
 tensor([2., 3., 2., 2., 2., 2., 1., 1.], dtype=torch.float64),
 tensor([1.0000, 0.6667, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 0.0000],
        dtype=torch.float64),
 0.8333333333333333)
```

vfdev-5 (commented):
Thanks for the PR @fco-dv ! Please, take a look at the comment on how to improve the test.

vfdev-5:
@fco-dv could you please udpate the tests according the suggestion: https://github.com/pytorch/ignite/pull/1646#discussion_r577214657

vfdev-5 (approved):
LGTM! Thanks @fco-dv !

PR #1661, opened 2021-02-21
chore(docs): add usage in supervised_training_steps

ydcjeff:
Follow up of #1589 

Description: add usage examples in supervised_training_steps

Check list:

### DIFF
--- a/ignite/engine/__init__.py
+++ b/ignite/engine/__init__.py
@@ -69,6 +69,15 @@ def supervised_training_step(
     Returns:
         Callable: update function.
 
+    Example::
+
+        model = ...
+        optimizer = ...
+        loss_fn = ...
+
+        update_fn = supervised_training_step(model, optimizer, loss_fn, 'cuda')
+        trainer = Engine(update_fn)
+
     .. versionadded:: 0.5.0
     """
 
@@ -114,6 +123,16 @@ def supervised_training_step_amp(
 
     Returns:
         Callable: update function
+
+    Example::
+
+        model = ...
+        optimizer = ...
+        loss_fn = ...
+        scaler = torch.cuda.amp.GradScaler(2**10)
+
+        update_fn = supervised_training_step_amp(model, optimizer, loss_fn, 'cuda', scaler=scaler)
+        trainer = Engine(update_fn)
 
     .. versionadded:: 0.5.0
     """
@@ -170,6 +189,15 @@ def supervised_training_step_apex(
     Returns:
         Callable: update function.
 
+    Example::
+
+        model = ...
+        optimizer = ...
+        loss_fn = ...
+
+        update_fn = supervised_training_step_apex(model, optimizer, loss_fn, 'cuda')
+        trainer = Engine(update_fn)
+
     .. versionadded:: 0.5.0
     """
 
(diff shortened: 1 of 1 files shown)


ydcjeff:
Should we remove `non_blocking` argument in `supervised_training_step_tpu`
as per
```py
non_blocking (bool, optional): if True and this copy is between CPU and GPU, the copy may occur asynchronously
```

ignite/engine/__init__.py
```diff
@@ -69,6 +69,15 @@ def supervised_training_step(
...
         Callable: update function.
 
+    Example::
+
+        model = ...
+        optimizer = ...
+        loss_fn = ...
+
+        update_fn = supervised_training_step(model, optimizer, loss_fn, 'cuda')
+        trainer = Engine(update_fn)
```
vfdev-5, line 79:
```suggestion

        from ignite.engine import Engne, supervised_training_step

        model = ...
        optimizer = ...
        loss_fn = ...

        update_fn = supervised_training_step(model, optimizer, loss_fn, 'cuda')
        trainer = Engine(update_fn)
```

Maybe, showing how to import could be helpful.

ignite/engine/__init__.py
```diff
@@ -115,6 +124,16 @@ def supervised_training_step_amp(
     Returns:
         Callable: update function
 
+    Example::
+
```
vfdev-5, line 128: Same here and below

vfdev-5 (approved):
Thanks a lot @ydcjeff !
Left a nit comment to add import.
As for non_blocking for TPU, we can keep the docstring, just adapt the text for TPU

ignite/engine/__init__.py
ydcjeff: Added in https://github.com/pytorch/ignite/pull/1661/commits/8ad516d07c05fa70d33acf93daeca4bac4181d96

ignite/engine/__init__.py
ydcjeff: Added in https://github.com/pytorch/ignite/pull/1661/commits/8ad516d07c05fa70d33acf93daeca4bac4181d96

ydcjeff:
> Do you know if it works for TPU this argument?

Asked in https://github.com/pytorch/xla/issues/2791

vfdev-5:
> > Do you know if it works for TPU this argument?
> 
> Asked in [pytorch/xla#2791](https://github.com/pytorch/xla/issues/2791)

Thanks @ydcjeff ! I would expect that everything is non-blocking in xla until an op that should sync and output a result. My intuition is that this parameter is not taken into account...

ydcjeff:
> > > Do you know if it works for TPU this argument?
> > 
> > 
> > Asked in [pytorch/xla#2791](https://github.com/pytorch/xla/issues/2791)
> 
> Thanks @ydcjeff ! I would expect that everything is non-blocking in xla until an op that should sync and output a result. My intuition is that this parameter is not taken into account...

Yes, I also think like that, since xla docs doesn't mention about `non_blocking`.
But, since I am not sure, I filed a question just to be sure. Let's wait for the answer.

ydcjeff:
So non_blocking is already taken care by xla, current docstring seems ok.
ready to go?

PR #1663, opened 2021-02-21
[DO NOT MERGE] code-style debugging

trsvchn:
PR to test code-style workflow behavior on forked repos

### DIFF
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -2,6 +2,8 @@ name: Format python code
 
 on:
   push:
+    branches:
+      - master
     paths-ignore:
       - "assets/**"
       - ".circleci/**"
@@ -10,6 +12,8 @@ on:
       - "conda.recipe"
       - "**.md"
   pull_request:
+    branches:
+      - master
     paths-ignore:
       - "assets/**"
       - ".circleci/**"
--- a/ignite/base/mixins.py
+++ b/ignite/base/mixins.py
@@ -1,10 +1,14 @@
 from collections import OrderedDict
 from collections.abc import Mapping
+
+
 class Serializable:
     _state_dict_all_req_keys = ()  # type: tuple
     _state_dict_one_of_opt_keys = ()  # type: tuple
+
     def state_dict(self) -> OrderedDict:
         pass
+
     def load_state_dict(self, state_dict: Mapping) -> None:
         if not isinstance(state_dict, Mapping):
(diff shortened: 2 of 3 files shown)


vfdev-5:
@trsvchn thanks for testing that! Let me enable code-formatting GA for that.

trsvchn:
ok, that was a test push with fork actions enabled

upstream workflow fails
but fork workflow pushes autopep8 fix, interesting...

trsvchn:
well, smth new:
https://github.com/pytorch/ignite/runs/1946964559#step:5:49

Looks like upstream workflow fails but fork's workflow pushes fine

going to disable fork actions

trsvchn:
I 've tried everything? but still:

> remote: Permission to pytorch/ignite.git denied to github-actions[bot].

vfdev-5:
@trsvchn thanks for investigating that. Probably, we have to search at `stefanzweifel/git-auto-commit-action@v4` source repo if there is something could be done with that...

trsvchn:
@vfdev-5 It Works!!

vfdev-5:
@trsvchn maybe we should disable pull_request option ? It still fails: https://github.com/pytorch/ignite/runs/1978982689

trsvchn:
@vfdev-5 Yes, I was thinking about that. But, how to handle cases with non-fork PRs? How often do we have them?

vfdev-5:
Well, they can be only created by maintainers so i'd say it is OK to ignore them now.

vfdev-5:
I also wonder why this autopep thing can not restart completely all CI tests, currently there are only 3 are shown...

trsvchn:
yeah, that makes sense, we can comment out them and waiting until GH team will fix double triggering :)
Anyway, push event will be active so, it is not a big loss

trsvchn:
> I also wonder why this autopep thing can not restart completely all CI tests, currently there are only 3 are shown...

huh, looks like a new opportunity to learn)

vfdev-5:
> > I also wonder why this autopep thing can not restart completely all CI tests, currently there are only 3 are shown...
> 
> huh, looks like a new opportunity to learn)

Let's take a break for a moment with that :)

trsvchn:
@vfdev-5 testing pr-labeler

trsvchn:
Oh, its not a good example since there are no matching files in PR

vfdev-5:
labeler works actually, see here : https://github.com/pytorch/ignite/pull/1733

PR #1664, opened 2021-02-21
feat(ci): create an issue automatically if cron jobs fail [skip ci]

### LINKED ISSUE #1659: Automatically create an issue if cron jobs fail
## 🚀 Feature
We currently have two cron jobs, `pytorch-version-tests` and `binaries-nightly-release`.
While doing #1589 , if I fail to skip some tests, that would fail in `pytorch-version-tests`, but not on master.
And, the failed cron jobs notifications will only be pushed to repo maintainers, not PR authors.
Since maintainers are usually track issue/pr via labels/milestones/projects, only a few look at github notifications.

Past example: #1580 

So why not create an issue automatically if cron jobs (2 currently) fail?
And, we can mention PR authors directly on that issues, less burden on maintainers and we should automate what can be automated, I think.

Inspires by : https://github.com/pytorch/vision/blob/master/.github/workflows/tests-schedule.yml

Relate actions:
- https://github.com/marketplace/actions/create-issue-from-file
- https://github.com/marketplace/actions/create-an-issue

ydcjeff:
Fixes #1659 

Description: create an issue automatically if cron jobs fail

Check list:

Sample created issue: https://github.com/ydcjeff/play-gh-actions/issues/28
Default env var: https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables
Link to the failed CI using:
```
Note: If you need to use a workflow run's URL from within a job, you can combine these environment variables: 

$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID
```

### DIFF
--- a/.github/failed_schedule_issue_template.md
+++ b/.github/failed_schedule_issue_template.md
@@ -1,0 +1,13 @@
+---
+title: Scheduled workflow failed
+labels:
+  - bug
+  - "module: datasets"
+---
+
+Oh no, something went wrong in the scheduled workflow **{{ env.GITHUB_WORKFLOW }}/{{ env.GITHUB_JOB }} with commit {{ env.GITHUB_SHA }}**.
+Please look into it:
+
+{{ env.GITHUB_SERVER_URL }}/{{ env.GITHUB_REPOSITORY }}/actions/runs/{{ env.GITHUB_RUN_ID }}
+
+Feel free to close this if this was just a one-off error.
--- a/.github/failed_schedule_issue_template.md
+++ b/.github/failed_schedule_issue_template.md
@@ -2,7 +2,6 @@
 title: Scheduled workflow failed
 labels:
   - bug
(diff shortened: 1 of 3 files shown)


.github/failed_schedule_issue_template.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Scheduled workflow failed
+labels:
+  - bug
+  - "module: datasets"
```
ydcjeff, line 5:
```suggestion
```

vfdev-5 (approved):
Looks good ! Thanks @ydcjeff !

vfdev-5:
@ydcjeff this night "PyTorch version tests" failed and we got spammed by 8 messages :) 
I think it would be better to conifgure something there to avoid that...

ydcjeff:
I miss that part. Yes it is possible.

Issue #1660, opened 2021-02-21
Fix code formatting github action


vfdev-5:
## 🐛 Bug description

Currently, auto code formatting GA: https://github.com/pytorch/ignite/actions/workflows/code-style.yml
does not on forks and fails while try to commit/push the fix to the fork. This gives a misleading error to the user without helping how to solve it. Currently this action is disabled.

The idea is to find out what is the cause of the failure and fix it if possible.

## Environment

- PyTorch Version (e.g., 1.4):
- Ignite Version (e.g., 0.3.0):
- OS (e.g., Linux):
- How you installed Ignite (`conda`, `pip`, source):
- Python version:
- Any other relevant information:

Devanshu24:
Hi @vfdev-5 !
I could try fixing this, the problem is the user has to enable GitHub actions on their fork (which GitHub disables by default)
We could add a line in CONTRIBUTING.md suggesting if you want to use auto formating please enable GitHub actions on your fork. :)

vfdev-5:
@Devanshu24 thanks for the info! Do you know if still there is a way to use/skip this action if user disabled actions ?

Devanshu24:
I do not at this moment, but I'll look it up! :)

trsvchn:
@vfdev-5 

Actions are disabled by default for forked repos (as @Devanshu24 already mentioned). When you're creating a PR to upstream repo (`pytorch/ignite`) it triggers worklfow from that repo (`pytorch/ignite`).

If you want to use GH Actions you need to toggle them manually for your fork. Then, if you're creating PR or pushing to the fork repo (`user/ignite`) it triggers local workflow (workflow file from the fork repo will be used).

So, I want to clarified, what kind of fail do we have 

> while try to commit/push the fix to the fork

do we assume that actions already activated for that fork?

trsvchn:
I've made some tests. Here #1663 you can see a PR with autopep8 fix made for a dummy python file in a fork w/failure. so, it works as expected.

But the most interesting thing happens when you're trying to add new commits to that branch, it triggers 2 same workflows simultaneously: one for push and one for pr:

![Screenshot from 2021-02-21 15-58-01](https://user-images.githubusercontent.com/30748114/108627359-cfae0180-745d-11eb-9f7e-596d5a26ac06.png)

trsvchn:
so we have double triggering here, and there is no way to fix it:

[...]

PR #1667, opened 2021-02-21
Add pr_target as a new code-style workflow triggering event

trsvchn:
Possible fix to #1660 

Description:

Adding new type event, as recommended here: [Using the Action in forks from public repositories](https://github.com/stefanzweifel/git-auto-commit-action#using-the-action-in-forks-from-public-repositories)

Check list:


vfdev-5:
@trsvchn thanks for the PR. Let's try here to make it commit to the fork on a code-format broken file. Maybe, such that we could see all of that in the action.
- Let's create a unformatted file (that we can finally remove).
- enable GA on your fork and commit to the PR -> We should see it fixing the broken file.
- disable GA on your fork and commit again to the PR originally broken file -> Normally, we should not see any failure with "Format python code / code-style (pull_request)"...

trsvchn:
1. Fork's GH Actions is on.
2. Committing broken file.

trsvchn:
1. Fork's GH Actions is off.
2. Committing broken file once again.

vfdev-5:
@trsvchn thanks ! Looks like there is no way to make it work if GA are disabled on fork ?

trsvchn:
@vfdev-5 yep :(

trsvchn:
@vfdev-5 I think this is related to this:

> The permissions for the GITHUB_TOKEN in forked repositories is read-only. 

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories-2

But in August 2020 they add pull_request_target that is granted a read/write repository token

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target

Looks, like it requires more research

vfdev-5:
Thanks for digging it up @trsvchn ! 

> Looks, like it requires more research

As you wish. Maybe, we can simply postpone on this issue and just disable code-style action for some moment...

vfdev-5:
Maybe, we can simply add the following condition `if: github.event.pull_request.head.repo.full_name == github.repository` when push action happens ?

trsvchn:
> Maybe, we can simply add the following condition `if: github.event.pull_request.head.repo.full_name == github.repository` when push action happens ?
> 
> Such that the action wont push and wont fail and if there is code-formatting issue it will be still reported by other actions.

Yeah, let's try it!

But anyway, I will continue that investigation :)

vfdev-5:
Sounds great, thanks !

vfdev-5:
Interesting idea @trsvchn ! Actually, how about instead of showing what to fix we could say:
1) either enable local GA
2) or install and run what we suggest in CONTRIBUTING
What do you think ?

Btw, I wonder how 121 chars line in ignite/metrics/psnr.py passed all previous checks ?

trsvchn:
> Btw, I wonder how 121 chars line in ignite/metrics/psnr.py passed all previous checks ?

@vfdev-5 Haha)) Yeah I was surprised as well)

> Actually, how about instead of showing what to fix we could say (post a message or in the code...):

Sure, that makes sense, we can raise a single error in actions log, I will prepare another demo :)

vfdev-5:
> Sure, that makes sense, we can raise a single error in actions log, I will prepare another demo :)

Or maybe, post a comment in the PR ?

trsvchn:
I am not sure that we can do that from the a workflow, and again not sure that event from the fork has a permission for that, I will check.

trsvchn:
> GitHub does not allow GitHub Actions running from a forked repo to modify parent repo, even to post comments on its own PR.

vfdev-5:
Yes, this seems correct. But here it is our own action running on our PR ... Actually, I think I have seen bots like that saying that the code is not correctly formatted and suggesting what to do ...

trsvchn:
@vfdev-5 Oh, yeah!  I think they have a kind of contributor access or smth, and use gh API for commenting and labeling.

trsvchn:
@vfdev-5 I think we can try similar approach to make a POST request from the workflow

vfdev-5:
OK, let's try that

trsvchn:
@vfdev-5 I have great news! I think I finally hacked `pull_request_target` (At least for my sandbox repo)!

Screenshot from 2021-02-25 13-26-23

The trick is that `pull_request_target` really works, but it doesn't start in PRs from forks that modify or create workflow file.

> ... instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets enabling the maintainer to safely comment on or label a pull request.

So, to enable fork events we need first to merge modified `code-style.yml` with `pull_request_target` enabled. Like I did here: https://github.com/trsvchn/gha-sandbox/pull/2

After that we can accept and handle PRs from the community)

https://github.com/trsvchn/gha-sandbox/pull/3

vfdev-5:
Sounds great ! Thanks a lot for digging into that, hope this could be helpful for you as well ! 
So, current version of our `code-style.yml` is OK or you think to update it to a finalized version ?

trsvchn:
@vfdev-5 Yes, it is, it's very helpful. 

yeah, I think `code-style.yml` is ready, I've tested this configuration in sandbox and it works. So we can start with this setup, anyway we can test it only after the merge)

vfdev-5:
OK, so I can merge it to the master and we'll see. So, current failure of `Format python code / code-style (pull_request)` is acceptable, right ?

trsvchn:
Yes, you are right! I have similar failure for `pull_request` event

vfdev-5 (approved):
Thanks @trsvchn !

trsvchn:
@vfdev-5 I reopened this https://github.com/pytorch/ignite/pull/1663 to see what happens

PR #1675, opened 2021-02-22
[BC-breaking] Move detach outside of loss function computation

### LINKED ISSUE #1674: Loss automatically detaching inputs breaks some workflows
## 🐛 Bug description

https://github.com/pytorch/ignite/blob/master/ignite/metrics/loss.py#L60

Before 0.4.3, the `y_pred` and `y` were passed without calling detach. Now because of detach a number of usecases are no longer possible.

An example is: https://docs.gpytorch.ai/en/v1.3.1/marginal_log_likelihoods.html#variationalelbo

The output of `model(train_x)` is an object which does not support detach. This leads to a crash when passing the `mll` function to Loss as a `loss_fn`.

Another example is https://github.com/y0ast/deterministic-uncertainty-quantification/blob/master/train_duq_cifar.py#L153

The loss is dependent on the gradient, which means that by calling detach the loss cannot be computed anymore.

I have been trying to work around it, but I can't figure out a nice way. https://pytorch.org/ignite/metrics.html#ignite.metrics.Average is not input size aware so it cannot correctly compute the average of the loss.

y0ast:
Fixes #1674

This is a proposal solution to the issue linked above. I am not 100% sure of the implications of this change.

However if it is deemed acceptable, I would hope that maybe it can be part of a 0.4.4 bugfix release.

### DIFF
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -57,7 +57,7 @@ class Loss(Metric):
             kwargs = {}  # type: Dict
         else:
             y_pred, y, kwargs = cast(Tuple[torch.Tensor, torch.Tensor, Dict], output)
-        average_loss = self._loss_fn(y_pred.detach(), y.detach(), **kwargs)
+        average_loss = self._loss_fn(y_pred, y, **kwargs).detach()
 
         if len(average_loss.shape) != 0:
             raise ValueError("loss_fn did not return the average loss.")


vfdev-5 (approved):
Looks good, thanks @y0ast !

sdesrozis (approved):
@y0ast Thank you !

vfdev-5:
@y0ast could you please send a follow-up PR that emulates your duq evaluator example, such that we verify this fix ? Thanks

y0ast:
@vfdev-5 do you mean in the form of a test? Or something else

vfdev-5:
@y0ast yes, exactly, a test which verifies that grads could still flow even with using `Loss`. Sorry, didn't mention that clearly.

Issue #1665, opened 2021-02-21
Use autodoc_typehints = "description" in conf.py


ydcjeff:
## 📚 Documentation

EDITED HERE: As a help wanted issue, please, see below comment https://github.com/pytorch/ignite/issues/1665#issuecomment-783240506 on what remains to do here. Thanks 

If there are many arguments in class/function/method, it's a little hard to read/scan all arguments (due to typehints).
Although there is a `Parameters` section, there might be missing some arguments (rarely though).
And we are duplicating type hints work in both signatures and docstring. Writing once can make less work on authors and less type hints mismatch can occur.

If we can remove type hints, it can be a little easier to read, and showing type hints only in `Parameters` section.

For example:
Before:
![Screen Shot 2021-02-21 at 22 52 16](https://user-images.githubusercontent.com/32727188/108631357-ff79fb00-7497-11eb-9409-79f9c7f12424.png)
After:
![Screen Shot 2021-02-21 at 22 51 40](https://user-images.githubusercontent.com/32727188/108631370-11f43480-7498-11eb-935f-c3fd5f69371f.png)

Changing to this docs style needs a little work. It needs to 
- add
```py
autoclass_content = "both"
autodoc_typehints = "description"  # new in sphinx 3.1
```
those lines in `conf.py`
- removing type hints in docstring
- leaving type hints only in signatures.

Let me know what you think.
cc: @vfdev-5 @sdesrozis @trsvchn

vfdev-5:
@ydcjeff thanks! Yes, it would be a nice improvement of the docs ! I also found that function's signature is cluttered by type hints which can hard to read. Let's introduce those configs to `conf.py`

trsvchn:
@ydcjeff Yes, I like it! BTW, I didn't know that we can do that so easily, I was expecting to use https://github.com/agronholm/sphinx-autodoc-typehints as a new dependency for that.

ydcjeff:
@vfdev-5 I will send PR tmr

@trsvchn yes, that `description` option is new in sphinx 3.1 so that we can do that

sdesrozis:
@ydcjeff typehints have to be removed from docstring, haven't they ? If so, It seems to be a huge work that we could scatter between us.

vfdev-5:
> @ydcjeff typehints have to be removed from docstring, haven't they ?

@sdesrozis why ? 

As far as I understand, it is just about adding to additional lines to `conf.py`.

ydcjeff:
[...]

PR #1676, opened 2021-02-22
docs: rm type hints in ignite.contrib.handlers (2)

ydcjeff:
Part of #1665 

Description:

Check list:

### DIFF
--- a/ignite/contrib/handlers/param_scheduler.py
+++ b/ignite/contrib/handlers/param_scheduler.py
@@ -40,7 +40,7 @@ class ParamScheduler(metaclass=ABCMeta):
         param_name: str,
         save_history: bool = False,
         param_group_index: Optional[int] = None,
-    ) -> None:
+    ):
 
         if not (
             isinstance(optimizer, Optimizer)
@@ -444,7 +444,7 @@ class ConcatScheduler(ParamScheduler):
 
     """
 
-    def __init__(self, schedulers: List[ParamScheduler], durations: List[int], save_history: bool = False) -> None:
+    def __init__(self, schedulers: List[ParamScheduler], durations: List[int], save_history: bool = False):
 
         if not isinstance(schedulers, Sequence):
             raise TypeError(f"Argument schedulers should be a sequence, but given {schedulers}")
@@ -672,7 +672,7 @@ class LRScheduler(ParamScheduler):
         trainer.add_event_handler(Events.ITERATION_COMPLETED, scheduler)
     """
 
-    def __init__(self, lr_scheduler: _LRScheduler, save_history: bool = False) -> None:
+    def __init__(self, lr_scheduler: _LRScheduler, save_history: bool = False):
 
         if not isinstance(lr_scheduler, _LRScheduler):
             raise TypeError(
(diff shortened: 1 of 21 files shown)


ignite/contrib/handlers/param_scheduler.py
```diff
@@ -194,7 +194,7 @@ def plot_values(cls, num_events: int, **scheduler_kwargs: Mapping) -> Any:
 
         Args:
             num_events: number of events during the simulation.
-            **scheduler_kwargs : parameter scheduler configuration kwargs.
+            scheduler_kwargs : parameter scheduler configuration kwargs.
```
vfdev-5, line 197:
```suggestion
            scheduler_kwargs: parameter scheduler configuration kwargs.
```

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -589,6 +589,7 @@ def simulate_values(  # type: ignore[override]
             durations: list of number of events that lasts a parameter scheduler from schedulers.
             param_names: parameter name or list of parameter names to simulate values.
                 By default, the first scheduler's parameter name is taken.
+            kwargs:
```
vfdev-5, line 592: this is a good question to what corresponds those kwargs

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -774,7 +775,7 @@ def create_lr_scheduler_with_warmup(
             by 100 simulated learning rate values.
 
     Returns:
-        ConcatScheduler: learning rate scheduler with linear warm-up.
+        learning rate scheduler with linear warm-up.
```
vfdev-5, line 778: Maybe, it's better to keep ConcatScheduler only ?

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -886,7 +887,7 @@ class PiecewiseLinear(ParamScheduler):
         param_group_index: optimizer's parameters group to use.
 
     Returns:
-        PiecewiseLinear: piecewise linear scheduler
+        piecewise linear scheduler instance
```
vfdev-5, line 890: Same here ?

ignite/contrib/handlers/param_scheduler.py
```diff
@@ -1083,11 +1085,12 @@ def simulate_values(cls, num_events: int, schedulers: List[_LRScheduler], **kwar
         """Method to simulate scheduled values during num_events events.
 
         Args:
-            num_events (int): number of events during the simulation.
-            schedulers (subclass of `torch.optim.lr_scheduler._LRScheduler`): lr_scheduler object to wrap.
+            num_events: number of events during the simulation.
+            schedulers: lr_scheduler object to wrap.
+            kwargs:
```
vfdev-5, line 1090:
```suggestion
            kwargs: kwargs passed to construct an instance of :class:`ignite.contrib.handlers.ParamGroupScheduler`.
```

vfdev-5 (commented):
Thanks a lot for working on that issue @ydcjeff !
I left few comments

ignite/contrib/handlers/param_scheduler.py
ydcjeff: Yes, `kwargs` is used in
```py
scheduler = cls(schedulers=schedulers, save_history=False, durations=durations, **kwargs)
```
`cls` is `ConcatScheduler` which only has 3 arguments, so kwargs is unnecessary?

vfdev-5: Yes, seems like that. Let's remove it from docstring and the function signature

vfdev-5 (commented):
Looks good ! A nit to remove kwargs from concat scheduler and good to go, I think.

vfdev-5 (approved):

PR #1679, opened 2021-02-22
add cohen kappa

### LINKED ISSUE #1673: add cohen kappa in contrib.metrics module
## 🚀 Feature
cohen's kappa is widely used in kaggle competitions, i think it would be great to add it in `ignite.contrib.metrics`

KickItLikeShika:
Fixes #1673 

Added Cohen's Kappa in `ignite.contrib.metrics`. Implemented `NonWeightedCohenKappa`, `LinearCohenKappa`, 'QuadraticCohenKappa' in `cohen_kappa` file. The implementation is similar to [average precision](urhttps://github.com/pytorch/ignite/blob/master/ignite/contrib/metrics/average_precision.pyl)

Check list:

### DIFF
--- a/ignite/contrib/metrics/cohen_kappa.py
+++ b/ignite/contrib/metrics/cohen_kappa.py
@@ -52,7 +52,7 @@ class CohenKappa(EpochMetric):
         weights (str): a string is used to define the type of Cohen's Kappa whether Non-Weighted or Linear or Quadratic
             (default: `None`)
         check_compute_fn (bool): Default False. If True, `cohen_kappa_score
-            <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.cohen_kappa_score.html>`_ 
+            <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.cohen_kappa_score.html>`_
             is run on the first batch of data to ensure there are
             no issues. User will be warned in case there are any issues computing the function.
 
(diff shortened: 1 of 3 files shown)


ignite/contrib/metrics/cohen_kappa.py
```diff
@@ -0,0 +1,131 @@
...
+        from sklearn.metrics import cohen_kappa_score
+    except ImportError:
+        raise RuntimeError("This contrib module requires sklearn to be installed.")
+
+    y_true = y_targets.numpy()
+    y_pred = y_preds.numpy()
+    return cohen_kappa_score(y_true, y_pred, weights=None)
+
+
+class NonWeightedCohenKappa(EpochMetric):
```
vfdev-5, line 19: I think it would be more simple to implement a single class with an additional arg `weights`.
try/except check may be better to put into constructor (same should have been done for AveragePrecision, I think).

vfdev-5 (commented):
@KickItLikeShika thanks for the PR, I left a comment on how to simplify the code. There are tests are missing and should be added to check if the implementation is working.

ignite/contrib/metrics/cohen_kappa.py
KickItLikeShika: But then the way the user uses this metrics it's going to be different from the other metrics

KickItLikeShika (commented):
@vfdev-5 I have tried to make it more simpler as you asked but i got some issues. The thing is the class `EpochMetric` we can not pass any additional args to it except the `y_true`, `y_pred`. We can not pass the weights arg. To make it simpler in a one class we have to see another approach. I'm ready if you told me so.

ignite/contrib/metrics/cohen_kappa.py
vfdev-5: You can implement it such that, weights is passed to `__init__` and before calling `super` with a function, you define this function according to `weights` value. Another way to do that is to wrap the function like
```python
def get_cohen_kappa_fn(weights):
    def wrapper(y_pred, y):
        y_true = y_targets.numpy()
        y_pred = y_preds.numpy()
        return cohen_kappa_score(y_true, y_pred, weights=weights)
    return wrapper
```

KickItLikeShika:
@vfdev-5 i have updated the implementation and added the test. Please have a look

ignite/contrib/metrics/cohen_kappa.py
```diff
@@ -0,0 +1,84 @@
+from typing import Callable
+
+import torch
+
+from ignite.metrics import EpochMetric
+
+
+def non_weighted_cohen_kappa_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> float:
+    try:
```
vfdev-5, line 9: Please, put this check inside `CohenKappa.__init__` and let's create a single function getter using the wrapper as I proposed previously. Thanks

KickItLikeShika: Okay i will close this pull request and create a new one with the updated implementation. Thanks

vfdev-5: You can commit here as well. Let's avoid closing and opening new PRs for updates.

PR #1681, opened 2021-02-22
Add gradient clipping to create_supervised_trainer()

### LINKED ISSUE #419: add gradient clipping to `create_supervised_trainer()`
It would be good to add gradient clipping to the trainers created by `create_supervised_trainer`. This is already provided by `torch.nn.utils.clip_grad_norm_`.

One possible implementation could be:

```python
import math
from torch.nn.utils import clip_grad_norm_

def create_supervised_trainer(model, optimizer, loss_fn,
                              device=None, non_blocking=False,
                              prepare_batch=_prepare_batch,
                              gradient_clip=math.inf):
    """
    Factory function for creating a trainer for supervised models.
    Args:
        model (`torch.nn.Module`): the model to train.
        optimizer (`torch.optim.Optimizer`): the optimizer to use.
        loss_fn (torch.nn loss function): the loss function to use.
        device (str, optional): device type specification (default: None).
            Applies to both model and batches.
        non_blocking (bool, optional): if True and this copy is between CPU and GPU, the copy may occur asynchronously
            tuple of tensors `(batch_x, batch_y)`.
        gradient_clip (float, optional): value to use to clip gradients.
    Note: `engine.state.output` for this engine is the loss of the processed batch.
    Returns:
        Engine: a trainer engine with supervised update function.
    """
    if device:
        model.to(device)

    def _update(engine, batch):
        model.train()
        optimizer.zero_grad()
        y_pred = model(x)
        loss = loss_fn(y_pred, y)
        loss.backward()
        clip_grad_norm_(model.parameters(), gradient_clip)
        optimizer.step()
        return loss.item()

    return Engine(_update)
```

KickItLikeShika:
Fixes #419

Added gradient clipping feature to `create_supervised_trainer()`, whatever the device or the amp mode.


ignite/contrib/metrics/cohen_kappa.py
```diff
@@ -0,0 +1,131 @@
+from typing import Callable
```
vfdev-5, line 1: This file should not be in this PR

KickItLikeShika: Has been fixed

PR #1685, opened 2021-02-23
docs: rm type hints in ignite.engine (3)

ydcjeff:
Part of #1665 
https://deploy-preview-1685--pytorch-ignite-preview.netlify.app/engine.html
Description:

Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -21,14 +21,14 @@ class Engine(Serializable):
     """Runs a given ``process_function`` over each batch of a dataset, emitting events as it goes.
 
     Args:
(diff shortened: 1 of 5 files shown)


ignite/engine/engine.py
```diff
@@ -118,6 +118,8 @@ def compute_mean_std(engine, batch):
 
     """
 
+    state: State
```
vfdev-5, line 121: Please, explain why to add these lines ?

ydcjeff: It is for adding type hints in attributes, but this should be done with upgrade sphinx 3.4 PR.
Removing now...

vfdev-5 (approved):
Thanks @ydcjeff !

PR #1691, opened 2021-02-23
docs(deps): up sphinx to 3.4.3, add napoleon_attr_annotations=True

ydcjeff:
Part of #1665 

Description: 
- upgrade sphinx to 3.4.3
- add `napoleon_attr_annotations=True` in conf.py

Check list:

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -206,8 +206,10 @@ todo_include_todos = True
 
 # -- Type hints configs ------------------------------------------------------
 
+autodoc_inherit_docstrings = True
 autoclass_content = "both"
 autodoc_typehints = "description"
+napoleon_attr_annotations = True
 
 # -- A patch that turns-off cross refs for type annotations ------------------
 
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -206,8 +206,10 @@
 
 # -- Type hints configs ------------------------------------------------------
 
+autodoc_inherit_docstrings = True
```
vfdev-5, line 209: Default is also True, but let's keep it like that

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html

vfdev-5 (approved):
Thanks @ydcjeff !

docs/source/conf.py
ydcjeff: Yes, we want to have `True` value, but keeping it just in case sphinx change to `False`

vfdev-5:
It's funny that the preview does not show
```
Notes[ + ]

Package Reference[ + ]

Contrib Package Reference[ + ]

Team[ + ]
```
the way how the docs lookk like now. @ydcjeff any ideas on how to keep the previous rendering with Package Reference, Contrib Package Reference and Team always expanded ?

ydcjeff:
I think it is related to the recent changes of `pytorch_sphinx_theme`

vfdev-5:
@ydcjeff so should we do something with it ?

PR #1692, opened 2021-02-23
add test for new loss detach behavior

y0ast:
Follow up of #1674 and #1675 

Add a test for the new detach behavior in the loss. This test only checks for non-failure, it doesn't check numerical correctness as that is expected to be tested elsewhere.

I verified manually that this test indeed fails before my previous PR.

### DIFF
--- a/tests/ignite/metrics/test_loss.py
+++ b/tests/ignite/metrics/test_loss.py
@@ -55,6 +55,7 @@ def test_non_averaging_loss():
 
 
 def test_gradient_based_loss():
+    # Tests https://github.com/pytorch/ignite/issues/1674
     x = torch.tensor([[0.1, 0.4, 0.5], [0.1, 0.7, 0.2]], requires_grad=True)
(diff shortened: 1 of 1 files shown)


tests/ignite/metrics/test_loss.py
```diff
@@ -54,6 +54,23 @@ def test_non_averaging_loss():
         loss.update((y_pred, y))
 
 
+def test_gradient_based_loss():
+    x = torch.tensor([[0.1, 0.4, 0.5], [0.1, 0.7, 0.2]], requires_grad=True)
```
vfdev-5, line 58:
```suggestion
    # Tests https://github.com/pytorch/ignite/issues/1674
    x = torch.tensor([[0.1, 0.4, 0.5], [0.1, 0.7, 0.2]], requires_grad=True)
```

vfdev-5 (approved):
Thanks a lot @y0ast !

PR #1690, opened 2021-02-23
add cohen kappa

KickItLikeShika:
Fixes #1673

Added Cohen's Kappa in `ignite.contrib.metrics`. Implemented 'CohenKappa' in `cohen_kappa.py` file. And added the tests in `test_cohen_kappa.py`.

Check list:

- [x] New tests are added (if a new feature is added)
- [x] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #1699, opened 2021-02-24
Added eps to avoid nans in canberra error

vfdev-5:
Description:
- Added eps to avoid nans in canberra error
- Should fix XLA issue in #1690

Check list:


vfdev-5:
@sdesrozis can you review this please and merge ?

sdesrozis (approved):
@vfdev-5 looks good!

An alternative implementation could be using `torch.nansum`

vfdev-5:
I'm afraid `nansum` is something recent and not supported by pth 1.3.1.

PR #1700, opened 2021-02-24
Improve contrib metrics

### LINKED ISSUE #1695: Improve `ignite.contirb.metrics` implementation
For these metrics:
- [Average Precision](https://github.com/pytorch/ignite/blob/master/ignite/contrib/metrics/average_precision.py)
- [Precision Recall Curve](https://github.com/pytorch/ignite/blob/master/ignite/contrib/metrics/precision_recall_curve.py)
- [Roc Auc](https://github.com/pytorch/ignite/blob/master/ignite/contrib/metrics/roc_auc.py) in `ignite.contrib.` 
We need to make the implementation simpler. Similar to the one here #1690. New arg `device` will be added, and we need to add the necessary distributed tests as well. See here for the tests https://github.com/pytorch/ignite/blob/master/tests/ignite/contrib/metrics/regression/test_canberra_metric.py#L99

KickItLikeShika:
Fixes #1695 

Improved the 3 implementation metrics in `ignite.contrib.metrics` [Average Precision](https://github.com/pytorch/ignite/blob/master/ignite/contrib/metrics/average_precision.py),  [Precision Recall Curve](https://github.com/pytorch/ignite/blob/master/ignite/contrib/metrics/precision_recall_curve.py), [Roc Auc](https://github.com/pytorch/ignite/blob/master/ignite/contrib/metrics/roc_auc.py) similar to [Cohen's Kappa](https://github.com/pytorch/ignite/blob/master/ignite/contrib/metrics/cohen_kappa.py) and added the `device` arg for the future support. Updated the tests, added distributed tests as well

Check list:

### DIFF
--- a/ignite/contrib/metrics/average_precision.py
+++ b/ignite/contrib/metrics/average_precision.py
@@ -1,19 +1,8 @@
-from typing import Callable
+from typing import Callable, Optional, Union
 
 import torch
 
 from ignite.metrics import EpochMetric
-
-
-def average_precision_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> float:
-    try:
-        from sklearn.metrics import average_precision_score
-    except ImportError:
-        raise RuntimeError("This contrib module requires sklearn to be installed.")
-
-    y_true = y_targets.numpy()
-    y_pred = y_preds.numpy()
-    return average_precision_score(y_true, y_pred)
 
 
 class AveragePrecision(EpochMetric):
@@ -22,14 +11,15 @@ class AveragePrecision(EpochMetric):
     sklearn.metrics.average_precision_score.html#sklearn.metrics.average_precision_score>`_ .
 
     Args:
-        output_transform (callable, optional): a callable that is used to transform the
+        output_transform: a callable that is used to transform the
             :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
-        check_compute_fn (bool): Default False. If True, `average_precision_score
+        check_compute_fn: Default False. If True, `average_precision_score
             <http://scikit-learn.org/stable/modules/generated/sklearn.metrics.average_precision_score.html
             #sklearn.metrics.average_precision_score>`_ is run on the first batch of data to ensure there are
             no issues. User will be warned in case there are any issues computing the function.
+        device: optional device specification for internal storage.
 
     AveragePrecision expects y to be comprised of 0's and 1's. y_pred must either be probability estimates or
     confidence values. To apply an activation to y_pred, use output_transform as shown below:
@@ -45,7 +35,33 @@ class AveragePrecision(EpochMetric):
 
     """
 
-    def __init__(self, output_transform: Callable = lambda x: x, check_compute_fn: bool = False) -> None:
+    def __init__(
+        self,
+        output_transform: Callable = lambda x: x,
+        check_compute_fn: bool = False,
+        device: Union[str, torch.device] = torch.device("cpu"),
+    ):
+
+        try:
+            from sklearn.metrics import average_precision_score
+        except ImportError:
+            raise RuntimeError("This contrib module requires sklearn to be installed.")
+
+        self.average_precision_compute = self.average_precision_compute_fn()
+
         super(AveragePrecision, self).__init__(
-            average_precision_compute_fn, output_transform=output_transform, check_compute_fn=check_compute_fn
+            self.average_precision_compute,
+            output_transform=output_transform,
+            check_compute_fn=check_compute_fn,
+            device=device,
         )
+
+    def average_precision_compute_fn(self) -> Callable[[torch.Tensor, torch.Tensor], float]:
+        from sklearn.metrics import average_precision_score
+
+        def wrapper(y_preds: torch.Tensor, y_targets: torch.Tensor) -> float:
+            y_true = y_targets.cpu().numpy()
+            y_pred = y_preds.cpu().numpy()
+            return average_precision_score(y_true, y_pred)
+
+        return wrapper
--- a/ignite/contrib/metrics/average_precision.py
+++ b/ignite/contrib/metrics/average_precision.py
@@ -5,6 +5,14 @@ import torch
 from ignite.metrics import EpochMetric
 
 
+def average_precision_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> float:
+    from sklearn.metrics import average_precision_score
+
+    y_true = y_targets.cpu().numpy()
+    y_pred = y_preds.cpu().numpy()
+    return average_precision_score(y_true, y_pred)
+
+
 class AveragePrecision(EpochMetric):
     """Computes Average Precision accumulating predictions and the ground-truth during an epoch
     and applying `sklearn.metrics.average_precision_score <http://scikit-learn.org/stable/modules/generated/
@@ -47,21 +55,9 @@ class AveragePrecision(EpochMetric):
         except ImportError:
             raise RuntimeError("This contrib module requires sklearn to be installed.")
 
-        self.average_precision_compute = self.average_precision_compute_fn()
-
         super(AveragePrecision, self).__init__(
-            self.average_precision_compute,
+            average_precision_compute_fn,
             output_transform=output_transform,
             check_compute_fn=check_compute_fn,
             device=device,
         )
-
-    def average_precision_compute_fn(self) -> Callable[[torch.Tensor, torch.Tensor], float]:
-        from sklearn.metrics import average_precision_score
-
-        def wrapper(y_preds: torch.Tensor, y_targets: torch.Tensor) -> float:
-            y_true = y_targets.cpu().numpy()
-            y_pred = y_preds.cpu().numpy()
-            return average_precision_score(y_true, y_pred)
-
-        return wrapper
--- a/ignite/contrib/metrics/precision_recall_curve.py
+++ b/ignite/contrib/metrics/precision_recall_curve.py
@@ -5,6 +5,20 @@ import torch
 from ignite.metrics import EpochMetric
 
 
+def precision_recall_curve_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> Union[Any, Any, Any]:
+    from sklearn.metrics import precision_recall_curve
+
+    y_true = y_targets.cpu().numpy()
+    y_pred = y_preds.cpu().numpy()
+    precision, recall, thresholds = precision_recall_curve(y_true, y_pred)
+    precision, recall, thresholds = (
+        torch.from_numpy(precision.copy()),
+        torch.from_numpy(recall.copy()),
+        torch.from_numpy(thresholds.copy()),
+    )
+    return precision, recall, thresholds
+
+
 class PrecisionRecallCurve(EpochMetric):
     """Compute precision-recall pairs for different probability thresholds for binary classification task
     by accumulating predictions and the ground-truth during an epoch and applying
@@ -44,25 +58,13 @@ class PrecisionRecallCurve(EpochMetric):
     ):
 
         try:
-            from sklearn.metrics import average_precision_score
+            from sklearn.metrics import precision_recall_curve
         except ImportError:
             raise RuntimeError("This contrib module requires sklearn to be installed.")
 
-        self.precision_recall_curve_compute = self.precision_recall_curve_compute_fn()
-
         super(PrecisionRecallCurve, self).__init__(
-            self.precision_recall_curve_compute,
+            precision_recall_curve_compute_fn,
             output_transform=output_transform,
             check_compute_fn=check_compute_fn,
             device=device,
         )
-
-    def precision_recall_curve_compute_fn(self) -> Tuple[Any, Any, Any]:
-        from sklearn.metrics import precision_recall_curve
-
-        def wrapper(y_preds: torch.Tensor, y_targets: torch.Tensor):
-            y_true = y_targets.cpu().numpy()
-            y_pred = y_preds.cpu().numpy()
-            return precision_recall_curve(y_true, y_pred)
-
-        return wrapper
--- a/ignite/contrib/metrics/precision_recall_curve.py
+++ b/ignite/contrib/metrics/precision_recall_curve.py
@@ -1,19 +1,8 @@
-from typing import Any, Callable, Tuple
+from typing import Any, Callable, Tuple, Union
 
 import torch
 
 from ignite.metrics import EpochMetric
-
-
-def precision_recall_curve_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> Tuple[Any, Any, Any]:
-    try:
-        from sklearn.metrics import precision_recall_curve
-    except ImportError:
-        raise RuntimeError("This contrib module requires sklearn to be installed.")
-
-    y_true = y_targets.numpy()
-    y_pred = y_preds.numpy()
-    return precision_recall_curve(y_true, y_pred)
 
 
 class PrecisionRecallCurve(EpochMetric):
@@ -23,14 +12,15 @@ class PrecisionRecallCurve(EpochMetric):
     sklearn.metrics.precision_recall_curve.html#sklearn.metrics.precision_recall_curve>`_ .
 
     Args:
-        output_transform (callable, optional): a callable that is used to transform the
+        output_transform: a callable that is used to transform the
             :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
-        check_compute_fn (bool): Default False. If True, `precision_recall_curve
+        check_compute_fn: Default False. If True, `precision_recall_curve
             <http://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_recall_curve.html
             #sklearn.metrics.precision_recall_curve>`_ is run on the first batch of data to ensure there are
             no issues. User will be warned in case there are any issues computing the function.
+        device: optional device specification for internal storage.
 
     PrecisionRecallCurve expects y to be comprised of 0's and 1's. y_pred must either be probability estimates
     or confidence values. To apply an activation to y_pred, use output_transform as shown below:
@@ -46,7 +36,33 @@ class PrecisionRecallCurve(EpochMetric):
 
     """
 
-    def __init__(self, output_transform: Callable = lambda x: x, check_compute_fn: bool = False) -> None:
+    def __init__(
+        self,
+        output_transform: Callable = lambda x: x,
+        check_compute_fn: bool = False,
+        device: Union[str, torch.device] = torch.device("cpu"),
+    ):
+
+        try:
+            from sklearn.metrics import average_precision_score
+        except ImportError:
+            raise RuntimeError("This contrib module requires sklearn to be installed.")
+
+        self.precision_recall_curve_compute = self.precision_recall_curve_compute_fn()
+
         super(PrecisionRecallCurve, self).__init__(
-            precision_recall_curve_compute_fn, output_transform=output_transform, check_compute_fn=check_compute_fn
+            self.precision_recall_curve_compute,
+            output_transform=output_transform,
+            check_compute_fn=check_compute_fn,
+            device=device,
         )
+
+    def precision_recall_curve_compute_fn(self) -> Tuple[Any, Any, Any]:
+        from sklearn.metrics import precision_recall_curve
+
+        def wrapper(y_preds: torch.Tensor, y_targets: torch.Tensor):
+            y_true = y_targets.cpu().numpy()
+            y_pred = y_preds.cpu().numpy()
+            return precision_recall_curve(y_true, y_pred)
+
+        return wrapper
--- a/ignite/contrib/metrics/roc_auc.py
+++ b/ignite/contrib/metrics/roc_auc.py
@@ -5,6 +5,28 @@ import torch
 from ignite.metrics import EpochMetric
 
 
+def roc_auc_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> float:
+    from sklearn.metrics import roc_auc_score
+
+    y_true = y_targets.cpu().numpy()
+    y_pred = y_preds.cpu().numpy()
+    return roc_auc_score(y_true, y_pred)
+
+
+def roc_auc_curve_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> Tuple[Any, Any, Any]:
+    from sklearn.metrics import roc_curve
+
+    y_true = y_targets.cpu().numpy()
+    y_pred = y_preds.cpu().numpy()
+    fpr, tpr, thresholds = roc_curve(y_true, y_pred)
+    fpr, tpr, thresholds = (
+        torch.from_numpy(fpr.copy()),
+        torch.from_numpy(tpr.copy()),
+        torch.from_numpy(thresholds.copy()),
+    )
+    return fpr, tpr, thresholds
+
+
 class ROC_AUC(EpochMetric):
     """Computes Area Under the Receiver Operating Characteristic Curve (ROC AUC)
     accumulating predictions and the ground-truth during an epoch and applying
@@ -48,24 +70,9 @@ class ROC_AUC(EpochMetric):
         except ImportError:
             raise RuntimeError("This contrib module requires sklearn to be installed.")
 
-        self.roc_auc_compute = self.roc_auc_compute_fn()
-
         super(ROC_AUC, self).__init__(
-            self.roc_auc_compute,
-            output_transform=output_transform,
-            check_compute_fn=check_compute_fn,
-            device=device,
+            roc_auc_compute_fn, output_transform=output_transform, check_compute_fn=check_compute_fn, device=device,
         )
-
-    def roc_auc_compute_fn(self) -> float:
-        from sklearn.metrics import roc_auc_score
-
-        def wrapper(y_preds: torch.Tensor, y_targets: torch.Tensor):
-            y_true = y_targets.cpu().numpy()
-            y_pred = y_preds.cpu().numpy()
-            return roc_auc_score(y_true, y_pred)
-
-        return wrapper
 
 
 class RocCurve(EpochMetric):
@@ -111,21 +118,9 @@ class RocCurve(EpochMetric):
         except ImportError:
             raise RuntimeError("This contrib module requires sklearn to be installed.")
 
-        self.roc_auc_curve_compute = self.roc_auc_curve_compute_fn()
-
         super(RocCurve, self).__init__(
-            self.roc_auc_curve_compute,
+            roc_auc_curve_compute_fn,
             output_transform=output_transform,
             check_compute_fn=check_compute_fn,
             device=device,
         )
-
-    def roc_auc_curve_compute_fn(self) -> Tuple[Any, Any, Any]:
-        from sklearn.metrics import roc_curve
-
-        def wrapper(y_preds: torch.Tensor, y_targets: torch.Tensor):
-            y_true = y_targets.numpy()
-            y_pred = y_preds.numpy()
-            return roc_curve(y_true, y_pred)
-
-        return wrapper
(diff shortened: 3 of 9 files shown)


ignite/contrib/metrics/average_precision.py
```diff
@@ -1,35 +1,25 @@
-from typing import Callable
+from typing import Callable, Optional, Union
 
 import torch
 
 from ignite.metrics import EpochMetric
 
 
-def average_precision_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> float:
```
vfdev-5, line 8: Tricky point here is that `average_precision_compute_fn` is a public method. Removing it will lead to a BC breaking change which is usually unwanted...

KickItLikeShika: Any idea to avoid that? and is there any other comments to edit everything and push again?

vfdev-5: I think you can just keep it where it is. I'll review this PR later.

KickItLikeShika: For `AveragePrecision` i will just update the test. Okay i will wait to see the full review

vfdev-5: I think the best what we can do is to keep this functions where they are and simplify things similar to what we did to CohenKappa:

```python
def average_precision_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> float:
     # no need to try/except here
     from sklearn.metrics import average_precision_score

    y_true = y_targets.numpy()
    y_pred = y_preds.numpy()
    return average_precision_score(y_true, y_pred)
```

```python
class AveragePrecision(EpochMetric):

    def __init__(self, ...):

        try:
            from sklearn.metrics import average_precision_score
        except ImportError:
            raise RuntimeError("This contrib module requires sklearn to be installed.")

        super(AveragePrecision, self).__init__(
            average_precision_compute_fn,
            output_transform=output_transform,
            check_compute_fn=check_compute_fn,
            device=device,
        )
```

KickItLikeShika: Is this for all metrics or just for `AveragePrecision`?

vfdev-5: Obviously, for all as otherwise it will be a BC-breaking change, right ?

KickItLikeShika: Okay then changing the try/except as provided and add the device arg will be good. Now for the tests, is there something to add/remove?

tests/ignite/contrib/metrics/test_average_precision.py
```diff
@@ -108,4 +148,100 @@ def update_fn(engine, batch):
     data = list(range(size // batch_size))
     ap = engine.run(data, max_epochs=1).metrics["ap"]
 
-    assert ap == np_ap
+    assert ap == pytest.approx(np_ap)
+
+
+def _test_distrib_compute(device):
```
vfdev-5, line 154: Integration test with Engine would be more interesting, I think. A minor testing issue here is that we perform only 1 update, but typically we do a lot of updates. This should be tested. @KickItLikeShika please take a look how we do tests for other core metrics and try to bake an integration test. You can indeed test that with CPU, please see CONTRIBUTING, I added a small section about how to run a distributed test. The idea is to define an Engine, attach the metric and feed in some random data which should be also different for every rank.

KickItLikeShika: What about updating the tests of `CohenKappa` the same way too? I will make tests similar to these here: [Accuracy](https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/test_accuracy.py), [Confusion Matrix](https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/test_confusion_matrix.py)

vfdev-5: Yes, CohenKappa should be also updated to more general testing way.

> I will make tests similar to these here: Accuracy, Confusion Matrix

Sounds good !

KickItLikeShika: Hey @vfdev-5, obviously i'm using `pytest.parametrize()` to test all weighting types in `CohenKappa`. Can i do that for distributed tests for test all the weighting types? I don't actually know but i tried that and it seemed working. Here is what i have done:
```
def _test_distirb_binary_input_N(device, weights):
    rank = idist.get_rank()
    
    def _test(metric_device):
        metric_device = torch.device(metric_device)
        ck = CohenKappa(device=metric_device, weights=weights)
        ..........
        ..........

    for _ in range(3):
        _test("cpu")
        if device.type != "xla":
            _test(idist.device())

@pytest.mark.distributed
@pytest.mark.parametrize("weights", [None, "linear", "quadratic"])
def test_distrib_cpu(distributed_context_single_node_gloo, weights):

    device = torch.device("cpu")
    _test_distirb_binary_input_N(device, weights)

```
What do you think?

vfdev-5: Looks great ! I do not know either if it can work but I think yes. What logs look like when you launch it with pytest ?

KickItLikeShika: Nothing suspicious, it works fine!

vfdev-5: I mean can you put an extract of it here with the tests related to CohenKappa

KickItLikeShika:
```
<frozen importlib._bootstrap>:219
<frozen importlib._bootstrap>:219
<frozen importlib._bootstrap>:219
<frozen importlib._bootstrap>:219
  <frozen importlib._bootstrap>:219: RuntimeWarning: numpy.ufunc size changed, may indicate binary incompatibility. Expected 192 from C header, got 216 from PyObject
```

vfdev-5: well, this does not help to understand :)
You can push your code to the PR and we can see how it works on the CI.

KickItLikeShika: @vfdev-5, i'm almost done, all the tests work fine but when running the tests with CPU as provided in CONTRIBUTING i get this error:
```
.......
tests/ignite/contrib/metrics/test_precision_recall_curve.py:277: in _test
    precision, recall, thresholds = prc.compute()
ignite/metrics/epoch_metric.py:145: in compute
    result = cast(float, idist.broadcast(result, src=0))
ignite/distributed/utils.py:405: in broadcast
    return _model.broadcast(tensor, src=src)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ignite.distributed.comp_models.native._NativeDistModel object at 0x7f6c69bf12e0>
tensor = (array([0.35      , 0.31578947, 0.33333333, 0.35294118, 0.375     ,
       0.33333333, 0.35714286, 0.30769231, 0.33333...434 , 0.61502165, 0.6177955 ,
       0.69359916, 0.69379395, 0.92483944, 0.9693086 , 0.9856615 ],
      dtype=float32))
src = 0

    def broadcast(self, tensor: Union[torch.Tensor, float, str], src: int = 0) -> Union[torch.Tensor, float, str]:
        if not isinstance(tensor, (torch.Tensor, Number, str)):
>           raise TypeError(f"Unhandled input type {type(tensor)}")
E           TypeError: Unhandled input type <class 'tuple'>

ignite/distributed/comp_models/base.py:174: TypeError

```
I have spent the past few hours searching this error but got nothing. And here is the code:
```

def _test_distirb_binary_input_N(device):
    rank = idist.get_rank()
    torch.manual_seed(12)
    
    def _test(metric_device):
        metric_device = torch.device(metric_device)
        prc = PrecisionRecallCurve(device=metric_device)

        torch.manual_seed(10 + rank)
        
        y_pred = torch.rand((10, ), device=device)
        y = torch.randint(0, 2, size=(10,), device=device).long()
        prc.update((y_pred, y))
        
        # gather y_pred, y
        y = idist.all_gather(y)

        np_y = y.cpu().numpy()
        np_y_pred = y_pred.cpu().numpy()

        sk_precision, sk_recall, sk_thresholds = precision_recall_curve(np_y, np_y_pred)
        precision, recall, thresholds = prc.compute()
        
        assert np.array_equal(precision, sk_precision)
        assert np.array_equal(recall, sk_recall)
        np.testing.assert_array_almost_equal(thresholds, sk_thresholds)
        ............
        ............
    for _ in range(3):
        _test("cpu")
        if device.type != "xla":
            _test(idist.device())

@pytest.mark.distributed
def test_distrib_cpu(distributed_context_single_node_gloo):

    device = torch.device("cpu")
    _test_distirb_binary_input_N(device)

```

vfdev-5: @KickItLikeShika the issue you encounter is with unsupported data type passed to our `broadcast` function as the error states:
```
        if not isinstance(tensor, (torch.Tensor, Number, str)):
>           raise TypeError(f"Unhandled input type {type(tensor)}")
E           TypeError: Unhandled input type <class 'tuple'>
```
normally, we should broadcast tensors, numbers or strings. But somehow a tuple is passed.

Broad cast happens in EpochMetric: 
```
ignite/metrics/epoch_metric.py:145: in compute
    result = cast(float, idist.broadcast(result, src=0))
```
on the result of self.compute_fn:
https://github.com/pytorch/ignite/blob/e6950b1f3a1f81762f31014b6741ed2c679860ba/ignite/metrics/epoch_metric.py#L141

As we say in the docs, the output of the function should be a scalar. Probably, you return something else in the function when compute PrecisionRecallCurve

KickItLikeShika: I can't really find a good and easy to use for the user, but i got another idea, `PrecisionRecallCurve` returns 3 arrays/tensors(precision, recall, thresholds), What about add an extra arg in the `PrecisionRecallCurve` to determine what you user wants to return `preicsion` or `recall` or `thresholds`. If you think this is a good idea, should i implement 3 functions for each return type? implementing it similar to `CohenKappa` and wieghts arg (BC-breaking)?

vfdev-5: Oh, I see better what is the issue: sklearn's precision_recall_curve returns a tuple and not a scalar. Thanks for pointing out that !

I think, what can be done is the following, let's improve EpochMetric to support various outputs. What can be done:
- add to EpochMetric: additional check that result is not None after this code
https://github.com/pytorch/ignite/blob/e6950b1f3a1f81762f31014b6741ed2c679860ba/ignite/metrics/epoch_metric.py#L141

- try to see if we could use [apply_to_type](https://github.com/pytorch/ignite/blob/e6950b1f3a1f81762f31014b6741ed2c679860ba/ignite/utils.py#L45) function to apply `idist.broadcast` on the result: `apply_to_type(result, (torch.Tensor, Number, str), partial(idist.broadcast, src=0))` ?

KickItLikeShika: @vfdev-5 `apply_to_type` raised a `TypeError`
```
E       TypeError: x must contain (<class 'torch.Tensor'>, <class 'float'>, <class 'str'>), dicts or lists; found <class 'numpy.ndarray'>
```
The updated `EpocMetric`

```
from functools import partial
from ignite.utils import apply_to_type
.........
.........
if idist.get_rank() == 0:
            result = self.compute_fn(_prediction_tensor, _target_tensor)
            if result is not None:
                apply_to_type(result, (torch.Tensor, float, str), partial(idist.broadcast, src=0))
.......
```

vfdev-5: We can return tensors from our compute_fn in Precision Recall Curve ?

KickItLikeShika: I proposed to return the user chooses to pass as arg `precision` or `recall` or `thresholds` and make 3 functions each of them returns something the user wants. One extra thing, whatever we do of `PrecisionRecallCurve` we will do it for `RocCurve` because it aslo return fpr, tpr, thresholds.

KickItLikeShika: But we can return tuple of tensors from compute_fn? if that's what you're asking i think yes

vfdev-5: Yes, let's return a tuple of tensors

KickItLikeShika: I tried this and everything now works fine! thank you sir!
and here is the implementation for `PreicsionRecallCurve` and `RocCurve` is similar
```
def precision_recall_curve_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> Union[Any, Any, Any]:
    from sklearn.metrics import precision_recall_curve

    y_true = y_targets.cpu().numpy()
    y_pred = y_preds.cpu().numpy()
    precision, recall, thresholds = precision_recall_curve(y_true, y_pred)
    precision, recall, thresholds = torch.from_numpy(precision.copy()), torch.from_numpy(recall.copy()), torch.from_numpy(thresholds.copy())
    return precision, recall, thresholds
```
The reason for using copy is sometimes the strides are negative then we will get this error:
```
ValueError: At least one stride in the given numpy array is negative, and tensors with negative strides are not currently supported. (You can probably work around this by making a copy of your array  with array.copy().)
```

KickItLikeShika: @vfdev-5, i have committed the new changed but it didn't pass the CI (expected), just to leave it for review and your tips to pass the CI, Thanks

KickItLikeShika: @vfdev-5 there are some `RuntimeErrors` please check it and tell me how can i fix that

ignite/contrib/metrics/precision_recall_curve.py
```diff
@@ -1,19 +1,22 @@
...
 
 from ignite.metrics import EpochMetric
 
 
-def precision_recall_curve_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> Tuple[Any, Any, Any]:
-    try:
-        from sklearn.metrics import precision_recall_curve
-    except ImportError:
-        raise RuntimeError("This contrib module requires sklearn to be installed.")
+def precision_recall_curve_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> Union[Any, Any, Any]:
```
vfdev-5, line 8:
```suggestion
def precision_recall_curve_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
```

ignite/contrib/metrics/precision_recall_curve.py
```diff
@@ -1,19 +1,22 @@
...
+    from sklearn.metrics import precision_recall_curve
 
-    y_true = y_targets.numpy()
-    y_pred = y_preds.numpy()
-    return precision_recall_curve(y_true, y_pred)
+    y_true = y_targets.cpu().numpy()
+    y_pred = y_preds.cpu().numpy()
+    precision, recall, thresholds = precision_recall_curve(y_true, y_pred)
```
vfdev-5, line 13: I think mypy shouldn't be happy with variable redefinition with different types...
```suggestion
    np_precision, np_recall, np_thresholds = precision_recall_curve(y_true, y_pred)
```

ignite/contrib/metrics/precision_recall_curve.py
```diff
@@ -1,19 +1,22 @@
...
+    precision, recall, thresholds = (
+        torch.from_numpy(precision.copy()),
```
vfdev-5, line 15: You mention negative strides here in numpy array. This is unusual I'd say... Can't make those arrays contiguous (which will certainly copy the data)?

vfdev-5:
There is a bunch of errors now:
- https://github.com/pytorch/ignite/pull/1700/checks?check_run_id=1994358057#step:8:341
- distrib tests are failing due to an error with test of average precision: https://github.com/pytorch/ignite/pull/1700/checks?check_run_id=1994358063#step:10:1200
(I do not see the exact error there neither)

ignite/contrib/metrics/roc_auc.py
```diff
@@ -1,30 +1,30 @@
...
+    from sklearn.metrics import roc_auc_score
 
-    y_true = y_targets.numpy()
-    y_pred = y_preds.numpy()
+    y_true = y_targets.cpu().numpy()
+    y_pred = y_preds.cpu().numpy()
     return roc_auc_score(y_true, y_pred)
 
 
 def roc_auc_curve_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> Tuple[Any, Any, Any]:
```
vfdev-5, line 16: Same here about type annotation for the output

KickItLikeShika:
@vfdev-5. Here are the errors for all the distrib tests
```
RuntimeError: [/opt/conda/conda-bld/pytorch_1603729096996/work/third_party/gloo/gloo/transport/tcp/unbound_buffer.cc:136] Timed out waiting 60000ms for send operation to complete
```
First everything was working fine, then i got these errors for all the tests, i thought the i got some issue on my machine but turns out no!

ignite/metrics/epoch_metric.py
```diff
@@ -139,6 +141,11 @@ def compute(self) -> float:
         if idist.get_rank() == 0:
             # Run compute_fn on zero rank only
             result = self.compute_fn(_prediction_tensor, _target_tensor)
+            if result is not None:
+                try:
+                    apply_to_type(result, (torch.Tensor, float), partial(idist.broadcast, src=0))
```
vfdev-5, line 146: This is wrong and this makes machine to hang

vfdev-5: The point was to raise a more comprehensive error message.
The idea is that we now add a feature to handle other types than only `torch.Tensor`: tuple, list or mapping of numbers, tensors etc. We also have to update the docstring for EpochMetric saying about supported outputs of compute_fn.
Well, I have to think a bit more about how to do that properly...

KickItLikeShika: I tried the solution you proposed 
```
if idist.get_rank() == 0:
    result = self.compute_fn(_prediction_tensor, _target_tensor)
    if result is not None:
        apply_to_type(result, (torch.Tensor, float, str), partial(idist.broadcast, src=0))
```
but it didn't pass the `test_epoc_metric` because EpocMetric inputs int sometimes so i changed it to 
```
if idist.get_rank() == 0:
    result = self.compute_fn(_prediction_tensor, _target_tensor)
    if result is not None:
        apply_to_type(result, (torch.Tensor, float, int, str), partial(idist.broadcast, src=0))
```
but that's obviously doesn't work with mypy, the second arg must be tuple, so i did it the way you see

vfdev-5:
@KickItLikeShika I think we should advance with this PR step by step and split it into several PR:

1st PR: improve EpochMetric to work with compute_fn outputs: scalars, tensors, tuple/list/mapping of tensors.
2nd PR: this PR

ignite/metrics/epoch_metric.py
vfdev-5:
```
if idist.get_rank() == 0:
    result = self.compute_fn(_prediction_tensor, _target_tensor)
    if result is not None:
        apply_to_type(result, (torch.Tensor, float, str), partial(idist.broadcast, src=0))
```
Let me explain why this is wrong:
- `apply_to_type` returns an output and is not inplace as you use it
- we can not call `idist.broadcast` for a single proc only (if idist.get_rank() == 0)

KickItLikeShika:
@vfdev-5 Should i open another PR for improving EpocMetric or we keep it all here? Can you please the steps required to improve EpocMetric?

vfdev-5:
Yes, two separate PRs. We pause on this PR now and will work on a new PR with improvements for EpochMetric only.

KickItLikeShika:
@vfdev-5, Sounds great! Please open an issue explaining what we should do, and i will start working on it indeed

KickItLikeShika:
Hey @vfdev-5, Please tell me if you got a solution for the EpocMetric update? If not, What about a similar function to `apply_to_type` but the second arg might be `tuple(torch.Tensor, float, int)`? because it the solution worked well but we needed to allow the input to be `int`(for epoch metric tests) or `float`(for most of the tests). Thank you.

vfdev-5:
Hey @KickItLikeShika, I do not understand your question, I think I've already explained above what to do with EpochMetric:
- https://github.com/pytorch/ignite/pull/1700#issuecomment-787151203
- ...

And `apply_to_type` works on multiple types without any modifications:
```python
import torch
from ignite.utils import apply_to_type


def func(x):
    print(x, type(x))
    return x * 2
    

for x in [torch.tensor([0.0]), 1, 1.0]:
    y = apply_to_type(x, (torch.Tensor, int, float), func)
    print("y=", y)

> tensor([1.]) <class 'torch.Tensor'>
y= tensor([2.])
1 <class 'int'>
y= 2
1.0 <class 'float'>
y= 2.0
```
I already explained you why your solution does not work in distributed, please, reread attentively our discussion here.

KickItLikeShika:
@vfdev-5, Thank for your explanation and i already tried the solution you provided above, but `apply_to_type` expects the second arg to be `Union[Type, Tuple[Type[Any], Any]]` so `apply_to_type(x, (torch.Tensor, int, float), func)` doesn't work well with mypy checks.

vfdev-5:
OK, I see better your concern. Let's just change apply_to_type's type hints to 
```
def apply_to_type(
    x: Union[Any, collections.Sequence, collections.Mapping, str, bytes],
    input_type: Union[Type, Tuple[Type, ...]],
    func: Callable,
) -> Union[Any, collections.Sequence, collections.Mapping, str, bytes]:
```

ignite/contrib/metrics/precision_recall_curve.py
KickItLikeShika: Yes i can use `numpy.asascontiguousarray` before converting to tensors, this will work, what about that?

vfdev-5: Actually, can you paste here a code to repro the output with negative strides in numpy array ?

KickItLikeShika:
```
y = np.random.randint(0, 2, size=(5,1))
y_pred = np.random.rand(5,1)
np_precision, np_recall, np_thresholds = precision_recall_curve(y, y_pred)
precision, recall, thresholds = torch.from_numpy(np_precision), torch.from_numpy(np_recall), torch.from_numpy(np_thresholds)
  File "<stdin>", line 1, in <module>
ValueError: At least one stride in the given numpy array is negative, and tensors with negative strides are not currently supported. (You can probably work around this by making a copy of your array  with array.copy().) 
```
The problem here is with `thresholds`
```
precision, recall, thresholds = torch.from_numpy(np_precision), torch.from_numpy(np_recall), torch.from_numpy(np_thresholds.copy())
```
This works fine

vfdev-5: Sounds to copy thresholds. Please, also put a comment about negatives stride in thresholds and a workaround by `.copy`.

KickItLikeShika:
Hey @vfdev-5, I have some questions:
- Obviously anything other than PyTorch tensor doesn't work. I have tried to loop through the tuple of tensors and broadcast each at a time like this: 
```
if type(result) == tuple:
    for tens in result:
        print("Naive print for debugging")
        tens = apply_to_type(tens, (torch.Tensor, float, int), partial(idist.broadcast, src=0))
```
It works fine and broadcasts the first tensor, but it doesn't work after that. Please tell me if you have an idea about broadcasting more than one tensor (list, tuple, etc)?
- The broadcasting step should be after this check
```
if idist.get_rank() == 0:
```
or this check 
```
if ws > 1:
```
Because it works fine if the `ws` is smaller than 2. 

This is the error occurs in this case:
```
 def broadcast(tensor,
                  src,
                  group=group.WORLD,
                  async_op=False):
        """
        Broadcasts the tensor to the whole group.
    
        ``tensor`` must have the same number of elements in all processes
        participating in the collective.
    
        Arguments:
            tensor (Tensor): Data to be sent if ``src`` is the rank of current
                process, and tensor to be used to save received data otherwise.
            src (int): Source rank.
            group (ProcessGroup, optional): The process group to work on
            async_op (bool, optional): Whether this op should be an async op
    
        Returns:
            Async work handle, if async_op is set to True.
            None, if not async_op or if not part of the group
    
        """
        _check_single_tensor(tensor, "tensor")
        if _rank_not_in_group(group):
            return
    
        opts = BroadcastOptions()
        opts.rootRank = src
        opts.rootTensor = 0
    
        if group == GroupMember.WORLD:
            _check_default_pg()
            work = _default_pg.broadcast([tensor], opts)
        else:
            group_src_rank = _get_group_rank(group, src)
            opts.rootRank = group_src_rank
            work = group.broadcast([tensor], opts)
        if async_op:
            return work
        else:
>           work.wait()
E           RuntimeError: [/pytorch/third_party/gloo/gloo/transport/tcp/pair.cc:566] Read error [127.0.0.1]:60647: Connection reset by peer
```

KickItLikeShika:
@vfdev-5, Here is the code and the error
```
if ws > 1:
            # broadcast result to all processes
            if type(result) == tuple:
                for tens in result:
                    print("Naive print for debugging")
                    
                    # All below gives the same error
                    tens = apply_to_type(tens, (torch.Tensor, float, int), partial(idist.broadcast, src=0))
                    # tens = idist.broadcast(tens, src=0)
                    # tens = cast(float, idist.broadcast(tens, src=0))
                
            else:
                result = cast(float, idist.broadcast(result, src=0))        
```
I'm just now trying to solve the tuple issue then i will improve it once i know to solve this.
Here is the error, that means that `apply_to_type` is fine. I think we have to see how to broadcast more than one tensor. It broadcasts the first tensor then fails.
```
def broadcast(tensor,
                  src,
                  group=group.WORLD,
                  async_op=False):
        """
        Broadcasts the tensor to the whole group.
    
        ``tensor`` must have the same number of elements in all processes
        participating in the collective.
    
        Arguments:
            tensor (Tensor): Data to be sent if ``src`` is the rank of current
                process, and tensor to be used to save received data otherwise.
            src (int): Source rank.
            group (ProcessGroup, optional): The process group to work on
            async_op (bool, optional): Whether this op should be an async op
    
        Returns:
            Async work handle, if async_op is set to True.
            None, if not async_op or if not part of the group
    
        """
        _check_single_tensor(tensor, "tensor")
        if _rank_not_in_group(group):
            return
    
        opts = BroadcastOptions()
        opts.rootRank = src
        opts.rootTensor = 0
    
        if group == GroupMember.WORLD:
            _check_default_pg()
            work = _default_pg.broadcast([tensor], opts)
        else:
            group_src_rank = _get_group_rank(group, src)
            opts.rootRank = group_src_rank
            work = group.broadcast([tensor], opts)
        if async_op:
            return work
        else:
>           work.wait()
E           RuntimeError: [/pytorch/third_party/gloo/gloo/transport/tcp/pair.cc:566] Read error [127.0.0.1]:7856: Connection reset by peer
```

vfdev-5:
Let me show you how I'd code compute method for `EpochMetric`:
```python
    def compute(self) -> float:
        if len(self._predictions) < 1 or len(self._targets) < 1:
            raise NotComputableError("EpochMetric must have at least one example before it can be computed.")

        _prediction_tensor = torch.cat(self._predictions, dim=0)
        _target_tensor = torch.cat(self._targets, dim=0)

        ws = idist.get_world_size()

        if ws > 1 and not self._is_reduced:
            # All gather across all processes
            _prediction_tensor = cast(torch.Tensor, idist.all_gather(_prediction_tensor))
            _target_tensor = cast(torch.Tensor, idist.all_gather(_target_tensor))
        self._is_reduced = True

        result = 0.0
        if idist.get_rank() == 0:
            result = self.compute_fn(_prediction_tensor, _target_tensor)

        # compute_fn outputs: scalars, tensors, tuple/list/mapping of tensors.
        # this should be clearly mentioned in the docstring
        if not isintance(result, (int, float, torch.Tensor, Sequence, Mapping)):
            raise TypeError("...")

        if ws > 1:
            # broadcast result to all processes
            # maybe, need a cast to satisfy mypy
            result = apply_to_type(result, (torch.Tensor, float, int), partial(idist.broadcast, src=0))

        return result
```

KickItLikeShika:
This PR will be closed, and I will work on a new PR and send it asap

PR #1702, opened 2021-02-25
Removed size limitation for str on collective ops

### LINKED ISSUE #1697: Remove 1024 limit for str on collective ops
## 🚀 Feature

Currently, if we have a limitation of 1024 chars for str input on collective ops:

https://github.com/pytorch/ignite/blob/314b87f490eeb812bca8b0b4911a4a60a847c010/ignite/distributed/comp_models/base.py#L92-L95

The idea is to remove this limitation. We can do that as following:
- for all inputs as str, we can first gather all lenghts
- find the max lengths
- pad the input to the max lenghts + add input length value (as it is done currently)
- perform collective op

Bug ref:  https://github.com/Project-MONAI/MONAI/issues/1633

vfdev-5:
Fixes #1697

Description:
- Removed size limitation for str on collective ops: all_gather, broadcast
- Added MIN, MAX, PRODUCT options for horovod all reduce
- Added more distrib tests
- minor consistency improvements

Check list:

### DIFF
--- a/.github/workflows/hvd-tests.yml
+++ b/.github/workflows/hvd-tests.yml
@@ -63,7 +63,7 @@ jobs:
       - name: Run Tests
         shell: bash -l {0}
         run: |
-          SKIP_DISTRIB_TESTS=${{ matrix.skip-distrib-tests }} bash tests/run_cpu_tests.sh
+          bash tests/run_cpu_tests.sh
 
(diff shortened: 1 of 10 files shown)


sdesrozis (approved):
@vfdev-5 It looks good !

PR #1707, opened 2021-02-25
chore: remove unused imports in tests

ydcjeff:
Fixes #{issue number}

Description: remove unused imports in tests

Check list:

### DIFF
--- a/tests/ignite/contrib/handlers/test_visdom_logger.py
+++ b/tests/ignite/contrib/handlers/test_visdom_logger.py
@@ -938,8 +938,6 @@ def test_integration_with_executor_as_context_manager(visdom_server, visdom_serv
 def no_site_packages():
     import sys
 
-    import visdom
-
     plx_module = sys.modules["visdom"]
     del sys.modules["visdom"]
     prev_path = list(sys.path)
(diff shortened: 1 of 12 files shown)


tests/ignite/contrib/handlers/test_visdom_logger.py
```diff
@@ -938,8 +938,6 @@ def update_fn(engine, batch):
 def no_site_packages():
     import sys
 
-    import visdom
-
     plx_module = sys.modules["visdom"]
```
vfdev-5, line 941: I think we should keep `import visdom` as otherwise it wont appear in sys modules. Maybe, same for other `no_site_packages` fixtures.

```suggestion
    visdom_module = sys.modules["visdom"]
```

ydcjeff: Test passed locally tho.

vfdev-5: I support that they can pass if there is a test that imports visdom before the test which calls `no_site_packages`... I think we should keep the import for consistency.

vfdev-5 (approved):
Looks good ! Thanks @ydcjeff

PR #1712, opened 2021-02-27
Add new blog post

trsvchn:
Fixes #{issue number}

Description:

Add new blog post about library

Check list:


vfdev-5:
Thanks for the link @trsvchn but it is yet another subjective comparision with incomplete info:
- we do support TPUs but they say only CPU and GPU
Well, I'm not sure that this represents any particular value for our community... 
What do you think ?

sdesrozis:
I agree with @vfdev-5 that this article does not take into account our `idist` module. It's a pity, we should make the effort to present it.

trsvchn:
@vfdev-5 @sdesrozis 
Was glad to hear your opinions!)

yeah, article is quite subjective and doesn't reflect all advantages of using ignite.

PR #1713, opened 2021-02-27
add match pytest for metrics

KickItLikeShika:
Added match for `pytest.raises` for most all the metrics don't have it

Check list:


vfdev-5 (approved):

PR #1718, opened 2021-02-28
Refactor test_loss.py

### LINKED ISSUE #1469: Parametrize metric tests
## 🚀 Feature

Use `pytest.mark.parametrize` for metric testing

Moh-Yakoub:
Fixes #1469

Description:  I did some refactor for test_loss. I noticed that we tend to repeat the same tensors over test cases. That can be made more efficient by using pytest.fixtures.

Once the approach is approved I will unify this across all other test cases in later PRs.

Check list:


vfdev-5:
@Moh-Yakoub thanks for the PR ! However, I'm not sure about the pros of this approach.
Btw, TPU CI failure is related.

Moh-Yakoub:
> @Moh-Yakoub thanks for the PR ! However, I'm not sure about the pros of this approach.
> Btw, TPU CI failure is related.

I will update the TPU test.
I believe using fixtures makes our testing more standardized. I feel that we shouldn't be repeating the fixtures - which is a constant value - across different test cases. also it allow us to add more tests in the future without just copying and pasting those values.

vfdev-5:
I see your point, I think also we should replace const data by random data as well with fixed seed. Looking at the tests before the PR's change, they do not look that good neither, distributed tests are very mean and should be updated, IMO. 
Personally, i'd prefer a common method reuse instead of fixtures... Especially those like
```
@pytest.fixture
def device():
    return None


@pytest.fixture
def requires_grad():
    return False
```

Moh-Yakoub:
> I see your point, I think also we should replace const data by random data as well with fixed seed. Looking at the tests before the PR's change, they do not look that good neither, distributed tests are very mean and should be updated, IMO.
> Personally, i'd prefer a common method reuse instead of fixtures... Especially those like
> 
> ```
> @pytest.fixture
> def device():
>     return None
> 
> 
> @pytest.fixture
> def requires_grad():
>     return False
> ```

I've updated the tests to use method and fixed the missing test.

vfdev-5 (approved):
Thanks @Moh-Yakoub

Issue #1644, opened 2021-02-15
Improve docker build scripts


vfdev-5:
## 🚀 Feature

Currently, we can build and publish dockers using Circle CI and it works more or less well, however, PR's modifying Dockerfile can not build docker images to ensure PR correctness (related to the way how the PR is checkout on Circle CI).

The idea is
- refactor all `build_all.sh` scripts (docker/main/build_all.sh, docker/hvd/build_all.sh and docker/msdp/build_all.sh) into a single shell script that could build a single docker image.
- use github actions to build only on PR a single docker image and introduce a single yml file for that (to see what can be done).
  -  publishing will use existing scripts
- add a test inside docker folder as a python script to check docker image versions for : torch, ignite ; can import cv2 ; maybe something else.

To Do list:

- [x] refactor `build_all.sh` into `build.sh`
- [x] run docker image build for PR on GA
- [ ] run tests on built image. Tests
  - For all images
    - can import torch and its version == required one
    - can import ignite and its version == required one
  - for all `-vision` images
    - can import opencv without driver issue
  - for all horovod images
    - can import horovod and its version == required one
  - for all msdp images
    - can import deepspeed and its version == required one

cc @fco-dv

sparkingdark:
I am not sure about the GitHub actions, that build the docker image from the pr, but I think the first point is an idea to merge all shell file to one main shell script.

what is the difference between the folders `main` , `hvd`, and `msdp`

vfdev-5:
@sparkingdark checkout https://github.com/pytorch/ignite/tree/master/docker and installation section. It is about having different frameworks in addition to pytorch and ignite.

sparkingdark:
Okay I will check that.

vfdev-5:
@trsvchn would you like to tackle this issue ?

trsvchn:
@vfdev-5 sure!

vfdev-5:
I'll start on this point
- refactor all build_all.sh scripts (docker/main/build_all.sh, docker/hvd/build_all.sh and docker/msdp/build_all.sh) into a single shell script that could build a single docker image.

If you could help with:
- create gtihub actions
- add a test inside docker folder as a python script to check docker image versions for : torch, ignite ; can import cv2 ; maybe something else.
would be great !

trsvchn:
@vfdev-5 Yes, I am going to start from this:

- create gtihub actions
[...]

PR #1733, opened 2021-03-03
Added docker image test script

vfdev-5:
Addresses #1644

Description:
- Added docker image test python script

Check list:


trsvchn:
there is python package docker? Nice!)

Since, I've started from `subprocess.run` :)

vfdev-5:
Yes, I tried that from my side to see if this could work. @trsvchn if you'd like to review, feel free to comment out and give a feedback.

vfdev-5:
And sorry for overlapping !

trsvchn:
No problem, sorry, I was busy yesterday. and now I know about `docker` python so it worth it :)

vfdev-5:
@trsvchn do you have an idea why it can't find installed `docker` package ?

trsvchn:
@vfdev-5 Oh, I think I know what is happening

trsvchn:
` $AGENT_TOOLSDIRECTORY` is `/opt/hostedtoolcache` 

and

```bash
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.7.10/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.7.10/x64/lib
```

Thus:

```bash
sudo rm -rf "/usr/local/share/boost"
sudo rm -rf "$AGENT_TOOLSDIRECTORY"
```
Removes python libs

vfdev-5:
Oh, I see. Should we just reinstall it after building ?

trsvchn:
But, wait, pyaml works fine

vfdev-5:
pyyaml is OK as it is used only before the suppression

trsvchn:
Yes, for base images
but we reuse it in later steps

trsvchn:
Let's try 

```
sudo rm -rf "/usr/local/share/boost"
sudo rm -rf "$AGENT_TOOLSDIRECTORY"
```

Just after the checkout, and before Python install actiob

vfdev-5:
@trsvchn seems like it works !

trsvchn:
Great!

vfdev-5:
@trsvchn I'm struggling with `test_image.py` script and the way how to expose details of the error when using docker python api. For example, here https://github.com/pytorch/ignite/pull/1733/checks?check_run_id=2046167278#step:6:605
import command is failing but we have no idea what is the error message. 
Basically, the following code is missing something.
```python
def run_python_cmd(cmd):
    try:
        out = client.containers.run(args.image, f"python -c '{cmd}'", auto_remove=True, stderr=True,)
        assert isinstance(out, bytes), type(out)
        out = out.decode("utf-8").strip()
    except docker.errors.ContainerError as e:
        raise RuntimeError(e)
    return out
```
Any ideas how to improve it ?

trsvchn:
@vfdev-5 sure! I'll test `test_image.py` in my sandbox

vfdev-5:
I'm thinking to do something like 
```python
def run_python_cmd(cmd):

    try_except_cmd = f"""

def main():
    {cmd}

try:
    main()
except Exception as e:
    import traceback
    print(traceback.format_exc())    
    """
    try:
        out = client.containers.run(args.image, f"python -c '{try_except_cmd}'", auto_remove=True, stderr=True,)
        assert isinstance(out, bytes), type(out)
        out = out.decode("utf-8").strip()

        if any([k in out.lower() for k in ["error", "exception"]]):
            raise RuntimeError(out)

    except docker.errors.ContainerError as e:
        raise RuntimeError(e)
    return out
```

but this looks like a hack...

trsvchn:
@vfdev-5 Yes!, I think the problem is in that we are raising exceptions, since this generates exit code 1 and it forces run method to raise exception:

```bash
$ python -c 'raise RuntimeError'
  File "<string>", line 1, in <module>
RuntimeError

$ echo $?
1
```

>         Raises:
>             :py:class:`docker.errors.ContainerError`
>                 If the container exits with a non-zero exit code and

we should handle this quietly

vfdev-5:
That's right. But what I'd like to have is to execute tests like `import cv2` and ensure that it passes and if not say that explicitly : "ModuleNotFoundError: No module named 'c2'".
I was just thinking if there is a proper way to get error message without 2 levels of exception trackings

trsvchn:
What about to try `detach=True`, with this param, run method returns logs:

>            If ``detach`` is ``True``, a :py:class:`Container` object is instead.

```python
container = client.containers.run('bfirsh/reticulate-splines', True)
container.logs() spline 1...\\nReticulating spline 2...\\n'
```

vfdev-5:
detach or not detach this is the question :)
Well, actually, the context is that I do not understand why there is no cv2 module in msdp-apex-vision image and I can not rerpo that locally...

trsvchn:
cv or not 2cv ... or cv2

trsvchn:
It seems to work fine

vfdev-5:
Let's see if we can finally update our docker images

PR #1724, opened 2021-03-01
GH Action for docker builds

trsvchn:
Related to #1644 and #1721

Description:

Check list:

- [ ] New tests are added (if a new feature is added)
- [ ] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #1736, opened 2021-03-04
Fix docker builds order

trsvchn:
The PR Fixes docker builds order as discussed here #1724 

Description:

Current setup looks like this:

```yml
jobs:
  setup:
    ...
  ...
  build-main:
    steps:
      - build base
        ...
      - build vision
        ...
      - build nlp
        ...
  ...
 ```

- 5 parallel build jobs + setup job
- each of 5 jobs builds 3 images: base + vision + nlp
- no artifacts
 
Check list:

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -21,8 +21,6 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Get a list of modified files
         run: echo "modified=${{ steps.files.outputs.files_created }} ${{ steps.files.outputs.files_updated }}" >> $GITHUB_ENV
-      - name: Debug Step
-        run: echo "${{ env.modified }}"
       - name: Set outputs
         id: set-modified
         run: echo "::set-output name=modified::${{ env.modified }}"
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: .github/workflows/docker-build.yml [hunk 1]
```diff
@@ -9,27 +9,20 @@ on:
...
       - name: Changed Files Exporter
         id: files
         uses: futuratrepadeira/changed-files@v3.3.0
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-
       - name: Get a list of modified files
         run: echo "modified=${{ steps.files.outputs.files_created }} ${{ steps.files.outputs.files_updated }}" >> $GITHUB_ENV
-
       - name: Debug Step
```
vfdev-5, line 24: I wonder if we need this "debug step" in the final version ?

vfdev-5 (approved):
Looks great, thanks @trsvchn !
I have a nit comment about debug step...

.github/workflows/docker-build.yml
trsvchn: No, I think no

PR #1742, opened 2021-03-04
Refactor release workflow

trsvchn:
Addresses https://github.com/pytorch/ignite/issues/1644#issuecomment-788071082

The aim of this PR is to merge two release jobs into one sequential, consisting of the two steps:

on release:
1. build and publish binaries
2. trigger circle ci to build and publish docker images

on push:
1. skip build and publish binaries steps
2. trigger circle ci to build and publish docker images


vfdev-5:
Thanks for the PR @trsvchn ! That could solve the issue we talked previously, but I fear a bit when we trigger this on other events... As this part is rather critical, I'd replicate the the lines here and would keep previous file as well... What do you think ?

trsvchn:
@vfdev-5 Exactly, I was thinking about this design:

So we remove release trigger for `docker-publish.yml` , only push

```
docker-publish.yml

on: 
  push:
    ...

jobs:
  build-publish-docker-job:
    ...
```

Then duplicate job from `docker-publish` as dependent from  `build-publish-binaries`

```
releases.yml

on: 
  release:
    ...

jobs:
  build-publish-binaries:
    ...

  build-publish-docker-job:
    needs:  build-publish-binaries:
    ...

```

So, is it okay to have duplicated jobs?

vfdev-5:
Well, I heard that we could somehow import yaml in yaml. If it is doable we can refactor that otherwise just duplicate. I do not want to touch already working action that is almost untestable here...

trsvchn:
Importing sounds fantastic, but I'm afraid this feature is on available right now:

https://github.com/actions/starter-workflows/issues/245

vfdev-5:
Well, anyway I think it is not a big deal to replicate 10 lines of yaml. We can also leave a comment about that and keep going :)

trsvchn:
Yes, agreed! DRY principle is all about coding, not configuration :D

vfdev-5:
I think DRY is a good principle but heard also opposite suggestions in order to avoid bad design etc... (https://programmingisterrible.com/post/176657481103/repeat-yourself-do-more-than-one-thing-and)

trsvchn:
@vfdev-5 thanks for the link! 

BTW, Romans used to say *Repetitio est mater studiorum* :)

vfdev-5 (approved):
Looks good ! Thanks @trsvchn !

PR #1743, opened 2021-03-04
Updated versions

vfdev-5:
Description:

Updated versions:
- pth to 1.8.0
- added python 3.9
- changed way to install torch xla

Check list:

### DIFF
--- a/.github/workflows/pytorch-version-tests.yml
+++ b/.github/workflows/pytorch-version-tests.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.6, 3.7, 3.8]
-        pytorch-version: [1.7.0, 1.6.0, 1.5.1, 1.4.0, 1.3.1]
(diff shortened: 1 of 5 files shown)


.github/workflows/pytorch-version-tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.6, 3.7, 3.8]
-        pytorch-version: [1.7.0, 1.6.0, 1.5.1, 1.4.0, 1.3.1]
+        pytorch-version: [1.7.1, 1.6.0, 1.5.1, 1.4.0, 1.3.1]
```
ydcjeff, line 17: Maybe add 1.8.0?

vfdev-5: Usually, we add only previous versions here as the latest stable one is running in unit-tests.yml...

PR #1745, opened 2021-03-05
[Feature] Pull Request Auto Labeler

trsvchn:
This PR adds automatic issue and pr labelers

Issue labeler adds labels based on the body content of the issue.

PR labeler adds labels s based on the paths that are modified in the PR.

Description:

Here is demo from my fork:

https://github.com/trsvchn/ignite/issues/13
https://github.com/trsvchn/ignite/pull/14

All labels are defined in config files in `.github/*-labeler.yml`

Some of them are disabled:

https://github.com/trsvchn/ignite/blob/b55ce8331316cba55b2d5cb12fe9c26d6182b067/.github/pr-labeler.yml#L9-L20

Since we need to create them in repo first, action doesn't create them for us just uses already available.

Check list:


vfdev-5:
Looks very cool @trsvchn ! Thanks for the FR !

trsvchn:
@vfdev-5 Please see the following experiments:

https://github.com/trsvchn/ignite/issues/13#issuecomment-791520336

trsvchn:
Issue labeler removes manually added labels

vfdev-5:
Yes, looks like that...

trsvchn:
As possible solution, we can use it only on issue opening, so can label it once

trsvchn:
Or create a subset of labels just for labeler: a kind of:
- "triage: docs"
- "triage: docker"
- etc

trsvchn:
@vfdev-5 

I've removed triggering on issue editing for Issue Labeler, since we are actively using labels on issues,

Triggering for PR Labeler left unchanged, since I've noticed we are not using labels for PR at all, so it is not going to affect any labeling system.

sdesrozis:
Beyond technical issues that you are facing, I really enjoy this FR :)

trsvchn:
@sdesrozis Thanks!) Yes, I'm looking for solution where we can "coexist" with auto-labeling system, since labelers don't have a clue who added this label etc, and there is no way to inform them: "I know no docs are modified, but this PR is logically connected to docs"

.github/issue-labeler.yml
```diff
@@ -0,0 +1,11 @@
+# Add/remove 'bug' label if issue contains the words 'Bug' or 'bug'
+bug:
+  - '(Bug|bug)'
```
vfdev-5, line 3: Thinking again about this auto-labelling. We already set labels if user selects right template. For bug I omitted pushing directly bug label as it can be something else, so it is up to maintainers to add "bug" after checking...

Maybe, we can keep labels for PRs only without any issue labelers as I'm not sure about our usage for that...
What do you think ?

trsvchn: @vfdev-5 Ok, now I see, I used bug template for testing and didn't notice that, for example, docs and question labels we add automatically on issue creating,

yes, I think in current configuration issue-labeler is useless for our situation, let's keep pr-labeler, and wait for more smarter issue-labeler :)

vfdev-5 (approved):
Looks good ! Thanks @trsvchn

PR #1755, opened 2021-03-07
 Refactor `test_accuracy` to use `pytest.mark.parametrize`

### LINKED ISSUE #1469: Parametrize metric tests

Moh-Yakoub:
Addresses #1469 (vfdev-5: we should not use "Fixes" as it closes the issue which larger than a single PR).

Description:

I noticed that the test test_accuracy was repeating the test cases and could have utilized the pytest parameterize decorator. every test method here was repeating the same code for measure accuracy. I utilized parameterize to remove that repeated code.

Check list:

### DIFF
--- a/tests/ignite/metrics/test_accuracy.py
+++ b/tests/ignite/metrics/test_accuracy.py
@@ -91,145 +91,45 @@ def test_binary_input_N():
         _test()
 
 
-def test_binary_input_NL():
-    # Binary accuracy on input of shape (N, L)
-    def _test():
+@pytest.mark.parametrize(
+    "y_pred,y,n_iters",
+    [
+        # Binary accuracy on input of shape (N, L)
+        (torch.randint(0, 2, size=(10, 5)).long(), torch.randint(0, 2, size=(10, 5)).long(), 1),
+        (torch.randint(0, 2, size=(10, 1, 5)).long(), torch.randint(0, 2, size=(10, 1, 5)).long(), 1),
+        (torch.randint(0, 2, size=(100, 8)).long(), torch.randint(0, 2, size=(100, 8)).long(), 16),
+        # Binary accuracy on input of shape (N, H, W, ...)
+        (torch.randint(0, 2, size=(4, 1, 12, 10)).long(), torch.randint(0, 2, size=(4, 1, 12, 10)).long(), 1),
+        (torch.randint(0, 2, size=(4, 1, 12, 10)).long(), torch.randint(0, 2, size=(4, 1, 12, 10)).long(), 1),
+        (torch.randint(0, 2, size=(100, 8, 8)).long(), torch.randint(0, 2, size=(100, 8, 8)).long(), 16),
+        # Binary accuracy on input of shape (N, 1, ...) - Multiclass input
+        (torch.randint(0, 2, size=(4, 1)).long(), torch.randint(0, 2, size=(4,)).long(), 1),
+        (torch.randint(0, 2, size=(4, 1, 12)).long(), torch.randint(0, 2, size=(4, 12)).long(), 1),
+        (torch.randint(0, 2, size=(100, 1, 8, 8)).long(), torch.randint(0, 2, size=(100, 8, 8)).long(), 16),
+        # Multiclass input data of shape (N, ) and (N, C)
+    ],
+)
+def test_binary_input(y_pred, y, n_iters):
+    def _test(y_pred, y, n_iters):
         acc = Accuracy()
 
-        y_pred = torch.randint(0, 2, size=(10, 5)).long()
-        y = torch.randint(0, 2, size=(10, 5)).long()
         acc.update((y_pred, y))
         np_y = y.numpy().ravel()
         np_y_pred = y_pred.numpy().ravel()
-        assert acc._type == "binary"
-        assert isinstance(acc.compute(), float)
-        assert accuracy_score(np_y, np_y_pred) == pytest.approx(acc.compute())
 
-        acc.reset()
-        y_pred = torch.randint(0, 2, size=(10, 1, 5)).long()
-        y = torch.randint(0, 2, size=(10, 1, 5)).long()
-        acc.update((y_pred, y))
-        np_y = y.numpy().ravel()
-        np_y_pred = y_pred.numpy().ravel()
-        assert acc._type == "binary"
-        assert isinstance(acc.compute(), float)
-        assert accuracy_score(np_y, np_y_pred) == pytest.approx(acc.compute())
-
-        # Batched Updates
-        acc.reset()
-        y_pred = torch.randint(0, 2, size=(100, 8)).long()
-        y = torch.randint(0, 2, size=(100, 8)).long()
-
-        n_iters = 16
-        batch_size = y.shape[0] // n_iters + 1
-
-        for i in range(n_iters):
-            idx = i * batch_size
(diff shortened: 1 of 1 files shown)


tests/ignite/metrics/test_accuracy.py
```diff
@@ -91,145 +91,45 @@ def _test():
...
-
-
-def test_binary_input_NHW():
-    # Binary accuracy on input of shape (N, H, W, ...)
-    def _test():
+@pytest.mark.parametrize(
+    "y_pred,y,n_iters",
+    [
+        # Binary accuracy on input of shape (N, L)
+        (torch.randint(0, 2, size=(10, 5)).long(), torch.randint(0, 2, size=(10, 5)).long(), 1),
```
vfdev-5, line 98: Nevertheless, test code is now more compact it does not do the same thing as previously.
We explicitly call `reset` to check that the same `acc` instance can be reused.

tests/ignite/metrics/test_accuracy.py
```diff
@@ -248,176 +148,49 @@ def test_multiclass_wrong_inputs():
...
-        np_y_pred = y_pred.numpy().argmax(axis=1).ravel()
-        np_y = y.numpy().ravel()
-        assert acc._type == "multiclass"
-        assert isinstance(acc.compute(), float)
-        assert accuracy_score(np_y, np_y_pred) == pytest.approx(acc.compute())
-
         acc.reset()
-        y_pred = torch.rand(4, 5, 10, 12, 8)
-        y = torch.randint(0, 5, size=(4, 10, 12, 8)).long()
+        y_pred = y_pred
```
vfdev-5, line 176: This is useless, right.

vfdev-5 (commented):
Thanks for the PR @Moh-Yakoub !
However, I think we have to be really careful with refactoring tests like that...
I left a comment with my concerns about that...

tests/ignite/metrics/test_accuracy.py
Moh-Yakoub: I just checked the accuracy method
```      def reset(self) -> None:
        self._num_correct = torch.tensor(0, device=self._device)
        self._num_examples = 0
        super(Accuracy, self).reset()``
So what it does is that it just resets the two variables that were updated in the `update` method.
When we use the parametrize method it creates a new instance which have the same effect. so In that case it won't affect the behavior of the test/instance. What do you think?

Comment (Moh-Yakoub) [reply tests/ignite/metrics/test_accuracy.py:98]: That's a good point. I think for this case - even for the accuracy one - we can have dedicated test for the reset method i.e. `test_reset` that will make two consecutive calls with reset in between to make sure we actually reset as expected. I feel it's a bit of too much overhead when we do that in every single test case we are having. I am up for both solutions though

Comment (vfdev-5) [reply tests/ignite/metrics/test_accuracy.py:98]: Probably, depending on the metric, `test_reset` can have as much case as other tests... 
When we explicitly call `reset`, it emulates the usage of the metric inside evaluator.

File: tests/ignite/metrics/test_accuracy.py [hunk 3]
```diff
@@ -91,145 +91,45 @@ def _test():
...
+        (torch.randint(0, 2, size=(100, 8, 8)).long(), torch.randint(0, 2, size=(100, 8, 8)).long(), 16),
+        # Binary accuracy on input of shape (N, 1, ...) - Multiclass input
+        (torch.randint(0, 2, size=(4, 1)).long(), torch.randint(0, 2, size=(4,)).long(), 1),
+        (torch.randint(0, 2, size=(4, 1, 12)).long(), torch.randint(0, 2, size=(4, 12)).long(), 1),
+        (torch.randint(0, 2, size=(100, 1, 8, 8)).long(), torch.randint(0, 2, size=(100, 8, 8)).long(), 16),
+        # Multiclass input data of shape (N, ) and (N, C)
+    ],
+)
+def test_binary_input(y_pred, y, n_iters):
+    def _test(y_pred, y, n_iters):
```
vfdev-5, line 113: Let's just create single instance here and pass it inside `_test` and call `reset`

vfdev-5 (approved):
LGTM !

tests/ignite/metrics/test_accuracy.py
```diff
@@ -91,145 +91,44 @@ def _test():
...
+        (torch.randint(0, 2, size=(4, 1, 12, 10)).long(), torch.randint(0, 2, size=(4, 1, 12, 10)).long(), 1),
+        (torch.randint(0, 2, size=(4, 1, 12, 10)).long(), torch.randint(0, 2, size=(4, 1, 12, 10)).long(), 1),
+        (torch.randint(0, 2, size=(100, 8, 8)).long(), torch.randint(0, 2, size=(100, 8, 8)).long(), 16),
+        # Binary accuracy on input of shape (N, 1, ...) - Multiclass input
+        (torch.randint(0, 2, size=(4, 1)).long(), torch.randint(0, 2, size=(4,)).long(), 1),
+        (torch.randint(0, 2, size=(4, 1, 12)).long(), torch.randint(0, 2, size=(4, 12)).long(), 1),
+        (torch.randint(0, 2, size=(100, 1, 8, 8)).long(), torch.randint(0, 2, size=(100, 8, 8)).long(), 16),
+        # Multiclass input data of shape (N, ) and (N, C)
+    ]
+    for y_pred, y, n_iters in test_cases:
```
KickItLikeShika, line 128: This is not the best way to generate test as the `test_cases` have been generated already, Please have a look [here](https://github.com/pytorch/ignite/blob/16b3a68b164aa140ee0e78161f8d2d183dea600e/tests/ignite/contrib/metrics/test_roc_auc.py#L90)

tests/ignite/metrics/test_accuracy.py
```diff
@@ -91,145 +91,44 @@ def _test():
...
+        # check multiple random inputs as random exact occurencies are rare
```
KickItLikeShika, line 129: This statement is not correct not, it should be correct after updating the test generation.

tests/ignite/metrics/test_accuracy.py
```diff
@@ -248,176 +147,45 @@ def test_multiclass_wrong_inputs():
...
+        (torch.rand(100, 5), torch.randint(0, 5, size=(100,)).long(), 16),
+        # Multiclass input data of shape (N, L) and (N, C, L)
+        (torch.rand(10, 4, 5), torch.randint(0, 4, size=(10, 5)).long(), 1),
+        (torch.rand(4, 10, 5), torch.randint(0, 10, size=(4, 5)).long(), 1),
+        (torch.rand(100, 9, 7), torch.randint(0, 9, size=(100, 7)).long(), 16),
+        # Multiclass input data of shape (N, H, W, ...) and (N, C, H, W, ...)
+        (torch.rand(4, 5, 12, 10), torch.randint(0, 5, size=(4, 12, 10)).long(), 1),
+        (torch.rand(100, 3, 8, 8), torch.randint(0, 3, size=(100, 8, 8)).long(), 16),
+    ]
+    for y_pred, y, batch_size in test_cases:
```
KickItLikeShika, line 185: The same here

vfdev-5:
@Moh-Yakoub I asked @KickItLikeShika to comment out here about the random data generation. Please, send a follow up PR with a fix for that. Thanks !

PR #1758, opened 2021-03-08
update epoch metrics to use collections

### LINKED ISSUE #1757: More EpochMetric's compute_fn output types
## 🚀 Feature

Following https://github.com/pytorch/ignite/pull/1700#issuecomment-787151203 we would like to improve `EpochMetric` by supporting more output types returned by `compute_fn`:
- EpochMetric should work with compute_fn outputs: scalars, tensors, tuple/list/mapping of tensors vs currently only scalars

A code snippet on possible implementation: https://github.com/pytorch/ignite/pull/1700#issuecomment-790413236
(It is not tested and maybe wrong, just an idea of how to do it).

It would be also helpful to raise a more user friendly error message if output type is not supported.

We need to update the docs as well.

Moh-Yakoub:
Fixes #1757

Description: As title. The main idea is to allow the epochmetrics to use collections of tensors

**This is  a WIP to gather feedback about the approach, once it's approved I will clean up the implementation and implement more test cases**

Check list:


ignite/metrics/epoch_metric.py
```diff
@@ -113,7 +116,7 @@ def update(self, output: Tuple[torch.Tensor, torch.Tensor]) -> None:
             except Exception as e:
                 warnings.warn(f"Probably, there can be a problem with `compute_fn`:\n {e}.", EpochMetricWarning)
 
-    def compute(self) -> float:
+    def compute(self) -> Union[int, float, Sequence, Mapping]:
```
vfdev-5, line 119: Let's say we support only sequence or mapping of tensors:
```suggestion
    def compute(self) -> Union[int, float, Sequence[torch.Tensor], Mapping[str, torch.Tensor]]:
```

ignite/metrics/epoch_metric.py
```diff
@@ -133,9 +136,15 @@ def compute(self) -> float:
             # Run compute_fn on zero rank only
             result = self.compute_fn(_prediction_tensor, _target_tensor)
 
+        # compute_fn outputs: scalars, tensors, tuple/list/mapping of tensors.
+        if not is_scalar_or_collection_of_tensor(result):
```
vfdev-5, line 140: I think there is no need to introduce a public `is_scalar_or_collection_of_tensor` for that. Let's create a small private function here.

vfdev-5 (commented):
Thanks for the PR @Moh-Yakoub ! The approach is good, let's simplify few things and it will be ok

ignite/metrics/epoch_metric.py
Moh-Yakoub: The problem with specifying the type of sequence/mapping is that it will break `mypy` as the `apply_to_type` returns a generic Sequence. 
I am not sure if we really need to update how `apply_to_type`for this usecase.

vfdev-5: Can't we use `cast` in this case somehow ?

Moh-Yakoub: Another thing I noticed. There are many test failures because the `precision_recall_curve` use the sklearn compute function which returns a response that doesn't fit with the new validation we setup in this PR i.e. it's not a scalar or a collection of tensors.
Shall I update the implementation of the `precision_recall_curve` to fix that validation? @vfdev-5

vfdev-5: @Moh-Yakoub it is related to https://github.com/pytorch/ignite/pull/1700 where we tried to fix that. I think what could be done is to comment out failing tests here and uncomment and fix failling tests in #1700 once this PR is merged.
What do you think ?
cc @KickItLikeShika

Moh-Yakoub: That makes sense. I will add a todo statement for that as well

vfdev-5: @Moh-Yakoub so, no way to cast the result to `Sequence[torch.Tensor], Mapping[str, torch.Tensor]` here ?

Moh-Yakoub: @vfdev-5  I've added a type cast, I am not sure if there's any better way to do that cast to allow the `Sequence[torch.Tensor], Mapping[str, torch.Tensor]` return type without annoying mypy.

If you are fine with current approach I will add couple of tests and that PR should be ready.

vfdev-5: Thanks for the working on that ! I think the best thing we can do here is to add "typing ignore" statement on the output of `apply_to_typefor` and keep the code simple.

ignite/utils.py
```diff
@@ -56,7 +56,7 @@ def apply_to_type(
     """
     if isinstance(x, input_type):
         return func(x)
-    if isinstance(x, (str, bytes)):
+    if isinstance(x, (str, bytes, float, int)):
```
vfdev-5, line 59: I'm not sure about this change. I'd prefer to keep it as it was unless this is a hard requirement.

ignite/metrics/epoch_metric.py
vfdev-5: @Moh-Yakoub I think the issue is that `Sequence` and `Mapping` are from collections and not typing. Please fix it and it will be good to go (think also to update PR to the master).

Moh-Yakoub:
@vfdev-5 I've noticed that specifying the output type of the sequence/mapping caused all tests to fail because of the following 
`TypeError: 'ABCMeta' object is not subscriptable`. any idea what maybe causing this?

I am not able to reproduce locally, all my tests pass locally

vfdev-5:
> @vfdev-5 I've noticed that specifying the output type of the sequence/mapping caused all tests to fail because of the following
> `TypeError: 'ABCMeta' object is not subscriptable`. any idea what maybe causing this?
> 
> I am not able to reproduce locally, all my tests pass locally

Maybe, this is the answer : https://github.com/pytorch/ignite/pull/1758#discussion_r593903851 
otherwise, please try to google that.

vfdev-5:
@Moh-Yakoub can we fix this one in priority, please ?

Moh-Yakoub:
> @Moh-Yakoub can we fix this one in priority, please ?

sure I am trying to resolve the ABC issue

ignite/metrics/epoch_metric.py
```diff
@@ -133,12 +138,35 @@ def compute(self) -> float:
...
+    """
+    if isinstance(x, (int, float, torch.Tensor)):
+        return True
+    if isinstance(x, Sequence):
+        return all([isinstance(item, torch.Tensor) for item in x])
+    if isinstance(x, Mapping):
+        return all([isinstance(item, torch.Tensor) for item in x.values()])
+    if isinstance(x, tuple) and hasattr(x, "_fields"):
+        return all([isinstance(item, torch.Tensor) for item in getattr(x, "_field")])
+    return False
```
vfdev-5, line 168: Let's already improve codecov here.

ignite/metrics/epoch_metric.py
```diff
@@ -133,12 +138,35 @@ def compute(self) -> float:
...
+        # compute_fn outputs: scalars, tensors, tuple/list/mapping of tensors.
+        if not _is_scalar_or_collection_of_tensor(result):
+            raise TypeError(
+                "output not supported: compute_fn should return scalar, tensor, tuple/list/mapping of tensors"
+            )
+
         if ws > 1:
             # broadcast result to all processes
-            result = cast(float, idist.broadcast(result, src=0))
+            return apply_to_type(result, torch.Tensor, partial(idist.broadcast, src=0))  # type: ignore[return-value]
```
vfdev-5, line 149: Shouldn't it be
```suggestion
            return apply_to_type(result, (torch.Tensor, int, float), partial(idist.broadcast, src=0))  # type: ignore[return-value]
```

Moh-Yakoub: Yes, my bad.

I've updated it, but needed to update the ignore type. hopefully now all tests will pass

ignite/metrics/epoch_metric.py
```diff
@@ -133,12 +138,39 @@ def compute(self) -> float:
...
+            raise TypeError(
+                "output not supported: compute_fn should return scalar, tensor, tuple/list/mapping of tensors"
+            )
+
         if ws > 1:
             # broadcast result to all processes
-            result = cast(float, idist.broadcast(result, src=0))
+            return apply_to_type(  # type: ignore
+                result,
+                (torch.Tensor, float, int),  # type: ignore
```
vfdev-5, line 151: Yeah, that's true, @Moh-Yakoub please, update `apply_to_type`  and remove type ignore from here. Thanks !

Moh-Yakoub: I thought we agreed it would be simpler to keep the method simple and just apply ignore for the type https://github.com/pytorch/ignite/pull/1758#discussion_r593812175

vfdev-5: I speaking about the second `# type: ignore`.

Moh-Yakoub:
@vfdev-5  I am getting a lot of ` RuntimeError: connect() timed out.` test failures and another one `Engine run is terminating due to exception: Mismatched data types: One rank had type int64, but another rank had type float32.` on Horovod. Do you suspect it's related to the broadcasting code?

vfdev-5:
@Moh-Yakoub yes, failure is real and actually we are wrong with the implementation when using broadcast on tensors.

The issue is with data types:
```
        result = 0.0  # <----- all procs has result as float scalar
        if idist.get_rank() == 0:
            result = self.compute_fn(_prediction_tensor, _target_tensor)  # <---- now result is tensor on rank 0

        # compute_fn outputs: scalars, tensors, tuple/list/mapping of tensors.
        if not _is_scalar_or_collection_of_tensor(result):
            raise TypeError(
                "output not supported: compute_fn should return scalar, tensor, tuple/list/mapping of tensors"
            )

        if ws > 1:
            # broadcast result to all processes
            # !!!! BUG : WE TRY TO BROADCAST TENSOR TO A SCALAR PLACEHOLDERS FOR OTHER PROCS...
            return apply_to_type(  # type: ignore
                result, (torch.Tensor, float, int), partial(idist.broadcast, src=0),
            )   
```

I have to think about that...

sdesrozis:
> @Moh-Yakoub yes, failure is real and actually we are wrong with the implementation when using broadcast on tensors.

I understand the need but I’m quite surprised. It means that every process would not know the handled type. It’s weird. Maybe it is because I’m familiar with strongly typed languages.

Could the return type of `self.compute_fn` be a union of possible types ?

vfdev-5:
> > @Moh-Yakoub yes, failure is real and actually we are wrong with the implementation when using broadcast on tensors.
> 
> I understand the need but I’m quite surprised. It means that every process would not know the handled type. It looks weird to me. Maybe it is because I’m familiar with strongly typed languages.
> 
> Could the return type of `self.compute_fn` be a union of possible types ?

@sdesrozis it is a union of known types: a scalar or a sequence/mapping/tuple of tensors.

ignite/metrics/epoch_metric.py
```diff
@@ -133,12 +138,37 @@ def compute(self) -> float:
...
+        # compute_fn outputs: scalars, tensors, tuple/list/mapping of tensors.
+        if not _is_scalar_or_collection_of_tensor(result):
+            raise TypeError(
+                "output not supported: compute_fn should return scalar, tensor, tuple/list/mapping of tensors"
+            )
+
         if ws > 1:
             # broadcast result to all processes
-            result = cast(float, idist.broadcast(result, src=0))
+            return apply_to_type(  # type: ignore
```
KickItLikeShika, line 149: I worked on this before, the Runtime errors is related to the broadcasting part. If you tried to print out anything while trying to handle tuple or list manually you will find out that it broadcasts the first tensor then it fails. I think what you need to figure out some way to broadcast more than just one tensor(tuples, lists, etc...) for solving this issue, I think then you will solve this issue. Good Luck!

vfdev-5:
@Moh-Yakoub I merged this PR #1839 and it should unblock this PR. So, we can now write safely like that :
```python
        result = None
        if idist.get_rank() == 0:
            result = self.compute_fn(_prediction_tensor, _target_tensor)

        # compute_fn outputs: scalars, tensors, tuple/list/mapping of tensors.
        if not _is_scalar_or_collection_of_tensor(result):
            raise TypeError(
                "output not supported: compute_fn should return scalar, tensor, tuple/list/mapping of tensors"
            )

        if ws > 1:
            # broadcast result to all processes
            return apply_to_type(  # type: ignore
                result, (torch.Tensor, float, int), partial(idist.broadcast, src=0, safe_mode=True),
            )
```

Moh-Yakoub:
@vfdev-5 Thanks a lot for the info and sorry for the late reply. I will work on this now.

vfdev-5 (commented):
@Moh-Yakoub I saw you updated the code according to https://github.com/pytorch/ignite/pull/1758#issuecomment-823659571 but now it still does not use `safe_mode` to broadcast... Please, update the PR and remove the comments. Thanks

Moh-Yakoub:
@vfdev-5  I've already removed the comments. related to broadcasts

I'm already using 

``` 
return apply_to_type(  # type: ignore
                result, (torch.Tensor, float, int), partial(idist.broadcast, src=0, safe_mode=True),
```


So the safe_mode is passed. Is there anything else to pass it to the broadcast method

vfdev-5:
@Moh-Yakoub any ideas why CI is still failing ?

tests/ignite/contrib/metrics/test_roc_curve.py
```diff
@@ -22,7 +22,8 @@ def test_no_sklearn(mock_no_sklearn):
         RocCurve()
 
 
-def test_roc_curve():
+# TODO uncomment those once #1700 is merge
```
vfdev-5, line 25: please, remove all these comments !

tests/ignite/contrib/metrics/test_average_precision.py
```diff
@@ -287,7 +287,7 @@ def test_distrib_gpu(distributed_context_single_node_nccl):
 
 @pytest.mark.distributed
 @pytest.mark.skipif(not idist.has_native_dist_support, reason="Skip if no native dist support")
-def test_distrib_cpu(distributed_context_single_node_gloo):
+def _test_distrib_cpu(distributed_context_single_node_gloo):
```
vfdev-5, line 290: This was a temp way to disable test, let's enable those tests once the CI is passing on epoch metric distrib tests.

tests/ignite/contrib/metrics/test_cohen_kappa.py
```diff
@@ -282,7 +282,7 @@ def test_distrib_gpu(distributed_context_single_node_nccl):
 
 @pytest.mark.distributed
 @pytest.mark.skipif(not idist.has_native_dist_support, reason="Skip if no native dist support")
-def test_distrib_cpu(distributed_context_single_node_gloo):
+def _test_distrib_cpu(distributed_context_single_node_gloo):
```
vfdev-5, line 285: Same here

tests/ignite/contrib/metrics/test_precision_recall_curve.py
```diff
@@ -25,7 +25,7 @@ def test_no_sklearn(mock_no_sklearn):
         pr_curve.compute()
 
 
-def test_precision_recall_curve():
+def _test_precision_recall_curve():
```
vfdev-5, line 28: Same here

ignite/metrics/epoch_metric.py
```diff
@@ -133,12 +138,37 @@ def compute(self) -> float:
             # Run compute_fn on zero rank only
             result = self.compute_fn(_prediction_tensor, _target_tensor)
 
+        # compute_fn outputs: scalars, tensors, tuple/list/mapping of tensors.
+        if not _is_scalar_or_collection_of_tensor(result):
+            raise TypeError(
+                "output not supported: compute_fn should return scalar, tensor, tuple/list/mapping of tensors"
```
vfdev-5, line 144:
```suggestion
                "output not supported: compute_fn should return scalar, tensor, tuple/list/mapping of tensors, "
                f"got {type(result)}"
```

ignite/metrics/epoch_metric.py
```diff
@@ -133,12 +138,37 @@ def compute(self) -> float:
             # Run compute_fn on zero rank only
             result = self.compute_fn(_prediction_tensor, _target_tensor)
 
+        # compute_fn outputs: scalars, tensors, tuple/list/mapping of tensors.
+        if not _is_scalar_or_collection_of_tensor(result):
```
vfdev-5, line 142: This check should be inside `if idist.get_rank() == 0:` I think

vfdev-5:
@Moh-Yakoub do you plan to finish wtih this PR ?

Moh-Yakoub:
> @Moh-Yakoub do you plan to finish with this PR ?

@vfdev-5 this PR have lagged behind now, Are there other PRs that have solved the issue, otherwise I can continue this one?

vfdev-5:
@Moh-Yakoub actually we figure out that this is not possible to do what we want here like that. I asked @KickItLikeShika to work on a simplified version of this features. I propose you to close this one and maybe tackle something else from the list: https://github.com/pytorch/ignite/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22

What do you think ?

Moh-Yakoub:
> @Moh-Yakoub actually we figured out that this is not possible to do what we want here like that. I asked @KickItLikeShika to work on a simplified version of this feature. I propose you to close this one and maybe tackle something else from the list: https://github.com/pytorch/ignite/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22
> 
> What do you think ?

Sure that sounds good, closing this now.

PR #1761, opened 2021-03-09
ci: add timeout-mintues in linkcheck job

ydcjeff:
Fixes https://github.com/pytorch/ignite/runs/2066997659?check_suite_focus=true

Description: `linkcheck` job is probably hanging now maybe due to some network requests issues.
At most it should run completely within a couple of minutes. Adding `timeout-minutes` 10 mins for the job to prevent long hanging CI check.

Check list:


vfdev-5 (approved):
Thanks @ydcjeff !

PR #1762, opened 2021-03-10
improve roc auc

KickItLikeShika:
Our 3rd PR to improve `ignite.contrib.metrics` module. I have updated the implementation of `roc_auc` (added the device arg) and updated the tests as we have done with [AveragePrecision](https://github.com/pytorch/ignite/blob/master/tests/ignite/contrib/metrics/test_average_precision.py).
Now we got `precision_recall_curve` and `roc_curve` left to be updated, but we will wait until `EpochMetrics` gets updated.

Check list:


ignite/contrib/metrics/roc_auc.py
```diff
@@ -30,7 +27,7 @@ def roc_auc_curve_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) ->
 class ROC_AUC(EpochMetric):
     """Computes Area Under the Receiver Operating Characteristic Curve (ROC AUC)
     accumulating predictions and the ground-truth during an epoch and applying
-    `sklearn.metrics.roc_auc_score <https://scikit-learn.org/stable/modules/generated/
+    `sklearn.metrics.roc_auc_score <http://scikit-learn.org/stable/modules/generated/
```
vfdev-5, line 30: Revert please

ignite/contrib/metrics/roc_auc.py
```diff
@@ -39,9 +36,10 @@ class ROC_AUC(EpochMetric):
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
         check_compute_fn: Default False. If True, `roc_curve
-            <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html#
+            <http://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html#
```
vfdev-5, line 39: Same here

tests/ignite/contrib/metrics/test_roc_auc.py
```diff
@@ -1,117 +1,129 @@
...
 
-    roc_auc_metric = ROC_AUC(output_transform=lambda x: (torch.sigmoid(x[1]), x[2]))
-    roc_auc_metric.attach(engine, "roc_auc")
+    test_cases = [
+        (torch.randint(0, 2, size=(10, 4)).long(), torch.randint(0, 2, size=(10, 4)).long(), 1),
+        (torch.randint(0, 2, size=(50, 7)).long(), torch.randint(0, 2, size=(50, 7)).long(), 1),
+        (torch.randint(0, 2, size=(100, 4)).long(), torch.randint(0, 2, size=(100, 4)).long(), 1),
+        (torch.randint(0, 2, size=(200, 6)).long(), torch.randint(0, 2, size=(200, 6)).long(), 1),
+        # updated batches
+        (torch.randint(0, 2, size=(10, 4)).long(), torch.randint(0, 2, size=(10, 4)).long(), 16),
```
vfdev-5, line 117: I think this test does not make sense as `batch_size = y.shape[0] // n_iters + 1 = 1`.

tests/ignite/contrib/metrics/test_roc_auc.py
```diff
@@ -1,117 +1,129 @@
...
-    np_y = np.zeros((size,), dtype=np.long)
-    np_y[size // 2 :] = 1
-    np.random.shuffle(np_y)
+def test_multilabel_input_N():
 
-    np_roc_auc = roc_auc_score(np_y, np_y_pred_sigmoid)
+    roc_auc = ROC_AUC()
 
-    batch_size = 10
+    def _test(y_pred, y, n_iters):
```
vfdev-5, line 98: What does this test check, it does not call `compute` and there is not consistency checking ? Am I missing something ?

vfdev-5 (commented):
There are few things to improve

tests/ignite/contrib/metrics/test_roc_auc.py
KickItLikeShika: That's right! that's how it should be:
```
for y_pred, y, batch_size in test_cases:
    ......

if batch_size > 1:
```
I think it should be updated now in `average_precision` and other tests, right?

tests/ignite/contrib/metrics/test_roc_auc.py
```diff
@@ -1,117 +1,129 @@
...
+        (torch.randint(0, 2, size=(50, 7)).long(), torch.randint(0, 2, size=(50, 7)).long(), 16),
+        (torch.randint(0, 2, size=(100, 4)).long(), torch.randint(0, 2, size=(100, 4)).long(), 16),
+        (torch.randint(0, 2, size=(200, 6)).long(), torch.randint(0, 2, size=(200, 6)).long(), 16),
+    ]
 
-    data = list(range(size // batch_size))
-    roc_auc = engine.run(data, max_epochs=1).metrics["roc_auc"]
-
-    assert roc_auc == np_roc_auc
+    for y_pred, y, n_iters in test_cases:
```
vfdev-5, line 123: I thinks we should swap the loops here and generate data inside the outer loop:
```python
for _ in range(10):
    test_cases = get_test_cases():
    for y_pred, y, n_iters in test_cases:
        _test(y_pred, y, n_iters)
```

tests/ignite/contrib/metrics/test_roc_auc.py
vfdev-5: Yes, something like that.
And yes, we should fix other as well if it is the same case. Even it would be better to code such test function once and reuse it instead of copying multiple times...

tests/ignite/contrib/metrics/test_roc_auc.py
KickItLikeShika: Updated all the functions the same way, what do you think?
```
def test_multilabel_input_N():

    roc_auc = ROC_AUC()

    def _test(y_pred, y, batch_size):
        ...........
        ...........
        assert isinstance(res, float)
        assert roc_auc_score(np_y, np_y_pred) == pytest.approx(res)

    def get_test_cases():
        test_cases = [
            (torch.randint(0, 2, size=(10, 4)).long(), torch.randint(0, 2, size=(10, 4)).long(), 1),
            (torch.randint(0, 2, size=(50, 7)).long(), torch.randint(0, 2, size=(50, 7)).long(), 1),
            (torch.randint(0, 2, size=(100, 4)).long(), torch.randint(0, 2, size=(100, 4)).long(), 1),
            (torch.randint(0, 2, size=(200, 6)).long(), torch.randint(0, 2, size=(200, 6)).long(), 1),
            # updated batches
            (torch.randint(0, 2, size=(10, 4)).long(), torch.randint(0, 2, size=(10, 4)).long(), 16),
            (torch.randint(0, 2, size=(50, 7)).long(), torch.randint(0, 2, size=(50, 7)).long(), 16),
            (torch.randint(0, 2, size=(100, 4)).long(), torch.randint(0, 2, size=(100, 4)).long(), 16),
            (torch.randint(0, 2, size=(200, 6)).long(), torch.randint(0, 2, size=(200, 6)).long(), 16),
        ]
        return test_cases

    for _ in range(10):
        test_cases = get_test_cases()
        for y_pred, y, batch_size in test_cases:
                _test(y_pred, y, batch_size)
```

vfdev-5: Looks good

tests/ignite/contrib/metrics/test_roc_auc.py
```diff
@@ -1,117 +1,139 @@
...
+        roc_auc.reset()
+        roc_auc.update((y_pred, y))
 
-    engine = Engine(update_fn)
+        np_y = y.numpy()
+        np_y_pred = y_pred.numpy()
 
-    roc_auc_metric = ROC_AUC(output_transform=lambda x: (x[1], x[2]))
-    roc_auc_metric.attach(engine, "roc_auc")
+        if n_iters > 1:
```
vfdev-5, line 66: I think we already discussed about this approach and changing it by `batch_size > 1`

KickItLikeShika: But this is in `test_binary_input_n()`, it makes sense! You will find it edited properly in `test_multilabel_input_n()`

vfdev-5:
@KickItLikeShika PR looks OK, could you please open an issue explaining that we need to refactor in a certain way our tests such we do not duplicate the code. Metrics of the same type could have similar test suites.

tests/ignite/contrib/metrics/test_roc_auc.py
vfdev-5: > But this is in test_binary_input_n(), it makes sense!

I'm not sure to understand why it makes sense here.

KickItLikeShika: We can do it the both ways no problem! And i can edit it if needed..

vfdev-5 (approved):
LGTM, thanks @KickItLikeShika !

Issue #1753, opened 2021-03-06
Reduce github repository size


vfdev-5:
Our `.git` starts to grow too fast, now. We discussed recently about that with @y0ast and repository size was ~500MB which is rather large. Now, it is 700MB:
```bash
/tmp/ignite# du -h -a -d 1
705M    ./.git
4.0K    ./codecov.yml
4.0K    ./.gitignore
28K     ./.circleci
4.0K    ./CODE_OF_CONDUCT.md
4.0K    ./pyproject.toml
916K    ./ignite
4.0K    ./requirements-dev.txt
4.0K    ./CITATION
92K     ./docker
1.2M    ./assets
28K     ./README.md
4.0K    ./LICENSE
12K     ./CONTRIBUTING.md
120K    ./.github
4.0K    ./setup.py
528K    ./docs
1.2M    ./tests
4.0K    ./.pre-commit-config.yaml
15M     ./examples
12K     ./conda.recipe
4.0K    ./setup.cfg
4.0K    ./mypy.ini
723M    .
```

That time I looked into ressources about reducing the size and it works locally, but it would be helpful to do that on GitHub as well.

Local compressing works:
```bash
/tmp/ignite# git gc --aggressive
/tmp/ignite# du -h -a -d 1
112M    ./.git
4.0K    ./codecov.yml
4.0K    ./.gitignore
28K     ./.circleci
4.0K    ./CODE_OF_CONDUCT.md
4.0K    ./pyproject.toml
916K    ./ignite
4.0K    ./requirements-dev.txt
4.0K    ./CITATION
92K     ./docker
1.2M    ./assets
28K     ./README.md
4.0K    ./LICENSE
12K     ./CONTRIBUTING.md
120K    ./.github
4.0K    ./setup.py
528K    ./docs
1.2M    ./tests
4.0K    ./.pre-commit-config.yaml
15M     ./examples
12K     ./conda.recipe
4.0K    ./setup.cfg
4.0K    ./mypy.ini
131M    .
```
but I have no idea how to apply that to `remote`...

cc @trsvchn @sdesrozis @fco-dv

ahmedo42:
This  [article](https://www.saschawillems.de/blog/2017/09/10/how-to-shrink-down-a-github-repository/) might be helpful, this will require closing open PR's and branches.

vfdev-5:
[...]

PR #1764, opened 2021-03-10
Keep only the latest commit for gh-pages

trsvchn:
Addressees  #1753 

Description:

Updated setup for GH pages action

Check list:


vfdev-5 (approved):
Thanks Taras, let's check if we can reduce the size now.

PR #1767, opened 2021-03-10
Improve idist to implement different init_methods

### LINKED ISSUE #1432: Improve idist to use Tcp and FileStore for init_method
## 🚀 Feature

`idist` supports default initilialization `env://`. It could be useful to support `tcp://` and `file://` modes (to tackle distributed computations on windows for instance).
 
Main issue is how to configure these modes ?

ahmedo42:
Fixes #1432 

Check list:


vfdev-5 (commented):
Thanks for the PR @ahmedo42 !
Seeing the code now, I think we have to rethink the way we set up process group...
Finally, I'm not very fan of doing the initialization using env var. Previously, the only way `env://` and thus env vars. But maybe, we have to generalize that to cover `tcp://master-addr:master-port` and `file://` ...

ahmedo42:
> I'm not very fan of doing the initialization using env var. Previously, the only way `env://` and thus env vars

@vfdev-5  Do you mean we should get rid of `setup_env_vars()` and `_setup_env_in_slurm()` , and maybe pass those params to `create_from_backend()` for example?

vfdev-5:
> > I'm not very fan of doing the initialization using env var. Previously, the only way `env://` and thus env vars
> 
> @vfdev-5 Do you mean we should get rid of `setup_env_vars()` and `_setup_env_in_slurm()` , and maybe pass those params to `create_from_backend()` for example?

I do not yet have a clear idea of that...

sdesrozis:
First we should write usage IMHO. Why not provide here simple use cases to show how the choice could be done by a user ? 

@ahmedo42 @vfdev-5 Thoughts ?

ahmedo42:
@sdesrozis  

These are some use cases I could think of.
Use case 1 :

```python
idist.initialize("gloo" ,init_method = 'file://test_folder/test_file' )

```

Use case 2;
```python
def training(local_rank, config, **kwargs):
    # ...
    print(idist.get_rank(), ": run with config:", config, "- backend=", idist.backend())
    # ...

master_addr = "master"
master_port = 5000
init_method = 'tcp://{master_addr}:{master_port}'
dist_config = {
    "nproc_per_node": 8,
    "nnodes": 2,
    "node_rank": args.node_rank,
    "master_addr": master_addr ,
    "master_port": master_port ,
     'init_method' : init_method
}

with idist.Parallel(backend="nccl", **dist_config) as parallel:
    parallel.run(training, config, a=1, b=2)

```

Use case 3:

``` python

def run(local_rank, *args, **kwargs):

    idist.set_local_rank(local_rank)
    # ...
    dist.init_process_group(**dist_info)
```
and if user doesn't provide an `init_method` we default to `env://`.

vfdev-5:
@sdesrozis , thanks for the idea about the use-cases. @ahmedo42 actually, we need to update the docs about this new feature.

tests/ignite/distributed/comp_models/test_native.py
```diff
@@ -35,6 +36,22 @@ def test__native_dist_model():
         _NativeDistModel.create_from_backend("abc")
 
 
+@pytest.mark.distributed
+@pytest.mark.skipif("WORLD_SIZE" in os.environ, reason="Skip if launched as multiproc")
+def test_native_init_method():
```
vfdev-5, line 41: @ahmedo42 are you sure this test is executed ?

ahmedo42: Yes that was an Issue I faced , and couldn't find  a way around it  , will change it in the new implementation

When running distributed tests using 
`export WORLD_SIZE=2 CUDA_VISIBLE_DEVICES="" pytest --dist=each --tx $WORLD_SIZE*popen//python=python tests/ -m distributed -vvv` it fails and throws the following error : 
 "PyTorch distributed configuration should define env variables" .
 I think it happens because we explicitly set world_size while other env variables are not defined..

vfdev-5: Yes, to run `export WORLD_SIZE=2 CUDA_VISIBLE_DEVICES="" pytest --dist=each --tx $WORLD_SIZE*popen//python=python tests/ -m distributed -vvv` we have to set up the test as here: https://github.com/pytorch/ignite/blob/e1ce313e7c9fcd6362477f7e2c1b0ce787db5630/tests/ignite/distributed/comp_models/test_native.py#L269-L270

with `local_rank, world_size` fixtures

vfdev-5: I'll commit here the way I'd like to implement it.
 and let you rework the docs and examples. We can also improve the tests together: should run also on windows and with spawn.

ahmedo42: Sounds great

sdesrozis:
I see only 2 ways :
* using environment variables (following the mind of `torch.launch.distributed`)
* setting explicitly an argument to `idist.Parallel` (`idist.initialize` and `idist.spawn`)

Maybe both could be provided. Or a third way :)

vfdev-5:
@ahmedo42 I updated the code as we discussed previously, could you and @sdesrozis review it please.

@ahmedo42 can you work not on docs and a follow-up PR for windows ?

tests/ignite/distributed/test_launcher.py
```diff
@@ -183,7 +183,7 @@ def test_idist_parallel_spawn_n_procs_native(init_method, backend, dirname):
     nproc_per_node = 4 if "gloo" == backend else torch.cuda.device_count()
     device = "cpu" if "gloo" == backend else "cuda"
     with idist.Parallel(backend=backend, nproc_per_node=nproc_per_node, init_method=init_method) as parallel:
-        parallel.run(_test_func, ws=nproc_per_node, device=device, backend=backend, true_init_method=init_method)
+        parallel.run(_test_func, ws=nproc_per_node, device=device)
```
vfdev-5, line 186: @ahmedo42 Is it correct ? `_test_func(index, ws, device, backend, true_init_method)` still requres `backend=backend, true_init_method=init_method`

ignite/distributed/utils.py
```diff
@@ -489,14 +493,20 @@ def train_fn(local_rank, a, b, c):
...
 
-            - "nccl" or "gloo" : timeout(=timedelta(minutes=30)).
+            - | "nccl" or "gloo" : ``timeout(=timedelta(minutes=30))``, ``init_method(=None)``,
+              | ``rank(=None)``, ``world_size(=None)``.
+              | By default, ``init_method`` will be "env://". See more info about parameters: `torch_init`_.
 
-            - "horovod" : comm(=None), more info: `hvd_init`_.
+            - | "horovod" : comm(=None), more info: `hvd_init`_.
 
+    .. _torch_init: https://horovod.readthedocs.io/en/latest/api.html#horovod.torch.init
```
ahmedo42, line 502: The two links are the same , is this the intention?

vfdev-5: Does documentation work if one of them removed ?

vfdev-5:
Horovod failure seems to be related: https://github.com/pytorch/ignite/pull/1767/checks?check_run_id=2115636257
I have to investigate it.

ignite/distributed/comp_models/native.py
```diff
@@ -46,40 +46,83 @@ def create_from_context() -> Optional["_NativeDistModel"]:
...
                 raise ValueError(f"Backend should be one of '{_NativeDistModel.available_backends}'")
 
             if dist.is_available() and dist.is_initialized():
                 raise RuntimeError("Can not create new distributed process group if default one is already initialized")
-            return _NativeDistModel(backend=backend, **kwargs)
 
-        def __init__(self, backend: Optional[str] = None, timeout: Optional[int] = None, **kwargs: Any) -> None:
+            if init_method is None:
+                if world_size is not None or rank is not None:
+                    raise ValueError("Arguments rank and world_size should be None if no init_method is provided")
```
vfdev-5, line 64: @ahmedo42 there also test to write to increase codecov according to the warnings here:
https://github.com/pytorch/ignite/pull/1767/files

sdesrozis:
It looks good !! I will go further asap.

ignite/distributed/utils.py
ahmedo42: No it doesn't , I'm just not sure that the first link should be pointing to horovod docs , shouldn't it point to pytorch distributed?

vfdev-5: Good remark, yes, it is a wrong copy-paste. It should point out to pytorch docs. Thanks !

ignite/distributed/launcher.py
```diff
@@ -170,10 +191,18 @@ def training(local_rank, config, **kwargs):
             (`nccl`, `gloo`). Mandatory argument if ``nnodes`` is specified and larger than one.
         master_port: optional argument, master node port for torch native backends
             (`nccl`, `gloo`). Mandatory argument if ``master_addr`` is specified.
+        init_method: optional argument to specify processing group initialization method for torch native
+            backends (`nccl`, `gloo`). Default, "env://". See more info: `torch_init`_.
         spawn_kwargs: kwargs to ``idist.spawn`` function.
 
+    .. _torch_init: https://horovod.readthedocs.io/en/latest/api.html#horovod.torch.init
```
vfdev-5, line 198: Here, the same thing

vfdev-5:
@ahmedo42 could you please fix the failure, please

vfdev-5 (approved):
LGTM, thanks @ahmedo42 !

PR #1772, opened 2021-03-12
Added Rouge metric

gucifer:
Added a simple Implementation for the ROUGE Metric.

Description:
Simple Implementation of the ROUGE metric (both Rouge-N and Rouge-L).

Check list:


sdesrozis:
@gucifer thank you ! As discussed, I will review / update this PR on my side. Therefore, we will ask to @vfdev-5 for the final review before merge.

ignite/metrics/rouge.py
```diff
@@ -0,0 +1,189 @@
...
+    ) -> None:
+        self._rougetotal: torch.Tensor = torch.tensor(0, device=device, dtype=torch.double)
+        self._num_examples: int = 0
+        self.beta: float = 0.0
+        self.n: int = 1
+        self.beta, self.n, self.rouge_fn, self.aggregate = self._check_parameters(beta, metric, aggregate)
+        self.alpha = 1 / (1 + self.beta ** 2)
+        super(Rouge, self).__init__(output_transform=output_transform, device=device)
+
+    def _check_parameters(self, beta: float, metric: str, aggregate: str) -> Tuple:
```
vfdev-5, line 122:
```suggestion
    def _check_parameters(self, beta: float, metric: str, aggregate: str) -> Tuple[float, int, Callable, str]:
```
and thus we can possibly remove redefinitions:
```python
        self.beta: float = 0.0
        self.n: int = 1
```

vfdev-5 (approved):
@sdesrozis and @gucifer, looks very nice now, thanks !
Nit picking comment

ignite/metrics/rouge.py
gucifer: Fixed it!

sdesrozis:
@vfdev-5 Currently I'm refactoring to have a corpus version. I will push it during the week-end.

sdesrozis:
@gucifer I worked on the PR today. Please, could you review ? Currently, it matches the results of the official perl ROUGE 1.5.5 scripts for single sentence. It almost works on multiple candidates but the average in the perl do resampling that slightly changes the results... I wrote a distributed test in which the process don't use the same data. Therefore, parallel reduction are validated.

Let's see if all works. Maybe it will need some fixes. 

@vfdev-5 Could you review the PR ? Sorry for the inconvenience.

ignite/metrics/rouge.py
```diff
@@ -0,0 +1,437 @@
...
+        candidate, references = output[0], output[1]
+        multiref_scores = [self.compute_score(candidate=candidate, reference=reference,) for reference in references]
+        score = self._mutliref_reducer(multiref_scores)
+        self._precision += score.precision
+        self._recall += score.recall
+        precision_recall = score.precision * score.recall
+        if precision_recall > 0:  # avoid zero division
+            self._fmeasure += precision_recall / ((1 - self._alpha) * score.precision + self._alpha * score.recall)
+        print("-********", candidate, references)
+        print(score.precision, score.recall, self._fmeasure)
```
vfdev-5, line 201: Do we need to keep this prints ?

ignite/metrics/rouge.py
```diff
@@ -0,0 +1,437 @@
...
+    def reset(self) -> None:
+        for m in self.internal_metrics:
+            m.reset()
+
+    @reinit__is_reduced
+    def update(self, output: Tuple[Sequence[Any], Sequence[Sequence[Any]]]) -> None:
+        for m in self.internal_metrics:
+            m.update(output)
+
+    @sync_all_reduce()
```
vfdev-5, line 432: I think we do not need to decorate the method if there is nothing to reduce ...

vfdev-5 (approved):
@sdesrozis looks great!
I left few comments.

ignite/metrics/rouge.py
sdesrozis: Sorry, I would have removed such debug prints...

ignite/metrics/rouge.py
```diff
@@ -399,15 +401,15 @@ class Rouge(Metric):
 
     def __init__(
         self,
-        metrics: Sequence[str] = None,
+        metrics: Optional[Sequence[str]] = None,
```
vfdev-5, line 403: Can't we set up one by default ?

ignite/metrics/rouge.py
sdesrozis: If the method is not decorated, a warning is raised when run in parallel. However, `Rouge` is a simple proxy to `RougeL` and `RougeN` which are parallel. If you have an idea, I take it !

ignite/metrics/rouge.py
```diff
@@ -171,35 +169,36 @@ def __init__(
...
         self._mutliref_reducer = self._get_multiref_reducer()
 
     def _get_multiref_reducer(self) -> MultiRefReducer:
         if self._multiref == "average":
             return MultiRefAverageReducer()
-        if self._multiref == "best":
+        elif self._multiref == "best":
             return MultiRefBestReducer()
+        else:
+            raise ValueError(f"multiref : wrong value (got : {self._multiref })")
```
vfdev-5, line 180:
```suggestion
            raise ValueError(f"multiref : wrong value (got : {self._multiref})")
```

ignite/metrics/rouge.py
```diff
@@ -171,35 +169,36 @@ def __init__(
         self._multiref = multiref
         valid_multiref = ["best", "average"]
         if self._multiref not in valid_multiref:
-            raise ValueError(f"aggregation must be in {valid_multiref} (got : {self._multiref })")
+            raise ValueError(f"multiref : valid values are {valid_multiref} (got : {self._multiref })")
```
vfdev-5, line 171:
```suggestion
            raise ValueError(f"multiref : valid values are {valid_multiref} (got : {self._multiref})")
```

ignite/metrics/rouge.py
vfdev-5: What kind of warning it is ? I'm thinking about `MetricsLambda.compute` which is not decorated:
https://github.com/pytorch/ignite/blob/cc76de461f63475f3b792c7c109fede95301556e/ignite/metrics/metrics_lambda.py#L88-L92

ignite/metrics/rouge.py
sdesrozis: Done.

ignite/metrics/rouge.py
sdesrozis: The warning below 

https://github.com/pytorch/ignite/blob/cc76de461f63475f3b792c7c109fede95301556e/ignite/metrics/metric.py#L215

ignite/metrics/rouge.py
```diff
@@ -0,0 +1,438 @@
...
+        self.internal_metrics: List[_BaseRouge] = []
+        for m in metrics:
+            if m == "Rouge-L":
+                self.internal_metrics.append(RougeL(multiref, alpha, output_transform, device))
+            else:
+                match = re.match("Rouge-(?P<ngram>[1-9])", m)
+                if not match:
+                    raise ValueError("Rouge metric name must be 'Rouge-L' or 'Rouge-N' for N in [1 ... 9]")
+                n = int(match.group("ngram"))
+                self.internal_metrics.append(RougeN(n, multiref, alpha, output_transform, device))
```
vfdev-5, line 420: Nit: maybe we could use input metrics as `Sequence[Union[str, int]]` and receive `["L", 1, 2, 4]` and simplify regexp etc ?

sdesrozis:
```
m = Rouge(metrics=["Rouge-1", "Rouge-2", "Rouge-4", "Rouge-L"], multiref="best", alpha=0.5)
```
would become
```
m = Rouge(metrics=[1, 2, 4, "L"], multiref="best", alpha=0.5)
```
I confess that is more compact but less clear to me. You choose :)

vfdev-5: Class `Rouge` is a group of Rouge-* variants. Maybe, it would make also sense to expose others: RougeN, RougeL in `ignite/metrics/__init__.py` ?

I'd code it even:
```python
m = Rouge(variants=[1, 2, 4, "L"], multiref="best", alpha=0.5)
```
But I have not strong opinion about that, just make things "simplier" probably

ignite/metrics/rouge.py
```diff
@@ -0,0 +1,438 @@
...
+        return Score(match=match, candidate=candidate, reference=reference)
+
+
+class MultiRefBestReducer(MultiRefReducer):
+    r"""
+    Reducer for selecting the best score
+    """
+
+    def __call__(self, scores: Sequence[Score]) -> Score:
+        return max(scores, key=lambda x: x.recall())
```
gucifer, line 149: Why are we using recall for calculating the best?

gucifer: Also, need to add perl as a dependency to run the perl package I think.

sdesrozis: > Why are we using recall for calculating the best?

That is the algorithm in the paper.

sdesrozis: > Also, need to add perl as a dependency to run the perl package I think.
> Check docs here:
> https://github.com/li-plus/rouge-metric
> 
> Also, if you wish to use the perl implementation to compare against I advise against using this module because the perl module is already downloaded and this package will only compare against that fixed version and not any future updates that might happen in the perl module.

Python 3.8.1 failed, others versions seems ok. Last thing to do before merge :)

gucifer: Can you explain why it failed? I thought it failed because of that

gucifer: Also, Looks great! Much cleaner than mine.

sdesrozis: Ok It is windows and MacOS tests.

Many options are possible for us 
* move to a pythonic ROUGE
* don't execute failing tests on such platforms
* install perl on windows and macos

There is a python version of rouge in `metric-rouge`. I tried it but results seemed different than perl rouge... Maybe we could try this further. 

@gucifer what do you think ?

gucifer: Sure! we can just use pythonic rouge. Let me find one which is most inline with Rouge155

gucifer:
@vfdev-5  @sdesrozis  I changed tests to py-rouge. It needs a specific nltk resource to run and we will have to download it everrytime. Can you review it and tell me if further changes are needed?

tests/ignite/metrics/test_rouge.py
```diff
@@ -0,0 +1,296 @@
...
+        engine = Engine(update)
+        m = Rouge(variants=[1, 2, "L"], alpha=0.5, device=metric_device)
+        m.attach(engine, "rouge")
+
+        engine.run(data=list(range(size)), max_epochs=1)
+
+        assert "rouge" in engine.state.metrics
+
+        evaluator = pyrouge.Rouge(
+            metrics=["rouge-n", "rouge-l", "rouge-w"],
```
sdesrozis, line 211:
```suggestion
            metrics=["rouge-n", "rouge-l"],
```

tests/ignite/metrics/test_rouge.py
```diff
@@ -0,0 +1,296 @@
+import os
+
+import nltk
```
sdesrozis, line 3: Useless ?

sdesrozis:
@gucifer 2 comments and we merge 🎂

tests/ignite/metrics/test_rouge.py
gucifer: we need to nltk to download punkt which is needed by py-rouge

sdesrozis (approved):
LGTM !!

sdesrozis:
Bravo @gucifer !!

docs/source/metrics.rst
```diff
@@ -323,6 +323,7 @@ Complete list of metrics
     precision.Precision
     PSNR
     recall.Recall
+    Rouge
```
vfdev-5, line 326: Please, add in another PR:
```
Rouge
RougeL
RougeN
```

ignite/metrics/rouge.py
```diff
@@ -0,0 +1,439 @@
...
+            raise ValueError(f"multiref : valid values are {valid_multiref} (got : {self._multiref})")
+        self._mutliref_reducer = self._get_multiref_reducer()
+
+    def _get_multiref_reducer(self) -> MultiRefReducer:
+        if self._multiref == "average":
+            return MultiRefAverageReducer()
+        elif self._multiref == "best":
+            return MultiRefBestReducer()
+        else:
+            raise ValueError(f"multiref : wrong value (got : {self._multiref})")
```
vfdev-5, line 180: @gucifer please add a test to cover this case in another PR

ignite/metrics/rouge.py
```diff
@@ -0,0 +1,439 @@
+import re
```
vfdev-5, line 1: We can remove `import re` now

vfdev-5:
Thanks a lot @sdesrozis and @gucifer for your hard work on this metric !

@gucifer I left few comments to address in a follow-up PR.

PR #1776, opened 2021-03-12
update test cases generation in accuracy metric

KickItLikeShika:
Updating test cases generation as [this](https://github.com/pytorch/ignite/blob/e4554604dca45529bebc03c8869fdeaeeb825132/tests/ignite/metrics/test_accuracy.py#L129) wasn't totally correct.

Check list:


vfdev-5 (approved):
Thanks @KickItLikeShika !

PR #1778, opened 2021-03-13
fix(docs): add heading for docs searching

### LINKED ISSUE #1766: improve documentation search results
## 📚 Documentation
Currently searching docs on duckduckgo and google results in poor search results.
DuckDuckGo: search precision metrics show source code result first while third one is a correct page.

Screen Shot 2021-03-10 at 22 57 27

Google: nearly same to google as well tho 2nd one is correct page.

Screen Shot 2021-03-10 at 22 58 08

For Algolia Docsearch: source code are shown first then correct ones.

Screen Shot 2021-03-10 at 23 04 31

---

This might be due to title of the page generated by `_modules` folder.
To improve this, we have the following options
1. use `autosummary` and generate the docs for only classes or functions or both just like [nn.Module](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module)
2. put heading 2 to every classes and functions before `autoclass`, `autofunction`
3. make `_modules` not to be indexed by search engines and algolia

Here's my suggestions:
- For search engine results, I think 1 > 2 > 3. IMO generating long docstring classes and functions to separate page would result better search result by search engines and docs readability (sometimes I got lost in long docstrings and methods). **generating such pages have a downside, we can't specify `:members:` in `.. autoclass` manually** since it will be already included in template. See [pytorch template](https://github.com/pytorch/pytorch/blob/master/docs/source/_templates/classtemplate.rst).
- For algolia docsearch, 1 == 2 > 3. Algolia docsearch searches h1-h6, so adding heading to every classes and functions can result better in-docs search.

What do you think ? @vfdev-5 @sdesrozis

ydcjeff:
Fixes #1766 

Description: Add heading 3 for docs searching

Check list:


vfdev-5:
@ydcjeff, thanks for the PR , however, I'm not sure if it solves the issue with searching for "Engine" word. The output is still irrelevant. In addition, removing `autosummary` makes the docs a bit ugly IMO: https://deploy-preview-1778--pytorch-ignite-preview.netlify.app/engine.html

ydcjeff:
I also feel like that and now playing with another style locally. will push in a minute

ydcjeff:
> I'm not sure if it solves the issue with searching for "Engine" word

Algolia indexed the current https://pytorch.org/ignite/ and displayed the results from it. So, changing the urls, headings in the PR will not reflect to this.

PR #1783, opened 2021-03-13
Remove unused imports in test

Moh-Yakoub:
Description: Those two imports were left from a previous PR and they are not used in the test. removed for cleanup

Check list:


vfdev-5 (approved):
Thanks @Moh-Yakoub !

Issue #1787, opened 2021-03-14
Make `autodoc_typehints` mode work with nitpicky mode


ydcjeff:
## 📚 Documentation
Currently we have imported the class in the handlers and metrics docs with `module.class_name` and `class_name`.
[handlers docs](https://pytorch.org/ignite/master/handlers.html) [[source](https://github.com/pytorch/ignite/blob/master/docs/source/handlers.rst)] and [metrics docs](https://pytorch.org/ignite/master/metrics.html#complete-list-of-metrics) [[source](https://github.com/pytorch/ignite/blob/master/docs/source/metrics.rst)]

This seems inconsistent. The reason why we have to import like this is we are using [nitpicky mode](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-nitpicky) to make sure all references are updated.
For example, [here returns EarlyStopping instance](https://pytorch.org/ignite/master/contrib/engines.html#ignite.contrib.engines.common.add_early_stopping_by_val_score) which is referenced to the class `EarlyStopping` which we need to write `ignite.handlers.early_stopping.EarlyStopping` to make reference work.

The idea is to have those references work even if we write the docs with full sub-module path or module path.
Current handlers docs:
```rst
.. autosummary::
    :nosignatures:
    :toctree: generated

    checkpoint.Checkpoint
    DiskSaver
    checkpoint.ModelCheckpoint
    early_stopping.EarlyStopping
    terminate_on_nan.TerminateOnNan
    TimeLimit
    timing.Timer
    global_step_from_engine
```
Handlers docs we want:
```rst
.. autosummary::
    :nosignatures:
    :toctree: generated

    Checkpoint
    DiskSaver
    ModelCheckpoint
    EarlyStopping
    TerminateOnNan
    TimeLimit
    Timer
    global_step_from_engine
```
This is also same for [metrics docs](https://github.com/pytorch/ignite/blob/master/docs/source/metrics.rst#complete-list-of-metrics)

Related sphinx conf : [`autodoc_typehints`](https://github.com/pytorch/ignite/blob/ba57568cf7d4343d22171645848c4af56a0a0ae4/docs/source/conf.py#L213) and [`nitpicky`](https://github.com/pytorch/ignite/blob/ba57568cf7d4343d22171645848c4af56a0a0ae4/docs/source/conf.py#L327-L341)

vfdev-5:
Related to this merged PR : https://github.com/pytorch/ignite/pull/1780

[...]

PR #1789, opened 2021-03-14
Rouge last details

gucifer:
Added a new test for multireference value check.
Added RougeN and RougeL to docs.
Check list:


gucifer:
@vfdev-5 can you check the issue with docs? They are the same old docs but they seem to be failing the linkcheck test

vfdev-5:
I think this should be related to https://github.com/pytorch/ignite/issues/1787
cc @ydcjeff

ydcjeff:
According to the error message
```
ignite.metrics.rouge.RougeL.rst:35:<autosummary>:1:py:obj reference target not found: ignite.metrics.rouge.RougeL.compute_score
```
`compute_score` of `RougeL` doesn't have a docstring while docs is trying to include that method.

vfdev-5:
Thanks a lot @ydcjeff ! 
@sdesrozis , @gucifer I was also wondering if those methods should be necessarily public ? Otherwise, let's put them private and thus no need to support them...

gucifer:
> Thanks a lot @ydcjeff !
> @sdesrozis , @gucifer I was also wondering if those methods should be necessarily public ? Otherwise, let's put them private and thus no need to support them...

Simple Rouge class will apply a training sample on all the metrics for which it is instantiated but if you wish to manually update only one metric this is a way to do so. Also, similarly metric_name is to find which object in Rouge is for which exact rouge metric.

vfdev-5:
I would say that `metric_name` can private. I understand its usage inside Rouge but If I instantiate `m = RougeL`, I'm not sure that I'll call `m.metric_name()` to get the class name.
If we keep `compute_score`, we have to explain the difference vs `compute` method.

gucifer:
> I would say that `metric_name` can private. I understand its usage inside Rouge but If I instantiate `m = RougeL`, I'm not sure that I'll call `m.metric_name()` to get the class name.
> If we keep `compute_score` public, we have to explain the difference vs `compute` method.

I will make them private then. Also, it is asking docs for pretty much every class, please check on that.

vfdev-5:
But I agree that every public class should be documented, otherwise put it private.

gucifer:
> But I agree that every public class should be documented, otherwise put it private. That's new module added by Jeff is called `nitpicky` :)

I will make everything private tomorrow if that's fine by you.

sdesrozis:
@gucifer Why this PR has so much commits ? 

Could you create a clean PR from a fresh new branch created from master ? It means you would have to first merge the upstream master in your fork master before create a new branch. Therefore, merged `rouge` will be in your new branch.

Thanks !

Issue #1790, opened 2021-03-14
Improve code coverage


vfdev-5:
Our project code coverage start to decrease with new features and reworked tests. We are now ~95% but were 96% recently and around ~98% the last year...

The idea is to identify the uncovered parts of code inspecting here, for example: https://codecov.io/gh/pytorch/ignite/tree/master/ignite and draft a PR with new tests.

Some of new test can be very simple and others could be very sophisticated to enter a particular code line.
Let's improve the coverage of all modules, except `distributed` for instance.

Moh-Yakoub:
That sounds like a group effort. shall we update the task with a checkbox list of packages so that we make sure that no two people will be working on the same package?

soma2000-lang:
@vfdev-5  Can I work on ignite/engine/engine.py?

soma2000-lang:
@vfdev I will try

soma2000-lang:
@vfdev-5  https://app.codecov.io/gh/pytorch/ignite/blob/master/ignite/engine/engine.py in this file I see almost everything is covered .Please correct me if I am wrong

vfdev-5:
@soma2000-lang yes 97% of the code is covered but we need to cover up to 100% if possible.

soma2000-lang:
Ok

mohameds-dev:
Hi, I am a senior CS student with industry experience in Python (Django) as well as my own side projects development.

I've practiced Test-Driven Development (TDD) as part of a Software Design class and I am familiar with how test coverage works in Python.

I am looking to practice open-source contribution by contributing to this project, specifically attempting to push the coverage to 100% [on this file as shown here](https://app.codecov.io/gh/pytorch/ignite/blob/master/ignite%2Fengine%2Fengine.py).

Could you assign me to this issue?

PR #1807, opened 2021-03-17
cov: add test for ignite.base

ydcjeff:
Part of #1790 

Description: Add test for ignite.base

Check list:

### DIFF
--- a/tests/ignite/base/test_mixins.py
+++ b/tests/ignite/base/test_mixins.py
@@ -1,8 +1,15 @@
+import pytest
+
 from ignite.base import Serializable
 
 
-def test_state_dict_and_load_state_dict():
+def test_state_dict():
+    s = Serializable()
+    with pytest.raises(NotImplementedError):
+        s.state_dict()
+
+
+def test_load_state_dict():
 
     s = Serializable()
-    s.state_dict()
     s.load_state_dict({})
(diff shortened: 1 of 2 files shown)


tests/ignite/base/test_mixins.py
```diff
@@ -1,7 +1,8 @@
 from ignite.base import Serializable
 
 
-def test_load_state_dict():
+def test_state_dict_and_load_state_dict():
 
     s = Serializable()
```
sdesrozis, line 6: I don't understand the test... more precisely I don't understand why this method is not abstract in the class or why it does not have a default implementation.

trsvchn: I think we can treat it as a "smoke test" for `Serializable` :)
About method, as a [common practice](https://docs.python.org/3/library/exceptions.html#NotImplementedError) we can raise there `NotImplementedError`:

> This exception is derived from RuntimeError. In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

```python
    def state_dict(self) -> OrderedDict:
        raise NotImplementedError
```

sdesrozis: Or use `abstractmethod` from `abc`

sdesrozis (approved):
@ydcjeff Thank you! LGTM

PR #1811, opened 2021-03-17
cov: add error checking tests for distributed/launcher.py 

ydcjeff:
Part of #1790 

Description: Add error checking tests for launcher.py

Check list:

### DIFF
--- a/tests/ignite/distributed/test_launcher.py
+++ b/tests/ignite/distributed/test_launcher.py
@@ -10,7 +10,7 @@ import ignite.distributed as idist
 from ignite.distributed.utils import has_hvd_support, has_native_dist_support, has_xla_support
 
 
-def test_parallel_error():
+def test_parallel_wrong_inputs():
     with pytest.raises(ValueError, match=r"Unknown backend 'abc'. Available backends:"):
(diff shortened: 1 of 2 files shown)


sdesrozis:
@ydcjeff Thank you ! Why not add wrong inputs for the static method `_setup_spawn_params` ?

tests/ignite/distributed/test_launcher.py
```diff
@@ -10,6 +10,29 @@
 from ignite.distributed.utils import has_hvd_support, has_native_dist_support, has_xla_support
 
 
+def test_parallel_error():
```
sdesrozis, line 13:
```suggestion
def test_parallel_wrong_inputs():
```

ydcjeff:
> @ydcjeff Thank you ! Why not add wrong inputs for the static method `_setup_spawn_params` ?

Since its private method, I go from `Parallel` also going from `Parallel` is readable.

sdesrozis:
> Since its private method, I go from `Parallel` also going from `Parallel` is readable.

My bad ! You are totally right, this method is covered by the tests.

PR #1812, opened 2021-03-17
cov: add tests in {events,deterministic}.py

ydcjeff:
Part of #1790 

Description: Add tests in events.py

Check list:

### DIFF
--- a/tests/ignite/engine/test_deterministic.py
+++ b/tests/ignite/engine/test_deterministic.py
@@ -22,7 +22,7 @@ from tests.ignite.engine import BatchChecker, setup_sampler
 
 
 def test_dengine_setup_seed_div_by_zero():
-    with pytest.raises(ZeroDivisionError, match=r"integer division or modulo by zero"):
+    with pytest.raises(ValueError, match=r"iter_counter should be positive value"):
         DeterministicEngine(lambda e, b: None)._setup_seed(iter_counter=0)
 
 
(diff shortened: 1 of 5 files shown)


tests/ignite/engine/test_event_handlers.py
```diff
@@ -6,7 +6,7 @@
 from pytest import raises
 
 from ignite.engine import Engine, Events, State
-from ignite.engine.events import EventsList
+from ignite.engine.events import EventsList, RemovableEventHandle
```
ydcjeff, line 9:
```suggestion
from ignite.engine.events import EventsList
```

tests/ignite/engine/test_deterministic.py
```diff
@@ -21,6 +21,11 @@
 from tests.ignite.engine import BatchChecker, setup_sampler
 
 
+def test_dengine_setup_seed_div_by_zero():
+    with pytest.raises(ZeroDivisionError, match=r"integer division or modulo by zero"):
```
trsvchn, line 25: Nice catch!
@vfdev-5 should we handle this `iter_counter=0` exception with our custom exception message?

ydcjeff: But `_setup_seed` is called internally with arguments which does not occur ZeroDivError

trsvchn: oh, I wasn't clear, I meant catch it before `Zero..` with custom `ValueError`

```python
if iter_counter == 0:
    raise ValueError
```

trsvchn (approved):
LGTM! Thanks @ydcjeff !

sdesrozis (approved):
@ydcjeff Thank you !

PR #1814, opened 2021-03-18
cov: add tests for auto.py

ydcjeff:
Part of #1790 

Description: Add tests for auto.py

Check list:


sdesrozis (approved):
@ydcjeff Thank you ! LGTM

PR #1818, opened 2021-03-18
script: add code style bash script 

ydcjeff:
Description: For those who don't use pre-commit and IDE extensions, a bash script for code formatting. And adding same args used in `unit-test` in `pre-commit`

The rest files are modified by pre-commit.

Check list:

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -86,7 +86,7 @@ jobs:
         shell: bash -l {0}
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: |
-          pip install mypy
+           bash ./tests/run_code_style.sh install
           bash ./tests/run_code_style.sh mypy
 
       - name: Run Tests
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -79,17 +79,15 @@ jobs:
       - name: Check code formatting
         shell: bash -l {0}
         run: |
-          pip install flake8 "black==19.10b0" "isort==5.7.0"
-          flake8 ignite/ tests/ examples/
-          black --check .
-          isort -c --profile black .
+          bash ./tests/run_code_style.sh install
+          bash ./tests/run_code_style.sh lint
 
       - name: Run Mypy
         shell: bash -l {0}
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: |
           pip install mypy
-          mypy --config-file mypy.ini
+          bash ./tests/run_code_style.sh mypy
 
       - name: Run Tests
         shell: bash -l {0}
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -86,7 +86,6 @@ jobs:
         shell: bash -l {0}
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: |
-           bash ./tests/run_code_style.sh install
           bash ./tests/run_code_style.sh mypy
 
       - name: Run Tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,11 +125,10 @@ If you choose not to use pre-commit, you can take advantage of IDE extensions co
 black manually to format files and commit them.
 
 ```bash
+# install required exact versions
+bash ./tests/run_code_style.sh install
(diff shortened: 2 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -87,7 +87,8 @@ git clone https://github.com/pytorch/ignite.git
 cd ignite
 python setup.py develop
 pip install -r requirements-dev.txt
-pip install flake8 "black==19.10b0" "isort==5.7.0" mypy
+pip install mypy
```
vfdev-5, line 90: shouldn't we install mypy with `run_code_style.sh install` ?

vfdev-5 (commented):
Thanks for the improvement, @ydcjeff !
I left a comment

CONTRIBUTING.md
ydcjeff: It will install mypy for mac and windows in the CI.
If its okay for you, we can add mypy in the script.

vfdev-5: I think it is OK as far as we do not execute it.

.github/workflows/unit-tests.yml
```diff
@@ -79,17 +79,15 @@ jobs:
...
-          black --check .
-          isort -c --profile black .
+          bash ./tests/run_code_style.sh install
+          bash ./tests/run_code_style.sh lint
 
       - name: Run Mypy
         shell: bash -l {0}
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: |
           pip install mypy
```
ydcjeff, line 89:
```suggestion
           bash ./tests/run_code_style.sh install
```

CONTRIBUTING.md
ydcjeff, line 90:
```suggestion
```

tests/run_code_style.sh
```diff
@@ -0,0 +1,17 @@
...
+    isort . --check --settings setup.cfg
+    black . --check --config pyproject.toml
+elif [ $1 = "fmt" ]; then
+    flake8 ignite tests examples --config setup.cfg
+    isort . --settings setup.cfg
+    black . --config pyproject.toml
+elif [ $1 = "mypy" ]; then
+    mypy --config-file mypy.ini
+elif [ $1 = "install" ]; then
+    pip install flake8 "black==19.10b0" "isort==5.7.0"
```
ydcjeff, line 16:
```suggestion
    pip install flake8 "black==19.10b0" "isort==5.7.0" mypy
```

.github/workflows/unit-tests.yml
```diff
@@ -79,17 +79,15 @@ jobs:
...
+          bash ./tests/run_code_style.sh install
+          bash ./tests/run_code_style.sh lint
 
       - name: Run Mypy
         shell: bash -l {0}
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: |
-          pip install mypy
-          mypy --config-file mypy.ini
+           bash ./tests/run_code_style.sh install
```
ydcjeff, line 89:
```suggestion
```

vfdev-5 (approved):
LGTM, thansk @ydcjeff !

tests/run_code_style.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -xeu
+
+if [ $1 = "lint" ]; then
+    flake8 ignite tests examples --config setup.cfg
+    isort . --check --settings setup.cfg
+    black . --check --config pyproject.toml
+elif [ $1 = "fmt" ]; then
+    flake8 ignite tests examples --config setup.cfg
```
vfdev-5, line 10: @ydcjeff this can lead to incorrect behaviour when flake8 fails and script will exit here due to `set -xeu`.
Why you replaced `autopep8 --recursive --in-place --aggressive --aggressive .` by flake8 ?
Let's restore previous behaviour with autopep8. Please, send a PR with the fix. Thanks

ydcjeff: `autopep8` is a formatter. We already have `black`. So we probably don't need two formatters.

vfdev-5: That's true about `autopep8`, I agree with you, let's keep only black then and remove `flake8` as if it fails script wont execute the following below commands. Thanks for pointing out, @ydcjeff  !

PR #1820, opened 2021-03-18
Add output detach for Canberra metric

KickItLikeShika:
Add `output.detach()` in `CanberraMetric`

Check list:


vfdev-5 (approved):
Thanks @KickItLikeShika !

PR #1823, opened 2021-03-18
Extend `sync_all_reduce` API

### LINKED ISSUE #1813: Support for max/min ops for `sync_all_reduce`
Running the distributed tests while `world_size > 1` makes the tests fail. I was trying to make the metric compatible with DDP with doing the same thing we have done with the other metrics but the results are different!
I tried to print out some things for a better understanding, everything seemed to be working fine (calculating the metric is very simple) but after doing `metric.compute()` returns a strange result! (even printing the values computed in `_update()` function seems fine!).

[Code snippet](https://gist.github.com/KickItLikeShika/fcd6a671cddbbfc7f787403c32fc792f)

KickItLikeShika:
Fixes #1813 

Extending `sync_all_reduce` API, added `op` arg to allow more reduction operations (`SUM`, `PRODUCT`, `MAX`, `MIN`)

Check list:


ignite/metrics/metric.py
```diff
@@ -526,7 +526,7 @@ def __setstate__(self, d: Dict) -> None:
         self.__dict__.update(d)
 
 
-def sync_all_reduce(*attrs: Any) -> Callable:
+def sync_all_reduce(*attrs: Any, op: str = "SUM") -> Callable:
```
sdesrozis, line 529: Ok, that is a way to do that. But we could want mix different ops
```python
@sync_all_reduce("sum_var", "max_var")
def compute(self):
    pass   
```

What do you think about the following suggestion ?
```python
@sync_all_reduce("sum_var", ("max_var", "MAX"))
def compute(self):
    pass   
```
I used a tuple, it is not very friendly but the idea is to provide a way to select the op to apply per variable. Maybe we could introduce specific tools:
```python
@sync_all_reduce("sum_var", max_reduced_variable("max_var"))
def compute(self):
    pass   
```
Interested by your opinion.

KickItLikeShika: The way i updated `sync_all_reduce` will allow minimal changes and that was my concern, i think chaining decorators with different `op` arg will solve the "select the op to apply per variable".

vfdev-5: We maydo it with several decorators ?
```python
@sync_all_reduce("sum_var")
@sync_all_reduce("max_var", op="MAX")
def compute(self):
    pass
```
And internally, resolve the decoration in a bit more sophisticated way...

KickItLikeShika: Thanks for the explanation @vfdev-5, that's what i mean by 
> chaining decorators with different `op` arg will solve the "select the op to apply per variable" issue.

vfdev-5: However, chaining decorators may look nice, i'm not a fan of multiple subfunc calls: each decorator brings 2 new functions. With 2 sync_all_reduce, we have 4 functions to call before calling `compute`. I'd like to avoid that...

vfdev-5: Maybe, Sylvain's solution like `@sync_all_reduce("sum_var", ("max_var", "MAX"))` would be the cleanest one which we can simplify a bit more as 
```
@sync_all_reduce("sum_var", "max_var:max")
def compute(self):
    pass   
```

KickItLikeShika: I can do that, should i make `@sync_all_reduce("sum_var", ("max_var", "MAX"))` tuples or what?

vfdev-5: Personnally, I prefer `@sync_all_reduce("sum_var", "max_var:max", "min_var:min", "another_var:sum")`.
@sdesrozis what do you think ?

sdesrozis: I prefer encode the op in the str rather than handle tuples 😉 It is very elegant.

KickItLikeShika:
Already tested these updates on #1813 and it works fine

sdesrozis:
Thank you @KickItLikeShika, it looks good. However, unitary tests of this improvement are needed
* test all the combination of op
* catch the error if a wrong op (for instance `"var:fakeop"`, `"var:"`, etc.)

ignite/metrics/metric.py
```diff
@@ -548,9 +548,15 @@ def another_wrapper(self: Metric, *args: Any, **kwargs: Any) -> Callable:
             if len(attrs) > 0 and not self._is_reduced:
                 if ws > 1:
                     for attr in attrs:
+                        if ":" in attr:
+                            idx = attr.index(":")
+                            op = attr[idx + 1 :].strip().upper()
```
sdesrozis, line 553: I’m wondering if this rule should not be included in `all_reduce`...

ignite/metrics/metric.py
```diff
@@ -548,9 +548,15 @@ def another_wrapper(self: Metric, *args: Any, **kwargs: Any) -> Callable:
...
+                            attr = attr[:idx]
+                        else:
+                            op = "SUM"
```
sdesrozis, line 556: Same here, maybe if no op is provided, we should call `idist.all_reduce(t)` and use the default of `all_reduce`.

ignite/metrics/metric.py
KickItLikeShika: You mean `strip().upper()`? it did that to handle `var: SUM` or `var: SUM `  just to strip the spaces and make it uppercase to handle `var: Sum`, `var:sum` or `var: sUM` as documented [here](https://pytorch.org/tutorials/intermediate/dist_tuto.html)

ignite/metrics/metric.py
KickItLikeShika: I will work on tests just after we agree on the implementation
> * test all the combination of op

That's already done in `torch.distirubted` but i can do that in `sync_all_reduce()`, what do you think?
> * catch the error if a wrong op (for instance `"var:fakeop"`, `"var:"`, etc.)

KickItLikeShika:
I will work on tests just after we agree on the implementation
> * test all the combination of op

That's already done in `torch.distirubted` but i can do that in `sync_all_reduce()`, what do you think?
> * catch the error if a wrong op (for instance `"var:fakeop"`, `"var:"`, etc.)

ignite/metrics/metric.py
KickItLikeShika: Can you please explain your point more?

ignite/metrics/metric.py
sdesrozis: The reduce method is the `idist` wrapper. Your api handles spaces and lower/upper case variations. My suggestion was about pushing this into the wrapper. 

But on second thoughts, it is not a good idea. Impossible to know if `SUm` is a bug or what is really desired. In fact, we should provide only one way to describe a reduction : 
* `var:max`
* `var:min`
* `var:sum`
* `var:prod`

So an exception should be raised on other string.
```python
valid_ops = ["min", "max", "sum", "prod"]
if op not in valid_ops:
    raise ValueError(f"Reduction operation is not valid (expected : {valid_ops}, got: {op}")
```
What do you think ?

KickItLikeShika: Definitely we need the exception part, but i thought keeping it as flexible as we can and easier to use is a positive thing, right?

ignite/metrics/metric.py
sdesrozis: If no op is provided, we could have a default as you did. Another option is to use the `idist.all_reduce` without `op` argument.
```python
for attr in attrs:
    op_kwargs = {}
    if ":" in attr:
        idx = attr.index(":")
        op = attr[idx + 1 :]
        valid_ops = ["min", "max", "sum", "prod"]
        if op not in valid_ops:
            raise ValueError(f"Reduction operation is not valid (expected : {valid_ops}, got: {op}")
        attr = attr[:idx]
        op_kwargs["op"] = op
    t = getattr(self, attr, None)
    if t is not None:
        t = idist.all_reduce(t, **op_kwargs)
        self._is_reduced = True
        setattr(self, attr, t)
```
Thoughts on that ?

ignite/metrics/metric.py
KickItLikeShika: Btw, `all_reduce()` can't handle `min, max, ...` in a lower case it must be `MIN, MAX, ..`

ignite/metrics/metric.py
KickItLikeShika: Seems cleaner, thanks for the thoughts

sdesrozis:
> That's already done in `torch.distirubted` but i can do that in `sync_all_reduce()`, what do you think?

Tests to update are here 
https://github.com/pytorch/ignite/blob/eca41b9e2cb66cb9c3ce45064873194a073ae1a8/tests/ignite/metrics/test_metric.py#L541

We must not forgot the update of the documentation too.

ignite/metrics/metric.py
```diff
@@ -548,9 +550,18 @@ def another_wrapper(self: Metric, *args: Any, **kwargs: Any) -> Callable:
             if len(attrs) > 0 and not self._is_reduced:
                 if ws > 1:
                     for attr in attrs:
+                        op_kwargs = {}
+                        if ":" in attr:
+                            idx = attr.index(":")
```
vfdev-5, line 555: Can't we do directly `attr, op = attr.split(":")` instead of using index etc ?

vfdev-5 (commented):
Thanks for the PR @KickItLikeShika ! A question about implementation.

ignite/metrics/metric.py
KickItLikeShika: If `:` is not existed it will raise an error 
```
ValueError: substring not found
```

KickItLikeShika: Yes that can be done after the check `if ":" in attr:`

vfdev-5 (approved):
LGTM!
@sdesrozis your thoughts and your turn to review :)

sdesrozis (approved):
@KickItLikeShika Thank you ! LGTM

KickItLikeShika:
@vfdev-5 @sdesrozis please check the CI failures, we get an assertion error here `assert (self.prod.cpu() == (self.prod_nocomp * 5) * (self.prod_nocomp * 5)).all()`, But it works fine locally!
Screenshot from 2021-03-21 17-49-34

vfdev-5:
@KickItLikeShika do you mean XLA tests? Have you tested it with xla locally ?

KickItLikeShika:
@vfdev-5 the CPU tests also fail https://github.com/pytorch/ignite/pull/1823/checks?check_run_id=2160328500#step:10:963

KickItLikeShika:
@vfdev-5 I realized the error, in the `prod` test i just considered `world_size = 1`, i will fix that, Thanks!

KickItLikeShika:
@vfdev-5 can you check this please? https://github.com/pytorch/ignite/pull/1823/checks?check_run_id=2160568207#step:8:1226

KickItLikeShika:
@sdesrozis can you please check this? https://github.com/pytorch/ignite/pull/1823/checks?check_run_id=2160568207#step:8:1226

tests/ignite/metrics/test_metric.py
```diff
@@ -554,20 +555,52 @@ def reset(self):
...
         @reinit__is_reduced
         def update(self, output):
+            # SUM op
             self.n += 1
             self.c += 1.23456
             self.a += 10.0
             self.b -= 5.0
 
+            # MAX op
+            mae = torch.abs(self.test_pred - self.test.view_as(self.test_pred)).max().item()
```
vfdev-5, line 592: I think the idea when we would like to test min/max here is to test it between different ranks =>
let's set up those vars as 
```python
self.m += idist.get_rank()
self.k -= idist.get_rank()
```
then inside `compute` it should be something like:
```python
assert self.m == idist.get_world_size() - 1
assert self.k == 10000 - idist.get_world_size()
```

KickItLikeShika: Regarding what's inside `compute` i guess it can't be something like `self.m == idist.get_world_size() - 1` if we increase `m` by `rank`, for example when `world_size = 2` `rank=[0, 1]`, i guess it doesn't make sense if we increased it by `rank` then we compare it with `world_size`, what do you think?

vfdev-5: I think it does make sense. More precisely it should be `num_updates * (world_size - 1) - 1`.
If ws is 2, for rank 0, `self.m = -1 + 0` after the first update and n-th update. For rank 1, `self.m = -1 + 1 * num_updates` after n-th update. Inside compute we ask to reduce `m:MAX` and thus `max(-1, -1 + num_updates)=-1 + num_updates * (ws - 1)`...

ignite/metrics/metric.py
```diff
@@ -548,9 +550,16 @@ def another_wrapper(self: Metric, *args: Any, **kwargs: Any) -> Callable:
             if len(attrs) > 0 and not self._is_reduced:
                 if ws > 1:
                     for attr in attrs:
+                        op_kwargs = {}
+                        if ":" in attr:
+                            attr, op = attr.split(":")
+                            valid_ops = ["MIN", "MAX", "SUM", "PRODUCT"]
+                            if op not in valid_ops:
+                                raise ValueError(f"Reduction operation is not valid (expected : {valid_ops}, got: {op}")
```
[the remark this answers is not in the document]
KickItLikeShika: Should i add the test inside `_test_distrib_sync_all_reduce_decorator` function or add it as separate test, what's better?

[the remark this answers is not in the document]
vfdev-5: Yes, let's create a separate test for that.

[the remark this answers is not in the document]
KickItLikeShika: Regarding the CI error, can you have a look please? https://github.com/pytorch/ignite/pull/1823/checks?check_run_id=2160568207#step:8:1226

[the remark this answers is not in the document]
vfdev-5: I found a bug today in idist all reduce with hvd when using min/max, maybe product is also impacted...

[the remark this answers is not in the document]
KickItLikeShika: hvd tests have been passed, Thanks a lot!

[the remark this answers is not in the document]
KickItLikeShika: hvd tests have been passed, Thanks a lot!

PR #1825, opened 2021-03-18
docs: link to master about page [skip ci]

ydcjeff:
Fixes #{issue number}

Description:

Check list:


vfdev-5:
@ydcjeff we can't test that on netlify as [skip ci] :(

vfdev-5 (approved):
Let's see. Thanks @ydcjeff

ydcjeff:
@vfdev-5 Docs build will pass, but `linkcheck` job will fail since it has `-W` which is "treats warning as error"

vfdev-5:
@ydcjeff can we ignore this single warning somehow otherwise it is annoying to have a failing test...

PR #1826, opened 2021-03-18
covtest: Adding test for handlers/checkpoint save_xla

Moh-Yakoub:
Part of #1790

Description: increase coverage of the following test https://codecov.io/gh/pytorch/ignite/src/master/ignite/handlers/checkpoint.py

Check list:


tests/ignite/handlers/test_checkpoint.py
```diff
@@ -1574,3 +1575,21 @@ def test_get_default_score_fn():
     score_fn = Checkpoint.get_default_score_fn("loss", -1)
     score = score_fn(engine)
     assert score == -0.123
+
+
+def test_save_xla_for_disk_saver(dirname):
+    with mock.patch("ignite.handlers.checkpoint.idist.has_xla_support"):
```
vfdev-5, line 1581: Instead of mocking we can execute this test with XLA tests as here for example : 
https://github.com/pytorch/ignite/blob/d8e6bec44e8307f22a0c926ef78a07c47255fd19/tests/ignite/handlers/test_checkpoint.py#L1215-L1221

vfdev-5 (commented):
@Moh-Yakoub thanks for the PR, I have a comment on the way it is done the testing.

tests/ignite/handlers/test_checkpoint.py
Moh-Yakoub: @vfdev-5  wouldn't `@pytest.mark.skipif(not idist.has_xla_support, reason="Not on TPU device") ` would be executed only for specific test runs i.e. when we are running tests on TPU device.
My assumption was to actually execute that test regardless of the running device of the test case, in that case the coverage would increase for the `checkpoint` module. Does this make sense?

vfdev-5: We have XLA tests executed here with GIthub Actions and codecov uploaded, please see : https://github.com/pytorch/ignite/blob/master/.github/workflows/tpu-tests.yml

I think this test have to be executed within the appropriate test suite.

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -1574,3 +1575,24 @@ def test_get_default_score_fn():
...
+        try:
+            with mock.patch("ignite.handlers.checkpoint._save_func") as _save_func_mock:
+                saver = DiskSaver(dirname, atomic=atomic, create_dir=False, require_empty=False)
+                fname = "test.pt"
+                saver(_to_save, fname)
+                assert _save_func_mock.called
+        except ModuleNotFoundError:
+            pass
+        _test_existance(atomic=False, _to_save=to_save_serializable)
+        _test_existance(atomic=True, _to_save=to_save_non_serializable)
```
vfdev-5, line 1598:
```suggestion
    _test_existance(atomic=False, _to_save=to_save_serializable)
    _test_existance(atomic=True, _to_save=to_save_non_serializable)
```
otherwise tests are not executed

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -1574,3 +1575,24 @@ def test_get_default_score_fn():
...
+@pytest.mark.skipif("NUM_TPU_WORKERS" in os.environ, reason="Skip if NUM_TPU_WORKERS is in env vars")
+@pytest.mark.skipif(not idist.has_xla_support, reason="Not on TPU device")
+def test_save_xla_for_disk_saver(dirname):
+    model = DummyModel()
+    to_save_serializable = {"model": model}
+    to_save_non_serializable = {"model": lambda x: x}
+
+    def _test_existance(atomic, _to_save):
+        try:
+            with mock.patch("ignite.handlers.checkpoint._save_func") as _save_func_mock:
```
vfdev-5, line 1590: Could you please explain why we have to mock this function ?

vfdev-5 (commented):
Thanks for the update @Moh-Yakoub , still few comments to address here

vfdev-5:
@Moh-Yakoub inspecting https://codecov.io/gh/pytorch/ignite/src/master/ignite/handlers/checkpoint.py it looks like it is justt he line 280 is not covered. Please, update this PR and rename it accordingly (not more related to xla). Thank you

PR #1830, opened 2021-03-19
Improve MNB metric for DDP

KickItLikeShika:
Improving `MeanNormalizedBias` implementation to be compatible with DDP and add the necessary tests.

Check list:


sdesrozis (approved):
@KickItLikeShika Thank you ! LGTM

PR #1832, opened 2021-03-20
WIP on fixing failing test

vfdev-5:
Description:
- WIP on fixing failing test: `test__native_dist_model_create_dist_gloo_1[FILE]` which uses `get_fixed_dirname`
- https://github.com/pytorch/ignite/runs/2152949660

Check list:

### DIFF
--- a/tests/ignite/conftest.py
+++ b/tests/ignite/conftest.py
@@ -18,20 +18,24 @@ def dirname():
 
 @pytest.fixture()
 def get_fixed_dirname(worker_id):
+    # multi-proc friendly fixed tmp dirname
+    path = "/tmp/fixed_tmp_dirname_"
+    lrank = int(worker_id.replace("gw", "")) if "gw" in worker_id else 0
+
     def getter(name="test"):
-        # multi-proc friendly fixed tmp dirname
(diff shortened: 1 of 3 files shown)


tests/ignite/conftest.py
```diff
@@ -25,7 +25,7 @@ def getter(name="test"):
         time.sleep(0.5 * lrank)
         os.makedirs(path, exist_ok=True)
         yield path
-        time.sleep(1.0 * lrank + 1.0)
+        time.sleep(1.0 * lrank + 2.0)
```
KickItLikeShika, line 28: I think a good approach to see if the issue is here or there is something else we can increase the `sleep time` to a big value and see what's going to happen, if it works then we start reducing it to some proper value

KickItLikeShika: Increasing the `sleep time` didn't work btw

vfdev-5: Thanks for inverstigating that @KickItLikeShika ! 
I also tried to repro that locally and everything works OK. Strange thing is about the actual problem here: `Aborted                 (core dumped)`.

PR #1833, opened 2021-03-20
[WIP] Adding torch ccl backend

### LINKED ISSUE #1728: Add oneCCL backend for idist
## 🚀 Feature

The idea is to add "oneCCL" backend by extending our [torch native dist model](https://github.com/pytorch/ignite/blob/395743d1f782ca84a7fb1a3c71afe2fa76b882a1/ignite/distributed/comp_models/native.py#L22).

Refs:
- https://pytorch.medium.com/optimizing-dlrm-by-using-pytorch-with-oneccl-backend-9f85b8ef6929
- https://github.com/intel/torch-ccl

Moh-Yakoub:
Fixes #1728 

Description:

This is a WIP to have a discussion about the support of CCL backend, this is my first backend implementation so I'm expecting it will need couple of iterations before it's fully complete.

Few points to discuss as well

1. Given the way we define the ccl, does it make sense to just add it as a supported backend in the native computation model as both implementation are pretty much identical?

Notes:

- Mypy is likely to fail as I haven't installed the `torch-ccl` yet.
- Once I collect feedback on the approach, I will implement a set of tests for that backend. 

Check list:

### DIFF
--- a/ignite/distributed/comp_models/torch_ccl.py
+++ b/ignite/distributed/comp_models/torch_ccl.py
@@ -1,0 +1,170 @@
+import os
+from distutils.version import LooseVersion
+from typing import Any, Callable, Mapping, Optional, Tuple, cast
+
+import torch
+import torch.multiprocessing as mp
+import torch.nn.parallel
+
+from ignite.distributed.comp_models.native import _NativeDistModel
+
+try:
+    import torch.distributed as dist
+    import torch_ccl
+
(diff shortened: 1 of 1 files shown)


vfdev-5:
@Moh-Yakoub thanks for the PR !

Let's try to write a small use-case example at first and then we'll see what do we need and the blockers...

PR #1834, opened 2021-03-21
Add bleu metric

sdesrozis:
Description:

Add BLEU metric. Refactor to introduce `nlp` module in `metrics`. Rouge metric has been moved in `nlp`.

The implementation is inspired from `nltk`.

Check list:


ignite/metrics/__init__.py
```diff
@@ -43,9 +42,6 @@
     "PSNR",
     "Recall",
     "RootMeanSquaredError",
-    "Rouge",
```
vfdev-5, line 46: @sdesrozis I think it is still good to expose nlp metrics here

ignite/metrics/nlp/bleu.py
```diff
@@ -0,0 +1,184 @@
...
+        self.smoother = _Smoother(method=smooth)
+        super(Bleu, self).__init__(output_transform=output_transform, device=device)
+
+    def corpus_bleu(
+        self, references: Sequence[Sequence[Any]], candidates: Sequence[Sequence[Any]],
+    ):
+        p_numerators = Counter()  # Key = ngram order, and value = no. of ngram matches.
+        p_denominators = Counter()  # Key = ngram order, and value = no. of ngram in ref.
+
+        assert len(references) == len(candidates), "The number of hypotheses and their reference(s) should be the same "
```
vfdev-5, line 124: We should do if not something raise exception and avoid using assert...

vfdev-5 (commented):
Looks good @sdesrozis , thanks 
I have few comments here and 
do we need empty `ignite/metrics/nlp/utils.py` here ?

gucifer:
> Looks good @sdesrozis , thanks
> I have few comments here and
> do we need empty `ignite/metrics/nlp/utils.py` here ?

The utils.py file holds the modified_precision function used in compute_corpus method line 131.
Also, categorizing metrics into sub-modules sounds like a good idea in my opinion as in the future the number of metrics supported by ignite will only grow.

ignite/metrics/nlp/bleu.py
```diff
@@ -0,0 +1,184 @@
...
+        denominators = [max(1, d) for d in denominators.values()]
+        return [n / d if n != 0 else epsilon / d for n, d in zip(numerators.values(), denominators)]
+
+    @staticmethod
+    def nltk_smooth2(numerators: Counter, denominators: Counter) -> Sequence[float]:
+        denominators = [max(1, d) for d in denominators.values()]
+        return [(n + 1) / (d + 1) for n, d in zip(numerators.values(), denominators)]
+
+    @staticmethod
+    def smooth2(numerators: Counter, denominators: Counter) -> Sequence[float]:
```
gucifer, line 48: This is not giving different results than nltk_smooth2 for me. Tested it with 3-4 values. Am I missing something?
```
from collections import Counter


def nltk_smooth2(numerators: Counter, denominators: Counter):
    denominators = [max(1, d) for d in denominators.values()]
    return [(n + 1) / (d + 1) for n, d in zip(numerators.values(), denominators)]


def smooth2(numerators: Counter, denominators: Counter):
    return [(n + 1) / (d + 1) for n, d in zip(numerators.values(), denominators.values())]


a = [1, 1, 2, 2, 2, 3, 3, 3, 4, 5]
b = [1, 1, 1, 2, 2, 3, 3, 4, 5, 5]
num = Counter(a)
den = Counter(b)
print(num, den)
for n, d in zip(num.values(), den):
    print(n, d)
print("here")
for n, d in zip(num.values(), den.values()):
    print(n, d)
print(nltk_smooth2(num, den))
print(smooth2(num, den))
```

sdesrozis: It is similar because you have ngram matching counts.

By default, nltk do for the denominator 

https://github.com/nltk/nltk/blob/1f50455ef3d754996e983699352c4d7bfa8a07cb/nltk/translate/bleu_score.py#L330

So for `smooth2`, if denominator (and denominator) was zero, the precision is : `(0 + 1) / (1 + 1)` rather than `(0 + 0) / (0 + 0)`.

For details https://www.aclweb.org/anthology/W14-3346.pdf

ignite/metrics/nlp/bleu.py
sdesrozis: Yep! I missed that!

sdesrozis:
> Looks good @sdesrozis , thanks 
> 
> I have few comments here and 
> 
> do we need empty `ignite/metrics/nlp/utils.py` here ?

This file should not be empty!! Let me see.

ignite/metrics/nlp/bleu.py
gucifer: No, I mean the nltk_smooth2 and smooth2 give the same result so why do we have both the functions. Can you give me a test case where the result for nltk_smooth2 and smooth2 are different?

vfdev-5: I also wonder what would happen if counter's keys are different ? Can we encounter such situation : 
```
num = {1: 1, 2: 2}
den = {2: 1, 3: 2}
```

ignite/metrics/nlp/bleu.py
```diff
@@ -0,0 +1,186 @@
...
+            for i in range(1, self.ngrams_order + 1):
+                numerator, denominator = modified_precision(refs, hyp, i)
+                p_numerators[i] += numerator
+                p_denominators[i] += denominator
+
+        # Returns 0 if there's no matching n-grams
+        # We only need to check for p_numerators[1] == 0, since if there's
+        # no unigrams, there won't be any higher order ngrams.
+        if p_numerators[1] == 0:
+            return 0
```
vfdev-5, line 141: Can we add a test for this case here or in another PR ?

sdesrozis: I will look but I think that is covered.

ignite/metrics/nlp/bleu.py
sdesrozis: @gucifer 

```python
import pytest
import warnings
from nltk.translate.bleu_score import corpus_bleu, SmoothingFunction

from ignite.metrics.nlp import Bleu


with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    reference = corpus_bleu([[["0"]]], [["0"]], smoothing_function=SmoothingFunction().method2)

m = Bleu(ngram=4, smooth="smooth2")

m.update((["0"], [["0"]]))

assert pytest.approx(reference) != m.compute()

m = Bleu(ngram=4, smooth="nltk_smooth2")

m.update((["0"], [["0"]]))

assert pytest.approx(reference) == m.compute()
```

sdesrozis: > No, I mean the nltk_smooth2 and smooth2 give the same result so why do we have both the functions. Can you give me a test case where the result for nltk_smooth2 and smooth2 are different?



> I also wonder what would happen if counter's keys are different ? Can we encounter such situation :
> 
> ```
> num = {1: 1, 2: 2}
> den = {2: 1, 3: 2}
> ```

It can't happen with the algorithm because operator `&` is used.

sdesrozis:
@vfdev-5 did you ever experience error with flaky codecov failure ?

vfdev-5:
@sdesrozis yes, codecov is start failing today

trsvchn:
https://github.com/codecov/codecov-action/issues/234

vfdev-5 (approved):
Thanks for great PR @sdesrozis 💪 LGTM

PR #1835, opened 2021-03-21
Join labeler and welcome into single workflow

trsvchn:
Fixes #{issue number}

Description:

This workflow will be triggered only on PR/Issue opened: Greeting step on issue opened + PR opened. Labeling step on PR opened.

The aim is to reduce a number of unnecessary GH actions runs.

Check list:


vfdev-5 (approved):
Looks good, thanks

vfdev-5 (approved):
Thanks @trsvchn

.github/workflows/triage.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Pull Request Labeler
+#         if: github.event_name == 'pull_request_target'
+        uses: actions/labeler@v3
+        with:
+          configuration-path: .github/pr-labeler.yml
```
sdesrozis, line 17: Does `or-labeller.yml` still useful ? Because it has been removed in the PR

trsvchn: Yes, workflow file`.github/workflows/pr-labeler.yml` is removed, and action moved to `.github/workflows/triage.yml` but `.github/pr-labeler.yml` is a config for that action, containing info about labels. I think we cannot put invalid workflow yml file inside workflows dir, that's why config file lives there.

.github/workflows/triage.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Pull Request Labeler
+#         if: github.event_name == 'pull_request_target'
+        uses: actions/labeler@v3
+        with:
+          configuration-path: .github/pr-labeler-config.yml
```
vfdev-5, line 17: So, this is our local file right ? But we do not have any file like that ?

trsvchn: Now we have
https://github.com/pytorch/ignite/pull/1835/files#diff-c07d414af06346d8cbebb019db1e0079729e57365ed5954a77058bca6c53bb62

trsvchn: after discussion https://github.com/pytorch/ignite/pull/1835#discussion_r598448898
we decided to rename it

vfdev-5: OK, I see, my bad

Issue #1836, opened 2021-03-21
GitHub Actions Workflow Improvements 


trsvchn:
There are some elements in our CI that can be improved:

For example, `code-style.yml` starts very often in non-code updates. The same for `docs.yml` for non-docs.

`pytorch-version-tests.yml` fails to open the issue because of access problem:
https://github.com/pytorch/ignite/runs/2157534748?check_suite_focus=true#step:9:8

- [x] code-style
- [x] docs
- [x] unit-tests: https://github.com/pytorch/ignite/pull/1838#issuecomment-804796165

vfdev-5:
@trsvchn why there was a message for the first contribution here ?

trsvchn:
@vfdev-5 Yeah, I think its a bug and removed it.

trsvchn:
But honestly, it worries me now

trsvchn:
@vfdev-5 Because, It was my second issue, but the first issue is closed so there was no open issue by me, is this what action call "first contribution"?

vfdev-5:
Can you try that on a separate repo to see what it works exactly. For me "first contribution" is the first and only the first issue issue (same for PR).

trsvchn:
Yes, exactly I'll try that!

trsvchn:
@vfdev-5 I think I got it, I know what is happening.

According to action source code first contribution is:
you have no issues at all
```ts
        if (issues.length === 0) {
            return true;
        }
```
or current issue number less or equal to any of issue numbers (I am not familiar with TS): 
```ts
        for (const issue of issues) {
            if (issue.number < curIssueNumber && !issue.pull_request) {
                return false;
            }
        }
        return true;
```
I've checked the actions logs, and turned out that it added comment not on `issue: opened`, but on `issue: edited`.
In our case on issue: edit it was kind of `issue.number === curIssueNumber` (I hope I used right amount of "=" to express "equality" :) ) since issues already contains edited issue but it returns that `true` after the for-loop:

https://github.com/actions/first-interaction/blob/bd33205aa5c96838e10fd65df0d01efd613677c1/lib/main.js#L111

And that's why it treats it as first-time contrib.

But in updated version our workflow (#1835) is going to start only on issue:opened, so we expect this branch:
```ts
        if (issues.length === 0) {
            return true;
        }
```

vfdev-5:
>  The same for docs.yml for non-docs.

[...]

PR #1838, opened 2021-03-21
Improve `code-style` workflow by tweaking path filters

trsvchn:
Addresses #1836 

Description:

Replace `path-ignore` with `path` filters:

> A matching negative pattern (prefixed with !) after a positive match will exclude the path.

Check list:


vfdev-5:
@trsvchn thanks for the PR, but why it is an improvement ?

trsvchn:
@vfdev-5 Good question! I think in updated configuration this workflow starts only when py file was modifed, except specified dirs. And no startups on bash, yml, toml, or cfg updates

With old filters:

```yml
    paths-ignore:
      - "assets/**"
      - ".circleci/**"
      - "docker/**"
      - "docs/**"
      - "conda.recipe"
      - "**.md"
```

we trigger `code-style` checks on any kind of files, except `md` files + ignore some dirs. So, it is not python oriented, for example I can change some bash files in `tests` or `.github` dirs, or yaml files in `.github` etc. This PR is also an example.

For me, the main aim of this workflow is fixing only python files and ignore anything else.

If you prefer to keep `paths-ignore` style, we can try smth like:

```yml
    paths-ignore:
      - "assets/**"
      - ".circleci/**"
      - "docker/**"
      - "docs/**"
      - "conda.recipe"
      - "**.md"

      - ".github/**"
      - "**.sh"
```

What do you think?

.github/workflows/code-style.yml
```diff
@@ -2,29 +2,32 @@ name: Format python code
...
-      - "assets/**"
-      - ".circleci/**"
-      - "docker/**"
-      - "docs/**"
-      - "conda.recipe"
-      - "**.md"
-  #   pull_request:
-  #     paths-ignore:
+    paths:
+      - "**.py"
```
trsvchn, line 6:
```suggestion
      - "**.py"
      - "setup.cfg"
      - ".github/workflows/code-style.yml"
```

trsvchn: Add some non-py files related to code-style checks

vfdev-5:
Here is an example when we would like to trigger unit tests but modified only `requirements-dev.txt` and currently it does not run any tests: https://github.com/pytorch/ignite/pull/1863

vfdev-5:
Finally I have read and understood your comment and your point. I agree with this approach. Let's finalize the PR and merge it

vfdev-5 (approved):
Thanks !

PR #1839, opened 2021-03-21
Added `safe_mode` for Idist broadcast 

vfdev-5:
Description:
- Added `safe_mode` for Idist broadcast 

Check list:


sdesrozis:
@vfdev-5 I’m not sure about the use case.

As far I have understood, only one process holds a valid tensor and others have none. So the type is reduced to build a valid empty tensor (same style).

Build such empty tensor should be pretty easy for the user, shouldn’t it ?

vfdev-5:
@sdesrozis here is the use-case where we are blocked a bit: https://github.com/pytorch/ignite/pull/1758

ignite/distributed/utils.py
```diff
@@ -361,13 +361,20 @@ def all_gather(tensor: Union[torch.Tensor, float, str]) -> Union[torch.Tensor, f
     return _model.all_gather(tensor)
 
 
-def broadcast(tensor: Union[torch.Tensor, float, str], src: int = 0) -> Union[torch.Tensor, float, str]:
+def broadcast(
+    tensor: Union[torch.Tensor, float, str, None], src: int = 0, use_none: bool = False
```
sdesrozis, line 365: `use_none` is not very expressive. Why not something like `allow_none` or `allow_undefined_tgt_tensors` or `safe_to_xxx` ? That is purely a matter of taste !!

ignite/distributed/utils.py
```diff
@@ -361,13 +361,20 @@ def all_gather(tensor: Union[torch.Tensor, float, str]) -> Union[torch.Tensor, f
...
+) -> Union[torch.Tensor, float, str]:
     """Helper method to perform broadcast operation.
 
     Args:
         tensor: tensor or number or str to broadcast to participating processes.
-            Make sure to respect dtype of torch tensor input for all processes, otherwise execution will crash.
+            Make sure to respect data type of torch tensor input for all processes, otherwise execution will crash.
```
sdesrozis, line 371: Why not add an example with `use_none` argument ?

sdesrozis (approved):
@vfdev-5 Perfect ! Maybe a matter of taste about the argument name `use_none` which is not very expressive but examples provided are.

vfdev-5:
@sdesrozis any other better ideas on how to name it :) ?

sdesrozis:
> @sdesrozis any other better ideas on how to name it :) ?

`use_none_target_tensors` ?

Moreover, we could have a specific safe broadcast mode replacing every target tensors to avoid errors.

vfdev-5:
> > @sdesrozis any other better ideas on how to name it :) ?
> 
> `use_none_target_tensors` ?

we can broadcast not only tensors but also numbers and strings

> Moreover, we could have a specific safe broadcast mode replacing every target tensors to avoid errors.

what do you mean by "replacing every target tensors to avoid errors." ?

sdesrozis:
> what do you mean by "replacing every target tensors to avoid errors." ?

Stupid scenario
```python
if rank == 0:
    t = tensor.empty((1,))
if rank == 1:
    t = 0
if rank == src: # src != 0 & src != 1
    t = 3
t = idist.broadcast(t, src=src, use_none=True) # here, max is tensor.empty (I suppose)
```

vfdev-5:
@sdesrozis i renamed `use_none` to `safe_mode` and now handle also the case you talked about: https://github.com/pytorch/ignite/pull/1839#issuecomment-804877954

Could you please review again, thanks !

sdesrozis (approved):
@vfdev-5 LGTM ! Safe mode with discard is nice !

PR #1842, opened 2021-03-22
fix: run `test_auto_methods_nccl` if ws > 1

ydcjeff:
Fixes failing master

Description:

Check list:


vfdev-5:
@ydcjeff any idea why it started failing now and not before ?

PR #1846, opened 2021-03-22
fix: run `test_auto_methods_nccl` if ws > 1 (#1842)

vfdev-5:
Original PR : #1842 by @ydcjeff 

* fix: run test_auto_methods_nccl if ws > 1

* chore: use auto_model instead of _test_auto_model

Fixes #{issue number}

Description:

Check list:


vfdev-5:
Thanks for the fix @ydcjeff !

PR #1847, opened 2021-03-22
ci: run pip as module for all os

ydcjeff:
Description:
Fixes fail to get python wheel cache in ubuntu CI.
ubuntu has been updated to 20.04, we can just use `pip`
For to work with cross platform, we need to with `python3 -m` as macos still has python2.7

Check list:


trsvchn:
@ydcjeff Thank you for your PR. I am just wondering about windows, since I was facing this kind of issue when running `python3 -m venv venv` on `windows-latest`:
```
/c/hostedtoolcache/windows/Python/3.7.9/x64/python3
Error: [WinError 2] The system cannot find the file specified
```
See: https://github.com/msys2/MINGW-packages/issues/5001#issue-414166629

So, I am not sure about `pip`, maybe it was `venv` issue or smth, did you try it on `windows-latest`

ydcjeff:
If use with `setup-python`, `python` alone is enough and it works

vfdev-5 (approved):
Thanks @ydcjeff

PR #1860, opened 2021-03-23
fix: call `.cpu()` before `.numpy()`, use cuda device 0

### LINKED ISSUE #1851: Fix failing test if cuda device is available
## 🐛 Bug description

If we execute the following tests:
```bash
pytest tests/ignite/metrics/test_psnr.py::test_psnr

> FAILED tests/ignite/metrics/test_psnr.py::test_psnr - TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

```bash
pytest tests/ignite/metrics/test_ssim.py::test_ssim

> FAILED tests/ignite/metrics/test_ssim.py::test_ssim - AssertionError: assert device(type='cuda', index=0) == device(type='cuda')
```

This is due to a presence of a GPU which is not properly handled here. On the CPU CI we set up `CUDA_VISIBLE_DEVICES=` to disable any gpu. 
Let's fix these tests.

cc @ydcjeff

ydcjeff:
Fixes #1851 

Description:
- Call `.cpu` before `.numpy()`
- Force use `cuda:0`

Check list:

### DIFF
--- a/tests/ignite/metrics/test_psnr.py
+++ b/tests/ignite/metrics/test_psnr.py
@@ -44,11 +44,11 @@ def _test_psnr(y_pred, y, data_range, device):
     assert isinstance(psnr_compute, torch.Tensor)
     assert psnr_compute.dtype == torch.float64
     assert psnr_compute.device == torch.device(device)
-    assert np.allclose(psnr_compute.numpy(), np_psnr / np_y.shape[0])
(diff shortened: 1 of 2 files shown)


sdesrozis (approved):
@ydcjeff Thank you for this catch

PR #1861, opened 2021-03-23
ci: separate create-issue job for pytorch version tests workflow [skip ci]

ydcjeff:
Fixes `Schedule workflow failed` issue being created multiple times
https://github.com/pytorch/ignite/pull/1664#issuecomment-804498443

https://github.com/pytorch/ignite/actions/runs/678571222

Description:

Check list:

### DIFF
--- a/.github/workflows/pytorch-version-tests.yml
+++ b/.github/workflows/pytorch-version-tests.yml
@@ -66,7 +66,7 @@ jobs:
     runs-on: ubuntu-latest
     # https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#needs-context
     needs: build
-    if: always() && needs.build.result != 'success'
+    if: always() && needs.build.result == 'failure'
     steps:
       - uses: actions/checkout@v2
       - uses: JasonEtco/create-an-issue@v2
--- a/.github/workflows/pytorch-version-tests.yml
+++ b/.github/workflows/pytorch-version-tests.yml
@@ -67,7 +67,7 @@ jobs:
     runs-on: ubuntu-latest
     # https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#needs-context
     needs: build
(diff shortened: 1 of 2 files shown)


vfdev-5 (approved):
Thanks a lot @ydcjeff !

.github/workflows/pytorch-version-tests.yml
```diff
@@ -62,9 +62,15 @@ jobs:
         run: |
           pytest tests -vvv
 
+  create-issue:
+    runs-on: ubuntu-latest
+    # https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#needs-context
+    needs: build
+    if: always() && needs.build.result != 'success'
```
vfdev-5, line 69: Can't we use `failure()` instead of `always() && ...` : https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#failure ?

ydcjeff: I have tried in the first commit and `create-issue` job doesn't run with `always()`

vfdev-5: Why can't we do : `if: needs.build.result == 'failure'` ?

ydcjeff: it would miss `cancelled` and `skipped` result.

vfdev-5: I think it is OK to skip `cancelled` as it is maintainer's action. Good point about "skipped" but I'm not sure about in which case it can appear and we'd like to be notified...

ydcjeff: So `== 'failure'` or `!= 'success'` ?

vfdev-5: `== 'failure'` is the winner :)

PR #1866, opened 2021-03-23
fix `Scheduled workflow failed` error

### LINKED ISSUE #1852: Scheduled workflow failed
Oh no, something went wrong in the scheduled workflow **PyTorch version tests/build with commit 83d663d6768e935f2dc4ef583608c4411b848aa3**.
Please look into it:

https://github.com/pytorch/ignite/actions/runs/677894582

Feel free to close this if this was just a one-off error.

Moh-Yakoub:
Fixes #1852

Description: One of the workflows were failing because the old versions didn't have the autocast method - https://github.com/pytorch/ignite/runs/2171095975?check_suite_focus=true - 
To allow backward compatibility I've added a try block to catch that attribute error so that it won't break the tests for the older versions.

Check list:

### DIFF
--- a/tests/ignite/engine/test_create_supervised.py
+++ b/tests/ignite/engine/test_create_supervised.py
@@ -460,35 +460,32 @@ def test_create_supervised_trainer_on_cuda_with_model_on_cpu():
 def test_create_supervised_evaluator():
     _test_create_supervised_evaluator()
     _test_mocked_supervised_evaluator()
-    try:
-        # older versions didn't have the autocast method so we skip the test
+
+    # older versions didn't have the autocast method so we skip the test for older builds
+    if LooseVersion(torch.__version__) > LooseVersion("1.6.0"):
         with mock.patch("torch.cuda.amp.autocast") as mock_torch_cuda_amp_module:
             _test_create_evaluation_step_amp(mock_torch_cuda_amp_module)
-    except AttributeError:
-        pass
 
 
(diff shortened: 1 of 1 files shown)


tests/ignite/engine/test_create_supervised.py
```diff
@@ -460,23 +460,35 @@ def test_create_supervised_trainer_on_cuda_with_model_on_cpu():
 def test_create_supervised_evaluator():
     _test_create_supervised_evaluator()
     _test_mocked_supervised_evaluator()
-    with mock.patch("torch.cuda.amp.autocast") as mock_torch_cuda_amp_module:
-        _test_create_evaluation_step_amp(mock_torch_cuda_amp_module)
+    try:
```
vfdev-5, line 463: Maybe we could simply skip those subtests with [pytest.skip](https://docs.pytest.org/en/stable/skipping.html#skipping-test-functions) similarly to `pytest.mark.skipif(LooseVersion(torch.__version__) < LooseVersion("1.6.0"), reason="Skip if < 1.6.0")` (without decorator) ?
```python
def test_function():
    if not valid_config():
        pytest.skip("unsupported configuration")
```

Moh-Yakoub: That's way cleaner actually, will do this now

vfdev-5 (approved):
LGTM, thanks @Moh-Yakoub !

PR #1869, opened 2021-03-24
Improve WHM metric for DDP

KickItLikeShika:
Improving `WaveHedgesDistance` implementation to be compatible with DDP and add the necessary tests.

Check list:


vfdev-5 (approved):
Looks good @KickItLikeShika ! Thanks

PR #1880, opened 2021-03-26
Added full read permissions to saved file (1876)

### LINKED ISSUE #1876: Checkpoint handler save files with wrong permissions
## 🐛 Bug description

Hi @vfdev-5 ,

We always run programs in dockers, usually it's root user, and found an issue that: pure PyTorch program saves model(torch.save) and has the read permission for non-root users, but ignite Checkpoint handler saves model and only has read permission for root users.
You can reproduce this issue by below 2 MONAI examples:
1. https://github.com/Project-MONAI/tutorials/blob/master/3d_segmentation/torch/unet_training_dict.py
The saved model will be:
```
-rw-r--r-- 1 root root 19253862 Mar 26 02:08 best_metric_model_segmentation3d_dict.pth
```
2. https://github.com/Project-MONAI/tutorials/blob/master/3d_segmentation/ignite/unet_training_dict.py
The saved model will be:
```
-rw------- 1 root root 57738580 Mar 26 01:54 net_checkpoint_50.pt
```

ignite version is 0.4.4, could you please help take a look at this issue?
As we are close to MONAI and Clara next release, it's an urgent bug for us.

Thanks in advance.

vfdev-5:
Fixes #1876 

Description:
- Added full read rights to saved file when atomic=True

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -1,6 +1,7 @@
 import collections.abc as collections
 import numbers
 import os
+import stat
 import tempfile
 import warnings
 from abc import ABCMeta, abstractmethod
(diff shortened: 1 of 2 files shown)


sdesrozis (approved):
It looks good !

sdesrozis:
It misses mkl libs for tpu...

vfdev-5:
I think it is the new pytorch release that fails our ci :)

PR #1881, opened 2021-03-26
Fix GPU failures

KickItLikeShika:
Add a hack to avoid results inconsistency between `np.median()` and `torch.median()`
Check list:

### DIFF
--- a/tests/ignite/contrib/metrics/regression/test_median_absolute_percentage_error.py
+++ b/tests/ignite/contrib/metrics/regression/test_median_absolute_percentage_error.py
@@ -131,9 +131,15 @@ def _test_distrib_compute(device):
 
         res = m.compute()
 
+        e = torch.abs(y - y_pred) / torch.abs(y)
+        np_res_like_torch = 100.0 * np.median(e.cpu().numpy())
+
         np_res = 100.0 * np.median(np.abs(np_y - np_y_pred) / np.abs(np_y))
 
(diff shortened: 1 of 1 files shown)


tests/ignite/contrib/metrics/regression/test_median_absolute_percentage_error.py
```diff
@@ -131,9 +131,15 @@ def _test(metric_device):
 
         res = m.compute()
 
+        e = torch.abs(y - y_pred) / torch.abs(y)
+        np_res_like_torch = 100.0 * np.median(e.cpu().numpy())
```
vfdev-5, line 135: Actually, why this is like pytorch ? The difference is between np.median and torch.median ?
I think there is no difference between : `np_res_like_torch` and `np_res`.

vfdev-5 (commented):
Thanks @KickItLikeShika for the PR.
I think we should generalize that for other tests where median related metrics are checked

tests/ignite/contrib/metrics/regression/test_median_absolute_percentage_error.py
KickItLikeShika: Mostly yes, but how should we do that?

vfdev-5:
@KickItLikeShika thanks for the update, I'll merge it to new branch to launch GPU tests.

PR #1882, opened 2021-03-26
Fix GPU failures (#1881)

vfdev-5:
Original PR #1881 

Check list:


vfdev-5:
@KickItLikeShika if tests are passing here, let's apply the same trick to other tests of median related metrics.

KickItLikeShika:
@vfdev-5 unfortunately it didn't work..

KickItLikeShika:
I found the error!  we did `np_res_prepend = 100.0 * np.median(e)` but it should be np_res_prepend = 100.0 * np.median(e_prepend)

KickItLikeShika:
@vfdev-5 It works fine now, thanks for providing such inspiring ideas! Thanks for your help!

PR #1883, opened 2021-03-26
Fix MAPE test for GPU failure

KickItLikeShika:
Add a hack to avoid results inconsistency between `np.median()` and `torch.median()`
Check list:

### DIFF
--- a/tests/ignite/contrib/metrics/regression/test_median_absolute_percentage_error.py
+++ b/tests/ignite/contrib/metrics/regression/test_median_absolute_percentage_error.py
@@ -135,7 +135,7 @@ def _test_distrib_compute(device):
         np_res = 100.0 * np.median(e)
 
         e_prepend = np.insert(e, 0, e[0], axis=0)
-        np_res_prepend = 100.0 * np.median(e)
+        np_res_prepend = 100.0 * np.median(e_prepend)
 
         # The results between numpy.median() and torch.median() are Inconsistant
         # when the length of the array/tensor is even. So this is a hack to avoid that.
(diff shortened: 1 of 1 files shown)


tests/ignite/contrib/metrics/regression/test_median_absolute_percentage_error.py
```diff
@@ -135,7 +135,7 @@ def _test(metric_device):
         np_res = 100.0 * np.median(e)
 
         e_prepend = np.insert(e, 0, e[0], axis=0)
-        np_res_prepend = 100.0 * np.median(e)
```
vfdev-5, line 138: I didn't notice that if we now compute the result as torch.median, no need to match between two results ?

KickItLikeShika: Yes we can do that, but the assertion here is between `res (computed by ignite)` and `np_res (computed by numpy)`, and that's what are we doing now (comparing between value computed by torch and value computed by numpy)

KickItLikeShika: >>> no need to match between two results ?
Can you explain that more?

vfdev-5: Do we need this : 
```
assert pytest.approx(res) in (np_res, np_res_prepend)
# vs
# assert pytest.approx(res) in np_res_prepend
```

KickItLikeShika: Well, i guess using the matching between two results is good to show that this `np_res_prepend` is a special case, what do you think?

vfdev-5: In this case, I'd do if something then check with this result otherwise check with another one.

KickItLikeShika: Something like this?
```
if idist.get_world_size() % 2 == 0:
        assert pytest.approx(res) == np_res_prepend
else:
        assert pytest.approx(res) == np_res
```

vfdev-5: Once again, ws is almost always even :) Second is better.

vfdev-5 (approved):
LGTM, thanks

PR #1887, opened 2021-03-26
Adding classification report metric

### LINKED ISSUE #1715: Adding Classification report utility function
## 🚀 Feature

I'd like to add the classification report utility to ignite. Similar to the same functionality in scikit-learn https://scikit-learn.org/stable/modules/generated/sklearn.metrics.classification_report.html#sklearn.metrics.classification_report

Moh-Yakoub:
Fixes #1715

Description:

Adding classification report contrib metric to use the corresponding sklearn one https://scikit-learn.org/stable/modules/generated/sklearn.metrics.classification_report.html#sklearn.metrics.classification_report and adding corresponding tests.

Check list:


ignite/contrib/metrics/classification_report.py
```diff
@@ -0,0 +1,71 @@
...
+    sklearn.metrics.classification_report.html`_ .
+
+    Args:
+        output_transform: a callable that is used to transform the
+            :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
+            form expected by the metric. This can be useful if, for example, you have a multi-output model and
+            you want to compute the metric with respect to one of the outputs.
+        labels: optional list of label indices to include in the report
+        check_compute_fn: Default False. If True, `classification-report
+            <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.classification-report.html>`_
```
vfdev-5, line 20: Bad link here

vfdev-5 (commented):
Thanks for the PR @Moh-Yakoub,
however, I'm not sure about that implementation as epoch metric. As suggested in the issue, we'd like to reuse ignite's precision, recall and fbeta implementations to create the report.

ignite/contrib/metrics/classification_report.py
Moh-Yakoub: @vfdev-5 I thought of reusing the existing metrics we have, but the sklearn classification report is more mature and also add a lot of other dimensions including support/averages and some other functionality. for me it made sense to include it like the other contrib metrics we already have.

sdesrozis: @Moh-Yakoub Thank you for your work. However, I'm not sure that using `sklearn` and `EpochMetric` is the best option. It implies much communication and having all predictions locally. Think of `ImageNet` usecase which could benefits from this report. 

Use our metrics (device aware, distributed, based on PyTorch tensors) makes really more sense for me. The main point is report in a smart way the results from `Precision`, `Recall` and `F1Score` (per class and averaged). Using `lambdaMetric` as @vfdev-5 mentioned would be straightforward.

What do you think ?

Moh-Yakoub: I will prepare an implementation with `lambdaMetric` and test how it would look like.

ignite/contrib/metrics/classification_report.py
```diff
@@ -0,0 +1,72 @@
...
+            check_compute_fn=check_compute_fn,
+            device=device,
+        )
+
+    def get_classification_report_fn(self) -> Callable[[torch.Tensor, torch.Tensor], Union[str, dict]]:
+        """Return a function computing classification report from scikit-learn."""
+        from sklearn.metrics import classification_report
+
+        def wrapper(y_targets: torch.Tensor, y_preds: torch.Tensor) -> Union[str, dict]:
+            y_true = y_targets.numpy()
```
KickItLikeShika, line 68: Add `.cpu()` here

ignite/contrib/metrics/classification_report.py
Moh-Yakoub: @sdesrozis  @vfdev-5  I was thinking about that implementation using `MetricsLambda`

At the moment the issues I am having are the following:

1. We don't support precision/recall per label. The main functionality of the sklean method is to generate those metrics per label. that means the only way to go around this in our library is to probably use an `output_transform`.

2. MetricsLambda doesn't automatically update recursively the underlying metrics (https://pytorch.org/ignite/master/generated/ignite.metrics.metrics_lambda.MetricsLambda.html) that means we have to update them by extending the update method. That's not a big issue though and straightforward to implement.

Let me know what you think of this

sdesrozis: @Moh-Yakoub I agree that this PR implies some improvements.

First, I suggest to write a small piece of code doing the report using our metrics and maybe some specific parts as you mentioned. We will see in a second stage what would be designed in a metric and how we would do it.

What do you think ?

vfdev-5: > We don't support precision/recall per label. 

This is incorrect, please take a look here: https://pytorch.org/ignite/master/generated/ignite.metrics.precision.Precision.html attribute `average`.

Moh-Yakoub: @vfdev-5 I think there's a misunderstanding due to the confusion between classes and labels. 

What I meant is this:

Using Sklearn.
```
y_true = [1, 1, 0, 1, 1]
y_pred = [0, 1, 1, 0, 0]
classification_report(y_true, y_pred, output_dict= True)
{'0': {'precision': 0.0, 'recall': 0.0, 'f1-score': 0.0, 'support': 1}, '1': {'precision': 0.5, 'recall': 0.25, 'f1-score': 0.3333333333333333, 'support': 4}, 'accuracy': 0.2, 'macro avg': {'precision': 0.25, 'recall': 0.125, 'f1-score': 0.16666666666666666, 'support': 5}, 'weighted avg': {'precision': 0.4, 'recall': 0.2, 'f1-score': 0.26666666666666666, 'support': 5}}
```

Now, if I use a metrics lambda using `Precision` and `Recall` with `false` `average`s what I get is this

```
{'precision': tensor([0.2500], dtype=torch.float64), 'recall': tensor([0.5000], dtype=torch.float64)}
```

So what happens is that yes, they returned a tensor but in reality it didn't support it per label - unless I am mis-configuring something here.

vfdev-5: @Moh-Yakoub if you'd like to treat `y_pred` as multiclass and not binary it should have `(N, 2)` shape:
```python
import torch

from ignite.metrics import Precision, Recall


pr = Precision(average=False)
re = Recall(average=False)
f1 = pr * re * 2 / (pr + re + 1e-20)

y_true = torch.tensor([1, 1, 0, 1, 1])
y_pred = torch.tensor([
    [1.0, 0.0], 
    [0.0, 1.0],
    [0.0, 1.0],
    [1.0, 0.0],
    [1.0, 0.0]
])

pr.update((y_pred, y_true))
re.update((y_pred, y_true))

print("Precision per class:", pr.compute())
print("Recall per class:", re.compute())
print("F1 per class:", f1.compute())
```
>
```
Precision per class: tensor([0.0000, 0.5000], dtype=torch.float64)
Recall per class: tensor([0.0000, 0.2500], dtype=torch.float64)
F1 per class: tensor([0.0000, 0.3333], dtype=torch.float64)
```

Moh-Yakoub: @vfdev-5  @sdesrozis  I've updated the implementation to use our functions + metrics lambda. I'd appreciate if you can take a look at it. once you approve the approach, I will add documentation and fully implement all tests.

sdesrozis: @Moh-Yakoub Your implementation looks good !

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,63 @@
...
+        self,
+        output_dict: bool = False,
+        output_transform: Optional[Callable] = None,
+        device: Union[str, torch.device] = torch.device("cpu"),
+        labels: Optional[List[str]] = None,
+    ):
+        self.precision = Precision(average=False, device=device,)
+        self.recall = Recall(average=False, device=device,)
+        self.averaged_precision = Precision(average=True, device=device,)
+        self.averaged_recall = Recall(average=True, device=device,)
```
vfdev-5, line 24: Can't reuse previous definitions with metrics arithmetics ?
```python
self.averaged_precision = self.precision.mean()
```

SImilarly, we can define F1 or reuse already present `metrics.fbeta` implementation.

vfdev-5 (commented):
Thanks for the update @Moh-Yakoub

ignite/metrics/classification_report.py
sdesrozis: Yep, `F1` score would be great !

Moh-Yakoub: Great, will add them with docs and test cases. Thanks for the review!

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+            y_true = torch.tensor([[1, 0],
+            [0, 1], [1, 0], [1, 1], [1, 0], [0, 1], [0, 1], [1, 0], [0, 1], [1, 0]])
+            y_pred = torch.randint(0, 2, size=(10,))
+
+            classification_report = ClassificationReport(output_dict=output_dict)
+            classification_report.update((y_true, y_pred))
+            res = classification_report.compute()
+
```
vfdev-5, line 40: Could you please give the output for the example, such that user have an idea of what that does.

vfdev-5 (commented):
Thanks for the updates @Moh-Yakoub 
I'll check it in details later. Just a small comment on docstring here.

Moh-Yakoub:
@vfdev-5 @sdesrozis Let me know what you think of the current implementation. Thanks

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,142 @@
...
+            y_true = torch.tensor([[1, 0],
+            [0, 1], [1, 0], [1, 1], [1, 0], [0, 1], [0, 1], [1, 0], [0, 1], [1, 0]])
+            y_pred = torch.randint(0, 2, size=(10,))
+
+            classification_report = ClassificationReport(output_dict=True, digits=2)
+            classification_report.update((y_true, y_pred))
+            res = classification_report.compute()
+
+            result should look like this:
```
vfdev-5, line 42:
```suggestion
            # result should look like this:
```

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,142 @@
...
+from ignite.metrics.fbeta import Fbeta
+from ignite.metrics.metric import Metric, reinit__is_reduced
+from ignite.metrics.metrics_lambda import MetricsLambda
+from ignite.metrics.precision import Precision
+from ignite.metrics.recall import Recall
+
+__all__ = ["ClassificationReport"]
+
+
+class ClassificationReport(MetricsLambda):
```
vfdev-5, line 15: Previously, we used methods to create composed metrics like this one. For example, FBeta, IoU, mIoU etc.
Do you think we can define `ClassificationReport` as a method instead of deriving from `MetricsLambda`. It seems to me that inheritance here is very artificial...

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,142 @@
...
+        Args:
+            beta: weight of precision in harmonic mean
+            output_dict: If True, return output as dict, otherwise return a str
+            output_transform: a callable that is used to transform the
+                :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
+                form expected by the metric. This can be useful if, for example, you have a multi-output model and
+                you want to compute the metric with respect to one of the outputs.
+            device: optional device specification for internal storage.
+            labels: Optional list of label indices to include in the report
+            digits: Number of digits for formatting output floating point values, by default it is 5
```
vfdev-5, line 29: In scikit-learn this option is for the output as str and no truncation for the output as dict. I think either we should propose the same or remove `digits` argument... What do you think ?

vfdev-5 (commented):
Thanks a lot @Moh-Yakoub for a great work on this metric ! I have few comments on if we could make it even nicer. Please let me know what do you think

ignite/metrics/classification_report.py
Moh-Yakoub: Good point I will remove the digits argument

ignite/metrics/classification_report.py
Moh-Yakoub: I agree on that point, my only issue with having a method returning the metrics lambda is with the update method. Metrics lambda update won't update underlying metrics automatically. it worked in the fbeta because we were computing final averages. I am not sure if it'd be possible here.

ignite/metrics/classification_report.py
vfdev-5: @Moh-Yakoub please hit "Resolve conversation" if the comment is/will be addressed and does not need more discussion. It is a convention that could be practical to follow when we contribute to OSS projects. Thanks.

ignite/metrics/classification_report.py
vfdev-5: Yes, that's true however there are two points here:
- MetricLambda does not work as user would expect with reset/update/compute API : 
https://github.com/pytorch/ignite/blob/f9504f1cea9783508d7ff6d852588167b2f4fa6f/ignite/metrics/metrics_lambda.py#L82-L87

It expects to be attached to the engine or to call update on ingredients. You are right, we can not call update on precision/recall/fbeta separately. This would be something to work on in another PR. If you could create an issue for that, I'd really appreciate.

- let's make this metric as a method and make it work with attach to engine API. Once `MetricLambda` can work with reset/update/compute API, this will be automatically inherited ...

Moh-Yakoub:
@vfdev-5  I've refactored the report and used a method instead of extending the class. In that way it's similar to `Fbeta` let me know if you have any furthur thoughts on this.

Moh-Yakoub:
@sdesrozis Any thoughts about the current implementation. I've adapted it to use a method instead of extending Metric?

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,108 @@
...
+
+from ignite.metrics.fbeta import Fbeta
+from ignite.metrics.metric import Metric
+from ignite.metrics.metrics_lambda import MetricsLambda
+from ignite.metrics.precision import Precision
+from ignite.metrics.recall import Recall
+
+__all__ = ["ClassificationReport"]
+
+"""Build a text report showing the main classification metrics. The report resembles in functionality
```
vfdev-5, line 14: Let's put the documentation into the method's docstring.

tests/ignite/metrics/test_classification_report.py
```diff
@@ -0,0 +1,297 @@
...
+        engine.run(data=data)
+
+        assert "cr" in engine.state.metrics
+        res = engine.state.metrics["cr"]
+        res2 = classification_report.compute()
+        assert res == res2
+
+        assert isinstance(res, dict if output_dict else str)
+        if not output_dict:
+            res = ast.literal_eval(res)
```
vfdev-5, line 155: Can you please explain why we need to use ast here ?

Moh-Yakoub: I've replaced it with the more standard `json` module

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,107 @@
...
+    fbeta = Fbeta(beta, average=False, precision=precision, recall=recall)
+    averaged_precision = precision.mean()
+    averaged_recall = recall.mean()
+    averaged_fbeta = fbeta.mean()
+
+    def _wrapper(
+        recall_metric: Metric, precision_metric: Metric, f: Metric, a_recall: Metric, a_precision: Metric, a_f: Metric,
+    ) -> Union[Collection[str], Dict]:
+        p_tensor, r_tensor, f_tensor = precision_metric, recall_metric, f
+        assert p_tensor.shape == r_tensor.shape
```
vfdev-5, line 88: Let's replace assert with `if not something: raise ValueError`

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,107 @@
...
+from ignite.metrics.precision import Precision
+from ignite.metrics.recall import Recall
+
+__all__ = ["ClassificationReport"]
+
+
+def ClassificationReport(
+    beta: int = 1,
+    output_dict: bool = False,
+    output_transform: Optional[Callable] = None,
```
vfdev-5, line 18: Let's use `output_transform` same as default one for `Metric` -> we can simply pass it below without
```python
output_transform=(lambda x: x) if output_transform is None else output_transform
```

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,107 @@
...
+    device: Union[str, torch.device] = torch.device("cpu"),
+    labels: Optional[List[str]] = None,
+) -> MetricsLambda:
+    """Build a text report showing the main classification metrics. The report resembles in functionality
+        'https://scikit-learn.org/stable/modules/generated/sklearn.metrics.classification_report.html#sklearn.metrics.classification_report'
```
vfdev-5, line 23:
```suggestion
    """Build a text report showing the main classification metrics. The report resembles in functionality to
         `scikit-learn classification_report <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.classification_report.html#sklearn.metrics.classification_report>`_
```

ignite/metrics/classification_report.py
vfdev-5: Please remove `digits` here

vfdev-5:
@Moh-Yakoub thanks for the update, it remains few things to update and will be good to go.

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,107 @@
...
+from ignite.metrics.fbeta import Fbeta
+from ignite.metrics.metric import Metric
+from ignite.metrics.metrics_lambda import MetricsLambda
+from ignite.metrics.precision import Precision
+from ignite.metrics.recall import Recall
+
+__all__ = ["ClassificationReport"]
+
+
+def ClassificationReport(
```
sdesrozis, line 15: It’s not a valid function name regarding pep8 naming rules.

vfdev-5: @sdesrozis true, but we are already doing that for IoU, mIoU, Fbeta functions ...

Moh-Yakoub: > @sdesrozis true, but we are already doing that for IoU, mIoU, Fbeta functions ...

I agree, so I kept it for consistency, otherwise we can rename all of them, but I believe it will be too much work.

ignite/metrics/classification_report.py
vfdev-5: @Moh-Yakoub ^^

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,101 @@
...
+    averaged_precision = precision.mean()
+    averaged_recall = recall.mean()
+    averaged_fbeta = fbeta.mean()
+
+    def _wrapper(
+        recall_metric: Metric, precision_metric: Metric, f: Metric, a_recall: Metric, a_precision: Metric, a_f: Metric,
+    ) -> Union[Collection[str], Dict]:
+        p_tensor, r_tensor, f_tensor = precision_metric, recall_metric, f
+        if p_tensor.shape != r_tensor.shape:
+            raise ValueError("Precision and Recall have mismatched shapes")
```
vfdev-5, line 81: In which case this can happen what user can do with that ? I think the error message is not clear about that.
If it is an internal check let's say that as 
```
f"Internal error: Precision and Recall have mismatched shapes: {p_tensor.shape} vs {r_tensor.shape}. "
"Please, open an issue with a refence on this error."
```

Moh-Yakoub: Great point, updating it now.

ignite/metrics/classification_report.py
```diff
@@ -78,7 +78,11 @@ def _wrapper(
     ) -> Union[Collection[str], Dict]:
         p_tensor, r_tensor, f_tensor = precision_metric, recall_metric, f
         if p_tensor.shape != r_tensor.shape:
-            raise ValueError("Precision and Recall have mismatched shapes")
+            raise ValueError(
+                f"Internal error: Precision and Recall have mismatched shapes: "
+                f"{p_tensor.shape} vs {r_tensor.shape}. Please, open an issue "
+                f"with a reference on this error."
```
vfdev-5, line 84:
```suggestion
                "Internal error: Precision and Recall have mismatched shapes: "
                f"{p_tensor.shape} vs {r_tensor.shape}. Please, open an issue "
                "with a reference on this error. Thank you!"
```

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,104 @@
...
+            labels: Optional list of label indices to include in the report
+
+        .. code-block:: python
+
+
+            y_true = torch.tensor([[1, 0],
+            [0, 1], [1, 0], [1, 1], [1, 0], [0, 1], [0, 1], [1, 0], [0, 1], [1, 0]])
+            y_pred = torch.randint(0, 2, size=(10,))
+
+            classification_report = ClassificationReport(output_dict=True, digits=2)
```
vfdev-5, line 44:
```suggestion
            classification_report = ClassificationReport(output_dict=True)
```

tests/ignite/metrics/test_classification_report.py
```diff
@@ -0,0 +1,297 @@
...
+    torch.manual_seed(12)
+
+    def _test(metric_device):
+
+        n_iters = 80
+        s = 16
+        offset = n_iters * s
+        y_true = torch.randint(0, 2, size=(offset * idist.get_world_size(),)).to(device)
+        y_preds = torch.randint(0, 2, size=(offset * idist.get_world_size(),)).to(device)
+        y_true_unflat = _unflatten_binary(y_true)
```
vfdev-5, line 24: I do not quite understand why input data is of that form. At first, it is not a binary case but 2 classes multi-class...

tests/ignite/metrics/test_classification_report.py
```diff
@@ -0,0 +1,297 @@
...
+    for _ in range(5):
+        # check multiple random inputs as random exact occurencies are rare
+        metric_devices = ["cpu"]
+        if device.type != "xla":
+            metric_devices.append(idist.device())
+        for metric_device in metric_devices:
+            _test(metric_device)
+
+
+def _test_integration_binary_labels(device, labels):
```
vfdev-5, line 68: This can be refactored with `_test_integration_binary` I think

tests/ignite/metrics/test_classification_report.py
```diff
@@ -0,0 +1,297 @@
...
+    for _ in range(5):
+        # check multiple random inputs as random exact occurencies are rare
+        metric_devices = ["cpu"]
+        if device.type != "xla":
+            metric_devices.append(idist.device())
+        for metric_device in metric_devices:
+            _test(metric_device, 3)
+
+
+def _test_integration_multilabel_with_labels(device, output_dict):
```
vfdev-5, line 178: Same here about refactoring with `_test_integration_multilabel`

vfdev-5 (commented):
@Moh-Yakoub thanks a lot for this PR! I understand that it took almost 3 weeks to you to implement, rework and add all the comments. 
However, more I look to the test and I think we have to pause a bit on that as I think I have to review it more carefully.

Moh-Yakoub:
> @Moh-Yakoub thanks a lot for this PR! I understand that it took almost 3 weeks to you to implement, rework and add all the comments.
> However, more I look to the test and I think we have to pause a bit on that as I think I have to review it more carefully.

Sure no worries, feel free to review the tests thoroughly and leave any comments you have on them. I will address all of them at once.

vfdev-5:
@Moh-Yakoub thinking more on that, I think the implementation should work on the same test cases as accuracy, precision, recall: binary (real binary and not 2 classes multclass), multiclass (maybe multilabel?) ...

Moh-Yakoub:
> @Moh-Yakoub thinking more on that, I think the implementation should work on the same test cases as accuracy, precision, recall: binary (real binary and not 2 classes multclass), multiclass (maybe multilabel?) ...

Maybe I am not following here. So you are suggesting replacing the current implementation to use binary as real binary ant not 2 classes.
I guess we cleared this point at the beginning of the implementation due to the fact that the only way we can calculate this via precisions/recall per label was to use the multiclass version.

vfdev-5:
> > @Moh-Yakoub thinking more on that, I think the implementation should work on the same test cases as accuracy, precision, recall: binary (real binary and not 2 classes multclass), multiclass (maybe multilabel?) ...
> 
> Maybe I am not following here. So you are suggesting replacing the current implementation to use binary as real binary ant not 2 classes.
> I guess we cleared this point at the beginning of the implementation due to the fact that the only way we can calculate this via precisions/recall per label was to use the multiclass version.

My comments are mostly about why we have to do 
```python
y_true = torch.randint(0, n_classes, size=(offset * idist.get_world_size(),)).to(device)
y_true_unflat = _unflatten_multilabel(y_true, n_classes)
```
for multi-class and similar thing for "binary".

I agree that for binary case, we ask user to provide binarized values using `output_transform`. So, in test we can follow the same test case strategy as for precision or recall: https://github.com/pytorch/ignite/blob/e6b2c3217aea5f4c47f747d7d6666a51d7b8ac79/tests/ignite/metrics/test_recall.py#L77

For multiclass, metric input should be 
```python
y_pred = torch.rand(B, C)
y_true  = torch.randint(0, C, size=(B, ))
```

Moh-Yakoub:
> > > @Moh-Yakoub thinking more on that, I think the implementation should work on the same test cases as accuracy, precision, recall: binary (real binary and not 2 classes multclass), multiclass (maybe multilabel?) ...
> > 
> > 
> > Maybe I am not following here. So you are suggesting replacing the current implementation to use binary as real binary ant not 2 classes.
> > I guess we cleared this point at the beginning of the implementation due to the fact that the only way we can calculate this via precisions/recall per label was to use the multiclass version.
> 
> My comments are mostly about why we have to do
> 
> ```python
> y_true = torch.randint(0, n_classes, size=(offset * idist.get_world_size(),)).to(device)
> y_true_unflat = _unflatten_multilabel(y_true, n_classes)
> ```
> 
> for multi-class and similar thing for "binary".
> 
> I agree that for binary case, we ask user to provide binarized values using `output_transform`. So, in test we can follow the same test case strategy as for precision or recall:
> 
> https://github.com/pytorch/ignite/blob/e6b2c3217aea5f4c47f747d7d6666a51d7b8ac79/tests/ignite/metrics/test_recall.py#L77
> 
> For multiclass, metric input should be
> 
> ```python
> y_pred = torch.rand(B, C)
> y_true  = torch.randint(0, C, size=(B, ))
> ```

@vfdev-5 I think I got what you mean, I've updated the test to be similar to the accuracy/precision tests. Let me know your thoughts.

vfdev-5:
@Moh-Yakoub thanks ! Looks good. What do you think about if we can cover binary case ?

Moh-Yakoub:
> @Moh-Yakoub thanks ! Looks good. What do you think about if we can cover binary case ?



> @Moh-Yakoub thanks ! Looks good. What do you think about if we can cover binary case ?

I wanted initially to add one, I wasn't sure how to make it properly as the use of metrics lambda require attaching to engine and we don't have a similar example in Fbeta to follow.

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,106 @@
...
+        labels: Optional list of label indices to include in the report
+
+    .. code-block:: python
+
+
+        y_true = torch.tensor([[1, 0],
+        [0, 1], [1, 0], [1, 1], [1, 0], [0, 1], [0, 1], [1, 0], [0, 1], [1, 0]])
+        y_pred = torch.randint(0, 2, size=(10,))
+
+        classification_report = ClassificationReport(output_dict=True)
```
trsvchn, line 46: Shouldn't we add `is_multilabel=True` here?

trsvchn: I've tried this example I've got
> NotComputableError: Recall must have at least one example before it can be computed.

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,106 @@
...
+        classification_report.update((y_true, y_pred))
+        res = classification_report.compute()
```
trsvchn, line 48: I've tried this example and I've got
> NotComputableError: Recall must have at least one example before it can be computed.

Moh-Yakoub: @trsvchn  Thanks a lot for reviewing this.
The example was from the old implementation where I was overriding MetricsLambda and had a custom update method. This is not applicable anymore as MetricsLambda methods don't by default update underlying metrics.

The only way is to attach it to an engine as specified by the test cases. I will update the example.

Thanks for raising this!

ignite/metrics/classification_report.py
```diff
@@ -0,0 +1,120 @@
...
+        y_true = torch.randint(0, 2, size=(offset * idist.get_world_size(),))
+        y_preds = torch.rand(offset * idist.get_world_size(), 2)
+
+        def update(engine, i):
+            return (
+                y_preds[i * s + rank * offset: (i + 1) * s + rank * offset, :],
+                y_true[i * s + rank * offset: (i + 1) * s + rank * offset],
+            )
+
+        engine = Engine(update)
```
vfdev-5, line 55: @Moh-Yakoub I think this is too complex for an example of class usage. Let's propose something more simple.

trsvchn: What about smth like with SSIM

```python
    # ...
    return y_pred, y

metric = ClassificationReport()
metric.attach(engine, "cr")
engine.run...
res = engine.state.metrics["cr"]
# result should be like
{
  "0": {
    "precision": 0.4891304347826087,
    "recall": 0.5056179775280899,
    "f1-score": 0.497237569060773
  },
  "1": {
    "precision": 0.5157232704402516,
    "recall": 0.4992389649923896,
    "f1-score": 0.507347254447022
  },
  "macro avg": {
    "precision": 0.5024268526114302,
    "recall": 0.5024284712602398,
    "f1-score": 0.5022924117538975
  }
}
```

Moh-Yakoub: @trsvchn Great suggestion! Thanks a lot

Moh-Yakoub:
@vfdev-5 Let me know what you think of the current PR.

I know there are a lot of areas that would need improvement for implementation and testing and I'd be up to do that for this and Fbeta in later PRs.

Thanks a lot.

vfdev-5:
@Moh-Yakoub thanks for pinging ! I think we can merge this PR as it takes almost ages to iterate over and do improvements in other PRs. For example, adding this metric somewhere in an example would be very helpful. Printing output dictionary in a handy way would be a great addition as well...

vfdev-5 (approved):
Thanks a lot for the PR @Moh-Yakoub !
LGTM!

PR #1888, opened 2021-03-26
Fix mkl linking issue

### LINKED ISSUE #1884: Fix TPU CI, libmkl_intel_lp64.so: cannot open shared object file: No such file or directory
- https://github.com/pytorch/ignite/pull/1880/checks?check_run_id=2205301550

I think this is related to this bad configuration:
https://github.com/pytorch/ignite/blob/05c5e19c839b1ef0138d695613997af1da2aef71/.github/workflows/tpu-tests.yml#L69

while we are using https://github.com/pytorch/ignite/blob/05c5e19c839b1ef0138d695613997af1da2aef71/.github/workflows/tpu-tests.yml#L29

trsvchn:
Fixes #1884

Description:

Solution as proposed here https://github.com/pytorch/ignite/issues/1884#issuecomment-808577870

Check list:

### DIFF
--- a/.github/workflows/tpu-tests.yml
+++ b/.github/workflows/tpu-tests.yml
@@ -64,9 +64,17 @@ jobs:
           pip install -r requirements-dev.txt
           python setup.py install
 
+      # https://github.com/pytorch/ignite/issues/1884
+      - name: Fix mkl Linking Issue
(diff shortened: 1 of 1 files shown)


vfdev-5 (approved):
LGTM

PR #1892, opened 2021-03-27
Fix TPU failure

KickItLikeShika:
Add eps to `WaveHdegesDistance` and reduce tolerence to fix TPU failure

Check list:

### DIFF
--- a/tests/ignite/contrib/metrics/regression/test_wave_hedges_distance.py
+++ b/tests/ignite/contrib/metrics/regression/test_wave_hedges_distance.py
@@ -84,7 +84,7 @@ def test_integration():
             _test(y_pred, y, batch_size)
 
 
-def _test_distrib_compute(device, tol=1e-5):
+def _test_distrib_compute(device):
     rank = idist.get_rank()
 
     def _test(metric_device):
@@ -108,7 +108,7 @@ def _test_distrib_compute(device, tol=1e-5):
 
         np_sum = (np.abs(np_y - np_y_pred) / (np.maximum.reduce([np_y_pred, np_y]) + 1e-30)).sum()
 
-        assert np_sum == pytest.approx(res, rel=tol)
+        assert np_sum == pytest.approx(res)
 
     for _ in range(3):
         _test("cpu")
@@ -116,7 +116,7 @@ def _test_distrib_compute(device, tol=1e-5):
             _test(idist.device())
 
 
-def _test_distrib_integration(device, tol=1e-5):
+def _test_distrib_integration(device):
 
     rank = idist.get_rank()
     torch.manual_seed(12)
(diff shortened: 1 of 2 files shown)


tests/ignite/contrib/metrics/regression/test_wave_hedges_distance.py
```diff
@@ -16,7 +16,7 @@ def test_wrong_input_shapes():
         m.update((torch.rand(4), torch.rand(4, 1)))
 
     with pytest.raises(ValueError, match=r"Input data shapes should be the same, but given"):
-        m.update((torch.rand(4, 1), torch.rand(4,)))
+        m.update((torch.rand(4, 1), torch.rand(4,),))
```
KickItLikeShika, line 19: It seems like one of the code-styling libraries (isort, black, flake8) added this

vfdev-5: This seems strange...

tests/ignite/contrib/metrics/regression/test_wave_hedges_distance.py
```diff
@@ -84,7 +84,7 @@ def get_test_cases():
             _test(y_pred, y, batch_size)
 
 
-def _test_distrib_compute(device):
+def _test_distrib_compute(device, tol=1e-5):
```
vfdev-5, line 87: Do we really need to change tolerence for xla ?

vfdev-5 (commented):
Thanks for the PR @KickItLikeShika , I have few questions here

tests/ignite/contrib/metrics/regression/test_wave_hedges_distance.py
KickItLikeShika: I did that just in case of we needed to reduce the tolerence for xla as we almost do (we did that in `FractionalBias`), what do you think?

vfdev-5: I'd prefer to adjust that if there is already a precision failure

KickItLikeShika: Sure! i will keep the tolerence `1e-5` to see what's going to happen

tests/ignite/contrib/metrics/regression/test_wave_hedges_distance.py
KickItLikeShika: Yes, it's the second time this happens! check this #1869 at the second commit i deleted it, i thought i added it by mistake, but turns out i didn't!

tests/ignite/contrib/metrics/regression/test_wave_hedges_distance.py
sdesrozis: IMO if `tol` is not currently needed (because it works), we should remove it, even if it could help. It is necessary to avoid under-quality but also over-quality in software engineering 😉

KickItLikeShika: Thanks for this @sdesrozis, appreciate your advice!

vfdev-5: @KickItLikeShika can you please remove `tol` at all

vfdev-5 (approved):
Thanks for the PR @KickItLikeShika !

PR #1896, opened 2021-03-30
updated wandb logger

### LINKED ISSUE #1894: Missing `close` method in `WandBLogger` and `PolyaxonLogger` ?
## 🐛 Bug description

Just found it unexpected. Not sure this is intended. There are `close` method overrides in provided loggers except for `WandBLogger` and `PolyaxonLogger`. [wandb states that wandb process have to close if we are on jupyter or colab env](https://docs.wandb.ai/integrations/huggingface#6-finish-your-w-and-b-run-notebook-only). Not sure about PolyaxonLogger.

We need to have one API to close all loggers even if users want to change using different loggers.

ahmedo42:
Fixes #1894

Description:

wandb logger should have close method.

Check list:

### DIFF
--- a/ignite/contrib/handlers/polyaxon_logger.py
+++ b/ignite/contrib/handlers/polyaxon_logger.py
@@ -87,7 +87,7 @@ class PolyaxonLogger(BaseLogger):
                 optimizer=optimizer,
                 param_name='lr'  # optional
             )
-            # We need to close the logger when we are done
+            # to manually end a run
             plx_logger.close()
 
     Args:
--- a/ignite/contrib/handlers/polyaxon_logger.py
+++ b/ignite/contrib/handlers/polyaxon_logger.py
@@ -87,6 +87,8 @@ class PolyaxonLogger(BaseLogger):
                 optimizer=optimizer,
                 param_name='lr'  # optional
             )
+            # We need to close the logger when we are done
+            wandb_logger.close()
 
     Args:
         args: Positional arguments accepted from
@@ -101,6 +103,7 @@ class PolyaxonLogger(BaseLogger):
             from polyaxon.tracking import Run
 
             self.experiment = Run(*args, **kwargs)
+
         except ImportError:
             try:
                 from polyaxon_client.tracking import Experiment
@@ -114,7 +117,10 @@ class PolyaxonLogger(BaseLogger):
                 )
 
     def close(self) -> None:
-        pass
+        try:
+            self.experiment.end()
+        except:
+            pass
 
     def __getattr__(self, attr: Any) -> Any:
         return getattr(self.experiment, attr)
--- a/ignite/contrib/handlers/polyaxon_logger.py
+++ b/ignite/contrib/handlers/polyaxon_logger.py
@@ -88,7 +88,7 @@ class PolyaxonLogger(BaseLogger):
                 param_name='lr'  # optional
             )
             # We need to close the logger when we are done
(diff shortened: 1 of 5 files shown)


sdesrozis (approved):
@ahmedo42 Thank you ! LGTM

vfdev-5:
@ahmedo42 to close #1894 we have to add empty close to PolyaxonLogger as well and additionally it would be helpful to check Polyaxon API if they still do not require any closing method.

ahmedo42:
> it would be helpful to check Polyaxon API if they still do not require any closing method.

I checked the docs and didn't find that they require one.

sdesrozis:
> I checked the docs and didn't find that they require one.

@ahmedo42 Are you sure ? I think there is a `tracking.end()` to manually end a run.

ahmedo42:
>  I think there is a `tracking.end()` to manually end a run.

Yes you're correct there is , but as far as we're using this logger it should be closed by calling  [Run.end()](https://polyaxon.com/docs/experimentation/tracking/client/#end) except when I simply call this I get the following error:

` 'Run' object has no attribute '_exit_handler' ` which entails that  the logger isn't initialized correctly perhaps?

Also I couldn't find any useful documentation for closing the logger when we use Experiment instead of Run

So wouldn't using [tracking module](https://polyaxon.com/docs/experimentation/tracking/module/) instead of Run and Experiment be more cleaner?

sdesrozis:
Sorry I meant `Run.end()`... Btw it is necessary to dig further the issue when calling this method. Maybe trying to reproduce this issue without ignite would help to understand if it's on our side or not ?

ahmedo42:
@sdesrozis  I filed an issue on Polyaxon and it's indeed a bug https://github.com/polyaxon/polyaxon/issues/1258#issuecomment-811511510

sdesrozis:
@ahmedo42 Thank you! 

So we have to handle an exception inside the method `close()` as suggested by Polyaxon team. Or do a correct initialisation (another PR ?)

ahmedo42:
@sdesrozis  I think catching the exception would be OK for now until they fix it ( even if it's a bit lazy 😄 ), if you think differently it  shouldn't be so difficult to open another PR.

sdesrozis:
> @sdesrozis  I think catching the exception would be OK for now until they fix it ( even if it's a bit lazy 😄 )

Let's do it like that 😊

ignite/contrib/handlers/polyaxon_logger.py
```diff
@@ -87,6 +87,8 @@ class PolyaxonLogger(BaseLogger):
                 optimizer=optimizer,
                 param_name='lr'  # optional
             )
+            # We need to close the logger when we are done
+            wandb_logger.close()
```
vfdev-5, line 91:
```suggestion
            plx_logger.close()
```

ignite/contrib/handlers/polyaxon_logger.py
```diff
@@ -87,6 +87,8 @@ class PolyaxonLogger(BaseLogger):
                 optimizer=optimizer,
                 param_name='lr'  # optional
             )
+            # We need to close the logger when we are done
```
vfdev-5, line 90: I think it is optional and not required by plx

sdesrozis:
@ahmedo42 Thank you! LGTM

sdesrozis (approved):
@ahmedo42 Thank you! LGTM

PR #1899, opened 2021-03-31
Improve `MAE` tests

KickItLikeShika:
Improve `MeanAbsoluteError` tests

Check list:

### DIFF
--- a/tests/ignite/metrics/test_mean_absolute_error.py
+++ b/tests/ignite/metrics/test_mean_absolute_error.py
@@ -1,5 +1,6 @@
 import os
 
+import numpy as np
 import pytest
 import torch
 
@@ -8,7 +9,7 @@ from ignite.exceptions import NotComputableError
 from ignite.metrics import MeanAbsoluteError
 
 
-def test_zero_div():
+def test_no_update():
     mae = MeanAbsoluteError()
     with pytest.raises(
         NotComputableError, match=r"MeanAbsoluteError must have at least one example before it can be computed"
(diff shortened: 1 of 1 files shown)


tests/ignite/metrics/test_mean_absolute_error.py
```diff
@@ -17,20 +18,47 @@ def test_zero_div():
...
+            n_iters = y.shape[0] // batch_size + 1
+            for i in range(n_iters):
+                idx = i * batch_size
+                mae.update((y_pred[idx : idx + batch_size], y[idx : idx + batch_size]))
+
+        np_res = (np.abs(np_y_pred - np_y)).sum() / np_y.shape[0]
+        assert isinstance(mae.compute(), float)
+        assert mae.compute() == np_res
+
+    def get_test_cases():
```
sdesrozis, line 41: @KickItLikeShika @vfdev-5 

wouldn’t we use `pytest.parametrize` here ?

vfdev-5: @sdesrozis we discussed that in some of previous PRs and as we run `_test` multiple times, we prefer to generate different data points. Using parametrize it would be random data but the same for `for _ in range(10)` loop. Valid question is whether we would need that loop...

KickItLikeShika: Sorry for the delay @sdesrozis, and thanks for explanation @vfdev-5. Is everything fine now?

sdesrozis: > Valid question is whether we really need that loop...

We need the loop if random values can generate different behaviour. I think that is not useful here.

sdesrozis (approved):
@KickItLikeShika LGTM

PR #1907, opened 2021-04-05
Improve docs github workflow action

Moh-Yakoub:
Addresses #1836 

Description:

Trigger docs build when a change is added to the docs folder.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -3,6 +3,8 @@ name: Build docs
 on:
   push:
     branches: [master]
+    paths:
+      - "docs/**"
   pull_request:
   release:
     types: [published]
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -8,7 +8,6 @@ on:
       - "tests/**"
       - "docker/**"
       - "assets/**"
-      - ".github/workflows/**"
   release:
     types: [published]
 
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -3,11 +3,6 @@ name: Build docs
 on:
   push:
     branches: [master]
-    paths-ignore:
-      - "tests/**"
-      - "docker/**"
-      - "assets/**"
-      - ".github/workflows/**"
   pull_request:
     paths-ignore:
       - "tests/**"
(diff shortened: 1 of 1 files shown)


vfdev-5:
Thanks for the PR @Moh-Yakoub but I also thought about that previously and it is not what we'd like. Please, see this comment: https://github.com/pytorch/ignite/issues/1836#issuecomment-813617977

vfdev-5:
@Moh-Yakoub please update this PR according to https://github.com/pytorch/ignite/issues/1836#issuecomment-813941963

.github/workflows/docs.yml
```diff
@@ -3,7 +3,17 @@ name: Build docs
 on:
   push:
     branches: [master]
+    paths-ignore:
+      - "tests/**"
+      - "docker/**"
+      - "assets/**"
+      - ".github/workflows/**"
```
trsvchn, line 10: Let's focus on `pull_request`, since we want to test our docs on `push` to master in all cases

```suggestion
```

.github/workflows/docs.yml
```diff
@@ -3,7 +3,17 @@ name: Build docs
...
   pull_request:
+    paths-ignore:
+      - "tests/**"
+      - "docker/**"
+      - "assets/**"
```
trsvchn, line 15: ok, `tests` and `docker` are fine, but we use some items from `assets`, so let's remove `assets`

```suggestion
```

.github/workflows/docs.yml
```diff
@@ -3,7 +3,17 @@ name: Build docs
...
+      - ".github/workflows/**"
```
trsvchn, line 16: Yeah, I was thinking to remove some workflows? but it going to be too verbose and messy (we also have some scripts there etc), so let's remove it, since we want to run docs workflow when `worlflows/docs.tml` changes as well.

```suggestion
```

trsvchn (changes requested):
Thank you for your PR @Moh-Yakoub ! 
I have left a few comments.

In general, let's focus on `pull_request` events and for now let's ignore `tests` and `docker` dirs

trsvchn (approved):
Thanks @Moh-Yakoub! LGTM!

PR #1905, opened 2021-04-04
Refactor `Precision` tests

KickItLikeShika:
Update tests, use `pytest.mark.parametrize()`, and improve test generation.

Check list:

- [ ] New tests are added (if a new feature is added)
- [ ] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #1908, opened 2021-04-06
Improve `RMSE` tests

KickItLikeShika:
Improve `RootMeanSquaredError` tests.

Check list:

### DIFF
--- a/tests/ignite/metrics/test_root_mean_squared_error.py
+++ b/tests/ignite/metrics/test_root_mean_squared_error.py
@@ -43,11 +43,27 @@ def test_compute():
     def get_test_cases():
 
         test_cases = [
-            (torch.randint(0, 10, size=(50,)), torch.randint(0, 10, size=(50,)), 1),
-            (torch.randint(-10, 10, size=(50, 1)), torch.randint(-10, 10, size=(50, 1)), 1),
+            (
+                torch.randint(0, 10, size=(50,), dtype=torch.float),
+                torch.randint(0, 10, size=(50,), dtype=torch.float),
+                1,
+            ),
+            (
+                torch.randint(-10, 10, size=(50, 1), dtype=torch.float),
+                torch.randint(-10, 10, size=(50, 1), dtype=torch.float),
+                1,
+            ),
             # updated batches
-            (torch.randint(0, 10, size=(50,)), torch.randint(0, 10, size=(50,)), 16),
-            (torch.randint(-10, 10, size=(50, 1)), torch.randint(-10, 10, size=(50, 1)), 16),
+            (
+                torch.randint(0, 10, size=(50,), dtype=torch.float),
+                torch.randint(0, 10, size=(50,), dtype=torch.float),
+                16,
+            ),
+            (
+                torch.randint(-10, 10, size=(50, 1), dtype=torch.float),
+                torch.randint(-10, 10, size=(50, 1), dtype=torch.float),
+                16,
+            ),
         ]
 
         return test_cases
--- a/tests/ignite/metrics/test_root_mean_squared_error.py
+++ b/tests/ignite/metrics/test_root_mean_squared_error.py
@@ -43,20 +43,16 @@ def test_compute():
     def get_test_cases():
 
         test_cases = [
-            (torch.randint(0, 10, size=(100,)), torch.randint(0, 10, size=(100,)), 1),
-            (torch.randint(-10, 10, size=(100, 1)), torch.randint(-10, 10, size=(100, 1)), 1),
-            (torch.randint(0, 10, size=(200,)), torch.randint(0, 10, size=(200,)), 1),
-            (torch.randint(-10, 10, size=(200, 1)), torch.randint(-10, 10, size=(200, 1)), 1),
+            (torch.randint(0, 10, size=(50,)), torch.randint(0, 10, size=(50,)), 1),
+            (torch.randint(-10, 10, size=(50, 1)), torch.randint(-10, 10, size=(50, 1)), 1),
             # updated batches
-            (torch.randint(0, 10, size=(100,)), torch.randint(0, 10, size=(100,)), 16),
-            (torch.randint(-10, 10, size=(100, 1)), torch.randint(-10, 10, size=(100, 1)), 16),
-            (torch.randint(0, 10, size=(200,)), torch.randint(0, 10, size=(200,)), 16),
-            (torch.randint(-10, 10, size=(200, 1)), torch.randint(-10, 10, size=(200, 1)), 16),
+            (torch.randint(0, 10, size=(50,)), torch.randint(0, 10, size=(50,)), 16),
+            (torch.randint(-10, 10, size=(50, 1)), torch.randint(-10, 10, size=(50, 1)), 16),
         ]
 
         return test_cases
 
-    for _ in range(10):
+    for _ in range(5):
         # check multiple random inputs as random exact occurencies are rare
         test_cases = get_test_cases()
         for y_pred, y, batch_size in test_cases:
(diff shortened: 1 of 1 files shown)


tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -17,24 +18,52 @@ def test_zero_div():
...
+
+        assert isinstance(res, float)
+        assert pytest.approx(res) == np_res
+
+    def get_test_cases():
+
+        test_cases = [
+            (torch.randint(0, 10, size=(100,)), torch.randint(0, 10, size=(100,)), 1),
+            (torch.randint(-10, 10, size=(100, 1)), torch.randint(-10, 10, size=(100, 1)), 1),
+            (torch.randint(0, 10, size=(200,)), torch.randint(0, 10, size=(200,)), 1),
```
sdesrozis, line 48: Is this test (and the following one) really relevant regarding to the test `size=(100,)` ?

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -17,24 +18,52 @@ def test_zero_div():
...
+            (torch.randint(-10, 10, size=(200, 1)), torch.randint(-10, 10, size=(200, 1)), 1),
+            # updated batches
+            (torch.randint(0, 10, size=(100,)), torch.randint(0, 10, size=(100,)), 16),
+            (torch.randint(-10, 10, size=(100, 1)), torch.randint(-10, 10, size=(100, 1)), 16),
+            (torch.randint(0, 10, size=(200,)), torch.randint(0, 10, size=(200,)), 16),
```
sdesrozis, line 53: Same question

sdesrozis:
@KickItLikeShika Thank you ! I left few comments.

tests/ignite/metrics/test_root_mean_squared_error.py
KickItLikeShika: Can you please explain `relevant`? you mean `size=100` and `size=200` are the same?

sdesrozis: I mean `size=200` does add better test coverage regarding to `size=100`. Why do we test this size ?

KickItLikeShika: Basically we are trying multiple test sizes `(100, 200)` and multiple ranges `([0, 10] and [-10, 10])`, what do you think?

vfdev-5: @KickItLikeShika the question is why 100 and not 10 ? Why 4 test cases and not just 2 ?

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -17,24 +18,52 @@ def test_zero_div():
...
+
+        return test_cases
 
-    rmse.reset()
-    y_pred = torch.Tensor([[3.0], [-3.0]])
-    y = torch.zeros(2)
-    rmse.update((y_pred, y))
-    assert isinstance(rmse.compute(), float)
-    assert rmse.compute() == 3.0
+    for _ in range(10):
```
vfdev-5, line 59: Same here: please apply 5 instead of 10

vfdev-5 (changes requested):
Thanks for the PR @KickItLikeShika but please apply the same logic as we discussed here: 
https://github.com/pytorch/ignite/pull/1905

tests/ignite/metrics/test_root_mean_squared_error.py
KickItLikeShika: There is not particular reason for using 100 instead of 10. and we're trying multiple sizes `4 tests cases not two` for a better coverage.
> We already talked about that in one of your previous PR and the idea is too save computational resources.
Yes we did, but computational load is not a problem here, what do you think?

sdesrozis: > There is no particular reason for using 100 instead of 10. and we're trying multiple sizes `4 tests cases not two` for a better coverage.

The coverage is not better IMO. The methods don't depend on values neither shapes.

KickItLikeShika: @sdesrozis thanks for the explanation!

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -17,24 +18,48 @@ def test_zero_div():
...
+        assert pytest.approx(res) == np_res
+
+    def get_test_cases():
+
+        test_cases = [
+            (torch.randint(0, 10, size=(50,)), torch.randint(0, 10, size=(50,)), 1),
+            (torch.randint(-10, 10, size=(50, 1)), torch.randint(-10, 10, size=(50, 1)), 1),
+            # updated batches
+            (torch.randint(0, 10, size=(50,)), torch.randint(0, 10, size=(50,)), 16),
+            (torch.randint(-10, 10, size=(50, 1)), torch.randint(-10, 10, size=(50, 1)), 16),
```
vfdev-5, line 50: Actually, why input tensors are int/long and not float as in the original implementation ?
@KickItLikeShika please pay attention on those details while porting things !

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -17,24 +18,64 @@ def test_zero_div():
...
+        assert isinstance(res, float)
+        assert pytest.approx(res) == np_res
+
+    def get_test_cases():
+
+        test_cases = [
+            (
+                torch.randint(0, 10, size=(50,), dtype=torch.float),
+                torch.randint(0, 10, size=(50,), dtype=torch.float),
+                1,
```
vfdev-5, line 49: Please, use `torch.rand()` and rescale it to [0, 10] interval. Use size=10 for single batch case.
Remove `,` after `1`:
```suggestion
                torch.randint(0, 10, size=(50,), dtype=torch.float),
                torch.randint(0, 10, size=(50,), dtype=torch.float),
                1
```

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -17,24 +18,64 @@ def test_zero_div():
...
+            ),
+            (
+                torch.randint(-10, 10, size=(50, 1), dtype=torch.float),
+                torch.randint(-10, 10, size=(50, 1), dtype=torch.float),
+                1,
```
vfdev-5, line 54: Same here for `torch.rand()` and rescale it to [-10, 10] interval and size=(10, 1)
```suggestion
                torch.randint(-10, 10, size=(50, 1), dtype=torch.float),
                torch.randint(-10, 10, size=(50, 1), dtype=torch.float),
                1
```

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -17,24 +18,64 @@ def test_zero_div():
...
+            ),
+            # updated batches
+            (
+                torch.randint(0, 10, size=(50,), dtype=torch.float),
+                torch.randint(0, 10, size=(50,), dtype=torch.float),
+                16,
```
vfdev-5, line 60: Same here for `torch.rand()` and rescale it to [0, 10] interval,
```suggestion
                torch.randint(0, 10, size=(50,), dtype=torch.float),
                torch.randint(0, 10, size=(50,), dtype=torch.float),
                16
```

vfdev-5 (changes requested):

tests/ignite/metrics/test_root_mean_squared_error.py
KickItLikeShika: Regarding the `,` after `1`, `black` adds the comma automatically when i commit the changes

vfdev-5 (approved):
Now looks much better. Thanks @KickItLikeShika !

PR #1915, opened 2021-04-08
Refactor confusion matrix multiclass tests

Moh-Yakoub:
Description: 

Refactor confusion matrix multiclass/binary tests and remove all repeated test case setups

Check list:


tests/ignite/metrics/test_confusion_matrix.py
```diff
@@ -55,145 +55,97 @@ def test_multiclass_wrong_inputs():
...
+            return [
+                (torch.rand(10, 4), torch.randint(0, 4, size=(10,)).long(), 4, 1),
+                (torch.rand(4, 10), torch.randint(0, 10, size=(4,)).long(), 10, 1),
+                (torch.rand(4, 2), torch.randint(0, 2, size=(4,)).long(), 2, 1),
+                # batched updates
+                (torch.rand(100, 5), torch.randint(0, 5, size=(100,)).long(), 5, 16),
+            ]
+
+        # check multiple random inputs as random exact occurencies are rare
+        for _ in range(10):
```
KickItLikeShika, line 86: For reducing computational load we can this make `5` instead of `10`

tests/ignite/metrics/test_confusion_matrix.py
```diff
@@ -55,145 +55,97 @@ def test_multiclass_wrong_inputs():
...
+    def get_test_cases():
+        return [
+            (torch.rand(10, 4, 5), torch.randint(0, 4, size=(10, 5)).long(), 4, 1),
+            (torch.rand(4, 10, 5), torch.randint(0, 10, size=(4, 5)).long(), 10, 1),
+            # batched updates
+            (torch.rand(100, 9, 7), torch.randint(0, 9, size=(100, 7)).long(), 9, 16),
+        ]
 
     # check multiple random inputs as random exact occurencies are rare
     for _ in range(10):
```
KickItLikeShika, line 117: This one too

tests/ignite/metrics/test_confusion_matrix.py
```diff
@@ -55,145 +55,97 @@ def test_multiclass_wrong_inputs():
...
+    def get_test_cases():
+        return [
+            (torch.rand(4, 5, 12, 10), torch.randint(0, 5, size=(4, 12, 10)).long(), 5, 1),
+            (torch.rand(4, 5, 10, 12, 8), torch.randint(0, 5, size=(4, 10, 12, 8)).long(), 5, 1),
+            # batched updates
+            (torch.rand(100, 3, 8, 8), torch.randint(0, 3, size=(100, 8, 8)).long(), 3, 16),
+        ]
 
     # check multiple random inputs as random exact occurencies are rare
     for _ in range(10):
```
KickItLikeShika, line 146: And this one

KickItLikeShika (commented):
Thanks for the PR! i have left some comments

tests/ignite/metrics/test_confusion_matrix.py
```diff
@@ -55,145 +55,97 @@ def test_multiclass_wrong_inputs():
...
 
-        num_classes = 10
         cm = ConfusionMatrix(num_classes=num_classes)
-        y_pred = torch.rand(4, num_classes)
-        y = torch.randint(0, num_classes, size=(4,)).long()
-        cm.update((y_pred, y))
-        np_y_pred = y_pred.numpy().argmax(axis=1).ravel()
-        np_y = y.numpy().ravel()
-        assert np.all(confusion_matrix(np_y, np_y_pred, labels=list(range(num_classes))) == cm.compute().numpy())
+        if batch_size == 1:
```
KickItLikeShika, line 61: Why do we need this check? we can update then check the `batch size`, right?

tests/ignite/metrics/test_confusion_matrix.py
```diff
@@ -55,145 +55,97 @@ def test_multiclass_wrong_inputs():
...
-        np_y = y.numpy().ravel()
-        assert np.all(confusion_matrix(np_y, np_y_pred, labels=list(range(num_classes))) == cm.compute().numpy())
+    def _test_NL(y_pred, y, num_classes, batch_size):
 
-        num_classes = 10
         cm = ConfusionMatrix(num_classes=num_classes)
-        y_pred = torch.rand(4, num_classes, 5)
-        y = torch.randint(0, num_classes, size=(4, 5)).long()
-        cm.update((y_pred, y))
+        if batch_size == 1:
```
KickItLikeShika, line 96: Same for this check

tests/ignite/metrics/test_confusion_matrix.py
```diff
@@ -55,145 +55,97 @@ def test_multiclass_wrong_inputs():
...
     # Multiclass input data of shape (N, H, W, ...)
-    def _test_NHW():
-        num_classes = 5
+    def _test_NHW(y_pred, y, num_classes, batch_size):
         cm = ConfusionMatrix(num_classes=num_classes)
-
-        y_pred = torch.rand(4, num_classes, 12, 10)
-        y = torch.randint(0, num_classes, size=(4, 12, 10)).long()
-        cm.update((y_pred, y))
+        if batch_size == 1:
```
KickItLikeShika, line 126: Here too

tests/ignite/metrics/test_confusion_matrix.py
```diff
@@ -55,145 +55,97 @@ def test_multiclass_wrong_inputs():
 
 def test_multiclass_input_N():
```
KickItLikeShika, line 56: Can we refactor all three `multiclass_input_` functions? we can just implement one function `test_multiclass_input` that has all three types of test cases.

tests/ignite/metrics/test_confusion_matrix.py
Moh-Yakoub: From execution point of view yes, it's doable. I am just keeping it to make the test case more readable

KickItLikeShika: It can work both ways, @vfdev-5 can we hear from you here?

tests/ignite/metrics/test_confusion_matrix.py
vfdev-5: @Moh-Yakoub let's code this test as here : https://github.com/pytorch/ignite/blob/a82327e1d908f122c359b9717a8003a45cf850e4/tests/ignite/metrics/test_accuracy.py#L64

```python

    def _test(y_pred, y, cm, batch_size):
         cm.reset()
         cm.update((y_pred, y))
         ...
         if batch_size > 1:
             ...

    for _ in range(5):
        for y_pred, y, num_classes, batch_size in get_test_cases():
            cm = ConfusionMatrix(num_classes=num_classes)
            _test(y_pred, y, cm, batch_size)

```

KickItLikeShika: @Moh-Yakoub please update it

Moh-Yakoub: @vfdev-5 I am not sure your code does the same check. The original test case looks like this
``
        cm = ConfusionMatrix(num_classes=num_classes)

        y_pred = torch.rand(100, num_classes)
        y = torch.randint(0, num_classes, size=(100,)).long()

        batch_size = 16

        for i in range(n_iters):
            idx = i * batch_size
            cm.update((y_pred[idx : idx + batch_size], y[idx : idx + batch_size]))``

If I try to do as you suggested, 
``
      cm.update((y_pred, y))
        if batch_size > 1:
            for i in range(n_iters):
                idx = i * batch_size
                cm.update((y_pred[idx : idx + batch_size], y[idx : idx + batch_size]))``


That means I am making a different sequence of update to the `cm` metric. My tests are failing because of that as well. not sure if i am missing anything, but how is this supposed to produce the same results if we have multiple update sequences.

vfdev-5: @Moh-Yakoub yes, you are right, the first update should be under the condition, it does not make much sense to put the same sequence twice when `batch_size > 1`. Thanks a lot for pointing out that !
How about that:
```python
    def _test(y_pred, y, cm, batch_size):
         cm.reset()
         ...
         if batch_size > 1:
             ... batched updates
         else:   
             cm.update((y_pred, y))   # single update
         ...

    for _ in range(5):
        for y_pred, y, num_classes, batch_size in get_test_cases():
            cm = ConfusionMatrix(num_classes=num_classes)
            _test(y_pred, y, cm, batch_size)
```
What do you think ?

@KickItLikeShika we have to take that into account for your PR on Accuracy, Recall, Precision etc... and fix that once we are agreed on the way to do that.

KickItLikeShika: @vfdev-5 that's a good idea! Thank you @Moh-Yakoub for such great ideas!

vfdev-5 (approved):
Thanks a lot @Moh-Yakoub !

PR #1919, opened 2021-04-09
Remove TODO from engine.py

radekosmulski:
I tested this - no change in GPU memory consumption regardless of whether the deleted line is uncommented or not.

When the batch is getting assigned to it lives in computer RAM, it only gets moved to `device` later. Meaning, the only memory that would be consumed here and not freed up would be computer RAM. From a design perspective, I am not sure that having `output.batch` set to None and having this additional line here would be worth it to collect whatever memory a single batch occupies.

But if that is something that would be believed to be worth it, I can run some tests to see what difference it makes wrt computer RAM consumption and report here.

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -807,6 +807,7 @@ class Engine(Serializable):
                 )
 
             while True:
+                self.state.batch = self.state.output = None
                 try:
                     # Avoid Events.GET_BATCH_STARTED triggered twice when data iter is restarted
                     if self.last_event_name != Events.DATALOADER_STOP_ITERATION:
@@ -851,9 +852,6 @@ class Engine(Serializable):
                 self._fire_event(Events.ITERATION_STARTED)
                 self.state.output = self._process_function(self, self.state.batch)
                 self._fire_event(Events.ITERATION_COMPLETED)
-
-                # TODO: remove refs on batch to avoid high mem consumption ? -> need verification
-                # self.state.batch = None
 
                 if self.should_terminate or self.should_terminate_single_epoch:
                     self._fire_event(Events.TERMINATE_SINGLE_EPOCH, iter_counter=iter_counter)
--- a/tests/ignite/engine/test_engine.py
+++ b/tests/ignite/engine/test_engine.py
@@ -969,14 +969,26 @@ def test_batch_is_released_before_new_one_is_loaded_cuda():
 @pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
 def test_output_is_released_before_new_one_is_assigned_cuda():
     torch.cuda.empty_cache()
-    mem_consumption = []
 
-    def update_fn(engine, batch):
-        output = torch.rand(10).cuda()
-        mem_consumption.append(torch.cuda.memory_allocated())
-        return output
+    def _test():
+        mem_consumption = []
 
-    engine = Engine(update_fn)
-    engine.run([0, 1], max_epochs=2)
+        def update_fn(engine, batch):
+            mem_consumption.append(torch.cuda.memory_allocated())
+            output = torch.rand(10).cuda()
+            mem_consumption.append(torch.cuda.memory_allocated())
+            return output
 
-    assert mem_consumption[0] == mem_consumption[1]
+        engine = Engine(update_fn)
+        engine.run([0, 1], max_epochs=2)
+
+        return mem_consumption
+
+    mem_consumption1 = _test()
+    # mem_consumption ~ [0, 512, 0, 512, 0, 512, 0, 512]
+    assert len(set(mem_consumption1)) == 2
+
+    mem_consumption2 = _test()
+    assert len(set(mem_consumption2)) == 2
+
+    assert mem_consumption1 == mem_consumption2
--- a/tests/ignite/engine/test_engine.py
+++ b/tests/ignite/engine/test_engine.py
@@ -940,18 +940,30 @@ def test_is_done_with_max_iters():
 @pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
 def test_batch_is_released_before_new_one_is_loaded_cuda():
     torch.cuda.empty_cache()
-    mem_consumption = []
-
-    def dataloader():
-        for _ in range(2):
-            batch = torch.randn(10).cuda()
-            mem_consumption.append(torch.cuda.memory_allocated())
-            yield batch
 
     engine = Engine(lambda e, b: None)
-    engine.run(dataloader(), max_epochs=2)
 
-    assert mem_consumption[0] == mem_consumption[1]
+    def _test():
+        mem_consumption = []
+
+        def dataloader():
+            for _ in range(4):
+                mem_consumption.append(torch.cuda.memory_allocated())
+                batch = torch.randn(10).cuda()
+                mem_consumption.append(torch.cuda.memory_allocated())
+                yield batch
+
+        engine.run(dataloader(), max_epochs=2, epoch_length=2)
+        return mem_consumption
+
+    mem_consumption1 = _test()
+    # mem_consumption should look like [0, 512, 512, 512, 512, 512, 512, 512]
+    assert len(set(mem_consumption1[1:])) == 1
+
+    mem_consumption2 = _test()
+    assert len(set(mem_consumption2[1:])) == 1
+
+    assert mem_consumption1 == mem_consumption2
 
 
 @pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
--- a/tests/ignite/engine/test_engine.py
+++ b/tests/ignite/engine/test_engine.py
@@ -938,7 +938,7 @@ def test_is_done_with_max_iters():
 
 
 @pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
-def test_batch_is_released_before_new_one_is_loaded_cuda():
+def test_batch_is_released_before_new_one_is_loaded_on_cuda():
     torch.cuda.empty_cache()
 
     engine = Engine(lambda e, b: None)
--- a/tests/ignite/engine/test_engine.py
+++ b/tests/ignite/engine/test_engine.py
@@ -935,3 +935,36 @@ def test_is_done_with_max_iters():
 
     state = State(iteration=250, epoch=1, max_epochs=3, epoch_length=100, max_iters=250)
     assert Engine._is_done(state)
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
+def test_batch_is_released_before_new_one_is_loaded_cuda():
+    torch.cuda.empty_cache()
+    mem_consumption = []
+
+    def dataloader():
+        for _ in range(2):
+            batch = torch.randn(10).cuda()
+            mem_consumption.append(torch.cuda.memory_allocated())
+            yield batch
+
+    engine = Engine(lambda e, b: None)
+    engine.run(dataloader(), max_epochs=2)
+
+    assert mem_consumption[0] == mem_consumption[1]
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
+def test_output_is_released_before_new_one_is_assigned_cuda():
+    torch.cuda.empty_cache()
+    mem_consumption = []
+
+    def update_fn(engine, batch):
+        output = torch.rand(10).cuda()
+        mem_consumption.append(torch.cuda.memory_allocated())
+        return output
+
+    engine = Engine(update_fn)
+    engine.run([0, 1], max_epochs=2)
+
+    assert mem_consumption[0] == mem_consumption[1]
--- a/tests/ignite/engine/test_engine.py
+++ b/tests/ignite/engine/test_engine.py
@@ -967,7 +967,7 @@ def test_batch_is_released_before_new_one_is_loaded_on_cuda():
 
 
 @pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
-def test_output_is_released_before_new_one_is_assigned_cuda():
+def test_output_is_released_before_new_one_is_assigned_on_cuda():
     torch.cuda.empty_cache()
 
     def _test():


radekosmulski:
Ok, well, I was curious what difference it makes with regards to computer RAM :slightly_smiling_face: 

With a batch of 142 images at 128x128, trying to hold everything else constant, running this multiple times, got consistent results: with setting `state.batch` to `None` ~30 mb are freed after training.

Not sure this is worth the bother :slightly_smiling_face:

vfdev-5:
@radekosmulski thanks for the PR ! The context for this commented code is related to https://github.com/pytorch/ignite/pull/476

Maybe, it worth rerun the tests from the discussions and see if something should be done or we can remove this commented line.

radekosmulski:
@vfdev-5 thank you for pointing me to #476 :slightly_smiling_face:

I think those are two separate issues. In `_prepare_batch` the tensors get sent to the `device`, possibly the GPU, but they never get assigned back.

```Python
def convert_tensor(
    x: Union[torch.Tensor, collections.Sequence, collections.Mapping, str, bytes],
    non_blocking: bool = False,
) -> Union[torch.Tensor, collections.Sequence, collections.Mapping, str, bytes]:
    """Move tensors to relevant device.

    Args:
        x: input tensor or mapping, or sequence of tensors.
        device: device type to move ``x``.
        non_blocking: convert a CPU Tensor with pinned memory to a CUDA Tensor
            asynchronously with respect to the host if possible
    """

    def _func(tensor: torch.Tensor) -> torch.Tensor:
        return tensor.to(device=device, non_blocking=non_blocking) if device is not None else tensor

    return apply_to_tensor(x, _func)


def apply_to_tensor(
    x: Union[torch.Tensor, collections.Sequence, collections.Mapping, str, bytes], func: Callable
) -> Union[torch.Tensor, collections.Sequence, collections.Mapping, str, bytes]:
    """Apply a function on a tensor or mapping, or sequence of tensors.

    Args:
        x: input tensor or mapping, or sequence of tensors.
        func: the function to apply on ``x``.
    """
    return apply_to_type(x, torch.Tensor, func)


def apply_to_type(
    x: Union[Any, collections.Sequence, collections.Mapping, str, bytes],
    input_type: Union[Type, Tuple[Type[Any], Any]],
    func: Callable,
) -> Union[Any, collections.Sequence, collections.Mapping, str, bytes]:
    """Apply a function on an object of `input_type` or mapping, or sequence of objects of `input_type`.

    Args:
        x: object or mapping or sequence.
        input_type: data type of ``x``.
        func: the function to apply on ``x``.
    """
    if isinstance(x, input_type):
        return func(x)
    if isinstance(x, (str, bytes)):
        return x
    if isinstance(x, collections.Mapping):
        return cast(Callable, type(x))({k: apply_to_type(sample, input_type, func) for k, sample in x.items()})
    if isinstance(x, tuple) and hasattr(x, "_fields"):  # namedtuple
        return cast(Callable, type(x))(*(apply_to_type(sample, input_type, func) for sample in x))
    if isinstance(x, collections.Sequence):
        return cast(Callable, type(x))([apply_to_type(sample, input_type, func) for sample in x])
    raise TypeError((f"x must contain {input_type}, dicts or lists; found {type(x)}"))
```

The tensors inside `self.state.batch` of `Engine` live on the CPU and never get moved to the `device`, only computer RAM consumption is affected, so I am thinking we might be good here :slightly_smiling_face: 

I am not sure what we can do about #476 though. On one hand, this behavior might not be completely apparent to the user, that this is the result they'll get when they return from `process_function` something that is large and that lives on the GPU.

On the other hand, one should probably assume that assigning tensors that have not been moved onto the CPU might lead to trouble.

I'll think more on #476 but don't see a good solution for now :slightly_frowning_face:

vfdev-5:
@radekosmulski thanks for your analysis. For me they are a bit related as we try to control the lifespan of data input and output. I was thinking that input batch can/should live between events: GET_BATCH_COMPLETED -> ITERATION_COMPLETED. So, for example, a trainer terminated its run with `trainer.state.batch is not None` and `trainer.state.output is not None` can be a bit strange. We can understand that they are the remainings from the last iteration, however IMO it would be better to remove them. In the same I was thinking that `trainer.state.output` should live until ITERATION_COMPLETED and be set to None after that, same as for batch in the comment.

However, it is a bit complicated to measure the impact of such change as well...

radekosmulski:
@vfdev-5 I am very new to the codebase and I am not sure I understand all the ramifications of the change you propose, but my initial reaction is that I like it :slightly_smiling_face: 

I typed out a long response but I keep changing it as I keep thinking about all the plausible scenarios - I can definitely see how people might be relying on the current behavior. At the same time, I like the clarity your proposed solution would introduce. It seems very elegant that the reference to the batch and output should only be retained for as long as it is needed for its original purpose, that is to support the `process_function` and the events that can be reasonably fired on ITERATION_COMPLETED.

Really appreciate the discussion!

sdesrozis:
My bad !!! I deleted a comment 😥

sdesrozis:
Btw I would share my thoughts about this issue.

I think it misses policies to handle lifespan of data in scope. @vfdev-5 said (in the message I deleted...) that it was strange that some data (batch and output) still in state after a run but it could be the same for every data.

radekosmulski:
Great to talk about this! :slightly_smiling_face: 

Please take this with a grain of salt - I don't have a lot of experience with ignite, but I am not completely sure the policies are needed, whether the additional complexity and lines of code that this would introduce are worth it. If I understand this correctly, with the changes proposed by @vfdev-5, things would go like this:

* `state.output` and `state.batch` are cleared after ITERATION_COMPLETED fires
* metrics are retained - if someone references a tensor on the GPU in a metric that they implement, they will run into trouble regardless of whether they use `ignite` or pure `pytorch`, this behavior shouldn't catch anyone off guard
* in this scenario, if one needs to retain something from `output` or `batch`, they should implement a handler to do so and store whatever they'd like on `state`, but I am not sure there will be many use cases for this

(the above captures all possible scenarios AFAICT and is quite easy to communicate and conceptualize)

I am having a hard time imagining one would need to retain output or batch if they really embrace the events / handler paradigm, but I might be wrong on that one :slightly_smiling_face:

Anyhow - the policies also sound interesting (maybe they are a more elegant way of handling this, and having this explicitly described in code might not be a bad thing), happy to help out with this however I can, regardless of which path is chosen.

vfdev-5:
Thanks for your opinions @radekosmulski @sdesrozis ! Great to talk about these internals :)

> I am having a hard time imagining one would need to retain output or batch if they really embrace the events / handler paradigm, but I might be wrong on that one

I agree, but it can still happend. If we opt to clear output / batch each iteration we should clearly say about that in the docs. Personally, i used that side effect to perform some visualizations. For example like that:
```python
        tb_logger.attach(
            evaluator,
            log_handler=predictions_gt_images_handler(...),
            event_name=Events.COMPLETED,
        )
```
but the then changed to another way to do that: https://github.com/pytorch/ignite/blob/13a42b90589112e67bcc6e1780e7c826a5a34199/examples/references/segmentation/pascal_voc2012/main.py#L138-L144

radekosmulski:
@vfdev-5 - yes, no problem :slightly_smiling_face: I thought this reproducer [here](https://github.com/pytorch/ignite/pull/476#issuecomment-480175292) was very nice. Happy to investigate this further, but could you please tell me what are we looking for beyond what was already demonstrated there?

I pushed some new code - I am not sure if this is the behavior that we want? I realize that this is all very basic. Should we arrive at an understanding of what we want, could you please help me spec out what should be implemented? I might need a bit of direction on writing idiomatic Python and OOP design but really happy to receive feedback, understand the rationale behind the proper way to write this and submit as many revisions as needed :slightly_smiling_face: 

Of course, I understand we are not even at the point here where any code needs to be written as not sure we have established what the solution should be, but just thought I'd throw it out there :slightly_smiling_face:

vfdev-5:
@radekosmulski thanks for digging that ! 

I think we would like to address here two points:
1) improve `state.batch` and `state.output` lifespan = these storages should live for one iteration and should not leak beyond. -> we can simply put in the code `self.state.batch = self.state.output = None` after all handlers attached to ITERATION_COMPLETED.
2) lifespan control can probably bring a memory reduction for batch and output when using the engine. Yes, this code snippet can be extended and used to measure GPU memory, to 
We can also argue that user can provide batch on GPU as well, so GPU memory measurements could show that we do not over-consume GPU memory while properly controlling the lifespan of `state.batch` and `state.output`.

Regarding the code you proposed, at first let's do that without exposing public attributes as it will be a public API that could be difficult to change later. We can simply do what we would like to achieve without generalizing:
`self.state.batch = self.state.output = None`

After, we can also write a unit test to execute it on GPU to ensure the behaviour. In this case we have to remove `[skip ci]` and I'll have to import the PR into another branch to launch GPU tests...

radekosmulski:
Thank you @vfdev-5 for your comment and for explaining all of this to me, appreciate it! :slightly_smiling_face: Very happy to work on this - might take me a day or two before I can get this done.

radekosmulski:
> 2. lifespan control can probably bring a memory reduction for batch and output when using the engine. Yes, this code snippet can be extended and used to measure GPU memory.
>    **We can also argue that user can provide batch on GPU as well**, so GPU memory measurements could show that we do not over-consume GPU memory while properly controlling the lifespan of `state.batch` and `state.output`.

Well, not the greatest of wins, but similarly to how we demonstrated earlier that there exists over-consumption of GPU memory with `engine.state.batch` here we can show that a similar effect holds when a user is providing a batch on the GPU.

image

And with setting `engine.state.batch` to `None` at end of the epoch this issue is gone:

image

Will add the line for setting `state.batch` and `state.output` to `None` and write the test!

radekosmulski:
BTW with manually written training loops in `PyTorch` I don't think I have seen this addressed, variables hang around across iterations, which is really cool - using ignite one gets additional GPU memory for free :wink:

radekosmulski:
@vfdev-5 I added tests and implemented the functionality.

There are some failing tests but that is because they rely on the old behavior. Can fix them without issue but just would like to ask if the proposed way of implementing this is okay? I was thinking that it might be nice to give the user the ability to bypass this clean-up for troubleshooting or while working on an implementation (and this would also make it easy to fix tests).

On second thought, maybe this would be a nicer way of implementing this:
```
# in __init__
self._debug = False

# in _run_once_on_dataset
if not self._debug: self.state.batch = self.state.output = None
```

Would appreciate any feedback and suggestions :slightly_smiling_face: Thank you!

vfdev-5:
@radekosmulski thanks for pointing out this. I agree that giving users the possibility to change that would be nice.

I was also thinking if we could keep the last batch and output of the last iteration to keep previous behaviour and maybe do not break user's code which does the same as we are doing in our tests.

As for `self._debug`, a boolean like this one may handle that, but with another name as "debug" is not clear what it does.
Maybe, something like `Engine.clear_batch_output` ?

radekosmulski:
Thank you @vfdev-5! `Engine.clear_batch_output` is a much better name, will make the change.

With regards to keeping the last batch, I am not sure that this would be a good idea. If the last batch is kept, we would hit the same issue as we have now should the `Engine` be run twice. Also, for as long as the engine would hang around (maybe a user would like to run an evaluator or there are multiple train stages) the memory of the GPU will continue to be used up by the old batch and output.

vfdev-5:
> With regards to keeping the last batch, I am not sure that this would be a good idea. If the last batch is kept, we would hit the same issue as we have now should the Engine be run twice. Also, for as long as the engine would hang around (maybe a user would like to run an evaluator or there are multiple train stages) the memory of the GPU will continue to be used up by the old batch and output.

I agree. We can in addition clear `batch` and `output` on `Engine.run` method in the very begining. I am afraid of breaking user's code which can use batch/output outside the `Engine.run`. For example, let's imagine a situation where I'm looking for the batch or output that fill some condition and I call `Engine.terminate()` and thus I'd like to use those batch/output in the code after running the engine.

radekosmulski:
> > With regards to keeping the last batch, I am not sure that this would be a good idea. If the last batch is kept, we would hit the same issue as we have now should the Engine be run twice. Also, for as long as the engine would hang around (maybe a user would like to run an evaluator or there are multiple train stages) the memory of the GPU will continue to be used up by the old batch and output.
> 
> I agree. We can in addition clear `batch` and `output` on `Engine.run` method in the very begining. I am afraid of breaking user's code which can use batch/output outside the `Engine.run`. For example, let's imagine a situation where I'm looking for the batch or output that fill some condition and I call `Engine.terminate()` and thus I'd like to use those batch/output in the code after running the engine.

In this case, maybe the solution I propose now could work even nicer. Let me know please what you think.

ignite/engine/engine.py
```diff
@@ -807,6 +807,7 @@ def _run_once_on_dataset(self) -> float:
                 )
 
             while True:
+                self.state.batch = self.state.output = None
```
vfdev-5, line 810: Awesome! I was thinking about that and wanted to comment out about that and you just implemented it 👍 
Thanks !

tests/ignite/engine/test_engine.py
```diff
@@ -935,3 +935,36 @@ def test_is_done_with_max_iters():
...
+
+    def update_fn(engine, batch):
+        output = torch.rand(10).cuda()
+        mem_consumption.append(torch.cuda.memory_allocated())
+        return output
+
+    engine = Engine(update_fn)
+    engine.run([0, 1], max_epochs=2)
+
+    assert mem_consumption[0] == mem_consumption[1]
```
vfdev-5, line 970:
```suggestion
    torch.cuda.empty_cache()

    def _test():
        mem_consumption = []

        def update_fn(engine, batch):
            mem_consumption.append(torch.cuda.memory_allocated())
            output = torch.rand(10).cuda()
            mem_consumption.append(torch.cuda.memory_allocated())
            return output

        engine = Engine(update_fn)
        engine.run([0, 1], max_epochs=2)

        return mem_consumption

    mem_consumption1 = _test()
    # mem_consumption ~ [0, 512, 0, 512, 0, 512, 0, 512]
    assert len(set(mem_consumption1)) == 2

    mem_consumption2 = _test()
    assert len(set(mem_consumption2)) == 2

    assert mem_consumption1 == mem_consumption2
```

tests/ignite/engine/test_engine.py
```diff
@@ -935,3 +935,36 @@ def test_is_done_with_max_iters():
 
     state = State(iteration=250, epoch=1, max_epochs=3, epoch_length=100, max_iters=250)
     assert Engine._is_done(state)
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
+def test_batch_is_released_before_new_one_is_loaded_cuda():
```
vfdev-5, line 941: We have a sort of rule to run non distributed tests on GPU, test should have `on_cuda` in the name (probably something to change as well)
```suggestion
def test_batch_is_released_before_new_one_is_loaded_on_cuda():
```

tests/ignite/engine/test_engine.py
```diff
@@ -935,3 +935,36 @@ def test_is_done_with_max_iters():
...
+            yield batch
+
+    engine = Engine(lambda e, b: None)
+    engine.run(dataloader(), max_epochs=2)
+
+    assert mem_consumption[0] == mem_consumption[1]
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
+def test_output_is_released_before_new_one_is_assigned_cuda():
```
vfdev-5, line 958:
```suggestion
def test_output_is_released_before_new_one_is_assigned_on_cuda():
```

vfdev-5 (commented):
Thanks for the update @radekosmulski !
I proposed a change for tests and it is good to go.

tests/ignite/engine/test_engine.py
```diff
@@ -935,3 +935,36 @@ def test_is_done_with_max_iters():
...
+    def dataloader():
+        for _ in range(2):
+            batch = torch.randn(10).cuda()
+            mem_consumption.append(torch.cuda.memory_allocated())
+            yield batch
+
+    engine = Engine(lambda e, b: None)
+    engine.run(dataloader(), max_epochs=2)
+
+    assert mem_consumption[0] == mem_consumption[1]
```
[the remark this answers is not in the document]
radekosmulski: looks great! much clearer to read

sdesrozis:
Thank you guys for this very interesting discussion.

About implementation, why not use specific handlers for adding reset of batch and output rather than hard code it inside the engine ?

vfdev-5:
> Thank you guys for this very interesting discussion.
> 
> About implementation, why not use specific handlers for adding reset of batch and output rather than hard code it inside the engine ?

This question I was asking myself for previous hardcoded features and the explanation we found is about keeping engine empty of internal handlers as in case of 1) introspection by user , it is better to not expose internals 2) if user removes all attached handlers it will change the engine assumed behavior...

sdesrozis:
> 1) introspection by user , it is better to not expose internals

If it's well done, I don't see issue concerning introspection. If we talk about software engineering, it is better to provide reusable components IMO. Handlers seems a good option and it is a coherent, extensible way to do things.

> 2) if user removes all attached handlers it will change the engine assumed behavior...

It depends on what we want to provide for users. Here, it could be a feature : do exactly what you want. It makes a lot of sense for me to have a very generic engine concept, potentially very agnostic to the usage.

vfdev-5:
@sdesrozis yes, I think you are right about these points, we can think about that and implement it for other release with refactored engine. Please, open an issue about that. Thanks !

vfdev-5 (approved):
LGTM, thanks @radekosmulski !

PR #1920, opened 2021-04-09
[skip netlify] rename a method to improve readability'

radekosmulski:
This is just a tiny PR. I was reading through `running_average.py` and was a bit confused what was going there because of the method name.

The output or state is not being updated, we are rather reading the value from the output and assigning to `src`.

Maybe `_read_output` would work even better? :thinking:


radekosmulski:
`_update_output` maybe actually makes more sense in the context of the naming across the library, not sure a small change like this makes sense anyhow, gonna close

PR #1930, opened 2021-04-11
Allow lr_finder to save files

### LINKED ISSUE #1926: [Feature Request] Save lr_finder plot instead of its showing
It would be great, if one can save plot of LRFinder module instead of showing it. Because, for example, in many cases we run our training scripts on Servers (e.g., AWS) and it is better to save the result plot (loss against learning rate plot) instead of showing it.
Thanks

Moh-Yakoub:
Fixes #1926

Description:

Allow saving files of the `LRFinder` module beside of showing it.
I've allowed passing params to add flexibility to the configuration of the saved plot

Check list:


ignite/contrib/handlers/lr_finder.py
```diff
@@ -186,7 +186,14 @@ def get_results(self) -> Dict[str, List[Any]]:
...
         return self._history
 
-    def plot(self, skip_start: int = 10, skip_end: int = 5, log_lr: bool = True) -> None:
+    def plot(
+        self,
+        skip_start: int = 10,
+        skip_end: int = 5,
+        log_lr: bool = True,
+        file_name: str = None,
+        file_params: dict = None,
```
vfdev-5, line 195:
```suggestion
        **kwargs: Any,
```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -186,7 +186,14 @@ def get_results(self) -> Dict[str, List[Any]]:
         """
         return self._history
 
-    def plot(self, skip_start: int = 10, skip_end: int = 5, log_lr: bool = True) -> None:
+    def plot(
+        self,
+        skip_start: int = 10,
+        skip_end: int = 5,
+        log_lr: bool = True,
+        file_name: str = None,
```
vfdev-5, line 194:
```suggestion
        filepath: str = None,
```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -202,6 +209,9 @@ def plot(self, skip_start: int = 10, skip_end: int = 5, log_lr: bool = True) ->
                 Default: 5.
             log_lr: True to plot the learning rate in a logarithmic
                 scale; otherwise, plotted in a linear scale. Default: True.
+            file_name: The file name to save the plot to. Default: None.
+            file_params: A set of params that can be passed to the
```
vfdev-5, line 213:
```suggestion
            kwargs: optional kwargs passed to ``plt.savefig`` if ``filepath`` is provided.
```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -238,6 +248,11 @@ def plot(self, skip_start: int = 10, skip_end: int = 5, log_lr: bool = True) ->
         plt.xlabel("Learning rate")
         plt.ylabel("Loss")
         plt.show()
+        if file_name is not None:
+            if file_params is not None:
+                plt.savefig(file_name, **file_params)
+            else:
+                plt.savefig(file_name)
```
vfdev-5, line 255:
```suggestion
            plt.savefig(filepath, **kwargs)
```

vfdev-5 (commented):
Thanks for the PR @Moh-Yakoub , please address my comments

ignite/contrib/handlers/lr_finder.py
```diff
@@ -202,6 +204,9 @@ def plot(self, skip_start: int = 10, skip_end: int = 5, log_lr: bool = True) ->
                 Default: 5.
             log_lr: True to plot the learning rate in a logarithmic
                 scale; otherwise, plotted in a linear scale. Default: True.
+            file_name: The file name to save the plot to. Default: None.
```
vfdev-5, line 207:
```suggestion
            filepath: The file name to save the plot to. Default: None.
```

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -230,6 +230,10 @@ def test_plot(lr_finder, to_save, dummy_engine, dataloader):
 
     lr_finder.plot()
     lr_finder.plot(skip_end=0)
+    lr_finder.plot(skip_end=0, file_name="dummy.jpg")
+    lr_finder.plot(
+        skip_end=0, file_name="dummy.jpg", file_params={"orientation": "portrait", "papertype": "a4", "format": "png",}
+    )
```
vfdev-5, line 236: This should be updated to

vfdev-5:
@Moh-Yakoub please fix mypy: https://github.com/pytorch/ignite/runs/2318812634

ignite/contrib/handlers/lr_finder.py
```diff
@@ -187,7 +187,7 @@ def get_results(self) -> Dict[str, List[Any]]:
         return self._history
 
     def plot(
-        self, skip_start: int = 10, skip_end: int = 5, log_lr: bool = True, filepath: str = None, **kwargs: Any,
+        self, skip_start: int = 10, skip_end: int = 5, log_lr: bool = True, filepath: str = "", **kwargs: Any,
```
vfdev-5, line 190:
```suggestion
        self, skip_start: int = 10, skip_end: int = 5, log_lr: bool = True, filepath: Optional[str] = None, **kwargs: Any,
```
Let's use optional instead.

Moh-Yakoub:
@vfdev-5 any more thoughts on this one?

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -230,6 +230,10 @@ def test_plot(lr_finder, to_save, dummy_engine, dataloader):
 
     lr_finder.plot()
     lr_finder.plot(skip_end=0)
+    lr_finder.plot(skip_end=0, filepath="dummy.jpg")
+    lr_finder.plot(
+        skip_end=0, filepath="dummy.jpg", orientation="landscape", papertype="a4", format="png",
```
vfdev-5, line 235: We do not check if file exists, right ? Can you please add this test ?

vfdev-5:
> @vfdev-5 any more thoughts on this one?

@Moh-Yakoub sorry for delay. I left another comment about the added test. Otherwise it looks good to be merged... 
Btw. have you checked if it works locally and we can see the image with the plot ?

tests/ignite/contrib/handlers/test_lr_finder.py
Moh-Yakoub: I've added the test as well as the extra try/check. That's really good catch. Thanks a lot

Moh-Yakoub:
> > @vfdev-5 any more thoughts on this one?
> 
> @Moh-Yakoub sorry for delay. I left another comment about the added test. Otherwise it looks good to be merged...
> Btw. have you checked if it works locally and we can see the image with the plot ?

Yes, I have tested using the example notebook and was able to reproduce the graph matching the output result.
Thanks a lot for reviewing this!

tests/ignite/contrib/handlers/test_lr_finder.py
vfdev-5: Actually, my comment was about the test:
```python
lr_finder.plot(skip_end=0, filepath="dummy.jpg")
assert Path("dummy.jpg").exists()
```
However, I agree that if we run it like : 
```python
lr_finder.plot(skip_end=0, filepath="/tmp/nonexisting-folder/dummy.jpg")
assert Path("/tmp/nonexisting-folder/dummy.jpg").exists()
```
we have to ensure parents folders creation as well.

Moh-Yakoub:
@vfdev-5 I've updated the test cases. the `savefig` doesn't create directory automatically if it doesn't exist. I added a test case to reflect that

vfdev-5 (approved):
Thanks for the PR @Moh-Yakoub 
I think we can improve it in another PR by returning `matplotlib.axes.Axes` as for example seaborn does for its drawing helper methods...

PR #1937, opened 2021-04-13
enable metric ordering on ProgressBar

radekosmulski:
**Description**: 

This allows for ordering of metrics on ProgressBar. The metrics will be displayed in the order that they are attached to the `Engine`.

I have not done any of the tasks from the checklist as I am not sure they are applicable here - happy to work on them if any would be needed!

Additional details on this change: #1936 

Check list:

### DIFF
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -285,7 +285,7 @@ class _OutputHandler(BaseOutputHandler):
                 rendered_metrics[key] = value
 
         if rendered_metrics:
-            logger.pbar.set_postfix(**rendered_metrics)  # type: ignore[attr-defined]
+            logger.pbar.set_postfix(rendered_metrics)  # type: ignore[attr-defined]
 
         global_step = engine.state.get_event_attrib_value(event_name)
         if pbar_total is not None:
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 """TQDM logger."""
 import warnings
+from collections import OrderedDict
 from typing import Any, Callable, List, Optional, Union
 
 import torch
@@ -270,7 +271,7 @@ class _OutputHandler(BaseOutputHandler):
 
         metrics = self._setup_output_metrics(engine)
 
-        rendered_metrics = {}
+        rendered_metrics = OrderedDict()
         for key, value in metrics.items():
             if isinstance(value, torch.Tensor):
                 if value.ndimension() == 0:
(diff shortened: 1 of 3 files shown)


radekosmulski:
Some tests are failing - will fix :slightly_smiling_face:

vfdev-5:
> Some tests are failing - will fix 🙂

Thanks for the PR @radekosmulski ! Yes, the failing tests should be adapted.

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -285,7 +285,7 @@ def __call__(self, engine: Engine, logger: ProgressBar, event_name: Union[str, E
                 rendered_metrics[key] = value
 
         if rendered_metrics:
-            logger.pbar.set_postfix(**rendered_metrics)  # type: ignore[attr-defined]
+            logger.pbar.set_postfix(rendered_metrics)  # type: ignore[attr-defined]
```
vfdev-5, line 288: @radekosmulski actually, I'm a bit confused about how we ensure the order with `rendered_metrics` being a dict ?
I think it wont work with plain dicts as they are not required to be ordered... 
The order is guaranteed since python 3.7 : https://docs.python.org/3/library/stdtypes.html?highlight=dict#dict

radekosmulski: That is a good point, I didn't think of this! I added a commit that I think should fix this.

As a summary -  with proposed changes, the ordering of metrics is set through the list of `metric_names` passed when attaching the progress bar.

For example,
```
pbar = ProgressBar(persist=False) \
    .attach(trainer, metric_names=['avg_loss', 'loss', 'gpu:0 mem(%)', 'gpu:0 util(%)'])
```

With `metric_names` set to `'all'` we again have a non-deterministic scenario as `engine.state.metrics` is a standard dict.

vfdev-5: Thanks @radekosmulski ! In case of "all" we can use a sorted list of keys and compute metrics variable like:
```python
metrics = OrderedDict([(k, engine.state.metrics[k]) for k in sorted(engine.state.metrics.keys())])
```

radekosmulski: There is one case that got me thinking...

If we keep `metrics = engine.state.metrics` on Python 3.7+ we can set the order of metrics with `'all'` by the order in which we attach metrics to the engine.

This to me seems like a nice behavior. On the other hand, it requires making the type annotations really verbose. I am also not sure what to think about nice functionality that is emergent, rather than explicitly defined in code.

With doing what you propose above, we are giving the user two clear paths:

- either give us `metric_names`
- or accept alphabetical ordering

Anyhow - maybe I am overthinking a completely irrelevant consideration in the longer run. This is something that got me thinking, I kind of personally liked the implicit behavior of metrics appearing in the order in which they were attached.

But thinking about it a bit more, and seeing how it looks in code, I can see how going with the `OrderedDict` with sorted key is the preferred way.

Please let me know your thoughts, which way would be better :)

vfdev-5: Good points, taking into account that python 3.6 support will be dropped since 2021-12-23 by python org: https://www.python.org/downloads/, let's just wrap engine's metric into OrderedDict for more clear code and we'll have attached order for python 3.7+ users
```python
metrics = OrderedDict(engine.state.metrics)
```

vfdev-5 (approved):
LGTM! Thanks @radekosmulski

PR #1944, opened 2021-04-16
[WIP] Possible fix for flaky hvd tests

### LINKED ISSUE #1942: Fix flaky Horovod CI
## 🐛 Bug description

- https://github.com/pytorch/ignite/runs/2357481982
- https://github.com/pytorch/ignite/runs/2360505516

Moh-Yakoub:
Fixes #1942

Description:

I propose that solution that adds a maximum of 3 retries with 2 seconds in between to fix the flaky hvd tests that are sometimes breaking the builds.

Check list:

### DIFF
--- a/tests/run_cpu_tests.sh
+++ b/tests/run_cpu_tests.sh
@@ -11,7 +11,7 @@ fi
 n=0
 until [ "$n" -ge 3 ]
 do
-   CUDA_VISIBLE_DEVICES="" pytest --tx 4*popen//python=python --cov ignite --cov-report term-missing --cov-report xml -vvv tests "${skip_distrib_opt[@]}"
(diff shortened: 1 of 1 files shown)


vfdev-5:
@Moh-Yakoub thanks for the PR ! At first I was thinking about inspecting if we can reproduce the issue locally and possibly fix it. I think it is related to distributed tests with FileStore...

Moh-Yakoub:
> @Moh-Yakoub thanks for the PR ! At first I was thinking about inspecting if we can reproduce the issue locally and possibly fix it. I think it is related to distributed tests with FileStore...

That makes sense. I felt this can be an alternate approach to not cause any furthur delays with PRs and on master while we fix that issue. But I agree that the inspection should be the long term goal

vfdev-5:
I closed related issue as CI seems to pass the last time. Let's close the PR as well

PR #1945, opened 2021-04-16
Fix issue #1763 : improve loggers to keep configuration 

### LINKED ISSUE #1763: Cannot set log level of `ignite.distributed` because of calls to setup_logger inside auto functions
## 🐛 Bug description

The `ignite.distributed` auto functions such as `ignite.distributed.auto_dataloader` use `setup_logger` to get a `logging.Logger` instance. However, this function always sets the logging level to `logging.INFO` and furthermore removes ands sets up the logging handlers. Functions should not be setting up a logger object if it is already initialized such that user code can specify a different log level, handlers, or log message format string in their code (e.g., main function).

See for example:
- https://github.com/pytorch/ignite/blob/6d490c8f3083f21ba8f82eb646d9052a06fae3e2/ignite/distributed/auto.py#L67
- https://github.com/pytorch/ignite/blob/6d490c8f3083f21ba8f82eb646d9052a06fae3e2/ignite/distributed/launcher.py#L201

Note that in case of `ignite.distributed.launcher.Parallel` one can modify the `self.logger` instance, but this logger already prints messages right after `setup_logger` in its `__init__` function, hence, making it impossible to adjust the log level, format, or handlers for these first messages also.

## Environment

- OS (e.g., Linux):
- Python version:

sdesrozis:
Fixes #1763

Description:

Idea is to be able to keep configuration of loggers.

```python
# first configuration
logger = setup_logger(name="my-logger", format="=== %(name)s %(message)s")
logger.info("first message")

# second configuration
setup_logger(name="my-logger", format="+++ %(name)s %(message)s", reset=True)
logger.info("second message")

# Logs will look like
# === my-logger first message
# +++ my-logger second message
```

By default, we keep existing configuration of loggers. 

For instance to change level of messages from `auto_*.py`
```python
logger = setup_logger("ignite.distributed.auto.auto_dataloader", level=logging.WARNING)
```
Same to change the level of messages from launcher
```python
logger = setup_logger(name="ignite.distributed.launcher.Parallel", level=logging.WARNING)

with idist.Parallel(backend="gloo", nproc_per_node=4) as parallel:
    parallel.run(main_fn)
```

Warning : the root logger should be reset to avoid additional default messages.
```python
logger = setup_logger(format="%(name)s %(message)s")
logger.info("a message")

# Logs will look like
# INFO:root:a message
# root a message
```

Check list:

### DIFF
--- a/ignite/distributed/launcher.py
+++ b/ignite/distributed/launcher.py
@@ -229,7 +229,7 @@ class Parallel:
         self.backend = backend
         self._spawn_params = None
         self.init_method = init_method
-        self.logger = setup_logger(__name__ + "." + self.__class__.__name__, distributed_rank=0)
+        # self.logger = setup_logger(__name__ + "." + self.__class__.__name__, distributed_rank=0)
         # distributed_rank=0 <=> explicit rank 0, avoid call idist. Critical for TPU on Colab, avoid context setup
 
         if self.backend is not None:
@@ -238,10 +238,10 @@ class Parallel:
                     nproc_per_node, nnodes, node_rank, master_addr, master_port, init_method, **spawn_kwargs
                 )
 
-        if self._spawn_params is not None:
-            self.logger.info(f"Initialized distributed launcher with backend: '{self.backend}'")
-            msg = "\n\t".join([f"{k}: {v}" for k, v in self._spawn_params.items() if v is not None])
-            self.logger.info(f"- Parameters to spawn processes: \n\t{msg}")
+        # if self._spawn_params is not None:
+        #     self.logger.info(f"Initialized distributed launcher with backend: '{self.backend}'")
+        #     msg = "\n\t".join([f"{k}: {v}" for k, v in self._spawn_params.items() if v is not None])
+        #     self.logger.info(f"- Parameters to spawn processes: \n\t{msg}")
 
     @staticmethod
     def _setup_spawn_params(
@@ -315,11 +315,15 @@ class Parallel:
         self.logger.info("End of run")
 
     def __enter__(self) -> "Parallel":
-        if (self.backend is not None) and self._spawn_params is None:
-            idist.initialize(self.backend, init_method=self.init_method)
+        if self.backend is not None:
             self.logger = setup_logger(__name__ + "." + self.__class__.__name__)
-            self.logger.info(f"Initialized processing group with backend: '{self.backend}'")
-
+            if self._spawn_params is None:
+                idist.initialize(self.backend, init_method=self.init_method)
+                self.logger.info(f"Initialized processing group with backend: '{self.backend}'")
+            else:
+                self.logger.info(f"Initialized distributed launcher with backend: '{self.backend}'")
+                msg = "\n\t".join([f"{k}: {v}" for k, v in self._spawn_params.items() if v is not None])
+                self.logger.info(f"- Parameters to spawn processes: \n\t{msg}")
         return self
 
     def __exit__(self, *args: Any, **kwargs: Any) -> None:
--- a/ignite/utils.py
+++ b/ignite/utils.py
@@ -91,6 +91,7 @@ def setup_logger(
     format: str = "%(asctime)s %(name)s %(levelname)s: %(message)s",
     filepath: Optional[str] = None,
     distributed_rank: Optional[int] = None,
+    reset: bool = False,
 ) -> logging.Logger:
     """Setups logger: name, level, format etc.
 
@@ -102,6 +103,7 @@ def setup_logger(
         filepath: Optional logging file path. If not None, logs are written to the file.
         distributed_rank: Optional, rank in distributed configuration to avoid logger setup for workers.
             If None, distributed_rank is initialized to the rank of process.
+        reset: if True, reset an existing logger rather than keep format, handlers, and level.
 
     Returns:
         logging.Logger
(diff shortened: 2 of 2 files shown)


vfdev-5:
@sdesrozis thanks for the PR !

> Warning : the root logger should be reset to avoid additional default messages.

Can you explain this point ? And maybe provide some code to see better what we have to do...


About this code:
```python
logger = setup_logger(name="ignite.distributed.launcher.Parallel", level=logging.WARNING)

with idist.Parallel(backend="gloo", nproc_per_node=4) as parallel:
    parallel.run(main_fn)
```

will this work with torch distributed launcher ? We setup `logger` for all ranks here ?

sdesrozis:
> > Warning : the root logger should be reset to avoid additional default messages.
> 
> Can you explain this point ? And maybe provide some code to see better what we have to do...

Sorry, my warning is wrong. We don't have additional messages but there is a specific behavior that could be not relevant.

I already warned about possible issues concerning the root logger. Indeed, this logger is the default one and always exists. So, because `reset=False` by default to let user's handlers and configuration, we use the default logger. It means `level=logging.WARNING`. 

Note that is it only impacting when a logger is created without a name.
```python
logger = setup_logger()
```
To tackle this, an idea would be to have a default name rather than `None`.

> About this code:
> 
> ```python
> logger = setup_logger(name="ignite.distributed.launcher.Parallel", level=logging.WARNING)
> 
> with idist.Parallel(backend="gloo", nproc_per_node=4) as parallel:
>     parallel.run(main_fn)
> ```
> 
> will this work with torch distributed launcher ? We setup `logger` for all ranks here ?

It works with `torch.distributed.launch` as well. In the spawn case, logger is used before and after spawning then, logger is sequential.

ignite/distributed/launcher.py
```diff
@@ -315,11 +309,15 @@ def training(local_rank, config, **kwargs):
         self.logger.info("End of run")
 
     def __enter__(self) -> "Parallel":
-        if (self.backend is not None) and self._spawn_params is None:
-            idist.initialize(self.backend, init_method=self.init_method)
-            self.logger = setup_logger(__name__ + "." + self.__class__.__name__)
-            self.logger.info(f"Initialized processing group with backend: '{self.backend}'")
-
+        self.logger = setup_logger(__name__ + "." + self.__class__.__name__)
```
vfdev-5, line 312: Nit: lint always suggests to set up attributes inside `__init__`.

ignite/utils.py
```diff
@@ -127,34 +129,68 @@ def setup_logger(
...
+    Example to change the level of an existing internal logger
+
+    .. code-block:: python
+
+        setup_logger(
+            name="ignite.distributed.launcher.Parallel",
+            level=logging.WARNING
+        )
+
+    Warning : the root logger should be reset because it already exists and default level is `logging.WARNING`
```
vfdev-5, line 154: This warning is not clear and what we have to do with that.

tests/ignite/test_utils.py
```diff
@@ -120,8 +120,8 @@ def test_setup_logger(capsys, dirname):
 
     def _test(stream):
 
-        trainer.logger = setup_logger("trainer", stream=stream, filepath=fp)
-        evaluator.logger = setup_logger("evaluator", stream=stream, filepath=fp)
+        trainer.logger = setup_logger("trainer", stream=stream, filepath=fp, reset=True)
```
vfdev-5, line 123: Can we add a test as in the issue: use auto method by default than override the logger and ensure that nothing in stdout ?

vfdev-5 (commented):
Thanks for the PR @sdesrozis ! Few comments to address please

ignite/distributed/launcher.py
sdesrozis: If we use spawn, logger defined in `__init__` will be only on master process. I've preferred defining only once but outside `__init__`

ignite/distributed/launcher.py
```diff
@@ -229,7 +229,6 @@ def __init__(
         self.backend = backend
         self._spawn_params = None
         self.init_method = init_method
-        self.logger = setup_logger(__name__ + "." + self.__class__.__name__, distributed_rank=0)
         # distributed_rank=0 <=> explicit rank 0, avoid call idist. Critical for TPU on Colab, avoid context setup
```
vfdev-5, line 232: Do you respect this comment also for TPUs ? 
We can remove the comment by the way

ignite/distributed/launcher.py
vfdev-5: Actually, I still do not understand the purpose of moving it here.

sdesrozis: We always define a logger per process. Each logger is initialized wrt his rank so process with `rank=0` is initialized to write in streams.

In the actual code, we use two times `setup_logger`
* for `spawn`. It is done before process were spawned so here, we have only one logger. As you can see, in the spawn flow, this logger is not used. 
https://github.com/pytorch/ignite/blob/a2343fa8caa5a8887af691f2eb06fa77e4b0845a/ignite/distributed/launcher.py#L232
* for `torch.distributed.launch`. The setup of the logger is done right after the process group initialization, it means rank is defined, so each logger can be properly initialized wrt his rank.
https://github.com/pytorch/ignite/blob/a2343fa8caa5a8887af691f2eb06fa77e4b0845a/ignite/distributed/launcher.py#L320

The following initialization seems weird to me because it is done before the initialization of process group.
https://github.com/pytorch/ignite/blob/a2343fa8caa5a8887af691f2eb06fa77e4b0845a/ignite/distributed/launcher.py#L232
Think about it in the `torch.distributed.launch`, each logger per process is setup as `rank=0`. That is not correct. I suppose that's why a second setup is called for `torch.distributed.launch` as I mentioned above. To me, the comment shows a problem, not a fix... We should not use internally logger before initialize the process group. Or, if we use it (even in colab), it should works.

ignite/distributed/launcher.py
sdesrozis: `idist` doesn't work on colab ? I completely missed that.

Btw, this setup should be exclusively done on colab, because it had side effect.
```python
parallel = idist.Parallel(backend="gloo", nproc_per_node=1)
parallel.logger.info("who print this ?")
```

I would prefer remove `self.logger` and catch a fresh one when we need to write something.

vfdev-5: Almost all idist functions work on colab but there are few restrictions to keep in mind when you set up the dist model: if we call `xm.*` methods in a single proc, it will initialize single tpu device context and wont be able to init 8 tpus with multiprocessing.

vfdev-5: I think we should also forbid the usage of `idist.Parallel` without context manager as in your example.

ignite/distributed/launcher.py
vfdev-5:
```
self.logger = setup_logger(__name__ + "." + self.__class__.__name__, distributed_rank=0)
```
I agree that this is a hack to avoid calls like `idist.get_rank()` which will init 1 proc XLA context before calling `idist.spawn` to spawn 8 procs.

vfdev-5: I was also thinking if we should only rely on idist to create the logger with `setup_logger` ?
For example, we can think of using other ways to resolve the problem of creating single writing logger by using filelock things ...

ignite/distributed/launcher.py
sdesrozis: > Almost all idist functions work on colab but there are few restrictions to keep in mind when you set up the dist model: if we call `xm.*` methods in a single proc, it will initialize single tpu device context and wont be able to init 8 tpus with multiprocessing.

It sounds like as we should not use loggers (mostly internally) before process group initialization.

ignite/distributed/launcher.py
sdesrozis: > I was also thinking if we should only rely on idist to create the logger with `setup_logger` ?
> For example, we can think of using other ways to resolve the problem of creating single writing logger by using filelock things ...

I don't know if it would work due to filesystem concerns, shared or not, permission or not, etc.

sdesrozis:
My thoughts about this PR:
* don't use the root logger of `logging` adding a default name rather than `None` in the method `setup_logger()`
* don't use loggers before call `init_process_group` to avoid configuration issues.

Pretty simple and straightforward.

What do you think ? @vfdev-5

vfdev-5:
@sdesrozis sorry for delay. How will we ensure these conditions from the user side ? Seems like only on docs  ?

sdesrozis:
> @sdesrozis sorry for delay. How will we ensure these conditions from the user side ? Seems like only on docs  ?

If we don't use root logger of `logging` (it means always have a name for the loggers, i.e. choose à default name), user can do what he wants, that does not conflict with our loggers.

On the other hand, we should not use loggers internally in `Parallel` before the processes would be spawned. That is on our side. If a user use a logger before that, we can't avoid it and we should not do because it is intentional. Btw, I think that users would use loggers in the spawned function, not outside.

vfdev-5:
@sdesrozis thanks for the code snippet ! I let you take a lead on that and decide what would be the optimal solution we could provide to the users.

vfdev-5:
@sdesrozis are we done with this PR or still WIP ?

sdesrozis:
I have a little improvement to push. I'm sorry for the delay, I will do it very soon.

vfdev-5:
@sdesrozis let me know when we are done with this PR : final review and merge ?

vfdev-5 (approved):
Thanks @sdesrozis ! LGTM

PR #1947, opened 2021-04-17
Inception Score Implementation

ahmedo42:
Related Issue #998 

Description:

To reproduce inception score on CIFAR10 check this notebook

https://colab.research.google.com/drive/1odZBAzltKl6OmNuA7zsnkeXLsh5YPEhi?usp=sharing

References:

- https://github.com/sbarratt/inception-score-pytorch
- https://github.com/NVlabs/stylegan2-ada-pytorch/blob/main/metrics/inception_score.py

Check list:

### DIFF
--- a/ignite/contrib/handlers/neptune_logger.py
+++ b/ignite/contrib/handlers/neptune_logger.py
@@ -42,7 +42,7 @@ class NeptuneLogger(BaseLogger):
     Args:
         api_token: Required in online mode. Neputne API token, found on https://neptune.ai.
             Read how to get your API key
-            https://docs.neptune.ai/administration/security-and-privacy/how-to-find-and-set-neptune-api-token.
+            https://docs.neptune.ai/integrations-and-supported-tools/ide-and-notebooks/jupyter-lab-and-jupyter-notebook/notebooks-1#where-is-neptune_api_token.
         project_name: Required in online mode. Qualified name of a project in a form of
            "namespace/project_name" for example "tom/minst-classification".
            If None, the value of NEPTUNE_PROJECT environment variable will be taken.
--- a/ignite/metrics/gan/inception_score.py
+++ b/ignite/metrics/gan/inception_score.py
@@ -1,0 +1,116 @@
+from typing import Callable, List, Optional, Union, cast
+
+import torch
+from torch import nn
+from torch.nn import functional as F
+
+import ignite.distributed as idist
+from ignite.exceptions import NotComputableError
+from ignite.metrics.metric import Metric, reinit__is_reduced
+
+__all__ = ["InceptionScore"]
+
+
+class InceptionScore(Metric):
+    r"""Computes the Inception Score for evaluating the quality of synthetic images.
+
+    More details can be found in `Improved Techniques for Training GANs`__.
+
+    __ https://arxiv.org/pdf/1606.03498.pdf
+
+
+    - ``update`` must receive output of the form ``y_pred`` or ``{'y_pred': y_pred}``.
+    - ``y_pred`` should have the following shape (batch_size, ) and contains syntheic images created by the generator.
+
+
+    .. warning::
+
+        Current implementation stores all input data before computing the metric.
+        This can potentially lead to a memory error if the input data is larger than available RAM.
+
+        In distributed configuration, all stored data is mutually collected across all processes
+        using all gather collective operation. This can potentially lead to a memory error.
+
+    Args:
+        splits: number of splits to calculate the mean inception score.
+
+        inception_model: model used for extracting class probabilities from images If not specified,
+            InceptionV3 Pretrained on ImageNet will be used.
+        output_transform: a callable that is used to transform the
+            :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
+            form expected by the metric. This can be useful if, for example, you have a multi-output model and
+            you want to compute the metric with respect to one of the outputs.
+        device: specifies which device updates are accumulated on. Setting the metric's
+            device to be the same as your ``update`` arguments ensures the ``update`` method is non-blocking. By
+            default, CPU.
+
+    Example:
+
+    .. code-block:: python
+
+        from ignite.metrics.gan import InceptionScore
+
+        m = InceptionScore(splits=4)
+
+        y_pred = torch.rand((16,3,299,299))  # size must match the input shape of the InceptionV3 model
+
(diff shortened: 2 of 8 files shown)


sdesrozis:
@ahmedo42 Thank you for this PR! As usual, I suppose that @vfdev-5 will be very fast to review but I will do my best too since this topic is very interesting to me.

ignite/contrib/handlers/neptune_logger.py
```diff
@@ -42,7 +42,7 @@ class NeptuneLogger(BaseLogger):
     Args:
         api_token: Required in online mode. Neputne API token, found on https://neptune.ai.
             Read how to get your API key
-            https://docs.neptune.ai/administration/security-and-privacy/how-to-find-and-set-neptune-api-token.
+            https://docs.neptune.ai/integrations-and-supported-tools/ide-and-notebooks/jupyter-lab-and-jupyter-notebook/notebooks-1#where-is-neptune_api_token.
```
vfdev-5, line 45: @ahmedo42 let's revert this change as I removed those links in the master.

ignite/metrics/gan/inception_score.py
```diff
@@ -0,0 +1,123 @@
...
+        self.n_splits = splits
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._probs = []  # type: List[torch.Tensor]
+
+    @reinit__is_reduced
+    def update(self, output: torch.Tensor) -> None:
+        generated = output.detach()
+        inception_output = self.inception_model(generated)
```
vfdev-5, line 95: We should add `with torch.no_grad()` context manager here to avoid grads creation.

ignite/metrics/gan/inception_score.py
```diff
@@ -0,0 +1,123 @@
...
+            _probs_tensor = cast(torch.Tensor, idist.all_gather(_probs_tensor))
+        self._is_reduced = True
+
+        result = 0.0
+        if idist.get_rank() == 0:
+            N = _probs_tensor.shape[0]
+            scores = torch.zeros((self.n_splits,))
+            for i in range(self.n_splits):
+                part = _probs_tensor[i * (N // self.n_splits) : (i + 1) * (N // self.n_splits)]
+                kl = part * (torch.log(part) - torch.log(torch.mean(part, dim=0)))
```
vfdev-5, line 116: Please, if we could reuse pytorch functions to compute KL. We can also probably avoid applying softmax in the `update` pass.

ignite/metrics/gan/inception_score.py
```diff
@@ -0,0 +1,123 @@
...
+                kl = torch.mean(torch.sum(kl, dim=1))
+                scores[i] = torch.exp(kl)
+            result = torch.mean(scores).item()
```
vfdev-5, line 119: Can't we accumulate `kl` on the `update` to avoid saving all `probs` ?

vfdev-5 (commented):
Thanks for the PR @ahmedo42 
I think we can have this metric with online implementation.

ignite/metrics/gan/inception_score.py
ahmedo42: If what you meant is this:
```python
    @reinit__is_reduced
    def update(self, output: torch.Tensor) -> None:
        generated = output.detach()
        with torch.no_grad():
            inception_output = self.inception_model(generated)
        probs = self._apply_softmax(inception_output)
        kl = self._compute_kl(probs)
        self._kl += kl
        self._num_examples += probs.shape[0]

    @sync_all_reduce("_kl","_num_examples")
    def compute(self) -> float:
        return self._kl.item() / self._num_examples
```
I think this is not valid. 

The problem is how to incorporate `splits` into the calculations , when you have the whole dataset it's easy to split it into well-defined segments ,calculate the `kl` for each and then average .

vfdev-5: @ahmedo42 you can think of `n_splits` of KL accumulators. Here is an examples with counter (which you can replace with KL) : 
```python
n_splits = 12
batch_size = 32
num_samples = 2000

n_iters = (num_samples + batch_size - 1) // batch_size 

# reset():
iteration = 0
counter = torch.zeros(n_splits)

for j in range(n_iters):
    
    # update():
    for i in range(batch_size):
        counter[iteration % n_splits] += 1
        iteration += 1

> 
> counter
> tensor([168., 168., 168., 168., 168., 168., 168., 168., 168., 168., 168., 168.])
```


Have you also seen this paper: https://arxiv.org/abs/1801.01973 where they suggest to get rid of `n_splits` ?

ahmedo42: @vfdev-5  thanks for the idea , this is the implementation based on what you said , the inception score from this is 10.3 (compared to 9.6 of the previous implementation under the same conditions):
```python

    @reinit__is_reduced
    def update(self, output: torch.Tensor) -> None:
        generated = output.detach()
        with torch.no_grad():
            inception_output = self.inception_model(generated)
        probs = self._apply_softmax(inception_output)
        batch_size = probs.shape[0]
        for i in range(batch_size):
            self._kl[self._iteration % self.n_splits] += self._compute_kl(probs[i])
            self._iteration += 1
        self._num_examples += probs.shape[0]

    @sync_all_reduce("_kl","_num_examples")
    def compute(self) -> float:
        return torch.mean(self._kl).item() / self._num_examples

    def _compute_kl(self,input_tensor:torch.Tensor):
        kl = input_tensor * (torch.log(input_tensor) - torch.log(torch.mean(input_tensor, dim=0)))
        kl = torch.sum(kl, dim=0)
        return torch.exp(kl)
```
I saw the paper and I think we should provide the original implementation ( as it's still widely used now ) , then we can extend that.

vfdev-5: I think this implementation is incorrect due to `_compute_kl` : 
- `torch.log(torch.mean(input_tensor, dim=0))` originally computes the average on `num_samples // n_splits` of tensors and in your case it is batch size
-  `torch.exp(kl)` should not be applied here.

vfdev-5:
@ahmedo42 do you plan to continue working on this PR ?

ahmedo42:
@vfdev-5 Sorry about the absence , I've been busy with college and will probably be for the next two weeks ( midterms ) , Iet's close it and then reopen it I think .

vfdev-5:
Closing the PR in favor of #2053

PR #1951, opened 2021-04-19
Return axes instead of plotting for LRFinder

Moh-Yakoub:
Description:

Based on the discussion we had here https://github.com/pytorch/ignite/pull/1930#pullrequestreview-637461553 I propose replacing the plot method with returning axis as seaborn API

Please not I specified the return type of the method as ANY since we don't import the matplotlib library by default underwhich axes are defined

Check list:

### DIFF
--- a/ignite/contrib/handlers/lr_finder.py
+++ b/ignite/contrib/handlers/lr_finder.py
@@ -76,7 +76,7 @@ class FastaiLRFinder:
         self._history = {}  # type: Dict[str, List[Any]]
         self._best_loss = None
         self._lr_schedule = None  # type: Optional[Union[LRScheduler, PiecewiseLinear]]
-        self.logger = logging.getLogger(__name__)
+        self.logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
 
     def _run(
         self,
@@ -182,14 +182,23 @@ class FastaiLRFinder:
 
     def get_results(self) -> Dict[str, List[Any]]:
         """
-        Returns: dictionary with loss and lr logs fromm the previous run
+        Returns:
+            Dictionary with loss and lr logs from the previous run
         """
         return self._history
 
-    def plot(self, skip_start: int = 10, skip_end: int = 5, log_lr: bool = True, **kwargs: Any,) -> Any:
+    def plot(
+        self,
+        skip_start: int = 10,
+        skip_end: int = 5,
+        log_lr: bool = True,
+        display_suggestion: bool = True,
+        axes: Any = None,
+        **kwargs: Any,
+    ) -> None:
         """Plots the learning rate range test.
 
-        This method requires `matplotlib` package to be installed:
+        This method requires ``matplotlib`` package to be installed:
 
         .. code-block:: bash
 
@@ -202,10 +211,10 @@ class FastaiLRFinder:
                 Default: 5.
             log_lr: True to plot the learning rate in a logarithmic
                 scale; otherwise, plotted in a linear scale. Default: True.
-            filepath: The file name to save the plot to. Default: None.
-            kwargs: optional kwargs passed to ``plt.subplots`` used to generate the results.
-
-        Returns: an matplotlib.axes.Axes object that can be used to plot results
+            display_suggestion: if True, red dot shows the suggested learning rate.
+            axes: An axes object used for plotting. Default: None.
+            kwargs: optional kwargs passed to ``plt.savefig`` if ``filepath`` is provided.
+                matplotlib method `savefig` to save the plot.
         """
         try:
             from matplotlib import pyplot as plt
@@ -214,7 +223,6 @@ class FastaiLRFinder:
                 "This method requires matplotlib to be installed. "
                 "Please install it with command: \n pip install matplotlib"
             )
-
         if not self._history:
             raise RuntimeError("learning rate finder didn't run yet so results can't be plotted")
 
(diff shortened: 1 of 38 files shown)


vfdev-5:
@Moh-Yakoub thanks for the PR ! Looks like it is backward compatibility (BC) breaking. We have to think about if it is possible to keep it compatible.

ignite/contrib/handlers/lr_finder.py
```diff
@@ -210,8 +203,9 @@ def plot(
             log_lr: True to plot the learning rate in a logarithmic
                 scale; otherwise, plotted in a linear scale. Default: True.
             filepath: The file name to save the plot to. Default: None.
```
vfdev-5, line 205: filepath was removed from args.

vfdev-5 (commented):
Inspecting more the notebook, I think we can still keep it as 
```python
lr_finder.plot()
```
without explicit matplotlib import and `.plot(ax)` call.
The point was talking about seaborn etc is that it can accept `ax` as argument => we can plot Loss curve on a given axis and thus user can embed into other subplots etc.

We are now removing filepath (which is OK), but maybe we could provide an example on how to save the plot on the disk as it was asked in the other issue...

Moh-Yakoub:
> Inspecting more the notebook, I think we can still keep it as
> 
> ```python
> lr_finder.plot()
> ```
> 
> without explicit matplotlib import and `.plot(ax)` call.
> The point was talking about seaborn etc is that it can accept `ax` as argument => we can plot Loss curve on a given axis and thus user can embed into other subplots etc.
> 

@vfdev-5 what do you think of passing an `ax` argument that can be None, and inevitably the user can either use it, or if it's not used, we just plot as the existing behavior. what do you think?

vfdev-5:
> > Inspecting more the notebook, I think we can still keep it as
> > ```python
> > lr_finder.plot()
> > ```
> > 
> > 
> > without explicit matplotlib import and `.plot(ax)` call.
> > The point was talking about seaborn etc is that it can accept `ax` as argument => we can plot Loss curve on a given axis and thus user can embed into other subplots etc.
> 
> @vfdev-5 what do you think of passing an `ax` argument that can be None, and inevitably the user can either use it, or if it's not used, we just plot as the existing behavior. what do you think?

Yes, `ax` arg with `ax=None` by default is something I was thinking about as you say here.
As a return type, we can return `ax` either created internally or the one passed as input arg...

ignite/contrib/handlers/lr_finder.py
```diff
@@ -193,7 +193,7 @@ def plot(
         skip_end: int = 5,
         log_lr: bool = True,
         display_suggestion: bool = True,
-        filepath: Optional[str] = None,
+        axes: Any = None,
```
vfdev-5, line 196: Let's call it `ax` as in seaborn: https://seaborn.pydata.org/generated/seaborn.lineplot.html#seaborn.lineplot

ignite/contrib/handlers/lr_finder.py
```diff
@@ -212,7 +212,7 @@ def plot(
             log_lr: True to plot the learning rate in a logarithmic
                 scale; otherwise, plotted in a linear scale. Default: True.
             display_suggestion: if True, red dot shows the suggested learning rate.
-            filepath: The file name to save the plot to. Default: None.
+            axes: An axes object used for plotting. Default: None.
```
vfdev-5, line 215:
```suggestion
            ax: Pre-existing axes for the plot. Default: None.
```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -264,18 +263,17 @@ def plot(
             losses = losses[skip_start:-skip_end]
 
         # Plot loss as a function of the learning rate
-        plt.plot(lrs, losses)
+        if axes is None:
+            f, axes = plt.subplots(**kwargs)
```
vfdev-5, line 267: This part of code should be executed before `if display_suggestion:` and we have to remove `fig, ax = plt.subplots()`, line 252.

ignite/contrib/handlers/lr_finder.py
```diff
@@ -212,7 +212,7 @@ def plot(
             log_lr: True to plot the learning rate in a logarithmic
                 scale; otherwise, plotted in a linear scale. Default: True.
             display_suggestion: if True, red dot shows the suggested learning rate.
-            filepath: The file name to save the plot to. Default: None.
+            axes: An axes object used for plotting. Default: None.
             kwargs: optional kwargs passed to ``plt.savefig`` if ``filepath`` is provided.
                 matplotlib method `savefig` to save the plot.
```
vfdev-5, line 217: I think I already asked you, please add an example code snippet to this docstring on how to save the figure.

vfdev-5 (commented):
Thanks for the update @Moh-Yakoub , some more updates are required and we can merge it

vfdev-5:
As there are more LR finder improvements coming (e.g. #1981), @Moh-Yakoub let's merge this PR before.

cc @KickItLikeShika for visibility.

ignite/contrib/handlers/lr_finder.py
```diff
@@ -193,7 +193,7 @@ def plot(
         skip_end: int = 5,
         log_lr: bool = True,
         display_suggestion: bool = True,
-        filepath: Optional[str] = None,
+        ax: Any = None,
```
vfdev-5, line 196:
```suggestion
        ax: Optional[Any] = None,
```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -212,9 +212,14 @@ def plot(
...
                 scale; otherwise, plotted in a linear scale. Default: True.
             display_suggestion: if True, red dot shows the suggested learning rate.
-            filepath: The file name to save the plot to. Default: None.
-            kwargs: optional kwargs passed to ``plt.savefig`` if ``filepath`` is provided.
-                matplotlib method `savefig` to save the plot.
+            ax: Pre-existing axes for the plot. Default: None.
+            kwargs: optional kwargs passed to ``plt.subplots`` if ``ax`` is not provided.
+
+        .. code-block:: python
+
```
vfdev-5, line 219:
```suggestion

```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -212,9 +212,13 @@ def plot(
...
                 scale; otherwise, plotted in a linear scale. Default: True.
             display_suggestion: if True, red dot shows the suggested learning rate.
-            filepath: The file name to save the plot to. Default: None.
-            kwargs: optional kwargs passed to ``plt.savefig`` if ``filepath`` is provided.
-                matplotlib method `savefig` to save the plot.
+            ax: Pre-existing axes for the plot. Default: None.
+            kwargs: optional kwargs passed to ``plt.subplots`` if ``ax`` is not provided.
+
+        .. code-block:: python
+            ax = lr_finder.plot(skip_end=0)
```
vfdev-5, line 219:
```suggestion
        .. code-block:: python

            ax = lr_finder.plot(skip_end=0)
```

vfdev-5 (approved):
LGTM, thanks @Moh-Yakoub

Issue #1938, opened 2021-04-13
Usage of persistent_workers with auto_dataloader for spawned processes


vfdev-5:
## 🚀 Feature

Since v1.7.0 pytorch introduced `persistent_workers` argument to `DataLoader`: https://pytorch.org/docs/1.7.0/data.html#torch.utils.data.DataLoader
This can reduce the time of dataloader creation for each epoch with native pytorch distributed config:
- with `persistent_workers=True`, our cifar10 example gives `CIFAR10-Training INFO: Engine run complete. Time taken: 00:02:55` on 2 procs
- with `persistent_workers=False`, the same example gives 

We have to take that into account for related dist backend when using `auto_dataloader`.

Moh-Yakoub:
@vfdev-5 Are you suggesting enabling that field for all our examples that utilize the DataLoader?

vfdev-5:
@Moh-Yakoub I'm not yet sure about what exactly we'd like to do. We have to check "horovod" and "xla" backends as well with `persistent_workers=True`.

sdesrozis:
The performance gap is quite impressive. However, I don't understand clearly the reason why. If I understand well, the option allows to keep worker threads in dataloaders through epochs. I don't understand why it accelerates the computation.

vfdev-5:
@sdesrozis there is no acceleration, actually, in practice that the app takes a lot of time in preparing the first batch for each epoch.

sdesrozis:
10s gained by destroy/allocate threads (and memory so far) for 24 epochs ? It seems a very huge cost.

PR #1952, opened 2021-04-19
[WIP][Proposition] Allow passing `persistent workers` to benchmarks

Moh-Yakoub:
Addresses #1938

Description: 

I am proposing adding the persisting workers to our bench-marking utils so that we can use it for benchmarks and assess the validity and impact of introducing that new param as explained in the issue.

Check list:

### DIFF
--- a/examples/contrib/cifar100_amp_benchmark/benchmark_fp32.py
+++ b/examples/contrib/cifar100_amp_benchmark/benchmark_fp32.py
@@ -11,14 +11,16 @@ from ignite.handlers import Timer
 from ignite.metrics import Accuracy, Loss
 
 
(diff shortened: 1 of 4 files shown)


vfdev-5:
Thanks for the PR @Moh-Yakoub , but why adding the parameter here ? 
Yes, we can also assess the impact of that on AMP, here is let's say the executing part : https://github.com/pytorch/ignite/blob/master/examples/notebooks/Cifar100_bench_amp.ipynb
But I think we wont see much difference here.

As said [here](https://github.com/pytorch/ignite/issues/1938#issuecomment-819092791) we'd like to see the impact on horovod and xla in priority.

PR #1961, opened 2021-04-22
Improve LRFinder

KickItLikeShika:
- Updated the way we find the suggested learning rate: for more details check this notebook: https://colab.research.google.com/drive/1_siUuBjZkHVc_PGS28gmF0SYzKDgMbKJ?usp=sharing#scrollTo=naYcP7kvRAXq (Thanks for @vfdev-5 for helping me)
- Added a red dot to show the suggested learning rate (provided in the notebook)
- Added MNIST tests for LRFinder.

And thanks for @sdesrozis for suggesting dividing the PR into more smaller PRs.

And regarding the [`apply_suggested_lr`]( https://github.com/pytorch/ignite/pull/1961#issuecomment-826281096), it will be addressed in another PR
Check list:

### DIFF
--- a/ignite/contrib/handlers/lr_finder.py
+++ b/ignite/contrib/handlers/lr_finder.py
@@ -255,7 +255,7 @@ class FastaiLRFinder:
                 sug_lr, corresponding_loss, s=75, marker="o", color="red", zorder=3,
             )
 
-        # handle skip_end=0 properly so the behaviour is the expected
+        # handle skip_end=0 properly
         if skip_end == 0:
             lrs = lrs[skip_start:]
             losses = losses[skip_start:]
--- a/ignite/contrib/handlers/lr_finder.py
+++ b/ignite/contrib/handlers/lr_finder.py
@@ -182,7 +182,8 @@ class FastaiLRFinder:
 
     def get_results(self) -> Dict[str, List[Any]]:
         """
-        Returns: Dictionary with loss and lr logs from the previous run
+        Returns:
+            Dictionary with loss and lr logs from the previous run
         """
         return self._history
 
@@ -279,8 +280,9 @@ class FastaiLRFinder:
 
     def lr_suggestion(self) -> Any:
         """
-        Returns: Learning rate at the minimum numerical gradient
-        (ignores the increasing part of the curve)
+        Returns:
+            Learning rate at the minimum numerical gradient
+            (ignoring the increasing part of the curve)
         """
         if not self._history:
             raise RuntimeError("learning rate finder didn't run yet so lr_suggestion can't be returned")
--- a/ignite/contrib/handlers/lr_finder.py
+++ b/ignite/contrib/handlers/lr_finder.py
@@ -93,7 +93,6 @@ class FastaiLRFinder:
         self._history = {"lr": [], "loss": []}
         self._best_loss = None
         self._diverge_flag = False
-        self._min_grad_idx = -1
 
         # attach LRScheduler to trainer.
         if num_iter is None:
@@ -237,6 +236,25 @@ class FastaiLRFinder:
 
         lrs = self._history["lr"]
         losses = self._history["loss"]
+
+        # Check to show the suggested learning rate
+        if display_suggestion:
+            sug_lr = self.lr_suggestion()
+            idx = self._history["lr"].index(sug_lr)
+
+            if skip_start >= idx:
+                warnings.warn(
+                    "skip_start is large, an that would make less clear, and it may affects displaying the suggestion",
+                    UserWarning,
+                )
+
+            corresponding_loss = self._history["loss"][int(idx)]
+
+            fig, ax = plt.subplots()
+            ax.scatter(
+                sug_lr, corresponding_loss, s=75, marker="o", color="red", zorder=3,
+            )
+
         if skip_end == 0:
             lrs = lrs[skip_start:]
             losses = losses[skip_start:]
@@ -244,20 +262,11 @@ class FastaiLRFinder:
             lrs = lrs[skip_start:-skip_end]
             losses = losses[skip_start:-skip_end]
 
-        # Check to show the suggested learning rate
-        if display_suggestion:
-            idx = self._min_grad_idx + skip_start - skip_end
-            sug_lr = lrs[idx]
-            equivalent_loss = losses[idx]
-            fig, ax = plt.subplots()
-            ax.scatter(
-                sug_lr, equivalent_loss, s=75, marker="o", color="red", zorder=3,
-            )
-
         # Plot loss as a function of the learning rate
         plt.plot(lrs, losses)
         if log_lr:
             plt.xscale("log")
+        plt.xlim([lrs[0], lrs[-1]])
         plt.xlabel("Learning rate")
         plt.ylabel("Loss")
         if filepath is not None:
@@ -283,8 +292,7 @@ class FastaiLRFinder:
         losses = torch.tensor(decreasing_losses)
         grads = torch.tensor([losses[i] - losses[i - 1] for i in range(1, len(losses))])
         min_grad_idx = grads.argmin() + 1
-        self._min_grad_idx = int(min_grad_idx)
-        return self._history["lr"][self._min_grad_idx]
+        return self._history["lr"][int(min_grad_idx)]
 
     @contextlib.contextmanager
     def attach(
--- a/ignite/contrib/handlers/lr_finder.py
+++ b/ignite/contrib/handlers/lr_finder.py
@@ -292,7 +292,7 @@ class FastaiLRFinder:
         if len(decreasing_losses) == 1:
             raise RuntimeError("FastaiLRFinder got unexpected curve shape, the curve should be somehow U-shaped")
         losses = torch.tensor(decreasing_losses)
-        grads = torch.tensor([losses[i] - losses[i - 1] for i in range(1, len(losses))])
+        grads = torch.tensor([0.5 * (losses[i + 1] - losses[i - 1]) for i in range(1, len(losses) - 1)])
         min_grad_idx = grads.argmin() + 1
         return self._history["lr"][int(min_grad_idx)]
 
--- a/ignite/contrib/handlers/lr_finder.py
+++ b/ignite/contrib/handlers/lr_finder.py
@@ -182,7 +182,7 @@ class FastaiLRFinder:
 
     def get_results(self) -> Dict[str, List[Any]]:
         """
-        Returns: dictionary with loss and lr logs from the previous run
+        Returns: Dictionary with loss and lr logs from the previous run
         """
         return self._history
 
@@ -279,7 +279,7 @@ class FastaiLRFinder:
 
     def lr_suggestion(self) -> Any:
         """
-        Returns: the learning rate at the minimum numerical gradient
+        Returns: Learning rate at the minimum numerical gradient
         (ignores the increasing part of the curve)
         """
         if not self._history:
--- a/ignite/contrib/handlers/lr_finder.py
+++ b/ignite/contrib/handlers/lr_finder.py
@@ -243,8 +243,8 @@ class FastaiLRFinder:
 
             if skip_start >= idx:
                 warnings.warn(
-                    "skip_start is large, that would make the plot less clear,"
-                    " and it may affects displaying the suggestion",
+                    "skip_start is larger than the suggested LR found"
+                    " and it will not be visible on the plot. Please, make the value smaller.",
                     UserWarning,
                 )
 
--- a/ignite/contrib/handlers/lr_finder.py
+++ b/ignite/contrib/handlers/lr_finder.py
@@ -232,9 +232,7 @@ class FastaiLRFinder:
         if skip_end < 0:
             raise ValueError("skip_end cannot be negative")
 
-        # Get the data to plot from the history dictionary. Also, handle skip_end=0
-        # properly so the behaviour is the expected
-
+        # Get the data to plot from the history dictionary.
         lrs = self._history["lr"]
         losses = self._history["loss"]
 
@@ -257,6 +255,7 @@ class FastaiLRFinder:
                 sug_lr, corresponding_loss, s=75, marker="o", color="red", zorder=3,
             )
 
+        # handle skip_end=0 properly so the behaviour is the expected
         if skip_end == 0:
             lrs = lrs[skip_start:]
             losses = losses[skip_start:]
--- a/ignite/contrib/handlers/lr_finder.py
+++ b/ignite/contrib/handlers/lr_finder.py
@@ -244,7 +244,8 @@ class FastaiLRFinder:
 
             if skip_start >= idx:
                 warnings.warn(
-                    "skip_start is large, an that would make less clear, and it may affects displaying the suggestion",
+                    "skip_start is large, that would make the plot less clear,"
+                    " and it may affects displaying the suggestion",
                     UserWarning,
                 )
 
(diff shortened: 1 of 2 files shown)


sdesrozis:
@KickItLikeShika Thank you for this work. It seems good.

I'm just wondering why don't generate loss curves using handcrafted output_transform functions ? The engine and model don't matter.

ignite/contrib/handlers/lr_finder.py
```diff
@@ -256,14 +269,22 @@ def plot(
...
+        # Ignore the increasing part of the curve
+        decreasing_losses = self._history["loss"][: int(min_loss_idx.item()) + 1]
+        if len(decreasing_losses) == 1:
+            raise RuntimeError("FastaiLRFinder got unexpected curve shape, the curve should be somehow U-shaped")
+        losses = torch.tensor(decreasing_losses)
+        grads = torch.tensor([losses[i] - losses[i - 1] for i in range(1, len(losses))])
         min_grad_idx = grads.argmin() + 1
-        return self._history["lr"][int(min_grad_idx)]
+        self._min_grad_idx = int(min_grad_idx)
+        return self._history["lr"][self._min_grad_idx]
```
vfdev-5, line 287: I think we can keep `_min_grad_idx` as a local var `min_grad_idx` instead of instance attribute `self._min_grad_idx`.

ignite/contrib/handlers/lr_finder.py
```diff
@@ -241,6 +244,16 @@ def plot(
             lrs = lrs[skip_start:-skip_end]
             losses = losses[skip_start:-skip_end]
 
+        # Check to show the suggested learning rate
+        if display_suggestion:
+            idx = self._min_grad_idx + skip_start - skip_end
```
vfdev-5, line 249: This can be wrong if user does not call `lr_suggestion` before.
I think it would be better to call `lr_suggestion` here instead of using `self._min_grad_idx` which we should not store as instance attribute.

ignite/contrib/handlers/lr_finder.py
```diff
@@ -241,6 +244,16 @@ def plot(
...
+            sug_lr = lrs[idx]
+            equivalent_loss = losses[idx]
```
vfdev-5, line 251: We can also either find the loss from the history which corresponds to suggested lr or simply draw vertical line.

vfdev-5 (commented):
Please, address the comments

KickItLikeShika:
@sdesrozis sorry I can't understand clearly, can you be more explicit please?

ignite/contrib/handlers/lr_finder.py
KickItLikeShika: Okay, plot the values right we have to find the index `sug_lr` not in the history but in the `lrs` list in the `plot` function example: 
```
        lrs = self._history["lr"]
        losses = self._history["loss"]
        if skip_end == 0:
            lrs = lrs[skip_start:]
            losses = losses[skip_start:]
        else:
            lrs = lrs[skip_start:-skip_end]
            losses = losses[skip_start:-skip_end]

        if display_suggestion:
            sug_lr = self.lr_suggestion()
            idx = lrs.index(sug_lr)
            print(idx)
            equivalent_loss = losses[int(idx)]
```
And that can go wrong and raise error like 
```
 # Check to show the suggested learning rate
        if display_suggestion:
            sug_lr = self.lr_suggestion()
            idx = lrs.index(sug_lr)
            ValueError: 0.00012589254117941674 is not in list
```
To avoid this we can find the index of the `sug_lr` in the list but this also can go wrong and the plot can be something like this
dummy

KickItLikeShika: The simplest solution is to manipulate the `skip_start`, for example we can put it initially to zero instead of 10 in the plot function and mention that in the docs, or even remove it!

vfdev-5: @KickItLikeShika I do not understand how skip_start can solve the issues. For me the simplest solution without changing the API is to draw vertical line instead of a circle if we can not determine correctly the loss value associated with suggested lr. Please, be more explicit in your suggestions...

KickItLikeShika: The problem isn't with finding the index of the loss, It's with finding the index of the suggested learning rate in the `lrs` list in the plot function, drawing a vertical line won't change anything in this case, and If we tried to find the suggested learning rate in the `history` the plot can be something like the image above

KickItLikeShika: Let's assume that `skip_start = 0`, almost the `self._history["lrs"]` will be equal to `lrs` list in the plot function (the `skip_end` won't matter as we mostly ignore that part), then we will have no issues finding the suggested learning rate, and that will get us a proper plot at the end

KickItLikeShika: I will call `lr_suggestion` then I got the value of `suggested_lr` that found in the whole history, with its index in the history I will find the corresponding loss in the history, then I can't apply the `skip_start` and `skip_end`, right?

KickItLikeShika: For more details:

If indexes found in the history:
```
the history of lrs: [1, 10]
the history of losses: [1, 10]
lrs list in plot function after skip star and end: [3, 9]
losses list in plot function after skip start and ed: [3, 9]
suggested_lr: 2
corresponding loss: 2
```
-------> the plot is wrong

If indexes found the lists
```
Raise a ValueError the lr is not found
```

vfdev-5: @KickItLikeShika please try that instead of speculating. I'm sure that it is not a big deal to make it work with out `self._min_grad_idx`. If there is a real blocker, put explicitly the code line and give an explanation why you think we can not go this way.

KickItLikeShika: I'm already trying everything
Here is the code you told me to try (I think we can't apply `skip_start` and `skip_end` as we have the values not the indexes)
```
if display_suggestion:
            sug_lr = self.lr_suggestion()
            idx = self._history["lr"].index(sug_lr)
            corresponding_loss = self._history["loss"][int(idx)]
            fig, ax = plt.subplots()
            ax.scatter(
                sug_lr, corresponding_loss, s=75, marker="o", color="red", zorder=3,
            )
```
And here is the result
dummy

vfdev-5: OK, have you tried to move this code
```python
        if skip_end == 0:
            lrs = lrs[skip_start:]
            losses = losses[skip_start:]
        else:
            lrs = lrs[skip_start:-skip_end]
            losses = losses[skip_start:-skip_end]
```
after `if display_suggestion: ...` code ??

KickItLikeShika: I tried it now, and it gives the same results

vfdev-5: OK, please find another solution without `self._min_grad_idx`.

KickItLikeShika: The result with `skip_start = 0` (always gives correct results)

dummy

To wrap up: The issue is related to the visualization interval, what happens is: if the suggested lr index is less than the `skip_start` value the red dot won't be on the curve, It will be before the curve as I mentioned above (assuming we will keep getting the index from the history not the lists)

Suggested solution:
- Make the `skip_start` initially to zero, and put a check inside the `if display_suggestion:` to check if the index of the lr is smaller than the `skip_start`, then we can raise a `ValueError` to ask to reduce the `skip_start` value, what do you think? I can think about another solution...

sdesrozis:
> @sdesrozis sorry I can't understand clearly, can you be more explicit please?

Sorry if my thoughts weren't clearly explained.

I meant we could use an empty engine
```python
trainer = Engine(lambda _, b: b)
```
and a fake optimizer
```python
tensor = torch.zeros([1], requires_grad=True)
optimizer = torch.optim.SGD([tensor], lr=0.0)
```
and a function `output_transform` that gives the loss we want
```python
x = np.linspace(-1.0, 1.0, num=10)**2
x = list(x)

def output_transform(_):
    return x.pop()
```

No need model, trainer, etc.

@KickItLikeShika What do you think ? 

HTH

KickItLikeShika:
@sdesrozis regarding the models/optimizers/dataloaders/MNIST, these were @vfdev-5 ideas, He worked on this issue once and he told me to consider his updates, please check his branch https://github.com/pytorch/ignite/compare/master...vfdev-5:issue-837-lrfinder-ddp

sdesrozis:
MNIST is a cheap integration test very interesting to consider.

What I suggested is to perfectly control the loss curves without deal with model, trainer and optimizer concerns. Btw it's just a suggestion.

KickItLikeShika:
Your suggestions are really nice! My opinion is simulating some real curves is something really good to be in our tests! And as you said it's cheap!

sdesrozis:
I suggested to use `output_transform` but a more straightforward way is to use a tuned `update_step` in the `Engine`!

```python
x = np.linspace(-1, 1, 100)**2
loss = list(x)

trainer = Engine(lambda _, b: loss.pop())
```

Sometimes, simple things are too simple.

vfdev-5:
> I suggested to use `output_transform` but a more straightforward way is to use a tuned `update_step` in the `Engine`!
> 
> ```python
> x = np.linspace(-1, 1, 100)**2
> loss = list(x)
> 
> trainer = Engine(lambda _, b: loss.pop())
> ```
> 
> Sometimes, simple things are too simple.

The main problem with this approach I see is that we have to store a list of real-like loss values. Simulating a parabola curve can be not that useful. That's why integration MNIST test in addition to unittest.

ignite/contrib/handlers/lr_finder.py
```diff
@@ -76,7 +76,7 @@ def __init__(self) -> None:
         self._history = {}  # type: Dict[str, List[Any]]
         self._best_loss = None
         self._lr_schedule = None  # type: Optional[Union[LRScheduler, PiecewiseLinear]]
-        self.logger = logging.getLogger(__name__)
+        self.logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
```
vfdev-5, line 79: Should we do something here with the logger @sdesrozis ?

ignite/contrib/handlers/lr_finder.py
```diff
@@ -182,7 +182,7 @@ def _detach(self, trainer: Engine) -> None:
 
     def get_results(self) -> Dict[str, List[Any]]:
         """
-        Returns: dictionary with loss and lr logs fromm the previous run
+        Returns: dictionary with loss and lr logs from the previous run
```
vfdev-5, line 185:
```suggestion
        Returns:
            Dictionary with loss and lr logs from the previous run
```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -256,12 +279,19 @@ def plot(
 
     def lr_suggestion(self) -> Any:
         """
-        Returns: learning rate at the minimum numerical gradient
+        Returns: the learning rate at the minimum numerical gradient
+        (ignores the increasing part of the curve)
```
vfdev-5, line 283:
```suggestion
        Returns: the learning rate at the minimum numerical gradient
        (ignores the increasing part of the curve)
```
```suggestion
        Returns:
            Learning rate at the minimum numerical gradient
            (ignoring the increasing part of the curve)
```

ignite/contrib/handlers/lr_finder.py
sdesrozis: The logger has a name according to the handler (i.e. not root logger), it can be used in the `Parallel` context and be overridden by users. It looks perfect.

sdesrozis:
> The main problem with this approach I see is that we have to store a list of real-like loss values. Simulating a parabola curve can be not that useful. That's why integration MNIST test in addition to unittest.

I agree that the MNIST test is very relevant. My suggestion was about unit tests only, to perfectly control the correctness and the behaviour of the lr finder. But that is not important, these are just thoughts I wanted to share.

vfdev-5:
> > The main problem with this approach I see is that we have to store a list of real-like loss values. Simulating a parabola curve can be not that useful. That's why integration MNIST test in addition to unittest.
> 
> I agree that the MNIST test is very relevant. My suggestion was about unit tests only, to perfectly control the correctness and the behaviour of the lr finder. But that is not important, these are just thoughts I wanted to share.

I agree with you that an intermediate test between unittests we have and mnist integration is missing.

ignite/contrib/handlers/lr_finder.py
vfdev-5, line 283:
```suggestion
        Returns:
            Learning rate at the minimum numerical gradient
            (ignoring the increasing part of the curve)
```

vfdev-5:
@KickItLikeShika should we also update notebook tutorial with this new code ?
How about providing a helper method with applying suggested lr to the optimizer ?

KickItLikeShika:
@vfdev-5 what exactly needs to be updated? (The example is fine for now at least as we didn't make many changes). And can you please explain this helper method suggestion?

vfdev-5:
@KickItLikeShika first, please pay attention on docstring format, especially on what I suggested as change and what you did.

You can see the difference if you render the docs or see the rendering on netlify ci job.

> @vfdev-5 what does need to be updated in the example? (The example is fine for now at least, as we didn't make many changes). And can you please explain this helper method suggestion?

OK, maybe nothing to change in the notebook right now, but one line with the helper method to apply lr to the optimizer.
We discussed that during the call, we would maybe to add a method to FastaiLRfinder:
```python
lr_finder.apply_suggested_lr(optimizer)
```
instead of the way we are doing it in the notebook.

ignite/contrib/handlers/lr_finder.py
```diff
@@ -229,11 +232,30 @@ def plot(
...
+
+        # Check to show the suggested learning rate
+        if display_suggestion:
+            sug_lr = self.lr_suggestion()
+            idx = self._history["lr"].index(sug_lr)
+
+            if skip_start >= idx:
+                warnings.warn(
+                    "skip_start is large, that would make the plot less clear,"
+                    " and it may affects displaying the suggestion",
```
vfdev-5, line 247:
```suggestion
                    "skip_start is larger then the suggested LR found"
                    " and it will not be visible on the plot. Please, make the value smaller.",
```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -256,12 +279,20 @@ def plot(
...
             raise RuntimeError("learning rate finder didn't run yet so lr_suggestion can't be returned")
         loss = self._history["loss"]
-        grads = torch.tensor([loss[i] - loss[i - 1] for i in range(1, len(loss))])
+        min_loss_idx = torch.tensor(loss).argmin()
+        # Ignore the increasing part of the curve
+        decreasing_losses = self._history["loss"][: int(min_loss_idx.item()) + 1]
+        if len(decreasing_losses) == 1:
+            raise RuntimeError("FastaiLRFinder got unexpected curve shape, the curve should be somehow U-shaped")
+        losses = torch.tensor(decreasing_losses)
+        grads = torch.tensor([losses[i] - losses[i - 1] for i in range(1, len(losses))])
```
vfdev-5, line 295: @KickItLikeShika finally, why we do not use more stable schema to determine grads with 3 points ?

KickItLikeShika: I tried to find a formula to find the grads with 3 points but I got nothing then, so I thought this way works really well so no need to change it, but I can dig more for the 3 points grads formula

vfdev-5: I had it already in my branch: `0.5 * (loss[i + 1] - loss[i - 1])` ?

KickItLikeShika: Thanks, I was getting it wrong! I will try to use it now and see the results

KickItLikeShika: 3 points grads give very stable results while ignoring the increasing part, see here: https://colab.research.google.com/drive/1_siUuBjZkHVc_PGS28gmF0SYzKDgMbKJ?usp=sharing#scrollTo=sBpNO3Q9n55r
. I will update the `lr_suggestion` method now

vfdev-5 (approved):
LGTM! Thanks @KickItLikeShika

Issue #1939, opened 2021-04-14
EpochOutputStore can optionally store the data to the state


vfdev-5:
## 🚀 Feature

The idea is to give an option to have the following:
```python
eos = EpochOutputStore()

evaluator = create_supervised_evaluator(model, metrics)
eos.attach(evaluator, name="eval_data")

# Adding the name is equivalent to
# name = "eval_data"
# @evaluator.on(Events.EPOCH_COMPLETED, name)
# def store_data(engine, name):
#     setattr(engine.state, name, eos.data)

state = evaluator.run(data)
print(len(state.eval_data))
```

In addition, it would make sense to move it from contrib to core as well, keeping BC.

radekosmulski:
Read the code of `EpochOutputStore` and I think I understand what this is about :slightly_smiling_face: Maybe this is something I could work on after I finish #1919?

Could I please ask what BC stands for?

ydcjeff:
BC - backward compatibility

vfdev-5:
Thanks Jeff! 

@radekosmulski Yes, BC is backward compatibility and a PR breaking it is called "BC-breaking" as far as I know.

vfdev-5:
@radekosmulski would you like to work on this issue ?

radekosmulski:
@vfdev-5 - would love to! Will read through the code today and either come here to ask questions if anything is unclear or will start working on a draft of the implementation :slightly_smiling_face:

radekosmulski:
I opened #1974 but it is probably quite basic, happy to implement any changes that might be needed :slightly_smiling_face: 

This is only a draft - as a next step, will move it out of contrib keeping BC.

PR #1974, opened 2021-04-27
give an option to store `EpochOutputStore` data on `engine.state`

radekosmulski:
Implements #1939

Description: When `EpochOutputStore` is attached and a `name` is provided, at `EPOCH_COMPLETED` `engine.state.{name}` is set to `eos.data` (in other words, `eos.data` is stored on `engine.state`).

Check list:

### DIFF
--- a/ignite/contrib/handlers/stores.py
+++ b/ignite/contrib/handlers/stores.py
@@ -1,4 +1,4 @@
-from typing import Callable, List, Tuple, Union
+from typing import Callable, List, Optional, Tuple, Union
 
 from ignite.engine import Engine, Events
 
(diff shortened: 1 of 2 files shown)


ignite/contrib/handlers/stores.py
```diff
@@ -45,8 +45,15 @@ def update(self, engine: Engine) -> None:
         output = self.output_transform(engine.state.output)
         self.data.append(output)
 
-    def attach(self, engine: Engine) -> None:
+    def store(self, engine: Engine) -> None:
```
sdesrozis, line 48: @radekosmulski I think the docstring should be update to show this new feature.

radekosmulski: sorry, completely missed that, will add! :slightly_smiling_face:

ignite/contrib/handlers/stores.py
```diff
@@ -45,8 +45,15 @@ def update(self, engine: Engine) -> None:
...
+        """Store `self.data` on `engine.state.{self.name}`"""
+        setattr(engine.state, self.name, self.data)
+
+    def attach(self, engine: Engine, name: Optional[str] = None) -> None:
         """Attaching `reset` method at EPOCH_STARTED and
         `update` method at ITERATION_COMPLETED."""
```
vfdev-5, line 54: @radekosmulski please also update the docstring here.

vfdev-5:
@radekosmulski thanks for the PR, looks good, just a docstring update remains. If you would like to move this class into core in another PR, it is OK for me (maybe even prefered way to do that).

radekosmulski:
I added the changes in two commits - one updates the docstrings, the other the example. I am not sure if the update to the example is needed? 

Generally, not sure if this is the right way to do it :slightly_smiling_face: 

Will move it out of `contrib` in a separate PR :slightly_smiling_face:

vfdev-5 (approved):
LGTM! Thanks @radekosmulski

PR #1987, opened 2021-05-05
Add `apply_suggested_lr` to LRFinder

KickItLikeShika:
Added `apply_suggested_lr` to LRFinder, added tests, updated tests to use MNIST, and updated the example.
Check list:

### DIFF
--- a/examples/notebooks/FastaiLRFinder_MNIST.ipynb
+++ b/examples/notebooks/FastaiLRFinder_MNIST.ipynb
@@ -197,14 +197,14 @@
     "lr_finder = FastaiLRFinder()\n",
     "to_save={'model': model, 'optimizer': optimizer}\n",
     "with lr_finder.attach(trainer, to_save, diverge_th=1.5) as trainer_with_lr_finder:\n",
-    "    trainer_with_lr_finder.run(trainloader)\n",
-    "    \n",
-    "trainer.run(trainloader, max_epochs=10)\n",
-    "\n",
-    "evaluator = create_supervised_evaluator(model, metrics={\"acc\": Accuracy(), \"loss\": Loss(nn.NLLLoss())}, device=device)\n",
-    "evaluator.run(testloader)\n",
-    "\n",
-    "print(evaluator.state.metrics)"
+    "    trainer_with_lr_finder.run(trainloader)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let's plot the found optimal learning rate."
    ]
   },
   {
@@ -218,18 +218,35 @@
     "from matplotlib import pyplot as plt\n",
     "ax = lr_finder.plot()\n",
     "plt.axes(ax)\n",
-    "plt.show()"
+    "plt.show()\n",
+    "\n",
+    "print(\"Suggested LR\", lr_finder.lr_suggestion())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "EHHNoOi8mvd6"
+   },
+   "source": [
+    "Let's compare training results: \n",
+    "- a) using a fixed learning rate (`lr=3e-4`)\n",
+    "- b) with applying optimal learning rate."
    ]
   },
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "id": "EHHNoOi8mvd6"
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
-    "lr_finder.lr_suggestion()"
+    "# with lr=3e-4\n",
+    "trainer.run(trainloader, max_epochs=10)\n",
+    "\n",
+    "evaluator = create_supervised_evaluator(model, metrics={\"acc\": Accuracy(), \"loss\": Loss(nn.NLLLoss())}, device=device)\n",
+    "evaluator.run(testloader)\n",
+    "\n",
+    "print(evaluator.state.metrics)"
    ]
   },
   {
@@ -246,16 +263,36 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "lr_finder.apply_suggested_lr(optimizer)\n",
+    "print(optimizer.param_groups[0]['lr'])"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# We can reset the parameters of our model\n",
+    "# or we can just create new instance for the model\n",
+    "new_model = Net()\n",
+    "new_model.to(device)\n",
+    "\n",
+    "new_trainer = create_supervised_trainer(new_model, optimizer, criterion, device=device)\n",
+    "ProgressBar(persist=True).attach(new_trainer, output_transform=lambda x: {\"batch loss\": x})"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
    "metadata": {
     "id": "DJqgyaFnmvd7"
    },
    "outputs": [],
    "source": [
-    "new_optimizer = lr_finder.apply_suggested_lr(optimizer)\n",
-    "\n",
-    "new_trainer = create_supervised_trainer(model, new_optimizer, criterion, device=device)\n",
-    "ProgressBar(persist=True).attach(new_trainer, output_transform=lambda x: {\"batch loss\": x})\n",
-    "\n",
     "new_trainer.run(trainloader, max_epochs=10)\n",
     "\n",
     "evaluator.run(testloader)\n",
--- a/examples/notebooks/FastaiLRFinder_MNIST.ipynb
+++ b/examples/notebooks/FastaiLRFinder_MNIST.ipynb
@@ -238,7 +238,7 @@
     "id": "NcT19wqkmvd6"
    },
    "source": [
-    "Let's now setup suggested learning rate to the optimizer and we can train the model with optimal learning rate.\n",
+    "Let's now apply suggested learning rate to the optimizer and create a new trainer, then we can train the model with optimal learning rate.\n",
     "\n",
     "*Note, that model and optimizer were restored to their initial states when `FastaiLRFinder` finished.*"
    ]
@@ -251,9 +251,13 @@
    },
    "outputs": [],
    "source": [
-    "optimizer.param_groups[0]['lr'] = lr_finder.lr_suggestion()\n",
+    "new_optimizer = lr_finder.apply_suggested_lr(optimizer)\n",
     "\n",
-    "trainer.run(trainloader, max_epochs=10)\n",
+    "new_trainer = create_supervised_trainer(model, new_optimizer, criterion, device=device)\n",
+    "ProgressBar(persist=True).attach(new_trainer, output_transform=lambda x: {\"batch loss\": x})\n",
+    "\n",
+    "new_trainer.run(trainloader, max_epochs=10)\n",
+    "\n",
     "evaluator.run(testloader)\n",
     "print(evaluator.state.metrics)"
    ]
@@ -280,7 +284,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.10"
+   "version": "3.8.3"
   },
   "pycharm": {
    "stem_cell": {
--- a/examples/notebooks/FastaiLRFinder_MNIST.ipynb
+++ b/examples/notebooks/FastaiLRFinder_MNIST.ipynb
@@ -180,7 +180,12 @@
     "criterion = nn.NLLLoss()\n",
     "model = Net()\n",
     "model.to(device)  # Move model before creating optimizer\n",
-    "optimizer = optim.SGD(model.parameters(), lr=3e-4, momentum=0.9)"
+    "optimizer = optim.SGD(model.parameters(), lr=3e-4, momentum=0.9)\n",
+    "\n",
+    "# save the initial state for the model and the optimizer\n",
+    "# to reset them later\n",
+    "init_model_state = model.state_dict()\n",
+    "init_opt_state = optimizer.state_dict()"
    ]
   },
   {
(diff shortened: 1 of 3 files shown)


ignite/contrib/handlers/lr_finder.py
```diff
@@ -310,6 +310,34 @@ def lr_suggestion(self) -> Any:
...
+            optimizer: the optimizer to apply the suggested learning rate(s) on.
+
+        Returns:
+                The optimizer with the new learning rate(s).
+        """
+        sug_lr = self.lr_suggestion()
+        if isinstance(sug_lr, list):
+            if len(sug_lr) != len(optimizer.param_groups):
+                raise RuntimeError(
+                    "Applying the suggested learning rate can't be done, the optimizers are not the same"
```
vfdev-5, line 330:
```suggestion
                    "The number of parameter groups does not match between "
                    "given optimizer and the one used for estimating the "
                    f"learning rate: {len(sug_lr)} vs {len(optimizer.param_groups)}"
```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -310,6 +310,34 @@ def lr_suggestion(self) -> Any:
...
+                )
+            i = 0
+            for lr in sug_lr:
+                optimizer.param_groups[i]["lr"] = lr
+                i += 1
```
vfdev-5, line 335:
```suggestion
            for i, lr in enumerate(sug_lr):
                optimizer.param_groups[i]["lr"] = lr
```

ignite/contrib/handlers/lr_finder.py
```diff
@@ -310,6 +310,34 @@ def lr_suggestion(self) -> Any:
         min_grad_idx = grads.argmin() + 1
         return self._history["lr"][int(min_grad_idx)]
 
+    def apply_suggested_lr(self, optimizer: Optimizer) -> Optimizer:
+        """
+        Applying the suggested learning rate(s) on the given optimizer.
```
vfdev-5, line 315: Technically, it is inplace method. Let's return nothing to enphasize that.

examples/notebooks/FastaiLRFinder_MNIST.ipynb
```diff
@@ -238,7 +238,7 @@
     "id": "NcT19wqkmvd6"
```
vfdev-5, line 238: Let's update this notebook as we discussed previously:
- Let's split this cell into multiple ones:
```
ProgressBar(persist=True).attach(trainer, output_transform=lambda x: {"batch loss": x})

lr_finder = FastaiLRFinder()
to_save={'model': model, 'optimizer': optimizer}
with lr_finder.attach(trainer, to_save, diverge_th=1.5) as trainer_with_lr_finder:
    trainer_with_lr_finder.run(trainloader)
    
trainer.run(trainloader, max_epochs=10)

evaluator = create_supervised_evaluator(model, metrics={"acc": Accuracy(), "loss": Loss(nn.NLLLoss())}, device=device)
evaluator.run(testloader)

print(evaluator.state.metrics)
```
->
```python
# CELL 1
ProgressBar(persist=True).attach(trainer, output_transform=lambda x: {"batch loss": x})

lr_finder = FastaiLRFinder()
to_save={'model': model, 'optimizer': optimizer}
with lr_finder.attach(trainer, to_save, diverge_th=1.5) as trainer_with_lr_finder:
    trainer_with_lr_finder.run(trainloader)
    
# CELL 2 - Text
Let's plot the found optimal learning rate.

# CELL 3
ax = lr_finder.plot()
plt.axes(ax)
plt.show()

print("Suggested LR", lr_finder.lr_suggestion())

# CELL 4 - Text
Let's compare training results: a) using a fixed learning rate (`lr=3e-4`) and b) with applying optimal learning rate.

# CELL 5
trainer.run(trainloader, max_epochs=10)

evaluator = create_supervised_evaluator(model, metrics={"acc": Accuracy(), "loss": Loss(nn.NLLLoss())}, device=device)
evaluator.run(testloader)

print(evaluator.state.metrics)
```

examples/notebooks/FastaiLRFinder_MNIST.ipynb
```diff
@@ -251,9 +251,13 @@
    },
    "outputs": [],
    "source": [
-    "optimizer.param_groups[0]['lr'] = lr_finder.lr_suggestion()\n",
+    "new_optimizer = lr_finder.apply_suggested_lr(optimizer)\n",
+    "\n",
+    "new_trainer = create_supervised_trainer(model, new_optimizer, criterion, device=device)\n",
```
vfdev-5, line 256: You have to use new model or restore weights from scratch.

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -337,21 +337,41 @@ def test_lr_suggestion(lr_finder, to_save, dummy_engine, dataloader):
...
 ):
 
-    with lr_finder.attach(
-        dummy_engine_single_param_groups, to_save_single_param_groups, end_lr=20, smooth_f=0.04
-    ) as trainer_with_finder:
-        trainer_with_finder.run(dataloader_plot)
+    with lr_finder.attach(dummy_engine_mnist, mnist_to_save) as trainer_with_finder:
+        trainer_with_finder.run(mnist_dataloader)
 
-    # Avoid RuntimeError
```
vfdev-5, line 349: Can you also update `test_lr_suggestion` and remove : 
```python
    # Avoid RuntimeError
    if torch.tensor(lr_finder._history["loss"]).argmin() < 2:
        lr_finder._history["loss"].insert(0, 10)
        lr_finder._history["loss"].insert(0, 100)
        lr_finder._history["lr"].insert(0, 10)
        lr_finder._history["lr"].insert(0, 100)

    assert 1e-4 <= lr_finder.lr_suggestion() <= 10
```
from there ?

vfdev-5 (commented):
Few comments to fix

tests/ignite/contrib/handlers/test_lr_finder.py
KickItLikeShika: I thought about that, but If we did that we will have to MNIST dataset, and if we did that, there will be no difference between `test_lr_suggestion` and `test_mnist_lr_suggestion`, so we can leave `test_lr_suggestion` as it is or we can delete it and rename `test_mnist_lr_suggestion` to `test_lr_suggestion`, and consider it the learning rate test function, what do you think?

vfdev-5: Let's do the following: let's manually create loss/lr history without calling
```python
    with lr_finder.attach(dummy_engine, to_save) as trainer_with_finder:
```
and check lr_suggestion on this synthetic data.
Another comment:
```python
assert 1e-4 <= lr_finder.lr_suggestion() <= 10
```
this assertion should be stricter, internal [1e-4, 10] is too large.

KickItLikeShika:
That's strange! these failures didn't happen locally!

KickItLikeShika:
It seems to be working fine now! Let's now rename `test_mnist_lr_suggestion` to `test_lr_suggestion_mnist`. And that's will help us double check everything

examples/notebooks/FastaiLRFinder_MNIST.ipynb
```diff
@@ -238,11 +255,36 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# We can reset the parameters of our model\n",
+    "# or we can just create new instance for the model\n",
+    "new_model = Net()\n",
+    "new_model.to(device)\n",
```
vfdev-5, line 282: If you use new model, you have to set parameters to the optimizer as it has old model's params right now

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -323,35 +317,105 @@ def test_lr_suggestion_unexpected_curve(lr_finder, to_save, dummy_engine, datalo
...
+    # 2 param_groups
+    lr_finder._history["lr"] = torch.tensor(np.linspace(0.01, 10, num=100)).reshape(50, 2)
+
+    # lr_finder.lr_suggestion() is supposed to return a list of values,
+    # but as we assign loss and lr to tensors, instead of lists, it will return tensors
+    suggested_lrs = lr_finder.lr_suggestion()
+
+    # Ignoring the increasing part of the curve in the assertion.
+    assert 0.1 <= suggested_lrs[0].item() <= 6
+    assert 0.1 <= suggested_lrs[1].item() <= 6
```
vfdev-5, line 348: Can we make these assertions more precise: `pytest.approx(suggested_lrs[0]) == 0.xyz`

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -323,35 +317,105 @@ def test_lr_suggestion_unexpected_curve(lr_finder, to_save, dummy_engine, datalo
...
-        lr_finder._history["lr"].insert(0, 10)
-        lr_finder._history["lr"].insert(0, 100)
+    with lr_finder.attach(dummy_engine_mnist, mnist_to_save) as trainer_with_finder:
+
+        with trainer_with_finder.add_event_handler(
+            Events.ITERATION_COMPLETED(once=max_iters), lambda _: trainer_with_finder.terminate()
+        ):
+            trainer_with_finder.run(mnist_dataloader)
 
     assert 1e-4 <= lr_finder.lr_suggestion() <= 10
```
vfdev-5, line 362: Same here, let's reduce the interval to a smaller one

vfdev-5:
@KickItLikeShika I'm running the notebook on the colab and I think we do not need to call `plt.axes(ax)` : 

```python
ax = lr_finder.plot()
plt.axes(ax)
plt.show()

print("Suggested LR", lr_finder.lr_suggestion())
```

examples/notebooks/FastaiLRFinder_MNIST.ipynb
```diff
@@ -251,11 +308,10 @@
    },
    "outputs": [],
    "source": [
-    "optimizer.param_groups[0]['lr'] = lr_finder.lr_suggestion()\n",
+    "new_trainer.run(trainloader, max_epochs=10)\n",
```
vfdev-5, line 311: I think it is better to reduce `max_epochs` to 5 epochs everywhere. In case of suggested LR it is rather large and on during the last 5 epochs batch loss jumps too much as LR is rather high.

examples/notebooks/FastaiLRFinder_MNIST.ipynb
```diff
@@ -218,18 +218,53 @@
...
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
+   "cell_type": "markdown",
    "metadata": {
     "id": "EHHNoOi8mvd6"
    },
+   "source": [
+    "Let's compare training results: \n",
```
vfdev-5, line 232: I think to convey a better message on how to use the LR finder we have to change the order of below runs.
The idea is to show how we can use LR finder in user code without re-creating model and optimizer again.
So, let's continue the training with suggested LR and create new model / optimizer with initial params.

By the way, the line "Note, that model and optimizer were restored to their initial states when FastaiLRFinder finished." should be put here and not below.

KickItLikeShika: > So, let's continue the training with suggested LR and create new model / optimizer with initial params.

Can you please explain this more?

vfdev-5: Think about how would you use this feature in your project:
```python
model = ...
optimizer = ...
trainer = ...

# do lr finder trick
# ...

lr_finder.apply_suggested_lr(optimizer)

trainer.run(...)
```

We do not want to create another model etc and this should be shown here as an example.

KickItLikeShika: I tried this approach already, but I found nothing on how to reset the parameters I found this discussion https://discuss.pytorch.org/t/reset-optimizer-stats/78516/5?u=kickitlikeshika , https://github.com/pytorch/pytorch/issues/37410

vfdev-5: Why would you like to reset parameters in this case ?

KickItLikeShika: Should we continue training with same old parameters? that will be easier to implement, but I though we needed to show the user that a new model/optimizer with found LR will perform way better, that's why I wanted to start from scratch again..

vfdev-5: Let me explain again my point. The notebook is nice now and almost OK, but it does not show the user how to use the feature in their project as we are creating new models/optim etc. In the doc I suggest to write it like that : 
```python
model = ...
optimizer = ...

init_model_state = model.state_dict()
init_opt_state = optimizer.state_dict()

# ----

ProgressBar(persist=True).attach(trainer, output_transform=lambda x: {"batch loss": x})

lr_finder = FastaiLRFinder()
to_save={'model': model, 'optimizer': optimizer}
with lr_finder.attach(trainer, to_save, diverge_th=1.5) as trainer_with_lr_finder:
    trainer_with_lr_finder.run(trainloader)

# ----

ax = lr_finder.plot()
plt.show()

print("Suggested LR", lr_finder.lr_suggestion())

# ----

Let's compare training results:

    a) with applying optimal learning rate.
    b) using a fixed learning rate (lr=3e-4)

# ----

lr_finder.apply_suggested_lr(optimizer)
print(optimizer.param_groups[0]['lr'])

# ----

trainer.run(trainloader, max_epochs=10)

evaluator = create_supervised_evaluator(model, metrics={"acc": Accuracy(), "loss": Loss(nn.NLLLoss())}, device=device)
evaluator.run(testloader)

print(evaluator.state.metrics)

# ----

# Reinit model / optim
model.load_state_dict(init_model_state)
optimizer.load_state_dict(init_opt_state)

# ----

trainer.run(trainloader, max_epochs=10)

evaluator = create_supervised_evaluator(model, metrics={"acc": Accuracy(), "loss": Loss(nn.NLLLoss())}, device=device)
evaluator.run(testloader)

print(evaluator.state.metrics)

```

vfdev-5 (approved):
LGTM, thanks for the PR @KickItLikeShika !

PR #1988, opened 2021-05-05
Move param handler

### LINKED ISSUE #1912: Move param handlers to core part
## 🚀 Feature

As discussed in https://github.com/Project-MONAI/MONAI/pull/1949#issuecomment-815283287 moving https://github.com/pytorch/ignite/blob/master/ignite/contrib/handlers/param_scheduler.py to the core part: https://github.com/pytorch/ignite/tree/master/ignite/handlers could be great addition for users, apps and toolkits which do not use contrib module (e.g. MONAI).

While moving the code, we would like to ensure BC-compatibilty for at least one future release. Once the code has been moved we can put deprecation message for classes and methods inside contrib module. `contrib/handlers/param_scheduler.py` will be removed in v0.6.0.

A follow-up feature to this feature request could be enlarging the scope of `ParamScheduler` to scheduling any parameter (e.g. hyperparameters) like it is done in MONAI: https://github.com/Project-MONAI/MONAI/pull/1949

cc @danieltudosiu

fco-dv:
Fixes #1912 

Description: Move param_scheduler handlers to core part, ensuring BC until v6.0 in the contrib sub-module.

Check list:


ignite/contrib/handlers/param_scheduler.py
```diff
@@ -13,6 +13,7 @@
 from torch.optim.optimizer import Optimizer
 
 from ignite.engine import Engine
+from ignite.utils import deprecated
```
sdesrozis, line 16: @fco-dv why not remove all the content of this file, import all from `ignite/handlers/param_scheduler.py` and add a depreciation warning ?

fco-dv: thanks @sdesrozis  yes it would be much better

vfdev-5:
@fco-dv how about doing things like for ClearML <-> Trains logger: https://github.com/pytorch/ignite/blob/master/ignite/contrib/handlers/trains_logger.py

sdesrozis (approved):
@fco-dv Thank you ! LGTM !

vfdev-5 (approved):
Thanks @fco-dv !

PR #1990, opened 2021-05-06
[skip ci] Updated nightly to latest stable pytorch-xla in teaser notebook

Priyansi:
Description:
With the nightly version of xla, the notebook produced the following error during installation - `OSError: libmkl_intel_lp64.so.1: cannot open shared object file: No such file or directory.`
The workarounds are too complicated for a simple teaser notebook. Therefore, changing the unstable nightly version to the stable one looked like the best option.

Check list:

### DIFF
--- a/assets/tldr/teaser.ipynb
+++ b/assets/tldr/teaser.ipynb
@@ -48,9 +48,7 @@
     "with_torch_launch = \"WORLD_SIZE\" in os.environ\n",
     "\n",
     "if in_colab:\n",
-    "    VERSION = \"nightly\"\n",
-    "    !curl https://raw.githubusercontent.com/pytorch/xla/master/contrib/scripts/env-setup.py -o pytorch-xla-env-setup.py\n",
-    "    !python pytorch-xla-env-setup.py --version $VERSION"
+    "    !pip install cloud-tpu-client==0.10 https://storage.googleapis.com/tpu-pytorch/wheels/torch_xla-1.8.1-cp37-cp37m-linux_x86_64.whl"
    ]
   },
   {
--- a/assets/tldr/teaser.ipynb
+++ b/assets/tldr/teaser.ipynb
@@ -48,8 +48,8 @@
     "with_torch_launch = \"WORLD_SIZE\" in os.environ\n",
     "\n",
     "if in_colab:\n",
-    "    VERSION = !curl -s https://api.github.com/repos/pytorch/xla/releases/latest | awk -F '\"' '/tag_name/{print $4}' | cut -c 2-\n",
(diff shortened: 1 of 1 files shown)


assets/tldr/teaser.ipynb
```diff
@@ -48,9 +48,7 @@
     "with_torch_launch = \"WORLD_SIZE\" in os.environ\n",
     "\n",
     "if in_colab:\n",
-    "    VERSION = \"nightly\"\n",
-    "    !curl https://raw.githubusercontent.com/pytorch/xla/master/contrib/scripts/env-setup.py -o pytorch-xla-env-setup.py\n",
-    "    !python pytorch-xla-env-setup.py --version $VERSION"
+    "    !pip install cloud-tpu-client==0.10 https://storage.googleapis.com/tpu-pytorch/wheels/torch_xla-1.8.1-cp37-cp37m-linux_x86_64.whl"
```
vfdev-5, line 51: Can't we install just latest stable instead of pinning the version to 1.8.1 ?

vfdev-5 (commented):
Thanks for the PR @Priyansi 
I left a comment

assets/tldr/teaser.ipynb
vfdev-5: Second option is OK for me, thanks 👍

Priyansi: Done! Is it alright now?

vfdev-5: @Priyansi have you tested it on Colab ?

Looks like it does not work : https://colab.research.google.com/github/pytorch/ignite/blob/3dce762c089f785eb103f6e2cd5e2961673950f6/assets/tldr/teaser.ipynb#scrollTo=Y6x4WuLQ3j2a

Priyansi: Oh my bad. I accidentally put a `$` before VERSION. Fixing it ASAP and running the entire notebook again.

vfdev-5: Btw, I think there could be a more simple/shorter way to fetch tag_name from the curl output...

Priyansi: If we are looking for a shorter regex, this would be the best option, according to me - `VERSION = !curl -s https://api.github.com/repos/pytorch/xla/releases/latest | grep -Po '"tag_name": "v\K.*?(?=")'`

Priyansi: Yes, I was thinking that too but it requires extra overhead that `grep` could easily solve. I made the above change. It should work now.

vfdev-5 (approved):
Thanks @Priyansi , LGTM

PR #1982, opened 2021-05-01
move EpochOutputStore out of contrib keeping BC

radekosmulski:
Fixes #1939

Description: Moves `EpochOutputStore` out of `contrib` keeping BC.

Check list:

- [ ] New tests are added (if a new feature is added)
- [ ] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #1993, opened 2021-05-08
chore: version change in EOS core handler [skip ci]

ydcjeff:
### Description

`EpochOutputStore` has been moved from contrib to core in #1982.
But the docstring said that this is new in v0.4.2 in core while it is actually moved in v0.5.0.

Check list:


vfdev-5 (approved):
LGTM! Thanks @ydcjeff

PR #1995, opened 2021-05-12
Issue 1986 Checkpointing write & remove order change

### LINKED ISSUE #1986: ModelCheckpoint improvement from "remove & write" to "write & remove"
Lately I experienced an issue with model checkpointing, so I wanted to move it to a discussion although I am unsure about this is a "bug", and thus I opened as "question". To sum it up, when using model checkpointing with certain a certain configuration, where 
`n_saved=1`, there is a potential risk to lose the checkpoint due to "remove first, and then write" logic. 

**Problem**
You can create a model checkpoint at specific timestamps to save a checkpoint. However, there is a potential risk that you can lose the checkpoint due to write errors, interruption, machine shut-down, or outage etc. DiskSaver used on ModelCheckpoint first checks if checkpoint count doesn't exceed `n_saved`, and after that it removes old/older checkpoints, and then writes new checkpoint. When `n_saved=1` this turns it into basic "remove existing & write", and if by any case write process is corrupted, then since the older checkpoint is deleted, you simply waste your training resources and time. Option `n_saved > 1` can create many dangling and redundant model checkpoints, and with large number of experiments and experimenting with huge models especially on cloud causes unnecessary files claiming large amount of storage space.

**Ideas**
The trivial idea is to set `n_saved > 1`, but this has some negative consequences that people may not want generally and avoid it. The second idea is to replace "remove first and then write" logic to "write first and then remove" logic. Is there any possible practical ideas ? 

NOTE: This issue is opened to discuss the situation, not meant to imply any feature request.

devrimcavusoglu:
Fixes #1986 

Description: Former "remove & write" implementation order is reversed to "write & remove", api call has been added for ClearMLSaver remove functionality.

Check list:


vfdev-5:
@devrimcavusoglu thanks for the PR, i'll review it asap

ignite/contrib/handlers/clearml_logger.py
```diff
@@ -806,7 +806,11 @@ def get_local_copy(self, filename: str) -> Optional[str]:
 
     @idist.one_rank_only()
     def remove(self, filename: str) -> None:
-        super(ClearMLSaver, self).remove(filename)
+        from clearml.storage.helper import StorageHelper
+
+        clearml_logger = self._task.get_logger()
+        helper = StorageHelper.get(filename, logger=clearml_logger)
+        helper.delete(filename)
```
vfdev-5, line 813: @devrimcavusoglu could you please explain why you are adding this code ?
I understand that the idea is to use clearml's v1.0 new feature and make `remove` work as expected for others.
I'm not sure how it would work with the previous logic where a queue is used...

cc @bmartinn any suggestion on how we could recode everything here ?

ignite/handlers/checkpoint.py
```diff
@@ -630,14 +626,18 @@ class DiskSaver(BaseSaveHandler):
         dirname: Directory path where the checkpoint will be saved
         atomic: if True, checkpoint is serialized to a temporary file, and then
             moved to final destination, so that files are guaranteed to not be damaged
-            (for example if exception occurs during saving).
+            (for example if exception occurs during saving). Setting ``atomic=True`` is
+            recommended if ``n_saved=1`` is set in checkpoint object. For details, see
+            issue `#1986`_
```
vfdev-5, line 631: I think we can remove the link to the issue here. We can create a paragraph below describing essential points from the issue and add "see notes below" or something like that here.

ignite/contrib/handlers/neptune_logger.py
```diff
@@ -78,7 +78,7 @@ class NeptuneLogger(BaseLogger):
                 project_name="shared/pytorch-ignite-integration",
                 experiment_name="cnn-mnist", # Optional,
                 params={"max_epochs": 10}, # Optional,
-                tags=["pytorch-ignite","minst"] # Optional
+                tags=["pytorch-ignite","mnist"] # Optional
```
vfdev-5, line 81: Could you please send a separate PR for this typo. It will be merged more quickly and wont require discussions and multiple iterations as this PR... Thanks !

ignite/contrib/handlers/clearml_logger.py
devrimcavusoglu: As you said, this would make use of clearml v1.0 new feature, and used to actually remove the remote or local file(s). For backward compatibility, I'd appreciate any suggestions.

bmartinn: No need to pass `logger` to `StorageHelper.get` , basically this should be fine:
```python
helper = StorageHelper.get(filename)
helper.delete(filename)
```

devrimcavusoglu: Thanks for the review & comment @bmartinn. I'll update PR with that change asap. I and @vfdev-5 were mainly concerned about backward compatibility as it may negatively impact users using `clearml<1.0`; however, I see that it is not a concern.

bmartinn: @devrimcavusoglu I rechecked backwards compatibility,  the only exception will be uploading models directly to `clearml-server` < 1.0. If this is the case the delete function might return an error as the http file server did not support the delete operation. I really think that just protecting with `except ValueError` should be enough. wdyt?
```python
try:
  helper.delete(filename)
except ValueError:
  # log something ?!
```

ignite/handlers/checkpoint.py
devrimcavusoglu: done

ignite/contrib/handlers/clearml_logger.py
devrimcavusoglu: @bmartinn Thanks for reviewing. Well, yes it should be sufficient I believe. As we protect the older implementation `super().remove()` inside `StorageHelper`. 

With the suggested change, all things should be okay. I'll add a warning (for `clearml-server<1.0` users) to point out for this.

vfdev-5: @bmartinn can you please confirm whether we still need `_CallbacksContext` and `_checkpoint_slots` code with clearml v1.X ?

I have an impression that if we could remove data from remote storage then we can drastically simplify `ClearMLSaver` for v1.X ?

devrimcavusoglu: @vfdev-5 Hi there, sorry for delayed reply (was in tight schedule). Option 2 seems more solid to me, and also more practical. Regarding the class seperation (former comment) for different versions, I think rather than pointing to V0 and V1, handling it under `ClearMLSaver` would be more concrete, wdyt ?

vfdev-5: > Regarding the class seperation (former comment) for different versions, I think rather than pointing to V0 and V1, handling it under ClearMLSaver would be more concrete, wdyt ?

I was thinking that doing things into two separate classes will help us later to quickly remove V0 when no one is using it anymore...

jkhenning: Hi @vfdev-5,

> @bmartinn can you please confirm whether we still need _CallbacksContext and _checkpoint_slots code with clearml v1.X ?

I think that if you handle storing a new checkpoint to a temporary-named newly uploaded, file, removing the previously uploaded file and renaming the temporary-named file to the previous name, you won't need the checkpoints 🙂

PR #1996, opened 2021-05-13
Improve LRFinder for DDP

### LINKED ISSUE #837: LRFinder in DDP
## 🚀 Feature

We need to check FastaiLRFinder's usage in DDP and probably imrove the docs, provide complete example of on its usage.

KickItLikeShika:
Fixes #837 

Improve LRFinder for DDP and add tests.

Check list:

### DIFF
--- a/ignite/contrib/handlers/lr_finder.py
+++ b/ignite/contrib/handlers/lr_finder.py
@@ -137,8 +137,7 @@ class FastaiLRFinder:
     def _log_lr_and_loss(self, trainer: Engine, output_transform: Callable, smooth_f: float, diverge_th: float) -> None:
         output = trainer.state.output
         loss = output_transform(output)
-        if idist.get_world_size() > 1:
-            loss = idist.all_reduce(loss)
+        loss = idist.all_reduce(loss)
         lr = self._lr_schedule.get_param()  # type: ignore[union-attr]
         self._history["lr"].append(lr)
         if trainer.state.iteration == 1:
--- a/ignite/contrib/handlers/lr_finder.py
+++ b/ignite/contrib/handlers/lr_finder.py
@@ -10,6 +10,7 @@ import torch
 from torch.optim import Optimizer
 from torch.optim.lr_scheduler import _LRScheduler
 
+import ignite.distributed as idist
 from ignite.contrib.handlers.param_scheduler import LRScheduler, PiecewiseLinear
 from ignite.engine import Engine, Events
 from ignite.handlers import Checkpoint
@@ -136,6 +137,8 @@ class FastaiLRFinder:
     def _log_lr_and_loss(self, trainer: Engine, output_transform: Callable, smooth_f: float, diverge_th: float) -> None:
         output = trainer.state.output
         loss = output_transform(output)
+        if idist.get_world_size() > 1:
+            loss = idist.all_reduce(loss)
         lr = self._lr_schedule.get_param()  # type: ignore[union-attr]
         self._history["lr"].append(lr)
         if trainer.state.iteration == 1:
--- a/tests/ignite/contrib/handlers/test_lr_finder.py
+++ b/tests/ignite/contrib/handlers/test_lr_finder.py
@@ -1,5 +1,6 @@
 import copy
 import os
+from unittest.mock import MagicMock
 
 import matplotlib
 import pytest
@@ -469,19 +470,7 @@ def test_plot_multiple_param_groups(
 def _test_distrib_log_lr_and_loss(device):
     from ignite.handlers import PiecewiseLinear
 
-    class DummyModel(nn.Module):
-        def __init__(self, n_channels=10, out_channels=10, flatten_input=False):
-            super(DummyModel, self).__init__()
-
-            self.net = nn.Sequential(
-                nn.Flatten() if flatten_input else nn.Identity(), nn.Linear(n_channels, out_channels)
-            )
-
-        def forward(self, x):
-            return self.net(x)
-
-    model = DummyModel()
-    optimizer = SGD(model.parameters(), lr=1e-4, momentum=0.0)
+    optimizer = MagicMock(spec=torch.optim.Optimizer)
     lr_finder = FastaiLRFinder()
 
     # minimal setup for lr_finder to make _log_lr_and_loss work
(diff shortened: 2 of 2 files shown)


KickItLikeShika:
I know the PR isn't perfect yet as the the tests were a bit confusing for me, but I need to to see the reviews, and Thanks for help!

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -462,4 +463,130 @@ def test_plot_multiple_param_groups(
...
+
+    model = DummyModel()
+    optimizer = SGD(model.parameters(), lr=1e-4, momentum=0.0)
+    lr_finder = FastaiLRFinder()
+
+    # minimal setup for lr_finder to make _log_lr_and_loss work
+    rank = idist.get_rank()
+    loss = 0.01 * (rank + 1)
+
+    _lr_schedule = PiecewiseLinear(optimizer, param_name="lr", milestones_values=[(0, 1e-4), (1, 10)])
```
vfdev-5, line 491: Let's see if we could use `unittest.mock` instead ? Such that we could remove model and optimizer from this test.

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -462,4 +463,130 @@ def test_plot_multiple_param_groups(
...
+    engine.state.output = loss
+    engine.state.iteration = 1
+    lr_finder._lr_schedule = _lr_schedule
+    lr_finder._history["loss"] = []
+    lr_finder._history["lr"] = []
+
+    lr_finder._log_lr_and_loss(engine, output_transform=lambda x: x, smooth_f=0.1, diverge_th=10.0)
+    # expected_loss = loss
+    # for i in range(rank - 1):
+    #     expected_loss += 0.01 * (i + 1)
```
vfdev-5, line 504: Can we remove that ?

vfdev-5 (commented):
Yes, let's polish it a bit.

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -462,4 +463,130 @@ def test_plot_multiple_param_groups(
...
+    to_save = {"model": model, "optimizer": optimizer}
+    engine = create_supervised_trainer(model, optimizer, nn.CrossEntropyLoss())
+    lr_finder = FastaiLRFinder()
+    with lr_finder.attach(engine, to_save) as trainer_with_finder:
+        trainer_with_finder.run(train_loader)
+
+    lr_finder.plot()
+    ax = lr_finder.plot(skip_end=0)
+    ax.figure.savefig("distrib_dummy.jpg")
+    assert os.path.exists("distrib_dummy.jpg")
```
vfdev-5, line 544: If each process is saving the figure, it will produce something bad...

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -462,4 +463,130 @@ def test_plot_multiple_param_groups(
...
+
+    assert 1e-4 <= lr_finder.lr_suggestion() <= 2
```
vfdev-5, line 546: Let's reduce the interval of expected values and let's add `lr_finder.apply_suggested_lr` to check

tests/ignite/contrib/handlers/test_lr_finder.py
KickItLikeShika: I don't mean what do you mean with something bad, but the figure looks fine
distrib_dummy

vfdev-5: @KickItLikeShika in distributed it is better to have only one process to write on disk to the same place.

KickItLikeShika: Thanks for pointing that out! I will dig more to know to handle that!

tests/ignite/contrib/handlers/test_lr_finder.py
vfdev-5: How about mocking `_lr_schedule` ?

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -462,4 +464,122 @@ def test_plot_multiple_param_groups(
...
+
+        def forward(self, x):
+            return self.net(x)
+
+    model = DummyModel(n_channels=784, out_channels=10, flatten_input=True)
+    model = model.to(device)
+
+    optimizer = SGD(model.parameters(), lr=1e-4, momentum=0.0)
+    to_save = {"model": model, "optimizer": optimizer}
+    engine = create_supervised_trainer(model, optimizer, nn.CrossEntropyLoss())
```
vfdev-5, line 523: I think here `device` arg is missing otherwise device mismatch. Reported here I think : https://github.com/pytorch/ignite/runs/2576224816

ignite/contrib/handlers/lr_finder.py
```diff
@@ -136,6 +137,8 @@ def _reset(self, trainer: Engine) -> None:
     def _log_lr_and_loss(self, trainer: Engine, output_transform: Callable, smooth_f: float, diverge_th: float) -> None:
         output = trainer.state.output
         loss = output_transform(output)
```
sdesrozis, line 139: @vfdev-5 Just a question, do we really need this test ? I mean, if sequential, all_reduce should do nothing.

ignite/contrib/handlers/lr_finder.py
```diff
@@ -136,6 +137,8 @@ def _reset(self, trainer: Engine) -> None:
...
+        if idist.get_world_size() > 1:
```
sdesrozis, line 140: @vfdev-5 Just a question, do we really need this test ? I mean, if sequential, all_reduce should do nothing.

vfdev-5: I agree, let's remove it

KickItLikeShika: @sdesrozis Thanks for the remark!

vfdev-5 (approved):
LGTM, thanks a lot @KickItLikeShika !

PR #1998, opened 2021-05-14
Move LRFinder to core

KickItLikeShika:
Moving LRFinder to core.

Check list:


sdesrozis:
@KickItLikeShika Thank you for the PR.

Have a look to `param_scheduler.py` which was recently moved from `contrib` to `handlers`.

https://github.com/pytorch/ignite/blob/ee2c7eeb8d07b56366868054b513f14e868f4d2e/ignite/contrib/handlers/param_scheduler.py

You need to first use the command `git copy` to duplicate the file and keep the history of commits. Secondly you have to depreciate the file using the same strategy than in `param_scheduler.py`. 

Could you update the PR following these comments ? Thank you !

sdesrozis (approved):
@KickItLikeShika Thanks a lot ! LGTM !

PR #2016, opened 2021-05-31
Enabled gpu for gloo [test on gpus]

vfdev-5:
Description:
- Enabled gpu for gloo

Check list:


sdesrozis (approved):
Use `torch.cuda.is_available()` and `idist.device()` rather than `torch.device("xxx")` is very nice !

LGTM.

PR #2017, opened 2021-05-31
Fix Precision compute for all zero predictions

### LINKED ISSUE #2007: Precision raises an NonComputableError for all zeros model predictions
## 🐛 Bug description

Following https://github.com/pytorch/ignite/issues/1991#issuecomment-844860265

```python
p = Precision()
p.update((torch.zeros(4), torch.randint(0, 2, (4,))))
p.compute()
> NotComputableError: Precision must have at least one example before it can be computed.
```
which is wrong as we do call `update`.

This is related to the check
```
/opt/conda/lib/python3.8/site-packages/ignite/metrics/precision.py in compute(self)
     51         is_scalar = not isinstance(self._positives, torch.Tensor) or self._positives.ndim == 0
     52         if is_scalar and self._positives == 0:
---> 53             raise NotComputableError(
     54                 f"{self.__class__.__name__} must have at least one example before it can be computed."
     55             )
```
where `self._positives` is 0 for all zeros predictions but is thought as uninitialized.

I think this should be related to a change in pytorch that previously had `torch.tensor([0, 0, 0]).sum()` as 1d tensor and now is 0d tensor.

## Environment

- PyTorch Version (e.g., 1.4): 1.8
- Ignite Version (e.g., 0.3.0): 0.4.4
- OS (e.g., Linux):
- Python version:

cc @liebkne

trsvchn:
Fixes #2007 

Description:

- Introduced `_updated` flag for update checks
- Recall is adjusted
- Precision and Recall Tests are updated

Check list:

### DIFF
--- a/ignite/metrics/precision.py
+++ b/ignite/metrics/precision.py
@@ -22,6 +22,7 @@ class _BasePrecisionRecall(_BaseClassification):
 
         self._average = average
         self.eps = 1e-20
+        self._updated = False
         super(_BasePrecisionRecall, self).__init__(
             output_transform=output_transform, is_multilabel=is_multilabel, device=device
         )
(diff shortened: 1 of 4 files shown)


vfdev-5:
@trsvchn thanks for the PR, it looks good ! Can you please add the issue case with all zeros predictions ?

trsvchn:
Oh, right! sure!

vfdev-5 (approved):
Thanks @trsvchn

PR #2019, opened 2021-05-31
Implemented Engine.state.data

### LINKED ISSUE #1923: Introduce state.data as end-to-end data structure
## 🚀 Feature

Following the discussions from https://github.com/Project-MONAI/MONAI/issues/1987 , we may think of providing a unified data structure where we could merge input batch and output to simplify data access inside user handlers.

@Nic-Ma could you please detail the pitch of the idea. Thanks a lot !

vfdev-5:
Fixes #1923

Description:
- Implemented Engine.state.data for convenient data access

Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -6,7 +6,7 @@ import warnings
 import weakref
 from collections import OrderedDict, defaultdict
 from collections.abc import Mapping
-from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Sequence, Tuple, Union
 
 from torch.utils.data import DataLoader
 
@@ -808,11 +808,13 @@ class Engine(Serializable):
 
             while True:
                 self.state.batch = self.state.output = None
+                self.state.data = {}
                 try:
                     # Avoid Events.GET_BATCH_STARTED triggered twice when data iter is restarted
                     if self.last_event_name != Events.DATALOADER_STOP_ITERATION:
                         self._fire_event(Events.GET_BATCH_STARTED)
                     self.state.batch = next(self._dataloader_iter)
+                    _add_to_state_data(self.state.data, self.state.batch, "input")
                     self._fire_event(Events.GET_BATCH_COMPLETED)
                     iter_counter += 1
                     should_exit = False
@@ -851,6 +853,7 @@ class Engine(Serializable):
                 self.state.iteration += 1
                 self._fire_event(Events.ITERATION_STARTED)
                 self.state.output = self._process_function(self, self.state.batch)
+                _add_to_state_data(self.state.data, self.state.output, "output")
                 self._fire_event(Events.ITERATION_COMPLETED)
 
                 if self.should_terminate or self.should_terminate_single_epoch:
@@ -871,3 +874,12 @@ class Engine(Serializable):
             self._handle_exception(e)
 
         return time.time() - start_time
+
+
+def _add_to_state_data(state_data: Dict[str, Any], data: Union[Dict[str, Any], Sequence, Any], name: str):
+    if isinstance(data, Sequence):
+        state_data.update({f"{name}_{i}": value for i, value in enumerate(data)})
+    elif isinstance(data, Mapping):
+        state_data.update(data)
+    else:
+        state_data[name] = data
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -876,7 +876,7 @@ class Engine(Serializable):
         return time.time() - start_time
 
 
-def _add_to_state_data(state_data: Dict[str, Any], data: Union[Dict[str, Any], Sequence, Any], name: str):
+def _add_to_state_data(state_data: Dict[str, Any], data: Union[Dict[str, Any], Sequence, Any], name: str) -> None:
     if isinstance(data, Sequence):
         state_data.update({f"{name}_{i}": value for i, value in enumerate(data)})
     elif isinstance(data, Mapping):
--- a/ignite/engine/events.py
+++ b/ignite/engine/events.py
@@ -366,9 +366,39 @@ class State:
         state.metrics           # dictionary with defined metrics if any
         state.times             # dictionary with total and per-epoch times fetched on
                                 # keys: Events.EPOCH_COMPLETED.name and Events.COMPLETED.name
+        state.data              # dictionary with unrolled batch and output. See below for details
+
+    Note:
+        ``state.data`` is a dictionary and contains ``state.batch`` and ``state.output`` data.
+
+        .. code-block:: python
+
+            # if batch is a sequence
+            # if output is not a sequence or dictionary
+            state.data = {
+                "input_0": batch[0],
+                "input_1": batch[1],
+                ...
+                "output": output
+            }
+
+            # if batch is a dictionary: {"image": x, "target": y, "meta": metadata, ...}
+            # if output is a dictionary: {"y_pred": y_pred, "batch_loss": batch_loss, ...}
+            state.data = {
+                "image": x,
+                "target": y,
+                "meta": metadata,
+                ...
+                "y_pred": y_pred,
+                "batch_loss": batch_loss,
+                ...
+            }
 
     Args:
         kwargs: keyword arguments to be defined as State attributes.
+
+    .. versionchanged:: 0.4.5
+        Added ``state.data``.
     """
 
     event_to_attr = {
@@ -397,6 +427,7 @@ class State:
             Events.EPOCH_COMPLETED.name: None,
             Events.COMPLETED.name: None,
         }  # type: Dict[str, Optional[float]]
+        self.data = {}  # type: Dict[str, Any]
 
         for k, v in kwargs.items():
             setattr(self, k, v)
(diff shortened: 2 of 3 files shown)


vfdev-5:
Hi @Nic-Ma , here is a PR implementing `engine.state.data` as discussed in #1923. Let me know what do you think ?

Nic-Ma:
Hi @vfdev-5 ,

Thanks for your quick update, the PR looks good to me.
Just want to double confirm that: when this PR is merged, I will update our own iteration() of engines:
```py
    def _iteration(self, engine: Engine, data: Dict[str, torch.Tensor]):
        inputs, targets = self.prepare_batch(data, engine.state.device, engine.non_blocking)
        engine.state.data.update({Keys.IMAGE: inputs, Keys.LABEL: targets})

        self.network.train()
        self.optimizer.zero_grad()
        engine.state.data[Keys.PRED] = self.inferer(inputs, self.network, *args, **kwargs)
        engine.fire_event(IterationEvents.FORWARD_COMPLETED)

        engine.state.data[Keys.LOSS] = self.loss_function(engine.state.data[Keys.PRED], targets).mean()
        engine.fire_event(IterationEvents.LOSS_COMPLETED)

        engine.state.data[Keys.LOSS].backward()
        engine.fire_event(IterationEvents.BACKWARD_COMPLETED)

        self.optimizer.step()
        engine.fire_event(IterationEvents.MODEL_COMPLETED)
        # will execute post transforms on `engine.state.data` when MODEL_COMPLETED

        return engine.state.data
```
So we will only use `engine.state.data` in iteration() or handlers, is there any potential issue?
@wyli, in my above proposal, the returned `engine.state.data` will be `engine.state.output`, do you agree?

Thanks.

ignite/engine/engine.py
```diff
@@ -808,11 +808,13 @@ def _run_once_on_dataset(self) -> float:
 
             while True:
                 self.state.batch = self.state.output = None
+                self.state.data = {}
                 try:
                     # Avoid Events.GET_BATCH_STARTED triggered twice when data iter is restarted
                     if self.last_event_name != Events.DATALOADER_STOP_ITERATION:
                         self._fire_event(Events.GET_BATCH_STARTED)
                     self.state.batch = next(self._dataloader_iter)
+                    _add_to_state_data(self.state.data, self.state.batch, "input")
```
sdesrozis, line 817: Why not “batch” rather than “input” ?

ignite/engine/events.py
```diff
@@ -366,9 +366,39 @@ class State:
...
+
+    Note:
+        ``state.data`` is a dictionary and contains ``state.batch`` and ``state.output`` data.
+
+        .. code-block:: python
+
+            # if batch is a sequence
+            # if output is not a sequence or dictionary
+            state.data = {
+                "input_0": batch[0],
```
sdesrozis, line 379: Here batch is accumulated during epoch, right?

ignite/engine/engine.py
```diff
@@ -871,3 +874,12 @@ def _run_once_on_dataset(self) -> float:
             self._handle_exception(e)
 
         return time.time() - start_time
+
+
+def _add_to_state_data(state_data: Dict[str, Any], data: Union[Dict[str, Any], Sequence, Any], name: str):
```
sdesrozis, line 879:
```suggestion
def _add_to_state_data(state_data: Dict[str, Any], data: Union[Dict[str, Any], Sequence, Any], name: str) -> None:
```

ignite/engine/engine.py
vfdev-5: This name is used if state.batch is list/tuple or non-dict and I found that "input_0", "input_1" etc is sort of better than
"batch_0", "batch_1" etc...

vfdev-5:
@Nic-Ma I think the way you are using `engine.state.data` is a bit different of how it is coded in this PR.
If used like that, then it is more simple to instantiate `engine.state.data` as an empty dict on `ITERATION_STARTED` and fill it as in `_iteration` without this PR.

With this PR, `_iteration` will be something like
```python

# before _iteration method, engine.state.data will contain data: Dict[str, torch.Tensor]
# as engine.state.data = {
#   k, v for k, v in data.items()
# }
 
    def _iteration(self, engine: Engine, data: Dict[str, torch.Tensor]):
        inputs, targets = self.prepare_batch(data, engine.state.device, engine.non_blocking)

        self.network.train()
        self.optimizer.zero_grad()
        engine.state.data[Keys.PRED] = self.inferer(inputs, self.network, *args, **kwargs)
        engine.fire_event(IterationEvents.FORWARD_COMPLETED)

        engine.state.data[Keys.LOSS] = self.loss_function(engine.state.data[Keys.PRED], targets).mean()
        engine.fire_event(IterationEvents.LOSS_COMPLETED)

        engine.state.data[Keys.LOSS].backward()
        engine.fire_event(IterationEvents.BACKWARD_COMPLETED)

        self.optimizer.step()
        engine.fire_event(IterationEvents.MODEL_COMPLETED)
        # will execute post transforms on `engine.state.data` when MODEL_COMPLETED

        return {}  # returns nothing as everything is already appended to data -> engine.state.output == {}

# after _iteration method, engine.state.data add the output of _iteration
# engine.state.data = {
#    k, v for k, v in data.items(),
#    Keys.PRED: ...
#    Keys.LOSS: ...
# }
```

sdesrozis (approved):
@vfdev-5 LGTM !

Nic-Ma:
Hi @vfdev-5 ,

I see your example code, can we return nothing as we only use `engine.state.data`? Something like:
```py
    def _iteration(self, engine: Engine, data: Dict[str, torch.Tensor]):
        inputs, targets = self.prepare_batch(data, engine.state.device, engine.non_blocking)
        # update `image` and `label` as they are moved to GPU tensor after `prepare_batch`
        engine.state.data.update({Keys.IMAGE: inputs, Keys.LABEL: targets})

        self.network.train()
        self.optimizer.zero_grad()
        engine.state.data[Keys.PRED] = self.inferer(inputs, self.network, *args, **kwargs)
        engine.fire_event(IterationEvents.FORWARD_COMPLETED)

        engine.state.data[Keys.LOSS] = self.loss_function(engine.state.data[Keys.PRED], targets).mean()
        engine.fire_event(IterationEvents.LOSS_COMPLETED)

        engine.state.data[Keys.LOSS].backward()
        engine.fire_event(IterationEvents.BACKWARD_COMPLETED)

        self.optimizer.step()
        engine.fire_event(IterationEvents.MODEL_COMPLETED)
        # will execute post transforms on `engine.state.data` when MODEL_COMPLETED
        # don't return
```

Thanks.

vfdev-5:
> I see your example code, can we return nothing as we only use `engine.state.data` ?

Yes, it is possible. However, it will still create `engine.state.data["output"] == None`. Maybe, we can filter out Nones. 
Another thought about returning None is that previously we had some user feature requests to skip an iteration and we were thinking to do that if output is None. It is not yet implemented but we may implement such logic of skipping an iteration in future. Maybe, we'll have to decide to skip it with some other specific value...

Nic-Ma:
Hi @vfdev-5 ,

Sounds good.
So I think we can accept to potentially create `engine.state.data["output"] == None` if return nothing, I will try to go this way when updated to ignite v0.4.5 later.

Thanks.

Nic-Ma (approved):

wyli:
thanks, just for clarification, now this PR implements `engine.state.data` to store almost the same piece of information as `batch`+`output`. so how to ensure the data consistency between these structures? can a user assume that `state.data` is always a shallow copy of `state.batch`+`state.output`, or when updating one of them, the user should take care of updating the other?

vfdev-5:
@wyli thanks for the comment. It's very good point to think about. Yes, that's true that currently they are unsynchronized in this PR. I was thinking about `state.data` as a view of batch and output -> if user change batch or output directly as `state.batch = another_batch` we need to make the change into data using batch as a property, for example.

I put this PR as a draft as I have an impression that what we need is `state.data` as a source and `batch` and `output` as views of the data...

Nic-Ma:
Hi @wyli @vfdev-5 ,

Another problem is that: if we **decollate** `engine.state.data` after the iteration, `engine.state.data` will change from a `dict of batch-first Tensor` to a `list of dict of channel-first Tensor`, which is against the initial definition.
So maybe we need to somehow define `engine.state.data` as unified `dataflow` but not limited to be a `dict`?

What do you think?

Thanks in advance.

Nic-Ma:
Hi @vfdev-5 ,

May I know your plan or proposal for this ticket?

Thanks.

vfdev-5:
Hi @Nic-Ma I think we need to think more carefully how to implement that and unfortunately I think this wont be part of the next release.

Nic-Ma:
Hi @vfdev-5 ,

Thanks for your updates, sure, I totally agree to think carefully about this overall feature.

ryanwongsa:
Hi @vfdev-5 

Sorry, if this has been answered already. But will this change also affect memory usage since it looks like you are storing the `batch` and `output` again in the `state`. I maybe misunderstanding the use case but it seems a bit redundant.

vfdev-5:
@ryanwongsa thanks for chiming in. Basically, we would like to provide a single structure as `state.data` instead of/in addition to `state.batch` and `state.output` (see #1923) without additional copy and sync between. This PR however does not do that and I have to probably close it right now in order to avoid misunderstandings...

PR #2021, opened 2021-06-01
Uniformly handling of metric types for all loggers

### LINKED ISSUE #1965: Make sure that all exp tracking system loggers handle metric types in the same way
## 🚀 Feature

Related to the discussion from here: https://github.com/pytorch/ignite/issues/1476#issuecomment-826314243

We would like to ensure that all our supported experiment tracking loggers handle metric types in the same way as here:

For Neptune (correct):
https://github.com/pytorch/ignite/blob/9230a7319047b37ce19d956e024fa1b86030c30a/ignite/contrib/handlers/neptune_logger.py#L344-L351

In a previous issue we found that the code for WandB is different (incorrect):
https://github.com/pytorch/ignite/blob/9230a7319047b37ce19d956e024fa1b86030c30a/ignite/contrib/handlers/wandb_logger.py#L271-L275

ToDo:
- check `OutputHandler` implementation for all supported loggers: tensorboard_logger.py, clearml_logger.py, mlflow_logger.py, neptune_logger.py, polyaxon_logger.py, tqdm_logger.py, visdom_logger.py, wandb_logger.py
and find those where it is not done as in the "correct" example of Neptune.
- reimplement those loggers using above "correct" example of Neptune.

If interested in working on this issue, please comment out in here such that it could be assigned to you.

Muktan:
Fixes #1965 

Description: 
Checked all the output handler class but only wandb_logger had changes all other loggers check the instance types. 
Most loggers added `elif isinstance(value, torch.Tensor) and value.ndimension() == 1:` condition and inside condition iterated over each item in value and logged it. But I have not added this for wandb_logger because it may change the initial logging behaviour.

Check list:

### DIFF
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """TQDM logger."""
+import numbers
 import warnings
 from collections import OrderedDict
 from typing import Any, Callable, List, Optional, Union
@@ -273,17 +274,16 @@ class _OutputHandler(BaseOutputHandler):
 
         rendered_metrics = OrderedDict()
         for key, value in metrics.items():
-            if isinstance(value, torch.Tensor):
-                if value.ndimension() == 0:
-                    rendered_metrics[key] = value.item()
-                elif value.ndimension() == 1:
-                    for i, v in enumerate(value):
-                        k = f"{key}_{i}"
-                        rendered_metrics[k] = v.item()
-                else:
-                    warnings.warn(f"ProgressBar can not log tensor with {value.ndimension()} dimensions")
-            else:
+            if isinstance(value, numbers.Number) or isinstance(value, str):
                 rendered_metrics[key] = value
+            elif isinstance(value, torch.Tensor) and value.ndimension() == 0:
+                rendered_metrics[key] = value.item()
+            elif isinstance(value, torch.Tensor) and value.ndimension() == 1:
+                for i, v in enumerate(value):
+                    k = f"{key}_{i}"
+                    rendered_metrics[k] = v.item()
+            else:
+                warnings.warn(f"ProgressBar can not log tensor with {value.ndimension()} dimensions")
 
         if rendered_metrics:
             logger.pbar.set_postfix(rendered_metrics)  # type: ignore[attr-defined]
--- a/ignite/contrib/handlers/tqdm_logger.py
+++ b/ignite/contrib/handlers/tqdm_logger.py
@@ -3,7 +3,7 @@
 import numbers
 import warnings
 from collections import OrderedDict
-from typing import Any, Callable, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Union
 
 import torch
 
@@ -272,7 +272,7 @@ class _OutputHandler(BaseOutputHandler):
 
         metrics = self._setup_output_metrics(engine)
 
-        rendered_metrics = OrderedDict()
+        rendered_metrics = OrderedDict()  # type: Dict[str, Union[str, float, numbers.Number]]
         for key, value in metrics.items():
             if isinstance(value, numbers.Number) or isinstance(value, str):
                 rendered_metrics[key] = value
--- a/tests/ignite/contrib/handlers/test_wandb_logger.py
+++ b/tests/ignite/contrib/handlers/test_wandb_logger.py
@@ -119,7 +119,7 @@ def test_output_handler_metric_names():
     mock_logger.log = MagicMock()
 
     wrapper(mock_engine, mock_logger, Events.ITERATION_STARTED)
-    mock_logger.log.assert_called_once_with({"tag/a": 55.56, "tag/c": "Some text"}, step=7, sync=None)
+    mock_logger.log.assert_called_once_with({"tag/a": 55.56}, step=7, sync=None)
 
     # all metrics
     wrapper = OutputHandler("tag", metric_names="all")
--- a/tests/ignite/contrib/handlers/test_wandb_logger.py
+++ b/tests/ignite/contrib/handlers/test_wandb_logger.py
@@ -119,7 +119,7 @@ def test_output_handler_metric_names():
     mock_logger.log = MagicMock()
 
     wrapper(mock_engine, mock_logger, Events.ITERATION_STARTED)
-    mock_logger.log.assert_called_once_with({"tag/a": 55.56}, step=7, sync=None)
+    mock_logger.log.assert_called_once_with({"tag/a": 55.56, "tag/c": "Some text"}, step=7, sync=None)
 
     # all metrics
     wrapper = OutputHandler("tag", metric_names="all")
--- a/tests/ignite/contrib/handlers/test_wandb_logger.py
+++ b/tests/ignite/contrib/handlers/test_wandb_logger.py
@@ -133,6 +133,30 @@ def test_output_handler_metric_names():
     wrapper(mock_engine, mock_logger, Events.ITERATION_STARTED)
     mock_logger.log.assert_called_once_with({"tag/a": 12.23, "tag/b": 23.45}, step=5, sync=None)
 
+    # log a torch vector
+    wrapper = OutputHandler("tag", metric_names="all")
+    mock_logger = MagicMock(spec=WandBLogger)
+    mock_logger.log = MagicMock()
+    vector = torch.tensor([0.1, 0.2, 0.1, 0.2, 0.33])
+    mock_engine = MagicMock()
+    mock_engine.state = State(metrics={"a": vector})
+    mock_engine.state.iteration = 5
+
+    wrapper(mock_engine, mock_logger, Events.ITERATION_STARTED)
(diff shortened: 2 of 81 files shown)


vfdev-5:
@Muktan thanks for the quick PR ! 

> Checked all the output handler class but only wandb_logger had changes all other loggers check the instance types.
Most loggers added elif isinstance(value, torch.Tensor) and value.ndimension() == 1: condition and inside condition iterated over each item in value and logged it. But I have not added this for wandb_logger because it may change the initial logging behaviour.

I think the point of the issue #1965 is to change to previous behaviour and make it uniform for all loggers.


Also, how about tqdm logger, I think we have to rework _OutputHandler as well : https://github.com/pytorch/ignite/blob/be2bff832b88cfbc0a21828a38c664930f297409/ignite/contrib/handlers/tqdm_logger.py#L275-L286

Muktan:
@vfdev-5 thanks for reply!

> I think the point of the issue #1965 is to change to previous behaviour and make it uniform for all loggers.
So do we have to change the actual behaviour.

For example for wandblogger it would be like:

```python
rendered_metrics = {}

if self.tag is not None:
     for name, value in metrics.items():
        if isinstance(value, numbers.Number) or isinstance(value, torch.Tensor) and value.ndimension() == 0: 
             rendered_metrics[f"{self.tag}/{name}"] = value
        elif isinstance(value, torch.Tensor) and value.ndimension() == 1: 
             for i, v in enumerate(value): 
                 rendered_metrics[f"{self.tag}/{name}/{i}"] = v
        else: 
             warnings.warn(f"WandBLogger output_handler can not log metrics value type {type(value)}") 

logger.log(rendered_metrics, step=global_step, sync=self.sync)
```
But the problem with that would occur when we run tests for the file. we have a test case (`test_output_handler_metric_names`) having `value` (`metrics={"a": 55.56, "c": "Some text"}`) as a string which in this case will raise the warning.
So, shall we update the test case? 


> Also, how about tqdm logger, I think we have to rework _OutputHandler as well :
https://github.com/pytorch/ignite/blob/be2bff832b88cfbc0a21828a38c664930f297409/ignite/contrib/handlers/tqdm_logger.py#L275-L286

Yes, for tqdm logger we have to rework _OutputHandler. 

I thought we have to make sure there is instance checking done. But it seems we need to follow the code structure of handling of metric types  as in neptune, for all other loggers, correct?
In that test cases won't fail?

vfdev-5:
> But the problem with that would occur when we run tests for the file. we have a test case (test_output_handler_metric_names) having value (metrics={"a": 55.56, "c": "Some text"}) as a string which in this case will raise the warning.
So, shall we update the test case?

That's a good point. I think logging a metric as string is not a common usage. Yes, I think updating the test cases is a solution.

vfdev-5:
> Okay, I will try to make a similar code structure for all the other loggers. Also, does the above wandb implementation look correct?

It looks like correct. I was also exploring mlflow logger implementation and explicit call `v.item()` probably could make sense here as well ?

Let me list all implementations and let's decide:
- ClearML
https://github.com/pytorch/ignite/blob/8cc474644504b2afaceec74790a8b3b43c352832/ignite/contrib/handlers/clearml_logger.py#L321-L330

- MLflow
https://github.com/pytorch/ignite/blob/8cc474644504b2afaceec74790a8b3b43c352832/ignite/contrib/handlers/mlflow_logger.py#L234-L244

- Neptune
https://github.com/pytorch/ignite/blob/8cc474644504b2afaceec74790a8b3b43c352832/ignite/contrib/handlers/neptune_logger.py#L342-L349

- Polyaxon
https://github.com/pytorch/ignite/blob/8cc474644504b2afaceec74790a8b3b43c352832/ignite/contrib/handlers/polyaxon_logger.py#L245-L255

- TensorBoard
https://github.com/pytorch/ignite/blob/8cc474644504b2afaceec74790a8b3b43c352832/ignite/contrib/handlers/tensorboard_logger.py#L287-L294

- TQDM
https://github.com/pytorch/ignite/blob/8cc474644504b2afaceec74790a8b3b43c352832/ignite/contrib/handlers/tqdm_logger.py#L274-L286

- Visdom
https://github.com/pytorch/ignite/blob/8cc474644504b2afaceec74790a8b3b43c352832/ignite/contrib/handlers/visdom_logger.py#L367-L382

- W&B
https://github.com/pytorch/ignite/blob/be2bff832b88cfbc0a21828a38c664930f297409/ignite/contrib/handlers/wandb_logger.py#L279-L283

Actually, I think polyaxon and mlflow and tqdm do the data type dispatch in the correct way:
```
if value is a number:
    log or append it to a dict as is: value
elif value is a 0-d Tensor:
    log or append it to a dict using: value.item() 
elif value is a 1-d Tensor:
    loop over values and
        log or append it to a dict using: v.item() 
else:
    warn about unhandled dtype
```

Let's implement this logic to all OutputHandlers. What do you think ?

Muktan:
Yes, `value.item()` for type tensor seems more apt.

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -273,17 +274,16 @@ def __call__(self, engine: Engine, logger: ProgressBar, event_name: Union[str, E
...
-                if value.ndimension() == 0:
-                    rendered_metrics[key] = value.item()
-                elif value.ndimension() == 1:
-                    for i, v in enumerate(value):
-                        k = f"{key}_{i}"
-                        rendered_metrics[k] = v.item()
-                else:
-                    warnings.warn(f"ProgressBar can not log tensor with {value.ndimension()} dimensions")
-            else:
+            if isinstance(value, numbers.Number) or isinstance(value, str):
```
Muktan, line 277: tqdm has logs string values so have to explicitly add  `isinstance(value, str)`

tests/ignite/contrib/handlers/test_wandb_logger.py
```diff
@@ -119,7 +119,7 @@ def test_output_handler_metric_names():
     mock_logger.log = MagicMock()
 
     wrapper(mock_engine, mock_logger, Events.ITERATION_STARTED)
-    mock_logger.log.assert_called_once_with({"tag/a": 55.56, "tag/c": "Some text"}, step=7, sync=None)
+    mock_logger.log.assert_called_once_with({"tag/a": 55.56}, step=7, sync=None)
```
Muktan, line 122: updated the wandb logger test, updating string value test case. Or, shall we add `isinstance(value, str)` in wandb logger as used in tqdm logger.

vfdev-5: This is a good question. I was looking at w&b docs if they can accept `str` data on their side. If you can test that explicitly it would be great and if they support str data then it would make sense to add this special case as tqdm...

Muktan: Ok, I will check and make changes accordingly.

Muktan: @vfdev-5 
I found some wandb doc links: https://docs.wandb.ai/guides/track/log, https://docs.wandb.ai/ref/python/log which shows logging of various objects like Image, 3D viz, audio.. etc
However, there does not seem to be any use cases in logging the string data type. 
So I think it would be fine if we exclude the string data type from logging.

vfdev-5: Yes, I was also inspecting this wandb docs and didn't find on str dtype. We can try to check that explicitly on a testing project to log a str and see what happens. I think they can handle that but metric value as str is something unusual

Muktan: Yes, metric values as str are unusual.

Muktan: I created a hugging face project that trains a Roberta-base on a sequence classification task and tried to log text metrics (eval/text).

```python
def compute_metrics(eval_pred):
        random_str = ["text1", "text2", "text3", "text4", "text5", "text6"]
        pred, act = eval_pred
        return {'score':np.sqrt(mean_squared_error(pred,act)), 'text': random_str[random.randint(0,len(random_str)-1)]}
```

But when I opened the run page (where the plots regarding logged metrics are plotted) I encountered following warning.

String logging.
Screenshot from 2021-06-02 11-18-48

Number logging.
Screenshot from 2021-06-02 11-24-06

However,
we can see the values (eval/text) being logged for string in the overview tab.
Screenshot from 2021-06-02 11-30-14

vfdev-5: Thanks a lot for checking that ! OK, let's add a special case for wandb as for tqdm.

vfdev-5 (approved):
Thanks a lot @Muktan ! Looks good to me !

vfdev-5:
@Muktan could you please fix mypy issues ?

Muktan:
@vfdev-5 shall I add the test for missing lines as indicated by codecov?

vfdev-5:
> @vfdev-5 shall I add the test for missing lines as indicated by codecov?

@Muktan yes, please

vfdev-5:
Thanks for the update @Muktan ! Looks good, let's what CI and codecov says.

Issue #2018, opened 2021-05-31
Fix warnings in CI


vfdev-5:
## 🚀 Feature

Recent PRs while moving modules from contrib to core introduced few warnings to fix:
- https://github.com/pytorch/ignite/runs/2694096058#step:12:2268
```
  /home/runner/work/ignite/ignite/ignite/contrib/handlers/__init__.py:2: DeprecationWarning: /home/runner/work/ignite/ignite/ignite/contrib/handlers/lr_finder.py has been moved to /ignite/handlers/lr_finder.py and will be removed in version 0.6.0.
   Please refer to the documentation for more details.
    from ignite.contrib.handlers.lr_finder import FastaiLRFinder
```
- https://github.com/pytorch/ignite/runs/2694096058#step:12:2274
```
  /home/runner/work/ignite/ignite/ignite/contrib/handlers/__init__.py:5: DeprecationWarning: /home/runner/work/ignite/ignite/ignite/contrib/handlers/param_scheduler.py has been moved to /ignite/handlers/param_scheduler.py and will be removed in version 0.6.0.
   Please refer to the documentation for more details.
    from ignite.contrib.handlers.param_scheduler import (
```

Other warnings to fix:
- https://github.com/pytorch/ignite/runs/2694096324#step:12:808
- https://github.com/pytorch/ignite/runs/2694096324#step:12:812
```
tests\ignite\handlers\test_checkpoint.py:4
  D:\a\ignite\ignite\tests\ignite\handlers\test_checkpoint.py:4: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
    from collections import Mapping, OrderedDict

tests/ignite/test_utils.py::test_deprecated
  D:\a\ignite\ignite\tests\ignite\test_utils.py:221: DeprecationWarning: This function has been deprecated since version 0.4.2 and will be removed in version 0.6.0.
   Please refer to the documentation for more details.
    assert func_check_warning() == 24
```

- https://github.com/pytorch/ignite/runs/2694096324#step:12:1220 (lr_finder)
- https://github.com/pytorch/ignite/runs/2694096324#step:12:1225 (lr_finder)
- https://github.com/pytorch/ignite/runs/2694096324#step:12:1230 (lr_finder)
```
 tests/ignite/contrib/handlers/test_lr_finder.py::test_lr_suggestion_unexpected_curve
[...]

PR #2023, opened 2021-06-01
Fixes Part of #2018: Fix warnings in CI

OBITORASU:
Fixes part of Issue #2018 

## Description:
I have tried my best to fix all the warnings aside from the `lr_finder` ones since somebody else was tagged to handle them. 

Fix 1:
https://github.com/pytorch/ignite/runs/2694096058#step:12:2268 
With reference to this warning in the CI, I updated `from ignite.contrib.handlers.lr_finder import FastaiLRFinder` to `from ignite.handlers.lr_finder import FastaiLRFinder` to suppress the warnings.

Fix 2:
https://github.com/pytorch/ignite/runs/2694096058#step:12:2274
Similar to Fix 1, I updated `from ignite.contrib.handlers.param_scheduler import (` to `from ignite.handlers.param_scheduler import (` to suppress the warnings.

Fix 3:
https://github.com/pytorch/ignite/runs/2694096324#step:12:808
Referring to this one, I changed the imports from `from collections import Mapping, OrderedDict` to `from collections.abc import Mapping` and `from collections import OrderedDict` respectively. This should fix the warnings as `Mapping` is a part of `collections.abc` while `OrderedDict` is not hence the two separate imports.

Fix 4:
**Other warnings (with reference to the original issue)**
All the instances of `np.long` which is deprecated were replaced with `np.int_` (compatible with Python int and C long) after referring to the respective release documentation where `np.long` was deprecated. 

For the CUDA warning namely `("torch.cuda.amp.GradScaler is enabled, but CUDA is not available.  Disabling.")`, I set `check_arg_mock.return_value = None, torch.cuda.amp.GradScaler(enabled=True)` to `check_arg_mock.return_value = None, torch.cuda.amp.GradScaler(enabled=False)` (Turned True to False) under the `test_create_supervised_training_scalar_assignment` function where the warning was being generated. 

**EDIT 1**
Fix 5:
As suggested, instead of:
```python
     with pytest.deprecated_call():
         func_check_warning()
     assert func_check_warning() == 24
```
Started using:
```python
    with pytest.deprecated_call():
        assert func_check_warning() == 24
```
Also migrated to using `pytest.warns` to reduce the overall lines of code and make things simpler.

Check list:

- [ ] New tests are added (if a new feature is added) (N/A)
- [ ] New doc strings: description and/or example code are in RST format (N/A)
- [ ] Documentation is updated (if required) (N/A)

### DIFF
--- a/ignite/handlers/lr_finder.py
+++ b/ignite/handlers/lr_finder.py
@@ -224,10 +224,7 @@ class FastaiLRFinder:
 
         """
         try:
-            import matplotlib
             from matplotlib import pyplot as plt
-
-            matplotlib.use("tkagg")
         except ImportError:
             raise RuntimeError(
                 "This method requires matplotlib to be installed. "
--- a/ignite/handlers/lr_finder.py
+++ b/ignite/handlers/lr_finder.py
@@ -224,7 +224,10 @@ class FastaiLRFinder:
 
         """
         try:
+            import matplotlib
             from matplotlib import pyplot as plt
+
+            matplotlib.use("tkagg")
         except ImportError:
             raise RuntimeError(
                 "This method requires matplotlib to be installed. "
@@ -302,12 +305,12 @@ class FastaiLRFinder:
         if not self._history:
             raise RuntimeError("learning rate finder didn't run yet so lr_suggestion can't be returned")
         loss = self._history["loss"]
-        min_loss_idx = torch.tensor(loss).argmin()
+        min_loss_idx = loss.clone().detach().argmin()
         # Ignore the increasing part of the curve
         decreasing_losses = self._history["loss"][: int(min_loss_idx.item()) + 1]
         if len(decreasing_losses) < 3:
             raise RuntimeError("FastaiLRFinder got unexpected curve shape, the curve should be somehow U-shaped")
-        losses = torch.tensor(decreasing_losses)
+        losses = decreasing_losses.clone().detach()
         grads = torch.tensor([0.5 * (losses[i + 1] - losses[i - 1]) for i in range(1, len(losses) - 1)])
         min_grad_idx = grads.argmin() + 1
         return self._history["lr"][int(min_grad_idx)]
--- a/ignite/handlers/lr_finder.py
+++ b/ignite/handlers/lr_finder.py
@@ -305,12 +305,12 @@ class FastaiLRFinder:
         if not self._history:
             raise RuntimeError("learning rate finder didn't run yet so lr_suggestion can't be returned")
         loss = self._history["loss"]
-        min_loss_idx = loss.clone().detach().argmin()
+        min_loss_idx = torch.tensor(loss).argmin()
         # Ignore the increasing part of the curve
         decreasing_losses = self._history["loss"][: int(min_loss_idx.item()) + 1]
         if len(decreasing_losses) < 3:
             raise RuntimeError("FastaiLRFinder got unexpected curve shape, the curve should be somehow U-shaped")
-        losses = decreasing_losses.clone().detach()
+        losses = torch.tensor(decreasing_losses)
         grads = torch.tensor([0.5 * (losses[i + 1] - losses[i - 1]) for i in range(1, len(losses) - 1)])
         min_grad_idx = grads.argmin() + 1
         return self._history["lr"][int(min_grad_idx)]
--- a/tests/ignite/contrib/handlers/test_lr_finder.py
+++ b/tests/ignite/contrib/handlers/test_lr_finder.py
@@ -414,7 +414,7 @@ def test_no_matplotlib(no_site_packages, lr_finder):
 
 def test_plot_single_param_group(lr_finder, mnist_to_save, dummy_engine_mnist, mnist_dataloader):
 
-    with lr_finder.attach(dummy_engine_mnist, mnist_to_save, end_lr=100, smooth_f=0.04) as trainer_with_finder:
+    with lr_finder.attach(dummy_engine_mnist, mnist_to_save, end_lr=20, smooth_f=0.04) as trainer_with_finder:
         trainer_with_finder.run(mnist_dataloader)
 
     lr_finder.plot()
@@ -443,7 +443,7 @@ def test_plot_multiple_param_groups(
 ):
 
     with lr_finder.attach(
-        dummy_engine_mulitple_param_groups, to_save_mulitple_param_groups, end_lr=100, smooth_f=0.04
+        dummy_engine_mulitple_param_groups, to_save_mulitple_param_groups, end_lr=20, smooth_f=0.04
     ) as trainer_with_finder:
         trainer_with_finder.run(dataloader_plot)
 
--- a/tests/ignite/contrib/handlers/test_lr_finder.py
+++ b/tests/ignite/contrib/handlers/test_lr_finder.py
@@ -414,7 +414,7 @@ def test_no_matplotlib(no_site_packages, lr_finder):
 
 def test_plot_single_param_group(lr_finder, mnist_to_save, dummy_engine_mnist, mnist_dataloader):
 
-    with lr_finder.attach(dummy_engine_mnist, mnist_to_save, end_lr=50, smooth_f=0.04) as trainer_with_finder:
+    with lr_finder.attach(dummy_engine_mnist, mnist_to_save, end_lr=100, smooth_f=0.04) as trainer_with_finder:
         trainer_with_finder.run(mnist_dataloader)
 
     lr_finder.plot()
@@ -443,7 +443,7 @@ def test_plot_multiple_param_groups(
 ):
 
     with lr_finder.attach(
-        dummy_engine_mulitple_param_groups, to_save_mulitple_param_groups, end_lr=50, smooth_f=0.04
+        dummy_engine_mulitple_param_groups, to_save_mulitple_param_groups, end_lr=100, smooth_f=0.04
     ) as trainer_with_finder:
         trainer_with_finder.run(dataloader_plot)
 
--- a/tests/ignite/contrib/handlers/test_lr_finder.py
+++ b/tests/ignite/contrib/handlers/test_lr_finder.py
@@ -539,8 +539,9 @@ def _test_distrib_integration_mnist(device):
 
 @pytest.mark.distributed
 @pytest.mark.skipif(not idist.has_native_dist_support, reason="Skip if no native dist support")
-def test_distrib_cpu(distributed_context_single_node_gloo):
-    device = torch.device("cpu")
+def test_distrib_gloo_cpu_or_gpu(distributed_context_single_node_gloo):
+
+    device = idist.device()
     _test_distrib_log_lr_and_loss(device)
     _test_distrib_integration_mnist(device)
 
@@ -548,8 +549,9 @@ def test_distrib_cpu(distributed_context_single_node_gloo):
 @pytest.mark.distributed
 @pytest.mark.skipif(not idist.has_native_dist_support, reason="Skip if no native dist support")
 @pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
-def test_distrib_gpu(distributed_context_single_node_nccl):
-    device = torch.device(f"cuda:{distributed_context_single_node_nccl['local_rank']}")
+def test_distrib_nccl_gpu(distributed_context_single_node_nccl):
+
+    device = idist.device()
     _test_distrib_log_lr_and_loss(device)
     _test_distrib_integration_mnist(device)
 
--- a/tests/ignite/contrib/handlers/test_lr_finder.py
+++ b/tests/ignite/contrib/handlers/test_lr_finder.py
@@ -414,7 +414,7 @@ def test_no_matplotlib(no_site_packages, lr_finder):
 
 def test_plot_single_param_group(lr_finder, mnist_to_save, dummy_engine_mnist, mnist_dataloader):
 
-    with lr_finder.attach(dummy_engine_mnist, mnist_to_save, end_lr=20, smooth_f=0.04) as trainer_with_finder:
+    with lr_finder.attach(dummy_engine_mnist, mnist_to_save, end_lr=50, smooth_f=0.04) as trainer_with_finder:
         trainer_with_finder.run(mnist_dataloader)
 
     lr_finder.plot()
@@ -443,7 +443,7 @@ def test_plot_multiple_param_groups(
 ):
 
     with lr_finder.attach(
-        dummy_engine_mulitple_param_groups, to_save_mulitple_param_groups, end_lr=20, smooth_f=0.04
+        dummy_engine_mulitple_param_groups, to_save_mulitple_param_groups, end_lr=50, smooth_f=0.04
     ) as trainer_with_finder:
         trainer_with_finder.run(dataloader_plot)
 
--- a/tests/ignite/contrib/metrics/test_precision_recall_curve.py
+++ b/tests/ignite/contrib/metrics/test_precision_recall_curve.py
@@ -28,7 +28,7 @@ def test_no_sklearn(mock_no_sklearn):
 def test_precision_recall_curve():
     size = 100
     np_y_pred = np.random.rand(size, 1)
-    np_y = np.zeros((size,), dtype=np.long)
+    np_y = np.zeros((size,), dtype=np.int_)
     np_y[size // 2 :] = 1
     sk_precision, sk_recall, sk_thresholds = precision_recall_curve(np_y, np_y_pred)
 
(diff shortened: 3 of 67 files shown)


OBITORASU:
A bit confused regarding https://github.com/pytorch/ignite/runs/2694096324#step:12:812 
While reading the code it seems that this warning is supposed to be a part of the test as suggested by a comment `# Tests that the function emits DeprecationWarning` for the function under consideration `func_check_warning()`. Should we be bothered about this one?

vfdev-5:
Thanks for the PR @OBITORASU !

> A bit confused regarding https://github.com/pytorch/ignite/runs/2694096324#step:12:812
While reading the code it seems that this warning is supposed to be a part of the test as suggested by a comment # Tests that the function emits DeprecationWarning for the function under consideration func_check_warning(). Should we be bothered about this one?

I think this code is responsible for emitting uncatched deprecation warning https://github.com/pytorch/ignite/blob/master/tests/ignite/test_utils.py#L219-L221

Instead of 
```python
    with pytest.deprecated_call():
        func_check_warning()
    assert func_check_warning() == 24
```
we should do
```python
    with pytest.deprecated_call():
        assert func_check_warning() == 24
```

I think also we can simplify all the code below:
https://github.com/pytorch/ignite/blob/be2bff832b88cfbc0a21828a38c664930f297409/tests/ignite/test_utils.py#L219-L234
using `pytest.warns` (usage: https://github.com/pytorch/ignite/blob/2a4ed134b6099541e1e15b35bea2b2c800036e4e/tests/ignite/contrib/engines/test_common.py#L147)

What do you think ?

tests/ignite/contrib/metrics/test_precision_recall_curve.py
```diff
@@ -28,7 +28,7 @@ def test_no_sklearn(mock_no_sklearn):
 def test_precision_recall_curve():
     size = 100
     np_y_pred = np.random.rand(size, 1)
-    np_y = np.zeros((size,), dtype=np.long)
+    np_y = np.zeros((size,), dtype=np.int_)
```
vfdev-5, line 31: Can't we do simply `np_y = np.zeros((size,))` without specifying the dtype ? By default it will be equivalent to `long` ?

OBITORASU: Sure I can update it that way as well.

OBITORASU:
> 
> 
> Thanks for the PR @OBITORASU !
> 
> > A bit confused regarding https://github.com/pytorch/ignite/runs/2694096324#step:12:812
> > While reading the code it seems that this warning is supposed to be a part of the test as suggested by a comment # Tests that the function emits DeprecationWarning for the function under consideration func_check_warning(). Should we be bothered about this one?
> 
> I think this code is responsible for emitting uncatched deprecation warning https://github.com/pytorch/ignite/blob/master/tests/ignite/test_utils.py#L219-L221
> 
> Instead of
> 
> ```python
>     with pytest.deprecated_call():
>         func_check_warning()
>     assert func_check_warning() == 24
> ```
> 
> we should do
> 
> ```python
>     with pytest.deprecated_call():
>         assert func_check_warning() == 24
> ```
> 
> I think also we can simplify all the code below:
> 
> https://github.com/pytorch/ignite/blob/be2bff832b88cfbc0a21828a38c664930f297409/tests/ignite/test_utils.py#L219-L234
> 
> using `pytest.warns` (usage:
> 
> https://github.com/pytorch/ignite/blob/2a4ed134b6099541e1e15b35bea2b2c800036e4e/tests/ignite/contrib/engines/test_common.py#L147
> )
> 
> What do you think ?

Trying to get this done as mentioned using `pytest.warns` to reduce the extra lines of code.

OBITORASU:
Made the requested changes. Removed the use of `dtypes` and updated `func_check_warning` function as requested and guided. PTAL :)

KickItLikeShika:
Thanks a lot for the PR @OBITORASU ! If you would like to tackle fixing `lr_finder` that would be great! and I can provide you help, and help reviewing the PR, as I worked on `lr_finder` a lot.

OBITORASU:
> 
> 
> Thanks a lot for the PR @OBITORASU ! If you would like to tackle fixing `lr_finder` that would be great! and I can provide you help, and help reviewing the PR, as I worked on `lr_finder` a lot.

Yeah sure I don't mind tackling the `lr_finder` issues at all. So I went through the logs of `lr_finder` and came up with the following solutions and troubles. I will list them down.


``` 
  D:\a\ignite\ignite\ignite\handlers\lr_finder.py:305: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
    min_loss_idx = torch.tensor(loss).argmin()
 ```
```
  D:\a\ignite\ignite\ignite\handlers\lr_finder.py:310: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
    losses = torch.tensor(decreasing_losses)
```

These two warnings are self explanatory and changing `torch.tensor(loss).argmin()` to `loss.clone().detach().argmin()` and `torch.tensor(decreasing_losses)` to `decreasing_losses.clone().detach()` should fix it.

For the next one namely:
```
c:\miniconda3\envs\test\lib\site-packages\torchvision\datasets\mnist.py:502: UserWarning: The given NumPy array is not writeable, and PyTorch does not support non-writeable tensors. This means you can write to the underlying (supposedly non-writeable) NumPy array using the tensor. You may want to copy the array to protect its data or make it writeable before converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally at  ..\torch\csrc\utils\tensor_numpy.cpp:143.)
    return torch.from_numpy(parsed.astype(m[2], copy=False)).view(*s)
```
I think changing `copy=False` to `copy=True` should suppress the warnings.


For the matplotlib warning that is:
```
D:\a\ignite\ignite\ignite\handlers\lr_finder.py:293: UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
    plt.show()
``` 
A simple addition of `matplotlib.use( 'tkagg')` should fix it in my opinion.

The remaining two warnings:

1.
 ```
  D:\a\ignite\ignite\ignite\handlers\lr_finder.py:164: UserWarning: Run completed without loss diverging, increase end_lr, decrease diverge_th or look at lr_finder.plot()
    warnings.warn(
 ```

2.
```
c:\miniconda3\envs\test\lib\site-packages\torch\nn\modules\loss.py:528: UserWarning: Using a target size (torch.Size([10])) that is different to the input size (torch.Size([1])). This will likely lead to incorrect results due to broadcasting. Please ensure they have the same size.
    return F.mse_loss(input, target, reduction=self.reduction)
 ```
I might need some guidance with these two. Do tell me if the approach for the previous warnings are fine or some potential changes are needed in my approach. Thanks in advance!

KickItLikeShika:
Your solutions seem OK, just make sure they work please. 
For this warning:
```
D:\a\ignite\ignite\ignite\handlers\lr_finder.py:164: UserWarning: Run completed without loss diverging, increase end_lr, decrease diverge_th or look at lr_finder.plot()
   warnings.warn(
```
here:
https://github.com/pytorch/ignite/blob/1d992ed8a10c6b47eb5ede5c76fa50be45e9bf7f/tests/ignite/contrib/handlers/test_lr_finder.py#L351
https://github.com/pytorch/ignite/blob/1d992ed8a10c6b47eb5ede5c76fa50be45e9bf7f/tests/ignite/contrib/handlers/test_lr_finder.py#L391

increasing the `end_lr` in `attach` should solve this:
```py
with lr_finder.attach(dummy_engine_mnist, mnist_to_save, end_lr=...) as trainer_with_finder:
```
And for this warning:
```
c:\miniconda3\envs\test\lib\site-packages\torch\nn\modules\loss.py:528: UserWarning: Using a target size (torch.Size([10])) that is different to the input size (torch.Size([1])). This will likely lead to incorrect results due to broadcasting. Please ensure they have the same size.
    return F.mse_loss(input, target, reduction=self.reduction)
```
You will see here the out channel of the model if `1D` by default: https://github.com/pytorch/ignite/blob/1d992ed8a10c6b47eb5ede5c76fa50be45e9bf7f/tests/ignite/contrib/handlers/test_lr_finder.py#L36

and the data we use if of shape `(100, 2, 10)`: https://github.com/pytorch/ignite/blob/1d992ed8a10c6b47eb5ede5c76fa50be45e9bf7f/tests/ignite/contrib/handlers/test_lr_finder.py#L137

So I think we need to change the out channel at the model to 10 instead of 1, but this can be a tricky fix and cause other problem, you need to get something that maybe allow minimum changes I guess.

ignite/handlers/lr_finder.py
```diff
@@ -302,12 +305,12 @@ def lr_suggestion(self) -> Any:
         if not self._history:
             raise RuntimeError("learning rate finder didn't run yet so lr_suggestion can't be returned")
         loss = self._history["loss"]
-        min_loss_idx = torch.tensor(loss).argmin()
+        min_loss_idx = loss.clone().detach().argmin()
```
vfdev-5, line 308: @OBITORASU thanks for pointing out that. @KickItLikeShika I think we should make sure that `loss` is a float and not a tensor requires grad:
https://github.com/pytorch/ignite/blob/65d9844de3b7faf23371dd4c82cd16d02d7e67aa/ignite/handlers/lr_finder.py#L138-L140
In this way we can reuse `torch.tensor(loss).argmin()` without any need to detach.

ignite/handlers/lr_finder.py
```diff
@@ -224,7 +224,10 @@ def plot(
 
         """
         try:
+            import matplotlib
             from matplotlib import pyplot as plt
+
+            matplotlib.use("tkagg")
```
vfdev-5, line 230: @OBITORASU do we really need that in the code ? I think either this should go to the tests or simply be removed...

ignite/handlers/lr_finder.py
OBITORASU: Doing a fallback on this one to test out the other changes, once that is done I will come back to fixing this one again.

ignite/handlers/lr_finder.py
OBITORASU: Trying some other approach then.

OBITORASU:
@KickItLikeShika increasing the `end_lr` value by quite a margin didn't seem to be working. Did I do anything wrong need a heads up on this one.

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -414,7 +414,7 @@ def test_no_matplotlib(no_site_packages, lr_finder):
 
 def test_plot_single_param_group(lr_finder, mnist_to_save, dummy_engine_mnist, mnist_dataloader):
 
-    with lr_finder.attach(dummy_engine_mnist, mnist_to_save, end_lr=20, smooth_f=0.04) as trainer_with_finder:
+    with lr_finder.attach(dummy_engine_mnist, mnist_to_save, end_lr=100, smooth_f=0.04) as trainer_with_finder:
```
KickItLikeShika, line 417: Thanks for the updates! as this doesn't work, please try to manipulate `diverge_th`, this may solve this issue

tests/ignite/contrib/handlers/test_lr_finder.py
```diff
@@ -443,7 +443,7 @@ def test_plot_multiple_param_groups(
 ):
 
     with lr_finder.attach(
-        dummy_engine_mulitple_param_groups, to_save_mulitple_param_groups, end_lr=20, smooth_f=0.04
+        dummy_engine_mulitple_param_groups, to_save_mulitple_param_groups, end_lr=100, smooth_f=0.04
```
KickItLikeShika, line 446: Here too

tests/ignite/contrib/handlers/test_lr_finder.py
KickItLikeShika: For more info about the `args`, please check the docs

OBITORASU:
@vfdev-5 @KickItLikeShika I tried to think around for possible fixes regarding the `lr_finder` warnings but none of the fixes that came to my mind worked out definitively, it mostly resulted in more errors and test failures. I think someone better suited to handle the `lr_finder` issues should take it up in my place, since I am having quite a bit of trouble handling them myself. I hope the other fixes are considerable under the PR for a merge. Thank you so much for your time and effort.

**NOTE: I have rolled back any changes introduced to handle `lr_finder` warnings back to default settings/values.**

tests/ignite/contrib/handlers/test_lr_finder.py
OBITORASU: I think I will skip on the `lr_finder` issues since I had a quite bit of trouble tackling them. I feel I am not capable enough to handle them myself and someone better than me should pick this one up, sorry for the inconvenience.

KickItLikeShika: No problem, If you need more time/send a separate PR for `lr_finder`, I can help you with that, If you don't want to work on `lr_finder`, no problem! thanks for the updates!

OBITORASU: Sure thing! Thanks for all the help and guidance till now. I will keep the PR as it is and wait for further review, I have rolled back the possible changes back to their default state as of now. If anything clicks for me again I will make sure to drop another follow up PR and ask for help if need be. Thank you so much!

vfdev-5:
@OBITORASU @KickItLikeShika maybe we can already merge the code with fixed warnings and in a later pr fix others ?

tests/ignite/engine/test_create_supervised.py
```diff
@@ -96,7 +96,7 @@ def test_create_supervised_training_scalar_assignment():
     optimizer = SGD(model.parameters(), 0.1)
 
     with mock.patch("ignite.engine._check_arg") as check_arg_mock:
-        check_arg_mock.return_value = None, torch.cuda.amp.GradScaler(enabled=True)
+        check_arg_mock.return_value = None, torch.cuda.amp.GradScaler(enabled=False)
```
vfdev-5, line 99: @OBITORASU could you please explain this, why setting it as False ? Thanks

OBITORASU:
> 
> 
> @OBITORASU @KickItLikeShika maybe we can already merge the code with fixed warnings and in a later pr fix others ?

That is what I was suggesting, if these current changes are considerable fixes for the issue, then we can have them merged as a part fix of #2018 and potentially wait for follow up PRs for the entire fix of the issue.

tests/ignite/engine/test_create_supervised.py
OBITORASU: Yes sure I set this one to False because of this error message in the logs `("torch.cuda.amp.GradScaler is enabled, but CUDA is not available. Disabling.")`. I assumed that this defaulted to False due to the lack of CUDA being available, I think this warning stopped showing up in the logs after flipping True to False as well.

vfdev-5:
@OBITORASU thanks a lot for the answer and updates ! 
The last small fix and this PR will be good to be merged:
Could you please fix this line : 
https://github.com/pytorch/ignite/blob/388cc3d8b9635605d702839296e9698b6c3a6484/ignite/contrib/engines/common.py#L26
As it produce this warning: https://github.com/pytorch/ignite/pull/2023/checks?check_run_id=2722741878#step:12:1230

OBITORASU:
Sure thing on it!

vfdev-5:
@OBITORASU thanks for the update ! I checked again for warnings about param_scheduler module if imported from contrib and there is another one : https://github.com/pytorch/ignite/pull/2023/checks?check_run_id=2723282106#step:12:1233

Could you please check and fix the places where we explicitly import from contrib like
```python
from ignite.contrib.handlers.param_scheduler import
```

vfdev-5:
@OBITORASU oh, seems like removing int32 type is not a good idea : https://github.com/pytorch/ignite/pull/2023/checks?check_run_id=2726663747#step:9:212
maybe we can follow numpy deprecation guidelines on how to replace it properly : 
- https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

=> `np.int -> np.int32` ?

OBITORASU:
> 
> 
> @OBITORASU oh, seems like removing int32 type is not a good idea : https://github.com/pytorch/ignite/pull/2023/checks?check_run_id=2726663747#step:9:212
> maybe we can follow numpy deprecation guidelines on how to replace it properly :
> 
>     * https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
> 
> 
> => `np.int -> np.int32` ?

Just noticed, doing it asap.

OBITORASU:
That should take care of both the problems the `np dtypes` one and the migration to `ignite.handler`. I went through the logs again and no problems relating to these two showed up as of now.

vfdev-5 (approved):
Thanks a lot @OBITORASU !

PR #2027, opened 2021-06-02
ENH: Updated Loss metric to use required_output_keys

### LINKED ISSUE #1415: Loss metric to use required_output_keys
## 🚀 Feature

Currently, if we have custom metrics that require data other then `y_pred` and `y`, [we suggest](https://discuss.pytorch.org/t/how-access-inputs-in-custom-ignite-metric/91221/6) to do the following: 
```python
metrics = {
    "Accuracy": Accuracy(),
    "Loss": Loss(criterion, output_transform=lambda out_dict: (out_dict["y_pred"], out_dict["y"])),
    "CustomMetric": CustomMetric()
}

evaluator = create_supervised_evaluator(
    model, 
    metrics=metrics, 
    output_transform=lambda x, y, y_pred: {"x": x, "y": y, "y_pred": y_pred}
)
```

where `CustomMetric` is defined as 
```python
class CustomMetric(Metric):

    required_output_keys = ("y_pred", "y", "x")
```

The idea is to extend this for `Loss` metric to support `required_output_keys`. The main issue with `Loss` now is with `(prediction, target, kwargs)` optional input, where `kwargs` is a dict for extra args for criterion function.

01-vyom:
Fixes #1415 

Description:
Similar to `Metric`, `Loss` metric now supports `required_output_keys`.

Check list:

### DIFF
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -30,9 +30,52 @@ class Loss(Metric):
             metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
             non-blocking. By default, CPU.
 
+    Attributes:
+        required_output_keys: dictionary defines required keys to be found in ``engine.state.output`` if the
+            latter is a dictionary. Default, ``("y_pred", "y", "criterion_kwargs")``. This is useful when the
+            criterion function requires additional argument,
+            which can be passed using ``criterion_kwargs``.
+            See notes below for an example.
+
+    Note:
+
+        Let's implement a Loss metric that requires ``y_pred``, ``y`` and ``criterion_kwargs`` as input
+        for ``criterion`` function. In the example below we show how to setup standard metric like Accuracy
+        and the Loss metric using an ``evaluator`` created
+        with :meth:`~ignite.engine.create_supervised_evaluator` method.
+
+        .. code-block:: python
+
+            import torch
+            import torch.nn as nn
+
+            from ignite.metrics import Accuracy, Loss
+            from ignite.engine import create_supervised_evaluator
+
+            model = ...
+
+            criterion = nn.NLLLoss()
+
+            metrics = {
+                "Accuracy": Accuracy(),
+                "Loss": Loss(criterion)
+            }
+
+            # global criterion kwargs
+            criterion_kwargs = {...}
+
+            evaluator = create_supervised_evaluator(
+                model,
+                metrics=metrics,
+                output_transform=lambda x, y, y_pred: {
+                    "x": x, "y": y, "y_pred": y_pred, "criterion_kwargs": criterion_kwargs}
+            )
+
+            res = evaluator.run(data)
+
     """
 
-    required_output_keys = None
+    required_output_keys = ("y_pred", "y", "criterion_kwargs")
 
     def __init__(
         self,
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -54,7 +54,7 @@ class Loss(Metric):
 
             model = ...
 
-            criterion = nll_loss()
+            criterion = nll_loss
 
             metrics = {
                 "Accuracy": Accuracy(),
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -33,12 +33,12 @@ class Loss(Metric):
     Attributes:
         required_output_keys: dictionary defines required keys to be found in ``engine.state.output`` if the
             latter is a dictionary. Default, ``("y_pred", "y", "criterion_kwargs")``. This is useful when the
-            criterion function requires additional argument, which can be passed using ``criterion_kwargs``.
+            criterion function requires additional arguments, which can be passed using ``criterion_kwargs``.
             See notes below for an example.
 
     Note:
 
-        Let's implement a Loss metric that requires ``y_pred``, ``y`` and ``criterion_kwargs`` as input
+        Let's implement a Loss metric that requires ``x``, ``y_pred``, ``y`` and ``criterion_kwargs`` as input
         for ``criterion`` function. In the example below we show how to setup standard metric like Accuracy
         and the Loss metric using an ``evaluator`` created with
         :meth:`~ignite.engine.create_supervised_evaluator` method.
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -33,28 +33,28 @@ class Loss(Metric):
     Attributes:
         required_output_keys: dictionary defines required keys to be found in ``engine.state.output`` if the
             latter is a dictionary. Default, ``("y_pred", "y", "criterion_kwargs")``. This is useful when the
-            criterion function requires additional argument,
-            which can be passed using ``criterion_kwargs``.
+            criterion function requires additional argument, which can be passed using ``criterion_kwargs``.
             See notes below for an example.
 
     Note:
 
         Let's implement a Loss metric that requires ``y_pred``, ``y`` and ``criterion_kwargs`` as input
         for ``criterion`` function. In the example below we show how to setup standard metric like Accuracy
-        and the Loss metric using an ``evaluator`` created
-        with :meth:`~ignite.engine.create_supervised_evaluator` method.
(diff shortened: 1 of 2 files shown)


01-vyom:
Following is the demo code to check the validity of the above change:

```python
import torch
import torch.nn as nn
from torch.nn.functional import nll_loss

from ignite.metrics import Accuracy, Loss
from ignite.engine import create_supervised_evaluator

model = nn.Linear(10, 3)

metrics = {
    "Accuracy": Accuracy(),
    "Loss": Loss(nll_loss)
}

# global criterion kwargs
criterion_kwargs = {"reduction": 'mean'}
# criterion_kwargs = {}

evaluator = create_supervised_evaluator(
    model,
    metrics=metrics,
    output_transform=lambda x, y, y_pred: {
        "x": x, "y": y, "y_pred": y_pred, "criterion_kwargs": criterion_kwargs}
)
data = [
    (torch.rand(4, 10), torch.randint(0, 3, size=(4, ))),
    (torch.rand(4, 10), torch.randint(0, 3, size=(4, ))),
    (torch.rand(4, 10), torch.randint(0, 3, size=(4, )))
]
res = evaluator.run(data)
```

As the `required_output_keys` contains `criterion_kwargs`, the user has to pass an empty dictionary for the original case with no `criterion_kwargs` for criterion.

vfdev-5:
@01-vyom thanks for the draft PR ! 
Let's now add some tests here:
- https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/test_loss.py
like here: 
- https://github.com/pytorch/ignite/blob/692487ce01ec2d212ca1693ebe53a46f255f5584/tests/ignite/metrics/test_metric.py#L98

Instead of `DummyMetric1`, we can create a metric derived from `Loss` and override update method to check passed args similarly to here: 
https://github.com/pytorch/ignite/blob/692487ce01ec2d212ca1693ebe53a46f255f5584/tests/ignite/metrics/test_metric.py#L28-L29

Please let me know if need more explanations.

01-vyom:
Ok, I will add a `test_output_mapping` test with a dummy `Loss` and an integration test similar to here:
https://github.com/pytorch/ignite/blob/692487ce01ec2d212ca1693ebe53a46f255f5584/tests/ignite/metrics/test_metric.py#L894

ignite/metrics/loss.py
```diff
@@ -33,28 +33,28 @@ class Loss(Metric):
...
             import torch.nn as nn
+            from torch.nn.functional import nll_loss
 
             from ignite.metrics import Accuracy, Loss
             from ignite.engine import create_supervised_evaluator
 
             model = ...
 
-            criterion = nn.NLLLoss()
+            criterion = nll_loss()
```
vfdev-5, line 57: I think this is incorrect, as `nll_loss()` is a function executed without any args and will give
```
TypeError: nll_loss() missing 2 required positional arguments: 'input' and 'target'
```

01-vyom: Yes, but I tried the above demo code [Passing criterion_kwargs] with nn.NLLLoss() and it gave the error that forward has no parameter "reduction". Any reason for this?  I will change the criterion to `nll_loss` from `nll_loss()`.

vfdev-5: yes, that's correct, we can do instead `criterion = nll_loss`.

Well, original thought was to provide additional kwargs to some custom loss functions that can require something else. For example a loss that computes on 
```
criterion(y_preds, y, image=x)
```
then we can pass `x` as `criterion_kwargs`. 
The idea is to show that there is no limitation for any kind of custom loss functions.

tests/ignite/metrics/test_loss.py
```diff
@@ -7,8 +8,83 @@
...
+        def update(self, output):
+            pass
+
+    loss_metric = DummyLoss(nll_loss)
+    assert loss_metric.required_output_keys is None
+    state = State(output=({"y1": 0, "y2": 1, "kwargs": {}}))
+    engine = MagicMock(state=state)
+
+    with pytest.raises(TypeError, match=r"Transformed engine output for DummyLoss metric should be a tuple/list"):
+        loss_metric.iteration_completed(engine)
```
vfdev-5, line 65: Thanks for updating the tests. I think we do not need those tests:
- test_output_as_mapping_wrong_keys
- test_output_as_mapping_keys_is_none
As it is already covered in test_metric.py...

Others look ok.

ignite/metrics/loss.py
01-vyom: Yes, looking at the whole codebase [at least the Loss, Metric, and Engine parts] I believe that we can achieve this idea very well without much hassle!!

tests/ignite/metrics/test_loss.py
01-vyom: Ok will make the changes

ignite/metrics/loss.py
```diff
@@ -30,9 +30,52 @@ class Loss(Metric):
...
+            criterion_kwargs = {...}
+
+            evaluator = create_supervised_evaluator(
+                model,
+                metrics=metrics,
+                output_transform=lambda x, y, y_pred: {
+                    "x": x, "y": y, "y_pred": y_pred, "criterion_kwargs": criterion_kwargs}
+            )
+
+            res = evaluator.run(data)
```
vfdev-5, line 74: @Priyansi could you please take a look and comment out if we need to improve this docstring ? Thanks

ignite/metrics/loss.py
```diff
@@ -30,9 +30,52 @@ class Loss(Metric):
...
+            }
+
+            # global criterion kwargs
+            criterion_kwargs = {...}
+
+            evaluator = create_supervised_evaluator(
+                model,
+                metrics=metrics,
+                output_transform=lambda x, y, y_pred: {
+                    "x": x, "y": y, "y_pred": y_pred, "criterion_kwargs": criterion_kwargs}
```
Priyansi, line 71: Here it isn't clear where the `x` is coming from since there is no mention of this variable in the above lines. Maybe we can include it in line 41 or just exclude this variable altogether?

ignite/metrics/loss.py
```diff
@@ -30,9 +30,52 @@ class Loss(Metric):
             metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
             non-blocking. By default, CPU.
 
+    Attributes:
+        required_output_keys: dictionary defines required keys to be found in ``engine.state.output`` if the
+            latter is a dictionary. Default, ``("y_pred", "y", "criterion_kwargs")``. This is useful when the
+            criterion function requires additional argument, which can be passed using ``criterion_kwargs``.
```
Priyansi, line 36: Typo: additional **arguments** since `criterion_kwargs` might contain more than one argument.

ignite/metrics/loss.py
01-vyom: This is the signature for create_supervised_evaluator of evaluate: 
https://github.com/pytorch/ignite/blob/dce97dfeacf1516e78b1cd9c6a37e57a1228ceed/ignite/engine/__init__.py#L518-L527

It contains lambda with 3 values `(x, y_pred, y)`, and by default, it takes 2 arguments `(y_pred, y)`. I can do something like this; refrain from passing x, but the lambda must contain x, whether it is used or not otherwise it will throw an error, code for this approach:

```python

evaluator = create_supervised_evaluator(
    model,
    metrics=metrics,
    output_transform=lambda x, y, y_pred: {
        "y": y, "y_pred": y_pred, "criterion_kwargs": criterion_kwargs}
)
```

Priyansi: Let's not remove `x` then since it won't make sense. We can just add `x` to line 41 like - requires `y_pred`, `y`, `x` and `criterion_kwargs`... 

As previously mentioned [here](https://pytorch.org/ignite/_modules/ignite/metrics/metric.html#Metric).

01-vyom: Ok

vfdev-5 (approved):
Thanks for the PR @01-vyom !

PR #2029, opened 2021-06-03
Fid Metric

gucifer:
Description:
First implementation and tests for the FID metric.
Check list:


gucifer:
Mypy error is cause scipy and torchvision do not exist in typeshed
typeshed in type hint database of mypy
https://stackoverflow.com/questions/60716482/error-skipping-analyzing-flask-mysqldb-found-module-but-no-type-hints-or-lib
What should I do?

gucifer:
> Mypy error is cause scipy and torchvision do not exist in typeshed
> typeshed in type hint database of mypy
> https://stackoverflow.com/questions/60716482/error-skipping-analyzing-flask-mysqldb-found-module-but-no-type-hints-or-lib
> What should I do?

nvm.... found the fix. Updated mypy.ini file.

sdesrozis:
I converted the PR to draft. Let’s implement the distributed case.

ignite/metrics/GAN/fid.py
```diff
@@ -0,0 +1,177 @@
...
+
+    In addition, a faster and online computation approach can be found in `Chen et al. 2014`__
+
+    __ https://arxiv.org/pdf/2009.14075.pdf
+
+    Remark:
+
+        This implementation is inspired by pytorch_fid package.
+
+    Args:
```
sdesrozis, line 72: `num_features` Is not documented

ignite/metrics/GAN/fid.py
```diff
@@ -0,0 +1,177 @@
...
+
+    Example:
+
+    .. code-block:: python
+        from ignite.metric.GAN import FID
+        import torch
+
+        train, test = torch.rand(10,2048), torch.rand(10,2048)
+
+        m = FID()
```
sdesrozis, line 91: This snippet doesn’t work

requirements-dev.txt
```diff
@@ -17,11 +19,13 @@ wandb
 mlflow
 neptune-client
 tensorboard
+torchvision
 pynvml
 clearml
 scikit-image
 py-rouge
 nltk
+pytorch_fid
```
sdesrozis, line 28:
```suggestion
```

tests/ignite/metrics/GAN/test_fid.py
```diff
@@ -0,0 +1,175 @@
...
+from ignite.metrics.GAN.fid import FID, InceptionExtractor, fid_score
+
+
+def test_fid_function():
+    train_samples, test_samples = torch.rand(10, 10), torch.rand(10, 10)
+
+    mu1, sigma1 = train_samples.mean(axis=0), cov(train_samples, rowvar=False)
+    mu2, sigma2 = test_samples.mean(axis=0), cov(test_samples, rowvar=False)
+
+    assert pytest.approx(
```
sdesrozis, line 18: Split this using temporary variables could be easier to read

tests/ignite/metrics/GAN/test_fid.py
```diff
@@ -0,0 +1,175 @@
...
+import pytest
+import pytorch_fid.fid_score as pytorch_fid_score
+import torch
+from numpy import cov
+
+import ignite.distributed as idist
+from ignite.metrics.GAN.fid import FID, InceptionExtractor, fid_score
+
+
+def test_fid_function():
```
sdesrozis, line 12:
```suggestion
def test_fid_score():
```

tests/ignite/metrics/GAN/test_fid.py
```diff
@@ -0,0 +1,175 @@
...
+    for cov1, cov2 in zip(sigma2, fid_sigma2):
+        assert torch.isclose(cov1.double(), cov2).all()
+
+
+def test_inception_extractor_wrong_inputs():
+    with pytest.raises(ValueError, match=r"Images should be of size 3x299x299 \(got torch.Size\(\[2, 2, 2, 0\]\)\)"):
+        InceptionExtractor()(torch.rand(2, 2, 2, 0))
+
+
+def _test_distrib_integration(device):
```
sdesrozis, line 75: This test is not distributed... have a look to others metric tests 

https://github.com/pytorch/ignite/blob/dce97dfeacf1516e78b1cd9c6a37e57a1228ceed/tests/ignite/metrics/test_precision.py#L402

requirements-dev.txt
gucifer: We are using this module for comparing our implementation. So we need it.

ignite/metrics/GAN/__init__.py
```diff
@@ -0,0 +1,5 @@
+from ignite.metrics.GAN.fid import FID
```
vfdev-5, line 1:
```suggestion
from ignite.metrics.gan.fid import FID
```

ignite/metrics/GAN/fid.py
```diff
@@ -0,0 +1,178 @@
+from typing import Callable, Sequence, Union
+
+import numpy as np
```
vfdev-5, line 3: Numpy and scipy are not dependencies for ignite ? Can we get rid of them ?

ignite/metrics/GAN/fid.py
```diff
@@ -0,0 +1,178 @@
...
+
+    __ https://arxiv.org/pdf/2009.14075.pdf
+
+    Remark:
+
+        This implementation is inspired by pytorch_fid package.
+
+    Args:
+        num_features: specifies number of features the evaluation samples should have
+        eps: small value added during matric division if denominator is 0.
```
vfdev-5, line 74: Do we really need to specify that as an argument ? Can't we just keep it internally ?

ignite/metrics/GAN/fid.py
gucifer: we need them for matrix square root operation.

ignite/metrics/GAN/fid.py
gucifer: It was previously done lazily but @sdesrozis and I think that we should have it more standardized.

ignite/metrics/GAN/fid.py
vfdev-5: you can use torch for that

ignite/metrics/GAN/fid.py
vfdev-5: My point is about why user should care about that ? Do we have a use-case where we need to set to `eps` another value ?
A standardized approach will be to have in somewhere in `ignite.eps` and reuse everywhere but not as class argument.

gucifer: My bad, I thought you were talking about num_features, we can have a global eps and just call ignite.eps. Point of eps is it should be negligible compared to the values of our data.

ignite/metrics/GAN/fid.py
gucifer: Torch doesn't have a function and writing one would be complicated for one. I can write it if you want but it might add a hundred or so lines of extra code.

vfdev-5: Well, your estimate of 100 lines seems too big.
- https://github.com/pytorch/pytorch/issues/25481
- https://github.com/scipy/scipy/blob/v1.6.3/scipy/linalg/_matfuncs_sqrtm.py#L114-L191

For now, let's keep scipy as dep but in this case we have to make it lazy instead of explicit import:
1) If no scipy => no FID metric import
2) user can import FID and have an issue on init if no scipy

ignite/metrics/GAN/fid.py
sdesrozis: I agree that `eps` is not relevant for the users. We should remove this parameter and use according to a default fixed value inside.

ignite/metrics/gan/fid.py
```diff
@@ -0,0 +1,177 @@
...
+
+    return diff.dot(diff) + np.trace(sigma1) + np.trace(sigma2) - 2 * tr_covmean
+
+
+class InceptionExtractor:
+    def __init__(self) -> None:
+        from torchvision import models
+
+        self.model = models.inception_v3(init_weights=True)
+        self.model.fc = torch.nn.Sequential()
```
vfdev-5, line 42:
```suggestion
        self.model.fc = torch.nn.Identity()
```
?
https://pytorch.org/docs/1.8.1/generated/torch.nn.Identity.html#torch.nn.Identity

ignite/metrics/gan/fid.py
```diff
@@ -0,0 +1,177 @@
...
+        self.model.eval()
+
+    def __call__(self, data: torch.Tensor) -> torch.Tensor:
```
vfdev-5, line 45: Shouldn't we do `with torch.no_grad()` ?

ignite/metrics/gan/fid.py
```diff
@@ -0,0 +1,177 @@
...
+
+    __ https://arxiv.org/pdf/1706.08500.pdf
+
+    In addition, a faster and online computation approach can be found in `Chen et al. 2014`__
+
+    __ https://arxiv.org/pdf/2009.14075.pdf
+
+    Remark:
+
+        This implementation is inspired by pytorch_fid package.
```
vfdev-5, line 70: We have to add a link to pytorch_fid

ignite/metrics/gan/fid.py
```diff
@@ -0,0 +1,177 @@
...
+        m.update((train,test))
+        print(m.compute())
+
+    .. versionadded:: 0.5.0
+    """
+
+    def __init__(
+        self,
+        num_features: int,
+        feature_extractor: Callable = lambda x: x,
```
vfdev-5, line 102: I think that default feature extractor should be inception v3

ignite/metrics/gan/fid.py
```diff
@@ -0,0 +1,177 @@
...
+        covmean = covmean.real
+
+    tr_covmean = np.trace(covmean)
+
+    return diff.dot(diff) + np.trace(sigma1) + np.trace(sigma2) - 2 * tr_covmean
+
+
+class InceptionExtractor:
+    def __init__(self) -> None:
+        from torchvision import models
```
vfdev-5, line 39: We should tnink better how to handle this dependency without making it as a hard requirement.

ignite/metrics/gan/fid.py
```diff
@@ -0,0 +1,177 @@
...
+
+__all__ = ["FID", "InceptionExtractor"]
+
+
+def fid_score(
+    mu1: torch.Tensor, mu2: torch.Tensor, sigma1: torch.Tensor, sigma2: torch.Tensor, eps: float = 1e-6
+) -> float:
+
+    diff = mu1 - mu2
+    import scipy
```
vfdev-5, line 17: This is very unfriendly way to do things. This method is called after 1 epoch done and user get ImportError after some computational time. We should avoid this way of coding things!

ignite/metrics/gan/fid.py
```diff
@@ -0,0 +1,177 @@
...
+            metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
+            non-blocking. By default, CPU.
+
+    Example:
+
+    .. code-block:: python
+        from ignite.metric.GAN import FID
+        import torch
+
+        train, test = torch.rand(10,2048), torch.rand(10,2048)
```
vfdev-5, line 90: Is it relevant as shape for the standard FID metric ?
Namings as train, test seem to be unconventional, we can put either put `y_pred`/`y` or `fake`/`real`

ignite/metrics/gan/fid.py
```diff
@@ -0,0 +1,177 @@
...
+class InceptionExtractor:
+    def __init__(self) -> None:
+        from torchvision import models
+
+        self.model = models.inception_v3(init_weights=True)
+        self.model.fc = torch.nn.Sequential()
+        self.model.eval()
+
+    def __call__(self, data: torch.Tensor) -> torch.Tensor:
+        if data.shape[1:] != (3, 299, 299):
```
vfdev-5, line 46: If inception model is headless (no fully-connected layer), can't we use any size input. In the metric code we just have to ensure that we always have the same size...

requirements-dev.txt
vfdev-5: You have it twice. See above.

ignite/metrics/gan/fid.py
gucifer: This just means that we have updated the mean and covariance based on all the training example given by user so far. Even if compute gives error as soon as user imports scipy they can rerun it and no computation done previously is lost.

gucifer: In the case where fid_score function is independently invoked, there is no extra computation so all cases are covered.

ignite/metrics/gan/fid.py
gucifer: FID is not just for image generation but also augmentation, style and feature transfer. In such cases real or fake is not appropriate. I will use `y_pred`/`y` instead.

ignite/metrics/gan/fid.py
gucifer: 299x299 is minimum size of image needed I can change it to >= 299 if needed

sdesrozis: > If inception model is headless (no fully-connected layer), can't we use any size input. 

That's right ! 

> In the metric code we just have to ensure that we always have the same size...

The user gives the feature size, so there is a test on shapes during iteration.

KickItLikeShika:
Friendly advice:
- It will be way better and you will save a lot of time for yourself if you ran the tests locally, for example:
```
pytest tests/ignite/metrics/gan/test_fid.py
```
- You had some issues with distributed tests before in this PR, as I was struggling a lot with DDP stuff (and still), but this page here I always use it as a reference, It will help you understand all the collective ops, try to have a look on it: https://pytorch.org/tutorials/intermediate/dist_tuto.html 
Hope my advice helps you.

ignite/metrics/gan/fid.py
sdesrozis: Please use exception and a user friendly message whether `scipy` is missing.

https://github.com/pytorch/ignite/blob/d16d15efbbbfc476702e91f3ab2bc757b839be26/ignite/contrib/metrics/cohen_kappa.py#L48

PR #2039, opened 2021-06-06
Fixes nightly CI failure

vfdev-5:
Description:
- Fixes test_auto.py::test_auto_methods_gloo for nightly where AssertionError is replaced by ValueError.

Check list:


KickItLikeShika:
I think we should restart GA workflow for Horovod and TPU tests

PR #2045, opened 2021-06-08
Handled different `engine.state.output` types for LRFinder

KickItLikeShika:
Handled tensors and floats from `engine.state.output` for LRFinder, and cover this with test.

Check list:

### DIFF
--- a/ignite/handlers/lr_finder.py
+++ b/ignite/handlers/lr_finder.py
@@ -137,6 +137,14 @@ class FastaiLRFinder:
     def _log_lr_and_loss(self, trainer: Engine, output_transform: Callable, smooth_f: float, diverge_th: float) -> None:
         output = trainer.state.output
         loss = output_transform(output)
+        if not isinstance(loss, float):
+            if isinstance(loss, torch.Tensor):
+                loss = loss.item()
+            else:
+                raise TypeError(
+                    "output of the engine should be of type float or torch.tensor, "
+                    f"but got output of type {type(loss).__name__}"
+                )
         loss = idist.all_reduce(loss)
         lr = self._lr_schedule.get_param()  # type: ignore[union-attr]
         self._history["lr"].append(lr)
--- a/ignite/handlers/lr_finder.py
+++ b/ignite/handlers/lr_finder.py
@@ -138,11 +138,14 @@ class FastaiLRFinder:
         output = trainer.state.output
         loss = output_transform(output)
         if not isinstance(loss, float):
-            if isinstance(loss, torch.Tensor) and (loss.ndimension() == 0):
+            if isinstance(loss, torch.Tensor) and (
+                (loss.ndimension() == 0) or (loss.ndimension() == 1 and len(loss) == 1)
+            ):
                 loss = loss.item()
             else:
                 raise TypeError(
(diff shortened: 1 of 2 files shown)


KickItLikeShika:
Mypy checks work fine locally, maybe we should restart the GA workflow

ignite/handlers/lr_finder.py
```diff
@@ -137,6 +137,14 @@ def _reset(self, trainer: Engine) -> None:
     def _log_lr_and_loss(self, trainer: Engine, output_transform: Callable, smooth_f: float, diverge_th: float) -> None:
         output = trainer.state.output
         loss = output_transform(output)
+        if not isinstance(loss, float):
+            if isinstance(loss, torch.Tensor):
```
vfdev-5, line 141: We have to check that tensor is 0-d or 1-d with one-element before applying `.item()`.

KickItLikeShika: I missed that, thanks for pointing that out!

ignite/handlers/lr_finder.py
```diff
@@ -137,6 +137,14 @@ def _reset(self, trainer: Engine) -> None:
     def _log_lr_and_loss(self, trainer: Engine, output_transform: Callable, smooth_f: float, diverge_th: float) -> None:
         output = trainer.state.output
         loss = output_transform(output)
+        if not isinstance(loss, float):
+            if isinstance(loss, torch.Tensor) and (loss.ndimension() == 0):
```
vfdev-5, line 141: You are still missing 1-d with one-element case

ignite/handlers/lr_finder.py
```diff
@@ -137,6 +137,17 @@ def _reset(self, trainer: Engine) -> None:
...
+        if not isinstance(loss, float):
+            if isinstance(loss, torch.Tensor) and (
+                (loss.ndimension() == 0) or (loss.ndimension() == 1 and len(loss) == 1)
+            ):
+                loss = loss.item()
+            else:
+                raise TypeError(
+                    "output of the engine should be of type float or 0d torch.Tensor "
+                    "or 1d torch.Tensor with 1 element, "
+                    f"but got output of type {type(loss).__name__}"
```
vfdev-5, line 149: This will show "Tensor" but I think we already know that loss is tensor, the point is to show shape maybe here.

ignite/handlers/lr_finder.py
```diff
@@ -137,6 +137,17 @@ def _reset(self, trainer: Engine) -> None:
     def _log_lr_and_loss(self, trainer: Engine, output_transform: Callable, smooth_f: float, diverge_th: float) -> None:
         output = trainer.state.output
         loss = output_transform(output)
+        if not isinstance(loss, float):
+            if isinstance(loss, torch.Tensor) and (
+                (loss.ndimension() == 0) or (loss.ndimension() == 1 and len(loss) == 1)
+            ):
+                loss = loss.item()
+            else:
```
vfdev-5, line 145: This looks a bit bulky but I have no idea on how to simplify it more than one line.

vfdev-5 (commented):
Thanks for the update @KickItLikeShika !

vfdev-5 (approved):
Thanks @KickItLikeShika !

PR #2047, opened 2021-06-09
Added mypy old version as dependency

gucifer:
Temporary Fixes #{2046}
Description:
Installs old mypy version as dependency to solve the errors.


vfdev-5 (approved):
Thanks @gucifer

PR #2052, opened 2021-06-11
add LRFinder to CIFAR10 Example

### LINKED ISSUE #2004: Add LRFinder to CIFAR10 example
We already have an example for LRFinder: https://github.com/pytorch/ignite/blob/master/examples/notebooks/FastaiLRFinder_MNIST.ipynb , However we'd like to add LRFinder to another example. 
So we want to update this example: https://github.com/pytorch/ignite/tree/master/examples/contrib/cifar10 . 
The idea is to add an option as `with_lrfinder` and if true, setup and execute LRFinder and `apply_suggested_lr` method.

Ishan-Kumar2:
Fixes #2004
Description:
Added LRFinder to CIFAR10 Example.

(PS. apart from adding the LRFinder the rest of the changes are done by running black on the file.)

Check list:

### DIFF
--- a/examples/contrib/cifar10/main.py
+++ b/examples/contrib/cifar10/main.py
@@ -16,6 +16,7 @@ from ignite.engine import Engine, Events
 from ignite.handlers import Checkpoint, DiskSaver, global_step_from_engine
 from ignite.metrics import Accuracy, Loss
 from ignite.utils import manual_seed, setup_logger
+from ignite.handlers import FastaiLRFinder
 
 
 def training(local_rank, config):
@@ -76,15 +77,20 @@ def training(local_rank, config):
     trainer = create_trainer(model, optimizer, criterion, lr_scheduler, train_loader.sampler, config, logger)
 
     # Let's now setup evaluator engine to perform model's validation and compute metrics
-    metrics = {
-        "Accuracy": Accuracy(),
-        "Loss": Loss(criterion),
-    }
+    metrics = {"Accuracy": Accuracy(), "Loss": Loss(criterion)}
 
     # We define two evaluators as they wont have exactly similar roles:
     # - `evaluator` will save the best model based on validation score
     evaluator = create_evaluator(model, metrics=metrics, config=config)
     train_evaluator = create_evaluator(model, metrics=metrics, config=config)
+
+    # Finding optimum Learning Rate using FastaiLRFinder
+    if config["lr_finder"] is True:
+        lr_finder = FastaiLRFinder()
+        to_save = {"model": model, "optimizer": optimizer}
+        with lr_finder.attach(trainer, to_save, diverge_th=5.0) as trainer_with_lr_finder:
+            trainer_with_lr_finder.run(train_loader)
+            lr_finder.apply_suggested_lr(optimizer)
 
     def run_validation(engine):
         epoch = trainer.state.epoch
@@ -156,6 +162,7 @@ def run(
     stop_iteration=None,
     with_clearml=False,
     with_amp=False,
+    lr_finder=False,
     **spawn_kwargs,
 ):
     """Main entry to train an model on CIFAR10 dataset.
@@ -216,11 +223,11 @@ def get_dataflow(config):
 
     # Setup data loader also adapted to distributed config: nccl, gloo, xla-tpu
     train_loader = idist.auto_dataloader(
-        train_dataset, batch_size=config["batch_size"], num_workers=config["num_workers"], shuffle=True, drop_last=True,
+        train_dataset, batch_size=config["batch_size"], num_workers=config["num_workers"], shuffle=True, drop_last=True
     )
 
     test_loader = idist.auto_dataloader(
-        test_dataset, batch_size=2 * config["batch_size"], num_workers=config["num_workers"], shuffle=False,
+        test_dataset, batch_size=2 * config["batch_size"], num_workers=config["num_workers"], shuffle=False
     )
     return train_loader, test_loader
 
@@ -317,17 +324,13 @@ def create_trainer(model, optimizer, criterion, lr_scheduler, train_sampler, con
         scaler.step(optimizer)
         scaler.update()
 
-        return {
-            "batch loss": loss.item(),
-        }
+        return {"batch loss": loss.item()}
 
     trainer = Engine(train_step)
     trainer.logger = logger
 
     to_save = {"trainer": trainer, "model": model, "optimizer": optimizer, "lr_scheduler": lr_scheduler}
-    metric_names = [
-        "batch loss",
-    ]
+    metric_names = ["batch loss"]
 
     common.setup_common_training_handlers(
         trainer=trainer,


Ishan-Kumar2:
Also I noticed that the `num_warmup_epochs` is set by default to 4. 
So if you keep that to default and change the `num_epochs` to something < 4, the `PiecewizeLinear` scheduler [[here]](https://github.com/pytorch/ignite/blob/86cec1165441a945c93c938c23544ac52a17a170/examples/contrib/cifar10/main.py#L249)   throws an error as it should. The error message doesnt say the reason for this (`num_epochs` >= `num_warmup_epochs`).
Should I add a simple check?

vfdev-5:
@Ishan-Kumar2 thanks for the PR, I let @KickItLikeShika to review it first.

> Also I noticed that the `num_warmup_epochs` is set by default to 4.
> So if you keep that to default and change the `num_epochs` to something < 4, the `PiecewizeLinear` scheduler [[here]](https://github.com/pytorch/ignite/blob/86cec1165441a945c93c938c23544ac52a17a170/examples/contrib/cifar10/main.py#L249) throws an error as it should. The error message doesnt say the reason for this (`num_epochs` >= `num_warmup_epochs`).
> Should I add a simple check?

Yes, you are right about that. We can add this check. On the other hand it does not make much sense to train on <4 epochs except of debuging the code :)

Ishan-Kumar2:
haha you are right @vfdev-5 training for <4 epochs doesnt make sense, but lets help out the debuggers :)

examples/contrib/cifar10/main.py
```diff
@@ -76,16 +76,21 @@ def training(local_rank, config):
...
     evaluator = create_evaluator(model, metrics=metrics, config=config)
     train_evaluator = create_evaluator(model, metrics=metrics, config=config)
 
+    # Finding optimum Learning Rate using FastaiLRFinder
+    if config["lr_finder"] is True:
+        lr_finder = FastaiLRFinder()
+        to_save = {"model": model, "optimizer": optimizer}
+        with lr_finder.attach(trainer, to_save, diverge_th=5.0) as trainer_with_lr_finder:
+            trainer_with_lr_finder.run(train_loader)
+            lr_finder.apply_suggested_lr(optimizer)
```
vfdev-5, line 92: This should be outside the context manager.

Ishan-Kumar2: fixed!

examples/contrib/cifar10/main.py
```diff
@@ -76,16 +76,21 @@ def training(local_rank, config):
...
     evaluator = create_evaluator(model, metrics=metrics, config=config)
     train_evaluator = create_evaluator(model, metrics=metrics, config=config)
 
+    # Finding optimum Learning Rate using FastaiLRFinder
+    if config["lr_finder"] is True:
+        lr_finder = FastaiLRFinder()
+        to_save = {"model": model, "optimizer": optimizer}
+        with lr_finder.attach(trainer, to_save, diverge_th=5.0) as trainer_with_lr_finder:
+            trainer_with_lr_finder.run(train_loader)
+        lr_finder.apply_suggested_lr(optimizer)
```
KickItLikeShika, line 92: I think putting some logs here about finding the optimal learning rate would be nice

examples/contrib/cifar10/main.py
```diff
@@ -76,16 +76,21 @@ def training(local_rank, config):
     trainer = create_trainer(model, optimizer, criterion, lr_scheduler, train_loader.sampler, config, logger)
```
KickItLikeShika, line 76: I hope you tried to run the example, because I think it won't work because of the `train_step` function, the return of `train_step` is a dict, and that's not supported for now with the LRFinder, the output must be of type float, and we are extending that in this PR by supporting tensors #2045 , so this must be edited https://github.com/pytorch/ignite/blob/ba8c63c358c96a5bdce428b708e7934c8e543a65/examples/contrib/cifar10/main.py#L329 
Check this method for more details https://github.com/pytorch/ignite/blob/master/ignite/handlers/lr_finder.py#L137

examples/contrib/cifar10/main.py
```diff
@@ -76,16 +76,21 @@ def training(local_rank, config):
...
     # We define two evaluators as they wont have exactly similar roles:
     # - `evaluator` will save the best model based on validation score
     evaluator = create_evaluator(model, metrics=metrics, config=config)
     train_evaluator = create_evaluator(model, metrics=metrics, config=config)
 
+    # Finding optimum Learning Rate using FastaiLRFinder
+    if config["lr_finder"] is True:
+        lr_finder = FastaiLRFinder()
+        to_save = {"model": model, "optimizer": optimizer}
+        with lr_finder.attach(trainer, to_save, diverge_th=5.0) as trainer_with_lr_finder:
```
KickItLikeShika, line 90: `diverge_th` is 5.0 by default https://github.com/pytorch/ignite/blob/86cec1165441a945c93c938c23544ac52a17a170/ignite/handlers/lr_finder.py#L352

KickItLikeShika (commented):
Thanks for the PR @Ishan-Kumar2, and sorry for not reviewing it quickly. Please try to run the example first, before committing the changes.

examples/contrib/cifar10/main.py
vfdev-5: @KickItLikeShika I think we can use `output_transform` to transform dict to float, see https://github.com/pytorch/ignite/blob/05f2b558f6e10bd8b05b6157684c392e65473638/ignite/handlers/lr_finder.py#L370
We can keep the output as it is now and add `output_transform` to `lr_finder.attach`

examples/contrib/cifar10/main.py
```diff
@@ -76,16 +76,21 @@ def training(local_rank, config):
...
-    }
+    metrics = {"Accuracy": Accuracy(), "Loss": Loss(criterion)}
 
     # We define two evaluators as they wont have exactly similar roles:
     # - `evaluator` will save the best model based on validation score
     evaluator = create_evaluator(model, metrics=metrics, config=config)
     train_evaluator = create_evaluator(model, metrics=metrics, config=config)
 
+    # Finding optimum Learning Rate using FastaiLRFinder
+    if config["lr_finder"] is True:
```
vfdev-5, line 87: Let's call the cmd argument as "use_lr_finder" instead of "lr_finder"

examples/contrib/cifar10/main.py
KickItLikeShika: Nice idea, thanks @vfdev-5

Ishan-Kumar2: Great, thanks for the input. Should I add the `output_transform` in the `utils.py` or in `main` only?

examples/contrib/cifar10/main.py
Ishan-Kumar2: fixed

examples/contrib/cifar10/main.py
Ishan-Kumar2: Sure, I have made the change.

examples/contrib/cifar10/main.py
Ishan-Kumar2: Yeah makes sense, I have added.

examples/contrib/cifar10/main.py
Ishan-Kumar2: I have added it in `main.py` for now, let me know if it has to be moved.

examples/contrib/cifar10/main.py
vfdev-5: @Ishan-Kumar2 I think we can just do : `output_transform=lambda output: output["batch loss"]` as we know what returns trainer. There is no need to do something generic.

Ishan-Kumar2: @vfdev-5 Okay, I have done that.

KickItLikeShika (approved):
Thanks for the updates @Ishan-Kumar2, LGTM!

examples/contrib/cifar10/main.py
```diff
@@ -76,16 +76,24 @@ def training(local_rank, config):
...
-    }
+    metrics = {"Accuracy": Accuracy(), "Loss": Loss(criterion)}
 
     # We define two evaluators as they wont have exactly similar roles:
     # - `evaluator` will save the best model based on validation score
     evaluator = create_evaluator(model, metrics=metrics, config=config)
     train_evaluator = create_evaluator(model, metrics=metrics, config=config)
 
+    # Finding optimum Learning Rate using FastaiLRFinder
+    if config["use_lr_finder"] is True:
```
vfdev-5, line 87:
```suggestion
    if config["use_lr_finder"]:
```

vfdev-5 (approved):
Thanks for the PR @Ishan-Kumar2 ! Looks good

Ishan-Kumar2:
Thanks a lot for the help and reviews @vfdev-5 and @KickItLikeShika. Excited to contribute more on ignite!

KickItLikeShika:
Thanks for you contributions @Ishan-Kumar2, kindly check your Discord please!

PR #2055, opened 2021-06-14
Added support for list of tensors as metric input 

### LINKED ISSUE #2013: Support list of tensors in Metrics
## 🚀 Feature

Related to discussion from here: https://github.com/Project-MONAI/MONAI/issues/2220#issuecomment-849250700

The feature request is to support list of tensors or scalars as input for metrics:
a) classification problem:
```python
acc = Accuracy()
output = (
    [torch.rand(10), torch.rand(10), torch.rand(10)],  # predictions
    [4, 5, 1],  # ground truth
)
acc.update(output)
print(acc.compute())
```

b) segmentation problem:
```python
acc = Accuracy()
output = (
    [torch.rand(10, 32, 32), torch.rand(10, 40, 30), torch.rand(10, 30, 24)],  # predictions
    [torch.randint(0, 10, size=(32, 32)), torch.randint(0, 10, size=(40, 30)),  torch.randint(0, 10, size=(30, 24))],  # ground truth
)
acc.update(output)
print(acc.compute())
```

c) Same for API with Engine.

cc @Nic-Ma

vfdev-5:
Fixes #2013

Description:
- Added support for list of tensors as metric input

Check list:

cc @Nic-Ma for visibility

### DIFF
--- a/ignite/metrics/metric.py
+++ b/ignite/metrics/metric.py
@@ -2,7 +2,8 @@ import warnings
 from abc import ABCMeta, abstractmethod
 from collections.abc import Mapping
 from functools import wraps
-from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Tuple, Union
+from numbers import Number
+from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Sequence, Tuple, Union
 
 import torch
 
@@ -279,6 +280,14 @@ class Metric(metaclass=ABCMeta):
         """Helper method to update metric's computation. It is automatically attached to the
         `engine` with :meth:`~ignite.metrics.metric.Metric.attach`.
 
+        Note:
+            ``engine.state.output`` is used to compute metric values.
+            The majority of implemented metrics accepts the following formats for ``engine.state.output``:
+            ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y}``. ``y_pred`` and ``y`` can be torch tensors or
+            list of tensors/numbers if applicable.
+
+
+
         Args:
             engine: the engine to which the metric must be attached
         """
@@ -296,7 +305,20 @@ class Metric(metaclass=ABCMeta):
                     f"it should contain {self.required_output_keys} keys, but given {list(output.keys())}"
                 )
             output = tuple(output[k] for k in self.required_output_keys)
-        self.update(output)
+
+        if isinstance(output, Sequence) and all([_is_list_of_tensors_or_numbers(o) for o in output]):
+            if not (len(output) == 2 and len(output[0]) == len(output[1])):
+                raise ValueError(
+                    f"Output should have 2 items of the same length, "
+                    f"got {len(output)} and {len(output[0])}, {len(output[1])}"
+                )
+            for o1, o2 in zip(output[0], output[1]):
+                # o1 and o2 are list of tensors or numbers
+                tensor_o1 = _to_batched_tensor(o1)
+                tensor_o2 = _to_batched_tensor(o2, device=tensor_o1.device)
+                self.update((tensor_o1, tensor_o2))
+        else:
+            self.update(output)
 
     def completed(self, engine: Engine, name: str) -> None:
         """Helper method to compute metric's value and put into the engine. It is automatically attached to the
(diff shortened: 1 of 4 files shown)


Nic-Ma (approved):

ignite/metrics/metric.py
```diff
@@ -296,7 +305,20 @@ def iteration_completed(self, engine: Engine) -> None:
...
+        if isinstance(output, Sequence) and all([_is_list_of_tensors_or_numbers(o) for o in output]):
+            if not (len(output) == 2 and len(output[0]) == len(output[1])):
+                raise ValueError(
+                    f"Output should have 2 items of the same length, "
+                    f"got {len(output)} and {len(output[0])}, {len(output[1])}"
+                )
+            for o1, o2 in zip(output[0], output[1]):
+                # o1 and o2 are list of tensors or numbers
+                tensor_o1 = _to_batched_tensor(o1)
+                tensor_o2 = _to_batched_tensor(o2, device=tensor_o1.device)
```
sdesrozis, line 318: Should we move the output to a device or let this be done in the training step function ?

sdesrozis:
Just one remark but it looks good !

ignite/metrics/metric.py
vfdev-5: here output is something coming from training step and in this case, o2 is a number that have to be wrapped with a tensor and put to appropriate device...

sdesrozis: My bad !!

sdesrozis (approved):
@vfdev-5 Ok !

Nic-Ma:
Hi @vfdev-5 @sdesrozis ,

Thanks for the quick PR, looks very good to me.
We also added support for `list of tensors` to MONAI last week: https://github.com/Project-MONAI/MONAI/blob/dev/monai/metrics/metric.py#L61
I have a quick question about the metrics support for **dict**:
Is it possible to make the global variable `required_output_keys` of ignite Metric to be an arg of `Metric.__init__()`?
We already received much feedback that our workflow programs have too many lambda functions and hard to understand. I am trying to reduce the `output_transform= lambda x: ...` in the code. If we can explicitly specify the `keys` of `y_pred` and `y`, maybe it's easier to use and understand?
CC @wyli

Thanks.

vfdev-5:
Hi @Nic-Ma , 

> Is it possible to make the global variable required_output_keys of ignite Metric to be an arg of Metric.__init__()?

thanks for the feedback, `required_output_keys` is a class attribute and I think it would be better to set `required_output_keys` globally instead of for each metric: 
```python
from ignite.metrics import Accuracy, Metric

Metric.required_output_keys = ("my_y_pred", "my_y_true")

acc = Accuracy()
acc.required_output_keys

engine = Engine(lambda e, b: None)
engine.state.output = {"my_y_pred": torch.rand(4, 10), "my_y_true": torch.randint(0, 10, size=(4, ))}

acc.iteration_completed(engine)
acc.compute()
> 0.125
```
what do you think ?

Nic-Ma:
Hi @vfdev-5 ,

Thanks for your proposal and sample code!
It can solve our problem, but maybe a little bit "hacking", especially when setting ignite args in the **JSON config** of NVIDIA Clara-Train SDK.
Our use case is something like:
```
"trainer": {
    "name": "SupervisedTrainer",
    "args": {
        ... ...
    }
},
post_transforms: [
    {
        "name": "CopyItemsD":
        "args": {"keys": "pred", "times": 1, "names": ["pred_inverted"]}
    },
    {
        "name": "InvertD",
        "args": {"keys": "pred_inverted"}
    },
    {
        "name": "SplitChannelD",
        "args": {"keys": "pred_inverted", "output_postfixes": ["background", "class1", "class2"]}
    }
]
"metrics": [
    {
        "name": "Accuracy",  # ignite metric
        "args": {"output_keys": ["pred", "label"]}
    },
    {
        "name": "MeanDice",  # subclass of ignite metric
        "args": {"output_keys": ["pred_inverted", "label_inverted"]}
    },
    {
        "name": "MeanDice",  # metric on class 1
        "args": {"output_keys": ["pred_inverted_class1", "label_inverted_class1"]}
    },
    {
        "name": "MeanDice",  # metric on class 2
        "args": {"output_keys": ["pred_inverted_class2", "label_inverted_class2"]}
    }
]
```
What do you think is the best way to support it?

Thanks.

vfdev-5:
Thanks for the details, Nic ! That's definitely requires `output_transform` to handle each case. Let me think about that.

> We already received much feedback that our workflow programs have too many lambda functions and hard to understand. I am trying to reduce the output_transform= lambda x: ... in the code.

In this case, maybe, it is possible to introduce one function to deal with those required for metrics like:
```python
def from_dict(*keys):
    def wrapper(output: dict):
        return tuple(output[k] for k in keys)
    return wrapper

acc = Accuracy(output_transform=from_dict("pred_inverted", "label_inverted"))
```
what do you think ?

Nic-Ma:
OK, thanks for the solution.
Let me try to add this API in MONAI utilities.

Thanks.

Nic-Ma:
Hi @vfdev-5 ,

BTW, may I know when do you plan to release ignite v0.4.5?

Thanks.

vfdev-5:
Hi Nic,
It should be done by this friday.

Nic-Ma:
Sounds great!
Thanks for letting me know.

PR #2070, opened 2021-06-21
Refactor stable release workflow [skip netlify]

trsvchn:
Fixes

Description:

This PR will split GH Action workflow for stable releases into 2 parallel workflows to increase "fault-tolerance" of the workflow:

- build  and publish on PyPI  -> trigger Docker builds and publish
- build  and publish on Anaconda

To 

Check list:

### DIFF
--- a/.github/workflows/stable-release-anaconda.yml
+++ b/.github/workflows/stable-release-anaconda.yml
@@ -28,7 +28,7 @@ jobs:
         shell: bash -l {0}
         env:
           ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
-          UPLOAD_USER: "pytorch"
+          UPLOAD_USER: ${{ secrets.UPLOAD_USER }}
         run: |
           chmod +x ./conda.recipe/build_and_upload.sh
           ./conda.recipe/build_and_upload.sh
--- a/.github/workflows/stable-release-anaconda.yml
+++ b/.github/workflows/stable-release-anaconda.yml
@@ -20,8 +20,6 @@ jobs:
         shell: bash -l {0}
         run: |
           conda install -y pytorch torchvision cpuonly -c pytorch
-          pip install -r requirements-dev.txt
-          pip install --upgrade --no-cache-dir twine
           python setup.py install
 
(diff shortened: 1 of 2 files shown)


trsvchn:
Can we just split it like that? Or job step with pypi builds requires "conda builds" step?
@vfdev-5 @ydcjeff

ydcjeff:
Is there a desired step by step workflow?

.github/workflows/stable-release-anaconda.yml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          miniconda-version: "latest"
+          python-version: 3.8
+
+      - name: Install dependencies
+        shell: bash -l {0}
+        run: |
+          conda install -y pytorch torchvision cpuonly -c pytorch
+          pip install -r requirements-dev.txt
+          pip install --upgrade --no-cache-dir twine
```
vfdev-5, line 24: no need to install twine

vfdev-5:
> Can we just split it like that? Or job step with pypi builds requires "conda builds" step?
> @vfdev-5 @ydcjeff

no pypi does not require conda build

vfdev-5:
> Is there a desired step by step workflow?

The context is that I was thinking that we could also change our release workflow and do pypi release and then upload remaining conda binary manually. For example to v0.4.5 we can already publish to pypi and manually do the upload for conda. So, in general if one job is failing we can still release a part ...

.github/workflows/stable-release-anaconda.yml
ydcjeff: Are those 2 `pip install` steps required?

vfdev-5: Good question, need to check if we need to run `pip install -r requirements-dev.txt`

vfdev-5: @trsvchn can you check that please ?

trsvchn: ok

vfdev-5 (approved):
Thanks @trsvchn !

Issue #2072, opened 2021-06-22
Bug Definition MBConv Class - EfficientNet_Cifar100_finetuning.ipynb


log-layer:
Hi, I think a found a bug in one of the PyTorch examples:

## 🐛 Bug description

In the iPython notebook "EfficientNet_Cifar100_finetuning.ipynb", there is a bug in the definition of the MBConv class:

In the third line after the sub-section `# Add identity skip`, the function `_drop_connect` is called without assigning the returned values to x. The code simply says `self._drop_connect(x)` instead of `x = self._drop_connect(x)`. The final operation performed inside `_drop_connect` does not occur in-place :)

sdesrozis:
@log-layer Thank you for pointing out this !!

What do you think about submitting a PR to correct this error ? It would be a great help.

log-layer:
@sdesrozis Thank you for the feedback, I just created a pull request. I hope that everything is ok, this is my first time performing a contribution.

PR #2073, opened 2021-06-22
Bug Fix MBConv Class - EffNet_Cifar100_finetuning

log-layer:
Fixes https://github.com/pytorch/ignite/issues/2072

🐛 Bug description
In the iPython notebook "EfficientNet_Cifar100_finetuning.ipynb", there is a bug in the definition of the MBConv class:

In the third line after the sub-section # Add identity skip, the function _drop_connect is called without assigning the returned values to x. The code simply says self._drop_connect(x) instead of x = self._drop_connect(x). The final operation performed inside _drop_connect does not occur in-place.

### DIFF
--- a/examples/notebooks/EfficientNet_Cifar100_finetuning.ipynb
+++ b/examples/notebooks/EfficientNet_Cifar100_finetuning.ipynb
@@ -351,7 +351,7 @@
     "        # Add identity skip\n",
     "        if x.shape == z.shape and self.with_skip:            \n",
     "            if self.training and self.drop_connect_rate is not None:\n",
-    "                self._drop_connect(x)\n",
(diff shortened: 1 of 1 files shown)


vfdev-5:
@log-layer thanks for the PR and nice catch !

Could you please also replace the following to make the code a bit nicer:
```diff
- self.drop_connect_rate = torch.tensor(drop_connect_rate, requires_grad=False)
+ self.drop_connect_rate = drop_connect_rate
```

log-layer:
@vfdev-5 Thank you for the feedback. I implemented the change that you suggested, and provided a small explanation.

vfdev-5 (approved):
LGTM! Thanks @log-layer

PR #2075, opened 2021-06-22
Update CUDA driver before running GPU tests on win

trsvchn:
Fixes #{issue number}

Description:

Check list:


KickItLikeShika:
Well, now we are using the full script but it doesn't work too, should we leave a comment on the issue (https://github.com/pytorch/vision/issues/4006) and tell the guy our problem?

vfdev-5 (approved):
Thanks a lot @trsvchn !

PR #2077, opened 2021-06-24
Lost the character "f" in f-string

Juddd:
Honestly if there's a bug here, I don't know how you get your presentation images


Juddd:
fix a bug

vfdev-5 (approved):
Thanks alot for catching that @Juddd !
Seems like when we were removing `.format` we lost that.

vfdev-5:
> Honestly if there's a bug here, I don't know how you get your presentation images

Probably, they come from another script.

By the way, if you find other typos, please send other quick fixes like that 👍

Issue #1284, opened 2020-09-12
Ensure that all core and contrib metrics can work in DDP


vfdev-5:
## 🚀 Feature

Tracking issue. All core metrics are working in DDP, except special case #1283.
The idea is to check all core and contrib metrics if work in DDP and open new issues if we need to work on some metric.

vfdev-5:
Remaining metrics:
- [ ] https://github.com/pytorch/ignite/blob/master/ignite/contrib/metrics/regression/mean_error.py
- [ ] https://github.com/pytorch/ignite/blob/master/ignite/contrib/metrics/regression/geometric_mean_relative_absolute_error.py

sayantan1410:
@vfdev-5 Hey, just a small question, to make precision_recall_curve.py work in DDP,  to write the metric from scratch instead of using the one from sci-kit learn ? Or there is some other way of doing it ?

PR #2083, opened 2021-06-26
Improved Mean Error for DDP

KickItLikeShika:
Fixes part of #1284 

Made mean error metric compatible with DDP and add the necessary tests. And added an integration tests.

Check list:

### DIFF
--- a/ignite/contrib/metrics/regression/mean_error.py
+++ b/ignite/contrib/metrics/regression/mean_error.py
@@ -37,11 +37,11 @@ class MeanError(_BaseRegression):
 
     @reinit__is_reduced
     def reset(self) -> None:
-        self._sum_of_errors = 0.0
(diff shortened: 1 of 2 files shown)


vfdev-5:
@KickItLikeShika thanks for the PR !

vfdev-5 (approved):

vfdev-5:
@KickItLikeShika seems like GPU tests wont pass: https://app.circleci.com/pipelines/github/pytorch/ignite/1993/workflows/4c558b7a-dc4d-48fd-bc89-e8fda3404d23/jobs/6059
(this is the PR that will fix the issue with FID/IS that currently is failing and stopping other metric tests)

KickItLikeShika:
Thanks for reporting that, unfortunately after trying to avoid sending follow-up PRs, we need to send another one, and add tolerence for distributed integration test, I will send the PR in minutes

PR #2084, opened 2021-06-26
Data Type mismatch Fix

gucifer:
Description:
Fixed Data Type error in Nvidia Apex.
Check list:

### DIFF
--- a/ignite/metrics/gan/fid.py
+++ b/ignite/metrics/gan/fid.py
@@ -182,8 +182,9 @@ class FID(_BaseInceptionMetric):
     def update(self, output: Sequence[torch.Tensor]) -> None:
 
         train, test = output
-        train_features = self._extract_features(train)
-        test_features = self._extract_features(test)
+
+        train_features = self._extract_features(train).to(torch.float64)
+        test_features = self._extract_features(test).to(torch.float64)
 
         if train_features.shape[0] != test_features.shape[0] or train_features.shape[1] != test_features.shape[1]:
             raise ValueError(
--- a/ignite/metrics/gan/fid.py
+++ b/ignite/metrics/gan/fid.py
@@ -183,8 +183,8 @@ class FID(_BaseInceptionMetric):
 
         train, test = output
 
-        train_features = self._extract_features(train).to(torch.float64)
-        test_features = self._extract_features(test).to(torch.float64)
+        train_features = self._extract_features(train)
+        test_features = self._extract_features(test)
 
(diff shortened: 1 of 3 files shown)


vfdev-5:
@gucifer I think you can use `.to(tensor)` to convert to appropriate dtype and do it only in one place instead of two places...

sdesrozis:
@gucifer IMO data should be moved in `_extract_features()`, so it would be done once.

gucifer:
> @gucifer I think you can use `.to(tensor)` to convert to appropriate dtype and do it only in one place instead of two places...
> @gucifer IMO data should be moved in `_extract_features()`, so it would be done once.
I tried that by changing last line of `_extract_features` to `return outputs.to(torch.float64)` but it does not seem to work. Anything, that requires class to class communication even from `_BaseInceptionMetric` to `FID/IS` gives the same error.

sdesrozis:
> Anything, that requires class to class communication even from `_BaseInceptionMetric` to `FID/IS` gives the same error.

What do you mean by `class communication` ?

gucifer:
> > Anything, that requires class to class communication even from `_BaseInceptionMetric` to `FID/IS` gives the same error.
> 
> What do you mean saying "class communication" ? Could you provide a snippet to illustrate ?

```
class A:
    def __init__(self):
        self.a = 0
    def func(self):
        self.a += B().func()

class B:
    def __init__(self):
        self.b = 1
    def func(self):
        return self.b
```
What I mean is parameter passing from method of one class to another.

sdesrozis:
So weird... It looks magical and I suspect a misunderstanding somewhere.

ignite/metrics/gan/fid.py
```diff
@@ -182,8 +182,9 @@ def reset(self) -> None:
     def update(self, output: Sequence[torch.Tensor]) -> None:
 
         train, test = output
-        train_features = self._extract_features(train)
-        test_features = self._extract_features(test)
+
+        train_features = self._extract_features(train).to(torch.float64)
+        test_features = self._extract_features(test).to(torch.float64)
```
vfdev-5, line 187: remove that

ignite/metrics/gan/fid.py
```diff
@@ -182,6 +182,7 @@ def reset(self) -> None:
     def update(self, output: Sequence[torch.Tensor]) -> None:
 
         train, test = output
+
```
vfdev-5, line 185:
```suggestion
```

vfdev-5:
@gucifer I'm closing this PR in favor of #2085 that should check the code agains GPU CI.

PR #2091, opened 2021-06-30
Extended `MetricsLambda` to update its underlying metrics

### LINKED ISSUE #1921: Make `MetricsLambda` update it's underlying metrics
## 🚀 Feature

As title.

The current state is that we cannot use `MetricsLambda` as a standard metric because it doesn't update/reset it's underlying metrics.
Adding this will expand the scope and usability of `MetricsLambda`

I've already an idea how to implement this based on our discussion https://github.com/pytorch/ignite/pull/1887#discussion_r610902667

So @vfdev-5 please assign the issue to me.

KickItLikeShika:
Fixes #1921 

Implemented the update method to make metrics lambda able to update its underlying methods, and I separated the attach from reset/update/compute API, then implemented the necessary tests.

Check list:

### DIFF
--- a/ignite/metrics/metrics_lambda.py
+++ b/ignite/metrics/metrics_lambda.py
@@ -116,7 +116,7 @@ class MetricsLambda(Metric):
     def attach(self, engine: Engine, name: str, usage: Union[str, MetricUsage] = EpochWise()) -> None:
         if self._updated:
             raise ValueError(
-                "The underlying metrics are already updated, " "can't attach while using reset/update/compute API."
+                "The underlying metrics are already updated, can't attach while using reset/update/compute API."
             )
         usage = self._check_usage(usage)
         # recursively attach all its dependencies (partially)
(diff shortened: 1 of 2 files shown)


ignite/metrics/metrics_lambda.py
```diff
@@ -105,6 +114,10 @@ def _internal_attach(self, engine: Engine, usage: MetricUsage) -> None:
                     engine.add_event_handler(usage.ITERATION_COMPLETED, metric.iteration_completed)
 
     def attach(self, engine: Engine, name: str, usage: Union[str, MetricUsage] = EpochWise()) -> None:
+        if self._updated:
+            raise ValueError(
+                "The underlying metrics are already updated, " "can't attach while using reset/update/compute API."
```
vfdev-5, line 119:
```suggestion
                "The underlying metrics are already updated, can't attach while using reset/update/compute API."
```

vfdev-5 (commented):
Thanks a lot for the PR @KickItLikeShika

ignite/metrics/metrics_lambda.py
KickItLikeShika: `black` magic again :)))

vfdev-5 (approved):
LGTM, thanks a lot @KickItLikeShika !

PR #2092, opened 2021-06-30
Add hostlist expand as alternative to scontrol

### LINKED ISSUE #2051: Alternative to the command scontrol
## 🚀 Feature

Using `idist` on a slurm cluster relies on the command `scontrol` for selecting an hostname as `MASTER_ADDR`.

https://github.com/pytorch/ignite/blob/86cec1165441a945c93c938c23544ac52a17a170/ignite/distributed/comp_models/native.py#L258

As discussed in #2048, we should provide another way whether the command missing, for instance, using container launcher as `pyxis`.

sdesrozis:
Fixes #2051 

Description:

Add an alternative to scontrol for expanding hostname list used by SLURM.

```python
hostnames = expand_hostlist("quartz[4-8]")
print(hostnames)

> ['quartz4', 'quartz5', 'quartz6', 'quartz7', 'quartz8']
```

Based on https://github.com/LLNL/py-hostlist

Check list:

### DIFF
--- a/ignite/utils.py
+++ b/ignite/utils.py
@@ -2,9 +2,8 @@ import collections.abc as collections
 import functools
 import logging
 import random
-import re
 import warnings
-from typing import Any, Callable, Dict, List, Optional, TextIO, Tuple, Type, TypeVar, Union, cast
+from typing import Any, Callable, Dict, Optional, TextIO, Tuple, Type, TypeVar, Union, cast
 
 import torch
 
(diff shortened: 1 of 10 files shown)


ignite/utils.py
```diff
@@ -271,3 +272,70 @@ def wrapper(*args: Any, **kwargs: Dict[str, Any]) -> Callable:
         return cast(F, wrapper)
 
     return decorator
+
+
+def expand_hostlist(nodelist: str) -> List[str]:
```
vfdev-5, line 277: I'm not sure if we need this in `ignite.utils` and as a public method. Let's keep it near in `ignite/distributed/comp_models/native.py` and as a private method...

vfdev-5 (commented):
@sdesrozis thanks for the PR

ignite/utils.py
sdesrozis: Ok 👍🏻

vfdev-5 (approved):
Thanks @sdesrozis !

PR #2094, opened 2021-07-02
Docs improvement and fixing bug with FID/IS metrics

gucifer:
Description:
Added explicitly that torchvision and scipy are needed for FID and InceptionScore.
Found a bug in InceptionScore due to pre-trained model returning logits instead of probabilities.
Fixed it by applying softmax function to deduce probabilities.
Check list:

### DIFF
--- a/ignite/metrics/gan/fid.py
+++ b/ignite/metrics/gan/fid.py
@@ -56,6 +56,8 @@ class FID(_BaseInceptionMetric):
 
     where :math:`\mu_1` and :math:`\sigma_1` refer to the mean and covariance of the train data and
     :math:`\mu_2` and :math:`\sigma_2` refer to the mean and covariance of the test data.
+    If you wish to use the default Inception Model for evaluation, it requires torchvision module.
+    Also, FID uses scipy module for matrix square root calculation.
 
     More details can be found in `Heusel et al. 2002`__
 
--- a/ignite/metrics/gan/fid.py
+++ b/ignite/metrics/gan/fid.py
@@ -129,7 +129,7 @@ class FID(_BaseInceptionMetric):
 
         if num_features is None and feature_extractor is None:
             num_features = 1000
-            feature_extractor = InceptionModel(return_features=False)
+            feature_extractor = InceptionModel(return_features=False, device=device)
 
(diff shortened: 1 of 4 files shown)


ignite/metrics/gan/inception_score.py
```diff
@@ -25,6 +25,9 @@ class InceptionScore(_BaseInceptionMetric):
 
     __ https://arxiv.org/pdf/1801.01973.pdf
 
+    .. note::
+        The default Inception model requires the `torchvision` module to be installed.
+
```
sdesrozis, line 30:
```suggestion
```

ignite/metrics/gan/fid.py
```diff
@@ -71,6 +71,10 @@ class FID(_BaseInceptionMetric):
 
         __ https://github.com/mseitzer/pytorch-fid
 
+    .. note::
+        The default Inception model requires the `torchvision` module to be installed.
+        FID also requires `scipy` module for matrix square root calculations.
```
vfdev-5, line 76:
```suggestion
        FID also requires `scipy` library for matrix square root calculations.
```

ignite/metrics/gan/utils.py
```diff
@@ -23,6 +23,7 @@ def __init__(self, return_features: bool, device: Union[str, torch.device] = "cp
             raise RuntimeError("This module requires torchvision to be installed.")
         super(InceptionModel, self).__init__()
         self._device = device
+        self.return_features = return_features
```
vfdev-5, line 26: Why do we need this as an attribute ?

gucifer: For the if else condition in the forward function

vfdev-5: We can append softmax layer once in the model definition.

gucifer: We can't just append to a sequential. We will have to create a new one.
If you are fine with that I will change the implementation that way.

sdesrozis (approved):
@gucifer  thank you ! LGTM !

vfdev-5:
@sdesrozis I let you merge it if OK

PR #2095, opened 2021-07-03
fix: add `py.typed` for type checkers

ydcjeff:
See [colab](https://colab.research.google.com/drive/1TjFg53UfdwR7__2qjRGCJ4Be0emwWsPT?usp=sharing).

Description:

Check list:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,7 @@ setup(
     license="BSD",
     # Package info
     packages=find_packages(exclude=("tests", "tests.*",)),
+    package_data={"ignite": ["py.typed"]},
     zip_safe=True,
     install_requires=requirements,
 )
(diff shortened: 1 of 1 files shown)


vfdev-5:
For context:
- https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages
- https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/

ydcjeff:
> Can you also add a test step similar to the one you have in the colab ?

I think this is not needed once we confirm it is working.

vfdev-5:
> > Can you also add a test step similar to the one you have in the colab ?
> 
> I think this is not needed once we confirm it is working.

There will be also failure for other domain libs like torchvision. OK for keeping this PR without test step.

Saw that MONAI does the same: https://github.com/Project-MONAI/MONAI/pull/626

vfdev-5 (approved):
Thanks @ydcjeff !

PR #2096, opened 2021-07-03
Fixed `GMRAE` implementation + Improved for DDP

### LINKED ISSUE #2087: Improve GMARE metric computation formula
Months ago I tried to improve GMARE to be compatible with DDP, but I had a very strange issue, and the distributed tests were failing, I thought I did something wrong (even I updated +10 metrics for DDP then) so I ignored it at that time, today I was trying to update it and send the PR but I had the same issue, the tests distributed tests are failing.
```
distributed_context_single_node_gloo = {'local_rank': 1, 'rank': 1, 'world_size': 2}

    @pytest.mark.distributed
    def test_distrib_gloo_cpu_or_gpu(distributed_context_single_node_gloo):
    
        device = idist.device()
>       _test_distrib_compute(device)


metric_device = device(type='cpu')

    def _test(metric_device):
        metric_device = torch.device(metric_device)
        m = GeometricMeanRelativeAbsoluteError(device=metric_device)
        torch.manual_seed(10 + rank)
    
        y_pred = torch.rand(size=(100,), device=device)
        y = torch.rand(size=(100,), device=device)
    
        m.update((y_pred, y))
    
        y = idist.all_gather(y)
    
        np_y = y.cpu().numpy()
        np_y_pred = y_pred.cpu().numpy()
    
        np_gmrae = np.exp(np.log(np.abs(np_y - np_y_pred) / np.abs(np_y - np_y.mean())).mean())
    
        # sum_y = 0
        # num_examples = 0
        # sum_of_errors = 0
        # np_gmrae = 0
        # sum_y += np_y.sum()
        # num_examples += np_y.shape[0]
        # y_mean = sum_y / num_examples
        # numerator = np.abs(y.view_as(y_pred) - y_pred)
        # denominator = np.abs(y.view_as(y_pred) - y_mean)
        # sum_of_errors += np.log(numerator / denominator).sum()
        # np_gmrae += np.exp((sum_of_errors / num_examples).mean())
    
>       assert m.compute() == pytest.approx(np_gmrae)
E       assert 1.3195266723632812 == 1.3320785760879517 ± 1.3e-06
E         +1.3195266723632812
E         -1.3320785760879517 ± 1.3e-06


FAILED tests/ignite/contrib/metrics/regression/test_geometric_mean_relative_absolute_error.py::test_distrib_gloo_cpu_or_gpu - assert 1.3195266723632812 == 1.3320785760879517 ± 1.3e-06
FAILED tests/ignite/contrib/metrics/regression/test_geometric_mean_relative_absolute_error.py::test_distrib_gloo_cpu_or_gpu - assert 1.3195266723632812 == 1.3320785760879517 ± 1.3e-06
```
And here is the code and the tests 
https://github.com/pytorch/ignite/compare/master...KickItLikeShika:improve-GMRAE-ddp

KickItLikeShika:
Fixes #2087, #1284 

- Fixed GMRAE implementation to be somehow similar to EpochMetric, all the `predictions` and `targets` gathered first then compute the GMRAE on all the gathered values to avoid the misleading results because of computing the mean over a single batch.
- Updated the integration tests.
- Made the metric compatible with DDP and added the necessary tests.

Check list:

### DIFF
--- a/ignite/contrib/metrics/regression/geometric_mean_relative_absolute_error.py
+++ b/ignite/contrib/metrics/regression/geometric_mean_relative_absolute_error.py
@@ -1,10 +1,11 @@
-from typing import Tuple
+from typing import Tuple, cast
 
 import torch
 
+import ignite.distributed as idist
 from ignite.contrib.metrics.regression._base import _BaseRegression
 from ignite.exceptions import NotComputableError
-from ignite.metrics.metric import reinit__is_reduced, sync_all_reduce
+from ignite.metrics.metric import reinit__is_reduced
 
 
 class GeometricMeanRelativeAbsoluteError(_BaseRegression):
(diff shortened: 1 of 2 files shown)


ignite/contrib/metrics/regression/geometric_mean_relative_absolute_error.py
```diff
@@ -34,23 +36,47 @@ class GeometricMeanRelativeAbsoluteError(_BaseRegression):
...
                 "GeometricMeanRelativeAbsoluteError must have at least one example before it can be computed."
             )
-        return torch.exp(torch.mean(cast(torch.Tensor, self._sum_of_errors) / self._num_examples)).item()
+
+        _prediction_tensor = torch.cat(self._predictions, dim=0)
+        _target_tensor = torch.cat(self._targets, dim=0)
+
+        ws = idist.get_world_size()
+
+        if ws > 1:
```
sdesrozis, line 64: I was wondering if all procs have the tensors, why let only proc 0 does the computation and add an extra communication. Why not duplicate the calculation over all procs and avoid the communication L80 ?

sdesrozis:
@KickItLikeShika Thank you, it looks good to me.

So finally, \bar{A} means the mean of A in the formula.

ignite/contrib/metrics/regression/geometric_mean_relative_absolute_error.py
```diff
@@ -34,23 +36,47 @@ class GeometricMeanRelativeAbsoluteError(_BaseRegression):
...
+            # All gather across all processes
+            _prediction_tensor = cast(torch.Tensor, idist.all_gather(_prediction_tensor))
+            _target_tensor = cast(torch.Tensor, idist.all_gather(_target_tensor))
+
+        result = 0.0
+        if idist.get_rank() == 0:
+            # Run compute_fn on zero rank only
+            result = torch.exp(
+                torch.log(
```
sdesrozis, line 73: Could you update the docstring to explain what means \bar{A} in the formula ?

it would be necessary to add a warning in docstring about computing cost and memory consumption of this metric.

ignite/contrib/metrics/regression/geometric_mean_relative_absolute_error.py
KickItLikeShika: I see your point, I don't actually what would be better between these two implementations, but my point here I wanted to keep the implementation similar to something that we already have (as almost the rest of the metric here has the same implementation), I wanted this to be almost the same as `EpochMetric`, I don't know what would be better here, please tell me what do you think?

vfdev-5: In case of EpochMetric we execute a custom function on cpu (think of a sklearn metric) and it doesn't make much sense to execute this function on all processes (especially if there are any parallelism in the implementation).

KickItLikeShika: Thanks for explanation. OK, I will change the implementation

sdesrozis: There could have been a compromise between communication and calculation. Anyway, it was a simple remark.

KickItLikeShika: Thanks @sdesrozis for pointing that out!

sdesrozis: > Thanks for explanation. OK, I will change the implementation

vfdev-5 (approved):

PR #2090, opened 2021-06-30
Any parameter scheduler

fco-dv:
Fixes #1913

Description:
Enlarge the scope of `ParamScheduler` like it is done in MONAI to schedule any hyperparameter while ensuring backward compatibility.

Check list:

- [x] New tests are added (if a new feature is added)
- [x] New doc strings: description and/or example code are in RST format
- [x] Documentation is updated (if required)

PR #2098, opened 2021-07-04
add EMA handler

### LINKED ISSUE #2097: Exponential moving average (EMA) handler
## 🚀 Feature
As discussed in the MONAI's issue thread, it would be great if EMA handler can be added to the ignite's handlers family.
EMA aims to keep an exponential moving averaged version of model weights. This smoothed version of weights are used for validation and inference.

Context: https://github.com/Project-MONAI/MONAI/issues/2514

sandylaker:
Fixes #2097 

Description:

Check list:

### DIFF
--- a/ignite/contrib/handlers/ema_handler.py
+++ b/ignite/contrib/handlers/ema_handler.py
@@ -1,6 +1,5 @@
-import warnings
 from copy import deepcopy
-from typing import Any, Optional, OrderedDict, Union
+from typing import Any, Dict, Optional, Union
 
 import torch
 import torch.nn as nn
@@ -33,10 +32,6 @@ class EMAHandler:
     - The handler allows for linearly warmup the momentum in the beginning when training process is not stable.
     - The handler supports saving the EMA backup to checkpoint via ``state_dict``, and loading from the checkpoint via
       ``load_state_dict``.
-    - In the end of each train epoch, the handler will swap weights with the model, so that the EMA weights
-      will be moved to the model and used for validation. In the beginning of each epoch, the handler will againswap
-      weights with the model, so that the EMA weights are moved back to the EMA handlers, and model weights are moved
-      back to the model.
 
 
     Note:
@@ -46,30 +41,6 @@ class EMAHandler:
           2. Build ``trainer`` (``ignite.engine.Engine``).
           3. Resume from checkpoint for ``model`` and ``ema_handler``.
           4. Attach ``ema_handler`` to ``trainer``.
-          5. Attach other handlers to ``trainer``.
-
-    Note:
-          When the trainer completes running, the EMA weights are moved to the model. If both the handler's
-          ``state_dict`` and model's ``state_dict`` are saved by ``Checkpoint`` handler, users can retrieve the EMA
-          weights as follows:
-
-          .. code-block:: python
-
-              to_load = {"model" model, "ema_handler", ema_handler}
-              ckpt = torch.load("checkpoint.pt")
-              Checkpoint.load_objects(to_load, ckpt)
-
-              # EMA weights are actually in the model's state_dict
-              smoothed_weights = model.state_dict()
-
-          Consider resuming training from a saved checkpoint: although the EMA weights are in the model when loaded
-          from a checkpoint, there is no need to manually swap the ``state_dict`` of handler and model. The Handler
-          will automatically perform this step when the training resumes.
-
-    Note:
-          Users need to register the EMA handler before the validation handler, so that swapping weights will be
-          performed before validation begins. It is recommended to register the EMA handler as the first handler for
-          the event ``Events.EPOCH_COMPLETED``.
 
     Examples:
           .. code-block:: python
@@ -78,8 +49,8 @@ class EMAHandler:
               model = nn.Linear(2, 1).to(device)
               # update the ema every 5 iterations
               ema_handler = EMAHandler(
-                model, momentum=0.0002, momentum_warmup=0.0001, interval=5, warmup_iters=10000, device=device)
-              trainer = Engine(step_fn)
+                  model, momentum=0.0002, momentum_warmup=0.0001, interval=5, warmup_iters=10000, device=device)
+              trainer = Engine(train_step_fn)
               to_load = {"model": model, "ema_handler", ema_handler, "trainer", trainer}
               if resume_from is not None:
                   Checkpoint.load_objects(to_load, checkpoint=resume_from)
@@ -89,6 +60,14 @@ class EMAHandler:
               to_save = to_load
               ckpt_handler = Checkpoint(to_save, DiskSaver(...), ...)
               trainer.add_event_handler(Events.EPOCH_COMPLETED, ckpt_handler)
+
+              # use ema model for validation
+              val_step_fn = get_val_step_fn(ema_handler.get_ema_model())
+              evaluator = Engine(val_step_fn)
+
+              @trainer.on(Events.EPOCH_COMPLETED)
+              def run_validation(engine):
+                  engine.run(val_data_loader)
 
               trainer.run(...)
 
@@ -128,14 +107,26 @@ class EMAHandler:
         self.warmup_iters = warmup_iters
         self.device = device
 
-        self.model = self._unwrap_model(model)
-        self.ema = deepcopy(self.model)
+        self.unwrapped_model = self._unwrap_model(model)
+        self.ema = deepcopy(self.unwrapped_model)
         self.ema.eval()
         if self.device is not None:
             self.ema.to(device=device)  # type: ignore
 
-    def state_dict(self, **kwargs: Any) -> OrderedDict[str, Tensor]:
-        """Return ``state_dict`` of the EMA copy.
+    def get_ema_model(self) -> nn.Module:
+        """Get the model with EMA weights.
+
+        Note:
+              This function returns an instance of ``nn.Module`` rather than ``DistributedDataParallel`` or
+              ``DataParallel``.
+
+        Returns:
+
+        """
+        return self.ema
+
+    def state_dict(self, **kwargs: Any) -> Dict[str, Tensor]:
+        """Return ``state_dict`` of the EMA model. It is a shortcut for ``get_ema_model().state_dict()``.
 
         Args:
             kwargs: arguments of ``nn.Module.state_dict``.
@@ -143,15 +134,15 @@ class EMAHandler:
         """
         return self.ema.state_dict(**kwargs)
 
-    def load_state_dict(self, state_dict: OrderedDict[str, Tensor], strict: bool = True) -> None:
-        """Load the ``sate_dict`` to the EMA copy.
+    def load_state_dict(self, state_dict: Dict[str, Tensor], strict: bool = True) -> None:
+        """Load the ``sate_dict`` to the EMA model. It is a shortcut for ``get_ema_model().load_state_dict()``.
 
         Args:
             state_dict: dict contains the weights.
             strict: whether to strictly enforce that the keys match.
 
         """
-        self.ema.load_state_dict(state_dict, strict=strict)
+        self.ema.load_state_dict(state_dict, strict=strict)  # type: ignore
 
     @staticmethod
     def _unwrap_model(model: nn.Module) -> nn.Module:
@@ -169,20 +160,11 @@ class EMAHandler:
         return min(self.momentum, momentum)
 
     @torch.no_grad()
-    def _swap_params(self, engine: Engine) -> None:
-        for ema_v, model_v in zip(self.ema.state_dict().values(), self.model.state_dict().values()):
-            if self.device is not None:
-                model_v = model_v.to(device=self.device)
-            tmp = model_v.data.clone()
-            model_v.data.copy_(ema_v.data)
-            ema_v.data.copy_(tmp)
-
-    @torch.no_grad()
     def _update(self, engine: Engine) -> None:
         curr_iter = engine.state.iteration
         momentum = self._get_momentum(curr_iter)
 
-        for ema_v, model_v in zip(self.ema.state_dict().values(), self.model.state_dict().values()):
+        for ema_v, model_v in zip(self.ema.state_dict().values(), self.unwrapped_model.state_dict().values()):
             if self.device is not None:
                 model_v = model_v.to(device=self.device)
             ema_v.mul_(1 - momentum).add_(model_v.data, alpha=momentum)
@@ -194,16 +176,4 @@ class EMAHandler:
             engine: Trainer to which the handler will be attached.
 
         """
-        # swap the parameters at the end of each epoch. If using Events.EPOCH_COMPLETED here, the handler of swap
-        # parameters might be called after validation handler.
-        # TODO Find a better hack to ensure that handler of swapping parameter is called before validation.
-        num_handlers_epoch_completed = engine._event_handlers[Events.EPOCH_COMPLETED]
-        if len(num_handlers_epoch_completed) > 0:
-            warnings.warn(
-                f"engine already has {num_handlers_epoch_completed} handlers at Events.EPOCH_COMPLETED, "
-                f"please manually check if the EMA handler is attached to the engine before other "
-                f"handlers"
-            )
-        engine.add_event_handler(Events.EPOCH_COMPLETED, self._swap_params)
-        engine.add_event_handler(Events.EPOCH_STARTED, self._swap_params)
         engine.add_event_handler(Events.ITERATION_COMPLETED(every=self.interval), self._update)
--- a/ignite/contrib/handlers/ema_handler.py
+++ b/ignite/contrib/handlers/ema_handler.py
@@ -63,7 +63,7 @@ class EMAHandler:
               smoothed_weights = model.state_dict()
 
           Consider resuming training from a saved checkpoint: although the EMA weights are in the model when loaded
-          from a checkpoint, there is no need to manually swap the ``sate_dict`` of handler and model. The Handler
+          from a checkpoint, there is no need to manually swap the ``state_dict`` of handler and model. The Handler
           will automatically perform this step when the training resumes.
 
     Note:
@@ -83,6 +83,7 @@ class EMAHandler:
               to_load = {"model": model, "ema_handler", ema_handler, "trainer", trainer}
               if resume_from is not None:
                   Checkpoint.load_objects(to_load, checkpoint=resume_from)
+              ema_handler.attach(trainer)
 
               # add other handlers
               to_save = to_load
@@ -103,24 +104,37 @@ class EMAHandler:
         warmup_iters: int = 100,
         device: Optional[Union[str, torch.device]] = None,
     ) -> None:
-        assert 0 < momentum < 1
-        assert 0 < momentum_warmup < 1
-        assert momentum_warmup <= momentum
-        assert isinstance(interval, int) and interval > 0
-        assert isinstance(warmup_iters, int) and warmup_iters > 0
-        assert isinstance(model, nn.Module)
+        if not 0 < momentum < 1:
+            raise ValueError(f"Invalid momentum: {momentum}")
+        if not 0 < momentum_warmup < 1:
+            raise ValueError(f"Invalid momentum_warmup: {momentum_warmup}")
+        if not momentum_warmup <= momentum:
+            raise ValueError(
+                f"momentum_warmup should be less than or equal to momentum, but got "
+                f"momentum_warmup: {momentum_warmup} and momentum: {momentum}"
+            )
+        if not (isinstance(interval, int) and interval > 0):
+            raise ValueError(f"Invalid interval: {interval}")
+        if not isinstance(warmup_iters, int) and warmup_iters > 0:
+            raise ValueError(f"Invalid warmup_iters: {warmup_iters}")
+        if not isinstance(model, nn.Module):
+            raise ValueError(
+                f"model should be an instance of nn.Module or its subclasses, but got"
+                f"model: {model.__class__.__name__}"
+            )
         self.momentum = momentum
         self.interval = interval
         self.momentum_warmup = momentum_warmup
         self.warmup_iters = warmup_iters
         self.device = device
 
-        self.ema = deepcopy(self._unwrap_model(model))
+        self.model = self._unwrap_model(model)
+        self.ema = deepcopy(self.model)
         self.ema.eval()
         if self.device is not None:
-            self.ema.to(device=device)
+            self.ema.to(device=device)  # type: ignore
 
-    def state_dict(self, **kwargs: Optional[Any]) -> OrderedDict[str, Tensor]:
+    def state_dict(self, **kwargs: Any) -> OrderedDict[str, Tensor]:
         """Return ``state_dict`` of the EMA copy.
 
         Args:
@@ -146,7 +160,7 @@ class EMAHandler:
         else:
             return model
 
-    def _get_momentum(self, curr_iter: int):
+    def _get_momentum(self, curr_iter: int) -> float:
         """Get current momentum, `curr_iter` should be 1-based. When `curr_iter = 1`, `momentum =
         self.momentum_warmup`; when `curr_iter >= self.warmup_iters`, `momentum = self.momentum`"""
         assert curr_iter >= 1
@@ -155,10 +169,8 @@ class EMAHandler:
         return min(self.momentum, momentum)
 
     @torch.no_grad()
-    def _swap_params(self, engine: Engine, model: nn.Module) -> None:
-
-        model = self._unwrap_model(model)
-        for ema_v, model_v in zip(self.ema.state_dict().values(), model.state_dict().values()):
+    def _swap_params(self, engine: Engine) -> None:
+        for ema_v, model_v in zip(self.ema.state_dict().values(), self.model.state_dict().values()):
             if self.device is not None:
                 model_v = model_v.to(device=self.device)
             tmp = model_v.data.clone()
@@ -166,27 +178,20 @@ class EMAHandler:
             ema_v.data.copy_(tmp)
 
     @torch.no_grad()
-    def _update(self, engine: Engine, model: nn.Module) -> None:
-
-        model = self._unwrap_model(model)
+    def _update(self, engine: Engine) -> None:
         curr_iter = engine.state.iteration
         momentum = self._get_momentum(curr_iter)
 
-        for ema_v, model_v in zip(self.ema.state_dict().values(), model.state_dict().values()):
+        for ema_v, model_v in zip(self.ema.state_dict().values(), self.model.state_dict().values()):
             if self.device is not None:
                 model_v = model_v.to(device=self.device)
             ema_v.mul_(1 - momentum).add_(model_v.data, alpha=momentum)
 
-    def attach(self, engine: Engine, model: nn.Module) -> None:
+    def attach(self, engine: Engine) -> None:
         """Attach the handler to engine.
-
-        Note:
-              If model is ``DataParallel`` or ``DistributedDataParallel``, the EMA smoothing will be applied to
-              ``model.module`` .
 
         Args:
             engine: Trainer to which the handler will be attached.
-            model: Model to which the EMA smoothing will be applied.
 
         """
         # swap the parameters at the end of each epoch. If using Events.EPOCH_COMPLETED here, the handler of swap
@@ -199,6 +204,6 @@ class EMAHandler:
                 f"please manually check if the EMA handler is attached to the engine before other "
                 f"handlers"
             )
-        engine.add_event_handler(Events.EPOCH_COMPLETED, self._swap_params, model)
-        engine.add_event_handler(Events.EPOCH_STARTED, self._swap_params, model)
-        engine.add_event_handler(Events.ITERATION_COMPLETED(every=self.interval), self._update, model)
+        engine.add_event_handler(Events.EPOCH_COMPLETED, self._swap_params)
+        engine.add_event_handler(Events.EPOCH_STARTED, self._swap_params)
+        engine.add_event_handler(Events.ITERATION_COMPLETED(every=self.interval), self._update)
--- a/ignite/contrib/handlers/ema_handler.py
+++ b/ignite/contrib/handlers/ema_handler.py
@@ -141,12 +141,10 @@ class EMAHandler:
 
     @staticmethod
     def _unwrap_model(model: nn.Module) -> nn.Module:
-        if isinstance(model, nn.Module):
-            return model
-        elif isinstance(model, (nn.parallel.DataParallel, nn.parallel.DistributedDataParallel)):
+        if isinstance(model, (nn.parallel.DataParallel, nn.parallel.DistributedDataParallel)):
             return model.module
         else:
-            raise ValueError("Invalid model class")
+            return model
 
     def _get_momentum(self, curr_iter: int):
         """Get current momentum, `curr_iter` should be 1-based. When `curr_iter = 1`, `momentum =
--- a/ignite/contrib/handlers/ema_handler.py
+++ b/ignite/contrib/handlers/ema_handler.py
@@ -1,0 +1,206 @@
+import warnings
+from copy import deepcopy
+from typing import Any, Optional, OrderedDict, Union
+
+import torch
+import torch.nn as nn
+from torch import Tensor
+
+from ignite.engine import Engine, Events
+
+__all__ = ["EMAHandler"]
+
+
+class EMAHandler:
+    r"""Exponential moving average (EMA) handler can be used to compute a EMA smoothed version of model weights.
+    The EMA weights are updated as follows:
+
+    .. math:: \theta_{\text{EMA}, t+1} = (1 - \lambda) \cdot \theta_{\text{EMA}, t} + \lambda \cdot \theta_{t}
+
+    where :math:`\theta_{\text{EMA}, t}` and :math:`\theta_{t}` are the EMA weights and model weights at
+    :math:`t`-th iteration, respectively; :math:`\lambda` is the update momentum.
+
+    Args:
+          model: the model for which EMA weights will be computed. If ``model`` is ``DataParallel`` or
+              ``DistributedDataParallel``, the EMA smoothing will be applied to ``model.module`` .
+          momentum: the update momentum, should be float in range :math:`\left(0, 1 \right)`.
+          momentum_warmup: the initial update momentum, the value should be smaller than ``momentum``. Momentum will
+              increase from this value to ``momentum`` linearly.
+          interval: update interval (in iterations).
+          warmup_iters: iterations of warmup.
+          device: device to store the EMA weights.
+
+    - The handler allows for linearly warmup the momentum in the beginning when training process is not stable.
+    - The handler supports saving the EMA backup to checkpoint via ``state_dict``, and loading from the checkpoint via
+      ``load_state_dict``.
+    - In the end of each train epoch, the handler will swap weights with the model, so that the EMA weights
+      will be moved to the model and used for validation. In the beginning of each epoch, the handler will againswap
+      weights with the model, so that the EMA weights are moved back to the EMA handlers, and model weights are moved
+      back to the model.
+
+
+    Note:
+          It is recommended to initialize and use an EMA handler in following order:
+
+          1. Initialize ``model`` (``nn.Module`` or ``DistributedDataParallel``) and ``ema_handler`` (``EMAHandler``).
+          2. Build ``trainer`` (``ignite.engine.Engine``).
+          3. Resume from checkpoint for ``model`` and ``ema_handler``.
+          4. Attach ``ema_handler`` to ``trainer``.
+          5. Attach other handlers to ``trainer``.
+
+    Note:
+          When the trainer completes running, the EMA weights are moved to the model. If both the handler's
+          ``state_dict`` and model's ``state_dict`` are saved by ``Checkpoint`` handler, users can retrieve the EMA
+          weights as follows:
+
+          .. code-block:: python
+
+              to_load = {"model" model, "ema_handler", ema_handler}
+              ckpt = torch.load("checkpoint.pt")
+              Checkpoint.load_objects(to_load, ckpt)
+
+              # EMA weights are actually in the model's state_dict
+              smoothed_weights = model.state_dict()
+
+          Consider resuming training from a saved checkpoint: although the EMA weights are in the model when loaded
+          from a checkpoint, there is no need to manually swap the ``sate_dict`` of handler and model. The Handler
+          will automatically perform this step when the training resumes.
+
+    Note:
+          Users need to register the EMA handler before the validation handler, so that swapping weights will be
+          performed before validation begins. It is recommended to register the EMA handler as the first handler for
+          the event ``Events.EPOCH_COMPLETED``.
+
+    Examples:
+          .. code-block:: python
+
+              device = torch.device("cuda:0")
+              model = nn.Linear(2, 1).to(device)
+              # update the ema every 5 iterations
+              ema_handler = EMAHandler(
+                model, momentum=0.0002, momentum_warmup=0.0001, interval=5, warmup_iters=10000, device=device)
+              trainer = Engine(step_fn)
+              to_load = {"model": model, "ema_handler", ema_handler, "trainer", trainer}
+              if resume_from is not None:
+                  Checkpoint.load_objects(to_load, checkpoint=resume_from)
+
+              # add other handlers
+              to_save = to_load
+              ckpt_handler = Checkpoint(to_save, DiskSaver(...), ...)
+              trainer.add_event_handler(Events.EPOCH_COMPLETED, ckpt_handler)
+
+              trainer.run(...)
+
+
+    """
+
+    def __init__(
+        self,
+        model: nn.Module,
+        momentum: float = 0.0002,
+        momentum_warmup: float = 0.0001,
+        interval: int = 1,
+        warmup_iters: int = 100,
+        device: Optional[Union[str, torch.device]] = None,
+    ) -> None:
+        assert 0 < momentum < 1
+        assert 0 < momentum_warmup < 1
+        assert momentum_warmup <= momentum
+        assert isinstance(interval, int) and interval > 0
+        assert isinstance(warmup_iters, int) and warmup_iters > 0
+        assert isinstance(model, nn.Module)
+        self.momentum = momentum
+        self.interval = interval
+        self.momentum_warmup = momentum_warmup
+        self.warmup_iters = warmup_iters
+        self.device = device
+
+        self.ema = deepcopy(self._unwrap_model(model))
+        self.ema.eval()
+        if self.device is not None:
+            self.ema.to(device=device)
+
+    def state_dict(self, **kwargs: Optional[Any]) -> OrderedDict[str, Tensor]:
+        """Return ``state_dict`` of the EMA copy.
+
+        Args:
+            kwargs: arguments of ``nn.Module.state_dict``.
+
+        """
+        return self.ema.state_dict(**kwargs)
+
+    def load_state_dict(self, state_dict: OrderedDict[str, Tensor], strict: bool = True) -> None:
+        """Load the ``sate_dict`` to the EMA copy.
+
+        Args:
+            state_dict: dict contains the weights.
+            strict: whether to strictly enforce that the keys match.
+
+        """
+        self.ema.load_state_dict(state_dict, strict=strict)
+
+    @staticmethod
+    def _unwrap_model(model: nn.Module) -> nn.Module:
+        if isinstance(model, nn.Module):
+            return model
+        elif isinstance(model, (nn.parallel.DataParallel, nn.parallel.DistributedDataParallel)):
+            return model.module
+        else:
+            raise ValueError("Invalid model class")
+
+    def _get_momentum(self, curr_iter: int):
+        """Get current momentum, `curr_iter` should be 1-based. When `curr_iter = 1`, `momentum =
+        self.momentum_warmup`; when `curr_iter >= self.warmup_iters`, `momentum = self.momentum`"""
+        assert curr_iter >= 1
+        denominator = max(1, self.warmup_iters - 1)
+        momentum = self.momentum_warmup + (self.momentum - self.momentum_warmup) * (curr_iter - 1) / denominator
+        return min(self.momentum, momentum)
+
+    @torch.no_grad()
+    def _swap_params(self, engine: Engine, model: nn.Module) -> None:
+
+        model = self._unwrap_model(model)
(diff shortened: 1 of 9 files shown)


vfdev-5:
@sandylaker thanks a lot for the PR ! I'll check the details soon, now I have a bit more generic comment that we can put this handler into the core module (not contrib): `ignite.handlers`.

sandylaker:
@vfdev-5 Github shows that "First-time contributors need a maintainer to approve running workflows", so could you please approve it for this PR? Thank you very much.

ignite/contrib/handlers/ema_handler.py
```diff
@@ -0,0 +1,204 @@
...
+
+              to_load = {"model" model, "ema_handler", ema_handler}
+              ckpt = torch.load("checkpoint.pt")
+              Checkpoint.load_objects(to_load, ckpt)
+
+              # EMA weights are actually in the model's state_dict
+              smoothed_weights = model.state_dict()
+
+          Consider resuming training from a saved checkpoint: although the EMA weights are in the model when loaded
+          from a checkpoint, there is no need to manually swap the ``sate_dict`` of handler and model. The Handler
```
vfdev-5, line 66:
```suggestion
          from a checkpoint, there is no need to manually swap the ``state_dict`` of handler and model. The Handler
```

ignite/contrib/handlers/ema_handler.py
```diff
@@ -0,0 +1,204 @@
...
+        interval: int = 1,
+        warmup_iters: int = 100,
+        device: Optional[Union[str, torch.device]] = None,
+    ) -> None:
+        assert 0 < momentum < 1
+        assert 0 < momentum_warmup < 1
+        assert momentum_warmup <= momentum
+        assert isinstance(interval, int) and interval > 0
+        assert isinstance(warmup_iters, int) and warmup_iters > 0
+        assert isinstance(model, nn.Module)
```
vfdev-5, line 111: In our codebase we use `if-not-then-raise` instead of `assert`. Could you please change that.

ignite/contrib/handlers/ema_handler.py
```diff
@@ -0,0 +1,204 @@
...
+        model = self._unwrap_model(model)
+        curr_iter = engine.state.iteration
+        momentum = self._get_momentum(curr_iter)
+
+        for ema_v, model_v in zip(self.ema.state_dict().values(), model.state_dict().values()):
+            if self.device is not None:
+                model_v = model_v.to(device=self.device)
+            ema_v.mul_(1 - momentum).add_(model_v.data, alpha=momentum)
+
+    def attach(self, engine: Engine, model: nn.Module) -> None:
```
vfdev-5, line 180: I see that we ask the model to EMA twice: in the constructor and here. I think it would be better to do that once and maybe in the constructor...

ignite/contrib/handlers/ema_handler.py
```diff
@@ -0,0 +1,204 @@
...
+        # TODO Find a better hack to ensure that handler of swapping parameter is called before validation.
+        num_handlers_epoch_completed = engine._event_handlers[Events.EPOCH_COMPLETED]
+        if len(num_handlers_epoch_completed) > 0:
+            warnings.warn(
+                f"engine already has {num_handlers_epoch_completed} handlers at Events.EPOCH_COMPLETED, "
+                f"please manually check if the EMA handler is attached to the engine before other "
+                f"handlers"
+            )
+        engine.add_event_handler(Events.EPOCH_COMPLETED, self._swap_params, model)
+        engine.add_event_handler(Events.EPOCH_STARTED, self._swap_params, model)
```
vfdev-5, line 203: Could you please explain why we need to swap the params every epoch started and completed ? 
Wouldn't be better to expose ema model somehow to the user and use it for validation directly ?

ignite/contrib/handlers/ema_handler.py
sandylaker: Fixed in the next commit

ignite/contrib/handlers/ema_handler.py
sandylaker: Thanks. It is fixed in next commit

ignite/contrib/handlers/ema_handler.py
sandylaker: Yes, I updated it in the next commit.

ignite/contrib/handlers/ema_handler.py
sandylaker: Please consider the case when users use DDP training. If we let the users to use the EMA model for validation, they need to initialize an extra DDP model for multi-gpu validation, right? Will it increase unnecessary communication cost? Please correct me if I am wrong.

vfdev-5: Actually, DDP wrapping is only needed if model params require grad (see https://pytorch.org/docs/master/_modules/torch/nn/parallel/distributed.html#DistributedDataParallel). We can consider EMA model as one with params without grads and thus we do not need to wrap it with DDP...

sandylaker: @vfdev-5 Thanks for the explanation. I have removed the functions for swapping parameters, and expose the ema model with a function `get_ema_model`.

vfdev-5:
@sandylaker thanks for the update, let me check and think about the API we would like to provide. Meanwhile, could you please put the code into `ignite.handlers` instead of `ignite.contrib.handlers`. Thanks

ignite/contrib/handlers/ema_handler.py
```diff
@@ -0,0 +1,179 @@
...
+    def _unwrap_model(model: nn.Module) -> nn.Module:
+        if isinstance(model, (nn.parallel.DataParallel, nn.parallel.DistributedDataParallel)):
+            return model.module
+        else:
+            return model
+
+    def _get_momentum(self, curr_iter: int) -> float:
+        """Get current momentum, `curr_iter` should be 1-based. When `curr_iter = 1`, `momentum =
+        self.momentum_warmup`; when `curr_iter >= self.warmup_iters`, `momentum = self.momentum`"""
+        assert curr_iter >= 1
```
vfdev-5, line 157: Could you please replace this assert with if-raise

vfdev-5:
@sandylaker inspecting the example from the docs: https://deploy-preview-2098--pytorch-ignite-preview.netlify.app/generated/ignite.handlers.ema_handler.emahandler#ignite.handlers.ema_handler.EMAHandler I was thinking more about the current API and I was wondering if it wouldn't be better to expose explicitly EMA model as the following:
```python
model = MyPyTorchModel(...)
model = idist.auto_model(model)  # it can be now a DDP or DP or torch.nn.Module

# Here we explicitly create EMA model
ema_model = EMAHandler.create_ema(model)
# ema_model can be either DP or torch.nn.Module

# now while saving or loading we use only torch objects
to_save = to_load = {"model": model, "ema_model", ema_model, "trainer", trainer}

trainer = ...
ema_handler = EMAHandler(
    ema_model, momentum=0.0002, momentum_warmup=0.0001, interval=5, warmup_iters=10000
)
ema_handler.attach(trainer)
```

Few comments, 
1) if user still uses DP (instead of DDP), then ema can be also a DP.
2) I think we need explicitly put while creating the ema model:
```python
    for param in ema_model.parameters():
        param.detach_()
```
3) Could you please check if buffers are also synchronized between model and ema_model ? I think it should be the case, however I'd like to make sure.
More precisely, I'm talking about something like:
```python
    # set EMA model's BN buffers as base model BN buffers:
    for m1, m2 in zip(model.modules(), ema_model.modules()):
        if isinstance(m1, nn.BatchNorm2d) and isinstance(m2, nn.BatchNorm2d):
            m2.running_mean = m1.running_mean
            m2.running_var = m1.running_var
```


@sandylaker what do you think ?

sandylaker:
@vfdev-5 Please feel free to correct me:

1. `create_ema` can be a static or class method, it does the following job: 
```python3
@staticmethod
def create_ema(model):
    if isinstance(model, DDP):
        ema_model = deepcopy(model.module)
    else:
        ema_model = deepcopy(model)
    for p in ema_model.parameters():
        p.detach_()
```

2. The handler needs to track the online model. But with `ema_handler = EMAHandler(ema_model, ...)` we can only track the ema model.
3. How should we handle the `device` argument above? In the current version, the handler allows for putting the ema model on CPU while online model on GPU. This can save some GPU memory but be slower.
4. If using the API above, `EMAHandler.state_dict()` and `EMAHandler.load_state_dict()` are redundant, right?

5. After we determine the API, I will add tests for testing if the buffers are synchronized. I think the current version only supports  persistent buffers. I.e., the buffers that are included in `state_dict`.

vfdev-5:
@sandylaker thanks for the update !

> 1. create_ema can be a static or class method, it does the following job:

Yes, the method can be a staticmethod for `EMAHandler` class as you suggested.


> 2. The handler needs to track the online model. But with ema_handler = EMAHandler(ema_model, ...) we can only track the ema model.

Oh, yes, you are right, my bad. If we add model as arg in the constructor or attach, than we have to ensure that model and ema_model both have the same structure... Finally I'm not fan of my suggested API... 

> 3. How should we handle the device argument above? In the current version, the handler allows for putting the ema model on CPU while online model on GPU. This can save some GPU memory but be slower.

I omitted device as I think this can lead to some device mismatch issues. I was imagining using ema_model for validation but validation data is on GPU and ema model is stored on CPU. We have to move something to a common device...

> 4. If using the API above, EMAHandler.state_dict() and EMAHandler.load_state_dict() are redundant, right?

Yes, if ema_model is exposed directly, we do not need these methods.

> 5. After we determine the API, I will add tests for testing if the buffers are synchronized. I think the current version only supports persistent buffers. I.e., the buffers that are included in state_dict.

Sounds good !

So, thinking more about your API and the one I proposed yesterday, if I understand correctly, in your API, ema_model is seen as a storage only and user does everything with the model itself and can sync the params with ema_model (e.g. for validation). In my case, I saw ema_model as another model to run the validation with...
Both approaches seems interesting... 

@sdesrozis @trsvchn any thoughts ?

sandylaker:
@vfdev-5 

> I omitted device as I think this can lead to some device mismatch issues. I was imagining using ema_model for validation but validation data is on GPU and ema model is stored on CPU. We have to move something to a common device...

May be we shall decrease the API's flexibility by only allow the ema model  and online model to be on the same device? This is relatively easy since we use `deepcopy` in the implementation.

>So, thinking more about your API and the one I proposed yesterday, if I understand correctly, in your API, ema_model is seen as a storage only and user does everything with the model itself and can sync the params with ema_model (e.g. for validation). In my case, I saw ema_model as another model to run the validation with...

In the previous version with `_swap_params`, the ema_model is essentially a storage. But in the current version, I expose the model by the method `EMAHandler.get_ema_model()`, it will returned an unwrapped model (`nn.Module`) and this model can be used for validation.

vfdev-5:
To recap:

Approach 1: ema_model is storage only
Pros: 
- if stored on CPU, no additional GPU memory consumption
- `model` is the only object to manipulate

Cons:
- Need special care from user's side to call switch params between model and ema_model. Probably, we can not do that automatically in general case


Approach 2: ema_model is another model
Pros:
- clear split between model and ema_model as there are 2 objects to work with.

Cons:
- Will need additional GPU memory space

sandylaker:
@vfdev-5 , Thanks for the recap:

> Approach 1 Cons: Need special care from user's side to call switch params between model and ema_model. Probably, we can not do that automatically in general case.

The handler can take care of swapping params. So in the end there will be three sub-handlers registered.
```python3
    # move ema weights to model for validation
    engine.add_event_handler(Events.EOCH_COMPLETED, self._swap_params)
    # move ema weights back to the storage
    engine.add_event_handler(Events.EPOCH_STARTED, self._swap_params)
    # ema update
    engine.add_event_handler(Events.ITERATION_COMPLETED, self._update)

The issue with these sub-handlers is actually that the users need to attach the `EMAHandler` to the engine before validation handler. Otherwise the validation function will be called before `_swap_params`.

**Comment by vfdev-5** [comment]
@sandylaker yes, using `engine.add_event_handler(Events.EPOCH_STARTED, self._swap_params)` is not ideal as we can also imagine that user would like to validate the model after a number of iterations before the epoch has ended and in this case, he/she should somehow call the swap method manually.

**Comment by sdesrozis** [comment]
@sandylaker @vfdev-5 Thank you for this interesting discussion about this nice feature.

Finally, the swapping should be done before and after the use of model with ema parameters. I feel like python contexts should help. 

```python
model = ...

ema_handler = EMAHandler(model, momentum=0.0002, momentum_warmup=0.0001, interval=5, warmup_iters=10000)

ema_handler.attach(trainer)

...

using ema_handler.model() as ema_model:
    ...
```

Although it is not perfect at all, this could be dig further.

sandylaker:
@sdesrozis , Thank you for the suggestion. So the swapping can be implemented as follows:
```python3
class EMAHandler:
    def __init__(self, model, ...):
        # model can be nn.Module or DP or DDP
        self.model = model
        # unwrap the DP or DDP to nn.Module
        self.ema = deepcopy(self._unwrap(model))

    ...

    @contextlib.contextmanager
    def ema_model():
        # move ema weights to online model
        self._swap_params()
        try:
            # self.model is nn.Module or DP or DDP
            yield self.model
        finally:
            # move ema weights back to handler's storage
            self._swap_params()
```

So this context manager can be used as follows:
```python3
# start validation
with ema_handler.ema_model() as ema_model:
    for x, y in data_loader:
        pred = ema_model(x.to(device))
        acc = (pred.argmax() == y).sum() / y.shape[0]
        acc = idist.all_gather(acc)
        ...
```
But how can we use this context manager together with the `Engine` workflow? @vfdev-5 what do you think?

sdesrozis:
This could be inserted in `val_test_fn` mentioned in the docstring or maybe 
```python
def run_validation(engine):
    with ema_handler.get_model():
        engine.run(val_data_loader)
```

Using decoration could be an idea too. Anyway, I don't know yet if using it as context is really a good idea.

vfdev-5:
Using context manager, the API can be similar to https://github.com/fadel/pytorch_ema/#example

I think we can maintain the following API which is rather close this implementation:
```python
model = MyPyTorchModel(...)
model = idist.auto_model(model)  # it can be now a DDP or DP or torch.nn.Module

trainer = ...
ema_handler = EMAHandler(
    model, momentum=0.0002, momentum_warmup=0.0001, interval=5, warmup_iters=10000
)
# Let's expose ema_model as attribute instead of a method
ema_model = ema_handler.ema_model
# Thus we can remove also EMAHandler.load_state_dict/state_dict methods

ema_handler.attach(trainer)

# now while saving or loading we use only torch objects
to_save = to_load = {"model": model, "ema_model", ema_model, "trainer", trainer}

# Let's also append ema_handler.momentum into trainer.state.ema_momentum
# Such that we could also log it
def _():
    print(trainer.state.ema_momentum)
```
@sandylaker any comments, feedback or suggestions please ?

sandylaker:
@vfdev-5 I also prefer the Approach 2. So I updated the API as you suggested. I also append `ema_momentum` to `Engine.state`. In addition, I added an test to test if the buffers of `ema_model` are also synchronized. Please check it out.

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,155 @@
...
+        """Update weights of ema model"""
+        momentum = engine.state.ema_momentum
+
+        for ema_v, model_v in zip(self.ema_model.state_dict().values(), self.unwrapped_model.state_dict().values()):
+            ema_v.mul_(1.0 - momentum).add_(model_v.data, alpha=momentum)  # type: ignore
+
+    def _update_ema_momentum(self, engine: Engine) -> None:
+        """Update momentum in engine.state"""
+        curr_iter = engine.state.iteration
+        momentum = self._get_momentum(curr_iter)
```
vfdev-5, line 144: I think `self.momentum` and momentum = `engine.state.ema_momentum` can be misleading. I would also expect `self.momentum` to be current momentum value.

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def attach(self, engine: Engine) -> None:
+        """Attach the handler to engine.
+
+        Args:
+            engine: Trainer to which the handler will be attached.
+
+        """
+        engine.add_event_handler(Events.ITERATION_STARTED, self._update_ema_momentum)
+        engine.add_event_handler(Events.ITERATION_COMPLETED(every=self.interval), self._update_weights)
```
vfdev-5, line 155: Instead of hard-coding the event here as `Events.ITERATION_COMPLETED(every=self.interval)` we can add that as the argument to `attach`:
```python
    def attach(
        self, 
        engine: Engine, 
        event_name: Union[str, Events, CallableEventWithFilter, EventsList] = Events.ITERATION_COMPLETED
    ) -> None:
        """Attach the handler to engine.
        Args:
            engine: Trainer to which the handler will be attached.
        """
        engine.add_event_handler(event_name, self._update_ema_momentum)
        engine.add_event_handler(event_name, self._update_weights)
```

vfdev-5 (commented):
Thanks a lot for the updates @sandylaker !
I left few other comments on improving the codebase and docs. Later i'll check the tests.

ignite/handlers/ema_handler.py
sandylaker: `self.momentum` refers to the momentum after warmup, `engine.state.ema_momentum` refers to the current momentum. Do you have any suggestion on re-naming these two variables?

vfdev-5: Actually, scheduling the momentum is another story I would say. We have currently a draft PR on any parameter scheduling https://github.com/pytorch/ignite/pull/2090 and maybe in future this could be used to set a warm-up or  gradual decay etc...
OK, let's keep it like that for instance..

sandylaker: I will leave a TODO marker here. So that we can update it after #2090 is merged

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,155 @@
...
+            raise ValueError(
+                f"model should be an instance of nn.Module or its subclasses, but got"
+                f"model: {model.__class__.__name__}"
+            )
+        self.momentum = momentum
+        self.interval = interval
+        self.momentum_warmup = momentum_warmup
+        self.warmup_iters = warmup_iters
+
+        self.unwrapped_model = self._unwrap_model(model)
```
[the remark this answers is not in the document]
sandylaker: We need the online model for the updating the EMA model. If we don't store it as an attribute, we need to provide in the `_update_ema_model` function, and thus in the `attach` function. This is unnecessary, right?

ignite/handlers/ema_handler.py
sandylaker: There are two issues with this API:

- `_update_ema_momemtum` and `_update_weights` use the same event, e.g., `Events.ITERATION_COMPLETED`. So we don't know which function will be called first. If the trainer first update EMA model, then it will raise an error since `Engine.state.ema_momentum` has not been set yet.
- Consider the case where two EMA handlers are attached to the same trainer. The one handler's momentum will override the other one's, depending on which one's `_update_ema_momentum` is called first. Because `Engine.state.ema_momentum` is a float. 

I tackle these issues in the next update, where I use two separate events `momentum_event` and `model_event`, and initialize `Engine.state.ema_momentum` as a dict.

vfdev-5: > _update_ema_momemtum and _update_weights use the same event, e.g., Events.ITERATION_COMPLETED. So we don't know which function will be called first. 

We know which function is called the first -> the first added -> `engine.add_event_handler(event_name, self._update_ema_momentum)`

> Consider the case where two EMA handlers are attached to the same trainer. The one handler's momentum will override the other one's, depending on which one's _update_ema_momentum is called first. Because Engine.state.ema_momentum is a float.

yes, this is a good point. I think we can do the following:
```python
    def attach(
        self, 
        engine: Engine, 
        event_name: Union[str, Events, CallableEventWithFilter, EventsList] = Events.ITERATION_COMPLETED,
        name="ema_momentum",
    ) -> None:

        if hasattr(engine.state, name):
            # warn about colliding attribute
        setattr(engine.state, name, 0.0)
        engine.add_event_handler(event_name, self._update_ema_momentum, name)
        engine.add_event_handler(event_name, self._update_weights)
```

ignite/handlers/ema_handler.py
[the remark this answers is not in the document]
vfdev-5: Yes, you are right, I forgot again about that. Let's call it `self.model`.

ignite/handlers/ema_handler.py
sandylaker: If we initialize `Engine.state.ema_momentum` as a dict, there is no need to warn. We can have multiple EMA handlers attached to the same engine, as long as the keys for `Engine.state.ema_momentum` are different. For example, one for generator and one of discriminator. You can have a look at the examples in docstring in the most recent commit. 

I will merge the `momentum_event` and `model_event` to a single event as you suggest.

vfdev-5: That could be a good idea. However I think it is a bit unusual to have a dict with a single key for a single EMA model (which should be a majority of usecases)...

sandylaker: I see. I will update it in the next commit. Thanks a lot for the suggestions

sandylaker: I updated the `attach` function as shown above. The tests are also updated. Please check them out.

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,207 @@
...
+    The EMA model are updated as follows:
+
+    .. math:: \theta_{\text{EMA}, t+1} = (1 - \lambda) \cdot \theta_{\text{EMA}, t} + \lambda \cdot \theta_{t}
+
+    where :math:`\theta_{\text{EMA}, t}` and :math:`\theta_{t}` are the EMA weights and online model weights at
+    :math:`t`-th iteration, respectively; :math:`\lambda` is the update momentum. The handler allows for linearly
+    warming up the momentum in the beginning when training process is not stable. Current momentum can be retrieved
+    from ``Engine.state.ema_momentum``.
+
+    .. versionadded:: 0.5.0
```
vfdev-5, line 23: Please put this line to the end of the docstring.

vfdev-5 (commented):
Thanks a lot for the update ! 
Just a docstring update.

tests/ignite/handlers/test_ema_handler.py
```diff
@@ -0,0 +1,296 @@
...
+        """Increment the weight by 1 at each iteration"""
+        model.weight.data.add_(1)
+        return 0
+
+    return step_fn
+
+
+@pytest.mark.parametrize("momentum_end", [-1, 2])
+def test_ema_invalid_momentum_end(momentum_end):
+    with pytest.raises(ValueError):
```
vfdev-5, line 42: Usually, we also provide `match="part of error message to match"` argument to ensure which message we check.

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,207 @@
...
+                  print(f"current momentum for generator: {engine.state.gen_ema_momentum}")
+                  print(f"current momentum for discriminator: {engine.state.disc_ema_momentum}")
+
+              engine.run(...)
+
+
+    """
+
+    def __init__(
+        self, model: nn.Module, momentum_start: float = 0.0001, momentum_end: float = 0.0002, warmup_iters: int = 100,
```
vfdev-5, line 124: Question, can we configure it without any warmup somehow ? It is always useful to perform a warm-up ?

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,207 @@
...
+            raise ValueError(f"curr_iter should be at least 1, but got {curr_iter}.")
+        denominator = max(1, self.warmup_iters - 1)
+        momentum = self.momentum_start + (self.momentum_end - self.momentum_start) * (curr_iter - 1) / denominator
+        return min(self.momentum_end, momentum)
+
+    def _update_ema_model(self, engine: Engine, name: str) -> None:
+        """Update weights of ema model"""
+        momentum = getattr(engine.state, name)
+        for ema_v, model_v in zip(self.ema_model.state_dict().values(), self.model.state_dict().values()):
+            print(name, ema_v.data, model_v.data)
```
vfdev-5, line 170: please. remove `print`

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def attach(
+        self,
+        engine: Engine,
+        name: str = "ema_momentum",
+        event: Union[str, Events, CallableEventWithFilter, EventsList] = Events.ITERATION_COMPLETED,
+    ) -> None:
+        """Attach the handler to engine. After the handler is attached, the ``Engine.state`` (a
+        ``dict``) will add an new attribute with name ``name``. Then, current momentum can be retrieved by
+        from ``Engine.state`` when the engine runs.
```
vfdev-5, line 187:
```suggestion
        """Attach the handler to engine. After the handler is attached, the ``Engine.state`` will add an new attribute with name ``name``. Then, current momentum can be retrieved by
        from ``Engine.state`` when the engine runs.
```

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,207 @@
...
+                trainer can have multiple EMA handlers.
+            event: event when the EMA momentum and EMA model are updated.
+
+        """
+        if hasattr(engine.state, name):
+            warnings.warn(
+                f"Attribute: '{name}' is already in Engine.state. Thus it might be "
+                f"overridden by other EMA handlers. Please select another name.",
+                UserWarning,
+            )
```
vfdev-5, line 201: In your previous commits with dict, I think you raised an error instead of the warning. I think it is good to stop execution instead of sliently overwritting things ...

ignite/handlers/ema_handler.py
sandylaker: > Can we configure it without any warmup somehow?
Yes, we can set `momentum_start` = `momentum_end`, so the momentum will be constant. For the future, we can use the `ignite`'s parameter scheduling as an optional argument. So if no scheduling is provided, we keep the constant momentum.

> Is it always useful to perform a warm-up?
In the early stage, the training is not stable, so we hope to keep a small momentum. Usually the the warmup will take 3-5 epoch.  But no empirical study shows that it is ALWAYS useful.

vfdev-5: Thanks for details ! Honestly, I prefer the previous API https://github.com/pytorch/ignite/pull/2098/commits/58312f7f36ffd7baea40b4948f847362d9e11926 : 
```
momentum: float = 0.0002,
momentum_warmup: Optional[float] = None, 
warmup_iters: Optional[int] = None,
```

In this case, once state parameter scheduling is done, warmup args can be deprecated or removed if no release was done before...

tests/ignite/handlers/test_ema_handler.py
sandylaker: Done

sandylaker:
In this update, I changed the `UserWarning` in `attach` to `ValueError`. I also added matched message to enhance the tests

tests/ignite/handlers/test_ema_handler.py
```diff
@@ -0,0 +1,294 @@
...
+
+def test_ema_invalid_model():
+    with pytest.raises(ValueError, match="model should be an instance of nn.Module or its subclasses"):
+        model = "Invalid Model"
+        EMAHandler(model)  # type: ignore
+
+
+@pytest.mark.distributed
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="Skip if no GPU")
+def test_ema_ema_model():
```
vfdev-5, line 71:
```suggestion
def test_ema_ema_model_on_cuda():
```
We have a convention to add `cuda` to make the test run on GPUs

ignite/handlers/ema_handler.py
vfdev-5: @sandylaker what do you think about the above comment ^^^ ?

sandylaker: @vfdev-5 Do you mean only change the name `momentum_start` and `momentum_end` in commit [3bfb9c7](https://github.com/pytorch/ignite/pull/2098/commits/3bfb9c7799626db2ddd89945219aecdd57ff7c8c) back to `momentum_warmup` and `momentum`? I have submitted a new commit to handle this issue.

vfdev-5: Yes, thanks! Also let's set by default the following : 
```
momentum: float = 0.0002,
momentum_warmup: Optional[float] = None, 
warmup_iters: Optional[int] = None,
```
such that there is no warm-up by default.

sandylaker: I updated in the most recent commit. I also added tests to test if the momentum is constant, when at least one of `momentum_warmup` and `warmup_iters` is ``None``.

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,211 @@
...
+    warming up the momentum in the beginning when training process is not stable. Current momentum can be retrieved
+    from ``Engine.state.ema_momentum``.
+
+    Args:
+          model: the online model for which an EMA model will be computed. If ``model`` is ``DataParallel`` or
+              ``DistributedDataParallel``, the EMA smoothing will be applied to ``model.module`` .
+          momentum_warmup: the initial update momentum during warmup phase, the value should be smaller than
+              ``momentum``. Momentum will linearly increase from this value to ``momentum`` in `warmup_iters`
+              iterations. If ``None``, no warmup will be performed.
+          momentum: the update momentum after warmup phase, should be float in range :math:`\left(0, 1 \right)`.
```
vfdev-5, line 28:
```suggestion
          momentum: the update momentum after warmup phase, should be float in range :math:`\left(0, 1 \right)`.
          momentum_warmup: the initial update momentum during warmup phase, the value should be smaller than
              ``momentum``. Momentum will linearly increase from this value to ``momentum`` in ``warmup_iters``
              iterations. If ``None``, no warmup will be performed.
```

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,211 @@
...
+              ``momentum``. Momentum will linearly increase from this value to ``momentum`` in ``warmup_iters``
+              iterations. If ``None``, no warmup will be performed.
+          warmup_iters: iterations of warmup. If ``None``, no warmup will be performed.
+
+    Attributes:
+          ema_model: the exponential moving averaged model.
+          model: the online model that is tracked by EMAHandler. It is ``model.module`` if ``model`` in
+              the initialization method is an instance of ``DistributedDataParallel``.
+          momentum_warmup: the initial update momentum.
+          momentum: the update momentum after warmup phase.
```
vfdev-5, line 36:
```suggestion
          momentum: the update momentum after warmup phase.
          momentum_warmup: the initial update momentum.
```

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,211 @@
...
+          4. Attach ``ema_handler`` to ``trainer``.
+
+    Examples:
+          .. code-block:: python
+
+              device = torch.device("cuda:0")
+              model = nn.Linear(2, 1).to(device)
+              # update the ema every 5 iterations
+              ema_handler = EMAHandler(
+                  model, momentum_warmup=0.0001, momentum=0.0002, warmup_iters=10000)
```
vfdev-5, line 60:
```suggestion
                  model, momentum=0.0002, momentum_warmup=0.0001, warmup_iters=10000)
```

tests/ignite/handlers/test_ema_handler.py
```diff
@@ -0,0 +1,319 @@
...
+    def _bn_step_fn(engine, batch):
+        return 1
+
+    engine = Engine(_bn_step_fn)
+    # engine will run 4 iterations
+    engine.run(_get_dummy_dataloader(), max_epochs=2)
+
+    ema_model = ema_handler.ema_model
+    torch.testing.assert_allclose(ema_model.running_mean, model.running_mean)
+    torch.testing.assert_allclose(ema_model.running_var, model.running_var)
```
vfdev-5, line 165: @sandylaker i was inspecting this test and would like to update it to
```python
import torch
import torch.nn as nn

model = nn.BatchNorm2d(2)
model.running_mean.data.fill_(1.5)
model.running_var.data.fill_(1.5)

def _bn_step_fn(engine, batch):
    x = torch.rand(4, 2, 32, 32)
    _ = model(x)    
    return 1

engine = Engine(_bn_step_fn)
ema_handler.attach(engine)

ema_model = ema_handler.ema_model

@engine.on(Events.ITERATION_COMPLETED)
def check_buffers():
    print(ema_model.running_mean, "vs", model.running_mean)
    torch.testing.assert_allclose(ema_model.running_mean, model.running_mean)
    torch.testing.assert_allclose(ema_model.running_var, model.running_var)

    
# engine will run 4 iterations
engine.run([0, 1], max_epochs=2)
    
torch.testing.assert_allclose(ema_model.running_mean, model.running_mean)
torch.testing.assert_allclose(ema_model.running_var, model.running_var)
```

and it breaks at
```
<ipython-input-5-101d8b59602f> in _update_ema_model(self, engine, name)
    175         momentum = getattr(engine.state, name)
    176         for ema_v, model_v in zip(self.ema_model.state_dict().values(), self.model.state_dict().values()):
--> 177             ema_v.mul_(1.0 - momentum).add_(model_v.data, alpha=momentum)
    178 
    179     def _update_ema_momentum(self, engine: Engine, name: str) -> None:

RuntimeError: result type Float can't be cast to the desired output type Long
```

sandylaker: Thanks for checking the test. This error is because the `BatchNorm2d` will create a new buffer `num_batches_tracked` after it forwards input, and the tensor in this buffer has type `long`.

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,211 @@
...
+
+        denominator = max(1, self.warmup_iters - 1)
+        momentum = self.momentum_warmup + (self.momentum - self.momentum_warmup) * (curr_iter - 1) / denominator
+        return min(self.momentum, momentum)
+
+    def _update_ema_model(self, engine: Engine, name: str) -> None:
+        """Update weights of ema model"""
+        momentum = getattr(engine.state, name)
+        for ema_v, model_v in zip(self.ema_model.state_dict().values(), self.model.state_dict().values()):
+            ema_v.mul_(1.0 - momentum).add_(model_v.data, alpha=momentum)
```
vfdev-5, line 177: I think we have to modify this part of code as the following:
```python
        for ema_v, model_v in zip(self.ema_model.state_dict().values(), self.model.state_dict().values()):
            if model_v.dtype in floating_dtypes:
                ema_v.mul_(1.0 - momentum).add_(model_v.data, alpha=momentum)
            else:
                ema_v.data.copy_(model_v.data)
```

sandylaker: @vfdev-5 what about changing the terms in `zip` to `parameters()`? This should exclude the buffers. After the parameters are updated, we can then copy the buffers directly. From the test, we expect the values in buffers of EMA model and online model to be the same, right?

vfdev-5: Yes, actually, you are right about that. We do not need to EMA on buffers but just copy them. We can do this once while creating ema_model.

sandylaker: @vfdev-5 I think we need to copy them on every update. For example, `_BatchNorm` will update their `running_mean` and `running_var` at each forward. So, if we only copy the buffers when creating the ``ema_model``. The ``ema_model`` will have default initialized buffers, e.g. `running_mean = Tensor([0, 0])`` and ``running_var = Tensor([1, 1])``.

vfdev-5: Actually, what we can do is to setup ema_model buffers to point to model's buffers without copying:
```
ref_model = nn.BatchNorm2d(3)
ema_model = nn.BatchNorm2d(3)

ema_model.load_state_dict(ref_model.state_dict())
for param in ema_model.parameters():
    param.detach_()
# set EMA model's BN buffers as base model BN buffers:
for m1, m2 in zip(ref_model.modules(), ema_model.modules()):
    if isinstance(m1, nn.BatchNorm2d) and isinstance(m2, nn.BatchNorm2d):
        m2.running_mean = m1.running_mean
        m2.running_var = m1.running_var

print(list(ref_model.buffers()), list(ema_model.buffers()))
> [tensor([0., 0., 0.]), tensor([1., 1., 1.]), tensor(0)] [tensor([0., 0., 0.]), tensor([1., 1., 1.]), tensor(0)]

_ = ref_model(torch.rand(12, 3, 32, 32))

print(list(ref_model.buffers()), list(ema_model.buffers()))
> [tensor([0.0502, 0.0504, 0.0499]), tensor([0.9083, 0.9084, 0.9084]), tensor(1)] [tensor([0.0502, 0.0504, 0.0499]), tensor([0.9083, 0.9084, 0.9084]), tensor(0)]

```

sandylaker: It is not due to the scalar type, but due to the fact that `num_batches_tracked` is created after the buffers are copied.

In addition, the example above use `m2.running_mean` to get the buffer variable. However, if the module is wrapped in other modules. ``getattr`` seems not working anymore.
```python3
class Model(nn.Module):
    def __init__(self):
        super().__init__()
        self.bn = nn.BatchNorm(2)

    def forward(self, x):
        return self.bn(x)
```
In this case, ``m = Model(); getattr(m, "bn.running_mean")`` will throw an error.

vfdev-5: Let me check that...

tests/ignite/handlers/test_ema_handler.py
```diff
@@ -0,0 +1,329 @@
...
+import torch.nn as nn
+from torch.nn.parallel import DataParallel, DistributedDataParallel
+from torch.utils.data import DataLoader, TensorDataset
+
+import ignite.distributed as idist
+from ignite.engine import Engine, Events
+from ignite.handlers import EMAHandler
+
+
+def _get_dummy_dataloader() -> DataLoader:
```
KickItLikeShika, line 15: Decorate these `_get_dummy_....` as `@pytest.fixture`would be nice I think, like we did in LRFinder tests here https://github.com/pytorch/ignite/blob/4d6d220e7be7a433cf2e16f92ca7cc6969ecea89/tests/ignite/contrib/handlers/test_lr_finder.py#L136

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,214 @@
...
+import torch.nn as nn
+
+from ignite.engine import CallableEventWithFilter, Engine, Events, EventsList
+
+__all__ = ["EMAHandler"]
+
+
+class EMAHandler:
+    r"""Exponential moving average (EMA) handler can be used to compute a smoothed version of model.
+    The EMA model are updated as follows:
```
KickItLikeShika, line 13: I think this should be `EMA model is updated` if I understand this right..

KickItLikeShika (commented):
Thanks for the awesome work @sandylaker! just quick few comments!

tests/ignite/handlers/test_ema_handler.py
sandylaker: Good idea, I will update it. Thanks.

ignite/handlers/ema_handler.py
```diff
@@ -0,0 +1,214 @@
...
+        return min(self.momentum, momentum)
+
+    def _update_ema_model(self, engine: Engine, name: str) -> None:
+        """Update weights of ema model"""
+        momentum = getattr(engine.state, name)
+        for ema_p, model_p in zip(self.ema_model.parameters(), self.model.parameters()):
+            ema_p.mul_(1.0 - momentum).add_(model_p.data, alpha=momentum)
+        # directly copy the buffers
+        for ema_b, model_b in zip(self.ema_model.buffers(), self.model.buffers()):
+            ema_b.copy_(model_b.data)
```
vfdev-5, line 180:
```suggestion
        # directly copy the buffers
        for ema_b, model_b in zip(self.ema_model.buffers(), self.model.buffers()):
            ema_b.data = model_b.data
```
maybe instead of copying we can create a view of the tensor (less expensive).

vfdev-5 (approved):
LGTM, thanks a lot @sandylaker !

vfdev-5:
@sandylaker I approved the PR and if the CI passes we can merge it and see if GPU CI is also passing. If it is not passing we will have to figure out the issue and send a fixing follow-up PR. Once everything is passing, it would be nice to check this new API on real models and trainings. You can then install nightly release: `pip install --pre pytorch-ignite` and test it on your projects.

sandylaker:
@vfdev-5 Thanks a lot for reviewing the code and giving me suggestions. I have tested the GPU tests (no distributed tests) on my own machine and they all passed. I will take care of the issues if the PR does not pass the CI. In addition, I will also update the API later when the #2090 is merged. For testing the EMA Handler on real-world applications, I think we can test with this [example](https://github.com/pytorch/ignite/blob/master/examples/contrib/cifar10/main.py). Besides the validation performance, we can also measure the extra time consumption of the EMA handler.

vfdev-5:
@sandylaker sounds great !

Btw, here is GPU CI status: https://app.circleci.com/pipelines/github/pytorch/ignite/2012/workflows/d16017e1-2f94-4706-bcbb-4ce91803ee76

sandylaker:
@vfdev-5 Thanks. By the way, why the codecov complains about the project coverage? We have added tests for every function.

sandylaker:
@vfdev-5 Thanks. By the way, why did `codecov` complain about the coverage? We have added tests for every function.

vfdev-5:
@sandylaker codecov's values and pass/fail marks are not always reliable... They show more or less accurately if PR does not cover new functionalities, but other checks can be ignored.

By the way, 2 GPUs tests are failing:
```
FAILED tests/ignite/handlers/test_ema_handler.py::test_ema_ema_model_on_cuda
FAILED tests/ignite/handlers/test_ema_handler.py::test_ema_final_weight_nccl_cuda[1]
FAILED tests/ignite/handlers/test_ema_handler.py::test_ema_final_weight_nccl_cuda[2]
FAILED tests/ignite/handlers/test_ema_handler.py::test_ema_final_weight_gloo_cuda[1]
FAILED tests/ignite/handlers/test_ema_handler.py::test_ema_final_weight_gloo_cuda[2]
```
https://app.circleci.com/pipelines/github/pytorch/ignite/2012/workflows/d16017e1-2f94-4706-bcbb-4ce91803ee76/jobs/6120/parallel-runs/0/steps/0-106

> test_ema_ema_model_on_cuda
>       assert (
            isinstance(ema_model, nn.Module)
            and not isinstance(ema_model, nn.parallel.DistributedDataParallel)
            and not isinstance(ema_model, nn.parallel.DataParallel)
        )

On N GPUs if no distributed group is created, auto_model wraps the input model with DataParallel, so technically our assertion here is not correct.

> test_ema_final_weight_nccl_cuda

It is a bit strange why `model` is `DataParallel` when I would expect to have a DDP...

sandylaker:
@vfdev-5 Let me submit an issue and fix it with a new PR.

vfdev-5:
I think we can omit creating new issue. Thanks

tests/ignite/handlers/test_ema_handler.py
```diff
@@ -0,0 +1,319 @@
...
+def test_ema_final_weight_cuda(get_dummy_model, interval):
+    device = torch.device("cuda:0")
+    _test_ema_final_weight(get_dummy_model(), device=device, ddp=False, interval=interval)
+
+
+@pytest.mark.distributed
+@pytest.mark.parametrize("interval", [1, 2])
+@pytest.mark.skipif(not idist.has_native_dist_support, reason="Skip if no native dist support")
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="Skip if no GPU")
+def test_ema_final_weight_nccl_cuda(get_dummy_model, distributed_context_single_node_nccl, interval):
```
vfdev-5, line 257: Please, remove `cuda` from the test name here (otherwise this test is also running with Windows and failing as no NCCL)
```diff
- test_ema_final_weight_nccl_cuda
+ test_ema_final_weight_nccl
```

tests/ignite/handlers/test_ema_handler.py
```diff
@@ -0,0 +1,319 @@
...
+    _test_ema_final_weight(get_dummy_model(), device=device, ddp=True, interval=interval)
+
+
+@pytest.mark.distributed
+@pytest.mark.parametrize("interval", [1, 2])
+@pytest.mark.skipif(not idist.has_native_dist_support, reason="Skip if no native dist support")
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="Skip if no GPU")
+def test_ema_final_weight_gloo_cuda(get_dummy_model, distributed_context_single_node_gloo, interval):
+    device = idist.device()
+    _test_ema_final_weight(get_dummy_model(), device=device, ddp=True, interval=interval)
```
vfdev-5, line 268: Let's update this test as following and make it run on distributed CPU and GPU
```python
@pytest.mark.distributed
@pytest.mark.parametrize("interval", [1, 2])
def test_distrib_gloo_cpu_or_gpu(get_dummy_model, distributed_context_single_node_gloo, interval):

    device = idist.device()
    _test_ema_final_weight(get_dummy_model(), device=device, ddp=True, interval=interval)
```

sandylaker: I have submitted a new PR. Please check it out.

vfdev-5:
@sandylaker following this discussion https://github.com/pytorch/ignite/pull/2098#issuecomment-877481124 , as GPU issues are fixed we can advance on 
> For testing the EMA Handler on real-world applications, I think we can test with this example. Besides the validation performance, we can also measure the extra time consumption of the EMA handler.

which could be nice tests.

sandylaker:
@vfdev-5 Hi, I will test it this weekend and let you know the results.

sandylaker:
@vfdev-5 I did some experiments yesterday. Here are the experiment settings:

- Dataset: I find it was too easy for the CNNs to achieve over 90% accuracy on the CIFAR-10 dataset. So I choose a more challenging dataset from a [kaggle competition](https://www.kaggle.com/c/cassava-leaf-disease-classification). I manually partition the dataset such that the training set contains 14977 images and the validation set contains 6420 images.
- Model: EfficientNet-B4 without pre-trained weights.
- Optimizer: Adam with learning rate 0.01 and weight decay 0.0004.
- EMA: momentum 0.02, no warmup.

Curves of losses and validation accuracy are in this [link](https://wandb.ai/sandylaker/ignite_leaf_disease?workspace=user-sandylaker). It can be seen that:

- The curves of validation accuracy and validation loss are less fluctuant when EMA is applied.
- It takes some iterations for the ema model to become stable, but the epoch length in our settings is 125, thus the validation loss fluctuates strongly in the beginning. This is also why we choose a large ema momentum.
- It take 9h16m for both runs (w/ or w/o EMA) . So the EMA handler is not a speed bottleneck of the training process.

Both runs achieve the same validation accuracy in the end. I think the EMA will enhance the performance when we train on a large-scale dataset and choose a smaller momentum like 0.001 or 0.0001.

Hope it helps.

vfdev-5:
Thanks for the tests and the feedback @sandylaker !

PR #2111, opened 2021-07-14
Added `start_lr` option to LRFinder

### LINKED ISSUE #2105: Add `start_lr` option to `LRFinder`
## 🚀 Feature

Currently we choose the start learning rate for the Learning Rate Finder as the learning rate specified with the optimizer
in case of linear `step_mode`
https://github.com/pytorch/ignite/blob/master/ignite/handlers/lr_finder.py#L124
And also in case of exp `step_mode` (we use torch `_LRScheduler`) they specify their `base_lrs` as the given learning rates given with the optimizer also 
https://github.com/pytorch/pytorch/blob/c371542efc31b1abfe6f388042aa3ab0cef935f2/torch/optim/lr_scheduler.py#L35
Which is not the best choice if the learning rate with the optimizer is not a small value.

The idea is to add `start_lr` argument to the `attach` method in the `LRFinder` to be used in both modes (linear, exp), that would be a nice improvement to to get a better results from the LRFinder.

KickItLikeShika:
Fixes #2105 

Added `start_lr` option to LRFinder and covered that with tests.

Check list:

### DIFF
--- a/ignite/handlers/lr_finder.py
+++ b/ignite/handlers/lr_finder.py
@@ -122,7 +122,7 @@ class FastaiLRFinder:
             start_lr = optimizer.param_groups[0]["lr"]
         # Initialize the proper learning rate policy
         if step_mode.lower() == "exp":
-            start_lr = [start_lr] * len(optimizer.param_groups)
+            start_lr = [start_lr] * len(optimizer.param_groups)  # type: ignore
             self._lr_schedule = LRScheduler(_ExponentialLR(optimizer, start_lr, end_lr, num_iter))
         else:
             self._lr_schedule = PiecewiseLinear(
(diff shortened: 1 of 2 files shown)


KickItLikeShika:
Can you please review this when you have some time? @vfdev-5

ignite/handlers/lr_finder.py
```diff
@@ -387,9 +391,10 @@ def attach(
                 iteration. It must return the loss of that iteration.
             num_iter: number of iterations for lr schedule between base lr and end_lr. Default, it will
                 run for ``trainer.state.epoch_length * trainer.state.max_epochs``.
+            start_lr: lower bound for lr search. Default, Learning Rate specified with the optimizer.
```
vfdev-5, line 394: Not related to this line. Could you please add to `to_save` docstring above and mention that it should contain optmizer by the key "optimizer". For example,
```
It should contain "optimizer" key for the optimizer.
```

vfdev-5 (approved):
LGTM !

ignite/handlers/lr_finder.py
KickItLikeShika: Done.

PR #2115, opened 2021-07-16
Moved state

KickItLikeShika:
The second PR to tackle refactoring the Engine: https://github.com/pytorch/ignite/wiki/Roadmap#refactor-engine

In this PR we move state from `ignite.engine.events.State` to `ignite.engine.state.State` 

Check list:

### DIFF
--- a/ignite/engine/events.py
+++ b/ignite/engine/events.py
@@ -10,7 +10,7 @@ if TYPE_CHECKING:
     from ignite.base.events_driven import EventsDriven
     from ignite.engine.engine import Engine
 
-__all__ = ["CallableEventWithFilter", "EventEnum", "Events", "EventsList", "RemovableEventHandle"]
+__all__ = ["CallableEventWithFilter", "EventEnum", "Events", "State", "EventsList", "RemovableEventHandle"]
 
 
 class CallableEventWithFilter:
@@ -345,6 +345,10 @@ class EventsList:
     def __or__(self, other: Union[Events, CallableEventWithFilter]) -> "EventsList":
         self._append(event=other)
         return self
+
+
+# Import state here to keep BC compatibility
+from ignite.engine.state import State
 
 
 class RemovableEventHandle:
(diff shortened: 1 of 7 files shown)


ignite/engine/events.py
```diff
@@ -2,17 +2,15 @@
...
+from typing import TYPE_CHECKING, Any, Callable, Iterator, List, Optional, Union
 
 from ignite.engine.utils import _check_signature
 
 if TYPE_CHECKING:
     from ignite.base.events_driven import EventsDriven
     from ignite.engine.engine import Engine
 
-__all__ = ["CallableEventWithFilter", "EventEnum", "Events", "State", "EventsList", "RemovableEventHandle"]
+__all__ = ["CallableEventWithFilter", "EventEnum", "Events", "EventsList", "RemovableEventHandle"]
```
vfdev-5, line 13: To keep BC-compatibility, let's try to import State here as well and revert `__all__`

vfdev-5 (commented):
@KickItLikeShika thanks ! 
I left a comment to improve BC

ignite/engine/events.py
KickItLikeShika: I am not sure if we can import state here, importing State here will give this
```
ImportError: cannot import name 'CallableEventWithFilter' from partially initialized module 'ignite.engine.events' (most likely due to a circular import) 
```

vfdev-5: I think it is possible to do that by checking carefully what we need to import in state.py

KickItLikeShika: We are using `CallableEventWithFilter` for type checking, but using `Events` is a must here, what do you think?

vfdev-5: Yes, that's correct about `CallableEventWithFilter` and `Events`. There is still a solution to solve this circular import issue, it is not very elegant but it would help to keep BC and seems like isort wont complaint.

KickItLikeShika: I think this can be resolved now, thanks for proposing the solution!

vfdev-5 (approved):
LGTM, thanks !

PR #2117, opened 2021-07-19
Update main.py

vfdev-5:
Description:
- Show logs on rank=0 only instead of local_rank=0

Check list:


sdesrozis (commented):
Ok !!

riyaj8888:
where is the config file present because config file is used multiple location in main.py script .
do anyone from where it is imported or what is its content?

PR #2119, opened 2021-07-20
Modify auto_dataloader() to not wrap user provided DistributedSampler

### LINKED ISSUE #2118: Do not wrap existing DistributedSampler in DistributedProxySampler in auto_dataloader()
## 🚀 Feature

I've looked at the implementation of ``ignite.distributed.auto_dataloader()`` and found that when the data loader has a sampler set, it will always wrap this sampler in a `DistributedProxySampler`. This works fine when the data loader sampler is of type `Sampler` and does itself not handle distribution of samples across processes. However, in my case I want to distribute samples differently across processes to minimize cache misses with my custom map-style dataset which buffers chunks of data. This places some constraints on how indices should be grouped based on rank, i.e., which indices should be assigned to the same process. Currently, the only work-around is to use an iterable dataset instead.

However, I would think that the lines:
https://github.com/pytorch/ignite/blob/0f7819a5d2f5c17697eaf4f01848a54ed43e6e53/ignite/distributed/auto.py#L86-L89

could be changed to keep the user provided instance (of a subclass) of `DistributedSampler` if set.

schuhschuh:
Closes #2118

Description:
Pass user provided `DistributedSampler` on to data loader without wrapping it.

Check list:

### DIFF
--- a/tests/ignite/distributed/test_auto.py
+++ b/tests/ignite/distributed/test_auto.py
@@ -57,12 +57,14 @@ def test_auto_dataloader_warning_distributed_sampler(distributed_context_single_
     rank = idist.get_rank()
     world_size = idist.get_world_size()
     auto_dataloader(dataset, sampler=DistributedSampler(dataset, num_replicas=world_size, rank=rank))
-    expected_warning = f"Found distributed sampler with rank={rank + 1}, but process rank is {rank}"
-    with pytest.warns(UserWarning, match=expected_warning):
-        auto_dataloader(dataset, sampler=DistributedSampler(dataset, num_replicas=world_size, rank=rank + 1))
-    expected_warning = f"Found distributed sampler with num_replicas={world_size + 1}, but world size is {world_size}"
-    with pytest.warns(UserWarning, match=expected_warning):
-        auto_dataloader(dataset, sampler=DistributedSampler(dataset, num_replicas=world_size + 1, rank=rank))
+    if world_size > 1:
+        wrong_rank = (rank + 1) % world_size
+        expected_warning = f"Found distributed sampler with rank={wrong_rank}, but process rank is {rank}"
+        with pytest.warns(UserWarning, match=expected_warning):
+            auto_dataloader(dataset, sampler=DistributedSampler(dataset, num_replicas=world_size, rank=wrong_rank))
+        expected_warning = f"Found distributed sampler with num_replicas={world_size + 1}, but world size is {world_size}"
+        with pytest.warns(UserWarning, match=expected_warning):
+            auto_dataloader(dataset, sampler=DistributedSampler(dataset, num_replicas=world_size + 1, rank=rank))
 
 
 @pytest.mark.tpu
--- a/tests/ignite/distributed/test_auto.py
+++ b/tests/ignite/distributed/test_auto.py
@@ -107,7 +107,10 @@ def _test_auto_dataloader(ws, nproc, batch_size, num_workers=1, sampler_name=Non
         if isinstance(data, IterableDataset):
             sampler_type = _InfiniteConstantSampler
         elif ws > 1:
-            sampler_type = DistributedSampler if sampler is None else DistributedProxySampler
+            if sampler is None or isinstance(sampler, DistributedSampler):
+                sampler_type = DistributedSampler
+            else:
+                sampler_type = DistributedProxySampler
         else:
             sampler_type = RandomSampler if sampler is None else type(sampler)
 
--- a/tests/ignite/distributed/test_auto.py
+++ b/tests/ignite/distributed/test_auto.py
@@ -62,9 +62,9 @@ def test_auto_dataloader_warning_distributed_sampler(distributed_context_single_
         expected_warning = f"Found distributed sampler with rank={wrong_rank}, but process rank is {rank}"
         with pytest.warns(UserWarning, match=expected_warning):
             auto_dataloader(dataset, sampler=DistributedSampler(dataset, num_replicas=world_size, rank=wrong_rank))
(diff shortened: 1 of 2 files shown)


vfdev-5 (commented):
Thanks for the PR @schuhschuh !
Could you please add few tests, I'll post another message with details.
It would be also good to raise a type error in `DistributedProxySampler` if sampler is of type `DistributedSampler`. What do you think ?

vfdev-5:
Let's add few tests here: 
1. Check warning if world_size or rank does not match. We can copy and adapt a test like this one:
https://github.com/pytorch/ignite/blob/0f7819a5d2f5c17697eaf4f01848a54ed43e6e53/tests/ignite/distributed/test_auto.py#L36-L43

2. Test with distributed sampler. We can adapt this test code:
https://github.com/pytorch/ignite/blob/0f7819a5d2f5c17697eaf4f01848a54ed43e6e53/tests/ignite/distributed/test_auto.py#L56-L61
and 
https://github.com/pytorch/ignite/blob/0f7819a5d2f5c17697eaf4f01848a54ed43e6e53/tests/ignite/distributed/test_auto.py#L156-L163

https://github.com/pytorch/ignite/blob/0f7819a5d2f5c17697eaf4f01848a54ed43e6e53/tests/ignite/distributed/test_auto.py#L175-L183

https://github.com/pytorch/ignite/blob/0f7819a5d2f5c17697eaf4f01848a54ed43e6e53/tests/ignite/distributed/test_auto.py#L193-L203

https://github.com/pytorch/ignite/blob/0f7819a5d2f5c17697eaf4f01848a54ed43e6e53/tests/ignite/distributed/test_auto.py#L223-L225

3. Check if type error is raised if sampler is distributed sampler passed into `DistributedProxySampler`
https://github.com/pytorch/ignite/blob/0f7819a5d2f5c17697eaf4f01848a54ed43e6e53/tests/ignite/distributed/test_auto.py#L271-L272

schuhschuh:
I'll look into these. The test for raising an error in `DistributedProxySampler` is added as part of #2120 instead.

tests/ignite/distributed/test_auto.py
```diff
@@ -57,6 +79,8 @@ def _test(data):
             sampler = None
         elif sampler_name == "WeightedRandomSampler":
             sampler = WeightedRandomSampler(weights=torch.ones(100), num_samples=100)
+        elif sampler_name == "DistributedSampler":
+            sampler = DistributedSampler(data, num_replicas=ws, rank=idist.get_rank())
```
vfdev-5, line 83: @schuhschuh I think we have to update also the check below:
```python
        if isinstance(data, IterableDataset):
            sampler_type = _InfiniteConstantSampler
        elif ws > 1:
            sampler_type = DistributedSampler if sampler is None else DistributedProxySampler
        else:
            sampler_type = RandomSampler if sampler is None else type(sampler)
```
?

schuhschuh: I thought so at first, too. But looking more closely, the `else` branch uses `type(sampler)` as the expected type, which in case of a user provided distributed sampler should be the type of the user provided sampler.

I also considered to disallow the use of a distributed sampler in `auto_dataloader()` when world size is 1 as in the `no_dist` test case. But given that a distributed sampler with world size 1 and rank 0 should be fine, I think it's best to allow this use case even if the data loading is not actually distributed?

schuhschuh: Ah, I see what you mean. The `ws > 1` case rather, where type no should be `DistributedSampler` instead of `DistributedProxySampler`. Will fix the test.

vfdev-5: > I also considered to disallow the use of a distributed sampler in auto_dataloader() when world size is 1 as in the no_dist test case. But given that a distributed sampler with world size 1 and rank 0 should be fine, I think it's best to allow this use case even if the data loading is not actually distributed?

Yes, I agree, let's allow this use case.

vfdev-5 (approved):
Thanks for the PR @schuhschuh ! Looks good to me 👍

schuhschuh:
Awesome, thanks for the guidance and quick response!

tests/ignite/distributed/test_auto.py
```diff
@@ -43,6 +49,22 @@ def test_auto_dataloader_warning(distributed_context_single_node_gloo):
...
+@pytest.mark.skipif(not idist.has_native_dist_support, reason="Skip if no native dist support")
+@pytest.mark.skipif("WORLD_SIZE" not in os.environ, reason="Skip if WORLD_SIZE not in env vars")
+def test_auto_dataloader_warning_distributed_sampler(distributed_context_single_node_gloo):
+    dataset = DummyDS()
+    rank = idist.get_rank()
+    world_size = idist.get_world_size()
+    auto_dataloader(dataset, sampler=DistributedSampler(dataset, num_replicas=world_size, rank=rank))
+    expected_warning = f"Found distributed sampler with rank={rank + 1}, but process rank is {rank}"
+    with pytest.warns(UserWarning, match=expected_warning):
+        auto_dataloader(dataset, sampler=DistributedSampler(dataset, num_replicas=world_size, rank=rank + 1))
```
vfdev-5, line 62: Unfortunately, this does not work. Maybe, like that:

```suggestion
        auto_dataloader(dataset, sampler=DistributedSampler(
            dataset, num_replicas=world_size, rank=(rank + 1) % world_size)
        )
```
https://github.com/pytorch/ignite/pull/2119/checks?check_run_id=3113040697#step:12:2537

schuhschuh: Indeed, just seen the error message as well.

vfdev-5:
Thanks for the fix @schuhschuh !

PR #2120, opened 2021-07-20
Raise error in DistributedProxySampler when sampler is already a DistributedSampler

schuhschuh:
Raise TypeError when sampler passed on to DistributedProxySampler is of type DistributedSampler. See #2118.


vfdev-5 (approved):
Thanks a lot for the PR @schuhschuh !

PR #2122, opened 2021-07-21
Refactored Events

KickItLikeShika:
The third PR to tackle refactoring the Engine: https://github.com/pytorch/ignite/wiki/Roadmap#refactor-engine

In this PR we move `CallableEventWithFilter`, `EventEnum`, `EventsList`, and `RemovableEventHandle` from `ignite.engine.events` to `ignite.base.base_events`

Check list:


ignite/base/base_events.py
```diff
@@ -0,0 +1,280 @@
+import numbers
```
vfdev-5, line 1: let's name it as `ignite.base.events`

README.md
```diff
@@ -208,7 +208,7 @@ Examples
 Custom events related to backward and optimizer step calls:
 
 ```python
-from ignite.engine import EventEnum
+from ignite.base.base_events import EventEnum
```
vfdev-5, line 211: Please, no BC-breaking changes.

KickItLikeShika: Reverted this, thanks!

docs/source/base.rst
```diff
@@ -1,7 +1,7 @@
 ignite.base
 ============
 
-Base module for Ignite Engine and State
+Base module for Ignite Engine, Events, and State.
```
vfdev-5, line 4:
```suggestion
Base module for Ignite Engine, Events and State.
```

ignite/base/events_driven.py
```diff
@@ -65,13 +66,15 @@ def __init__(self) -> None:
         self.logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
 
     def register_events(
-        self, *event_names: Union[List[str], List[EventEnum]], event_to_attr: Optional[dict] = None
+        self,
+        *event_names: "Union[List[str], List[ignite.engine.events.EventEnum]]",
```
vfdev-5, line 70: Doing this as string does not look good. Let's find another solution

KickItLikeShika: Can you please explain your point?

sdesrozis:
As far as I understand, this PR moves code from `ignite.engine.events` to `ignite.base.events`. I don't have the full detailed picture, yet it looks good and should help the users to build their own engine using `EventsDriven`.

KickItLikeShika:
Thanks a lot for commenting @sdesrozis, let me explain more: we also had problems with circular importing when we tried to import `EventsDrivenState` at `State` (because state will extend it), and so other issues that require some modifications to the design, so we moved all classes except `Event` from `ignite.engine.events` to `ignite.base.events`. The problem we had in this PR is to keep the docs in `ignite.engine.events` and also in `ignite.base.events` but we can't do this because Sphinx will raise duplication label warnings, so we handled that by referencing the events from base at the engine and also explain that in the note in `engine.rst`. So this PR should be OK now.

docs/source/engine.rst
```diff
@@ -19,13 +19,25 @@ ignite.engine.events
...
+    :toctree: generated
+
+    Events
+
+.. currentmodule:: ignite.base.events
+
+.. note::
+    The classes below are also exist in :ref:`base-events`, still
+    can be imported from here, there are no difference, but for future support
+    we recommend importing them from :ref:`base-events`.
```
vfdev-5, line 33:
```suggestion
    The classes below are also defined in :ref:`base-events` and still
    can be imported from this module for backward compatibility. We 
    recommend importing them from :ref:`base-events`.
```

ignite/engine/events.py
```diff
@@ -1,174 +1,13 @@
...
-            @trainer.on(BackpropEvents.BACKWARD_STARTED)
-            def function_before_backprop(engine):
-                # ...
-    """
-
-    pass
+CallableEventWithFilter = CallableEventWithFilter
+EventEnum = EventEnum
+EventsList = EventsList
+RemovableEventHandle = RemovableEventHandle
```
vfdev-5, line 10: Do we need these links now ?

vfdev-5 (approved):
Thanks @KickItLikeShika , lgtm

PR #2125, opened 2021-07-27
Integrated LRFinder with Cifar10 example

KickItLikeShika:
#2004 

Integrated LRFinder with Cifar10 example and updated some code.

Check list:

### DIFF
--- a/examples/contrib/cifar10/main.py
+++ b/examples/contrib/cifar10/main.py
@@ -84,11 +84,13 @@ def training(local_rank, config):
         ) as trainer_with_lr_finder:
             trainer_with_lr_finder.run(train_loader)
         logger.info(f"\tOptimal Learning Rate: {lr_finder.lr_suggestion()}")
-        if config["with_clearml"]:
-            lr_finder_plot = lr_finder.plot()
-            task.logger.report_matplotlib_figure(
-                title="LRFinder plot", series="LR vs Loss", figure=lr_finder_plot, report_image=True
-            )
+        # Remove the attached lr_scheduler to create a new one
+        trainer.remove_event_handler(lr_scheduler, Events.ITERATION_STARTED)
+        # if config["with_clearml"]:
+        #     lr_finder_plot = lr_finder.plot()
+        #     task.logger.report_matplotlib_figure(
+        #         title="LRFinder plot", series="LR vs Loss", figure=lr_finder_plot
+        #     )
         # Create new scheduler with the new optimal learning rate
         config["learning_rate"] = lr_finder.lr_suggestion()
         le = config["num_iters_per_epoch"]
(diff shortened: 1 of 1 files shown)


examples/contrib/cifar10/main.py
```diff
@@ -324,31 +370,19 @@ def train_step(engine, batch):
...
     )
 
-    resume_from = config["resume_from"]
-    if resume_from is not None:
-        checkpoint_fp = Path(resume_from)
-        assert checkpoint_fp.exists(), f"Checkpoint '{checkpoint_fp.as_posix()}' is not found"
-        logger.info(f"Resume from a checkpoint: {checkpoint_fp.as_posix()}")
-        checkpoint = torch.load(checkpoint_fp.as_posix(), map_location="cpu")
-        Checkpoint.load_objects(to_load=to_save, checkpoint=checkpoint)
-
```
vfdev-5, line 351: I do not expect to see these changes. Why should we move this code elsewhere ?

examples/contrib/cifar10/main.py
```diff
@@ -75,6 +75,50 @@ def training(local_rank, config):
...
+        ) as trainer_with_lr_finder:
+            trainer_with_lr_finder.run(train_loader)
+        logger.info(f"\tOptimal Learning Rate: {lr_finder.lr_suggestion()}")
+        # Remove the attached lr_scheduler to create a new one
+        trainer.remove_event_handler(lr_scheduler, Events.ITERATION_STARTED)
+        # if config["with_clearml"]:
+        #     lr_finder_plot = lr_finder.plot()
+        #     task.logger.report_matplotlib_figure(
+        #         title="LRFinder plot", series="LR vs Loss", figure=lr_finder_plot
+        #     )
```
vfdev-5, line 93: Why this code is commented ?

examples/contrib/cifar10/main.py
KickItLikeShika: `to_save` needs to be updated as we will update `lr_scheduler` if `use_lr_finder` is true

examples/contrib/cifar10/main.py
```diff
@@ -75,6 +75,50 @@ def training(local_rank, config):
...
+        # Create new scheduler with the new optimal learning rate
+        config["learning_rate"] = lr_finder.lr_suggestion()
+        le = config["num_iters_per_epoch"]
+        milestones_values = [
+            (0, 0.0),
+            (le * config["num_warmup_epochs"], config["learning_rate"]),
+            (le * config["num_epochs"], 0.0),
+        ]
+        lr_scheduler = PiecewiseLinear(optimizer, param_name="lr", milestones_values=milestones_values)
```
vfdev-5, line 102: Instead of redefining `lr_scheduler` instance, can't we rewrite just its `milestones_values` ?

examples/contrib/cifar10/main.py
KickItLikeShika: Doesn't work as expected yet, I'm trying to understand but in their docs they don't explain this part well

vfdev-5 (commented):
Thanks for the PR @KickItLikeShika however, these changes does not look solid to me...
Let's figure out how to use lr finder without all these side modifications.

examples/contrib/cifar10/main.py
KickItLikeShika: Yes that can be done, thanks for the recommendation!

KickItLikeShika: Sorry, I tried this but it won't work that way, because of this
https://github.com/pytorch/ignite/blob/master/ignite/handlers/param_scheduler.py#L932

vfdev-5: You can just update middle values of `lr_scheduler.values` and put a comment explaining what you are doing.

KickItLikeShika: That's done, thanks for the suggestions!

PR #2128, opened 2021-07-30
Refactored Engine

KickItLikeShika:
The 4th PR to tackle refactoring the Engine: https://github.com/pytorch/ignite/wiki/Roadmap#refactor-engine

In this PR we refactored `Engine` and some made some other necessary changes.

Check list:


ignite/base/events_driven.py
```diff
@@ -282,12 +283,20 @@ def __init__(
     def __getattr__(self, attr: str) -> Any:
         evnts = None
         if attr in self._attr_to_events:
-            evnts = self._attr_to_events[attr]
+            if attr == "epoch":
+                evnts = self._attr_to_events["epoch"][0]  # Fetch EPOCH_STARTED
+            elif attr == "iteration":
+                evnts = self._attr_to_events["iteration"][2]  # Fetch ITERATION_STARTED
+            else:
+                evnts = self._attr_to_events[attr]
```
vfdev-5, line 291: as we talked in VC, this part of code is to remove

KickItLikeShika: I am not sure if that's can be done in the way we talked about, if we replaced `event_to_attr` arg with `attr_to_event` and if it's None, we will need to put manually so we won't avoid using `epoch` and `iteration` here. And also `_attr_to_event` is updated anyway within `update_mapping` method, so `iteration` will be fetched from `GET_BATCH_STARTED` also. So I'm thinking about either we override `__getattr__` in State or a more nice solution is to just override `update_mapping` in state to only map `ITERATION_STARTED` and `EPOCH_STARTED` in case of iteration/epoch
`update_mapping` in state maybe should be something like this, I tried it and it works fine
```py
 if event_to_attr:
            for k, v in event_to_attr.items():
                if v == 'epoch' and k != Events.EPOCH_STARTED:
                    attr_evnts = self._attr_to_events["epoch"]
                    continue
                if v == 'iteration' and k != Events.ITERATION_STARTED:
                    attr_evnts = self._attr_to_events["iteration"]
                    continue
                else:
                    attr_evnts = self._attr_to_events[v]
                if k not in attr_evnts:
                    attr_evnts.append(k)
```
What do you think?

vfdev-5: @KickItLikeShika again "iteration" and "epoch" are specific to Engine and State and shouldn't appear in any case in EventsDrivenState.

KickItLikeShika: Yes, that's why I said we will extend `update_mapping` to state and do that?

vfdev-5: This solution does not look better neither

ignite/base/events_driven.py
```diff
@@ -307,14 +316,15 @@ def __setattr__(self, attr: str, value: Any) -> None:
 
         super().__setattr__(attr, value)
 
-    def update_mapping(self, event_to_attr: Mapping[Any, str]) -> None:
+    def update_mapping(self, event_to_attr: Optional[Dict[Any, Any]]) -> None:
```
vfdev-5, line 319: Why it should be optional ? Empty method does not make much sense

KickItLikeShika: Mypy is complaining about it should be optional, because in `register_events` in Engine `event_to_attr` is optional https://github.com/pytorch/ignite/blob/b2ff869c07fae30640cd37416596b20e76fa4dd1/ignite/engine/engine.py#L228
so either we do this or we ignore mypy in the line I sent above

vfdev-5: or either you do `if event_to_attr is not None`

ignite/base/events_driven.py
KickItLikeShika: Can you please explain your point why it doesn't look nice, for me:
- Avoided mentioning `epoch`/`iteration` in `EventsDrivenState` and made this specific to `State`.
- Kept `EventsDrivenState` generic as it is with the same simple implementation.
- Not much duplication as `updating_mapping` is a 5 lines method.
I can push the updates so you can explain your concerns better. Thanks.

KickItLikeShika: Anyway, It doesn't work as expected, I will look for another solution, thanks.

KickItLikeShika:
I want to note that our solution in this case is more limited than we think, we can't in anyway (I think) override `attr_to_event` and leave only `ITERATION/EPOCH_STARTED` there, because we use `attr_to_event` to in `__setattr__` in `EventsDrivenState` to set `allowed_events_count` values, and if we excluded the rest of the events that would cause problems as what happened in `test_gradient_on_resume_cpu` in `test_deterministic`. So what the solution should be similar to what I have done now. Thanks.

ignite/base/events_driven.py
```diff
@@ -114,6 +116,8 @@ def add_event_handler(self, event_name: Any, handler: Callable, *args: Any, **kw
         Returns:
             :class:`~ignite.base.events.RemovableEventHandle`, which can be used to remove the handler.
         """
+        from ignite.engine.events import Events
+
```
vfdev-5, line 120: Put this import near the usage, line 134 : `event_args = (Exception(),) if event_name == Events.EXCEPTION_RAISED else ()`
and leave a comment saying that we have to redesign this part of code.

KickItLikeShika:
Added an option to `EventsDrivenState` if the user wants to specify how he wants to retrieve a specific attribute, if the user didn't specify that, the attribute will be retrieved by default from `event_at_attr`.

vfdev-5:
> Added an option to `EventsDrivenState` if the user wants to specify how he wants to retrieve a specific attribute, if the user didn't specify that, the attribute will be retrieved by default from `event_at_attr`.

Looks better, but I think we should get rid of `event_to_attr` and ask user to provide a meaningful mapping such that everyone fully understands what is shown on `get` and what is done on `set`. Current usage of `max` of the values is not good. We have to provide more deterministic solution.

KickItLikeShika:
@vfdev-5 by `meaningful mapping` you mean the user provides `attr_to_evetns directly`?

KickItLikeShika:
@vfdev-5 can you please review this?

ignite/base/events_driven.py
```diff
@@ -257,34 +262,35 @@ class EventsDrivenState:
     Args:
         engine: ignite engine :class:`~ignite.base.events_driven.EventsDriven` that used to access
             the allowed events and their counters.
-        event_to_attr: mapping consists of the events from :class:`~ignite.engine.events.Events`
-            or any other custom events added by :meth:`~ignite.base.events_driven.EventsDriven.register_events`.
+        attr_to_events: mapping consists of the attributes mapped to a list events from
+            :class:`~ignite.engine.events.Events` or any other custom events added
+            by :meth:`~ignite.base.events_driven.EventsDriven.register_events`.
```
vfdev-5, line 267: Here user do not understand which exactly event is used with the attribute.

ignite/base/events_driven.py
```diff
@@ -304,14 +310,22 @@ def __setattr__(self, attr: str, value: Any) -> None:
 
         super().__setattr__(attr, value)
 
-    def update_mapping(self, event_to_attr: Mapping[Any, str]) -> None:
-        """Looks for the attribute and check if not existed to add it.
+    def update_mapping(self, event_to_attr: Mapping["Events", str]) -> None:
```
vfdev-5, line 313: We can change the signature and replace `event_to_attr` by `attr_to_events`

ignite/base/events_driven.py
```diff
@@ -257,34 +262,35 @@ class EventsDrivenState:
...
 
         self.engine = engine  # type: Optional[EventsDriven]
-
-        self._attr_to_events = defaultdict(list)  # type: Mapping[str, Any]
-        if event_to_attr is not None:
-            for k, v in event_to_attr.items():
-                self._attr_to_events[v].append(k)
+        self._attr_to_events = defaultdict(list)  # type: Mapping[str, List[Events]]
+        if attr_to_events:
+            self._attr_to_events = attr_to_events
```
vfdev-5, line 284: Recode this lines into a single line definition

vfdev-5: Check also for the given type and raise issues if something else is given

ignite/base/events_driven.py
KickItLikeShika: I put a note to clarify this part
> Note:
        The attributes will be retrieved according to the first element in the list of the event at `attr_to_events`.

ignite/engine/engine.py
```diff
@@ -123,26 +123,33 @@ def compute_mean_std(engine, batch):
...
-        self.state = State()
+        self._state = State(engine=self)
         self._state_dict_user_keys = []  # type: List[str]
-        self._allowed_events = []  # type: List[EventEnum]
 
         self._dataloader_iter = None  # type: Optional[Iterator[Any]]
         self._init_iter = []  # type: List[int]
 
-        self.register_events(*Events)
+        self.register_events(*Events, event_to_attr=State.event_to_attr)
```
vfdev-5, line 137: Does it work ?

ignite/engine/engine.py
```diff
@@ -333,23 +282,7 @@ def has_event_handler(self, handler: Callable, event_name: Optional[Any] = None)
...
-                if self._compare_handlers(handler, h):
-                    return True
-        return False
-
-    @staticmethod
-    def _compare_handlers(user_handler: Callable, registered_handler: Callable) -> bool:
-        if hasattr(registered_handler, "_parent"):
-            registered_handler = registered_handler._parent()  # type: ignore[attr-defined]
-        return registered_handler == user_handler
+        return super(Engine, self).has_event_handler(handler, event_name=event_name)
```
vfdev-5, line 285: Finally, I think we can remove this methods and ensure that docs are taken from base class

ignite/base/events_driven.py
vfdev-5: You should say about this note here, then. Otherwise, it wont be read.

ignite/base/events_driven.py
KickItLikeShika: That will require refactoring in the tests, also another another thought I wanted to keep both options for the user, either to give `attr_to_events` in the constructor or `event_to_attr` in `update_mapping`, what do you think?

ignite/base/events_driven.py
```diff
@@ -257,34 +262,34 @@ class EventsDrivenState:
     Args:
         engine: ignite engine :class:`~ignite.base.events_driven.EventsDriven` that used to access
             the allowed events and their counters.
-        event_to_attr: mapping consists of the events from :class:`~ignite.engine.events.Events`
-            or any other custom events added by :meth:`~ignite.base.events_driven.EventsDriven.register_events`.
+        attr_to_events: mapping consists of the attributes mapped to a list events from
+            :class:`~ignite.engine.events.Events` or any other custom events added
+            by :meth:`~ignite.base.events_driven.EventsDriven.register_events`.
+            The attributes will be retrieved according to the first element in the list
+            of the event.
```
vfdev-5, line 269: Please, rephrase and pay attention on typos

ignite/engine/engine.py
KickItLikeShika: I have been trying to find a way to duplicate the docs from a class to another by the way we want and I didn't find anything, if we want to remove these methods from here, we can mention the list of methods for example under the `Engine` in `engine.rst`, something similar to what we did with `events`.
For example:
```
ignite.engine.engine
--------------------------

.. currentmodule:: ignite.engine.engine

.. autosummary::
    :nosignatures:
    :toctree: generated

    Engine
 
.. currentmodule:: ignite.base.event_driven

.. note::
    Engine class is now derived from :class:`ignite.base.event_driven`. 

# and we can list all methods that engine include either from engine of events_driven
```

vfdev-5: OK, let's keep things as they are for now

ignite/base/events_driven.py
```diff
@@ -257,34 +262,33 @@ class EventsDrivenState:
     Args:
         engine: ignite engine :class:`~ignite.base.events_driven.EventsDriven` that used to access
             the allowed events and their counters.
-        event_to_attr: mapping consists of the events from :class:`~ignite.engine.events.Events`
-            or any other custom events added by :meth:`~ignite.base.events_driven.EventsDriven.register_events`.
+        attr_to_events: mapping consists of the attributes mapped to a list events from
+            :class:`~ignite.engine.events.Events` or any other custom events added
+            by :meth:`~ignite.base.events_driven.EventsDriven.register_events`.
+            Getting attributes is done based the on first element in the list of the events.
```
vfdev-5, line 268:
```suggestion
            Getting attribute values is done based the on first element in the list of the events.
```

ignite/base/events_driven.py
vfdev-5: @KickItLikeShika why did you resolve the comments without updating the code ?

ignite/base/events_driven.py
```diff
@@ -257,34 +262,33 @@ class EventsDrivenState:
...
         evnts = None
         if attr in self._attr_to_events:
             evnts = self._attr_to_events[attr]
 
         if self.engine and evnts:
-            # return max of available event counts
-            counts = [self.engine._allowed_events_counts[e] for e in evnts]
-            return max(counts)
+            # return first of available event counts
+            return self.engine._allowed_events_counts[evnts[0]]
```
vfdev-5, line 291: We should make sure that the list is not empty somewhere.

ignite/base/events_driven.py
```diff
@@ -304,14 +308,30 @@ def __setattr__(self, attr: str, value: Any) -> None:
...
+                Getting attributes is done based on the first element in the list of the events.
         """
-        for k, v in event_to_attr.items():
-            attr_evnts = self._attr_to_events[v]
-            if k not in attr_evnts:
-                attr_evnts.append(k)
+        if (
+            not isinstance(attr_to_events, dict)
+            or not isinstance(list(attr_to_events.keys())[0], str)
+            or not isinstance(list(attr_to_events.values())[0], list)
```
vfdev-5, line 324: All keys should be str and all values should be lists

ignite/base/events_driven.py
KickItLikeShika: Oh sorry my bad, I will update it now

KickItLikeShika:
I have made a mistake sorry pulling the updates, I will revert the commit

ignite/base/events_driven.py
```diff
@@ -304,14 +308,30 @@ def __setattr__(self, attr: str, value: Any) -> None:
...
-            attr_evnts = self._attr_to_events[v]
-            if k not in attr_evnts:
-                attr_evnts.append(k)
+        if (
+            not isinstance(attr_to_events, dict)
+            or not all(isinstance(key, str) for key in list(attr_to_events.keys()))
+            or not all(isinstance(value, list) for value in list(attr_to_events.values()))
+        ):
+            raise TypeError(
+                "'attr_to_events' must be a dictionary, it shoule map keys of type string to list of Events"
```
vfdev-5, line 327:
```suggestion
                "'attr_to_events' must be a dictionary, it should map keys of type string to list of Events"
```

ignite/base/events_driven.py
```diff
@@ -304,14 +308,30 @@ def __setattr__(self, attr: str, value: Any) -> None:
 
         super().__setattr__(attr, value)
 
-    def update_mapping(self, event_to_attr: Mapping[Any, str]) -> None:
-        """Looks for the attribute and check if not existed to add it.
+    def update_mapping(self, attr_to_events: Mapping[str, List["Events"]]) -> None:
+        """Maps each attribute to a list of the corresponding events in the same given order.
```
vfdev-5, line 312: This docstring description is unclear.

ignite/base/events_driven.py
```diff
@@ -304,14 +308,30 @@ def __setattr__(self, attr: str, value: Any) -> None:
...
+            )
+
+        for k, v in attr_to_events.items():
+            if k not in self._attr_to_events:
+                self._attr_to_events[k] = v  # type: ignore
+            elif k in self._attr_to_events:
+                attr_evnts = self._attr_to_events[k]
+                for evnt in v:
+                    if evnt not in attr_evnts:
+                        attr_evnts.append(evnt)
```
vfdev-5, line 337: I think we should not alter existing attributes in the map, otherwise the API is cumbersome.

ignite/base/events_driven.py
```diff
@@ -304,14 +308,30 @@ def __setattr__(self, attr: str, value: Any) -> None:
...
+        if (
+            not isinstance(attr_to_events, dict)
+            or not all(isinstance(key, str) for key in list(attr_to_events.keys()))
+            or not all(isinstance(value, list) for value in list(attr_to_events.values()))
+        ):
+            raise TypeError(
+                "'attr_to_events' must be a dictionary, it shoule map keys of type string to list of Events"
+            )
+
+        for k, v in attr_to_events.items():
```
vfdev-5, line 330: Why we can't use `dict.update` method ?

ignite/base/events_driven.py
```diff
@@ -304,14 +306,26 @@ def __setattr__(self, attr: str, value: Any) -> None:
...
+                Getting attribute values is done based on the first element in the list of the events.
         """
-        for k, v in event_to_attr.items():
-            attr_evnts = self._attr_to_events[v]
-            if k not in attr_evnts:
-                attr_evnts.append(k)
+        if (
+            not isinstance(attr_to_events, dict)
+            or not all(isinstance(key, str) for key in list(attr_to_events.keys()))
+            or not all(isinstance(value, list) for value in list(attr_to_events.values()))
```
vfdev-5, line 323: it is useless to wrap with `list`

ignite/base/events_driven.py
vfdev-5: Can you consider this question ?

KickItLikeShika: Yes, updating here will update the already existed keys, if the user missed up with them or something that will lead to update them, which is we decided not to do this
> I think we should not alter existing attributes in the map, otherwise the API is cumbersome.

vfdev-5: I'm very unsure about `update_mapping` API and what we would like there. I'd like to simplify everything make strictly necessary things

docs/source/concepts.rst
```diff
@@ -266,13 +266,13 @@ These events could be used to attach any handler and are fired using :meth:`~ign
 If you want to use filtering with custom events (e.g. ``CustomEvents.CUSTOM_STARTED(every=5)``), you need to do 3 more things:
 
 - ``engine.state`` should have corresponding attributes for the events, e.g. ``engine.state.custom_started``
-- you need to pass a dict `event_to_attr` to :meth:`~ignite.engine.engine.Engine.register_events`, which maps between events and state attributes, e.g.
+- you need to pass a dict `attr_to_events` to :meth:`~ignite.engine.engine.Engine.register_events`, which maps between state attributes and events, e.g.
```
vfdev-5, line 269: Probably, we need to explain why list and that the first is taken for the attribute value.

Please, remove the notes below about inscreasing the counter and the warning

ignite/base/events_driven.py
```diff
@@ -62,17 +64,20 @@ def __init__(self) -> None:
         self.logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
 
     def register_events(
-        self, *event_names: Union[List[str], List[EventEnum]], event_to_attr: Optional[dict] = None,
+        self, *event_names: Union[List[str], List[EventEnum]], attr_to_events: Optional[dict] = None,
```
vfdev-5, line 67: I think we have to move `attr_to_events` somewhere else. Here, it is unused and does not make much sense.
Maybe, we'll have to add *kwargs to make mypy happy while inheriting from this class and overridding the method

ignite/contrib/handlers/base_logger.py
```diff
@@ -172,17 +172,21 @@ def attach(
...
         """
+        # create a list of the available events
+        state_events = []
+        for event in State.attr_to_events.values():
+            state_events.extend(event)
+
         if isinstance(event_name, EventsList):
             for name in event_name:
-                if name not in State.event_to_attr:
+                if name not in state_events:
```
vfdev-5, line 182: I think you have to dig more into the code to understand why we did that previously and thus decide how to make a proper code change

ignite/engine/engine.py
```diff
@@ -123,28 +123,35 @@ def compute_mean_std(engine, batch):
...
+        return self._state
+
+    @state.setter
+    def state(self, new_state: State) -> None:
+        self._state = new_state
+        self._state.engine = self
+
     def register_events(
-        self, *event_names: Union[List[str], List[EventEnum]], event_to_attr: Optional[dict] = None
+        self, *event_names: Union[List[str], List[EventEnum]], attr_to_events: Optional[dict] = None
```
vfdev-5, line 154: You can not remove an arg like that without deprecating it

ignite/engine/engine.py
```diff
@@ -207,45 +217,20 @@ class TBPTT_Events(EventEnum):
...
-        return wrapper
-
-    def _assert_allowed_event(self, event_name: Any) -> None:
-        if event_name not in self._allowed_events:
-            self.logger.error(f"attempt to add event handler to an invalid event {event_name}")
-            raise ValueError(f"Event {event_name} is not a valid event for this {self.__class__.__name__}.")
+        super(Engine, self).register_events(*event_names, attr_to_events=attr_to_events)
+        if attr_to_events is not None:
+            for attribute, events in attr_to_events.items():
+                self.state.update_attribute_mapping(attribute, events)
```
vfdev-5, line 233:
```suggestion
                self._state.update_attribute_mapping(attribute, events)
```

ignite/engine/state.py
```diff
@@ -33,21 +35,22 @@ class State:
         kwargs: keyword arguments to be defined as State attributes.
     """
 
-    event_to_attr = {
```
vfdev-5, line 36: You have to keep it for BC as `event_to_attr` is a public class attribute. Try to see if we can deprecate it on its usage somehow (e.g. as property?)

ignite/engine/state.py
```diff
@@ -64,18 +67,26 @@ def __init__(self, **kwargs: Any) -> None:
         for k, v in kwargs.items():
             setattr(self, k, v)
 
-        self._update_attrs()
+        # Update the attributes if it's a standalone state
+        # to keep BC compatibility
```
vfdev-5, line 71: Can you put more explanation about what exactly we are trying to keep

ignite/engine/state.py
```diff
@@ -64,18 +67,26 @@ def __init__(self, **kwargs: Any) -> None:
...
 
     def _update_attrs(self) -> None:
-        for value in self.event_to_attr.values():
-            if not hasattr(self, value):
-                setattr(self, value, 0)
+        for key in self.attr_to_events.keys():
+            if not hasattr(self, key):
+                setattr(self, key, 0)
 
     def get_event_attrib_value(self, event_name: Union[str, Events, CallableEventWithFilter]) -> int:
```
vfdev-5, line 80: Let's deprecate this method as now it is engine's responsibility

KickItLikeShika: I'm not sure about doing this now, I suggest leave this part for now as didn't think of a replacement yet, check this https://github.com/pytorch/ignite/search?q=get_event_attrib_value, it's used in engine/handlers/loggers/tests/..

ignite/engine/engine.py
KickItLikeShika: I'm thinking about leaving both for now `event_to_attr` and `attr_to_events`, and do something like this inside the method
```py
if event_to_attr is not None:
        # raise warning about deprecation
        # create attr_to_events from event_to_attr

# then continue the same as now
```

ignite/base/events_driven.py
KickItLikeShika: Oh, I see, thanks for pointing to this

ignite/engine/state.py
KickItLikeShika: That raised the deprecation warning in case of just using it as an instance attribute, but I think that's not enough, I am trying to find a way to raise a deprecation warning while using it as a class attribute

KickItLikeShika:
I don't understand why this PR is closed and the branch is deleted! The branch still in my fork here https://github.com/pytorch/ignite/compare/refactor-engine...KickItLikeShika:improve-engine

vfdev-5:
It is you who closed it and deleted: 
image

KickItLikeShika:
Yeah I saw that but that's very weird!! the branch still in my fork and locally also!

vfdev-5:
image

PR #2130, opened 2021-08-01
Mean Average Precision

gucifer:
Fixes #{520}

Description:
Created a COCO Style Implementation for mAP metric. Both of them are calculated at once since both require similar calculation and have similar use case. The overall implementation is complicated so any advice to simplify it is appreciated.
Check list:


sdesrozis:
We already discussed a plan one week ago :

(1) remove `mAR`, use one `max_det` 
(2) do a comparison with the original `pycocotools` implementation 
(3) do an api based on preds/gts of images, it means mostly move pieces of code 
(4) optimize the code to save space/cpu, it means change the data structures

The first thing you had to do was having a test, yet I don't see any test.

sdesrozis:
@gucifer It misses tests, from unitary to ddp.

ignite/metrics/mAP.py
```diff
@@ -0,0 +1,402 @@
...
+i_xmax = 4
+i_ymax = 5
+i_area = 6
+i_crowd = 7
+g_ignore = 8
+d_confidence = 8
+
+
+class AP(Metric):
+    r"""Calculates Average Precision for object detection data.
```
sdesrozis, line 50: Compute the Mean Average Precision, right ?

ignite/metrics/mAP.py
```diff
@@ -0,0 +1,402 @@
...
+i_ymin = 3
+i_xmax = 4
+i_ymax = 5
+i_area = 6
+i_crowd = 7
+g_ignore = 8
+d_confidence = 8
+
+
+class AP(Metric):
```
sdesrozis, line 49: Why not MeanAveragePrecision ?

ignite/metrics/mAP.py
```diff
@@ -0,0 +1,402 @@
...
+            if iscrowd:
+                u = da
+            else:
+                u = da + ga - i
+            ious[m - d - 1, g] = i / u
+
+    return ious
+
+
+i_id = 0
```
sdesrozis, line 37: Is it possible to wrap in an enum inside the class ?

ignite/metrics/mAP.py
```diff
@@ -0,0 +1,402 @@
...
+            mAP = AP(area_rngs=area_rngs, max_det=max_det, iou_thrs=iou_thrs, rec_thrs=rec_thrs)
+
+            mAP.update([[1,gts],[1,dts]])
+
+            mAP.compute()
+
+    .. versionadded:: 0.5.0
+    """
+
+    def __init__(
```
sdesrozis, line 114: Could we use some default values here ?

ignite/metrics/mAP.py
```diff
@@ -0,0 +1,402 @@
...
+            for eval_imgs in gather_dicts:
+                for key in eval_imgs:
+                    keys.add(key)
+            combined_eval_imgs: defaultdict = defaultdict(list)
+            for key in keys:
+                for eval_imgs in gather_dicts:
+                    combined_eval_imgs[key] += eval_imgs[key]
+            self.eval_imgs = combined_eval_imgs
+
+    def compute(self) -> torch.Tensor:
```
sdesrozis, line 396: Could the output be a dict rather than a tensor ? It would help the user and ease the print of the results

ignite/metrics/mAP.py
```diff
@@ -0,0 +1,402 @@
...
+
+            import torch
+            from ignite.metrics import AP
+
+            gts = torch.tensor([[1,73,249.1,50.7,375.3,277.2,101928.52,0,0],
+                    [2,73,152.3,50.29,47,312.38, 4097.4,0,0]])
+            dts = torch.tensor([[1,73,176.38,50.3,463.62,312.4,144825.6,0,0.328],
+                    [2,73,355.5,111.87,210,93.75,19687.5,0,0.64]])
+
+            area_rngs = {
```
sdesrozis, line 97: Use defaults to simplify the usage

ignite/metrics/mAP.py
```diff
@@ -0,0 +1,402 @@
...
+                'all' : [0, float("inf")],
+            }
+            max_det = 100
+
+            iou_thrs = torch.tensor([0.5 + (i/20) for i in range(10)])
+            rec_thrs = torch.tensor([i/100 for i in range(101)])
+
+            mAP = AP(area_rngs=area_rngs, max_det=max_det, iou_thrs=iou_thrs, rec_thrs=rec_thrs)
+
+            mAP.update([[1,gts],[1,dts]])
```
sdesrozis, line 107: This example is not user friendly. Explain what the arguments are supposed to be.

ignite/metrics/mAP.py
gucifer: I don't understand what are suggesting, can you please elaborate a bit?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,420 @@
+from collections import defaultdict
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
+
+import torch
+
+from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
+
```
sdesrozis, line 7: it misses `__all__ = [...]`

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,420 @@
...
+
+def iou(y: torch.Tensor, y_pred: torch.Tensor, crowd: List) -> torch.Tensor:
```
sdesrozis, line 9: Is there a test of this method ? To keep it private (and don't show it in the doc), please use a private naming (like `_iou`)

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,420 @@
...
+    m = len(y_pred)
+    n = len(y)
+
+    ious = torch.zeros(m, n)
+
+    for g in range(n):
+        Y = y[g].tolist()
```
sdesrozis, line 16: The naming of the `Y` variable does not respect the PEP8 convention. Please fix and check the file.

ignite/metrics/mean_average_precision.py
sdesrozis, line 9: Add a comment to explain the format of the tensors.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,420 @@
...
+        ya = Y[2] * Y[3]
+        iscrowd = crowd[g]
+        for d in range(m):
+            P = y_pred[d].tolist()
+            pa = P[2] * P[3]
+            ious[m - d - 1, g] = 0
```
sdesrozis, line 22: Why the array is filled in reverse order ?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,420 @@
...
+            w = min(P[2] + P[0], Y[2] + Y[0]) - max(P[0], Y[0])
+            h = min(P[3] + P[1], Y[3] + Y[1]) - max(P[1], Y[1])
+            if w <= 0 or h <= 0:
+                continue
+            i = w * h
+            if iscrowd:
```
sdesrozis, line 28: Could we have iscrowd optional ?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,420 @@
...
+
+    Example:
+
+        .. code-block:: python
+
+            import torch
+            from ignite.metrics import MeanAveragePrecision
+
+            # Detection Format:
+            # (id, class, xmin, ymin, width, height, area, crowd, ignore/confidence)
```
sdesrozis, line 81: area, crowd, ignore are not so trivial. Could you explain ? This docstring is the root of the doc, it should be extremely clear.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+    def compute(self) -> Dict:
+        self._gather_all()
+        self._accumulate()
+
+        results = {
+            "all@0.5": self._summarize(iou_thr=0.5),
+            "all@0.75": self._summarize(iou_thr=0.75),
+        }
+
+        for area in self.object_area_ranges:
```
sdesrozis, line 406: `all` shoudl be excluded

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+
+        if iou_thresholds is None:
+            iou_thresholds = [0.5, 0.55, 0.6, 0.65, 0.7, 0.75, 0.8, 0.85, 0.9, 0.95]
+        self.iou_thresholds = torch.tensor(iou_thresholds, device=device)
+
+        if rec_thresholds is None:
+            rec_thresholds = [i / 100 for i in range(101)]
+        self.rec_thresholds = torch.tensor(rec_thresholds, device=device)
+
+        class labels:
```
sdesrozis, line 140: Naming is not PEP8 compliant.

Anyway, the labels would be included in the class. Sorry, it was not a good idea to have it in a specific class. Please, add the labels statically in the class

```python
class MeanAveragePrecision(Metric):
    __id = 0
   __category_id = 1
   ...

   def __init(...):
      ....
```

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+            raise ValueError(f"detections_tensor should be of size [num_detections, 9], got {y_pred_img.shape}")
+
+        if len(y_img.shape) != 2 or y_img.shape[1] != 9:
+            raise ValueError(f"ground_truths should be of size [num_detections, 9], got {y_img.shape}")
+
+        y_pred_img = y_pred_img.to(self._device)
+        y_img = y_img.to(self._device)
+
+        categories = set()
+        categorywise_y: defaultdict = defaultdict(lambda: torch.zeros(0, 9).to(self._device))
```
sdesrozis, line 175: Why not allocate the tensor with the good shape ? and avoid to stack ?

sdesrozis, line 175: Could you improve the design to avoid the stack ?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+                        ious,
+                    )
+                )
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self.category_ids: set = set()
+        self.eval_imgs: defaultdict = defaultdict(list)
+
+    def _compute_iou(self, y: torch.Tensor, y_pred: torch.Tensor) -> torch.Tensor:
```
sdesrozis, line 206: Add a few comments to help the understanding please

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+        y_pred = y_pred[torch.argsort(y_pred[:, self.labels.d_confidence])]
+
+        crowd = [g[self.labels.i_crowd] for g in y]
+        ious = iou(
```
sdesrozis, line 210: why not use the full `y` and `y_pred` instead of extracting parts of the data ? It means having `iou` in the class and not as a generic and public method.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+            y[:, self.labels.i_xmin : self.labels.i_area], y_pred[:, self.labels.i_xmin : self.labels.i_area], crowd
+        ).to(self._device)
+
+        return ious
+
+    def _evaluate_image(
```
sdesrozis, line 216: The name of the method is not relevant. Choose a better name and add a few comments please.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+            "y_pred_ignore": y_pred_ignore,
+        }
+
+    def _accumulate(self) -> None:
+        num_iou_thr = len(self.iou_thresholds)
+        num_rec_thr = len(self.rec_thresholds)
+        num_categories = len(self.category_ids)
+        num_area = len(self.object_area_ranges)
+
+        max_det = self.num_detection_max
```
sdesrozis, line 290: useless

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+
+                    pr = pr.tolist()
+                    q = q.tolist()
+
+                    for i in range(nd - 1, 0, -1):
+                        if pr[i] > pr[i - 1]:
+                            pr[i - 1] = pr[i]
+
+                    inds = torch.searchsorted(rc, self.rec_thresholds, right=False)
+                    try:
```
sdesrozis, line 334: The exception seems weird. Could you explain ?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+                        for ri, pi in enumerate(inds):
+                            q[ri] = pr[pi]
+                    except:
+                        pass
+                    precision[t, :, c, a] = torch.tensor(q)
+        # return precision, recall
```
sdesrozis, line 340: Does this comment have any meaning ?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+        self.precision = precision
+
+    def _summarize(self, iou_thr: Optional[float] = None, area_rng: str = "all") -> float:
+        aind = [i for i, a_rng in enumerate(self.object_area_ranges) if a_rng == area_rng]
```
sdesrozis, line 344:
```python
range_index = list(self.object_area_ranges.keys()).index(area_range)
```
?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+
+        self.category_ids = all_category_ids
+
+    @sync_all_reduce()
+    def compute(self) -> Dict:
+        self._gather_all()
+        self._accumulate()
+
+        results = {
+            "all@0.5": self._summarize(iou_thr=0.5),
```
sdesrozis, line 402: if 0.5 and 0.75 are not values from tresholds, what happen ?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+        import torch.distributed as dist
+
+        def is_dist_avail_and_initialized() -> bool:
+            if not dist.is_available():
+                return False
+            if not dist.is_initialized():
+                return False
+            return True
+
+        if not is_dist_avail_and_initialized():
```
sdesrozis, line 370:
```python
if not dist.is_available():
    return
if not dist.is_initialized():
    return
```
Please keep it simple...

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,409 @@
...
+            return
+
+        eval_gather_dicts: List[Dict] = [defaultdict(list)] * dist.get_world_size()
+        dist.all_gather_object(eval_gather_dicts, self.eval_imgs)
+
+        category_gather_list: List[Set] = [set()] * dist.get_world_size()
+        dist.all_gather_object(category_gather_list, self.category_ids)
+
+        keys = set()
```
sdesrozis, line 379:
```python
combined_eval_imgs = defaultdict(list)
for eval_imgs in eval_gather_dicts:
    for key, value in eval_imgs.items():
        combined_eval_imgs[key] += value
```
What do you think ?

ignite/metrics/mean_average_precision.py
gucifer: why should all be excluded? We have to compute it atleast once.

ignite/metrics/mean_average_precision.py
gucifer: It improves code readability to pass only the bounding boxes. Otherwise it would become difficult to track what is being passed where, if a new person reads it.

gucifer: And for the public iou method, I can make it a static method inside MeanAveragePrecision class.

ignite/metrics/mean_average_precision.py
gucifer: Because of confidence sorting used in self._eval_images and self._compute_iou.

ignite/metrics/mean_average_precision.py
gucifer: We could, but since we already have crowd field in the input, better to check it here itself than checking it elsewhere first and passing None.

ignite/metrics/mean_average_precision.py
gucifer: I kept it as it is since it will get complicated....., did you check it again like we discussed on discord?

ignite/metrics/mean_average_precision.py
gucifer: If the recording threshold(self.rec_thresholds) doesn't exist in the recall(rc) tensor, it will return len(rc) + 1 for that value. So we use try and except to skip past that value.

ignite/metrics/mean_average_precision.py
gucifer: It will return -1.

ignite/metrics/mean_average_precision.py
sdesrozis: I mean crowd may not be an input. Typically, this could not be an input for a lot of datasets

ignite/metrics/mean_average_precision.py
sdesrozis: From https://github.com/pytorch/vision/blob/759c5b6dc2dfe46735bf724839f815b5cd79af09/torchvision/models/detection/mask_rcnn.py#L27

It would be simpler to handle a dict format rather than a plain tensor format. What do you think ? It would be easier to have optional crowd, ignore, etc.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,422 @@
...
+    Example:
+
+        .. code-block:: python
+
+            import torch
+            from ignite.metrics import MeanAveragePrecision
+
+            # Detection Format:
+            # (id, class, xmin, ymin, width, height, area, crowd, ignore/confidence)
+            # area: refers to the area covered by the bounding box, is usually
```
sdesrozis, line 85: `area` can be deduced from height and width, it can be removed.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,422 @@
...
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
+
+import torch
+
+from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
+
+__all__ = ["MeanAveragePrecision"]
+
+
+def _iou(y: torch.Tensor, y_pred: torch.Tensor, crowd: List) -> torch.Tensor:
```
sdesrozis, line 11: Does this function tested ?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,422 @@
...
+        self.eval_imgs: defaultdict = defaultdict(list)
+
+    def _compute_iou(self, y: torch.Tensor, y_pred: torch.Tensor) -> torch.Tensor:
+        # Sort predictions by confidence.
+        y_pred = y_pred[torch.argsort(-y_pred[:, self.__confidence])]
+
+        crowd = [g[self.__crowd] for g in y]
+
+        # Pass the bounding boxes for ground truths and predictions and crowd to obtain ious.y
+        ious = _iou(y[:, self.__xmin : self.__area], y_pred[:, self.__xmin : self.__area], crowd).to(self._device)
```
sdesrozis, line 219: why `y` and `y_pred` are not passed directly ?

ignite/metrics/mean_average_precision.py
gucifer: Compare it against pycocotools iou function in test?

tests/ignite/metrics/test_mean_average_precision.py
```diff
@@ -0,0 +1,306 @@
...
+    for i in cocoEval._dts:
+        for img in cocoEval._dts[i]:
+            id = img["id"]
+            img_id = img["image_id"]
+            class_id = img["category_id"]
+            area = img["area"]
+            confidence = img["score"]
+            xmin, ymin, xmax, ymax = img["bbox"]
+            img_dts[img_id] = torch.vstack(
+                [img_dts[img_id], torch.tensor([id, class_id, xmin, ymin, xmax, ymax, confidence, area, crowd])]
```
sdesrozis, line 64: `crowd` parameter never initialized

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,447 @@
...
+
+        if iou_thresholds is None:
+            iou_thresholds = [0.5, 0.55, 0.6, 0.65, 0.7, 0.75, 0.8, 0.85, 0.9, 0.95]
+        self.iou_thresholds = torch.tensor(iou_thresholds, device=device)
+
+        if rec_thresholds is None:
+            rec_thresholds = [i / 100 for i in range(101)]
+        self.rec_thresholds = torch.tensor(rec_thresholds, device=device)
+
+        self.__id = 0
```
sdesrozis, line 152: It is a parameter from user side. We can't use it...

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,440 @@
...
+    Example:
+
+        .. code-block:: python
+
+            import torch
+            from ignite.metrics import MeanAveragePrecision
+
+            # Input Format:
+            # Ground Truth
+            # [
```
sdesrozis, line 88: Sizes of tensor (ie B) have no sense. We talked about a list of dict, one per image. So image_id should be an int (or a tensor of size one), etc.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,440 @@
...
+                )
+
+    @reinit__is_reduced
+    def update(self, outputs: Tuple[Dict, Dict]) -> None:
+
+        for output in outputs:
+
+            y_pred_img, y_img = output
+
+            assert y_img["image_id"] == y_pred_img["image_id"]
```
sdesrozis, line 186: We don’t use assert in ignite, but exception.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,440 @@
...
+                sorted_y_pred_bbox = y_pred_bbox[torch.argsort(-y_pred_img["score"][y_pred_ind])]
+
+                y_ind = y_category_dict[category]
+                y_id = y_img["id"][y_ind]
+                y_bbox = y_img["bbox"][y_ind]
+                crowd = y_img["iscrowd"][y_ind] if "iscrowd" in y_img else None
+                y_ignore = y_img["ignore"][y_ind] if "ignore" in y_img else torch.zeros(len(y_ind))
+                y_area = y_img["area"][y_ind] if "area" in y_img else (y_img["bbox"][:, 2] * y_img["bbox"][:, 3])[y_ind]
+
+                ious = _iou(y_bbox, sorted_y_pred_bbox, crowd).to(self._device)
```
sdesrozis, line 220: I don’t see why y_pred’s boxes should be sorted by score. Moreover, that is not mentioned in the iou method.

sdesrozis:
I move to draft instead of PR. A lot of work remaining. Maybe it could worth to discuss, implement and test on our side to create a new clean PR. We will see after GSoC.

ignite/metrics/mean_average_precision.py
gucifer: iou method is generic, for 'a' y_bbox and 'b' y_pred_bbox it return a matrix of size 'a x b' containing all the respective ious. We are passing it sorted list because in eval_image function we are sorting y_preds according to confidence as well for greedy matching.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,441 @@
...
+            if len(area_range) != 2 or area_range[0] >= area_range[1]:
+                raise ValueError(
+                    f"object_area_ranges must be a dict associating to each key (str) a tuple of float values (a, b) "
+                    f"where a < b (got: key={area}, value={area_range})"
+                )
+
+    @reinit__is_reduced
+    def update(self, outputs: Tuple[Dict, Dict]) -> None:
+
+        for output in outputs:
```
sdesrozis, line 182: `outputs` is defined as a tuple but seems considered as a list.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,441 @@
...
+
+            y_pred_img, y_img = output
+
+            if y_img["image_id"] != y_pred_img["image_id"]:
```
sdesrozis, line 186: A test is required.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,441 @@
...
+                raise ValueError("Ground Truth and Predictions should be for the same image.")
+
+            if y_img["image_id"] in self.image_ids:
```
sdesrozis, line 189: A test is required

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,441 @@
...
+        self, y: List, y_pred: List, area_rng: Tuple[float, float], ious: torch.Tensor
+    ) -> Optional[Dict]:
+        """
+            Evaluates iou_threshold wise y and y_pred matches.
+        """
+
+        y_id, y_area, y_ignore, y_crowd = y
+        y_pred_id, y_pred_area, y_pred_score = y_pred
+
+        if len(y_id) == 0 and len(y_pred_id) == 0:
```
sdesrozis, line 249: A test is required

ignite/metrics/mean_average_precision.py
sdesrozis: I don’t see where it matters…

vfdev-5:
Close as done in https://github.com/pytorch/ignite/pull/2901

Issue #2134, opened 2021-08-04
Move `LRFinder` and `ParamScheduler` tests


KickItLikeShika:
We already moved `LRFinder` and `ParamScheduler` form contrib to core at #1988 and #1998, however we missed moving the tests to core either. So we need to move the tests.

PR #2135, opened 2021-08-04
Moved `LRFinder` tests to core

KickItLikeShika:
Fixes part of #2134 

Moved `LRFinder` tests from contrib to core via `git mv`

Check list:


vfdev-5 (approved):
Thanks

PR #2137, opened 2021-08-04
Updated BaseOutputHandler to accept state attributes - Tensorboard

Ishan-Kumar2:
Addresses #2100  (using "fixes" will close the issue)

Description:

Adds feature to log state attributes in Tensorboard 

Check list:

- [X] New tests are added (if a new feature is added)
- [X] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #2140, opened 2021-08-05
Moved metrics data_type dispatch to base class

Ishan-Kumar2:
Description:
Moves the data type check and dispatch from derived class `OutputHandler` to the base class function `_setup_output_metrics` as mentioned in #2137 
This PR handles only for Tensorboard, the derived class part has to be removed in a similar manner for all other trackers.
 
Check list:


vfdev-5:
@Ishan-Kumar2 thanks a lot for the PR ! It looks good. Could you please make the same changes for other loggers like Visdom, ClearML, etc ? Such that this idea is applied for all loggers.

ignite/contrib/handlers/base_logger.py
```diff
@@ -103,7 +103,26 @@ def _setup_output_metrics(self, engine: Engine) -> Dict[str, Any]:
...
+                continue
+            if isinstance(value, numbers.Number):
+                metrics_dict[f"{self.tag}/{name}"] = value
+            elif isinstance(value, torch.Tensor) and value.ndimension() == 0:
+                metrics_dict[f"{self.tag}/{name}"] = value.item()
+            elif isinstance(value, torch.Tensor) and value.ndimension() == 1:
+                for i, v in enumerate(value):
+                    metrics_dict[f"{self.tag}/{name}/{i}"] = v.item()
+            else:
+                if isinstance(value, str) and log_text:
```
Ishan-Kumar2, line 120: Since only some loggers allow text, I have handled it using a flag like this. Let me know if there is a better way for doing this.

Ishan-Kumar2:
@vfdev-5 I have made the changes for the rest of the loggers (and some tests) too, let me know if something needs to be changed :)

ignite/contrib/handlers/base_logger.py
```diff
@@ -103,7 +103,26 @@ def _setup_output_metrics(self, engine: Engine) -> Dict[str, Any]:
...
+                    metrics_dict[f"{self.tag}/{name}"] = value
+                else:
+                    warnings.warn(f"Logger output_handler can not log metrics value type {type(value)}")
```
Ishan-Kumar2, line 123: Here since we dont have the name of the logger in the base class I have raised a generic warning using the term "Logger". Would it be better to pass the name of the Logger to the base class and have that in the warning?

ignite/contrib/handlers/tqdm_logger.py
```diff
@@ -270,23 +266,17 @@ def __call__(self, engine: Engine, logger: ProgressBar, event_name: Union[str, E
...
-                warnings.warn(f"ProgressBar can not log tensor with {value.ndimension()} dimensions")
-
-        if rendered_metrics:
-            logger.pbar.set_postfix(rendered_metrics)  # type: ignore[attr-defined]
+        rendered_metrics = self._setup_output_metrics(engine, log_text=True)
+        metrics = OrderedDict()
+        for key, value in rendered_metrics.items():
+            keys_list = key.split("/")
+            keys_list.pop(0)  # tqdm has tag as description
+            key = "_".join(keys_list)
```
vfdev-5, line 274: I was wondering if it would be better instead of doing split by "/" to have keys as tuple of `(tag, name)` or `(tag, name, i)` such that we can create appropriate string for each logger by `"/".join(key)` ?

ignite/contrib/handlers/base_logger.py
vfdev-5: I think it is OK, thanks !

ignite/contrib/handlers/base_logger.py
vfdev-5: I think we can use `self.__class__.__name__` instead of Logger.

ignite/contrib/handlers/tqdm_logger.py
Ishan-Kumar2: Yeah, can do that although since most of them follow the "/" separated convention (like Polyaxon, visdom and wandb) the dict can be directly passed to the logger without the need of any loops in the derived class. If the uniform tuple method is more useful I can go with that.

vfdev-5: Yes, I also thought about that, but splitting the key looks like patching. Let's try to use tuples.

ignite/contrib/handlers/base_logger.py
Ishan-Kumar2: That won't be helpful as the name of the derived class is OutputHandler in all cases (not Logger name) right?

ignite/contrib/handlers/tqdm_logger.py
Ishan-Kumar2: Yeah, fixed!

ignite/contrib/handlers/base_logger.py
```diff
@@ -103,7 +103,25 @@ def _setup_output_metrics(self, engine: Engine) -> Dict[str, Any]:
                 output_dict = {"output": output_dict}
 
             metrics.update({name: value for name, value in output_dict.items()})
-        return metrics
+
+        metrics_dict = {}  # type: Dict[Tuple[str, ...], Union[str, float, numbers.Number]]
+
+        for name, value in metrics.items():
+            if value is None:
+                continue
```
vfdev-5, line 111: Value can not be None, I think

ignite/contrib/handlers/base_logger.py
vfdev-5: OK, let's keep Logger.

ignite/contrib/handlers/mlflow_logger.py
```diff
@@ -231,30 +229,26 @@ def __call__(self, engine: Engine, logger: MLflowLogger, event_name: Union[str,
...
-            else:
-                warnings.warn(f"MLflowLogger output_handler can not log metrics value type {type(value)}")
-
         # Additionally recheck metric names as MLflow rejects non-valid names with MLflowException
         from mlflow.utils.validation import _VALID_PARAM_AND_METRIC_NAMES
 
-        for key in list(rendered_metrics.keys()):
+        for keys in list(rendered_metrics.keys()):
+            key = " ".join(keys)
+            print(key, keys)
```
vfdev-5, line 237: Please remove `print`

ignite/contrib/handlers/mlflow_logger.py
```diff
@@ -231,30 +229,26 @@ def __call__(self, engine: Engine, logger: MLflowLogger, event_name: Union[str,
...
                     "will be ignored and not logged to MLflow"
                 )
-                del rendered_metrics[key]
+                del rendered_metrics[keys]
+
+        metrics = {}
+        for key, value in rendered_metrics.items():
+            print(key)
+            key = " ".join(key)
+            metrics[key] = value
```
vfdev-5, line 249: Let's do this step before the previous one where we can remove invalid keys. Such that we call `key = " ".join(key)` once

ignite/contrib/handlers/wandb_logger.py
```diff
@@ -279,21 +276,12 @@ def __call__(self, engine: Engine, logger: WandBLogger, event_name: Union[str, E
...
-                    for i, v in enumerate(value):
-                        rendered_metrics[f"{self.tag}/{name}/{i}"] = v.item()
-                else:
-                    warnings.warn(f"WandBLogger output_handler can not log metrics value type {type(value)}")
-
-        logger.log(rendered_metrics, step=global_step, sync=self.sync)
+        rendered_metrics = self._setup_output_metrics(engine, log_text=True)
+        metrics = {}
+        for key, value in rendered_metrics.items():
+            metrics["/".join(key)] = value
```
vfdev-5, line 282: Actually, I just realized that we could introduce another flag like `log_text` and output keys either as tuples or as a string with "/". Such that we can simplify the code for TB, W&B and other similar loggers.

Ishan-Kumar2: Thats a good idea, I have added it

ignite/contrib/handlers/mlflow_logger.py
Ishan-Kumar2: Yeah makes sense, fixed!

ignite/contrib/handlers/mlflow_logger.py
Ishan-Kumar2: Sorry for that

ignite/contrib/handlers/base_logger.py
```diff
@@ -103,7 +105,35 @@ def _setup_output_metrics(self, engine: Engine) -> Dict[str, Any]:
...
-        return metrics
+
+        metrics_dict = {}  # type: Dict[Any, Union[str, float, numbers.Number]]
+
+        for name, value in metrics.items():
+            if isinstance(value, numbers.Number):
+                if key_tuple:
+                    metrics_dict[(self.tag, name)] = value
+                else:
+                    metrics_dict[f"{self.tag}/{name}"] = value
```
vfdev-5, line 116: Let's define a key transforming function once, like here:
```python
key_tuple = False

if key_tuple:
    key_tf = lambda tag, name, *args: (tag, name) + args
else:
    key_tf = lambda tag, name, *args: "/".join((tag, name) + args)
    
key_tf("tag", "name"), key_tf("tag", "name", "1")
> ('tag/name', 'tag/name/1')
> (('tag', 'name'), ('tag', 'name', '1'))
```
and only use it without multiple if-else.

Ishan-Kumar2: Yup, thats much better. I have changed

ignite/contrib/handlers/base_logger.py
```diff
@@ -103,7 +105,33 @@ def _setup_output_metrics(self, engine: Engine) -> Dict[str, Any]:
...
+
+        if key_tuple:
+
+            def key_tf(tag, name, *args):
+                return (tag, name) + args
+
+        else:
+
+            def key_tf(tag, name, *args):
+                return "/".join((tag, name) + args)
```
vfdev-5, line 119: Can't we use lambda functions ?

Ishan-Kumar2: I tried with that, flake8 gives this [warning](https://www.flake8rules.com/rules/E731.html).
Should I ignore that and add lambda only?

vfdev-5: Yes, you are right. Let's do that like that:
```python
def key_tuple(tag, name, *args):
    return (tag, name) + args

def key_str(tag, name, *args):
    return "/".join((tag, name) + args)

key_tf = key_tuple if key_tuple else key_str
```

Ishan-Kumar2: Yes thats better, fixed it.

tests/ignite/contrib/handlers/test_base_logger.py
```diff
@@ -58,34 +58,34 @@ def test_base_output_handler_setup_output_metrics():
 
     # Only metric_names
     handler = DummyOutputHandler("tag", metric_names=["a", "b"], output_transform=None)
-    metrics = handler._setup_output_metrics(engine=engine)
-    assert metrics == true_metrics
+    metrics = handler._setup_output_metrics(engine=engine, key_tuple=False)
+    assert metrics == {"tag/a": 0, "tag/b": 1}
```
vfdev-5, line 62: We can remove unused `true_metrics` definition, line 55

vfdev-5 (commented):
Thanks for the update @Ishan-Kumar2 
Just a nit fix and good to go

tests/ignite/contrib/handlers/test_base_logger.py
Ishan-Kumar2: Its being used in line 56 for passing in the metrics to the `State` right now. Should I remove it as a variable and directly pass the dictionary?

Ishan-Kumar2: Also, I added the type annotations to the new functions which was causing the check to fail.

vfdev-5 (approved):
Thanks a lot @Ishan-Kumar2 !
LGTM

PR #2147, opened 2021-08-09
Refactored Engine

KickItLikeShika:
The 4th PR to tackle refactoring the Engine: https://github.com/pytorch/ignite/wiki/Roadmap#refactor-engine

In this PR we refactored Engine and made some other necessary changes.

Check list:


vfdev-5:
@KickItLikeShika have incorporated all requested changes from the closed PR ?

docs/source/concepts.rst
```diff
@@ -266,22 +266,15 @@ These events could be used to attach any handler and are fired using :meth:`~ign
 If you want to use filtering with custom events (e.g. ``CustomEvents.CUSTOM_STARTED(every=5)``), you need to do 3 more things:
 
 - ``engine.state`` should have corresponding attributes for the events, e.g. ``engine.state.custom_started``
-- you need to pass a dict `event_to_attr` to :meth:`~ignite.engine.engine.Engine.register_events`, which maps between events and state attributes, e.g.
+- you need to pass a dict `attr_to_events` to :meth:`~ignite.engine.engine.Engine.register_events`, which maps between state attributes and a list of events, when setting the attribute to any value, all the counters for the events in the corresponding list will be updated accordingly. For getting the attirubte value, the counter value of the first event in the corresponding list will be returned automatically.
```
vfdev-5, line 269:
```suggestion
- you need to pass a dict `attr_to_events` to :meth:`~ignite.engine.engine.Engine.register_events`, which maps between state attributes and a list of events, when setting the attribute to any value, all the counters for the events in the corresponding list will be updated accordingly. For getting the attribute value, the counter value of the first event in the corresponding list will be returned automatically.
```

ignite/contrib/handlers/base_logger.py
```diff
@@ -172,17 +172,17 @@ def attach(
         Returns:
             :class:`~ignite.base.events.RemovableEventHandle`, which can be used to remove the handler.
         """
+
         if isinstance(event_name, EventsList):
             for name in event_name:
-                if name not in State.event_to_attr:
+                if name not in engine._allowed_events:
```
vfdev-5, line 178: Do we have a public method to check that ?

ignite/engine/engine.py
```diff
@@ -157,6 +167,10 @@ def register_events(
         Args:
             event_names: Defines the name of the event being supported. New events can be a str
                 or an object derived from :class:`~ignite.base.events.EventEnum`. See example below.
+            attr_to_events: mapping consists of the state attributes mapped to a list events from
+                :class:`~ignite.engine.events.Events` or any other custom events added
+                by :meth:`~ignite.base.events_driven.EventsDriven.register_events`.
+                Getting attribute values is done based the on first element in the list of the events.
             event_to_attr: A dictionary to map an event to a state attribute.
```
vfdev-5, line 174: Update the docstring about the deprecation

ignite/engine/engine.py
```diff
@@ -207,45 +221,36 @@ class TBPTT_Events(EventEnum):
                 TIME_ITERATION_STARTED = "time_iteration_started"
                 TIME_ITERATION_COMPLETED = "time_iteration_completed"
 
-            TBPTT_event_to_attr = {
-                TBPTT_Events.TIME_ITERATION_STARTED: 'time_iteration',
-                TBPTT_Events.TIME_ITERATION_COMPLETED: 'time_iteration'
+            TBPTT_attr_to_events = {
+                'time_iteration': [TBPTT_Events.TIME_ITERATION_STARTED,
+                                    TBPTT_Events.TIME_ITERATION_COMPLETED]
```
vfdev-5, line 226:
```suggestion
                'time_iteration': [
                    TBPTT_Events.TIME_ITERATION_STARTED,
                    TBPTT_Events.TIME_ITERATION_COMPLETED
                ]
```

ignite/engine/state.py
```diff
@@ -64,18 +79,34 @@ def __init__(self, **kwargs: Any) -> None:
...
+        # epoch and iteration included. We do this to keep BC compatibility,
+        # as we now don't keep iteration and epoch here in state,
+        # instead of that, we use engine._allowed_events_counts to keep track of them.
+        if self.engine is None:
+            self._update_attrs()
+
+    # @property
+    # def event_to_attr(self) -> event_to_attr:
+    #     warnings.warn("'event_to_attr' is deprecated, please use 'attr_to_evets' instead.")
+    #     return State.event_to_attr
```
vfdev-5, line 93: If this does not work, let's remove that.

KickItLikeShika:
Yes I did, only this one left for more investigation https://github.com/pytorch/ignite/pull/2128#discussion_r684786098

ignite/engine/state.py
```diff
@@ -64,18 +79,34 @@ def __init__(self, **kwargs: Any) -> None:
...
-            if not hasattr(self, value):
-                setattr(self, value, 0)
+        for key in self.attr_to_events.keys():
+            if not hasattr(self, key):
+                setattr(self, key, 0)
 
     def get_event_attrib_value(self, event_name: Union[str, Events, CallableEventWithFilter]) -> int:
         """Get the value of Event attribute with given `event_name`."""
-        if event_name not in State.event_to_attr:
+        event_to_attr = {}
```
vfdev-5, line 102: Shouldn't we deprecate this method as well ?

KickItLikeShika: I think we should leave this till we find a replacement, check this please https://github.com/pytorch/ignite/pull/2128#discussion_r684754648

vfdev-5: I think, this method can go to EventsDrivenState and fetch the value from Engine. If no engine, raise an error.

ignite/contrib/handlers/base_logger.py
KickItLikeShika: That's already tested everywhere in `tests/ignite/contrib/engine/test_common` and other places also

ignite/engine/state.py
KickItLikeShika: Thanks for this comment, this comment pointed me to a dangerous bug, I will try to explain it here
If something like this happened 
```py
engine = Engine(...)
engine.state = State(...)
```
normally we set the new state attributes only, and set engine to self. But the bug was that:
- `state._attr_to_events` is now None, which is wrong.
- `allowed_events_count` is not set to the new values of the attributes.
I discovered that when I was trying to get move this method to `EventsDrivenState` and get the values vai `allowed_events_counts` to drop `event_to_attr` totally. I'm working on fixing this now.

KickItLikeShika: My thoughts on this:
- for fixing the `state._attr_to_evets` issue, I'm thinking about passing the default `attr_to_events` driven in `super()` while initializing state if `state._attr_to_events` is None.
- for fixing `allowed_events_counts` I'm trying different solutions now.

KickItLikeShika: Almost solved it, another point about changing the code `get_event_attrib_value`, We have to keep the code the same way for now, in all tests we mock the engine, so `state.engine` is none, so we can't use `allowed_events_..` to get anything.

ignite/contrib/handlers/base_logger.py
vfdev-5: In tests it is OK, but in the code it is a bit different. It is only used in 3 places: https://github.com/pytorch/ignite/search?q=_allowed_events . OK, for now to use it like that. Maybe, later we can create a public method for that.

ignite/engine/engine.py
```diff
@@ -123,28 +123,42 @@ def compute_mean_std(engine, batch):
...
+        return self._state
+
+    @state.setter
+    def state(self, new_state: State) -> None:
+        self._state = new_state
+        self._state.engine = self
+        for k, v in self._state.kwargs.items():
+            setattr(self._state, k, v)
+            if k == "epoch" or k == "iteration":
+                delattr(self._state, k)
```
vfdev-5, line 155: This part of the code looks strange. Please, put a comment in the code and explain here, why we are doing that.

KickItLikeShika: Added a comment here anyway, I agree that I should clarify that.

vfdev-5: I think, we have to "repeat" the same steps as we do in the constructor:
```python
self._state = State(engine=self)
self.register_events(*Events, attr_to_events=State.attr_to_events)
```

ignite/engine/state.py
```diff
@@ -45,9 +61,9 @@ class State:
     }  # type: Dict[Union[str, "Events", "CallableEventWithFilter"], str]
 
     def __init__(self, **kwargs: Any) -> None:
+        super(State, self).__init__(attr_to_events=State.attr_to_events, **kwargs)
 
-        self.iteration = 0
-        self.epoch = 0
+        self.kwargs = kwargs
```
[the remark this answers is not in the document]
KickItLikeShika: Found a better way using `stata.__dict__`, I tried this before but there where some errors, I solved it doing 
`for k, v in zip(list(self._state.__dict__.keys()), list(self._state.__dict__.values())):`
here is the solution
```py
        for k, v in zip(list(self._state.__dict__.keys()), list(self._state.__dict__.values())):
            if k in self._state._attr_to_events.keys():
                setattr(self._state, k, v)
                delattr(self._state, k)
```

ignite/engine/engine.py
```diff
@@ -123,28 +123,46 @@ def compute_mean_std(engine, batch):
...
+        self._state = new_state
+        self._state.engine = self
+        # After setting state._attr_to_events and engine we have to set
+        # engine._allowed_events_counts values according the added kwargs,
+        # also remove epoch and iteration if added as attributes,
+        # as we use engine._allowed_events_counts to get their values
+        for k, v in self._state.kwargs.items():
+            setattr(self._state, k, v)
+            if k in self._state._attr_to_events.keys():
+                delattr(self._state, k)
```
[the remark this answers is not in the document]
KickItLikeShika: How overriding `state.engine` setter can sync automatically? what we can do here is to override `state.engine` setter and divide this code into two places, this part of syncing the attributes and counters can be moved to `state.engine` setter.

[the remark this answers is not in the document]
vfdev-5: > How overriding state.engine setter can sync automatically? 

Like that : 
```python
    @state.setter
    def state(self, new_state: State) -> None:
        self._state = new_state
        self._state.engine = self
```
and the class of `_state` implements engine as a property and sync required part internally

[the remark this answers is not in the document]
KickItLikeShika: Done.

ignite/engine/state.py
```diff
@@ -64,18 +79,49 @@ def __init__(self, **kwargs: Any) -> None:
...
+        if self._engine is None:
+            self._update_attrs()
+
+    @property
+    def engine(self) -> Optional[EventsDriven]:
+        return self._engine
+
+    @engine.setter
+    def engine(self, new_engine: EventsDriven) -> None:
+        self._engine = new_engine
```
vfdev-5, line 96: Any reason to keep it here and not in its base class EventsDrivenState ?

ignite/base/events_driven.py
```diff
@@ -257,61 +257,143 @@ class EventsDrivenState:
...
+
+                def register_events(self, *event_names, attr_to_events=None) -> None:
+                    super(ToyEngine, self).register_events(*event_names)
+                    if attr_to_events is not None:
+                        for attribute, events in attr_to_events.items():
+                            self._state.update_attribute_mapping(attribute, events)
+
+            toy_engine = ToyEngine()
+            toy_engine.state.beta = 80
+            # toy_engine._allowed_events_counts is synchronized automatically
```
vfdev-5, line 387: Let's avoid exposing private attributes.

ignite/engine/engine.py
```diff
@@ -1,24 +1,24 @@
...
+from ignite.base.mixins import Serializable
+from ignite.engine.events import Events
 from ignite.engine.state import State
 from ignite.engine.utils import _check_signature, _to_hours_mins_secs
 
 __all__ = ["Engine"]
 
 
-class Engine(Serializable):
+class Engine(Serializable, EventsDriven):
```
vfdev-5, line 21: Can't we now inherit from `EventsDrivenWithState` ?

KickItLikeShika: Yes but we can't remove the property because of the setter, I will inherit from `EvetsDrivenWithState` but not changes in the code, after forbidding setting state we can remove the property and the setter in `Engine`.

ignite/base/events_driven.py
```diff
@@ -127,6 +124,9 @@ def add_event_handler(self, event_name: Any, handler: Callable, *args: Any, **kw
 
         self._assert_allowed_event(event_name)
 
+        # TODO: remove this import after refactroing Events
```
vfdev-5, line 127:
```suggestion
        # TODO: remove this import after refactoring Events
```

ignite/engine/state.py
```diff
@@ -64,18 +79,29 @@ def __init__(self, **kwargs: Any) -> None:
...
-        if event_name not in State.event_to_attr:
+        event_to_attr = {}
+        for attribute, events in self._attr_to_events.items():
+            for event in events:
+                event_to_attr[event] = attribute
+
+        if event_name not in event_to_attr:
             raise RuntimeError(f"Unknown event name '{event_name}'")
-        return getattr(self, State.event_to_attr[event_name])
+        return getattr(self, event_to_attr[event_name])  # type: ignore
```
vfdev-5, line 104: I think `get_event_attrib_value` can go to `EventsDrivenState` and be implemented like:
```python
    def get_event_attrib_value(self, event_name) -> int:
        if event_name not in self.engine._allowed_events_counts:
            raise RuntimeError(f"Unknown event name '{event_name}'")        
        return self.engine._allowed_events_counts[event_name]
```

Issue #2100, opened 2021-07-06
Update BaseOutputHandler to accept state attributes


vfdev-5:
## 🚀 Feature

Here is the feature I would like to have, for example, with TensorboardLogger, but we can implement that for all exp tracking systems:
```python
tb_logger = TensorboardLogger(log_dir="experiments/tb_logs")

tb_logger.attach_output_handler(
    trainer,
    event_name=Events.ITERATION_COMPLETED,
    tag="training",
    state_attrs=["alpha", "beta"]  # <->  trainer.state.alpha and trainer.state.beta
)
```
where `state_attrs` defines some attributes from `trainer.state` to log to TB.

Currently, this could be done as 
```python
tb_logger = TensorboardLogger(log_dir="experiments/tb_logs")

def trainer_state_logger(state_attrs, tag=None):

    def wrapper(engine, logger, event_name):
        assert engine == trainer

        global_step = engine.state.get_event_attrib_value(event_name)
        tag_prefix = f"{tag}/" if tag else ""
        for name in state_attrs:
            value = getattr(engine, name, None)
            if value is None:
                continue
            logger.writer.add_scalar(f"{tag_prefix}{name}", value, global_step)
    
tb_logger.attach(
    trainer,
    event_name=Events.ITERATION_COMPLETED,
    log_handler=trainer_state_logger(["alpha", "beta"], tag="training")
)
```

We would like to support state attributes types same as for metrics:
See https://github.com/pytorch/ignite/blob/4d6d220e7be7a433cf2e16f92ca7cc6969ecea89/ignite/contrib/handlers/tensorboard_logger.py#L287-L296

This issues can be handled in several PRs that should implement the feature for the following exp tracking systems and loggers:

- [x] TensorBoard logger
- [x] Visdom logger
- [x] ClearML logger
- [x] Neptune logger
- [x] WandB logger
- [x] MLFlow logger
- [x] Polyaxon logger
- [ ] tqdm logger

If you would like work on this issue, please comment out and say which part you would like to cover.
A PR should contain new code, tests and documentation updates. Please see our contributing guide: https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md
A draft PR can be sent and maintainers can help to iterate over the remaining tasks.

Ishan-Kumar2:
[...]

PR #2148, opened 2021-08-09
Added state_attributes for Visdom logger

Ishan-Kumar2:
Addresses #2100 

Description:
Adds feature to log state attributes in Visdom

Check list:


vfdev-5 (approved):
Thanks @Ishan-Kumar2 , looks good !

PR #2151, opened 2021-08-10
Added state attributes from ClearML logger

Ishan-Kumar2:
Addresses #2100 

Description:
Adds feature to log state attributes in ClearML
Check list:

### DIFF
--- a/ignite/contrib/handlers/clearml_logger.py
+++ b/ignite/contrib/handlers/clearml_logger.py
@@ -269,6 +269,22 @@ class OutputHandler(BaseOutputHandler):
                 global_step_transform=global_step_transform
             )
 
+        Another example where the State Attributes ``trainer.state.alpha`` and ``trainer.state.beta``
+        are also logged along with the NLL and Accuracy after each iteration:
+
+        .. code-block:: python
+
+            clearml_logger.attach(
+                trainer,
+                log_handler=OutputHandler(
+                    tag="training",
+                    metric_names=["nll", "accuracy"],
+                    state_attributes=["alpha", "beta"],
+                ),
+                event_name=Events.ITERATION_COMPLETED
+            )
+
+
     Args:
         tag: common title for all produced plots. For example, "training"
         metric_names: list of metric names to plot or a string "all" to plot all available
@@ -282,6 +298,7 @@ class OutputHandler(BaseOutputHandler):
             Default is None, global_step based on attached engine. If provided,
             uses function output as global_step. To setup global step from another engine, please use
             :meth:`~ignite.contrib.handlers.clearml_logger.global_step_from_engine`.
+        state_attributes: list of attributes of the ``trainer.state`` to plot.
 
     Note:
         Example of `global_step_transform`:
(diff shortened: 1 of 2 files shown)


ignite/contrib/handlers/clearml_logger.py
```diff
@@ -282,6 +298,7 @@ def global_step_transform(*args, **kwargs):
             Default is None, global_step based on attached engine. If provided,
             uses function output as global_step. To setup global step from another engine, please use
             :meth:`~ignite.contrib.handlers.clearml_logger.global_step_from_engine`.
+        state_attributes: list of attributes of the ``trainer.state`` to plot.
```
vfdev-5, line 301: Please add `versionchanged` after the Note as explained here: https://github.com/pytorch/ignite/pull/2137#issuecomment-895112574

vfdev-5 (commented):
Thanks for the PR ! 
I left a comment about updating docstring

ignite/contrib/handlers/clearml_logger.py
Ishan-Kumar2: I have added ``versionchanged`` here, if it is fine I'll update the other two also.

vfdev-5: No, you added it in the wrong place. It should be in the end of the docstring and make sure to respect indents like here: https://github.com/pytorch/ignite/blob/c6ff5b8533e4c9b0dbd0c2b89606edf9a1ea8647/ignite/handlers/stores.py#L38-L39

ignite/contrib/handlers/clearml_logger.py
```diff
@@ -269,6 +269,25 @@ def global_step_transform(*args, **kwargs):
                 global_step_transform=global_step_transform
             )
 
+        ..  versionchanged:: 0.5.1
+        OutputHandler now accepts an optional list of `state_attributes`
+        which are attributes of the ``trainer.state`` to be logged.
```
vfdev-5, line 274:
```suggestion
        ..  versionchanged:: 0.5.1
            now accepts an optional list of `state_attributes`        
```

ignite/contrib/handlers/clearml_logger.py
```diff
@@ -291,6 +307,9 @@ def global_step_transform(*args, **kwargs):
             def global_step_transform(engine, event_name):
                 return engine.state.get_event_attrib_value(event_name)
 
+    ..  versionchanged:: 0.5.0
+        Now accepts an optional list of `state_attributes`
+        which are attributes of the ``trainer.state`` to be logged.
```
vfdev-5, line 312:
```suggestion
        accepts an optional list of `state_attributes`
```

ignite/contrib/handlers/clearml_logger.py
```diff
@@ -291,6 +307,8 @@ def global_step_transform(*args, **kwargs):
             def global_step_transform(engine, event_name):
                 return engine.state.get_event_attrib_value(event_name)
 
+    ..  versionchanged:: 0.5.0
+        Accepts an optional list of `state_attributes`
```
vfdev-5, line 311:
```suggestion
        accepts an optional list of `state_attributes`
```

Issue #2101, opened 2021-07-09
Add missing distributed tests to core handlers


vfdev-5:
## 🚀 Feature

- Early stopping:

Currently, early stopping tests does not run any distributed tests with horovod:
https://github.com/pytorch/ignite/blob/a75d228ddae9558f925384bf3a199832c78de8c3/tests/ignite/handlers/test_early_stopping.py#L336-L352

ToDo:
- [x] add hvd distributed test similarly to 
https://github.com/pytorch/ignite/blob/a75d228ddae9558f925384bf3a199832c78de8c3/tests/ignite/metrics/test_accuracy.py#L576-L588

- [x] add xla tests similarly to 
https://github.com/pytorch/ignite/blob/a75d228ddae9558f925384bf3a199832c78de8c3/tests/ignite/metrics/test_accuracy.py#L591-L617

- Checkpoint:

Same for Checkpoint tests
https://github.com/pytorch/ignite/blob/a75d228ddae9558f925384bf3a199832c78de8c3/tests/ignite/handlers/test_checkpoint.py#L1190-L1212

ToDo:
- [ ] add hvd distributed test similarly to 
https://github.com/pytorch/ignite/blob/a75d228ddae9558f925384bf3a199832c78de8c3/tests/ignite/metrics/test_accuracy.py#L576-L588

If you would like work on this issue, please comment out and say which part you would like to cover.
A PR should contain new code, tests and documentation updates. Please see our contributing guide: https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md
A draft PR can be sent and maintainers can help to iterate over the remaining tasks.

Ishan-Kumar2:
Interested in this, will raise a PR soon!

vfdev-5:
@Ishan-Kumar2 it remains adding hvd tests to Checkpoint. Would you like to help with that as well to finalize this issue ?

Ishan-Kumar2:
@vfdev-5 sure, I will raise a PR soon.

PR #2165, opened 2021-08-16
Added distributed tests with Horovod and XLA for early_stopping

Ishan-Kumar2:
Addresses #2101

Description:
Added distributed tests with Horovod and XLA for early_stopping.

Check list:

### DIFF
--- a/docs/source/about.rst
+++ b/docs/source/about.rst
@@ -10,16 +10,15 @@ This project was started by Alykhan Tejani `@alykhantejani <https://github.com/a
 Francisco Massa `@fmassa <https://github.com/fmassa>`_ and Joost van Amersfoort `@y0ast <https://github.com/y0ast>`_
 in the end of 2017 and was highly inspired by `torchnet <https://github.com/pytorch/tnt>`_.
 
-In 2018 project core teams has grown and the first public release was published in June 2018.
-Since then, several releases have appeared following a ~5 month cycle, and a thriving international community has
-been leading the development.
+In 2018, the project core team had grown and the first public release was published in June 2018.
+Since then, several releases have appeared following a ~5 month cycle, with a thriving international community leading the development.
 
 Governance
 ----------
 PyTorch-Ignite is a `NumFOCUS Affiliated Project <https://numfocus.org/sponsored-projects/affiliated-projects>`_ and 
 open-source community effort. 
-Project's affiliation to `PyTorch organization <https://github.com/pytorch>`_ is due to historical reasons and it 
-has the same implication to the project's governance as for other community projects from
+The Project's affiliation to `PyTorch organization <https://github.com/pytorch>`_ is due in large part to historical reasons and it 
+has the same implication to the project's governance as the other community projects from the
 `PyTorch Ecosystem <https://pytorch.org/ecosystem/>`_.
 
 The decision making process and governance structure of the project is described in the :ref:`governance document<PyTorch-Ignite governance>`.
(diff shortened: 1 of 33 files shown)


sdesrozis:
@Ishan-Kumar2 Thank you very much for this PR !!

However, it seems that it does not work perfectly. Feel free to ask help if needed.

Ishan-Kumar2:
Hi @sdesrozis, I have made a change (adding device definition for None case). Was relying on the checks to run to see if it works, for some reason they are not running.

Ishan-Kumar2:
Hi @sdesrozis the hvd tests are working now ([here](https://github.com/pytorch/ignite/runs/3346667860#step:9:827)).
The xla tests are still failing because in ``_test_distrib_integration_engine_early_stopping``  there is an Accuracy metric which is passed to device. So should I change the device to cpu when it is xla for this test?

sdesrozis:
@Ishan-Kumar2 it seems there are still some code formatting errors https://github.com/pytorch/ignite/pull/2165/checks?check_run_id=3346963218#step:8:46

Ishan-Kumar2:
@sdesrozis I have fixed it, now both the hvd tests and xla tests are passing.
Let me know if they are correct. 
I noticed that in the tests for metrics the inputs are sliced using the rank of the device, so that all devices are testing on a different set of values like [here](https://github.com/pytorch/ignite/blob/2a4ed134b6099541e1e15b35bea2b2c800036e4e/tests/ignite/metrics/test_mean_pairwise_distance.py#L74). Is the same method also applicable here?

vfdev-5:
> I noticed that in the tests for metrics the inputs are sliced using the rank of the device, so that all devices are testing on a different set of values like here

@Ishan-Kumar2 yes, we are doing that for metrics to explicitly check that metrics are correctly computed across devices. In case of early stopping, all process should have the same data and technically should all require to stop if metrics are not improving... So, no need to make the data rank-dependent, I'd say.

Ishan-Kumar2:
> In case of early stopping, all process should have the same data and technically should all require to stop if metrics are not improving

Yup that makes sense. 

In terms of the checks, I am not sure why some are failing, they seem unrelated to this PR.

Ishan-Kumar2:
@vfdev-5 
On a slightly different issue, I noticed that in a lot of tests the way they are handled is that there is a helper function (`_test`) inside the test and a loop which calls it for differnt test cases for example [here](https://github.com/pytorch/ignite/blob/0e0200d7911776f5cd378f6ef29fde169836837c/tests/ignite/contrib/metrics/test_average_precision.py#L66-L109) 
Do you think it would be better to remove the loop and instead have the various values of tests as pytest parameters (`pytest.mark.parameterize`)?

For the above link `test_binary_and_multilabel_inputs` this would look like
```python

@pytest.mark.parametrize("y_pred, y, batch_size",[
            (torch.randint(0, 2, size=(50,)).long(), torch.randint(0, 2, size=(50,)).long(), 1),
            ...
        ])
def test_binary_and_multilabel_inputs(y_pred, y, batch_size):
    ...
    assert isinstance(res, float)
    assert average_precision_score(np_y, np_y_pred) == pytest.approx(res)

```

sdesrozis:
I think a good way is to mimic what is done in others metrics' tests 😉

sdesrozis:
> In terms of the checks, I am not sure why some are failing (`test_deterministic`), they seem unrelated to this PR.

@Ishan-Kumar2 It seems that we have some troubles with the PyTorch-Nigthly... We will investigate.

vfdev-5 (approved):
LGTM, thanks @Ishan-Kumar2

PR #2166, opened 2021-08-16
Quick fixes to about section

logankilpatrick:
Fixes working in about and history section.


sdesrozis:
@logankilpatrick Thank you very much !

PR #2173, opened 2021-08-22
Improve the Timer's example

### LINKED ISSUE #2171: Improve the example of Timer's usage
## 📚 Documentation

The example in the `Timer's documentation suggest to estimate the processing time for a batch as follows:
```python
timer.attach(
    trainer, 
    start=Events.EPOCH_STARTED, 
    resume=Events.ITERATION_STARTED, 
    pause=Events.ITERATION_COMPLETED, 
    step=Events.ITERATION_COMPLETED)
```
It is to note that this timer will be reset at the start of each epoch. When we use multiple workers in the data loader, the first couple of iterations at each epoch often take longer than the later ones. The reset behavior will incur an inaccurate estimation of the remaining training time (ETA), even when the `average` flag is set to `True`. Specifically, the ETA is computed as `(engine.state.max_iters - engine.state.iteration) * time_per_iter`. So the small fluctuation in `time_per_iter` will be magnified by remaining number of iterations. To address this problem, we can let the timer only start once in the whole training process:
```python
 timer.attach(
    trainer, 
    start=Events.EPOCH_STARTED(once=1), 
    resume=Events.ITERATION_STARTED, 
    pause=Events.ITERATION_COMPLETED, 
    step=Events.ITERATION_COMPLETED)
```
I have empirically verified the effectiveness of this modification.

sandylaker:
Signed-off-by: sandylaker <yawei.li@tum.de>

Fixes #2171

Description:

Check list:

### DIFF
--- a/ignite/handlers/timing.py
+++ b/ignite/handlers/timing.py
@@ -70,7 +70,7 @@ class Timer:
         >>> trainer = Engine(training_update_function)
         >>> timer = Timer(average=True)
         >>> timer.attach(trainer,
-        ...              start=Events.EPOCH_STARTED,
+        ...              start=Events.STARTED,
         ...              resume=Events.ITERATION_STARTED,
         ...              pause=Events.ITERATION_COMPLETED,
         ...              step=Events.ITERATION_COMPLETED)


vfdev-5 (approved):
Thanks for the PR @sandylaker

sandylaker:
@vfdev-5 It seems that some tests failed. But I only modified one line of code in the docs. Could you please also have a look? Thanks a lot.

vfdev-5:
@sandylaker no worries, nightlies are currently failing due to recent changes in random sampler. I merged the PR as it changes the docs only.

Issue #2178, opened 2021-08-24
Using Bleu for batch input


Ishan-Kumar2:
The example provided in docs for Bleu is for Single input. In that case the output from the engine should be in this format
```python

def evaluate_step():
  ...
  predictions = "Predicted Sentence 1".split()
  references = ["Reference Sentence 1".split(), "Reference Sentence 1.2".split()]
  return (predictions, references)

```

When calculating Bleu score for a Batch what is the format for output from the engine?
It should be something like
```python
#For a batch of size 2
predictions = ["Predicted Sentence 1".split(), "Predicted Sentence 2".split()]
references = [["Reference Sentence 1.1".split(), "Reference Sentence 1.2".split()], ["Reference Sentence 2.1".split(), "Reference Sentence 2.2".split()]]
```
Doing this gives an error. 
```python
TypeError: unhashable type: 'list'
```
The typing for `update` requires predictions to be a `Sequence[Any]` and References to be `Sequence[Sequence[Any]]]`. Does this mean the batch input is not possible?

vfdev-5:
Cc @sdesrozis @gucifer

sdesrozis:
@Ishan-Kumar2 Thanks for this question.

You are right, we didn't expose a batchwise api. Note that the internal function to compute the bleu score could be use in that sense. However, it will compute a bleu score per batch which would be different to a batch bleu score.

https://github.com/pytorch/ignite/blob/cc2c1aa9e31c297f8eb8fe14f3301dd201be8e48/ignite/metrics/nlp/bleu.py#L190

Would you like to help on this ? It would be a great improvement to fix the batch usage.

Ishan-Kumar2:
@sdesrozis thanks for the update, I am interested in contributing to this.
I agree we should implement the micro avg version also. Then the `corpus_bleu` can be renamed to `sentence_bleu` and the new `corpus_bleu` should be used for corpus/batch level calculations.

sdesrozis:
Firstly, the `_corpus_bleu` should be split in 2 parts : (1) the accumulation of ngrams counter (maybe the lenghts of cand/hyp too) and (2) the computation of bp, smoothing and geometric mean. The sentence bleu score (ie macro avg) is (1) + (2) for each sentence and (3) average. The corpus bleu score (ie micro avg) is (1) for each sentence then (2). 

[...]

PR #2180, opened 2021-08-27
added Bleu calculation on batch input

Ishan-Kumar2:
Addresses #2178 

Description:
Added Bleu calculation on batch of input.
Check list:

@sdesrozis


tests/ignite/metrics/nlp/test_bleu.py
```diff
@@ -121,23 +121,52 @@ def test_corpus_bleu_smooth2(candidate, references):
...
     value = bleu._corpus_bleu([corpus.references_1], [corpus.cand_1])
     value += bleu._corpus_bleu([corpus.references_2], [corpus.cand_2a])
     value += bleu._corpus_bleu([corpus.references_2], [corpus.cand_2b])
     value += bleu._corpus_bleu([corpus.references_2], [corpus.cand_3])
     assert bleu.compute() == value / 4
 
 
+def test_bleu_batch():
+
+    # Batch size 1
```
sdesrozis, line 143: This test is useless since we did all previous tests using batch size 1

tests/ignite/metrics/nlp/test_bleu.py
```diff
@@ -121,23 +121,52 @@ def test_corpus_bleu_smooth2(candidate, references):
...
+    refs = [corpus.references_2]
+    hypotheses = [corpus.cand_2a]
+    bleu = Bleu(ngram=4)
+    bleu.update((hypotheses, refs))
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        reference_bleu_score = (corpus_bleu([refs[0]], [hypotheses[0]]) + corpus_bleu([refs[0]], [hypotheses[0]])) / 2
+    assert bleu.compute() == reference_bleu_score
+
```
sdesrozis, line 153: It would be nice to compare the batch version, batch 1 version (loop over the batch) and the nltk bleu as you did.

tests/ignite/metrics/nlp/test_bleu.py
```diff
@@ -148,7 +177,7 @@ def _test_distrib_integration(device):
 
     data = []
     for c in corpus.chunks:
-        data += idist.get_world_size() * [c]
+        data += idist.get_world_size() * [([c[0]], [c[1]])]
```
sdesrozis, line 180: I would prefer modifying the corpus.chunks and keep this as the original. Adding batch in the corpus.chunks would be necessary to validate batch + ddp.

sdesrozis:
> @sdesrozis I have updated tests, please have a look.

@Ishan-Kumar2 I left a few comments. Thanks.

tests/ignite/metrics/nlp/test_bleu.py
Ishan-Kumar2: Added [here](https://github.com/pytorch/ignite/pull/2180/files#diff-4f64ad98d488d32957c8f58462fdb23ab0c3ea93f7b8150ff49c2a39665074b4R159-R170). This is what you were looking for right?

tests/ignite/metrics/nlp/test_bleu.py
```diff
@@ -121,23 +121,55 @@ def test_corpus_bleu_smooth2(candidate, references):
...
     assert bleu.compute() == value / 4
 
 
+def test_bleu_batch():
+    bleu = Bleu(ngram=4)
+
+    # Batch size 3
+    hypotheses = [corpus.cand_1, corpus.cand_2a, corpus.cand_2b]
+    refs = [corpus.references_1, corpus.references_2, corpus.references_2]
+    bleu.reset()
```
sdesrozis, line 147:
```suggestion
```

tests/ignite/metrics/nlp/test_bleu.py
```diff
@@ -121,23 +121,55 @@ def test_corpus_bleu_smooth2(candidate, references):
...
+        warnings.simplefilter("ignore")
+        reference_bleu_score = (
+            corpus_bleu([refs[0]], [hypotheses[0]])
+            + corpus_bleu([refs[1]], [hypotheses[1]])
+            + corpus_bleu([refs[2]], [hypotheses[2]])
+        ) / 3
+    assert bleu.compute() == reference_bleu_score
+
+    value = 0
+    for _hypotheses, _refs in zip(hypotheses, refs):
```
sdesrozis, line 160: I think it’s possible to do one loop instead of the two.

sdesrozis:
@Ishan-Kumar2 Thanks ! A minor remark left.

Ishan-Kumar2:
@sdesrozis incorporated the changes, thanks!

sdesrozis:
@Ishan-Kumar2 it looks good to me. However, some tests of the rouge metric are now broken. This is due to the `corpus.chunks` refactoring. Could you fix please ?

Ishan-Kumar2:
@sdesrozis I fixed this. On looking at the Rouge metric, I think we need to add batch functionality there also. I noticed the Rouge docstring is a little better as it explains what format the `y`, `y_pred` have to be in like [this](https://github.com/pytorch/ignite/blob/master/ignite/metrics/nlp/rouge.py#L194-L196). Should I add description like this to bleu also?

sdesrozis:
@Ishan-Kumar2 I agree that the rouge metrics could be improve to handle batch. 

Please, improve the docstring would be nice ! Thanks !

sdesrozis (approved):
@Ishan-Kumar2 Thank you ! LGTM !

ignite/metrics/nlp/bleu.py
```diff
@@ -187,8 +191,9 @@ def reset(self) -> None:
     @reinit__is_reduced
     def update(self, output: Tuple[Sequence[Any], Sequence[Sequence[Any]]]) -> None:
         y_pred, y = output
-        self._sum_of_bleu += self._corpus_bleu(references=[y], candidates=[y_pred])
-        self._num_sentences += 1
+        for _y_pred, _y in zip(y_pred, y):
+            self._sum_of_bleu += self._corpus_bleu(references=[_y], candidates=[_y_pred])
+            self._num_sentences += 1
```
vfdev-5, line 196: Looks like this breaks BC ?

sdesrozis: Yes it does. I was thinking since we don't have an official notebook, we could improve the usage to handle batches. What do you think ?

vfdev-5: I have an impression that type hints in `update` method are incorrect, especially checking the ones of `_corpus_bleu`

Ishan-Kumar2: If you want to keep compatibility with a warning, one way would be like this to add a check if single instances are passed like this https://github.com/pytorch/ignite/pull/2180#issuecomment-907169670

vfdev-5: OK, I understand now. Sorry, I haven't read everything here and now I agree with Sylvain and let's make the BC-breaking change. What about type hints, why they are the same as previous ones ?

Ishan-Kumar2: Okay, I'll update the type hints

vfdev-5:
Thanks @Ishan-Kumar2 , merging this PR

PR #2182, opened 2021-08-31
Fix ema test

### LINKED ISSUE #2181: CI XLA failing test: test_ema_final_weight_distrib_single_device_xla
tests/ignite/handlers/test_ema_handler.py::test_ema_final_weight_distrib_single_device_xla

- https://github.com/pytorch/ignite/pull/2180/checks?check_run_id=3471518836#step:9:147

```
_test_ema_final_weight(get_dummy_model(), device=device, ddp=True)

...

actual = tensor([[4.0625, 4.0625]], device='xla:0', dtype=torch.float64)
expected = tensor([[4.0625, 4.0625]], device='xla:0', dtype=torch.float64)

> Operator __and__ is only supported for integer or boolean type tensors, got: f64[1,2]{1,0}
```

cc @sandylaker

sandylaker:
Signed-off-by: sandylaker <yawei.li@tum.de>

Fixes #2181

Description: use `Tensor.new_full` to automatically cast the dtype of expected tensors to `torch.float32`.

Check list:

### DIFF
--- a/tests/ignite/handlers/test_ema_handler.py
+++ b/tests/ignite/handlers/test_ema_handler.py
@@ -205,8 +205,8 @@ def test_ema_two_handlers(get_dummy_model):
     engine.run(range(2), max_epochs=2)
     ema_weight_1 = ema_handler_1.ema_model.weight.data
     ema_weight_2 = ema_handler_2.ema_model.weight.data
-    torch.testing.assert_allclose(ema_weight_1, torch.full((1, 2), 4.0625))
-    torch.testing.assert_allclose(ema_weight_2, torch.full((1, 2), 3.5))
+    torch.testing.assert_allclose(ema_weight_1, ema_weight_1.new_full((1, 2), 4.0625))
+    torch.testing.assert_allclose(ema_weight_2, ema_weight_2.new_full((1, 2), 3.5))
 
     assert engine.state.ema_momentum_1 == 0.5
     assert engine.state.ema_momentum_2 == 0.5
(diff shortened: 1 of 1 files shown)


sandylaker:
@vfdev-5 After a little bit investigation, I found the problem was caused by the **nightly-version** of `torch_xla`. 

`torch.testing.all_close` will raise an error if the two tensors are of `torch.float32` or `torch.float64`. However this error will not occur when using the stable version of `torch_xla`.

You can try out this [Colab Notebook](https://colab.research.google.com/drive/1bfN3g-Z-XexiPZ5_PdgCKKs1YuckQenT?usp=sharing).

vfdev-5:
@sandylaker thanks a lot for investigating that ! If it is a reproducible bug with xla nightly, maybe we can open an issue on pytorch/xla ?
I tried the notebook and it fails with importing torch_xla and mkl:
```
ImportError: /usr/local/lib/python3.7/dist-packages/_XLAC.cpython-37m-x86_64-linux-gnu.so: undefined symbol: _ZN2at4_ops4item4callERKNS_6TensorE
```

Otherwise, I'd like to quickly check if replacing torch.testing.allclose with `assert expected.allclose(output)`

sandylaker:
@vfdev-5 This error is caused by the mismatch between` torch` and `torch_xla`.  Unfortunately, I cannot remember how I solved it yesterday. However, if you run the notebook with `torch_xla==1.9`. You will find that `torch.testing.assert_allclose` does not throw any error. So I think the root cause is the nightly version of `torch_xla`.

I will change the `assert_allclose` expressions as you suggest.

vfdev-5:
Thanks ! 

> By the way, how are the tensors compared in other modules' tests?

we were using assert t1.allclose(t2), `torch.testing` is very new and marked as prototype (https://pytorch.org/docs/1.9.0/testing.html)

vfdev-5 (approved):
Thanks @sandylaker !

PR #2190, opened 2021-09-10
Transformer Example Update - AutoModelforSequenceClassification

### LINKED ISSUE #2189: Restrict model options in transformers examples
In transformers example https://github.com/pytorch/ignite/tree/master/examples/contrib/transformers it's up to user to override the default model which is `bert-base-uncased`, a lot of models take similar inputs to `BERT` and similar outputs too, but models like `distilbert-base-uncased`, `distilroberta-base`, `bart-base` (and many other models) will not work here as they work on a bit different way regarding to the inputs and outputs of the model, check here for more info: https://huggingface.co/transformers/model_doc/distilbert.html#distilbertmodel

So we will get an error similar to this while using `DistilBERT`
```
KeyError: 'token_type_ids'
```
And also if we used something like `XLNet` that won't work well and we will have dimensions issue, because `XLNet` doesn't return a `pooler_output`, check here https://huggingface.co/transformers/model_doc/xlnet.html#transformers.XLNetModel

So what we should do is:
- We don't provide the option of choosing models, and we just keep BERT, and if the user wants to change the model, he can do that manually
- Keep it that way and mention in the docs that, the models that can be used here are `BERT`, `RoBERTa` (and if there is others we mentions them).

EDIT: After deciding to go with the first option, what we need to do now is to remove `model` from the argument, and set `config['model'] = 'bert-base-uncased'` manually inside `run` method, and mention in the docs we are using `BERT` by default, and maybe leave a note about if the user wants to try another model, he should that himself.

Ishan-Kumar2:
Fixes #2189 

Description:
Changes the model from `AutoModel` to `AutoModelforSequenceClassification` so that any model which supports Sequence Classification [list](https://huggingface.co/transformers/model_doc/auto.html#transformers.AutoModelForSequenceClassification) can be used without the need to make changes to the code.
Check list:

### DIFF
--- a/examples/contrib/transformers/README.md
+++ b/examples/contrib/transformers/README.md
@@ -34,6 +34,15 @@ python main.py run
 ```
 If needed, please, adjust the batch size to your GPU device with `--batch_size` argument.
 
+The default model is `bert-base-uncased` incase you need to change that use the `--model` argument, for details on which models can be used refer [here](https://huggingface.co/transformers/v3.0.2/model_doc/auto.html#automodelforsequenceclassification)
+
+Example: 
+```bash
+#Using DistilBERT which has 40% less parameters than bert-base-uncased
+python main.py run --model="distilbert-base-uncased"
+```
+
+
 For details on accepted arguments:
 
 ```bash
--- a/examples/contrib/transformers/dataset.py
+++ b/examples/contrib/transformers/dataset.py
@@ -11,24 +11,19 @@ class TransformerDataset(torch.utils.data.Dataset):
     def __getitem__(self, idx):
         text = str(self.texts[idx])
         text = " ".join(text.split())
-        inputs = self.tokenizer.encode_plus(
-            text, None, add_special_tokens=True, max_length=self.max_length, truncation=True
+        inputs = self.tokenizer(
+            text,
+            None,
+            add_special_tokens=True,
+            max_length=self.max_length,
+            truncation=True,
+            padding="max_length",
+            return_tensors="pt",
         )
+        inputs = {k: v.type(torch.long).squeeze(0) for k, v in inputs.items()}
 
-        ids = inputs["input_ids"]
-        token_type_ids = inputs["token_type_ids"]
-        mask = inputs["attention_mask"]
-        padding_length = self.max_length - len(ids)
-
-        ids = ids + ([0] * padding_length)
-        mask = mask + ([0] * padding_length)
-        token_type_ids = token_type_ids + ([0] * padding_length)
-        return {
-            "input_ids": torch.tensor(ids, dtype=torch.long),
-            "attention_mask": torch.tensor(mask, dtype=torch.long),
-            "token_type_ids": torch.tensor(token_type_ids, dtype=torch.long),
-            "label": torch.tensor(self.labels[idx], dtype=torch.float),
-        }
+        labels_pt = torch.tensor(self.labels[idx], dtype=torch.float)
+        return inputs, labels_pt
 
     def __len__(self):
         return len(self.labels)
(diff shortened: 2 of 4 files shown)


sdesrozis:
Tried on Colab and it seems this does not work fine... @Ishan-Kumar2 could you check please ? Thanks !!

KickItLikeShika:
Thanks for investigating @sdesrozis. Thanks for the PR @Ishan-Kumar2, please make sure your code works before pushing your updates, thanks!

Ishan-Kumar2:
Hi @sdesrozis, I tested on my machine and on colab just now, its seems to be working for me. Could you share the notebook you running or please have a look at this one(I stopped the run after 1 epoch)?
https://colab.research.google.com/drive/1xJ6tB8GVf_1lZfj3xV3rFwdbwi41R9qW#scrollTo=EIf3dk6PC979
Thanks

sdesrozis:
@Ishan-Kumar2 I confirm it does not work using GPU which is not enabled in your notebook. Anyway, it would be great that your branch works fine (version 0.5 instead of 0.4.6)

```
%%shell 

git clone https://github.com/Ishan-Kumar2/ignite.git
cd ignite
git checkout transformers_ex_refactor
python setup.py install
cd examples/contrib/transformers
pip install -r requirements.txt
python main.py run --batch_size=2 --max_length=10
```

Ishan-Kumar2:
@sdesrozis thanks for the review, I realized there was a slight mistake in the transfer of the input to GPU. I have fixed it now and tested on Colab, it works fine. Please take a look thanks!

KickItLikeShika:
Thanks for the updates @Ishan-Kumar2, the first issue has been solved, but there is another issue related to GPU memory in Colab with the default `max_length` and `batch_size`, after running th evaluation it crashes because GPU is out of memory, I suggest using  `gc.collect()` + `torch.cuda.empty_cache()` + being careful with what needs to be deleted/what needs to stay on CPU to avoid this issue, and also I suggest decrease the default batch_size a bit, 128 is too large for a transformer, please fix this issue. Thanks, Ishan!

Ishan-Kumar2:
@KickItLikeShika, you are right the eval does cause an out of memory error on default params. 
It is because the batch size is made twice that of train [here](https://github.com/pytorch/ignite/pull/2190/files#R228) (So it becomes 256). 
The training part takes about 8.5gb of GPU memory which is fine with most Colab GPUs, so I think just changing the val batch size equal to train batch size should be good enough? Or should I reduce the default train batch size to 64 and keep the double in val (128).

KickItLikeShika:
Hey @Ishan-Kumar2, any updates of this?

Ishan-Kumar2:
@KickItLikeShika, added the commit. Sorry for the delay, I was stuck on why there was a sudden increase in GPU memory usage when going from Epoch 0 to Epoch 1.

KickItLikeShika (approved):
Very nice work, Ishan! thanks a lot for the updates!

Ishan-Kumar2:
@KickItLikeShika I have made a minor change reverting the validation dataloader batch_size to 2 * train_loader batch_size. This wasn't the reason for the out of memory earlier.

Priyansi:
Hey @Ishan-Kumar2 and @KickItLikeShika , thank you for working on this PR. Just wanted to let you know that we are in the process of shifting ignite/examples to https://github.com/pytorch-ignite/examples. While this PR is a great improvement, this might not be relevant and can be removed once we make tutorials out of this, one of which is already in the process: https://github.com/pytorch-ignite/examples/pull/32 which uses `AutoModelForSequenceClassification`.

sdesrozis (approved):
@Ishan-Kumar2 Thanks ! LGTM !

PR #2200, opened 2021-09-16
Enabled LRFinder to run multiple epochs

### LINKED ISSUE #1357: FastaiLRFinder does not run more than 1 epoch. Why?
I have been trying to use the `FastaiLRFinder` to find the best learning rate for my module.

If we create the trainer with the function `create_supervised_trainer` as below:

```
trainer = create_supervised_trainer(
    model, optimizer, criterion, device, output_transform=custom_output_transform
)
```

and run it:

```
with lr_finder.attach(
            trainer,
            to_save=to_save,
            num_iter=50,  
            end_lr=1.,
            step_mode='exp') as lr_finder_training:
        lr_finder_training.run(train_loader)
```

A warning will come up say: "UserWarning: Desired num_iter 50 is unreachable with the current run setup of 15 iteration (1 epochs) from ignite.contrib.handlers.param_scheduler import (LRScheduler, PiecewiseLinear)"

My dataloader has 15 batches to iterate, which means FastaiLRFinder does not allow you to run more than 1 epoch. Why?

According to their source code, we can see [here](https://github.com/pytorch/ignite/blob/14cdc046c7cac2db8a1bf838dcd09cbbe0e9c33b/ignite/contrib/handlers/lr_finder.py#L87) that this verification limits the user to run more iterations than the maximum of epochs in their dataloader.

But why? Am I missing something important here?

KickItLikeShika:
Fixes #1357

Previously LRFinder raises a warning if `num_iter` was equivalent to more than 1 epoch, fixed that, and it runs more than 1 epoch now.

Check list:

### DIFF
--- a/ignite/handlers/lr_finder.py
+++ b/ignite/handlers/lr_finder.py
@@ -98,26 +98,15 @@ class FastaiLRFinder:
         self._best_loss = None
         self._diverge_flag = False
 
-        # print(trainer.state.epoch_length)
-        # print(trainer.state.max_epochs)
-        # print(num_iter // trainer.state.epoch_length)
         # attach LRScheduler to trainer.
         if num_iter is None:
             num_iter = trainer.state.epoch_length * trainer.state.max_epochs
         else:
             max_iter = trainer.state.epoch_length * trainer.state.max_epochs  # type: ignore[operator]
-            print(f"max iters: {max_iter}")
             if max_iter < num_iter:
                 max_iter = num_iter
                 trainer.state.max_iters = num_iter
                 trainer.state.max_epochs = num_iter // trainer.state.epoch_length
-            print(f"max iters: {max_iter}")
-            if num_iter > max_iter:
-                warnings.warn(
-                    f"Desired num_iter {num_iter} is unreachable with the current run setup of {max_iter} iteration "
-                    f"({trainer.state.max_epochs} epochs)",
-                    UserWarning,
-                )
 
         if not trainer.has_event_handler(self._reached_num_iterations):
             trainer.add_event_handler(Events.ITERATION_COMPLETED, self._reached_num_iterations, num_iter)
@@ -504,7 +493,6 @@ class _ExponentialLR(_LRScheduler):
     def __init__(self, optimizer: Optimizer, start_lr: float, end_lr: float, num_iter: int, last_epoch: int = -1):
         self.end_lr = end_lr
         self.num_iter = num_iter
-        self.shit = 0
         super(_ExponentialLR, self).__init__(optimizer, last_epoch)
 
         # override base_lrs
@@ -512,8 +500,5 @@ class _ExponentialLR(_LRScheduler):
 
     def get_lr(self) -> List[float]:  # type: ignore
         curr_iter = self.last_epoch + 1  # type: ignore[attr-defined]
-        # print(self.num_iter)
-        self.shit += 1
-        print(self.shit)
         r = curr_iter / self.num_iter
         return [base_lr * (self.end_lr / base_lr) ** r for base_lr in self.base_lrs]  # type: ignore[attr-defined]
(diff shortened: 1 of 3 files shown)


tests/ignite/handlers/test_lr_finder.py
```diff
@@ -321,6 +321,18 @@ def test_detach_terminates(lr_finder, to_save, dummy_engine, dataloader, recwarn
...
+def test_different_num_iters(lr_finder, to_save, dummy_engine, dataloader):
+    with lr_finder.attach(dummy_engine, to_save, num_iter=200) as trainer_with_finder:
+        trainer_with_finder.run(dataloader)
+    len_lrs_few_iters = len(lr_finder._history["lr"])
+
+    with lr_finder.attach(dummy_engine, to_save, num_iter=1000) as trainer_with_finder:
+        trainer_with_finder.run(dataloader)
+    len_lrs_many_iters = len(lr_finder._history["lr"])
+
+    assert len_lrs_many_iters - len_lrs_few_iters > 500
```
vfdev-5, line 333: I think the test should ensure that trainer_with_finder.state.iteration after the run is equal to num_iter

ignite/handlers/lr_finder.py
```diff
@@ -103,12 +103,10 @@ def _run(
...
-            if num_iter > max_iter:
-                warnings.warn(
-                    f"Desired num_iter {num_iter} is unreachable with the current run setup of {max_iter} iteration "
-                    f"({trainer.state.max_epochs} epochs)",
-                    UserWarning,
-                )
+            if max_iter < num_iter:
+                max_iter = num_iter
+                trainer.state.max_iters = num_iter
+                trainer.state.max_epochs = num_iter // trainer.state.epoch_length  # type: ignore[operator]
```
vfdev-5, line 109: I wonder if this should not be ceil instead of floor...

vfdev-5 (commented):
I left few comments to improve the PR

tests/ignite/handlers/test_lr_finder.py
KickItLikeShika: I tried that, the number of iterations always is not 100% equal to `num_iters`, it's always slightly smaller (that was always the case, even before these updates), how much smaller depends on the scheduler we are using, so If we would like to this, we should do range, for example
```py
num_iter = 100
num_iter - trainer_with_finder.state.iteration < 10
```

ignite/handlers/lr_finder.py
KickItLikeShika: I think it should stay floor, for example 150 num_iters should be 1 epoch I think, and have a look here here for example https://github.com/pytorch/ignite/blob/master/tests/ignite/handlers/test_lr_finder.py#L306

tests/ignite/handlers/test_lr_finder.py
vfdev-5: That's true ! Maybe we can set the divergence threshold rather high such that we wont exit before ? In this case they should be exact. Or specify end_lr not that high such that loss value wont go too high neither ...

ignite/handlers/lr_finder.py
vfdev-5: Thanks for the link. I think this test shows that the API is a bit weird:
- dataloader size is 100
- user requires 150 iters to do
```python
    with lr_finder.attach(dummy_engine, to_save, num_iter=150, diverge_th=float("inf")) as trainer_with_finder:
        with pytest.warns(UserWarning):
        # We should do 150 iters and not the size of the dataloader:
        assert dummy_engine.state.iteration != len(dataloader)
        assert dummy_engine.state.iteration == 150
```

KickItLikeShika: Thanks for the clarification, you're right, this test should be edited according to the updates, as the lrfinder now is able to run more than one epoch, will edit it

vfdev-5 (approved):
Thanks for the PR, Ahmed ! Looks good

Issue #2048, opened 2021-06-10
distributed program hangs in SLURM


Nic-Ma:
## 🐛 Bug description

Hi @vfdev-5 ,

We got an urgent issue from MONAI and Clara users that the distributed program hangs in NVIDIA NSL-B platform, which is based on SLURM.
You can reproduce the issue with this simple example:
https://github.com/Project-MONAI/tutorials/blob/master/acceleration/distributed_training/unet_training_workflows.py
It will hang when creating ignite Accurary metric, seems related to this line:
https://github.com/pytorch/ignite/blob/v0.4.4.post1/ignite/distributed/comp_models/native.py#L107
After removing the Accurary metric from the example, it hangs when training started and hasn't timeout yet.
Please note that this example can run successfully with **ignite 0.4.2**.
And we also tried the pure PyTorch dist example in the same hardware and software env, it can run successfully:
https://github.com/Project-MONAI/tutorials/blob/master/acceleration/distributed_training/unet_training_ddp.py

Could you please help analyze the reason and give some advice? It blocks our cooperation with another team now.

Thanks in advance.

## Environment

- PyTorch Version (e.g., 1.4): 1.8.1
- Ignite Version (e.g., 0.3.0): 0.4.4
- OS (e.g., Linux): Ubuntu 18.04
- How you installed Ignite (`conda`, `pip`, source): pip

sdesrozis:
Hi @Nic-Ma ! Sorry to hear that we have such an issue with SLURM.

In that script, you use `torch.distributed` method to create process group. We have the `ignite.distributed` (`idist`) used on our side, in order to handle SLURM, parallel contexts (slurm, horovod, tpu, etc.) with one code version. The interaction is generally transparent, yet I suspect that SLURM could require a little modification.

We have a way to manage the initialisation of a distributed app based on `idist.Parallel`. See [here](https://github.com/sdesrozis/why-ignite/blob/main/basics/3_ignite/helloworld.py). If this helper is not used, it works generally fine, however it could require a synchronisation for `idist`. 


We first need to try adding the following line after the process group initialisation
```
idist.sync()
```
and see if it helps.

Nic-Ma:
Hi @sdesrozis ,

Thanks for your detailed analysis.
So I think you mean we should use `ignite.distributed` to init the process group? Let me have a try.

Thanks.

sdesrozis:
It could also help us for understanding what is running wrong to have the launch command.

[...]

PR #2206, opened 2021-09-19
Refactored PTH DDP env vars creation in SLURM

### LINKED ISSUE #2202: SLURM and conflicting environment variables
## 🚀 Feature

Discussed here https://github.com/pytorch/ignite/issues/2048#issuecomment-921153418

Currently, if slurm environment variables are detected, an exception may be thrown if conflicting environment variables are defined (LOCAL_RANK, RANK, WORLD_SIZE). Typically, these variables are defined in `idist` using the slurm variables. This mutual exclusion is safe and avoid misuse of PyTorch launcher with slurm. 

https://github.com/pytorch/ignite/blob/922907217928fee7766046227506e54b28ae4283/ignite/distributed/comp_models/native.py#L219

However, some tools offered by nvidia (for docker/singularity, see https://github.com/NVIDIA/enroot/blob/v3.1.0/conf/hooks/extra/50-slurm-pytorch.sh) seem to do exactly the same things in the environment.

The idea is to verify the consistency of the environment (slurm wrt others), and raise an exception only if a conflict is detected.

https://github.com/pytorch/ignite/blob/922907217928fee7766046227506e54b28ae4283/ignite/distributed/comp_models/native.py#L250

vfdev-5:
Fixes #2202

Related to https://github.com/pytorch/ignite/issues/2048

Description:
- Refactored PTH DDP env vars creation in SLURM

Check list:

### DIFF
--- a/ignite/distributed/comp_models/native.py
+++ b/ignite/distributed/comp_models/native.py
@@ -3,7 +3,7 @@ import re
 import subprocess
 import warnings
 from distutils.version import LooseVersion
-from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple, cast
+from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple, Union, cast
 
 import torch
 import torch.distributed as dist
@@ -215,12 +215,13 @@ if has_native_dist_support:
             all_env_vars_defined = [k in os.environ for k in env_vars]
 
             if "SLURM_JOB_ID" in os.environ:
-                if any(all_env_vars_defined):
-                    raise RuntimeError(
-                        f"Defined env variables '{env_vars}' should not be specified with SLURM. Typically, this "
-                        "happens when `torch.distributed.launch` or `torch.multiprocessing.spawn` are used. Please be "
-                        "sure to use the `srun` command instead."
-                    )
+                # TO REMOVE:
+                # if any(all_env_vars_defined):
+                #     raise RuntimeError(
+                #         f"Defined env variables '{env_vars}' should not be specified with SLURM. Typically, this "
+                #         "happens when `torch.distributed.launch` or `torch.multiprocessing.spawn` are used. Please be "
+                #         "sure to use the `srun` command instead."
+                #     )
                 if rank is not None or world_size is not None:
                     raise ValueError("Arguments rank and world_size should not be specified with SLURM")
                 self._setup_env_in_slurm()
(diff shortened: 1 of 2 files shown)


ignite/distributed/comp_models/native.py
```diff
@@ -215,12 +215,6 @@ def setup_env_vars(self, rank: Optional[int] = None, world_size: Optional[int] =
             all_env_vars_defined = [k in os.environ for k in env_vars]
```
sdesrozis, line 215: L214-215 should be moved in the `else` statement L22

ignite/distributed/comp_models/native.py
```diff
@@ -499,3 +491,84 @@ def _expand_hostlist(nodelist: str) -> List[str]:
...
+    def _setup_ddp_vars_from_slurm_env(environ: Dict[str, str]) -> Dict[str, Union[str, int]]:
+        """Method to setup DDP env vars required by PyTorch from SLURM env
+        """
+        # 1) Tools like enroot can have hooks to translate slurm env vars to RANK, LOCAL_RANK, WORLD_SIZE etc
+        # See https://github.com/NVIDIA/enroot/blob/v3.1.0/conf/hooks/extra/50-slurm-pytorch.sh
+        # 2) User can use torch.distributed.launch tool to schedule on N local GPUs using 1 node, 1 task by SLURM
+        # To cover case 1), let's ensure that defined RANK == SLURM_PROCID, LOCAL_RANK == SLURM_LOCALID,
+        #   WORLD_SIZE == SLURM_NTASKS. We will use defined MASTER_ADDR and MASTER_PORT instead of defining
+        #   them by our means
+        # To cover case 2), let's check that defined RANK >= SLURM_PROCID, LOCAL_RANK >= SLURM_LOCALID,
```
sdesrozis, line 504: If I understand what is done, the idea is to ensure that in such a case, the user didn't use `srun` as a mistake 
```
srun python -m torch.distributed.launch ...
```
Therefore, every process should have a rank, local rank and world size greater or equal to what is defined by slurm.

Is it correct ?

ignite/distributed/comp_models/native.py
```diff
@@ -499,3 +491,84 @@ def _expand_hostlist(nodelist: str) -> List[str]:
...
+        #   WORLD_SIZE >= SLURM_NTASKS, SLURM_JOB_NUM_NODES == 1
+
+        ddp_vars: Dict[str, Union[str, int]] = {
+            "RANK": int(environ["SLURM_PROCID"]),
+            "LOCAL_RANK": int(environ["SLURM_LOCALID"]),
+            "WORLD_SIZE": int(environ["SLURM_NTASKS"]),
+            "MASTER_ADDR": "",
```
sdesrozis, line 511: It would be better to have `None` instead of these defaults, for instance L558 would become 
```
if ddp_vars["MASTER_ADDR"] is None:
    ...
```
but it's a detail.

vfdev-5: I did that in the very first commits but to make mypy happy I had to tweak things like that. Can check if can figure out another way to have None and our output type hint

ignite/distributed/comp_models/native.py
vfdev-5: The case 2 is to cover use-case like : 
```
srun -N1 -n1 -G8 python -m torch.distributed.launch\
        --nproc_per_node=8 --nnodes=1 --node_rank=0 \
        --master_addr="localhost" --master_port=1234 \
        main.py
```

ignite/distributed/comp_models/native.py
vfdev-5: L22 ?

ignite/distributed/comp_models/native.py
sdesrozis: case 2)

`RANK >= SLURM_PROCID`, `LOCAL_RANK >= SLURM_LOCALID` means that one process is spawn on the node by `srun` 
but 
`WORLD_SIZE >= SLURM_NTASKS` sounds weird. `SLURM_NTASKS` is the max number of tasks checked by slurm. If `WORLD_SIZE` is greater to that value, the scheduler should kill the process, because more than allocated ressources for the job are used.

What do you think ?

sdesrozis: Otherwise, it works on my side.

vfdev-5: The example above `srun -N1 -n1` allocates `SLURM_NTASKS=1`, but launcher creates ws=8. That's why `WORLD_SIZE >= SLURM_NTASKS`. Am I missing something ?

sdesrozis: Yes but the slurm scheduler can kill the job if more ressources than allocated are used. I suppose it depends on how the scheduler is configured. Imagine you schedule a job defining 4 tasks and in fact you use 8, it could be a big issue in production. Anyway, I think that does not really matter. This is a user constraint, we can't handle.

sdesrozis (approved):
LGTM !

PR #2217, opened 2021-09-25
Fix 2114: Sphinx copy-button plugin

### LINKED ISSUE #2114: Improve doc examples rendering
Currently we are rendering our examples in the docs like this https://pytorch.org/ignite/generated/ignite.engine.engine.Engine.html#ignite.engine.engine.Engine which is not very neat.
What we want to do is to find a better way to make it more good looking, and update our docs examples

image

- Replace "Examples" with some sphinx tag or something else, for example, like scikit-image does:

image

toxa23:
Fixes #2114

Description:
part 2 of 7: Added Sphinx plugin "copy-button" that allows copy code examples to clipboard

Check list:

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,4 @@
 sphinx==3.4.3
 -e git+https://github.com/pytorch/pytorch_sphinx_theme.git#egg=pytorch_sphinx_theme
 sphinxcontrib-katex
+sphinx-copybutton==0.4.0
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -57,6 +57,7 @@ extensions = [
     "sphinx.ext.napoleon",
     "sphinx.ext.viewcode",
     "sphinx.ext.autosectionlabel",
+    "sphinx_copybutton",
 ]
 
 # katex options
@@ -123,7 +124,7 @@ html_favicon = "_templates/_static/img/ignite_logomark.svg"
 html_static_path = ["_static", "_templates/_static"]
 
 html_context = {
-    "css_files": [
+    "extra_css_files": [
         # 'https://fonts.googleapis.com/css?family=Lato',
         # '_static/css/pytorch_theme.css'
         "_static/css/ignite_theme.css",


vfdev-5 (approved):
LGTM! Thanks @toxa23 

@ydcjeff could you please review this PR as well. Thanks

docs/requirements.txt
```diff
@@ -1,3 +1,4 @@
 sphinx==3.4.3
 -e git+https://github.com/pytorch/pytorch_sphinx_theme.git#egg=pytorch_sphinx_theme
 sphinxcontrib-katex
+sphinx-copybutton==0.4.0
```
ydcjeff, line 4: Since this is a small extension, do we need to pin the version number?

docs/source/conf.py
```diff
@@ -123,7 +124,7 @@
 html_static_path = ["_static", "_templates/_static"]
 
 html_context = {
-    "css_files": [
+    "extra_css_files": [
```
ydcjeff, line 127: Why is this change needed?

toxa23: Because `css_files` is a builtin context variable used by `app.add_css_file` in Sphinx' plugins setup. The copy button provides a css file. If we still want to overwrite `css_files` then we have to duplicate the copy button styles in `ignite_theme.css`.

docs/requirements.txt
toxa23: At least is is guaranteed to work with pinned Sphinx version.

ydcjeff (approved):
LGTM

docs/requirements.txt
oke-aditya: In my opinion you can set it to >=0.9.0, pytorch sphinx theme works with it fine. (Tested it for torchvision)

vfdev-5: @oke-aditya well, copy-button latest is 0.4.0. >=0.9.0 is gallery you are talking about : https://github.com/pytorch/vision/blob/b7c235892332c49a86c4be1b00fb6a672a3f9930/docs/requirements.txt#L4

oke-aditya: My bad. Sorry for that

PR #2219, opened 2021-09-25
Related to #2114: reformat docstrings in the 'engine' folder

### LINKED ISSUE #2114: Improve doc examples rendering

toxa23:
Related to #2114 

Description:
part 4 of 7, reformat docstrings in the 'engine' folder, make sure all examples have "Examples:" header, move "Args:" block to the top of method documentation

Check list:


vfdev-5 (approved):
Looks good, thanks @toxa23

PR #2223, opened 2021-09-25
Added gradient accumulation to supervised training step functions

### LINKED ISSUE #1662: add gradient normalization and accumulation in supervised_training_step_* functions
## 🚀 Feature

After #1589 , `grad_norm` is left to add (#419), and also [gradient accumulation](https://pytorch.org/ignite/master/faq.html#gradients-accumulation) is also common in supervised training, we should also add that to have full supervised training.

Ishan-Kumar2:
Fixes #1662

Description:
Added Gradient Accumulation feature in all the supervised training steps. 

@vfdev-5  I have used a variable `gradient_accumulation_steps` which is set to 1 by default instead of a flag and integer. If this is fine, I can start adding the tests for this.

Check list:

### DIFF
--- a/ignite/engine/__init__.py
+++ b/ignite/engine/__init__.py
@@ -91,8 +91,8 @@ def supervised_training_step(
 
     if gradient_accumulation_steps <= 0:
         raise ValueError(
-            "Gradient Accumulation steps can't be <= 0, if you don't want to use "
-            "gradient accumulation set it to 1 (default)"
+            "Gradient_accumulation_steps must be strictly positive. "
+            "No gradient accumulation if the value set to one (default)."
         )
 
     def update(engine: Engine, batch: Sequence[torch.Tensor]) -> Union[Any, Tuple[torch.Tensor]]:
@@ -168,8 +168,8 @@ def supervised_training_step_amp(
 
     if gradient_accumulation_steps <= 0:
         raise ValueError(
-            "Gradient Accumulation steps can't be <= 0, if you don't want to use "
-            "gradient accumulation set it to 1 (default)"
+            "Gradient_accumulation_steps must be strictly positive. "
+            "No gradient accumulation if the value set to one (default)."
         )
 
     def update(engine: Engine, batch: Sequence[torch.Tensor]) -> Union[Any, Tuple[torch.Tensor]]:
@@ -250,8 +250,8 @@ def supervised_training_step_apex(
 
     if gradient_accumulation_steps <= 0:
         raise ValueError(
-            "Gradient Accumulation steps can't be <= 0, if you don't want to use "
-            "gradient accumulation set it to 1 (default)"
+            "Gradient_accumulation_steps must be strictly positive. "
+            "No gradient accumulation if the value set to one (default)."
         )
 
     def update(engine: Engine, batch: Sequence[torch.Tensor]) -> Union[Any, Tuple[torch.Tensor]]:
@@ -324,8 +324,8 @@ def supervised_training_step_tpu(
 
     if gradient_accumulation_steps <= 0:
         raise ValueError(
-            "Gradient Accumulation steps can't be be <= 0, if you don't want to use "
-            "gradient accumulation set it to 1 (default)"
+            "Gradient_accumulation_steps must be strictly positive. "
+            "No gradient accumulation if the value set to one (default)."
         )
 
     def update(engine: Engine, batch: Sequence[torch.Tensor]) -> Union[Any, Tuple[torch.Tensor]]:
@@ -452,12 +452,6 @@ def create_supervised_trainer(
     device_type = device.type if isinstance(device, torch.device) else device
     on_tpu = "xla" in device_type if device_type is not None else False
     mode, _scaler = _check_arg(on_tpu, amp_mode, scaler)
-
-    if gradient_accumulation_steps <= 0:
-        raise ValueError(
-            "Gradient Accumulation steps can't be <= 0, if you don't want to use "
-            "gradient accumulation set it to 1 (default)"
-        )
 
     if mode == "amp":
         _update = supervised_training_step_amp(
--- a/ignite/engine/__init__.py
+++ b/ignite/engine/__init__.py
@@ -85,7 +85,7 @@ def supervised_training_step(
         trainer = Engine(update_fn)
 
     .. versionadded:: 0.4.5
-    .. versionadded:: 0.5.0
+    .. versionchanged:: 0.5.0
         Added Gradient Accumulation.
     """
 
@@ -93,7 +93,9 @@ def supervised_training_step(
         model.train()
         x, y = prepare_batch(batch, device=device, non_blocking=non_blocking)
         y_pred = model(x)
-        loss = loss_fn(y_pred, y) / gradient_accumulation_steps
+        loss = loss_fn(y_pred, y)
+        if gradient_accumulation_steps > 1:
+            loss = loss / gradient_accumulation_steps
         loss.backward()
         if engine.state.iteration % gradient_accumulation_steps == 0:
             optimizer.step()
@@ -149,7 +151,7 @@ def supervised_training_step_amp(
         trainer = Engine(update_fn)
 
     .. versionadded:: 0.4.5
-    .. versionadded:: 0.5.0
+    .. versionchanged:: 0.5.0
         Added Gradient Accumulation.
     """
 
@@ -163,7 +165,9 @@ def supervised_training_step_amp(
         x, y = prepare_batch(batch, device=device, non_blocking=non_blocking)
         with autocast(enabled=True):
             y_pred = model(x)
-            loss = loss_fn(y_pred, y) / gradient_accumulation_steps
+            loss = loss_fn(y_pred, y)
+            if gradient_accumulation_steps > 1:
+                loss = loss / gradient_accumulation_steps
         if scaler:
             scaler.scale(loss).backward()
             if engine.state.iteration % gradient_accumulation_steps == 0:
@@ -223,7 +227,7 @@ def supervised_training_step_apex(
         trainer = Engine(update_fn)
 
     .. versionadded:: 0.4.5
-    .. versionadded:: 0.5.0
+    .. versionchanged:: 0.5.0
         Added Gradient Accumulation.
     """
 
@@ -236,7 +240,9 @@ def supervised_training_step_apex(
         model.train()
         x, y = prepare_batch(batch, device=device, non_blocking=non_blocking)
         y_pred = model(x)
-        loss = loss_fn(y_pred, y) / gradient_accumulation_steps
+        loss = loss_fn(y_pred, y)
+        if gradient_accumulation_steps > 1:
+            loss = loss / gradient_accumulation_steps
         with apex_amp.scale_loss(loss, optimizer) as scaled_loss:
             scaled_loss.backward()
         if engine.state.iteration % gradient_accumulation_steps == 0:
@@ -290,7 +296,7 @@ def supervised_training_step_tpu(
         trainer = Engine(update_fn)
 
     .. versionadded:: 0.4.5
-    .. versionadded:: 0.5.0
+    .. versionchanged:: 0.5.0
        Added Gradient Accumulation argument for all supervised training methods.
     """
     try:
@@ -302,7 +308,9 @@ def supervised_training_step_tpu(
         model.train()
         x, y = prepare_batch(batch, device=device, non_blocking=non_blocking)
         y_pred = model(x)
-        loss = loss_fn(y_pred, y) / gradient_accumulation_steps
+        loss = loss_fn(y_pred, y)
+        if gradient_accumulation_steps > 1:
+            loss = loss / gradient_accumulation_steps
         loss.backward()
         if engine.state.iteration % gradient_accumulation_steps == 0:
             xm.optimizer_step(optimizer, barrier=True)
@@ -413,7 +421,7 @@ def create_supervised_trainer(
         - Added ``amp_mode`` argument for automatic mixed precision.
         - Added ``scaler`` argument for gradient scaling.
 
-    .. versionadded:: 0.5.0
+    .. versionchanged:: 0.5.0
         Added Gradient Accumulation argument for all supervised training methods.
     """
 
--- a/ignite/engine/__init__.py
+++ b/ignite/engine/__init__.py
@@ -51,6 +51,7 @@ def supervised_training_step(
     non_blocking: bool = False,
     prepare_batch: Callable = _prepare_batch,
     output_transform: Callable = lambda x, y, y_pred, loss: loss.item(),
+    gradient_accumulation_steps: int = 1,
 ) -> Callable:
     """Factory function for supervised training.
 
@@ -67,7 +68,8 @@ def supervised_training_step(
             tuple of tensors `(batch_x, batch_y)`.
         output_transform: function that receives 'x', 'y', 'y_pred', 'loss' and returns value
             to be assigned to engine's state.output after each iteration. Default is returning `loss.item()`.
-
+        gradient_accumulation_steps: Number of steps the gradients should be accumulated across.
+            (default: 1 (means no gradient accumulation))
     Returns:
         Callable: update function.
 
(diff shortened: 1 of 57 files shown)


sdesrozis:
@Ishan-Kumar2 Thanks, it looks good ! IMO you can now work on the tests :)

Ishan-Kumar2:
@sdesrozis added the tests, please take a look.

ignite/engine/__init__.py
```diff
@@ -83,16 +85,19 @@ def supervised_training_step(
         trainer = Engine(update_fn)
 
     .. versionadded:: 0.4.5
+    .. versionadded:: 0.5.0
```
vfdev-5, line 88:
```suggestion
    .. versionchanged:: 0.5.0
```

ignite/engine/__init__.py
```diff
@@ -141,6 +149,8 @@ def supervised_training_step_amp(
...
```
vfdev-5, line 152:
```suggestion
    .. versionchanged:: 0.5.0
```

ignite/engine/__init__.py
```diff
@@ -83,16 +85,19 @@ def supervised_training_step(
...
+        Added Gradient Accumulation.
     """
 
     def update(engine: Engine, batch: Sequence[torch.Tensor]) -> Union[Any, Tuple[torch.Tensor]]:
         model.train()
-        optimizer.zero_grad()
         x, y = prepare_batch(batch, device=device, non_blocking=non_blocking)
         y_pred = model(x)
-        loss = loss_fn(y_pred, y)
+        loss = loss_fn(y_pred, y) / gradient_accumulation_steps
```
vfdev-5, line 96:
```suggestion
        loss = loss_fn(y_pred, y) / gradient_accumulation_steps if gradient_accumulation_steps > 1 else loss_fn(y_pred, y)
```
to avoid additional useless kernel execution

ignite/engine/__init__.py
```diff
@@ -207,6 +223,8 @@ def supervised_training_step_apex(
         trainer = Engine(update_fn)
 
     .. versionadded:: 0.4.5
+    .. versionadded:: 0.5.0
```
vfdev-5, line 226:
```suggestion
    .. versionchanged:: 0.5.0
```

vfdev-5 (commented):
Thanks for the PR @Ishan-Kumar2 
I left few comments to address

ignite/engine/__init__.py
Ishan-Kumar2: Okay I have added the check as a seperate line, flake8 was making it a little less readable if we kept it like this in the same line.

Ishan-Kumar2: Also I forced the commit by mistake which caused your commit to be overwritten, I have merged the master again sorry.

sdesrozis: Not sure that it adds overhead. The AD should detect a division by a scalar so no impact on the derivatives and only one extra division on the value.

vfdev-5: It's about performances (not AD) and skip +1 CUDA kernel execution

ignite/engine/__init__.py
```diff
@@ -150,18 +162,23 @@ def supervised_training_step_amp(
...
         with autocast(enabled=True):
             y_pred = model(x)
             loss = loss_fn(y_pred, y)
+            if gradient_accumulation_steps > 1:
+                loss = loss / gradient_accumulation_steps
         if scaler:
             scaler.scale(loss).backward()
-            scaler.step(optimizer)
-            scaler.update()
+            if engine.state.iteration % gradient_accumulation_steps == 0:
```
sdesrozis, line 173: Bug. If `gradient_accumulation_steps < 0`, `optimizer` never updated.

ignite/engine/__init__.py
```diff
@@ -150,18 +162,23 @@ def supervised_training_step_amp(
```
sdesrozis, line 162: Add an exception catching if `gradient_accumulation_steps < 0`

ignite/engine/__init__.py
sdesrozis, line 173: Bug if `gradient_accumulation_steps < 0`

ignite/engine/__init__.py
```diff
@@ -384,27 +420,65 @@ def create_supervised_trainer(
...
+
+    .. versionchanged:: 0.5.0
+        Added Gradient Accumulation argument for all supervised training methods.
     """
 
     device_type = device.type if isinstance(device, torch.device) else device
     on_tpu = "xla" in device_type if device_type is not None else False
     mode, _scaler = _check_arg(on_tpu, amp_mode, scaler)
 
+    if gradient_accumulation_steps == 0:
```
sdesrozis, line 432: This exception should catch if `gradient_accumulation_steps <= 0`.
IMO it should be added in every functions since they should be used standalone.

tests/ignite/engine/test_create_supervised.py
```diff
@@ -172,6 +172,73 @@ def _test_create_mocked_supervised_trainer(
                             assert training_step_mock.called
 
 
+def _test_gradient_accumulation(
```
sdesrozis, line 175: Is it possible to update the test `_test_create_supervised_trainer` instead of introduce this new one which seems to be a copy/paste with a little modification ?

Ishan-Kumar2: Okay in that case should I parameterize over gradient_accumulation_steps like this?

```python
test_create_supervised_parameters=("gradient_accumulation_steps,loss,weight,bias",[(1,...),(3,...)])

def _test_create_supervised_trainer(gradient_accumulation_steps,loss,weight,bias,....):
... 

@pytest.mark.parametrize(*test_create_supervised_parameters)
def test_create_supervised_trainer_with_cpu(gradient_accumulation_steps, loss, weight, bias):
    _test_create_supervised_trainer(gradient_accumulation_steps, loss, weight, bias, trainer_device="cpu")
```

Ishan-Kumar2: Or just call it twice with ``gradient_accumulation_steps``  as 1 and not 1

sdesrozis: > Okay in that case should I parameterize over gradient_accumulation_steps like this?
> 
> ```python
> test_create_supervised_parameters=("gradient_accumulation_steps,loss,weight,bias",[(0,...),(1,...),(3,...)])
> 
> def _test_create_supervised_trainer(gradient_accumulation_steps,loss,weight,bias,....):
> ... 
> 
> @pytest.mark.parametrize(*test_create_supervised_parameters)
> def test_create_supervised_trainer_with_cpu(gradient_accumulation_steps, loss, weight, bias):
>     _test_create_supervised_trainer(gradient_accumulation_steps, loss, weight, bias, trainer_device="cpu")
> ```

Yes, `gradient_accumulation_steps`, `loss`, `weight`, and `bias` arguments should be added.

Ishan-Kumar2:
Hi @sdesrozis I have included the gradient_accumulation tests in the `_test_create_supervised_trainer` only. 

A slight problem  is that all the calling functions like `test_create_supervised_trainer` also call `_test_create_mocked_supervised_trainer` which just needs to be called once but due to pytest parameterisation over the parent it also gets called 3 times.

ignite/engine/__init__.py
```diff
@@ -51,6 +51,7 @@ def supervised_training_step(
     non_blocking: bool = False,
     prepare_batch: Callable = _prepare_batch,
     output_transform: Callable = lambda x, y, y_pred, loss: loss.item(),
+    gradient_accumulation_steps: int = 1,
```
sdesrozis, line 54: L103 defines the usage. IMO we can define this parameter as optional int > 1, default to `None`. What do you think ?

ignite/engine/__init__.py
```diff
@@ -83,16 +85,27 @@ def supervised_training_step(
         trainer = Engine(update_fn)
 
     .. versionadded:: 0.4.5
+    .. versionchanged:: 0.5.0
+        Added Gradient Accumulation.
     """
 
+    if gradient_accumulation_steps <= 0:
+        raise ValueError(
+            "Gradient Accumulation steps can't be <= 0, if you don't want to use "
```
sdesrozis, line 94: If we use an optional parameter, the message could be cleaner.

ignite/engine/__init__.py
```diff
@@ -384,27 +444,65 @@ def create_supervised_trainer(
...
+
+    .. versionchanged:: 0.5.0
+        Added Gradient Accumulation argument for all supervised training methods.
     """
 
     device_type = device.type if isinstance(device, torch.device) else device
     on_tpu = "xla" in device_type if device_type is not None else False
     mode, _scaler = _check_arg(on_tpu, amp_mode, scaler)
 
+    if gradient_accumulation_steps <= 0:
```
sdesrozis, line 456: Since every method does the test, it seems useless here.

sdesrozis:
> Hi @sdesrozis I have included the gradient_accumulation tests in the `_test_create_supervised_trainer` only.
> 
> A slight problem is that all the calling functions like `test_create_supervised_trainer` also call `_test_create_mocked_supervised_trainer` which just needs to be called once but due to pytest parameterisation over the parent it also gets called 3 times.

@Ishan-Kumar2 Thank you. However, too much code duplicated. Let me think about it.

Ishan-Kumar2:
> @Ishan-Kumar2 Thank you. However, too much code duplicated. Let me think about it.

How about I write a small wrapper test which does the check for zero and then calls with given args like this.
```python
def _test_create_supervised_wrapper(gradient_accumulation_steps, loss, weight, bias, **kwargs):
    if gradient_accumulation_steps == 0:
        with pytest.raises(ValueError, match="Gradient Accumulation steps can't be <= 0"):
            _test_create_supervised_trainer(gradient_accumulation_steps, loss, weight, bias, **kwargs)
    else:
        _test_create_supervised_trainer(gradient_accumulation_steps, loss, weight, bias, **kwargs)
```
and then just call this from all the tests? 
like this
```python

@pytest.mark.parametrize(*test_create_supervised_parameters)
def test_create_supervised_trainer_on_cuda(gradient_accumulation_steps, loss, weight, bias):
    _test_create_supervised_wrapper(gradient_accumulation_steps,loss,weight,bias,model_device=model_device,trainer_device=trainer_device)
    _test_create_mocked_supervised_trainer(model_device=model_device, trainer_device=trainer_device)
```

It would reduce the repetition of check for 0.

sdesrozis:
I think we could refactor the tests to have a better code base. We should firstly define a trainer used in the tests as follow
```python
def _default_create_supervised_trainer(
    gradient_accumulation_steps = 1,
    model_device: Optional[str] = None,
    trainer_device: Optional[str] = None,
    trace: bool = False,
    amp_mode: str = None,
    scaler: Union[bool, "torch.cuda.amp.GradScaler"] = False,
):    
    model = Linear(1, 1)

    if model_device:
        model.to(model_device)

    model.weight.data.zero_()
    model.bias.data.zero_()
    optimizer = SGD(model.parameters(), 0.1)

    if trace:
        example_input = torch.randn(1)
        model = torch.jit.trace(model, example_input)

    if amp_mode == "apex" and model_device == trainer_device == "cuda":
        from apex import amp

        model, optimizer = amp.initialize(model, optimizer, opt_level="O2")

    trainer = create_supervised_trainer(
        model,
        optimizer,
        mse_loss,
        device=trainer_device,
        output_transform=lambda x, y, y_pred, loss: (y_pred, loss.item()),
        amp_mode=amp_mode,
        scaler=scaler,
        gradient_accumulation_steps=gradient_accumulation_steps,
    )
    
    assert model.weight.data[0, 0].item() == approx(0.0)
    assert model.bias.item() == approx(0.0)

    return trainer, model
```
Thus we can limit the coupling between model / trainer definitions and the validation.

Secondly, we could do the same for the evaluator.

I think, we must handle the case `gradient_accumulation_steps == 0` outside the parametrize.
```python
def _test_create_supervised_trainer_wrong_accumulation(model_device, trainer_device):
    with pytest.raises(ValueError, match="Gradient Accumulation steps can't be <= 0"):
        _default_create_supervised_trainer(
            gradient_accumulation_steps=0,
            model_device=model_device,
            trainer_device=trainer_device
        )
```
and for instance 
```python
@pytest.mark.tpu
@pytest.mark.skipif("NUM_TPU_WORKERS" in os.environ, reason="Skip if no NUM_TPU_WORKERS in env vars")
@pytest.mark.skipif(not idist.has_xla_support, reason="Skip if no PyTorch XLA package")
@pytest.mark.parametrize(*test_create_supervised_parameters)
def test_create_supervised_trainer_on_tpu(gradient_accumulation_steps, loss, weight, bias):
    model_device = trainer_device = "xla"
    _test_create_supervised_trainer_wrong_accumulation(model_device=model_device, trainer_device=trainer_device)
    _test_create_supervised_trainer(gradient_accumulation_steps, loss, weight, bias,
                                    model_device=model_device,
                                    trainer_device=trainer_device)
    _test_create_mocked_supervised_trainer(model_device=model_device, trainer_device=trainer_device)
```

@Ishan-Kumar2 What do you think ?

Ishan-Kumar2:
@sdesrozis yes this looks good, I'll incorporate these changes.

tests/ignite/engine/test_create_supervised.py
```diff
@@ -89,23 +118,10 @@ def _test_create_supervised_trainer(
...
-        trainer = create_supervised_trainer(
-            model,
-            optimizer,
-            mse_loss,
-            device="cpu",
-            output_transform=lambda x, y, y_pred, loss: (y_pred, loss.item()),
-            amp_mode=None,
-            scaler=True,
-        )
+        trainer, model = _default_create_supervised_trainer(model_device="cpu", trainer_device="cpu", scaler=True)
```
sdesrozis, line 124:
```suggestion
        trainer, _ = _default_create_supervised_trainer(model_device="cpu", trainer_device="cpu", scaler=True)
```

tests/ignite/engine/test_create_supervised.py
```diff
@@ -121,30 +137,11 @@ def _test_create_mocked_supervised_trainer(
...
-
-                        model, optimizer = amp.initialize(model, optimizer, opt_level="O2")
-
-                    trainer = create_supervised_trainer(
-                        model,
-                        optimizer,
-                        mse_loss,
-                        device=trainer_device,
-                        output_transform=lambda x, y, y_pred, loss: (y_pred, loss.item()),
+                    trainer, model = _default_create_supervised_trainer(
```
sdesrozis, line 141:
```suggestion
                    trainer, _ = _default_create_supervised_trainer(
```

tests/ignite/engine/test_create_supervised.py
```diff
@@ -224,19 +240,9 @@ def _test_mocked_supervised_evaluator(
...
-            model.bias.data.zero_()
-
-            if trace:
-                example_input = torch.randn(1, 1)
-                model = torch.jit.trace(model, example_input)
-
-            evaluator = create_supervised_evaluator(model, device=evaluator_device, amp_mode=amp_mode)
+            model, evaluator = _default_create_supervised_evaluator(
+                model_device=model_device, evaluator_device=evaluator_device, trace=trace, amp_mode=amp_mode
+            )
```
sdesrozis, line 245: L252  
```
evaluator.run(data)
```

sdesrozis (approved):
@Ishan-Kumar2 Thank you ! It looks good even if `neptune`'s tests are failing.

tests/ignite/engine/test_create_supervised.py
Ishan-Kumar2: Incorporated these changes, thanks!

Ishan-Kumar2:
> @Ishan-Kumar2 I was thinking about the point you noticed. I would slightly change the problem to solve this and remove the `pytest.parametrize`. The idea is to consider a very simple and predictable problem.
> 
> ```
> model = Linear(1, 1, bias=False)
> ```
> 
> and
> 
> ```
> x = torch.tensor([[1.], [1.], [1.], [1.], [1.]])
> y = x
> 
> data = [(_x, _x) for _x in x]
> ```
> 
> thus (no accumulation) the variation of the weight is
> 
> ```
> theta := theta + 2 * lr * (1 - theta)
> ```
> 
> No longer `pytest.parametrize` needed. What do you think ?

Hi @sdesrozis, this is a nice solution, I agree it works. I have added it with the bias (I didn't notice you proposed using without it). Let me know if it is fine.

sdesrozis (approved):
@Ishan-Kumar2 Thank your very much ! LGTM ! Let's merge this PR asap.

Ishan-Kumar2:
@sdesrozis sure, let me know if anything needs to be updated. The test failures seem unrelated.

vfdev-5:
Thanks @Ishan-Kumar2 , merging the PR

PR #2225, opened 2021-09-27
Relates to #2114. bugfix for docstring examples with huge bottom padding

toxa23:
Relates to #2114

Description:
docstrings examples CSS styles should be applied to the `<pre>` tag rather than .highlight, otherwise another `<pre>` style is added and a big bottom padding becomes highlighted too.

Check list:


vfdev-5 (approved):
Thanks a lot for the fix @toxa23 !

PR #2228, opened 2021-09-28
Fix test_auto CI errors

sdesrozis:
Description:

Fix CI errors due to exception variation in `test_auto`

Check list:


vfdev-5 (approved):
Let's nail it down once for all. Thanks @sdesrozis

Issue #2224, opened 2021-09-25
Update sphinx to 3.5.4 or latest


vfdev-5:
## 📚 Documentation

Let's bump Sphinx version to 3.5.4 or later for our docs building.

Anubha13kumari:
hi @vfdev-5 .first timer this side .i want to work on this issue please guide me .

vfdev-5:
Hi @Anubha13kumari , please check out https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md#writing-documentation and try to build locally our docs. Next, change sphinx version in `docs/requirements.txt`, reinstall the new version and rebuild the docs. Please report here how it works and send a PR with the change.

ydcjeff:
As a side note, updating to latest (4.2.0) is [failing to build the old versioned docs](https://github.com/ydcjeff/ignite/runs/3731615939?check_suite_focus=true).

We should think about how to manage old versioned docs.

vfdev-5:
Looks like it is an issue with https://github.com/pytorch-ignite/sphinxcontrib-versioning which we can update as well from our side

ydcjeff:
> Looks like it is an issue with https://github.com/pytorch-ignite/sphinxcontrib-versioning which we can update as well from our side

@vfdev-5 I think at some point, we might have used deprecated API of latest sphinx to work with old versioned docs since old docs used old sphinx features.

vfdev-5:
@trsvchn can you check how much effort it can take to update https://github.com/pytorch-ignite/sphinxcontrib-versioning  ? thanks

PR #2233, opened 2021-09-29
changed sphinx version

Anubha13kumari:
Fixes #{issue number}

Description:

Check list:

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
-sphinx==3.4.3
+sphinx==4.2.0
 -e git+https://github.com/pytorch/pytorch_sphinx_theme.git#egg=pytorch_sphinx_theme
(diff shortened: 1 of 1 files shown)


vfdev-5:
@Anubha13kumari thanks for the PR. I think we have to pause on that a bit as it will break doc building on master according to https://github.com/pytorch/ignite/issues/2224#issuecomment-929090224

trsvchn:
@Anubha13kumari thanks for the PR! 

In order to get it work, we have to replace this line

https://github.com/pytorch/ignite/blob/5e5e6ee4a8de7e280669f8324959b11ee4f08a89/docs/source/conf.py#L136

with these:

```suggestion
html_permalinks = True
html_permalinks_icon = "#"
```

since `html_add_permalinks` is deprecated

PR #2236, opened 2021-10-02
Related to #2207: Friendly Errors (#2207)

### LINKED ISSUE #2207: Raise friendly errors in case of `num_epochs` is less than `num_warmup_epochs` in CIFAR10 and Transformers examples
I was trying to benchmark `torch.inference_mode` vs `torch.no_grad` on CIFAR10 example, so I just wanted to run for 3 epochs and average the benchmarks, the default warmup epochs is 4, so the scheduler will raise this error, which is not very clear to new users 
```
f"Milestones should be increasing integers, but given {pair[0]} is smaller "
ValueError: Milestones should be increasing integers, but given 291 is smaller than the previous milestone 388
```
So I suggest to check `num_epochs` vs `num_warmup_epochs` inside `run` method and raise more clear error message for the users.

theory-in-progress:
Fixes #2207

Description:  
- Edited to raise friendly errors if ``num_epochs`` is less than ``num_warmup_epochs``
- Edited main.py in directories ``cifar10``, ``cifar10_qat`` and ``transformers``.

### DIFF
--- a/examples/contrib/cifar10/main.py
+++ b/examples/contrib/cifar10/main.py
@@ -187,6 +187,11 @@ def run(
         **spawn_kwargs: Other kwargs to spawn run in child processes: master_addr, master_port, node_rank, nnodes
 
     """
+    # check to see if the num_epochs is greater than num_warmup_epochs issue user friendly Error
+    if num_warmup_epochs>num_epochs:
+        raise ValueError(f"""The minimum number of warmup epochs is {num_warmup_epochs}, either increase the num_epochs 
+        or decrease the num_warmup_epochs""")
+
     # catch all local parameters
     config = locals()
     config.update(config["spawn_kwargs"])
--- a/examples/contrib/cifar10_qat/main.py
+++ b/examples/contrib/cifar10_qat/main.py
@@ -174,6 +174,11 @@ def run(
         **spawn_kwargs: Other kwargs to spawn run in child processes: master_addr, master_port, node_rank, nnodes
 
     """
+    # check to see if the num_epochs is greater than num_warmup_epochs issue user friendly Error
+    if num_warmup_epochs>num_epochs:
+        raise ValueError(f"""The minimum number of warmup epochs is {num_warmup_epochs}, either increase the num_epochs 
+        or decrease the num_warmup_epochs""")
+
     # catch all local parameters
     config = locals()
     config.update(config["spawn_kwargs"])
--- a/examples/contrib/transformers/main.py
+++ b/examples/contrib/transformers/main.py
@@ -192,6 +192,11 @@ def run(
         with_amp (bool): if True, enables native automatic mixed precision. Default, False.
         **spawn_kwargs: Other kwargs to spawn run in child processes: master_addr, master_port, node_rank, nnodes
     """
+    # check to see if the num_epochs is greater than num_warmup_epochs issue user friendly Error
+    if num_warmup_epochs>num_epochs:
+        raise ValueError(f"""The minimum number of warmup epochs is {num_warmup_epochs}, either increase the num_epochs 
+        or decrease the num_warmup_epochs""")
+
     # catch all local parameters
     config = locals()
     config.update(config["spawn_kwargs"])


examples/contrib/cifar10/main.py
```diff
@@ -187,6 +187,11 @@ def run(
         **spawn_kwargs: Other kwargs to spawn run in child processes: master_addr, master_port, node_rank, nnodes
 
     """
+    # check to see if the num_epochs is greater than num_warmup_epochs issue user friendly Error
+    if num_warmup_epochs>num_epochs:
+        raise ValueError(f"""The minimum number of warmup epochs is {num_warmup_epochs}, either increase the num_epochs 
+        or decrease the num_warmup_epochs""")
```
vfdev-5, line 193: @theory-in-progress Please fix code formatting issues and use single quote `"` instead of 3.
To reformat the code please check our contributing guide.

theory-in-progress:
Hi @vfdev-5, I've made the necessary code formatting to compile according to the guidelines. Should I close this pull request and create a new pull request?

examples/contrib/cifar10/main.py
```diff
@@ -187,6 +187,13 @@ def run(
         **spawn_kwargs: Other kwargs to spawn run in child processes: master_addr, master_port, node_rank, nnodes
 
     """
+    # check to see if the num_epochs is greater than num_warmup_epochs issue user friendly Error
```
KickItLikeShika, line 190:
```suggestion
    # check to see if the num_epochs is greater than num_warmup_epochs
```

KickItLikeShika: Apply this change everywhere else please

examples/contrib/cifar10/main.py
```diff
@@ -187,6 +187,13 @@ def run(
...
+    if num_warmup_epochs > num_epochs:
+        raise ValueError(
+            f"The minimum number of warmup epochs is {num_warmup_epochs}, either increase the num_epochs or decrease"
```
KickItLikeShika, line 193: The error message is not understandable for me, it should be the minimum number of (epochs) must be greater than or equal to num_warmup_epochs, am I missing something?
To be more clear I suggest to do something like
```
epochs can't be less than num_warmup_epochs, etc....
```

KickItLikeShika: I won't spam and put a lot of comments, this also the case of other updates. Thanks!

KickItLikeShika (commented):
Thanks a lot for your work @theory-in-progress! Just a couple of comments.

examples/contrib/cifar10/main.py
theory-in-progress: Right! Sorry for that, will make the changes soon enough and commit. Thanks for the inputs!

examples/contrib/cifar10/main.py
```diff
@@ -187,6 +187,10 @@ def run(
         **spawn_kwargs: Other kwargs to spawn run in child processes: master_addr, master_port, node_rank, nnodes
 
     """
+    # check to see if the num_epochs is greater than num_warmup_epochs
+    if num_warmup_epochs > num_epochs:
+        raise ValueError("num_epochs cannot be less than num_warmup_epochs")
```
KickItLikeShika, line 192:
```suggestion
        raise ValueError("num_epochs cannot be less than num_warmup_epochs, please increase num_epochs or decrease num_warmup_epochs")
```

KickItLikeShika (approved):
LGTM! Thanks @theory-in-progress!

vfdev-5:
I think code formatting wont pass

theory-in-progress:
Please run the tests now @vfdev-5, The code formatting ``bash ./tests/run_code_style.sh fmt`` ran once successfully but when I tried running it the second, third and fourth time it gave the error ``INTERNAL ERROR: Black produced code that is not equivalent to the source. `` please advise on how to proceed further. Thanks!

vfdev-5:
> INTERNAL ERROR: Black produced code that is not equivalent to the source

@theory-in-progress, it should be an issue with black. I've never seen it previously. On the other hand the new code looks correctly formatted.

theory-in-progress:
Oof, Not sure on how to proceed now, Sorry for the trouble I'm causing.

vfdev-5:
Broken CI on nightly is unrelated and reveals an issue with pytorch nightly. I think we can merge your PR now.

examples/contrib/cifar10/main.py
```diff
@@ -187,6 +187,12 @@ def run(
         **spawn_kwargs: Other kwargs to spawn run in child processes: master_addr, master_port, node_rank, nnodes
 
     """
+    # check to see if the num_epochs is greater than num_warmup_epochs
+    if num_warmup_epochs > num_epochs:
```
vfdev-5, line 191: @theory-in-progress @KickItLikeShika What happen if num_warmup_epochs == num_epochs ?

theory-in-progress: Yeah, I think that's the right configuration, the original error according to @KickItLikeShika was raised when he set the ``num_epochs`` less than ``num_warmup_epochs``

vfdev-5: Well, even if it does not raise an error, does it make much sense to execute only warm-up phase ?

KickItLikeShika: Thanks for noting, @vfdev-5! Yup no worries about that, the scheduler can handle that.

vfdev-5: @KickItLikeShika can you detail on how exactly it will handle that ?

KickItLikeShika: Sure, check here https://github.com/pytorch/ignite/blob/master/ignite/handlers/param_scheduler.py#L928, we (only) check if the current miles is less than the previous one or not

vfdev-5: @KickItLikeShika I'm asking precisely about : 

> even if it does not raise an error, does it make much sense to execute only warm-up phase ?
For example, num_epochs=5 and num_warmup_epochs=5 --> training will be done on lr starting from 0.0 up to 0.4 and that's it.
If we expect num_epochs >= num_warmup_epochs + 1 --> training will be done on lr starting from 0.0 up to 0.4 and down to 0.0

KickItLikeShika: Sorry, the page wasn't reloaded since yesterday and didn't see your comments, IMO this error should be for the error only, here we handled raising the error incase of num_epochs is less, otherwise, the error message won't make much sense. On the other hand if we would like to inform the user about the equality of num_epochs and num_warmup_epochs we should raise a warning in this case, what do you think?

vfdev-5: I think it is OK to raise an error for a unrealistic use-case as well and not a warning

KickItLikeShika: Cool, @theory-in-progress please address this comment, and make sure to edit the error message as well, thanks!

theory-in-progress: Sure!

theory-in-progress: Sure I'll try that.

theory-in-progress: Nope, I tried creating new venv three times, each time I get the same error for pre-commit and then  ``3 files failed to reformat`` when I run the ``run_code_style.sh fmt``

theory-in-progress:
Well, I think the Code formatting is giving the error, sigh, I tried running the code formatting with the correct versions of isort and black, but It's just not happening...

examples/contrib/cifar10/main.py
```diff
@@ -187,6 +187,12 @@ def run(
         **spawn_kwargs: Other kwargs to spawn run in child processes: master_addr, master_port, node_rank, nnodes
 
     """
+    # check to see if the num_epochs is greater than or equal to num_warmup_epochs
+    if num_warmup_epochs >= num_epochs:
```
sdesrozis, line 191: If `black` doesn't work as expected, or whatever, please try a correction like 
```python
    if num_warmup_epochs >= num_epochs:
        raise ValueError(
            "num_epochs cannot be less than or equal to num_warmup_epochs, please increase num_epochs "
            "or decrease num_warmup_epochs"
        )
```
and let's see.

sdesrozis:
@vfdev-5 It seems that `black` doesnt format the comments, see [here](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#comments). Do I understand well ?

Issue #2230, opened 2021-09-28
Use doctest to execute examples in the docstring


sdesrozis:
## 📚 Documentation

Example snippets are inserted into the docstring to show usage. However, these pieces of code might be deprecated, or might not work. The module `doctest` aims to execute the tests written in docstring and it works fine with sphinx. It could help to detect earlier doc issues and provide a better doc. 

Useful links
* https://docs.python.org/3/library/doctest.html 
* https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html

Note that some tests was written according to `doctest`
https://github.com/pytorch/ignite/blob/b3d9849c2243be6068a64b5d1cf9c0ede727a2ea/ignite/handlers/timing.py#L30

The main point is the refactoring of every test to pass the validation. It means have consistent pieces of code (define training fn, engines, etc.) to provide reproducible examples.

vfdev-5:
Interesting idea. I think Jeff or Taras proposed already something like that previously. Probably we wont be able to execute all doc snippets like ddp etc

sdesrozis:
> Interesting idea. I think Jeff or Taras proposed already something like that previously. Probably we wont be able to execute all doc snippets like ddp etc

Of course, but it could help to keep most examples correct.

DevPranjal:
Hello @sdesrozis, @vfdev-5, I'd like to work on the Ignite docs, can you provide some more details about this issue? Do we have to check each doctest for deprecation? Or do we have to refactor the doctests to a follow a particular template?

sdesrozis:
A good first point would be to check whether `doctest` would help. Maybe we could choose a module and see how it helps. Meanwhile we must assess the required effort for the rest of the doc.

I propose to focus on a few metrics. Select ones having a piece of example code inside the docstring and check how `doctest` + sphinx extension works. What do you think ?

Anyway, I think this issue is a huge effort. We have to do it carefully according to others actions on the doc. Concretely the first point you would do should help to discuss and plan with the team.

DevPranjal:
Since this is a fast developing library, I think it would be great to have doctests to stay updated on the docs. I will start looking into how doctests work with sphinx.

vfdev-5:
@sdesrozis can we close this issue as solved or still there is something to be done ?

sdesrozis:
It's done !

PR #2241, opened 2021-10-04
feat(doctest): SSIM metric doctest

ydcjeff:
ref #2230 

Description:
- Add doctest for SSIM metric
- Add doctest gh job

Check list:


ignite/metrics/ssim.py
```diff
@@ -36,14 +36,17 @@ class SSIM(Metric):
...
-        .. code-block:: python
-
-            def process_function(engine, batch):
-                # ...
-                return y_pred, y
-            engine = Engine(process_function)
-            metric = SSIM(data_range=1.0)
-            metric.attach(engine, "ssim")
+        >>> def process_function(engine, batch):
+        ...     y_pred, y = batch
```
vfdev-5, line 40: Can't it work with code-block ? I'm not a big fan of `>>>`

ydcjeff: It does not work AFAIK.

vfdev-5: With `>>>` when code is copied it is with these useless `>>>`. I think it would be better to use other tags instead of `>>>` which requires each line modification.

ydcjeff: sphinx-copybutton can be modified to strip `>>>` in `conf.py`.

sdesrozis: I agree the @vfdev-5's remark about `>>>`. Maybe we could explore an alternative and see ?

ydcjeff: @vfdev-5 @sdesrozis change to using `.. testcode::` and `.. testoutput::`. If we are using those, we need to use `print` to test the expected output.

sdesrozis: Does the rendering of the doc show the output ?

sdesrozis: >  If we are using those, we need to use `print` to test the expected output.
Not sure to understand

ydcjeff: I mean if we are using `.. testcode::` and `.. testoutput::`, we need to use `print` like `print(state.metrics['ssim'])` to test the expected output.

> Does the rendering of the doc show the output ?

Yes, it shows.

ydcjeff:
@sdesrozis https://deploy-preview-2241--pytorch-ignite-preview.netlify.app/generated/ignite.metrics.ssim

sdesrozis:
Each examples above are tested by `make doctest`. I think a section title would be valuable before the output (`Expected output(s)` ?).

What do you think ?

ydcjeff:
Mostly python doctest style `>>>` is used alongside with output. So it's better to show output as well maybe with "Output" section if we want.

vfdev-5:
@sdesrozis @ydcjeff I let you open new tracking issue or update existing one to see clearly the task and how we can involve contributors to that. Sylvain, I let you guide this work and merge this PR or wait for something. Thanks!

PR #2246, opened 2021-10-05
minor black error fix

Ishan-Kumar2:
Fixes :
Description:

Fixes error in black formatting which is causing all the tests to fail.

Check list:


vfdev-5 (approved):
Thanks @Ishan-Kumar2

Issue #2213, opened 2021-09-24
[FR] Add kwargs to `idist.barrier`


vfdev-5:
## 🚀 Feature

A feature request to add kwargs to `idist.barrier`.

- Verify if all supported backends now can accept additional args to their `barrier` methods
- Add kwargs and route it to backend implementations
  - warn if kwargs are specified but can not be used by the backend  
- Update documentation
- Add test

PR #2254, opened 2021-10-09
Fixed NCCL warning caused by barrier if using idist

### LINKED ISSUE #2212: Fix NCCL warning caused by barrier if using `idist`
## 🚀 Feature

The following code will produce warnings with the latest PyTorch : 
```python
# python -u -m torch.distributed.run --nproc_per_node=2 main.py
import time

def training(local_rank, config, **kwargs):
    # ...
    time.sleep(0.1 * local_rank)
    print(idist.get_rank(), ": run with config:", config, "- backend=", idist.backend())
    # ...

    
if __name__ == "__main__":

    backend = "nccl"
    config = {}
    with idist.Parallel(backend=backend) as parallel:
        parallel.run(training, config, a=1, b=2)
```
Output:
```
[W ProcessGroupNCCL.cpp:1569] Rank 0 using best-guess GPU 0 to perform barrier as devices used by this process are currently unknown. This can potentially cause a hang if this rank to GPU mapping is incorrect.Specify device_ids in barrier() to force use of a particular device.
[W ProcessGroupNCCL.cpp:1569] Rank 1 using best-guess GPU 1 to perform barrier as devices used by this process are currently unknown. This can potentially cause a hang if this rank to GPU mapping is incorrect.Specify device_ids in barrier() to force use of a particular device.
2021-09-24 03:23:13,045 ignite.distributed.launcher.Parallel INFO: Initialized processing group with backend: 'nccl'
```

These NCCL warnings are caused by this code:
https://github.com/pytorch/ignite/blob/053411d73047e278f22008f9f0f1775d0e51dcd5/ignite/distributed/comp_models/native.py#L128-L129

Solutions:
- verify the API of `dist.barrier` and specify participating processes
- remove barrier and argue why

vfdev-5:
Fixes #2212

Description:
- Fixed NCCL warning caused by barrier if using idist

Check list:

### DIFF
--- a/ignite/distributed/comp_models/native.py
+++ b/ignite/distributed/comp_models/native.py
@@ -125,11 +125,19 @@ if has_native_dist_support:
 
             dist.init_process_group(backend, init_method=init_method, **init_pg_kwargs)
 
-            # https://github.com/facebookresearch/maskrcnn-benchmark/issues/172
-            dist.barrier()
-
             if torch.cuda.is_available():
                 torch.cuda.set_device(self._local_rank)
+
+            # Call barrier after init_process_group as in
+            # https://github.com/facebookresearch/maskrcnn-benchmark/issues/172
+            # Define device ids for NCCL to avoid warnings
+            # [W ProcessGroupNCCL.cpp:1569] Rank 0 using best-guess GPU 0 to perform barrier as devices used by
+            # this process are currently unknown. This can potentially cause a hang if this rank to GPU mapping
+            # is incorrect.Specify device_ids in barrier() to force use of a particular device.
+            device_ids = None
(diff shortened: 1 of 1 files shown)


sdesrozis:
IMO This could be the default implementation of `barrier` for the native model regarding to #2213.

vfdev-5:
I'm not very sure about this warning. I checked detectron2 and inspired from its code. Prior to fixing it, I checked for the warning with pytorch nightly and surprisingly there were no warning... As this is pure NCCL thing, i'd avoid generalizing that.
I also think that the following code is partially responsible for the warning:
```
if torch.cuda.is_available():
    torch.cuda.set_device(self._local_rank)
```
which was called after `barrier`...

Issue #2265, opened 2021-10-14
[docs] Add doctests for metrics


ydcjeff:
### 📚 Documentation

Following #2230 , I listed the metrics places (suggested by @sdesrozis) where doctests can be added.

#### Metrics

- [x] Average (#2332)
- [x] GeometricAverage (#2332)
- [x] VariableAccumulation (This class is an helper, no test needed)
- [x] Accuracy (#2345)
- [x] ConfusionMatrix (#2336)
- [x] ClassificationReport (#2346)
- [x] DiceCoefficient (#2336)
- [x] JaccardIndex (#2336)
- [x] IoU (#2336)
- [x] mIoU (#2336)
- [x] EpochMetric (#2351)
- [x] Fbeta (#2340)
- [x] Loss (#2335)
- [x] MeanAbsoluteError (#2280)
- [x] MeanPairwiseDistance (#2307)
- [x] MeanSquaredError (#2280)
- [x] MetricsLambda (#2352)
- [x] MultiLabelConfusionMatrix (#2347)
- [x] Precision (#2340)
- [x] PSNR (#2311)
- [x] Recall (#2340)
- [x] RootMeanSquaredError (#2307)
- [x] RunningAverage (#2348)
- [x] SSIM (#2241)
- [x] TopKCategoricalAccuracy (#2284)
- [x] Bleu (#2317)
- [x] Rouge (#2317)
- [x] RougeL (#2317)
- [x] RougeN (#2317)
- [x] InceptionScore (#2349)
- [x] FID (#2349)

#### Contrib Metrics

- [x] AveragePrecision (#2341)
- [x] CohenKappa (#2321)
- [x] PrecisionRecallCurve (#2341)
- [x] ROC_AUC (#2341)
- [x] RocCurve (#2341)
- [x] CanberraMetric (#2323)
- [x] FractionalAbsoluteError (#2323)
- [x] FractionalBias (#2323)
- [x] GeometricMeanAbsoluteError (#2323)
- [x] GeometricMeanRelativeAbsoluteError (#2324)
- [x] ManhattanDistance (#2324)
- [x] MaximumAbsoluteError (#2324)
- [x] MeanAbsoluteRelativeError (#2324)
- [x] MeanError (#2324)
- [x] MeanNormalizedBias (#2324)
- [x] MedianAbsoluteError (#2324)
- [x] MedianAbsolutePercentageError (#2324) 
- [x] MedianRelativeAbsoluteError (#2324)
[...]

PR #2287, opened 2021-10-22
[skip ci] Doctest for Accuracy

Chandan-h-509:
Addresses #2265

Description: Added doctests for 'Accuracy'

Check list:

- [X] New tests are added (if a new feature is added)
- [X] New doc strings: description and/or example code are in RST format
- [X] Documentation is updated (if required)

PR #2307, opened 2021-10-31
Doctests for `RMSE` and `MeanPairwiseDistance`

DevPranjal:
Addresses #2265 

Description: Added doctests for `RMSE` and `MeanPairwiseDistance`

Check list:

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -341,6 +341,14 @@ from ignite.metrics import *
 from ignite.utils import *
 
 manual_seed(666)
+
+# create default evaluator for doctests
+
+def process_function(engine, batch):
+    y_pred, y = batch
+    return y_pred, y
+
+default_evaluator = Engine(process_function)
 """
 
 
--- a/ignite/metrics/mean_pairwise_distance.py
+++ b/ignite/metrics/mean_pairwise_distance.py
@@ -26,6 +26,32 @@ class MeanPairwiseDistance(Metric):
         device: specifies which device updates are accumulated on. Setting the
             metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
             non-blocking. By default, CPU.
+
+    Examples:
+        To use with ``Engine`` and ``process_function``, simply attach the metric instance to the engine.
+        The output of the engine's ``process_function`` needs to be in the format of
+        ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y, ...}``. If not, ``output_tranform`` can be added
+        to the metric to transform the output into the form expected by the metric.
+
+        ``y_pred`` and ``y`` should have the same shape.
+
+        .. testcode::
+
+            metric = MeanPairwiseDistance(p=4)
+            metric.attach(default_evaluator, 'l4_distance')
+            preds = torch.Tensor([
+                [1, 2, 4, 1],
+                [2, 3, 1, 5],
+                [1, 3, 5, 1],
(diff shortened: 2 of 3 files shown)


docs/source/conf.py
```diff
@@ -341,6 +341,14 @@ def run(self):
...
 
 manual_seed(666)
+
+# create default evaluator for doctests
+
+def process_function(engine, batch):
+    y_pred, y = batch
+    return y_pred, y
+
+default_evaluator = Engine(process_function)
```
DevPranjal, line 351: I have added a global for the default evaluator to which @vfdev-5 was referring to in [this comment](https://github.com/pytorch/ignite/pull/2287#discussion_r736646796).
I also felt we could mention it somewhere... What could be a right place to mention it, if necessary? 
Thanks :)

sdesrozis: @DevPranjal Thank you. I think we already discussed that point, and the idea is to have simple and copy/paste pieces of code, even if a lot of code could be reused.

@vfdev-5 what do you think ?

ignite/metrics/mean_pairwise_distance.py
```diff
@@ -26,6 +26,32 @@ class MeanPairwiseDistance(Metric):
...
+        The output of the engine's ``process_function`` needs to be in the format of
+        ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y, ...}``. If not, ``output_tranform`` can be added
+        to the metric to transform the output into the form expected by the metric.
+
+        ``y_pred`` and ``y`` should have the same shape.
+
+        .. testcode::
+
+            metric = MeanPairwiseDistance(p=4)
+            metric.attach(default_evaluator, 'l4_distance')
```
sdesrozis, line 41:
```suggestion
            metric.attach(default_evaluator, 'mpd')
```
or anything else but

docs/source/conf.py
vfdev-5: @sdesrozis I do not see the point to make API examples cluttered with other "unrealistic" code just to make them executable.

vfdev-5 (approved):
LGTM, thanks @DevPranjal !

docs/source/conf.py
sdesrozis: The point is the user can't copy/paste the code to test. I thought we would like to provide users with easily reproducible examples.

Anyway, let's choose whether we hide things and let's deploy 😊

vfdev-5: I know, but it means for this metric we'll have the following:
```python
import torch
from torch import nn, optim
from ignite.engine import *
from ignite.handlers import *
from ignite.metrics import *
from ignite.utils import *

manual_seed(666)

    y_pred, y = batch
    return y_pred, y

default_evaluator = Engine(process_function)

# --- API example 
metric = MeanPairwiseDistance(p=4)
metric.attach(default_evaluator, 'mpd')
preds = torch.Tensor([
                [1, 2, 4, 1],
                [2, 3, 1, 5],
                [1, 3, 5, 1],
                [1, 5, 1 ,11]
])
target = preds * 0.75
state = default_evaluator.run([[preds, target]])
print(state.metrics['mpd'])
```

You have half of the code snippet of preps.

sdesrozis: We are talking about more or less 30 code lines, and using lambda function, we could reduce this. Not sure that it's worth hiding rather than exposing.

Anyway, that's a matter of taste.

PR #2326, opened 2021-11-14
Paramscheduler emahandler

### LINKED ISSUE #2295: Make work EMAHandler with state parameter scheduler
## 🚀 Feature

I'd like to make custom EMA decay scheduling with, for example, `PiecewiseLinearStateScheduler`. Here is failing code:
```python
import ignite
from ignite.handlers import PiecewiseLinearStateScheduler, EMAHandler
import torch
import torch.nn as nn
import matplotlib.pyplot as plt
%matplotlib inline

print(ignite.__version__)

model = nn.Linear(2, 1)
trainer = Engine(lambda e, b: model(b))
data = torch.rand(100, 2)

param_name = "ema_decay"

ema_handler.attach(trainer, name=param_name, event=Events.ITERATION_COMPLETED)

ema_decay_scheduler = PiecewiseLinearStateScheduler(
    param_name=param_name,
    milestones_values=[(0, 0.0), (10 * len(data), 0.999)]
)
ema_decay_scheduler.attach(trainer, Events.ITERATION_COMPLETED)

ema_decay_history = []
engine.add_event_handler(Events.EPOCH_COMPLETED, lambda _ : ema_decay_history.append(engine.state.ema_decay))

engine.run(data, max_epochs=20)
```
it gives: 
```bash
ValueError: Attribute: 'ema_decay' is already defined in the Engine.state.This may be a conflict between multiple StateParameterScheduler handlers.Please choose another name.
```

By the way, error message is missing spaces between sentences.

cc @sandylaker

fco-dv:
Fixes #2295

Description:
Add `create_new` parameter to ``attach`` method . This flag is used to create ``param_name`` on ``engine.state`` and is taking into account whether ``param_name`` attribute already exists or not on ``engine.state``. Overrides existing attribute by default.

Check list:

### DIFF
--- a/ignite/handlers/state_param_scheduler.py
+++ b/ignite/handlers/state_param_scheduler.py
@@ -50,8 +50,6 @@ class StateParamScheduler(BaseParamScheduler):
                     f"This may be a conflict between multiple handlers. "
                     f"Please choose another name."
                 )
-            else:
-                setattr(engine.state, self.param_name, None)
         else:
             if not self.create_new:
                 warnings.warn(
--- a/ignite/handlers/state_param_scheduler.py
+++ b/ignite/handlers/state_param_scheduler.py
@@ -46,7 +46,7 @@ class StateParamScheduler(BaseParamScheduler):
         if hasattr(engine.state, self.param_name):
             if self.create_new:
                 raise ValueError(
-                    f"Attribute: `{self.param_name}` already exists in the engine.state. "
+                    f"Attribute '{self.param_name}' already exists in the engine.state. "
                     f"This may be a conflict between multiple handlers. "
                     f"Please choose another name."
                 )
--- a/ignite/handlers/state_param_scheduler.py
+++ b/ignite/handlers/state_param_scheduler.py
@@ -15,6 +15,9 @@ class StateParamScheduler(BaseParamScheduler):
         param_name: name of parameter to update.
         save_history: whether to log the parameter values to
             `engine.state.param_history`, (default=False).
+        create_new: in case `param_name` already exists in `engine.state`, whether to authorize `StateParamScheduler`
+            create a new parameter based on `StateParamScheduler` class name.
+            By default, this option is `False` meaning that `StateParamScheduler` will override `param_name` values.
 
     Note:
         Parameter scheduler works independently of the internal state of the attached engine.
--- a/ignite/handlers/state_param_scheduler.py
+++ b/ignite/handlers/state_param_scheduler.py
@@ -13,11 +13,9 @@ class StateParamScheduler(BaseParamScheduler):
 
     Args:
         param_name: name of parameter to update.
-        save_history: whether to log the parameter values to
-            `engine.state.param_history`, (default=False).
-        create_new: in case `param_name` already exists in `engine.state`, whether to authorize `StateParamScheduler`
-            create a new parameter based on `StateParamScheduler` class name.
-            By default, this option is `False` meaning that `StateParamScheduler` will override `param_name` values.
+        save_history: whether to log the parameter values to ``engine.state.param_history``, (default=False).
+        create_new: whether to create ``param_name`` on ``engine.state`` taking into account whether ``param_name``
+            attribute already exists or not. Overrides existing attribute by default, (default=False).
 
     Note:
         Parameter scheduler works independently of the internal state of the attached engine.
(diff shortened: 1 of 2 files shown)


ignite/handlers/state_param_scheduler.py
```diff
@@ -13,6 +15,9 @@ class StateParamScheduler(BaseParamScheduler):
         param_name: name of parameter to update.
         save_history: whether to log the parameter values to
             `engine.state.param_history`, (default=False).
+        create_new: in case `param_name` already exists in `engine.state`, whether to authorize `StateParamScheduler`
```
vfdev-5, line 18: @fco-dv please use double-ticks ```` instead of single-tick when addresses code. Here and above.

```suggestion
        create_new: in case ``param_name`` already exists in ``engine.state``, whether to authorize ``StateParamScheduler``
```

ignite/handlers/state_param_scheduler.py
```diff
@@ -43,16 +47,35 @@ def attach(
 
         """
         if hasattr(engine.state, self.param_name):
-            raise ValueError(
-                f"Attribute: '{self.param_name}' is already defined in the Engine.state."
-                f"This may be a conflict between multiple StateParameterScheduler handlers."
-                f"Please choose another name."
-            )
+            if not self.create_new:
```
vfdev-5, line 50: I do not quite understand this if/else clauses, especially "else" one.
I was thinking about the following logic:

1) `param_name` NOT in `engine.state`  AND create_new is True => simply create new attribute
2) `param_name` NOT in `engine.state`  AND create_new is False => warn that we will create new attribute, but to remove this warning, create_new should be True
3) `param_name` in `engine.state`  AND create_new is True => raise error as we can not create new attribute as it already exists in the state.
4) `param_name` in `engine.state`  AND create_new is False => silently override existing attribute

What do you think ?

vfdev-5 (commented):
Thanks for the PR @fco-dv ! I left few comments

ignite/handlers/state_param_scheduler.py
fco-dv: Oh yes , I've misunderstood the case 3, thought we were handling the creation  of state parameter in that case ... So yes Indeed, it's better and simpler ! Thanks

ignite/handlers/state_param_scheduler.py
```diff
@@ -43,16 +44,24 @@ def attach(
...
         """
         if hasattr(engine.state, self.param_name):
-            raise ValueError(
-                f"Attribute: '{self.param_name}' is already defined in the Engine.state."
-                f"This may be a conflict between multiple StateParameterScheduler handlers."
-                f"Please choose another name."
-            )
+            if self.create_new:
+                raise ValueError(
+                    f"Attribute: `{self.param_name}` already exists in the engine.state. "
```
vfdev-5, line 49:
```suggestion
                    f"Attribute: '{self.param_name}' already exists in the engine.state. "
```

ignite/handlers/state_param_scheduler.py
```diff
@@ -43,16 +44,24 @@ def attach(
...
+                    f"This may be a conflict between multiple handlers. "
+                    f"Please choose another name."
+                )
+        else:
+            if not self.create_new:
+                warnings.warn(
+                    f"Attribute: `{self.param_name}` is not defined in the engine.state. "
+                    f"{type(self).__name__} will create it. Remove this warning by setting `create_new=True`."
```
vfdev-5, line 57:
```suggestion
                    f"Attribute '{self.param_name}' is not defined in the engine.state. "
                    f"{type(self).__name__} will create it. Remove this warning by setting create_new=True."
```

tests/ignite/handlers/test_state_param_scheduler.py
```diff
@@ -414,3 +389,58 @@ def __call__(self, event_index):
...
+    create_new = True
+
+    ema_handler = EMAHandler(model)
+    ema_handler.attach(trainer, name=param_name, event=Events.ITERATION_COMPLETED)
+    ema_decay_scheduler = PiecewiseLinearStateScheduler(
+        param_name=param_name,
+        milestones_values=[(0, 0.0), (10 * len(data), 0.999)],
+        save_history=save_history,
+        create_new=create_new,
+    )
```
vfdev-5, line 413: I think you do not need to add EMAHandler to check `ValueError`.

vfdev-5: Can we add a working test with EMAHandler and PiecewiseLinearStateScheduler like in the issue ?

vfdev-5 (approved):
Thanks for the update @fco-dv 
Can you improve tests a bit. Thanks !

tests/ignite/handlers/test_state_param_scheduler.py
fco-dv: Yes sure , thanks for your comments @vfdev-5 !

tests/ignite/handlers/test_state_param_scheduler.py
```diff
@@ -414,3 +391,69 @@ def __call__(self, event_index):
...
+    param_name = "ema_decay"
+
+    ema_handler = EMAHandler(model)
+    ema_handler.attach(trainer, name=param_name, event=Events.ITERATION_COMPLETED)
+
+    ema_decay_scheduler = PiecewiseLinearStateScheduler(
+        param_name=param_name, milestones_values=[(0, 0.0), (10, 0.999),], save_history=True
+    )
+    ema_decay_scheduler.attach(trainer, Events.ITERATION_COMPLETED)
+    trainer.run(data, max_epochs=20)
```
vfdev-5, line 459: @fco-dv looks good, but can we check somehow that these values were used in EMAHandler ? 
Maybe you can mock it and instead of updating a model, just ensure that values are the expected ones

fco-dv: Thanks for the hints @vfdev-5 , will try this approach!

fco-dv: @vfdev-5  don't you think we need now to remove the inner parameter scheduler associated to `_get_momentum` defined in the `EMAHandler`, otherwise I guess the values used by  `_update_ema_model` would not be those of `StateParamScheduler` ?

vfdev-5: Yes, we can deprecate them I think. Cc @sandylaker

sandylaker: @vfdev-5 @fco-dv Thanks for the cc. Yes the `_get_momentum` function should be deprecated. Once an `EMAHandler` is attached, a state parameter `momentum` will be created, and it will be scheduled by the `StateParamScheduler` if the user has specialized one. Therefore, we only need to read the `momentum` value from the engine's state when updating the ema model. I will create a PR after this one is merged to the master branch.

fco-dv: @sandylaker thanks for the clarification! Seems that our "working test" involving `EMAHandler` + `StateParamScheduler` would be more relevant after your PR... What do you think @vfdev-5, maybe we could add this test after the next PR ?

vfdev-5: Sounds good to me !

PR #2329, opened 2021-11-16
[skip ci][skip actions][skip netlify] add `run_code_style` for windows

sdesrozis:
Description:

Add a Windows `cmd` version of the script `run_code_style`. I'm not an expert of such a batch language but it seems to work pretty well as far I have tested.

Check list:

### DIFF
--- a/tests/run_code_style.bat
+++ b/tests/run_code_style.bat
@@ -1,0 +1,29 @@
+@ECHO OFF
+
+if "%1" == "lint" goto lint
+if "%1" == "fmt" goto fmt
+if "%1" == "mypy" goto mypy
+if "%1" == "install" goto install
+goto end
+
+:lint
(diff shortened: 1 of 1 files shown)


vfdev-5:
@sdesrozis thanks for the PR, can't we use bash scripts on windows as we probably be able to do on ci ?
Otherwise, should it also work with powershell ?

sdesrozis:
Yes, it works well with powershell. Using bash should be a good option, but it is not the usage in my company. Note that a win make.bat script is provided for documentation generation. I assumed that such scripts could be introduced as they really make the work easier.

Anyway, it increases the maintenance, I completely understand if this PR is rejected.

vfdev-5:
Ok for me to have it, thanks Sylvain !

PR #2333, opened 2021-11-24
Ema momentum

### LINKED ISSUE #2294: EMAHandler with warm-up counter-intuitive.
## ❓ Questions/Help/Support

Hello there,
First of all, thanks for the cool package, it's just super useful.

I recently checked out the new version of ignite, which provides a cool new feature: [EMAHandler](https://pytorch.org/ignite/_modules/ignite/handlers/ema_handler.html#EMAHandler).
The EMAHandler uses a momentum value (\alpha), which in contrast to most implementations that I've seen is the inverse decay value (i.e., 1- \alpha = decay).
Usually, when training an EMA model, the online model parameters are unreliable in the first couple epochs, thus in these cases, one might use a high \alpha value of i.e., 0.999, which leads to mainly copying the original parameters over to the EMA model.
Then after some epochs, the EMA models parameters are more reliable and thus \alpha can be set to i.e., 0.001 in order to incrementally update the model.

Thus, in most implementations, there is a linear warmup for the momentum value from i.e., 0.0 to the target decay value (i.e., 0.999).
However, in the current EMAHandler, the warmup momentum (\alpha) is required to be [strictly less than the target momentum](https://github.com/pytorch/ignite/blob/4a37e35dcb9548bb777dc86ab649ff390c6fb631/ignite/handlers/ema_handler.py#L134).
This, therefore leads to only two choices for the momentum: 

1. Start with a high \alpha value and a low (i.e., 0.0) warmup momentum, which just means that at the beginning \theta_{EMA} is used and late into training the online parameters \theta_t.
2. Don't use momentum warmup at all.

In the current EMAHandler implementation, it is impossible to start with a low \alpha (0.0) and increase that value to a high decay (0.999).

My question is, is there something I missed here or why is the implementation using warmup values strictly smaller than the target momentum?

sandylaker:
Fixes #2294 

Description: Deprecate the momentum scheduling function in the EMAHandler. 

Check list:

### DIFF
--- a/ignite/handlers/ema_handler.py
+++ b/ignite/handlers/ema_handler.py
@@ -5,8 +5,24 @@ from typing import Optional, Union
 import torch.nn as nn
 
 from ignite.engine import CallableEventWithFilter, Engine, Events, EventsList
+from ignite.handlers.state_param_scheduler import LambdaStateScheduler
 
 __all__ = ["EMAHandler"]
+
+
+class EMAWarmUp:
+    def __init__(self, momentum_warmup: float, warmup_iters: int, momentum: float) -> None:
+        self.momentum_warmup = momentum_warmup
+        self.warmup_iters = warmup_iters
+        self.momentum = momentum
+
+    def __call__(self, event_index: int) -> float:
+        denominator = max(1, self.warmup_iters - 1)
+        curr_momentum = self.momentum_warmup + (self.momentum - self.momentum_warmup) * (event_index - 1) / denominator
+        if self.momentum >= self.momentum_warmup:
+            return min(self.momentum, curr_momentum)
+        else:
+            return max(self.momentum, curr_momentum)
 
 
 class EMAHandler:
@@ -23,8 +39,8 @@ class EMAHandler:
           model: the online model for which an EMA model will be computed. If ``model`` is ``DataParallel`` or
               ``DistributedDataParallel``, the EMA smoothing will be applied to ``model.module`` .
           momentum: the update momentum after warmup phase, should be float in range :math:`\left(0, 1 \right)`.
-          momentum_warmup: the initial update momentum during warmup phase. This argument is not used.
-          warmup_iters: iterations of warmup. This argument is not used.
+          momentum_warmup: the initial update momentum during warmup phase.
+          warmup_iters: iterations of warmup.
 
     Attributes:
           ema_model: the exponential moving averaged model.
@@ -36,11 +52,6 @@ class EMAHandler:
           The EMA model is already in ``eval`` mode. If model in the arguments is an ``nn.Module`` or
           ``DistributedDataParallel``, the EMA model is an ``nn.Module`` and it is on the same device as the online
           model. If the model is an ``nn.DataParallel``, then the EMA model is an ``nn.DataParallel``.
-
-    .. warning::
-        The arguments `momentum_warmup` and `warmup_iters` will be deprecated in the future. In the current version,
-        these two argument have no effect on the momentum. I.e., the momentum will be constant during the training.
-        If users need to change the momentum, please use a ``StateParamScheduler``.
 
 
     Note:
@@ -89,6 +100,17 @@ class EMAHandler:
 
               trainer.run(...)
 
+          The following example shows how to perform warm-up to the EMA momentum:
+          .. code-block::python
+              device = torch.device("cuda:0")
+              model = nn.Linear(2, 1).to(device)
+              # linearly change the EMA momentum from 0.2 to 0.002 in the first 100 iterations,
+              # then keep a constant EMA momentum of 0.002 afterwards
+              ema_handler = EMAHandler(model, momentum=0.002, momentum_warmup=0.2, warmup_iters=100)
+              engine = Engine(step_fn)
+              ema_handler.attach(engine, name="ema_momentum")
+              engine.run(...)
+
           The following example shows how to attach two handlers to the same trainer:
 
           .. code-block:: python
@@ -127,25 +149,18 @@ class EMAHandler:
     ) -> None:
         if not 0 < momentum < 1:
             raise ValueError(f"Invalid momentum: {momentum}")
-        if momentum_warmup is not None:
-            warnings.warn(
-                "Argument 'momentum_warmup' will be deprecated in the future. In the current version,"
-                "it has no effect on the ema momentum. Please use the ParamStateScheduler to schedule"
-                "the ema momentum."
-            )
-        if warmup_iters is not None:
-            warnings.warn(
-                "Argument 'warmup_iters' will be deprecated in the future. In the current version, "
-                "it has no effect on the ema momentum. Please use the ParamStateScheduler to schedule"
-                "the ema momentum."
-            )
+        self.momentum = momentum
+        if momentum_warmup is not None and warmup_iters is not None:
+            self.momentum_scheduler = None
+            self._momentum_lambda_obj = EMAWarmUp(momentum_warmup, warmup_iters, momentum)
+        else:
+            self._momentum_lambda_obj = None  # type: ignore[assignment]
+
         if not isinstance(model, nn.Module):
             raise ValueError(
                 f"model should be an instance of nn.Module or its subclasses, but got"
                 f"model: {model.__class__.__name__}"
             )
-        # TODO: in the next version, rename momentum to init_momentum, which is more rigorous
-        self.momentum = momentum
 
         if isinstance(model, nn.parallel.DistributedDataParallel):
             model = model.module
@@ -195,10 +210,19 @@ class EMAHandler:
         if hasattr(engine.state, name):
             if warn_if_exists:
                 warnings.warn(
-                    f"Attribute '{name}' already exists in Engine.state. Turn off this warning by setting"
+                    f"Attribute '{name}' already exists in Engine.state. It might because 1. the engine has loaded its "
+                    f"state dict or 2. {name} is already created by other handlers. Turn off this warning by setting"
                     f"warn_if_exists to False.",
                     category=UserWarning,
                 )
         else:
             setattr(engine.state, name, self.momentum)
+
+        if self._momentum_lambda_obj is not None:
+            self.momentum_scheduler = LambdaStateScheduler(
+                self._momentum_lambda_obj, param_name="ema_momentum"
+            )  # type: ignore[assignment]
+
+            # first update the momentum and then update the EMA model
+            self.momentum_scheduler.attach(engine, event)  # type: ignore[attr-defined]
         engine.add_event_handler(event, self._update_ema_model, name)
--- a/ignite/handlers/ema_handler.py
+++ b/ignite/handlers/ema_handler.py
@@ -101,7 +101,9 @@ class EMAHandler:
               trainer.run(...)
 
           The following example shows how to perform warm-up to the EMA momentum:
-          .. code-block::python
+
+          .. code-block:: python
+
               device = torch.device("cuda:0")
               model = nn.Linear(2, 1).to(device)
               # linearly change the EMA momentum from 0.2 to 0.002 in the first 100 iterations,
--- a/ignite/handlers/ema_handler.py
+++ b/ignite/handlers/ema_handler.py
@@ -1,6 +1,6 @@
 import warnings
 from copy import deepcopy
-from typing import Union
+from typing import Optional, Union
 
 import torch.nn as nn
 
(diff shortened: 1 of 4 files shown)


vfdev-5:
Thanks for the PR @sandylaker !
Unfortunately, as we already released a stable version with the args you are removing here, we can not just remove them without deprecation cycle. Let's do the following, we keep the docstring as it was and add deprecation message suggesting to use `StateParamScheduler` and that we will remove these args in future releases. 
In the code we can first fix #2294 issue and also show deprecation warning if any of these params is used.

sandylaker:
@vfdev-5 Thanks for pointing out that. 

1. Do we have a specific class or function/decorator for the deprecation warning? Or should I adopt the Python's built-in `DeprecationWarning`?
2. Besides warning, should we keep using the internal function for updating the momentum? I.e., we just throw a warning when the handler performs warm-up. But in this case the `StateParamScheduler` still cannot control the momentum.

vfdev-5:
> Do we have a specific class or function/decorator for the deprecation warning? Or should I adopt the Python's built-in DeprecationWarning?

@sandylaker I think we can do as we are doing for `seed` arg in `Engine.run`, https://github.com/pytorch/ignite/blob/8a1de70e2152936ffcb051f3bd618138130626ec/ignite/engine/engine.py#L598

> Besides warning, should we keep using the internal function for updating the momentum? I.e., we just throw a warning when the handler performs warm-up. But in this case the StateParamScheduler still cannot control the momentum.

Sorry, I didn't quite follow everything in the code for `StateParamScheduler`, but if it is either one or another, then let's again do as for `seed` <=> if user uses it the warning is shown but no seed is set = nothing done.



cc @fco-dv

sandylaker:
@vfdev-5 So that means: we encourage the users to use `StateParamScheduler`. If they create an handler like this

```python3
ema_handler = EMAHandler(momentum=0.2, momentum_warmup=0.02, warmup_iters=100)
```
Then, 
1. a warning message is shown and saying that `momentum_warmup` and `warmup_iters` will be deprecated in the future.
2. The warm-up args have no actual effect. I.e., the momentum will be constant, unless users use `StateParamScheduler` to control it. This also means that `_update_momentum` and `_get_momentum` can be removed, we just read the momentum from `Engine.state`.
3. We keep the current docstring and the args.

vfdev-5:
@sandylaker yes, I was thinking about something like that. Do you think this is too restrictive ? In terms of API, I agree that it would require 2-3 additional lines with a state param scheduler in order to have the same behaviour. On the other hand we can keep both things: `momentum_warmup, warmup_iters` args and usage of state param handler and clearly say in docs for EMAHandler that if `momentum_warmup, warmup_iters` args are specified then they will overwrite any other scheduling (via state param scheduler or just a manual control).
What do you think or any other suggestions ?

sandylaker:
@vfdev-5 I think it is better to do like this:

>1. a warning message is shown and saying that momentum_warmup and warmup_iters will be deprecated in the future.
>2. The warm-up args have no actual effect, no warmup-up will be performed. I.e., the momentum will be constant, unless users use StateParamScheduler to control it. This also means that _update_ema_momentum and _get_momentum can be removed, we just read the momentum from Engine.state.
>3. We keep the current docstring and the args.

Because some users hope to schedule the `momentum` with a custom schedule. So we can provide this feature in the next version. 
At the same time, we will warn the users who is using the current version that they the warmup function will be deprecated. 
Users only need to update their code when they have set the warmup args.

So I think doing this will not break the BC.

vfdev-5:
@sandylaker sorry for delay, OK, I agree with you and let's make it as you are suggesting.

ignite/handlers/ema_handler.py
```diff
@@ -15,32 +16,32 @@ class EMAHandler:
...
 
     Note:
           The EMA model is already in ``eval`` mode. If model in the arguments is an ``nn.Module`` or
           ``DistributedDataParallel``, the EMA model is an ``nn.Module`` and it is on the same device as the online
           model. If the model is an ``nn.DataParallel``, then the EMA model is an ``nn.DataParallel``.
 
+    .. warning::
+        The arguments `momentum_warmup` and `warmup_iters` will be deprecated in the future. In the current version,
+        these two argument have no effect on the momentum. I.e., the momentum will be constant during the training.
+        If users need to change the momentum, please use a ``StateParamScheduler``.
```
vfdev-5, line 43: @sandylaker let's add an explicit example on how to do the momentum warm-up.

cc @fco-dv

ignite/handlers/ema_handler.py
```diff
@@ -125,25 +125,27 @@ def __init__(
...
             )
-        if warmup_iters is not None and not (isinstance(warmup_iters, int) and warmup_iters > 0):
-            raise ValueError(f"Invalid warmup_iters: {warmup_iters}")
         if not isinstance(model, nn.Module):
             raise ValueError(
                 f"model should be an instance of nn.Module or its subclasses, but got"
                 f"model: {model.__class__.__name__}"
             )
-        self.momentum_warmup = momentum_warmup
+        # TODO: in the next version, rename momentum to init_momentum, which is more rigorous
```
vfdev-5, line 147: Maybe, we could make it as a property to fetch `getattr(engine.state, name)` ?

ignite/handlers/ema_handler.py
```diff
@@ -179,36 +165,40 @@ def _update_ema_model(self, engine: Engine, name: str) -> None:
...
-            raise ValueError(
-                f"Attribute: '{name}' is already in Engine.state. Thus it might be "
-                f"overridden by other EMA handlers. Please select another name."
-            )
-
-        setattr(engine.state, name, 0.0)
-
-        # first update momentum, then update ema model
-        engine.add_event_handler(event, self._update_ema_momentum, name)
+            if warn_if_exists:
```
vfdev-5, line 196: nit: 
```python
if hasattr(engine.state, name) and warn_if_exists:
```

vfdev-5 (commented):
Thanks for the updates @sandylaker (and sorry for delay)
I left few other comments...

It would be nice to provide an explicit example in the docs on how to do the warm-up. Further more I wonder how could we simplify more explicit state param scheduler creation and make it a one-liner... @fco-dv any ideas ?

ignite/handlers/ema_handler.py
sandylaker: @vfdev-5 You mean the current momentum? Yes, but we need to first store the initial momentum with an attribute, say `init_momentum`. Then, when we attach the handler to the engine, we know how to initialize the momentum. After that, we can then retrieve the current momentum from the engine.

A problem is that a property usually does not allow to pass the second argument in addition to `self`. It means that if we want to make it a property, we need to write it like this:
```python3
@property
def momentum(self):
    return getattr(self.engine.state, self.name)
```
So we need to store the engine and the attribute name of the engine state. This might lead to a circular reference between `Engine` and `EMAHandler`.

If you think the `getattr` too ugly. We can make simply wrap it like this:

```python3
def get_curr_momentum(self, engine, name):
    return getattr(self.engine.state, self.name)
```
Since this function is called in `_update_ema_model`, the arguments `engine` and `name` are available.

vfdev-5: No, you are right, probably we can not delegate everything to engine.state as we still need the initial value. I was hesitating about the renaming `momentum` -> `init_momentum`.
Let's keep it as it is and keep your comment as well. Thanks for the explanation.

ignite/handlers/ema_handler.py
sandylaker: @vfdev-5

`hasattr(engine.state, name) = True` happens in two cases: 1. at the beginning of training, an attribute with the same name is created by another handler. 2.The engine resumes from a checkpoint. In the second case, the users might already know that the attribute exists and they can turn off the warning, and we should not reset the attribute value.

In other words, when `hasattr(engine.state, name) = True` and `warn_if_exists = False`, we do nothing.

vfdev-5: Oh, I see, sorry about that. You are right !

ignite/handlers/ema_handler.py
fco-dv: @vfdev-5 an example could be something like this I suppose, but I agree with your idea of simplification, maybe we could define a common `EMAWarmUp` class somewhere as it seems like users will always use these kind of scheduling ?

```python

    from ignite.handlers import EMAHandler

    model = nn.Linear(2, 1)
    trainer = Engine(lambda e, b: model(b))
    data = torch.rand(100, 2)
    
    class EMAWarmUp:
        def __init__(self, momentum_warmup, warmup_iters, momentum):
            self.momentum_warmup = momentum_warmup
            self.warmup_iters = warmup_iters
            self.momentum = momentum

        def __call__(self, event_index):
            denominator = max(1, self.warmup_iters - 1)
            momentum = self.momentum_warmup + (self.momentum - self.momentum_warmup) * (event_index - 1) / denominator
            return min(self.momentum, momentum)

    ema_state_parameter_scheduler = LambdaStateScheduler(
        param_name="momentum",
        lambda_obj=EMAState(momentum_warmup=0.0001, warmup_iters=10, momentum=0.002),
    )
    trainer = Engine(lambda e, b: model(b))

    ema_handler.attach(trainer, name="momentum", event=Events.ITERATION_COMPLETED)

    ema_state_parameter_scheduler.attach(trainer, Events.EPOCH_COMPLETED)
    trainer.add_event_handler(Events.EPOCH_COMPLETED, lambda _: print(trainer.state.momentum))

    trainer.run(data, max_epochs=20)
```

vfdev-5: I think we can just hide this `EMAWarmUp` and `ema_state_parameter_scheduler` and `ema_state_parameter_scheduler.attach` inside `ema_handler = EMAHandler(model, momentum=0.002, momentum_warmup=0.0001, warmup_iters=10)` : 

```python

class EMAHandler:

    def __init__(self, model, momentum=0.002, momentum_warmup=None, warmup_iters=None):
        # ...
        if momentum_warmup and warmup_iters:
            self.momentum_scheduler = # lazy LambdaStateScheduler creation

    def attach(self, engine, name, ...):
        # ...
        if self.momentum_scheduler is not None:
             # attach self.momentum_scheduler.attach(trainer, ...)
```

fco-dv: Looks good !

sandylaker: @vfdev-5 and @fco-dv There is a problem when implementing the API as the example above:

We need to pass a state parameter name to `LambdaStateScheduler`. However, we do not know the name until we pass it to `EMAHandler.attach`. 

A dirty fixup would be initializing `self.momentum_scheduler = None` in the `__init__` function. Next, we create the `self.momentum_scheduler = LambdaStateScheduler(...)` in `EMAHandler.attach` function.

fco-dv: Yes you're right and thanks for your reply @sandylaker!  for the fix, what you've mentioned could be a solution.  Parameter name may also be passed to the EMAHandler constructor  ? ( like for  `ParamScheduler` and `StateParamScheduler` classes), but I don't which one is the best approach ... maybe @vfdev-5 has another idea.

vfdev-5: @sandylaker yes, that's right we can not create `self.momentum_scheduler` directly that's why marked as `lazy LambdaStateScheduler creation`.
Here is a detailed implementation I was thinking:
```python
from ignite.handlers import LambdaStateScheduler

class EMAWarmUp:
    def __init__(self, momentum, momentum_warmup, warmup_iters):
        self.momentum = momentum        
        self.momentum_warmup = momentum_warmup
        self.warmup_iters = warmup_iters

    def __call__(self, event_index):
        denominator = max(1, self.warmup_iters - 1)
        momentum = self.momentum_warmup + (self.momentum - self.momentum_warmup) * (event_index - 1) / denominator
        return min(self.momentum, momentum)

class EMAHandler:

    def __init__(self, model, momentum=0.002, momentum_warmup=None, warmup_iters=None):
        # ...
        if momentum_warmup and warmup_iters:
            self.momentum_scheduler = None
            self._momentum_lambda_obj = EMAWarmUp(
                momentum=momentum, momentum_warmup=momentum_warmup, warmup_iters=warmup_iters
            )

    def attach(self, engine, name, event):
        # ...
        if self._momentum_lambda_obj is not None:
            self.momentum_scheduler = LambdaStateScheduler(
                param_name=name,
                lambda_obj=self._momentum_lambda_obj
            )            
            self.momentum_scheduler.attach(engine, Events.ITERATION_STARTED)
        # ...
```

What do you think ?

vfdev-5: @sandylaker @fco-dv do you think we can implement above solution in this PR ?

sandylaker: @vfdev-5 Sorry for the delay. 
Yes, I actually have the same idea as you. I will update this PR in this week.

tests/ignite/handlers/test_ema_handler.py
```diff
@@ -47,18 +47,51 @@ def test_ema_invalid_momentum(get_dummy_model, momentum):
...
-        EMAHandler(get_dummy_model(), momentum_warmup=momentum_warmup, momentum=momentum)
+    # momentum_warmup < momentum
+    model_1 = get_dummy_model()
+    engine_1 = Engine(_get_dummy_step_fn(model_1))
+    ema_handler_1 = EMAHandler(model_1, momentum, momentum_warmup_1, warmup_iters)
+    ema_handler_1.attach(engine_1)
+    engine_1.add_event_handler(
+        Events.ITERATION_COMPLETED, check_ema_momentum, momentum_warmup_1, momentum, warmup_iters
+    )
+    engine_1.run(range(2))
```
vfdev-5, line 84: Shouldn't we make at least 10 iters to properly check the scheduler ?

tests/ignite/handlers/test_ema_handler.py
```diff
@@ -47,18 +47,51 @@ def test_ema_invalid_momentum(get_dummy_model, momentum):
...
+
+    # momentum_warmup > momentum
+    model_2 = get_dummy_model()
+    engine_2 = Engine(_get_dummy_step_fn(model_2))
+    ema_handler_2 = EMAHandler(model_2, momentum, momentum_warmup_2, warmup_iters)
+    ema_handler_2.attach(engine_2)
+    engine_2.add_event_handler(
+        Events.ITERATION_COMPLETED, check_ema_momentum, momentum_warmup_2, momentum, warmup_iters
+    )
+    engine_2.run(range(2))
```
vfdev-5, line 94: Same here, IMO

tests/ignite/handlers/test_ema_handler.py
vfdev-5, line 84:
```suggestion
    engine_1.run(range(10))
```

tests/ignite/handlers/test_ema_handler.py
vfdev-5, line 94:
```suggestion
    engine_2.run(range(10))
```

vfdev-5 (approved):
Thanks a lot for the updates @sandylaker !
I'll update the tests and can merge the PR then

tests/ignite/handlers/test_ema_handler.py
```diff
@@ -98,56 +131,21 @@ def test_ema_load_state_dict(get_dummy_model):
...
-    engine.run(range(2))
-
-    # no warmup_iters
-    engine = Engine(step_fn)
-    ema_handler = EMAHandler(model, momentum=0.002, momentum_warmup=0.001, warmup_iters=None)
+    ema_handler = EMAHandler(model, momentum=0.002)
     ema_handler.attach(engine)
-    # attach the assertion handler after ema_handler, so the momentum is first updated and then tested
     engine.add_event_handler(Events.ITERATION_COMPLETED, assert_const_momentum, ema_handler.momentum)
     engine.run(range(2))
```
vfdev-5, line 146:
```suggestion
    engine.run(range(10))
```

PR #2344, opened 2021-12-10
Fix #2343

### LINKED ISSUE #2343: ClearrMLLogger freezes logging in distributed env
## 🐛 Bug description

Instantiating a `ClearMLLogger`instance in a distributed training freezes the logging of the subprocesses. The bug can be reproduced by executing the following script: https://github.com/H4dr1en/trains/commit/642c1130ad1f76db10ed9b8e1a4ff0fd7e45b3cc

I will open a PR to fix it.

## Environment

- PyTorch Version (e.g., 1.7.1):
- Ignite Version (e.g., 0.5.0):
- OS (e.g., Linux): linux
- How you installed Ignite (`conda`, `pip`, source): pip/source
- Python version: 3.8.5

H4dr1en:
Fixes #2343 

Description: ClearMLLogger tries to retrieve current task before trying to create a new one.

Check list:

### DIFF
--- a/ignite/contrib/handlers/clearml_logger.py
+++ b/ignite/contrib/handlers/clearml_logger.py
@@ -148,6 +148,7 @@ class ClearMLLogger(BaseLogger):
 
             self._task = _Stub()
         else:
+            # Try to retrieve current task before trying to create a new one
             self._task = Task.current_task()
             if self._task is None: 
                 self._task = Task.init(
--- a/ignite/contrib/handlers/clearml_logger.py
+++ b/ignite/contrib/handlers/clearml_logger.py
@@ -52,7 +52,7 @@ class ClearMLLogger(BaseLogger):
         kwargs: Keyword arguments accepted from
             `clearml.Task
             <https://clear.ml/docs/latest/docs/references/sdk/task#taskinit>`_.
-            All arguments are optional. If a ClearML Task has already been created, 
+            All arguments are optional. If a ClearML Task has already been created,
             kwargs will be ignored and the current ClearML Task will be used.
 
     Examples:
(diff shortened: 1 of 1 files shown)


vfdev-5:
@H4dr1en thanks for the PR ! Do you think we can write a distributed test for checking that everything is working ?

ignite/contrib/handlers/clearml_logger.py
```diff
@@ -147,12 +148,14 @@ def __setattr__(self, attr: str, val: Any) -> None:
 
             self._task = _Stub()
         else:
-            self._task = Task.init(
-                project_name=kwargs.get("project_name"),
-                task_name=kwargs.get("task_name"),
-                task_type=kwargs.get("task_type", Task.TaskTypes.training),
-                **experiment_kwargs,
-            )
+            self._task = Task.current_task()
```
vfdev-5, line 151: Could you please leave a comment here on what we are trying to do like the PR's description:
"ClearMLLogger tries to retrieve current task before trying to create a new one." ?

H4dr1en:
@vfdev-5 This might be very tricky to test: we want to make sure that the communication between the subprocesses and the main process works well, so that clearml can log successfully in the clearml-server. I have now idea how to test that :D

vfdev-5 (approved):
@H4dr1en thanks a lot for the PR ! 
For the tests I agree that this could be a bit tricky to write and there is no error to catch. 
Let's merge this like that and if possible to add a test we can do that later in a follow-up PR.
I still would like to hear from ClearML team.
@bmartinn could you please review as well. Thanks !


EDIT: There is a lint to fix and TPU tests are unrelated

PR #2345, opened 2021-12-10
[skip ci] add doctest for Accuracy metric

sdesrozis:
Addresses #2265

Description:  add doctest for `Accuracy` metric

@ydcjeff 

Check list:

### DIFF
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -105,18 +105,6 @@ class Accuracy(_BaseClassification):
     - `y` and `y_pred` must be in the following shape of (batch_size, num_categories, ...) and
       num_categories must be greater than 1 for multilabel cases.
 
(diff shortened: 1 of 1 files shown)


ignite/metrics/accuracy.py
```diff
@@ -126,6 +114,93 @@ def thresholded_output_transform(output):
...
+
+            0.6666...
+
+        Multiclass case
+
+        .. testcode:: 2
+
+            metric = Accuracy()
+            metric.attach(default_evaluator, "accuracy")
+            y_true = torch.Tensor([2, 0, 2, 1, 0, 1]).long()
```
vfdev-5, line 141: @sdesrozis tensors should be created as `torch.tensor(data)` and not `torch.Tensor(data)` (it was deprecated since v0.4.0). `torch.Tensor` is only used for matching object vs class, e.g. `isinstance(t, torch.Tensor)`.
Please fix that everywhere in doctest strings. Thanks

PR #2369, opened 2021-12-18
Added data as None option to Engine.run

### LINKED ISSUE #2354: Engine.run make input data optional
## 🚀 Feature

In certain use-cases, for example in RL or multi-data cases, it does not make much sense to call `engine.run` with an artificial data that only defines epoch lenght:
```python
trainer.run(data, max_epochs=100)
```
For example, 
https://github.com/pytorch/ignite/blob/4759e01034f53bb0334f661b8a8d61f76f6be9e4/examples/reinforcement_learning/reinforce.py#L70

https://github.com/pytorch/ignite/blob/4759e01034f53bb0334f661b8a8d61f76f6be9e4/examples/reinforcement_learning/reinforce.py#L118

The idea is to make `data` argument optional. Proposed API
```python
trainer.run(max_epochs=100, epoch_length=10_000)
```
In this case, `training_step(engine, batch)` will receive `None` as batch value.

cc @sdesrozis

vfdev-5:
Fixes #2354

Description:
- Added data as None option to `Engine.run`

Check list:

### DIFF
--- a/ignite/engine/deterministic.py
+++ b/ignite/engine/deterministic.py
@@ -198,8 +198,8 @@ class DeterministicEngine(Engine):
 
     def _setup_engine(self) -> None:
         if self.state.dataloader is None:
-            raise RuntimeError(
-                "Internal error, self.state.dataloader is None. Please, file an issue if you encounter this error."
+            raise ValueError(
+                "Deterministic engine does not support the option of data=None. Please, provide data as iterable"
             )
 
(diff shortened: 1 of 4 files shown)


sdesrozis (approved):
@vfdev-5 Thank you, LGTM !

trsvchn (approved):
LGTM! Thanks @vfdev-5 !

PR #2379, opened 2021-12-21
links changed to redirected links 

bibhabasumohapatra:
Description:
image

due to redirected links , link-test in CI failed. This changes links to redirected link.

### DIFF
--- a/docs/source/about.rst
+++ b/docs/source/about.rst
@@ -47,7 +47,7 @@ Join Core Team
 
 We are looking for motivated contributors to become collaborators and help out with the project.
 We can start considering a candidate after several successfully merged GitHub pull requests.
-If you are interested, for more details, please, contact Victor (vfdev-5#4458) on `PyTorch-Ignite Discord <https://discord.gg/djZtm3EmKj>`_.
+If you are interested, for more details, please, contact Victor (vfdev-5#4458) on `PyTorch-Ignite Discord <https://discord.com/invite/djZtm3EmKj>`_.
 
 
 Citing PyTorch-Ignite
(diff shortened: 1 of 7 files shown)


vfdev-5:
@bibhabasumohapatra link-test is not always failing and rather flaky8. For example, on master right now it passes. I wonder if we really need that much changes ?

trsvchn (commented):
@bibhabasumohapatra Thank you for the PR, but I don't think redirections cause linkcheck job faillures. To my knowledge http status codes for redirections (3xx) are not indicators of "bad links". as you can see linkcheck job works OK for the master:
https://github.com/pytorch/ignite/actions/workflows/docs.yml

In your case linkcheck failure was related not to redirections but to server error 500:
https://github.com/pytorch/ignite/runs/4586125674?check_suite_focus=true#step:6:116

>This error response is a generic "catch-all" response. Usually, this indicates the server cannot find a better 5xx error code to response. Sometimes, server administrators log error responses like the 500 status code with more details about the request to prevent the error from happening again in the future.

[MDN Web Docs: 500 Internal Server Error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500)

bibhabasumohapatra:
> @bibhabasumohapatra Thank you for the PR, but I don't think redirections cause linkcheck job faillures. To my knowledge http status codes for redirections (3xx) are not indicators of "bad links". as you can see linkcheck job works OK for the master: https://github.com/pytorch/ignite/actions/workflows/docs.yml
> 
> In your case linkcheck failure was related not to redirections but to server error 500: https://github.com/pytorch/ignite/runs/4586125674?check_suite_focus=true#step:6:116
> 
> > This error response is a generic "catch-all" response. Usually, this indicates the server cannot find a better 5xx error code to response. Sometimes, server administrators log error responses like the 500 status code with more details about the request to prevent the error from happening again in the future.
> 
> [MDN Web Docs: 500 Internal Server Error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500)

oh!  good to know that ... I will close this PR

PR #2380, opened 2021-12-21
added py39  pyproject.toml

bibhabasumohapatra:
adds py39 to pyproject.toml after updated black


vfdev-5 (approved):
Thanks @bibhabasumohapatra

PR #2409, opened 2022-01-04
removing 1.8.1 and 1.9.1 support for python 3.9

bibhabasumohapatra:
#2383

Description: removing python 3.9 PyTorch 1.8.1 and 1.7.1 support for PyTorch version tests

### DIFF
--- a/.github/workflows/pytorch-version-tests.yml
+++ b/.github/workflows/pytorch-version-tests.yml
@@ -26,6 +26,10 @@ jobs:
             python-version: 3.9 
           - pytorch-version: 1.6.0
             python-version: 3.9 
+          - pytorch-version: 1.7.1
+            python-version: 3.9
+          - pytorch-version: 1.8.1
+            python-version: 3.9
 
     steps:
       - uses: actions/checkout@v2


.github/workflows/pytorch-version-tests.yml
```diff
@@ -26,6 +26,10 @@ jobs:
             python-version: 3.9 
           - pytorch-version: 1.6.0
             python-version: 3.9 
+          - pytorch-version: 1.7.1
```
vfdev-5, line 29: Let's add a comment why we excluding these confs

bibhabasumohapatra: ok,  
for 1.7.1 and 1..8.1 it is temporary right?

bibhabasumohapatra: > Let's add a comment why we excluding these confs

#### Like:

-  `python-version:` is list of versions of python.
-  `pytorch-version: ` is list of versions of PyTorch.
- and in `exclude:` we can give the particular PyTorch and python version to remove from testing.
Does sound right?

vfdev-5: No, a comment explaning why we excluding 3.9/1.7.1 and 3.9/1.8.1 and put a link on https://github.com/pytorch/ignite/issues/2383

bibhabasumohapatra: okay.

.github/workflows/pytorch-version-tests.yml
```diff
@@ -26,6 +26,12 @@ jobs:
             python-version: 3.9 
           - pytorch-version: 1.6.0
             python-version: 3.9 
+          # disabling python 3.9 support with PyTorch 1.7.1 and 1.8.1,to stop repeated pytorch-version test fail.
```
vfdev-5, line 29:
```suggestion
          # disabling python 3.9 support with PyTorch 1.7.1 and 1.8.1, to stop repeated pytorch-version test fail.
```

.github/workflows/pytorch-version-tests.yml
```diff
@@ -5,7 +5,7 @@ on:
   schedule:
     # Run at 00:00 UTC Every Day
     - cron: "0 0 * * *"
-
+
```
vfdev-5, line 8:
```suggestion

```

.github/workflows/pytorch-version-tests.yml
```diff
@@ -5,7 +5,6 @@ on:
   schedule:
     # Run at 00:00 UTC Every Day
     - cron: "0 0 * * *"
```
vfdev-5, line 7:
```suggestion
    - cron: "0 0 * * *"

```

vfdev-5 (approved):
Thanks @bibhabasumohapatra

PR #2416, opened 2022-01-10
set master_addr master_port default values

### LINKED ISSUE #2312: ignite.distributed.Parallel: master_port seems always to be 2222 when using env:// init_method
I want to use another master_port in my machine. My code is below:

```python
def run(
    backend='nccl',
    resume=True,
    nproc_per_node=4,
    with_clearml=False,
    with_amp=False,
    **spawn_kwargs
):
    config = locals()
    config.update(config["spawn_kwargs"])
    del config["spawn_kwargs"]
    spawn_kwargs["nproc_per_node"] = nproc_per_node
    if backend == "xla-tpu" and with_amp:
        raise RuntimeError("The value of with_amp should be False if backend is xla")

    os.environ['CUDA_VISIBLE_DEVICES']   = "0,1,2,3"

    with ignite.distributed.Parallel(backend=backend, **spawn_kwargs) as parallel:

        parallel.run(training, config)


if __name__ == '__main__':
    fire.Fire(run)
```

I use the command to run this:

`python normal_train_ddp.py  --matser-port 50001`

I found it always use master port 2222.

I found this code.  It seems ignore my master_port setting when init_method is None or env://. 

https://github.com/pytorch/ignite/blob/8a1de70e2152936ffcb051f3bd618138130626ec/ignite/distributed/comp_models/native.py#L385-L388

## Environment

- PyTorch Version (e.g., 1.4): 1.8
- OS (e.g., Linux): Linux
- Python version: 3.8

fco-dv:
Fixes #2312

Description: Fix static `master_addr` `master_port` values in `spawn` method by setting default values in method prototype.

Check list:

### DIFF
--- a/ignite/distributed/comp_models/native.py
+++ b/ignite/distributed/comp_models/native.py
@@ -354,8 +354,8 @@ if has_native_dist_support:
             nproc_per_node: int = 1,
             nnodes: int = 1,
             node_rank: int = 0,
-            master_addr: Optional[str] = None,
-            master_port: Optional[int] = None,
+            master_addr: Optional[str] = "127.0.0.1",
+            master_port: Optional[int] = 2222,
             backend: str = "nccl",
             init_method: Optional[str] = None,
             **kwargs: Any,
@@ -380,10 +380,8 @@ if has_native_dist_support:
                 spawn_kwargs["start_method"] = kwargs.get("start_method", default_start_method)
                 start_processes = mp.start_processes
 
-            if init_method in [None, "env://"]:
+            if init_method is None:
                 init_method = "env://"
-                master_addr = "127.0.0.1"
-                master_port = 2222
 
             start_processes(
                 _NativeDistModel._dist_worker_task_fn,
--- a/ignite/distributed/comp_models/native.py
+++ b/ignite/distributed/comp_models/native.py
@@ -386,6 +386,10 @@ if has_native_dist_support:
                     master_port = 2222
                 if master_addr is None:
                     master_addr = "127.0.0.1"
+            elif master_addr is not None:
+                raise ValueError("master_addr should be None if init_method is provided other then 'env://'")
+            elif master_port is not None:
+                raise ValueError("master_port should be None if init_method is provided other then 'env://'")
 
             start_processes(
                 _NativeDistModel._dist_worker_task_fn,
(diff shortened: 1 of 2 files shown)


ignite/distributed/comp_models/native.py
```diff
@@ -354,8 +354,8 @@ def spawn(  # type: ignore[override]
             nproc_per_node: int = 1,
             nnodes: int = 1,
             node_rank: int = 0,
-            master_addr: Optional[str] = None,
-            master_port: Optional[int] = None,
+            master_addr: Optional[str] = "127.0.0.1",
+            master_port: Optional[int] = 2222,
```
vfdev-5, line 358: This fixes the issue, but if user defines `init_method` as something unrelated to tcp store (e.g. filestore), then seeing defined `master_addr` and `master_port` and `init_method` together can make things a bit cumbersome...
How about keeping them as None (same as previously) but defining as you do in case of init_method is None and they are undefined (are Nones) ?

fco-dv: Yes I agree I was also thinking about that, but didn't know clearly the impact for the other methods, will fix this , thanks for your comment @vfdev-5

vfdev-5:
@fco-dv thanks for the update. Can we make a test of this feature, please ?

vfdev-5:
@fco-dv I updated tests and the code, please review. Thanks!

tests/ignite/distributed/comp_models/test_native.py
```diff
@@ -499,31 +509,56 @@ def _test__native_dist_model_spawn(backend, num_workers_per_machine, device, ini
...
+        init_method = None
+        spawn_kwargs["master_addr"] = "0.0.0.0"
+        spawn_kwargs["master_port"] = 2345
 
-    num_workers_per_machine = torch.cuda.device_count()
+    nproc = torch.cuda.device_count()
     _test__native_dist_model_spawn(
-        "nccl", num_workers_per_machine=num_workers_per_machine, device="cuda", init_method=init_method
+        "nccl", num_workers_per_machine=nproc, device="cuda", init_method=init_method, **spawn_kwargs
     )
```
fco-dv, line 563: may add also this like in the `gloo` case but it's a bit redundant ... ?
```python
if init_method not in [None, "env://"]:
    with pytest.raises(ValueError, match=r"master_addr should be None if init_method is provided"):
        _test__native_dist_model_spawn(
            "nccl",   ... , master_addr="abc"
        )
    with pytest.raises(ValueError, match=r"master_port should be None if init_method is provided"):
        _test__native_dist_model_spawn(
            "nccl", ... ,  master_port=123
        )
```

vfdev-5: yes, it is redundant and IMO does not depend on the backend, this do not check for all backends.

PR #2421, opened 2022-01-14
fixing issue #2328 from PyData's sprint

### LINKED ISSUE #2328: WeightsHistHandler should plot all weights (incl those without grad)
## 🚀 Feature

```python

            name = name.replace(".", "/")
            logger.writer.add_histogram(
            )

```

abeeraleisa:
Fixes #2328

Description:

Check list:
- Updated the class WeightsHistHandler 
-  Tested the newly added changes on the class only 
-  Didn't update any documents


vfdev-5:
Closing in favor of #2550

PR #2422, opened 2022-01-14
Replace all os.path in ignite and tests by pathlib.Path good first issue help wanted

elaugh9:
Fixes #{2418}

Description:
I worked on steps 1, 2, 3. Step 4 is still not done. 
  
Check list:


sdesrozis:
@elaugh9 Thanks for your work ! IMO you should modify the raw data of notebooks rather than use jupyter.

Issue #2325, opened 2021-11-13
Update pytorch.org/ignite site and reference pytorch-ignite.ai


vfdev-5:
## 📚 Documentation

Our project new website is online and available. We need to update pytorch.org/ignite to reflect only API and provide links to pytorch-ignite.ai as source for examples/tutorials/how-to guides etc.

### Redirections:

Quickstart -> https://pytorch-ignite.ai/tutorials/beginner/01-getting-started/
FAQ -> https://pytorch-ignite.ai/how-to-guides/
Concepts -> TBD
Examples -> TBD
Github -> Github (no change)
About us -> https://pytorch-ignite.ai/about/community/

### Index page (https://pytorch.org/ignite/)

Remove almost all the content and add the logo and the link to https://pytorch-ignite.ai. Like below (but better aligned):

<div>
  <img width=200 src=https://pytorch-ignite.ai/_images/ignite_logo.svg />
  <p>https://pytorch-ignite.ai</p>
</div>

### Add another explicit link to https://pytorch-ignite.ai from the header ?

The idea is to have something like: 
```
Quickstart    Concepts    Examples     FAQ     GitHub     About us    https://pytorch-ignite.ai
```
Any thoughts ?

iamhardikat11:
Is this issue open to be taken up?

vfdev-5:
Hi @iamhardikat11 , this issue is a bit critical for the project, but yes, you can help with a part of it, for example, replacing the links for: about us, FAQ, Quickstart (as suggested in the description) and remove those .rst files.

iamhardikat11:
Yup, I can work on that. I will Submit my PR By today Night.

Moh-Yakoub:
> Hi @iamhardikat11 , this issue is a bit critical for the project, but yes, you can help with a part of it, for example, replacing the links for: about us, FAQ, Quickstart (as suggested in the description) and remove those .rst files. What do you think ?

@vfdev-5 Any other parts that I can help with at the moment?

vfdev-5:
@Moh-Yakoub yes, you can help. Nothing is done at the moment. You can submit a PR with docs updates for the redirections (see the description). What do you think ?

Moh-Yakoub:
> @Moh-Yakoub yes, you can help. Nothing is done at the moment. You can submit a PR with docs updates for the redirections (see the description). What do you think ?

Great, I will start doing that. Thanks a lot!

Moh-Yakoub:
[...]

PR #2431, opened 2022-01-18
Update pytorch.org/ignite site and reference pytorch-ignite.ai #2325

Moh-Yakoub:
Related to #2325

Description: I've updated some of the links as specified in the associated tasks and updated docs accordingly

To test it
```
1. cd docs
2. make html
3. cd build/
4. python -m http.server 8989
```

Checked the links and they referencing the updated links instead of the old html.

**I didn't change the main URL (https://pytorch.org/ignite')  to the new URL because it will break the pages not available for the new URL (for instance: examples), will change the base to the new URL once all of the URLs are available on the (pytorch-ignite.ai/) URL**

Check list:

### DIFF
--- a/docs/source/_templates/layout.html
+++ b/docs/source/_templates/layout.html
@@ -129,7 +129,7 @@
       <div class="main-menu">
         <ul>
           <li>
-            <a href="{{ pathto('quickstart') }}">Quickstart</a>
+            <a href="{{ theme_variables.external_urls['quickstart'] }}">Quickstart</a>
           </li>
 
           <li>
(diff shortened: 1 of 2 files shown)


vfdev-5 (approved):
@Moh-Yakoub thanks for the PR ! It looks good and we can merge it.
However, in order to finalize the migration we need to make a bit more updates to the API ref website. I'll update the issue description.
Basically, we need to have consistent reference to https://pytorch-ignite.ai
The index page have to minimally expose a reference https://pytorch-ignite.ai ...

PR #2435, opened 2022-01-19
Fix for issues related to mypy

### LINKED ISSUE #2361: Fix mypy issues with mypy==0.920
## 🐛 Bug description

Our code base mypy check is failing:
```
ignite/handlers/time_profilers.py:638: error: Incompatible return value type
(got
"List[List[Union[Any, str, float, Tuple[Union[str, float], Union[str, float]]]]]",
expected "List[List[Union[str, float]]]")  [return-value]
            return event_handler_stats
                   ^
ignite/handlers/param_scheduler.py:829: error: unused "type: ignore" comment
                scheduler = cls(save_history=False, lr_scheduler=lr_schedu...
                ^
ignite/handlers/checkpoint.py:721: error: unused "type: ignore" comment
                    tmp_file = tmp.file  # type: ignore
                    ^
ignite/contrib/handlers/neptune_logger.py:588: error: unused "type: ignore"
comment
                torch.save(checkpoint, tmp.file)  # type: ignore[attr-defi...
                ^
Found 4 errors in 4 files (checked 103 source files)
```
for mypy==0.920

Let's make our codebase compatible.

## Environment

- OS (e.g., Linux):
- Python version:

sayantan1410:
Fixes #2361 

Check list:
- [X] Bug Fix

Let me know if there is any mistake @vfdev-5

### DIFF
--- a/ignite/contrib/handlers/neptune_logger.py
+++ b/ignite/contrib/handlers/neptune_logger.py
@@ -1,3 +1,4 @@
+# mypy: no-warn-unused-ignores
 """Neptune logger and its helper handlers."""
 import tempfile
 from typing import Any, Callable, List, Mapping, Optional, Union
--- a/ignite/contrib/handlers/neptune_logger.py
+++ b/ignite/contrib/handlers/neptune_logger.py
@@ -1,4 +1,3 @@
-# mypy: no-warn-unused-ignores
 """Neptune logger and its helper handlers."""
 import tempfile
 from typing import Any, Callable, List, Mapping, Optional, Union
@@ -586,7 +585,7 @@ class NeptuneSaver(BaseSaveHandler):
         with tempfile.NamedTemporaryFile() as tmp:
             # we can not use tmp.name to open tmp.file twice on Win32
             # https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile
-            torch.save(checkpoint, tmp.file)  # type: ignore[attr-defined]
+            torch.save(checkpoint, tmp.file)
             self._logger.log_artifact(tmp.name, filename)
 
     @idist.one_rank_only(with_barrier=True)
--- a/docs/source/_templates/layout.html
+++ b/docs/source/_templates/layout.html
@@ -129,7 +129,7 @@
       <div class="main-menu">
         <ul>
           <li>
-            <a href="{{ pathto('quickstart') }}">Quickstart</a>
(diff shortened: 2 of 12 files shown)


ignite/contrib/handlers/neptune_logger.py
```diff
@@ -1,3 +1,4 @@
+# mypy: no-warn-unused-ignores
```
vfdev-5, line 1: @sayantan1410 do you know which objects/variables are unused ? 
My question is why we put this and can't we fix the warning otherwise ?

vfdev-5 (commented):
Thanks for the PR @sayantan1410 
I left a question

ignite/contrib/handlers/neptune_logger.py
sayantan1410: Like instead of ignoring the type errors for the entire file should I do it for like each of the variable or something like that ?

vfdev-5: Yes, the purpose is to fix the error instead of ignoring it

sayantan1410: Okay, trying to fix that. Thank you

sayantan1410:
@vfdev-5 I have tried to look at each of the error codes in the lines where the errors where generating, Let me know if I have made some mistake.

vfdev-5 (approved):
LGTM, thanks for the update @sayantan1410 !
Let's see if mypy is passing and we can merge it.

vfdev-5:
@sayantan1410 can you fix flake8 error:
```
+ flake8 ignite tests examples --config setup.cfg
ignite/handlers/checkpoint.py:721:36: W291 trailing whitespace
```
- https://github.com/pytorch/ignite/runs/4882332288?check_suite_focus=true#step:11:57

sayantan1410:
@vfdev-5 Sorry, probably there were two white spaces and I removed one and that's why it failed again, actually the extra white space did not show in my code editor.

vfdev-5:
@sayantan1410 if you are using VSCode, you can trim extra white spaces with this option:
- https://stackoverflow.com/questions/30884131/remove-trailing-spaces-automatically-or-with-a-shortcut

vfdev-5:
@sayantan1410 failing test is codecov patch, this one is not always reliable indicator and we ignore it sometimes

sayantan1410:
@vfdev-5 Okay, thank you, also yes I was using VS Code and I changed the settings for trimming white spaces to true.

PR #2442, opened 2022-01-25
Fix doctest for LR Scheduler

sdesrozis:
Description:

Check list:


vfdev-5 (approved):
Thanks @sdesrozis !

PR #2457, opened 2022-02-09
Add toggle for default setup in docs

### LINKED ISSUE #2436: [doctest] Provide definition of `default_evaluator` and other default objects
## 📚 Documentation

It was already discussed and also pointed out in the forum that it is unclear how default objects are defined inside doctest examples. For example: https://pytorch.org/ignite/generated/ignite.metrics.Accuracy.html#ignite.metrics.Accuracy

```python
metric = Accuracy()
metric.attach(default_evaluator, "accuracy")
y_true = torch.Tensor([1, 0, 1, 1, 0, 1])
y_pred = torch.Tensor([1, 0, 1, 0, 1, 1])
state = default_evaluator.run([[y_pred, y_true]])
print(state.metrics["accuracy"])
```

We can think of different solutions:
- 1) for all doctests, add a link on https://github.com/pytorch/ignite/blob/111fc7a0c57bb4d96d94f300f8c2f188f7bda148/docs/source/conf.py#L334-L378
- 2) visible code is as above but if code is copied we add automatically all defaults (not good as a solution)
- 3) other ideas ?

DevPranjal:
Fixes #2436 

Description: Added toggles for default setup in docs to avoid confusion

Check list:


vfdev-5:
@DevPranjal there is a warning that makes docs build to fail: 
```
/home/runner/work/ignite/ignite/docs/source/defaults.rst: WARNING: document isn't included in any toctree
```
I haven't that warning when built locally

docs/source/defaults.rst
```diff
@@ -0,0 +1,47 @@
+.. toggle::
+
+    .. testcode:: default, 1, 2, 3, 4, 5
```
vfdev-5, line 3: I think this can be just 
```suggestion
    .. code-block:: python
```

DevPranjal: If I write `.. code-block:: python` the block would not run and hence the default setup would no be defined. This is also confirmed when I try it.

docs/source/conf.py
```diff
@@ -331,51 +343,51 @@ def run(self):
 ]
 
 # doctest config
-doctest_global_setup = """
-from collections import OrderedDict
+# doctest_global_setup = """
```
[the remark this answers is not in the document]
DevPranjal: I think we can just run the default setup within the `.. testcode::` directive. This will allow us to change at only one place in the future if any more setup comes in

DevPranjal:
> @DevPranjal there is a warning that makes docs build to fail
I haven't that warning when built locally

I had the warning locally too, but the built docs were displaying fine. After [searching](https://stackoverflow.com/questions/15249340/warning-document-isnt-included-in-any-toctree-for-included-file#:~:text=Does%20this%20warning%20always%20appear,rather%20than%20included%20via%20toctree%3F&text=AFAIK%2C%20every%20.,%2Fa%2F13109215%2F407651.), I have added the `:orphan:` metadata tag on top in `defaults.rst`

docs/source/conf.py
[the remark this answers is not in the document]
vfdev-5: You are right, thanks for pointing out that ! Let's try to do as you suggest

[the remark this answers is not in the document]
DevPranjal: So shall I just delete the commented block completely?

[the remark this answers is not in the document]
vfdev-5: Yes, we can delete it

DevPranjal:
@vfdev-5 I have edited the files as discussed to fix the problems with `:orphan:` tag.

docs/source/defaults.rst
```diff
@@ -0,0 +1,49 @@
+:orphan:
+
+.. toggle::
+
+    .. testcode:: default, 1, 2, 3, 4, 5
```
vfdev-5, line 5: @DevPranjal what does `default, 1, 2, 3, 4, 5` ?

vfdev-5 (approved):
Thanks a lot @DevPranjal , looks great !
Just a small question and we are good to go

docs/source/defaults.rst
DevPranjal: In the doctests we have (in some places) grouped the tests like `1, 2, 3` so that variables from one doctest are no used in the other. Here `1, 2, 3` implies that this testcode will defined for each of those groups (and separately). Not putting these will define this setup only for the `default` group. 

There is also an alternative to this. The docs say we can put `*` instead of naming all the groups, but that does not seem to work when I tried it out.

vfdev-5: @DevPranjal let's move with that and if @sdesrozis have another idea we can make a follow-up PR

DevPranjal: Great!

vfdev-5:
@DevPranjal can you please fetch the latest commit from your fork to your local repository (and thus have [1341dff](https://github.com/pytorch/ignite/pull/2457/commits/1341dffe5b81564d04ba4854ae154bf80e59d194) commit) and make an empty commit with amend (just `git commit --amend`) and push force to your fork such that we can restart the CI and `autopep 8` will have nothing to fix in terms of code formatting. Currently the latest commit does not run doctests. Thanks

PR #2460, opened 2022-02-12
Switch formatting from black+isort to µfmt (black+µsort)

jreese:
Description:

This switches linting, formatting, and pre-commit checks from using black and isort directly to using [µfmt](https://ufmt.omnilib.dev), which runs [µsort](https://usort.readthedocs.io) and black as an atomic formatting step. This matches what torchvision uses (pytorch/vision#4384).

Covered files were formatted using `./tests/run_code_style.sh fmt` and `pre-commit run` validated the results.

Check list:

### DIFF
--- a/tests/run_code_style.sh
+++ b/tests/run_code_style.sh
@@ -10,5 +10,5 @@ elif [ $1 = "fmt" ]; then
 elif [ $1 = "mypy" ]; then
     mypy --config-file mypy.ini
 elif [ $1 = "install" ]; then
-    pip install flake8 "black==21.12b0" "usort==1.0.1" "ufmt==1.3.1.post1" "mypy==0.910"
+    pip install flake8 "black==21.12b0" "usort==1.0.1" "ufmt==1.3.1.post1" "mypy"
 fi
--- a/tests/run_code_style.sh
+++ b/tests/run_code_style.sh
@@ -4,13 +4,11 @@ set -xeu
 
 if [ $1 = "lint" ]; then
     flake8 ignite tests examples --config setup.cfg
-    isort . --check --settings setup.cfg
-    black . --check --config pyproject.toml
+    ufmt diff .
 elif [ $1 = "fmt" ]; then
-    isort . --settings setup.cfg
-    black . --config pyproject.toml
+    ufmt format .
 elif [ $1 = "mypy" ]; then
     mypy --config-file mypy.ini
 elif [ $1 = "install" ]; then
-    pip install flake8 "black==21.12b0" "isort==5.7.0" "mypy"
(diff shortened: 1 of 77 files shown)


tests/run_code_style.sh
```diff
@@ -4,13 +4,11 @@ set -xeu
...
+    ufmt diff .
 elif [ $1 = "fmt" ]; then
-    isort . --settings setup.cfg
-    black . --config pyproject.toml
+    ufmt format .
 elif [ $1 = "mypy" ]; then
     mypy --config-file mypy.ini
 elif [ $1 = "install" ]; then
-    pip install flake8 "black==21.12b0" "isort==5.7.0" "mypy"
+    pip install flake8 "black==21.12b0" "usort==1.0.1" "ufmt==1.3.1.post1" "mypy==0.910"
```
vfdev-5, line 13: Why do you fix mypy to 0.910 ?

vfdev-5 (commented):
@jreese thanks for the PR ! Yes, ufmt works well in torchvision and maintainers are rather happy with it. 
I left a comment and let me see what other maintainers think about the change as code formatting could be rather opinionated :)

tests/run_code_style.sh
jreese: It was pinned in `run_code_style.bat` as well, so I just assumed it was intended to have pinned versions for everything. Happy to change it back to unpinned if that's desired.

vfdev-5: oh, I see, we forgot to unpin it there. Yes, please unpin it.

jreese: Done!

tests/run_code_style.sh
```diff
@@ -4,13 +4,11 @@ set -xeu
 
 if [ $1 = "lint" ]; then
     flake8 ignite tests examples --config setup.cfg
-    isort . --check --settings setup.cfg
-    black . --check --config pyproject.toml
+    ufmt diff .
```
vfdev-5, line 7: @jreese i wonder why here you use `ufmt diff .` and in `tests/run_code_style.bat` : `ufmt check .` ?

jreese: Ah, I originally had it as check in both places to match existing behavior, but switched it to diff when I was working out the correct configuration. I personally think diff is more useful, and would suggest using that in both places, but would defer that decision to your opinion.

vfdev-5: I see. If they both behave similarly in CI and diff shows the diff. I'm ok with `ufmt diff .` everywhere.

jreese:
Rebased onto latest and unified on using `usort diff`.

vfdev-5 (approved):
Thanks @jreese !

Issue #583, opened 2019-08-22
Fix Gradients accumulation example 


dobosevych:
Link: https://pytorch.org/ignite/faq.html
It should be
```
accumulation_steps = 4

def update_fn(engine, batch):
    model.train()

    x, y = prepare_batch(batch, device=device, non_blocking=non_blocking)
    y_pred = model(x)
    loss = criterion(y_pred, y) / accumulation_steps
    loss.backward()

    if engine.state.iteration % accumulation_steps == 0:
        optimizer.step()
        optimizer.zero_grad()

    return loss.item()

trainer = Engine(update_fn)
```
instead of 
```
accumulation_steps = 4

def update_fn(engine, batch):
    model.train()

    if engine.state.iteration % accumulation_steps == 0:
        optimizer.zero_grad()

    x, y = prepare_batch(batch, device=device, non_blocking=non_blocking)
    y_pred = model(x)
    loss = criterion(y_pred, y) / accumulation_steps
    loss.backward()

    if engine.state.iteration % accumulation_steps == 0:
        optimizer.step()

    return loss.item()

trainer = Engine(update_fn)
```

vfdev-5:
@dobosevych thanks! yes, there is a bug in FAQ with
```python
    if (engine.state.iteration + 1) % accumulation_steps == 0:
        optimizer.step()
```
Your code is shorter, I'll change the FAQ to this.

PR #2470, opened 2022-02-16
Fix gradients flushed at the end for 'supervised_train_step' (#2459)

### LINKED ISSUE #2459: GradsScalarHandler logs 0 gradients if default update function is used
## 🐛 Bug description

**Steps to reproduce:**

**Solution proposal**

```
            optimizer.zero_grad()
        model.train()
        y_pred = model(x)
        loss = loss_fn(y_pred, y)
        loss.backward()
            optimizer.step()
```

- OS (e.g., Linux):
- Python version: 3.9.7

egaznep:
I  have not changed the behavior of AMP, APEX, TPU as I don't have the means to test it.

Fixes #2459 

Description: https://github.com/pytorch/ignite/issues/583 reported that gradient accumulation did not work at all. The fix done at the time broke gradient loggability for sake of correct gradient accumulation. This fix achieves both objectives.

The behavior for AMP, APEX and TPU is not changed as I cannot test them.

Check list:


vfdev-5:
@egaznep can you update the code for amp, apex and tpu please ?

egaznep:
> @egaznep can you update the code for amp, apex and tpu please ?

I just found the time today and did the changes. Is there a "training test" of some sort that is built-in that I can perform before submitting? Also sadly I don't have docker and can't install it either as I don't have admin rights on this machine.

vfdev-5:
@egaznep we have mnist example which uses create_supervised_trainer: https://github.com/pytorch/ignite/blob/master/examples/mnist/mnist.py   but we didn't expose grad accumulation option.
I think you can update the code and I'll test it myself

vfdev-5:
@egaznep thanks for the update! Can I ask you to add a test to https://github.com/pytorch/ignite/blob/master/tests/ignite/engine/test_create_supervised.py that checking that model grads are not empty between iterations (and thus could be logged) ?

vfdev-5:
Closing in favor of #2560

egaznep:
> Closing in favor of #2560

I was extremely busy with my master's thesis. I am sorry that I could not complete the remaining tasks on time. I was planning to address the changes soon but it seems that someone else has already done it nicely. Thank you all for maintaining this immensely useful library.

vfdev-5:
@egaznep no problems, hope your master thesis is done and in a good shape !
Yes, your commits from this PR were reused in above referenced PR such that your contribution is also reported in the commits history.

egaznep:
@vfdev-5 yes it finally is! Thank you, I appreciate it.

PR #2471, opened 2022-02-16
Doc fixes

### LINKED ISSUE #2157: `Timer`'s misleading behaviour when epoch completion time calculated
When I tried calculating the time taken to complete a single epoch via `Timer`, the handlers attached to `trainer` before `Timer` were executed first, and thus their time also got recorded by the `Timer` too for a single epoch. Therefore the true time taken for epoch completion, provided by `trainer.state.times`, is less than what `Timer` calculated. This can be misleading. More clarification on how this actually works in the docs would be appreciated. Or the `Timer`'s functionality can be enhanced to `step` before all other handlers attached to an event could also be helpful. Notebook to quickly verify this [here](https://colab.research.google.com/drive/1xX7slM9BiDSWYGId5zB_8ISneJ0wezAM?usp=sharing).

FarehaNousheen:
Fixes #2157

Description:

Check list:

### DIFF
--- a/ignite/handlers/timing.py
+++ b/ignite/handlers/timing.py
@@ -22,7 +22,8 @@ class Timer:
     Note:
         When using ``Timer(average=True)`` do not forget to call ``timer.step()`` every time an event occurs. See
         the examples below.
-
+    Note:
+        Timer if attached to measure single epoch time, it can also record the time of all handlers attached prior to Timer. To obtain the actual time taken for the epoch completion, handlers will have to be rearranged.
     Examples:
         Measuring total time of the epoch:
 


vfdev-5:
@FarehaNousheen can you reformat the code using: `run_code_style.bat fmt` ?

ignite/handlers/timing.py
```diff
@@ -22,7 +22,8 @@ class Timer:
     Note:
         When using ``Timer(average=True)`` do not forget to call ``timer.step()`` every time an event occurs. See
         the examples below.
-
+    Note:
+        Timer if attached to measure single epoch time, it can also record the time of all handlers attached prior to Timer. To obtain the actual time taken for the epoch completion, handlers will have to be rearranged.
```
vfdev-5, line 26:
```suggestion
         If ``Timer`` is attached to measure single epoch time, it can also record the time of all handlers attached prior to it. To obtain the actual time taken for the epoch completion, handlers have to be rearranged.
```

FarehaNousheen:
Hi Victor,
Let me try this.

vfdev-5:
Closing in favor of https://github.com/pytorch/ignite/pull/3827

PR #2487, opened 2022-02-20
Fixed issue when loading a single non-nn.Module object

### LINKED ISSUE #2479: Checkpoint loading doesn't work if object doesn't contain a torch.nn.Module
## 🐛 Bug description

Running the following code:
```python
Checkpoint.load_objects(to_load=to_load, checkpoint=checkpoint)
```

results in the following error:
```
load_state_dict() got an unexpected keyword argument 'strict'
```

if ```to_load``` doesn't contain a ```torch.nn.Module```. 

For example, ```{"optimizer": optimizer}``` fails, but ```{"model": torch.nn.Linear(10, 10)}``` succeeds.

## Notebook to reproduce
https://colab.research.google.com/drive/1G-iVC-ixMQrw90AEp5mDuSjxUAPyO8YF?usp=sharing

## Environment
Run on google colab with:
torch version 1.10.0+cu111
ignite version 0.4.8

vfdev-5:
Fixed #2479

Description:
- Fixed issue when loading a single non-nn.Module object

Check list:


sdesrozis:
A few mypy issues but it seems good to me !

PR #2496, opened 2022-02-27
Feature/#2465 lr scheduler attach events

### LINKED ISSUE #2465: Make LRScheduler attachable to Events.ITERATION_STARTED
## 🚀 Feature

Currently, the correct way to use `LRScheduler` wrapper for pytorch >= 1.1.0 is the following:
```python
from torch.optim.lr_scheduler import StepLR

torch_lr_scheduler = StepLR(optimizer, step_size=3, gamma=0.1)
scheduler = LRScheduler(torch_lr_scheduler)

def print_lr():
    print(optimizer.param_groups[0]["lr"])

# In this example, we assume to have installed PyTorch>=1.1.0
# (with new `torch.optim.lr_scheduler` behaviour) and
# we attach scheduler to Events.ITERATION_COMPLETED
# instead of Events.ITERATION_STARTED to make sure to use
# the first lr value from the optimizer, otherwise it is will be skipped:
trainer.add_event_handler(Events.ITERATION_COMPLETED, scheduler)

trainer..run([0] * 8, max_epochs=1)
```
>
```
0.1
0.1
0.1
0.010
0.010
0.010
0.001
0.001
```

- https://github.com/pytorch/ignite/pull/2463

however, other schedulers should be used as following ([link](https://pytorch.org/ignite/generated/ignite.handlers.param_scheduler.PiecewiseLinear.html#ignite.handlers.param_scheduler.PiecewiseLinear))
```python
milestones_values = [(1, 1.0), (3, 0.8), (5, 0.2)]
scheduler = PiecewiseLinear(
    optimizer, "lr", milestones_values=milestones_values)

trainer.add_event_handler(Events.ITERATION_STARTED, scheduler)

def print_lr():
    print(optimizer.param_groups[0]["lr"])

trainer.run([0] * 6, max_epochs=1)
```

The idea is to improve `LRScheduler` such that we could attach it to `Events.ITERATION_STARTED` and have a coherent API. It will be a BC-breaking change, but for good.

So, desired example using `LRScheduler` should be:
```python
from torch.optim.lr_scheduler import StepLR

torch_lr_scheduler = StepLR(optimizer, step_size=3, gamma=0.1)
scheduler = LRScheduler(torch_lr_scheduler)

trainer.add_event_handler(Events.ITERATION_STARTED, scheduler)

def print_lr():
    print(optimizer.param_groups[0]["lr"])

trainer.run([0] * 8, max_epochs=1)
```
Currently, this gives a wrong behaviour as the first 0.1 wasn't consumed by the training step.
```
0.1
0.1
0.010
0.010
0.010
0.001
0.001
```
The idea could be to retain the first value and reapply it once and then keep everything as it is now.

yuta0821:
Fixes #2465

Description:
Based on discussion in the #2465, I updated the code so that the LRScheduler will attach to Events.STARTED.
Main Contribution
- ignite/handlers/param_scheduler.py
  - add `use_legacy`  on class LRScheduler and explanation of it
  - update example code
  - updata `__call__`  method
  - update `simulate_values`  such as change the timing of  `__call__`  method of scheduler
  - update create_lr_scheduler_with_warmup
- ignite/contrib/engines/common.py
  - remove some code to adjust to this change
- tests/ignite/handlers/test_param_scheduler.py
  - add some test to make sure that the `use_legacy`  works

Matters for consultation
Currently implementing the following part using `use_legacy = True`, but feel it is not a very smart implemantation.
```python
# ignite/contrib/engines/common.py 1022
lr_scheduler = LRScheduler(lr_scheduler, save_history=save_history, use_legacy=True)
```
However, in the current implementation, we need to set the initial learning rate at the end of each milestone, and the change for issue #2465 in implementation will probably result in one extra initial learning rate. I couldn't think of an easy way to fix this. I think if don't take `use_legacy=True`, probably may need to change the code in many places.

I ran the following test and pass them
- bash tests/run_cpu_tests.sh
- bash ./tests/run_code_style.sh mypy
- bash ./tests/run_code_style.sh fmt

Check list:

### DIFF
--- a/ignite/handlers/param_scheduler.py
+++ b/ignite/handlers/param_scheduler.py
@@ -861,6 +861,7 @@ class LRScheduler(ParamScheduler):
                 "the first lr value from the optimizer, otherwise it is will be skipped"
             )
             self.lr_scheduler.last_epoch += 1  # type: ignore[attr-defined]
+        self.keep_first_lr = keep_first_lr
         if keep_first_lr:
             self.lr_scheduler._get_lr_called_within_step = True  # type: ignore[attr-defined]
             self.first_lr = cast(List[float], self.lr_scheduler.get_lr())
@@ -874,9 +875,8 @@ class LRScheduler(ParamScheduler):
     def get_param(self) -> Union[float, List[float]]:
         """Method to get current optimizer's parameter value"""
         # Emulate context manager for pytorch>=1.4
-        if hasattr(self, "first_lr"):
+        if self.keep_first_lr and self.lr_scheduler.last_epoch == 0:
             lr_list = self.first_lr
-            del self.first_lr
         else:
             self.lr_scheduler._get_lr_called_within_step = True  # type: ignore[attr-defined]
             lr_list = cast(List[float], self.lr_scheduler.get_lr())
--- a/ignite/handlers/param_scheduler.py
+++ b/ignite/handlers/param_scheduler.py
@@ -795,6 +795,7 @@ class LRScheduler(ParamScheduler):
         lr_scheduler: lr_scheduler object to wrap.
         save_history: whether to log the parameter values to
             `engine.state.param_history`, (default=False).
+        use_legacy: whether for LRScheduler to behave previously, (default=False).
 
     Examples:
 
@@ -808,19 +809,15 @@ class LRScheduler(ParamScheduler):
             from torch.optim.lr_scheduler import StepLR
 
             torch_lr_scheduler = StepLR(default_optimizer, step_size=3, gamma=0.1)
-
             scheduler = LRScheduler(torch_lr_scheduler)
+
+            # In this example, we assume to have installed PyTorch>=1.1.0
+            # (with new `torch.optim.lr_scheduler` behaviour)
+            default_trainer.add_event_handler(Events.ITERATION_STARTED, scheduler)
 
             @default_trainer.on(Events.ITERATION_COMPLETED)
             def print_lr():
                 print(default_optimizer.param_groups[0]["lr"])
-
-            # In this example, we assume to have installed PyTorch>=1.1.0
-            # (with new `torch.optim.lr_scheduler` behaviour) and
-            # we attach scheduler to Events.ITERATION_COMPLETED
-            # instead of Events.ITERATION_STARTED to make sure to use
-            # the first lr value from the optimizer, otherwise it is will be skipped:
-            default_trainer.add_event_handler(Events.ITERATION_COMPLETED, scheduler)
 
             default_trainer.run([0] * 8, max_epochs=1)
 
@@ -838,7 +835,7 @@ class LRScheduler(ParamScheduler):
     .. versionadded:: 0.4.5
     """
 
-    def __init__(self, lr_scheduler: _LRScheduler, save_history: bool = False):
+    def __init__(self, lr_scheduler: _LRScheduler, save_history: bool = False, use_legacy: bool = False):
 
         if not isinstance(lr_scheduler, _LRScheduler):
             raise TypeError(
@@ -852,11 +849,18 @@ class LRScheduler(ParamScheduler):
             param_name="lr",
             save_history=save_history,
         )
+        if use_legacy:
+            warnings.warn(
+                "we attach scheduler to Events.ITERATION_COMPLETED"
+                "instead of Events.ITERATION_STARTED to make sure to use"
+                "the first lr value from the optimizer, otherwise it is will be skipped"
+            )
+            self.lr_scheduler.last_epoch += 1  # type: ignore[attr-defined]
         self._state_attrs += ["lr_scheduler"]
 
     def __call__(self, engine: Optional[Engine], name: Optional[str] = None) -> None:
-        self.lr_scheduler.last_epoch += 1  # type: ignore[attr-defined]
         super(LRScheduler, self).__call__(engine, name)
+        self.lr_scheduler.last_epoch += 1  # type: ignore[attr-defined]
 
     def get_param(self) -> Union[float, List[float]]:
         """Method to get current optimizer's parameter value"""
@@ -904,9 +908,9 @@ class LRScheduler(ParamScheduler):
             values = []
             scheduler = cls(save_history=False, lr_scheduler=lr_scheduler, **kwargs)
             for i in range(num_events):
+                scheduler(engine=None)
                 params = [p[scheduler.param_name] for p in scheduler.optimizer_param_groups]
                 values.append([i] + params)
-                scheduler(engine=None)
 
             obj = torch.load(cache_filepath.as_posix())
             lr_scheduler.load_state_dict(obj["lr_scheduler"])
@@ -1015,7 +1019,7 @@ def create_lr_scheduler_with_warmup(
             if init_lr != param_group_warmup_end_value:
                 milestones_values.append((warmup_duration, init_lr))
 
-            lr_scheduler = LRScheduler(lr_scheduler, save_history=save_history)
+            lr_scheduler = LRScheduler(lr_scheduler, save_history=save_history, use_legacy=True)
         else:
             init_lr = lr_scheduler.get_param()
             if init_lr == param_group_warmup_end_value:
--- a/ignite/handlers/param_scheduler.py
+++ b/ignite/handlers/param_scheduler.py
@@ -795,7 +795,7 @@ class LRScheduler(ParamScheduler):
         lr_scheduler: lr_scheduler object to wrap.
         save_history: whether to log the parameter values to
             `engine.state.param_history`, (default=False).
-        use_legacy: whether for LRScheduler to behave previously, (default=False).
+        use_legacy: if True, scheduler should be attached to ``Events.ITERATION_COMPLETED``, (default=False).
 
     Examples:
 
@@ -811,8 +811,6 @@ class LRScheduler(ParamScheduler):
             torch_lr_scheduler = StepLR(default_optimizer, step_size=3, gamma=0.1)
             scheduler = LRScheduler(torch_lr_scheduler)
 
-            # In this example, we assume to have installed PyTorch>=1.1.0
-            # (with new `torch.optim.lr_scheduler` behaviour)
             default_trainer.add_event_handler(Events.ITERATION_STARTED, scheduler)
 
             @default_trainer.on(Events.ITERATION_COMPLETED)
@@ -851,7 +849,7 @@ class LRScheduler(ParamScheduler):
         )
         if use_legacy:
             warnings.warn(
-                "we attach scheduler to Events.ITERATION_COMPLETED"
+                "Please make sure to attach scheduler to Events.ITERATION_COMPLETED"
                 "instead of Events.ITERATION_STARTED to make sure to use"
                 "the first lr value from the optimizer, otherwise it is will be skipped"
             )
--- a/ignite/handlers/param_scheduler.py
+++ b/ignite/handlers/param_scheduler.py
@@ -875,7 +875,7 @@ class LRScheduler(ParamScheduler):
     def get_param(self) -> Union[float, List[float]]:
         """Method to get current optimizer's parameter value"""
         # Emulate context manager for pytorch>=1.4
-        if self.keep_first_lr and self.lr_scheduler.last_epoch == 0:
+        if self.keep_first_lr and self.lr_scheduler.last_epoch == 0:  # type: ignore[attr-defined]
             lr_list = self.first_lr
         else:
             self.lr_scheduler._get_lr_called_within_step = True  # type: ignore[attr-defined]
--- a/ignite/handlers/param_scheduler.py
+++ b/ignite/handlers/param_scheduler.py
@@ -796,6 +796,7 @@ class LRScheduler(ParamScheduler):
         save_history: whether to log the parameter values to
             `engine.state.param_history`, (default=False).
         use_legacy: if True, scheduler should be attached to ``Events.ITERATION_COMPLETED``, (default=False).
+        keep_first_lr: whether to keep first lr. Only needed for internal use. Do not touch!
 
     Examples:
 
@@ -833,7 +834,13 @@ class LRScheduler(ParamScheduler):
     .. versionadded:: 0.4.5
     """
 
-    def __init__(self, lr_scheduler: _LRScheduler, save_history: bool = False, use_legacy: bool = False):
+    def __init__(
+        self,
+        lr_scheduler: _LRScheduler,
+        save_history: bool = False,
+        use_legacy: bool = False,
+        keep_first_lr: bool = False,
+    ):
 
         if not isinstance(lr_scheduler, _LRScheduler):
             raise TypeError(
@@ -854,6 +861,10 @@ class LRScheduler(ParamScheduler):
                 "the first lr value from the optimizer, otherwise it is will be skipped"
             )
             self.lr_scheduler.last_epoch += 1  # type: ignore[attr-defined]
+        if keep_first_lr:
+            self.lr_scheduler._get_lr_called_within_step = True  # type: ignore[attr-defined]
+            self.first_lr = self.lr_scheduler.get_lr()
+            self.lr_scheduler._get_lr_called_within_step = False  # type: ignore[attr-defined]
         self._state_attrs += ["lr_scheduler"]
 
     def __call__(self, engine: Optional[Engine], name: Optional[str] = None) -> None:
@@ -863,9 +874,14 @@ class LRScheduler(ParamScheduler):
     def get_param(self) -> Union[float, List[float]]:
         """Method to get current optimizer's parameter value"""
         # Emulate context manager for pytorch>=1.4
-        self.lr_scheduler._get_lr_called_within_step = True  # type: ignore[attr-defined]
-        lr_list = cast(List[float], self.lr_scheduler.get_lr())
-        self.lr_scheduler._get_lr_called_within_step = False  # type: ignore[attr-defined]
+        if hasattr(self, "first_lr"):
+            lr_list = self.first_lr
+            del self.first_lr
+        else:
+            self.lr_scheduler._get_lr_called_within_step = True  # type: ignore[attr-defined]
+            lr_list = cast(List[float], self.lr_scheduler.get_lr())
+            self.lr_scheduler._get_lr_called_within_step = False  # type: ignore[attr-defined]
+
         if len(lr_list) == 1:
             return lr_list[0]
         else:
@@ -1009,23 +1025,19 @@ def create_lr_scheduler_with_warmup(
         else:
             param_group_warmup_end_value = warmup_end_value
 
-        milestones_values = [(0, warmup_start_value), (warmup_duration - 1, param_group_warmup_end_value)]
-
         if isinstance(lr_scheduler, _LRScheduler):
             init_lr = param_group["lr"]
-
-            if init_lr != param_group_warmup_end_value:
-                milestones_values.append((warmup_duration, init_lr))
-
-            lr_scheduler = LRScheduler(lr_scheduler, save_history=save_history, use_legacy=True)
+            lr_scheduler = LRScheduler(lr_scheduler, save_history=save_history, keep_first_lr=True)
         else:
             init_lr = lr_scheduler.get_param()
-            if init_lr == param_group_warmup_end_value:
-                if warmup_duration > 2:
-                    d = (param_group_warmup_end_value - warmup_start_value) / (warmup_duration - 1)
-                    milestones_values[-1] = (warmup_duration - 2, param_group_warmup_end_value - d)
-                else:
-                    milestones_values.pop(-1)
+
+        milestones_values = [(0, warmup_start_value), (warmup_duration - 1, param_group_warmup_end_value)]
+        if init_lr == param_group_warmup_end_value:
+            if warmup_duration > 2:
+                d = (param_group_warmup_end_value - warmup_start_value) / (warmup_duration - 1)
+                milestones_values[-1] = (warmup_duration - 2, param_group_warmup_end_value - d)
+            else:
+                milestones_values.pop(-1)
 
         warmup_schedulers.append(
             PiecewiseLinear(
--- a/ignite/handlers/param_scheduler.py
+++ b/ignite/handlers/param_scheduler.py
@@ -854,8 +854,8 @@ class LRScheduler(ParamScheduler):
         )
         if use_legacy:
             warnings.warn(
-                "Please make sure to attach scheduler to Events.ITERATION_COMPLETED"
-                "instead of Events.ITERATION_STARTED to make sure to use"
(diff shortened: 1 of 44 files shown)


ignite/handlers/param_scheduler.py
```diff
@@ -1015,7 +1019,7 @@ def print_lr():
             if init_lr != param_group_warmup_end_value:
                 milestones_values.append((warmup_duration, init_lr))
 
-            lr_scheduler = LRScheduler(lr_scheduler, save_history=save_history)
+            lr_scheduler = LRScheduler(lr_scheduler, save_history=save_history, use_legacy=True)
```
vfdev-5, line 1022: Can we check if we could use `use_legacy=False` ? Or what is blocking here ?

vfdev-5 (commented):
Thanks a lot for the PR @yuta0821 !
I left few comments and will check in details the modifications in tests.

ignite/handlers/param_scheduler.py
yuta0821: if `use_legacy=False`,  the result look like the example below
- example1
```python
from torch.optim.lr_scheduler import ExponentialLR

tensor = torch.zeros([1], requires_grad=True)
optimizer = torch.optim.SGD([tensor], lr=0.01)
torch_lr_scheduler = StepLR(optimizer, step_size=3, gamma=0.1)

trainer = Engine(dummy_update)
scheduler = create_lr_scheduler_with_warmup(torch_lr_scheduler,
                                            warmup_start_value=0.0,
                                            warmup_end_value=0.1,
                                            warmup_duration=3)

trainer.add_event_handler(Events.ITERATION_STARTED, scheduler)

def print_lr():
    print(optimizer.param_groups[0]["lr"])

_ = trainer.run([0] * 8, max_epochs=1)
```
- output1
```
0.0
0.05
0.1
0.01
0.01
0.01
0.01
0.001
```
- example2
```python
from torch.optim.lr_scheduler import ExponentialLR

tensor = torch.zeros([1], requires_grad=True)
optimizer = torch.optim.SGD([tensor], lr=0.1)
torch_lr_scheduler = StepLR(optimizer, step_size=3, gamma=0.1)

trainer = Engine(dummy_update)
scheduler = create_lr_scheduler_with_warmup(torch_lr_scheduler,
                                            warmup_start_value=0.0,
                                            warmup_end_value=0.1,
                                            warmup_duration=3)

trainer.add_event_handler(Events.ITERATION_STARTED, scheduler)

def print_lr():
    print(optimizer.param_groups[0]["lr"])

_ = trainer.run([0] * 8, max_epochs=1)
```
- output2
```
0.0
0.05
0.1
0.1
0.1
0.1
0.010000000000000002
0.010000000000000002
```
The initial value of stepLR (= 0.1 or 0.01) is counted one extra time than step_size (=3).

vfdev-5: Thanks for the details @yuta0821 , can you please debug this a bit more and explicitly say which scheduler is responsible for adding LR value (0.1 or 0.01). I'm not quite sure to understand why exactly this happens. Thanks !

vfdev-5: @yuta0821 sorry for delay and thanks for the explanation! I haven't checked it in details but will do as soon as it could be possible from my side (~4-5 days).
@sdesrozis can you help with that if you have some bandwidth ?

yuta0821: @vfdev-5 
Thank you for your reply, there is no problem.

sdesrozis: @vfdev-5 Ok I will check

sdesrozis: @yuta0821 The point is `use_legacy=True` mustn’t be used internally because we will depreciate in the futur. All the things should work correctly using the new behaviour.

Anyway, have you checked the ConcatScheduler ? It seems the concatenation should not work as excepted without `use_legacy=True`, right ?

yuta0821: > @yuta0821 The point is use_legacy=True mustn’t be used internally because we will depreciate in the futur. All the things should work correctly using the new behaviour.

Ok, I understand.

> Anyway, have you checked the ConcatScheduler ? 

Yes, I have checked the `ConcatScheduler`.  
> It seems the concatenation should not work as excepted without use_legacy=True, right ?

The 'excepted' is 'expected' ? If `use_legacy` is True, the initial `lr` value of optimizer is repeated extra once.  
I think the concatenation between `LRScheduler` and `PiecewiseLinear` in `ConcatScheduler` causes unexpected behavior.

yuta0821: It performs the same operation as `use_legacy`, but wouldn't it be preferable to add the argument `skip_initial_value` as a variable to be used for the internal function ?

sdesrozis: Yeah, expected rather than excepted 😅

As far I understand, using `use_legacy=False`, the first `lr` comes from the optimizer. Whatever schedulers used, the schedulers concatenation will produce a repetition at each joint. 

Having an internal option as you suggested sounds good to me. I mean rename `use_legacy` to `skip_initial_value` is fine. Although, we have to keep `use_legacy` for the users.

What do you think ?

yuta0821: @sdesrozis 
I am sorry for having consulted with you.
It seems that this problem can be solved by setting the internal variable `keep_first_value=True` only when `create_lr_scheduler_with_warmup` is called, as shown below, to store the initial value in `LRScheduler`.
```python
class LRScheduler(ParamScheduler):
    def __init__(self, lr_scheduler: _LRScheduler, save_history: bool = False, use_legacy: bool = False, keep_first_lr: bool = False):
        if keep_first_lr:
            self.lr_scheduler._get_lr_called_within_step = True  # type: ignore[attr-defined]
            self.first_lr = self.lr_scheduler.get_lr()
            self.lr_scheduler._get_lr_called_within_step = False  # type: ignore[attr-defined]

    def get_param(self) -> Union[float, List[float]]:
        """Method to get current optimizer's parameter value"""
        # Emulate context manager for pytorch>=1.4
        if hasattr(self, "first_lr"):
            lr_list = self.first_lr
            del self.first_lr
        else:
       
def create_lr_scheduler_with_warmup( ):
    if isinstance(lr_scheduler, _LRScheduler):
        init_lr = param_group["lr"]
        lr_scheduler = LRScheduler(lr_scheduler, save_history=save_history, keep_first_lr=True)
    else:
        init_lr = lr_scheduler.get_param()
```
I am running the existing test now. I will commit once all tests pass!

sdesrozis: It seems that the tests are ko...

yuta0821: Sorry, I added `type: ignore[attr-defined]` !

sdesrozis:
@yuta0821 Thanks very much for this work. I have to think about the api though. The `use_legacy` arg is for BC (so to be removed one day) but the internal only `keep_fist_lr` makes me a little doubtful. This arg is only required using ConcatScheduler, could we do smarter ? Anyway, it works and that is pretty cool. Let's just think about it. I suppose that @vfdev-5 would have an opinion on that.

yuta0821:
@sdesrozis Yes, I agree, I don't want to add more args without any thought. However, I haven't yet figured out how to control the `lr` value of an `LRScheduler` bound to a `ConcatScheduler` without using `keep_first_lr`. Originally, this was implemented by adjusting the `milestones_values` of `PeacewiseLinear`,  but chage of this PR no longer works. I'll see if there's a smarter way to do this.

sdesrozis:
My toughts are the target would be having the following code working fine
```python
from torch.optim.lr_scheduler import StepLR

scheduler1 = StepLR(optimizer, step_size=3, gamma=0.1)
scheduler1 = LRScheduler(scheduler1)

scheduler2 = StepLR(optimizer, step_size=3, gamma=0.01)
scheduler2 = LRScheduler(scheduler2)

scheduler = ConcatScheduler(schedulers=[scheduler1, scheduler2], durations=[4, ])
```

Whether `use_legacy` is used or not for `scheduler2`, it should produce the same output. It means `ConcatScheduler` should enable `keep_first_lr` of `LRScheduler` internally without any user api. The minimal would be to enable using a method rather than at init.
```python
scheduler2 = LRScheduler(scheduler2)
scheduler2.keep_first_lr = True
```
I don't know yet...

yuta0821:
@sdesrozis I'd like to confirm the behavior of following code using `use_legacy=True`. Is this output is right ?
```python
from torch.optim.lr_scheduler import StepLR

optimizer = torch.optim.SGD([tensor], lr=0.01)
torch_scheduler1 = StepLR(optimizer, step_size=3, gamma=0.1)
scheduler1 = LRScheduler(torch_scheduler1, use_legacy=True)

torch_scheduler2 = StepLR(optimizer, step_size=3, gamma=0.01)
scheduler2 = LRScheduler(torch_scheduler2, use_legacy=True)

scheduler = ConcatScheduler(schedulers=[scheduler1, scheduler2], durations=[4, ])

def dummy_update(engine, batch):
    pass
trainer = Engine(dummy_update)

def print_lr():
    print(optimizer.param_groups[0]["lr"])
trainer.add_event_handler(Events.ITERATION_COMPLETED, scheduler)

_ = trainer.run([0] * 9, max_epochs=1)
```
```
0.01
0.01
0.01
0.001
0.001
0.001
0.001
1e-05
1e-05
```
> Whether `use_legacy` is used or not for schedulers, it should produce the same output. It means `ConcatScheduler` should enable `keep_first_lr` of `LRScheduler` internally without any user api.

OK, I understand what message means. Thank you for the clear explanation. Surely, we must enable `keep_first_lr` of `LRScheduler` internally without any user api.

ignite/handlers/param_scheduler.py
```diff
@@ -1008,16 +1017,17 @@ def print_lr():
...
 
         milestones_values = [(0, warmup_start_value), (warmup_duration - 1, param_group_warmup_end_value)]
-
         if isinstance(lr_scheduler, _LRScheduler):
             init_lr = param_group["lr"]
 
             if init_lr != param_group_warmup_end_value:
                 milestones_values.append((warmup_duration, init_lr))
 
+            lr_scheduler.last_epoch = 1
```
yuta0821, line 1026: I set `lr_scheduelr.last_epoch` to one here so that the initial `lr` value `of optimizer` avoids to overlap with `milestones_values`.

vfdev-5: @yuta0821 I'd avoid hard-coding values like that, can we make it relative somehow ? Why it should be explicitly as 1 and not 2 or 3 ?

ignite/handlers/param_scheduler.py
```diff
@@ -1008,16 +1017,17 @@ def print_lr():
             param_group_warmup_end_value = warmup_end_value
 
         milestones_values = [(0, warmup_start_value), (warmup_duration - 1, param_group_warmup_end_value)]
-
         if isinstance(lr_scheduler, _LRScheduler):
             init_lr = param_group["lr"]
 
             if init_lr != param_group_warmup_end_value:
```
vfdev-5, line 1023: @yuta0821 I wonder if this part of code shouldn't be changed to avoid extra step with initial optimizer's lr ?

Following your examples (example1 and example2): https://github.com/pytorch/ignite/pull/2496#discussion_r815573334

Example 1: 
```
optimizer = torch.optim.SGD([tensor], lr=0.01)
torch_lr_scheduler = StepLR(optimizer, step_size=3, gamma=0.1)
scheduler = create_lr_scheduler_with_warmup(torch_lr_scheduler,
                                            warmup_start_value=0.0,
                                            warmup_end_value=0.1,
                                            warmup_duration=3)
```

What would be an expected output regardless the implementation ? IMO, something like:
```
0.0     # iter0, warmup
0.05    # iter1, warmup
0.1     # iter2, warmup
0.01    # iter3, torch scheduler
0.01    # iter4, torch scheduler
0.01    # iter4, torch scheduler
0.001
```

Example 2: 
```
optimizer = torch.optim.SGD([tensor], lr=0.1)
torch_lr_scheduler = StepLR(optimizer, step_size=3, gamma=0.1)
scheduler = create_lr_scheduler_with_warmup(torch_lr_scheduler,
                                            warmup_start_value=0.0,
                                            warmup_end_value=0.1,
                                            warmup_duration=3)
```

What would be an expected output regardless the implementation ? IMO, something like:
```
0.0     # iter0, warmup
0.05    # iter1, warmup
0.1     # iter2, warmup
0.01    # iter3, torch scheduler
0.01    # iter4, torch scheduler
0.01    # iter4, torch scheduler
0.001
```

vfdev-5: Sorry, I wanted to say "this part of code SHOULD be changed" and not "shouldn't".

vfdev-5: Yes, and is this overlap due to : 
```python
if init_lr != param_group_warmup_end_value:
    milestones_values.append((warmup_duration, init_lr))
```

yuta0821: Yes, it is.
However, if this part of code is removed, the following easy test doesn't work well.

```python
optimizer = torch.optim.SGD([tensor], lr=0.01)
torch_lr_scheduler = StepLR(optimizer, step_size=3, gamma=0.1)
scheduler = create_lr_scheduler_with_warmup(torch_lr_scheduler,
                                            warmup_start_value=0.0,
                                            warmup_end_value=0.1,
                                            warmup_duration=3)

```

```
0.0      # iter0, warmup
0.05     # iter1, warmup
0.1      # iter2, warmup
0.1      # iter3, torch scheduler <- this result is not good
0.1      # iter4, torch scheduler
0.1      # iter4, torch scheduler
0.01     # iter5, torch scheduler
```

vfdev-5: @yuta0821 thanks for checking that ! OK, let's keep this part of code.

ignite/handlers/param_scheduler.py
vfdev-5: @yuta0821 how about here, if we were doing `lr_scheduler.last_epoch += 1`, instead of `lr_scheduler.last_epoch = 1` ?

yuta0821: > @yuta0821 I'd avoid hard-coding values like that, can we make it relative somehow ? Why it should be explicitly as 1 and not 2 or 3 ?

I chose 1 to skip the extra initial value which is equal the value of `last_epoch = 0`.

> yuta0821 how about here, if we were doing lr_scheduler.last_epoch += 1, instead of lr_scheduler.last_epoch = 1 ?

When the `create_lr_scheduler_with_warmup` is constructed, `lr_scheduler.last_epoch = 0` is assumed, so `lr_scheduler.last_epoch = 1` is same as `lr_scheduler.last_epoch += 1`.

vfdev-5: `lr_scheduler` is an input argument where user can manually set `lr_scheduler.last_epoch` to something different to 0. If we set it internally to 1 this can be a "surprising" behaviour. I do not know why user would want to set it to something (maybe while restoring from a state), but setting it to 1 can't be that innocent, I think.

yuta0821: Adding `lr_scheduler.last_epoch += 1`  to skip the initial `lr` value can't be that innocent, either ?
I think this part of code plays the same role as following code in `LRScheduler.__init__`.
```python
if use_legacy:
    warnings.warn(
        "Please make sure to attach scheduler to Events.ITERATION_COMPLETED"
        "instead of Events.ITERATION_STARTED to make sure to use"
        "the first lr value from the optimizer, otherwise it is will be skipped"
     )
    self.lr_scheduler.last_epoch += 1  # type: ignore[attr-defined]
```

vfdev-5: That's true, if we do `lr_scheduler.last_epoch += 1` or `lr_scheduler.last_epoch = 1` it would be equivalent to `use_legacy=True` once `lr_scheduler` is wrapped with `LRScheduler`.

vfdev-5:
> @sdesrozis I'd like to confirm the behavior of following code using `use_legacy=True`. Is this output is right ?
> 
> ```python
> from torch.optim.lr_scheduler import StepLR
> 
> optimizer = torch.optim.SGD([tensor], lr=0.01)
> torch_scheduler1 = StepLR(optimizer, step_size=3, gamma=0.1)
> scheduler1 = LRScheduler(torch_scheduler1, use_legacy=True)
> 
> torch_scheduler2 = StepLR(optimizer, step_size=3, gamma=0.01)
> scheduler2 = LRScheduler(torch_scheduler2, use_legacy=True)
> 
> scheduler = ConcatScheduler(schedulers=[scheduler1, scheduler2], durations=[4, ])
> 
> def dummy_update(engine, batch):
>     pass
> trainer = Engine(dummy_update)
> 
> @trainer.on(Events.ITERATION_COMPLETED)
> def print_lr():
>     print(optimizer.param_groups[0]["lr"])
> trainer.add_event_handler(Events.ITERATION_COMPLETED, scheduler)
> 
> _ = trainer.run([0] * 9, max_epochs=1)
> ```
> 
> ```
> 0.01
> 0.01
> 0.01
> 0.001
> 0.001
> 0.001
> 0.001
> 1e-05
> 1e-05
> ```

@yuta0821 the output looks correct to me. We have 
```
--- scheduler 1 
0.01
0.01
0.01
0.001
--- scheduler 2
0.001
0.001
0.001
1e-05
1e-05
```

What happens if using `use_legacy=False` and attaching to appropriate event ?

yuta0821:
@vfdev-5 
Sorry for late reply.
> What happens if using use_legacy=False and attaching to appropriate event ?

If set `trainer.add_event_handler` to the proper position, the behavior with `use_legacy = False` becomes as follows.
With `keep_lr_first`, the behavior of this code is the same as with `use_legacy = True`.

```python
from torch.optim.lr_scheduler import StepLR

optimizer = torch.optim.SGD([tensor], lr=0.01)
torch_scheduler1 = StepLR(optimizer, step_size=3, gamma=0.1)
scheduler1 = LRScheduler(torch_scheduler1, use_legacy=False)

torch_scheduler2 = StepLR(optimizer, step_size=3, gamma=0.01)
scheduler2 = LRScheduler(torch_scheduler2, use_legacy=False)

scheduler = ConcatScheduler(schedulers=[scheduler1, scheduler2], durations=[4, ])

def dummy_update(engine, batch):
    pass
trainer = Engine(dummy_update)
trainer.add_event_handler(Events.ITERATION_STARTED, scheduler)

def print_lr():
    print(optimizer.param_groups[0]["lr"])
    
_ = trainer.run([0] * 9, max_epochs=1)
```
```
--- scheduler 1 
0.01
0.01
0.01
0.001
--- scheduler 2 
0.001
0.001
0.001
1e-05
1e-05
```

vfdev-5:
@yuta0821 sorry for delay on this PR, i'm still missing a complete understanding of the problem here. I'll try to figure out what happens exactly on my side and try to suggest a way to handle that. Adding an argument for internal usage is not a good design.

yuta0821:
@vfdev-5 

> @yuta0821 sorry for delay on this PR, i'm still missing a complete understanding of the problem here. I'll try to figure out what happens exactly on my side and try to suggest a way to handle that. Adding an argument for internal usage is not a good design.
> Thanks for your patience and working on the PR !

No, sir, not at all. Thanks a lot for your help. If I can contribute in any way, I'll be happy to do so.

ignite/handlers/param_scheduler.py
```diff
@@ -1008,22 +1027,18 @@ def print_lr():
...
-                    milestones_values[-1] = (warmup_duration - 2, param_group_warmup_end_value - d)
-                else:
-                    milestones_values.pop(-1)
+
+        if init_lr == param_group_warmup_end_value:
+            if warmup_duration > 2:
+                d = (param_group_warmup_end_value - warmup_start_value) / (warmup_duration - 1)
+                milestones_values[-1] = (warmup_duration - 2, param_group_warmup_end_value - d)
+            else:
+                milestones_values.pop(-1)
```
vfdev-5, line 1041: @yuta0821 can you explain why you did this modification ?

vfdev-5:
@yuta0821 i understand the problem and why you need to introduce `keep_first_lr` argument.
In two words for others, the problem with `create_lr_scheduler_with_warmup` is that either 1) we have an additional transitional lr value between `PiecewiseLinear` and `LRScheduler`:
```- PiecewiseLinear updates LR
1 0.0
- PiecewiseLinear updates LR
2 0.03
- PiecewiseLinear updates LR
3 0.06
- PiecewiseLinear updates LR
4 0.09
- PiecewiseLinear updates LR
5 0.12
- PiecewiseLinear updates LR  <----- additional step to remove
6 0.01
- LRScheduler updates LR 
7 0.01
- LRScheduler updates LR
8 0.01
- LRScheduler updates LR
9 0.01
- LRScheduler updates LR
10 0.005
```
or 2) `LRScheduler` modifies LR in a relative way => if `PiecewiseLinear` stops one step before, then `LRScheduler` will work from this value, e.g `0.12` from example above. This also gives a wrong behavior.

I see here two ways to move forward:
- either we add a more meaningful public arg to `LRScheduler`, e.g. `relative: bool = True` and thus we setup `create_lr_scheduler_with_warmup` with `LRScheduler(..., relative=False)`
- do something like `lr_scheduler.lr_scheduler.last_epoch += 1` inside `create_lr_scheduler_with_warmup`:
```python
lr_scheduler = LRScheduler(lr_scheduler, save_history=save_history)
lr_scheduler.lr_scheduler.last_epoch += 1
```

yuta0821:
> @yuta0821 i understand the problem and why you need to introduce keep_first_lr argument.

Thank you for your analysis. Your summary is exactly what I was thinking.

vfdev-5:
@yuta0821 I'm trying to figure out how to merge your work to master step by step.
You updated tests/ignite/handlers/test_param_scheduler.py , do you think we could make merge without updates introduced in ignite/handlers/param_scheduler.py ?
If yes, could you please send a PR with only updated tests ? Thanks a lot for your patience

yuta0821:
I think we could make merge neither `tests/ignite/handlers/test_param_scheduler.py` nor `tests/ignite/contrib/engines/test_common.py`  without updates introduced `ignite/handlers/param_scheduler.py`.
This is because the changes in tests such like `trainer.add_event_handler(Events.ITERATION_STARTED, scheduler)` are added to tackle on changes caused by `ignite/handlers/param_scheduler.py`.

vfdev-5:
@yuta0821 I pushed few commits to make this PR merged. Basically, it is one of your suggestions. Currently, I do not have a better way to fix it. Let's keep it as it is.
Thanks again for starting this work and sorry for delay it took to make it landed

vfdev-5 (approved):

yuta0821:
Thank you for your review.
> I pushed few commits to make this PR merged. Basically, it is one of your suggestions. Currently, I do not have a better way to fix it. Let's keep it as it is.

I understand. If I come up with a better way to fix it, I will create a new issue.

PR #2498, opened 2022-03-03
Add filename components in Checkpoint

### LINKED ISSUE #2480: Pass filename components into Checkpoint.load_objects
## 🚀 Feature

Currently, objects can be loaded like this:

```python

Checkpoint.load_objects(to_load=to_load, checkpoint=checkpoint_fp)
```

It would be nice to have the following as an option:
```python

checkpointer = Checkpoint(...)
checkpointer .load_objects(to_load=to_load, filename_components={"name": name, "global_step":global_step})
```

In other words, pass in components of ```self.filename_pattern```. The Checkpoint object already has ```self.filename_pattern```, ```self.filename_prefix```, ```self.ext```, and it should also have the ```dirname``` that is passed into ```__init__```. So most of the components of constructing the filename are already defined in the Checkpoint object. Getting the user to recreate the full file path is unnecessary.

sdesrozis:
Fixes #2480 

Description:

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -327,7 +327,7 @@ class Checkpoint(Serializable):
         self.include_self = include_self
         self.greater_or_equal = greater_or_equal
 
-    def __get_filename_pattern(self, global_step: Optional[int]) -> str:
+    def _get_filename_pattern(self, global_step: Optional[int]) -> str:
         if self.filename_pattern is None:
             filename_pattern = self.setup_filename_pattern(
                 with_prefix=len(self.filename_prefix) > 0,
@@ -414,7 +414,7 @@ class Checkpoint(Serializable):
                     name = k
                 checkpoint = checkpoint[name]
 
-            filename_pattern = self.__get_filename_pattern(global_step)
+            filename_pattern = self._get_filename_pattern(global_step)
 
             filename_dict = {
                 "filename_prefix": self.filename_prefix,
@@ -662,7 +662,7 @@ class Checkpoint(Serializable):
 
         global_step = filename_components.get("global_step", None)
 
-        filename_pattern = self.__get_filename_pattern(global_step)
+        filename_pattern = self._get_filename_pattern(global_step)
 
         checkpoint = self._setup_checkpoint()
         name = "checkpoint"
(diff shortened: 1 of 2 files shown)


ignite/handlers/checkpoint.py
```diff
@@ -327,6 +327,18 @@ def __init__(
         self.include_self = include_self
         self.greater_or_equal = greater_or_equal
 
+    def __get_filename_pattern(self, global_step: Optional[int]) -> str:
```
vfdev-5, line 330: @sdesrozis why using dunder `__` instead of single underscore `_` ,  `_get_filename_pattern` ?

vfdev-5 (commented):
Looks good to me, just a nit about refactored function name

ignite/handlers/checkpoint.py
sdesrozis: For privacy but one underscore would be enough.

ignite/handlers/checkpoint.py
```diff
@@ -519,41 +523,51 @@ def _check_objects(objs: Mapping, attr: str) -> None:
...
+
+                    to_save = {"weights": model, "optimizer": optimizer}
+
+                    trainer.add_event_handler(Events.EPOCH_COMPLETED(every=2), handler, to_save)
+                    trainer.run(torch.randn(10, 1), 5)
+
+                    to_load = to_save
+                    checkpoint_fp = Path(tmpdirname) / 'myprefix_checkpoint_40.pt'
+                    checkpoint = torch.load(checkpoint_fp)
+                    Checkpoint.load_objects(to_load=to_load, checkpoint=checkpoint)
```
vfdev-5, line 564: @sdesrozis this example is not working: 
```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [2], in <cell line: 10>()
     21 to_load = to_save
     22 # load checkpoint myprefix_checkpoint_40.pt
---> 23 checkpoint.load_objects(to_load=to_load, global_step=40)

TypeError: load_objects() missing 1 required positional argument: 'checkpoint'
```

ignite/handlers/checkpoint.py
```diff
@@ -519,41 +523,51 @@ def _check_objects(objs: Mapping, attr: str) -> None:
...
         Examples:
             .. code-block:: python
 
+                import tempfile
+                from pathlib import Path
+
                 import torch
+
                 from ignite.engine import Engine, Events
                 from ignite.handlers import ModelCheckpoint, Checkpoint
```
vfdev-5, line 546: We can also remove unused Checkpoint

PR #2520, opened 2022-03-21
add image detection example

louis-she:
An image detection example using PyTorch Faster RCNN with Aim. Some screenshots:

image

image

image

### DIFF
--- a/ignite/distributed/comp_models/native.py
+++ b/ignite/distributed/comp_models/native.py
@@ -425,7 +425,10 @@ if has_native_dist_support:
             if group is not None and not isinstance(group, dist.ProcessGroup):
                 raise ValueError("Argument group should be list of int or ProcessGroup")
             reduce_op = self._reduce_op_map[op]
-            dist.all_reduce(tensor, reduce_op, group=group)
+            if group is not None:
+                dist.all_reduce(tensor, reduce_op, group=group)
+            else:
+                dist.all_reduce(tensor, reduce_op)
             return tensor
 
         def _do_all_gather(self, tensor: torch.Tensor) -> torch.Tensor:


examples/contrib/detection/README.md
```diff
@@ -0,0 +1,30 @@
+# FasterRCNN Example with Ignite
+
+In this example, we show how to use _Ignite_ to train a image detection model with PyTorch built-in Faster RCNN implementation.
+
+- Using VOC2012 datasets (~2GB).
+- Vlidate with COCO mAP.
```
vfdev-5, line 6:
```suggestion
- Validate with COCO mAP.
```

examples/contrib/detection/README.md
```diff
@@ -0,0 +1,30 @@
...
+- Visualization.
+- Aim as experiment manager.
+
+## Requirements:
+
+- pytorch-ignite: `pip install pytorch-ignite`
+- [torchvision](https://github.com/pytorch/vision): `pip install torchvision`
+- [aim](https://github.com/aimhubio/aim): `pip install aim`
```
vfdev-5, line 14: I think Albumentations and pycocotools deps are missing here and in requirements.txt

examples/contrib/detection/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Alternatively, install the all requirements using `pip install -r requirements.txt`.
+
+## Usage:
+
+```bash
+python faster_rcnn.py
```
vfdev-5, line 21: maybe, we can rename it as `main.py` and support more torchvision's det models (same as for classification) ?

examples/contrib/detection/faster_rcnn.py
```diff
@@ -0,0 +1,286 @@
...
+    def update_model(engine, batch):
+        model.train()
+        images, targets = batch
+        images = list(image.to(device) for image in images)
+        targets = [{k: v.to(device) for k, v in t.items() if isinstance(v, torch.Tensor)} for t in targets]
+
+        if device == "cuda":
+            with torch.cuda.amp.autocast(enabled=True):
+                loss_dict = model(images, targets)
+                loss = sum(loss for loss in loss_dict.values())
```
vfdev-5, line 145: I think pytorch also support cpu autocast: https://pytorch.org/docs/stable/amp.html?highlight=autocast#torch.autocast
so we can unify the code...

examples/contrib/detection/faster_rcnn.py
```diff
@@ -0,0 +1,286 @@
...
+                pred_labels = [Dataset.class2name[l.item()] for l in pred["labels"]]
+                pred_boxes = pred["boxes"].long()
+                image = draw_bounding_boxes(image, pred_boxes, pred_labels, colors="red")
+
+                target_labels = [Dataset.class2name[l.item()] for l in target["labels"]]
+                target_boxes = target["boxes"].long()
+                image = draw_bounding_boxes(image, target_boxes, target_labels, colors="green")
+
+                aim_images.append(aim.Image(image.numpy().transpose((1, 2, 0))))
+        aim_logger.experiment.track(aim_images, name="vis", step=trainer.state.epoch)
```
vfdev-5, line 216: I'm not quite sure to understand why we need to accumulate and submit on iterations...
Can't we just submit on iterations without accumulating ?

examples/contrib/detection/faster_rcnn.py
```diff
@@ -0,0 +1,286 @@
...
+        tag="val",
+        metric_names=["mAP"],
+        global_step_transform=global_step_from_engine(trainer, Events.ITERATION_COMPLETED),
+    )
+
+    trainer.run(train_dataloader, max_epochs=epochs)
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser()
```
vfdev-5, line 257: Just a question if it won't be more easy to use [python-fire](https://github.com/google/python-fire) instead of arg parser ?

examples/contrib/detection/mean_ap.py
```diff
@@ -0,0 +1,252 @@
+"""
```
vfdev-5, line 1: Just a comment for ourselves that we have to advance on implementing mAP from our side

examples/contrib/detection/mean_ap.py
```diff
@@ -0,0 +1,252 @@
...
+            coco_eval.cocoDt = coco_dt
+            coco_eval.params.imgIds = list(img_ids)
+            img_ids, eval_imgs = evaluate(coco_eval)
+
+            self.eval_imgs[iou_type].append(eval_imgs)
+
+    def synchronize_between_processes(self):
+        for iou_type in self.iou_types:
+            self.eval_imgs[iou_type] = np.concatenate(self.eval_imgs[iou_type], 2)
+            create_common_coco_eval(self.coco_eval[iou_type], self.img_ids, self.eval_imgs[iou_type])
```
vfdev-5, line 107: This method looks a bit weird as according to its name it should sync between procs...

vfdev-5 (commented):
@louis-she thanks a lot for the PR !
This is a great start for finally adding detection example to ignite !
I left some comments on how to improve some parts of it...

examples/contrib/detection/mean_ap.py
```diff
@@ -0,0 +1,252 @@
...
+        img_dict["height"] = img.shape[-2]
+        img_dict["width"] = img.shape[-1]
+        dataset["images"].append(img_dict)
+        bboxes = targets["boxes"].clone()
+        bboxes[:, 2:] -= bboxes[:, :2]
+        bboxes = bboxes.tolist()
+        labels = targets["labels"].tolist()
+        areas = targets["area"].tolist()
+        iscrowd = targets["iscrowd"].tolist()
+        if "masks" in targets:
```
sdesrozis, line 41: Here, the implementation could be simplified since no masks neither key points are used.

examples/contrib/detection/faster_rcnn.py
```diff
@@ -0,0 +1,286 @@
...
+    parser.add_argument("--batch-size", type=int, default=4, help="batch size to use")
+    parser.add_argument("--lr", type=float, default=0.01, help="initial lr to use")
+    parser.add_argument(
+        "--download", action="store_true", default=False, help="either to download dataset automatically"
+    )
+    parser.add_argument("--device", type=str, default="cuda", help="gpu or cpu")
+    parser.add_argument("--image-size", type=int, default=512, help="image size to train")
+    parser.add_argument("--experiment-name", type=str, default="test", help="name of one experiment")
+    args = parser.parse_args()
+    run(
```
sdesrozis, line 276: idist could be used for ddp

examples/contrib/detection/faster_rcnn.py
```diff
@@ -0,0 +1,286 @@
...
+        "cow",
+        "dog",
+        "horse",
+        "sheep",
+        "person",
+    )
+    name2class = {v: k + 1 for k, v in enumerate(classes)}
+    class2name = {k + 1: v for k, v in enumerate(classes)}
+
+    def __getitem__(self, index: int) -> Tuple[Any, Any]:
```
sdesrozis, line 63: It seems that what is done here could be implemented as a transformation. It should be moved in another file IMO.

sayantan1410:
Also, the tests are failing because of code-formatting error, try running
`bash ./tests/run_code_style.sh install`
`bash ./tests/run_code_style.sh fmt`

examples/contrib/detection/faster_rcnn.py
louis-she: Can submit on iterations only if we set `step` as `trainer.state.iteration`, or `aim` will override the previous images.

Gather visualization images in epoch can give a good view when using aim I think, but indeed it will eat more memory.

examples/contrib/detection/faster_rcnn.py
louis-she: Do you mean the name and class number conversion? If yes I think it's ok to place here since there are nothing to do with the images.

examples/contrib/detection/faster_rcnn.py
louis-she: I don't have 2 GPUs to test distribution now. I'll try to write the distribution code and, maybe test on AWS?

examples/contrib/detection/main.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def update_model(engine, batch):
+        model.train()
+        images, targets = batch
+        images = list(image.to(device) for image in images)
+        targets = [{k: v.to(device) for k, v in t.items() if isinstance(v, torch.Tensor)} for t in targets]
+
+        with torch.cuda.amp.autocast(enabled=True):
+            loss_dict = model(images, targets)
+            loss = sum(loss for loss in loss_dict.values())
```
vfdev-5, line 168: Still `cuda.amp.autocast`, can it work on CPU ?

examples/contrib/detection/main.py
```diff
@@ -0,0 +1,329 @@
...
+    @visualizer.on(Events.EPOCH_STARTED)
+    def reset_vis_images(engine):
+        engine.state.model_outputs = []
+
+    @visualizer.on(Events.ITERATION_COMPLETED)
+    def add_vis_images(engine):
+        engine.state.model_outputs.append(engine.state.output)
+
+    @visualizer.on(Events.ITERATION_COMPLETED)
+    def submit_vis_images(engine):
```
sdesrozis, line 208: Does this ok in distributed ? I would have added a `@one_rank_only()` decorator.

louis-she: I havn't tested it on multi gpu core yet, there is a `@one_rank_only()` protection on the `trainer`'s `epoch_end` so the visualizer engine will not be called on other ranks.

louis-she:
@vfdev-5 @sdesrozis 
added retinanet support and distributed training, have tested on 2 GPUs.

louis-she:
One of the ci is failed cause `codecov `-4.27%`. I think there are no need to write test code for example right? Should we add `example` folder in the `ignore` in `codecov.yml` ?

louis-she:
I see that there is a PR of `mAP` implementation here https://github.com/pytorch/ignite/pull/2130, but there was no update for like half a year.

Can I open a new PR to implemente mAP?

examples/contrib/detection/main.py
```diff
@@ -0,0 +1,352 @@
...
+            loss_dict = model(images, targets)
+            loss = sum(loss for loss in loss_dict.values())
+
+        optimizer.zero_grad()
+        scaler.scale(loss).backward()
+        scaler.step(optimizer)
+        scaler.update()
+
+        loss_items = {k: v.item() for k, v in loss_dict.items()}
+        loss_items["loss_average"] = loss.item() / 4
```
vfdev-5, line 164:
```suggestion
        loss_items["loss_average"] = loss.item() / len(loss_dict)
```
?

PR #2523, opened 2022-03-22
Feature fixed weights hist handler issue 2328

### LINKED ISSUE #2328: WeightsHistHandler should plot all weights (incl those without grad)

sadra-barikbin:
Fixes #2328 

Description: 

- Added `whitelist` arg to `WeightsHistHandler` to be able to log specific weights to `TensorboardLogger`.

Check list:


vfdev-5:
@sadra-barikbin thanks for the PR, the issue was to improve current `WeightsHistHandler` without introducing another class.

sadra-barikbin:
Oh. So I add a `whitelist` parameter to `WeightsHistHandler`. If it was `None`, I log all of weights, otherwise just specified weights. Is this OK?

vfdev-5:
@sadra-barikbin please see this comment about whitelist option: https://github.com/pytorch/ignite/issues/2328#issuecomment-970316886
We want non-exact module name matching but partial.

sadra-barikbin:
In this PR one could give submodules themselves as well, e.g. `model.fc` that log all of its inner weights. Isn’t this enough?

vfdev-5:
No, I think this can be not enough. Imagine a situation for a NN (e.g. resnet) and we would like to log all convs without BN, FC etc.

sadra-barikbin:
How is it now? User could pass a function which gets weight and decides if it should be logged.

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -419,20 +426,105 @@ class WeightsHistHandler(BaseWeightsHistHandler):
...
+            )
+
+        .. code-block:: python
+
+            from ignite.contrib.handlers.tensorboard_logger import *
+
+            # Create a logger
+            tb_logger = TensorboardLogger(log_dir="experiments/tb_logs")
+
+            # Log weights which their name include 'conv'.
```
vfdev-5, line 454:
```suggestion
            # Log weights which name include 'conv'.
```

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -419,20 +426,105 @@ class WeightsHistHandler(BaseWeightsHistHandler):
...
+                    if m == item:
+                        name = n
+                        break
+                if name is not None:
+                    for n, p in item.named_parameters(prefix=name):
+                        self.weights[n] = p
+                else:
+                    raise ValueError(f"Module {item} is not among model's submodules")
+            else:
+                raise ValueError(f"Whitelist weights shoud be string or nn.Module, given {type(item)}")
```
vfdev-5, line 518: This part of code looks very bulky with all these try/except clauses.
Can we figure out how to simplify a bit the code ?

ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -419,20 +426,105 @@ class WeightsHistHandler(BaseWeightsHistHandler):
...
     def __call__(self, engine: Engine, logger: TensorboardLogger, event_name: Union[str, Events]) -> None:
         if not isinstance(logger, TensorboardLogger):
             raise RuntimeError("Handler 'WeightsHistHandler' works only with TensorboardLogger")
 
         global_step = engine.state.get_event_attrib_value(event_name)
         tag_prefix = f"{self.tag}/" if self.tag else ""
-        for name, p in self.model.named_parameters():
-            if p.grad is None:
-                continue
+        weights = self.weights.items() if self.weights is not None else self.model.named_parameters()
```
vfdev-5, line 526: Can't we define `self.weights = self.model.named_parameters()` in the constructeur if `whitelist is None` instead of here ?

vfdev-5 (commented):
Thanks for the PR @sadra-barikbin 
I left few comments on to improve the code

ignite/contrib/handlers/tensorboard_logger.py
sadra-barikbin: I changed it. How is it?

vfdev-5: Well, still looks too complicated.
Let's try to simplify it in the following way:
```python
weights = {}
for name, param in model.named_parameters():    
    for item in whitelist:        
        if name.startswith(item):
            weights[name] = param
```
Here I assume that `item` is string. However, in your API `item` can be a `nn.Module`. The idea is to rewrite `whitelist` into `List[str]` and replace entries like `model.fc` or `model.layer3` by string values...

sadra-barikbin: You mean to remove entry type of callable as well?

sadra-barikbin: I changed `whitelist` to be `List[str]`. Is it OK now?

vfdev-5: Very sorry for delay @sadra-barikbin , actually, `callable` type was OK and gives a lot of flexibility.

vfdev-5:
@sadra-barikbin thanks for the updates ! Can you now run [this example](https://github.com/pytorch/ignite/blob/master/examples/contrib/mnist/mnist_with_tensorboard_logger.py) with specifying `whitelist=["conv", ]` here: https://github.com/pytorch/ignite/blob/82a929d3e452ccaba8c7767148231e741b6335e5/examples/contrib/mnist/mnist_with_tensorboard_logger.py#L136

and report here the result ?

sdesrozis (approved):
@sadra-barikbin Thanks very much, LGTM ! Although a comment still to be addressed.

sadra-barikbin:
Hi, I ran the example. Result is depicted in picture below:

image

By the way, there was mistakes in lines 138 and 142 in `event_name` parameter.

vfdev-5 (approved):
Thanks @sadra-barikbin !

vfdev-5:
@sadra-barikbin can you add please a similar arg to other weights and grad logging handlers ?

sadra-barikbin:
Sure. Shall I open a new issue for them? You mean just Tensorboard ones?

vfdev-5:
yes, please open an issue and suggest the classes where we could add such new argument.
By the way, I think we forgot about docs tag for new args in this PR: [`.. versionadded::`](https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md#writing-documentation) (check the codebase for examples of usage)
Please send a follow-up update. Thanks

PR #2530, opened 2022-03-25
Update mean_squared_error.py

leotac:
Update docstring of MSE so that it's consistent with the actual behavior.

Description:

MSE is actually computing the sum of squared differences for each element i before taking the average. The current documentation only covers the case where y is a vector (predicting scalar values), but in cases where the prediction has higher dimension than 1 the formula is not valid/clear. 
With the change, it would be valid for both cases.

Check list:


vfdev-5:
@leotac thanks for the update! We'll check docs rendering once GHA will be working again.

vfdev-5 (approved):
LGTM,

Issue #2522, opened 2022-03-22
Improve tests using pytest parametrize


vfdev-5:
## Improve tests using pytest parametrize

The idea is to improve our testing code using [`pytest.mark.parametrize`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-mark-parametrize).

Currently some tests are implemented as : 
```python
def test_something():

    def _test(a):
       # internal function doing some tests
       # For example, a dummy check below
       assert 0 < a < 5

    a = 1
    _test(a)

    a = 2
    _test(a, b, c)
```

We would like to implement that using `pytest.mark.parametrize`:
```python

@pytest.mark.parametrize("a", [1, 2])
def test_something(a):
    # internal function doing some tests
    # For example, a dummy check below
    assert 0 < a < 5
```

Another example PR doing that for a test case: https://github.com/pytorch/ignite/pull/2521/files

## What to do:
- [ ] Read CONTRIBUTING to see how to develop and execute a specific test: https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md#run-tests 
- [ ] Explore [test codebase](https://github.com/pytorch/ignite/tree/master/tests/ignite) and identify tests to improve. Please pay attention that **some tests** most probably can't be reimplemented with `pytest.mark.parametrize` as they can generated random data. Let's skip all `_test_distrib_*` tests as they can be particularly complicated to port.
- [ ] Once a test to improve is identified, comment out here about it and propose a draft code snippet of how you think to reimplement it using `pytest.mark.parametrize`.
- [ ] In parallel, you can try to reimplement it locally and test if the new test is still passing. Please make sure not to modify test assertions to make test pass as your PR wont be accepted.
- [ ] Send a draft PR

Thanks!

nmcguire101:
Hi, I would like to implement this.

vfdev-5:
@nmcguire101 please follow "What to do" guidelines

divo12:
In https://github.com/pytorch/ignite/blob/3a286b1d13a4a0476b3ee8e8046f16465818c9f6/tests/ignite/metrics/gan/test_fid.py#L158
https://github.com/pytorch/ignite/blob/3a286b1d13a4a0476b3ee8e8046f16465818c9f6/tests/ignite/metrics/gan/test_inception_score.py#L68
the _test function could be improved in the _test_distrib_integration function
Draft code:
```
def _test_distrib_integration(device):

    from ignite.engine import Engine

[...]

PR #2531, opened 2022-03-27
updated test_param_scheduler replacing multiple calls with pytest.parametrize

Ishan-Kumar2:
Addresses #2522

Description:
Updated handlers/test_param_scheduler replacing multiple calls with pytest.parametrize

Haven't converted 

https://github.com/pytorch/ignite/blob/69ea3c8fef56191335771665c6acb9937f7a8668/tests/ignite/handlers/test_param_scheduler.py#L812 
and https://github.com/pytorch/ignite/blob/69ea3c8fef56191335771665c6acb9937f7a8668/tests/ignite/handlers/test_param_scheduler.py#L1228

because it has conditional calls. We could have different fixtures for all the cases but I believe that would make the code less readable, let me know what you think would be best for these tests.
Check list:

### DIFF
--- a/tests/ignite/handlers/test_param_scheduler.py
+++ b/tests/ignite/handlers/test_param_scheduler.py
@@ -637,12 +637,12 @@ def test_lr_scheduler_asserts():
     [
         (StepLR, ({"step_size": 5, "gamma": 0.5})),
         (ExponentialLR, ({"gamma": 0.78})),
-        (MultiplicativeLR, ({"lr_lambda": lambda epoch: 0.95})),
+        (MultiplicativeLR if has_multiplicative_lr else None, ({"lr_lambda": lambda epoch: 0.95})),
     ],
 )
 def test_lr_scheduler(torch_lr_scheduler_cls, kwargs):
 
-    if torch_lr_scheduler_cls == MultiplicativeLR and not has_multiplicative_lr:
+    if torch_lr_scheduler_cls is None:
         return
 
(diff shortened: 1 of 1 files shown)


tests/ignite/handlers/test_param_scheduler.py
```diff
@@ -637,66 +632,69 @@ def test_lr_scheduler_asserts():
...
 
 
-def test_lr_scheduler():
-    def _test(torch_lr_scheduler_cls, **kwargs):
+@pytest.mark.parametrize(
+    "torch_lr_scheduler_cls, kwargs",
+    [
+        (StepLR, ({"step_size": 5, "gamma": 0.5})),
+        (ExponentialLR, ({"gamma": 0.78})),
+        (MultiplicativeLR, ({"lr_lambda": lambda epoch: 0.95})),
```
vfdev-5, line 640: This wont work as older versions wont be able to import the class

vfdev-5:
@Ishan-Kumar2 thanks for the PR, I left a comment to address.

> We could have different fixtures for all the cases but I believe that would make the code less readable, let me know what you think would be best for these tests.

I agree, let's keep them as they are now...

tests/ignite/handlers/test_param_scheduler.py
Ishan-Kumar2: I have added a fix, please have a look thanks!

vfdev-5 (approved):
@Ishan-Kumar2 LGTM, thanks for the PR !

PR #2543, opened 2022-04-08
Added: Improved tests using pytest parametrize

### LINKED ISSUE #2522: Improve tests using pytest parametrize

nmcguire101:
Related to #2522 (edited by @vfdev-5 to avoid closing the issue when this PR is merged)

Description:
Improved test tests\ignite\contrib\metrics\regression\test__base.py using pytest parametrize

Check list:

### DIFF
--- a/tests/ignite/metrics/test_root_mean_squared_error.py
+++ b/tests/ignite/metrics/test_root_mean_squared_error.py
@@ -18,7 +18,7 @@ def test_zero_sample():
 
 
 @pytest.fixture(params=[0, 1, 2, 3])
-def generate_tests(request):
+def test_data(request):
     return [
         (torch.empty(10).uniform_(0, 10), torch.empty(10).uniform_(0, 10), 1),
         (torch.empty(10, 1).uniform_(-10, 10), torch.empty(10, 1).uniform_(-10, 10), 1),
--- a/tests/ignite/metrics/test_root_mean_squared_error.py
+++ b/tests/ignite/metrics/test_root_mean_squared_error.py
@@ -33,7 +33,7 @@ def test_compute(n_times, generate_tests):
 
     rmse = RootMeanSquaredError()
 
-    (y_pred, y, batch_size) = generate_tests
+    y_pred, y, batch_size = test_data
     rmse.reset()
     if batch_size > 1:
         n_iters = y.shape[0] // batch_size + 1
--- a/tests/ignite/metrics/test_root_mean_squared_error.py
+++ b/tests/ignite/metrics/test_root_mean_squared_error.py
@@ -28,12 +28,13 @@ def generate_tests():
     ]
 
 
-@pytest.mark.parametrize("n_times", range(5))
+@pytest.mark.parametrize("n_times", range(3))
 def test_compute(n_times, generate_tests):
 
     rmse = RootMeanSquaredError()
 
-    def _test(y_pred, y, batch_size):
+    for test in generate_tests:
+        (y_pred, y, batch_size) = test
         rmse.reset()
         if batch_size > 1:
             n_iters = y.shape[0] // batch_size + 1
@@ -51,10 +52,6 @@ def test_compute(n_times, generate_tests):
 
         assert isinstance(res, float)
         assert pytest.approx(res) == np_res
-
-    for test in generate_tests:
-        (y_pred, y, batch_size) = test
-        _test(y_pred, y, batch_size)
 
 
 def _test_distrib_integration(device, tol=1e-6):
--- a/tests/ignite/metrics/test_root_mean_squared_error.py
+++ b/tests/ignite/metrics/test_root_mean_squared_error.py
@@ -27,7 +27,8 @@ def test_zero_sample():
         (torch.empty(50, 1).uniform_(-10, 10), torch.empty(50, 1).uniform_(-10, 10), 16),
     ],
 )
-def test_compute(test_cases):
+@pytest.mark.parametrize("dummy_val", [(0, 1, 2, 3, 4)])
+def test_compute(test_cases, dummy_val):
 
     rmse = RootMeanSquaredError()
 
@@ -50,10 +51,8 @@ def test_compute(test_cases):
         assert isinstance(res, float)
         assert pytest.approx(res) == np_res
 
-    for _ in range(5):
-        # check multiple random inputs as random exact occurencies are rare
-        (y_pred, y, batch_size) = test_cases
-        _test(y_pred, y, batch_size)
+    (y_pred, y, batch_size) = test_cases
+    _test(y_pred, y, batch_size)
 
 
 def _test_distrib_integration(device, tol=1e-6):
--- a/tests/ignite/metrics/test_root_mean_squared_error.py
+++ b/tests/ignite/metrics/test_root_mean_squared_error.py
@@ -17,24 +17,23 @@ def test_zero_sample():
         rmse.compute()
 
 
-@pytest.fixture
-def generate_tests():
+@pytest.fixture(params=[0, 1, 2, 3])
+def generate_tests(request):
     return [
         (torch.empty(10).uniform_(0, 10), torch.empty(10).uniform_(0, 10), 1),
         (torch.empty(10, 1).uniform_(-10, 10), torch.empty(10, 1).uniform_(-10, 10), 1),
         # updated batches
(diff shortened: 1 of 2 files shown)


vfdev-5:
@nmcguire101 thanks for the PR, however I'm not sure if this change is an improvement. This test with asserts I think does not require any reparametrization. You can also see that your PR removes 14 lines and add 19 lines.
Let's improve other tests.

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -40,23 +50,10 @@ def _test(y_pred, y, batch_size):
...
-        ]
-
-        return test_cases
-
     for _ in range(5):
         # check multiple random inputs as random exact occurencies are rare
-        test_cases = get_test_cases()
-        for y_pred, y, batch_size in test_cases:
-            _test(y_pred, y, batch_size)
+        (y_pred, y, batch_size) = test_cases
```
vfdev-5, line 55: @nmcguire101 If you take a look attentively what does `test_cases = get_test_cases()` previously, you can see that in the loop `for _ in range(5):` all `test_cases` are different. In your PR however, they are all the same and thus there is no point to make the loop.
Can you check however whether `test_cases` become different if you add another dummy arg to `@pytest.mark.parametrize` to represent the loop `for _ in range(5):` ?

vfdev-5: @nmcguire101 Can you check however whether `test_cases` are different with `dummy_val` ? Thanks

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -40,23 +51,8 @@ def _test(y_pred, y, batch_size):
...
-
-        return test_cases
-
-    for _ in range(5):
-        # check multiple random inputs as random exact occurencies are rare
-        test_cases = get_test_cases()
-        for y_pred, y, batch_size in test_cases:
-            _test(y_pred, y, batch_size)
+    (y_pred, y, batch_size) = test_cases
+    _test(y_pred, y, batch_size)
```
vfdev-5, line 55: The idea of using `pytest.mark.parametrize` to get rid of `_test` inner method

tests/ignite/metrics/test_root_mean_squared_error.py
nmcguire101: I realized there was an error in my last PR, this latest one adds a few lines but generates unique values. This fix may be trivial as it only gets rid of two lines. If that's the case I'll keep looking for other tests to improve.

vfdev-5: I see, I also checked that parametrize does not generate different data for different with `dummy_val`.
As commented previously, the idea of using pytest.mark.parametrize to get rid of _test inner method.
Can you check fixture parametrize: https://docs.pytest.org/en/6.2.x/fixture.html#fixture-parametrize to see if we could get what we want: 1) get rid of `_test`, 2) have different test values (`y_pred, y, batch_size`) and 3) run the check with a parametrized loop for 3 times ?

nmcguire101: I was able to get rid of _test and generate unique test values for `y_pred,` y, batch_size`. However, I am confused what you mean by run the check with a parametrized loop for 3 times because based on the original code it should be run 5 times (unless I am misunderstanding what you're referring to here). I will submit a new PR after getting clarification on the third point.

nmcguire101: I was able to get rid of _test and generate unique test values for `y_pred,` `y,` `batch_size`. However, I am confused what you mean by run the check with a parametrized loop for 3 times because based on the original code it should be run 5 times (unless I am misunderstanding what you're referring to here). I will submit a new PR after getting clarification on the third point.

vfdev-5: Sorry, i wanted just to reduce the number of times to run the test. 5 -> 3

vfdev-5: @nmcguire101 do you think we can get rid of `for test in generate_tests:` loop using fixture parametrize ?

nmcguire101: I'm not sure that would work without defining four different fixtures, one for each test case. Do you have a different idea?

nmcguire101: I took another look at it and found a way to get rid of the for loop.

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -17,46 +17,40 @@ def test_zero_sample():
         rmse.compute()
 
 
-def test_compute():
+@pytest.fixture(params=[0, 1, 2, 3])
+def generate_tests(request):
```
vfdev-5, line 21:
```suggestion
def test_data(request):
```

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -17,46 +17,40 @@ def test_zero_sample():
...
+        (torch.empty(10).uniform_(0, 10), torch.empty(10).uniform_(0, 10), 1),
+        (torch.empty(10, 1).uniform_(-10, 10), torch.empty(10, 1).uniform_(-10, 10), 1),
+        # updated batches
+        (torch.empty(50).uniform_(0, 10), torch.empty(50).uniform_(0, 10), 16),
+        (torch.empty(50, 1).uniform_(-10, 10), torch.empty(50, 1).uniform_(-10, 10), 16),
+    ][request.param]
+
+
+@pytest.mark.parametrize("n_times", range(3))
+def test_compute(n_times, generate_tests):
```
vfdev-5, line 32:
```suggestion
def test_compute(n_times, test_data):
```

tests/ignite/metrics/test_root_mean_squared_error.py
```diff
@@ -17,46 +17,40 @@ def test_zero_sample():
...
-        ]
-
-        return test_cases
-
-    for _ in range(5):
-        # check multiple random inputs as random exact occurencies are rare
-        test_cases = get_test_cases()
-        for y_pred, y, batch_size in test_cases:
-            _test(y_pred, y, batch_size)
+    (y_pred, y, batch_size) = generate_tests
```
vfdev-5, line 36:
```suggestion
    y_pred, y, batch_size = test_data
```

vfdev-5 (commented):
Thanks for the update @nmcguire101 
Few naming nits and good to go, I think

vfdev-5 (approved):
Thanks @nmcguire101 
LGTM

vfdev-5:
@nmcguire101 would you mind to apply similar ideas to other tests ?

nmcguire101:
Yes, I will do that. Thank you for all your help

PR #2546, opened 2022-04-13
ADD COCO mAP metric

louis-she:


ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,89 @@
+__all__ = ["MeanAveragePrecision"]
+
+from typing import Callable, List, Optional, Sequence, Union
+
+import torch
+from torchvision.ops import box_iou
```
vfdev-5, line 6: Let's do something like that for torchvision dependency:
https://github.com/pytorch/ignite/blob/a0383ff22a98e53b556382a29a06bb9e2995645c/ignite/metrics/gan/fid.py#L169-L172

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(
+        self,
+        iou_thresholds: Optional[Union[List[float], torch.Tensor]] = None,
+        output_transform: Callable = lambda x: x,
+        device: Union[str, torch.device] = torch.device("cpu"),
+    ) -> None:
+        if iou_thresholds is None:
+            iou_thresholds = torch.range(0.5, 0.99, 0.05)
+        self.iou_thresholds = torch.tensor(iou_thresholds, device=device)
+        self.rec_thresholds = torch.range(0, 1, 0.01, device=device)
```
vfdev-5, line 21: Maybe, `torch.linspace(0, 1, 100)` instead ?

vfdev-5 (commented):
Thanks a lot for the PR @louis-she 
I haven't yet checked update/compute methods but just leave few coding comments.
I'll go deeper into the implementation a bit later...

ignite/metrics/mean_average_precision.py
louis-she: It's `torch.linspace(0, 1, 101)`

louis-she:
@vfdev-5 Thanks for reviewing the code. I've added some comments of the `__init__` and `update` methods to explain what the input should be. What's your opinion about that part?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,113 @@
+__all__ = ["MeanAveragePrecision"]
+
+from typing import Callable, List, Optional, Tuple, Union
+
+import torch
+
+try:
+    from torchvision.ops import box_iou
+except ImportError:
+    raise RuntimeError("This module requires torchvision to be installed.")
```
vfdev-5, line 10: @louis-she please put this into `MeanAveragePrecision` constructor. Otherwise, while imported in `__init__` it will directly fail if no torchvision.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,113 @@
...
+                non-blocking. By default, CPU.
+        """
+        if iou_thresholds is None:
+            iou_thresholds = torch.range(0.5, 0.99, 0.05)
+        self.iou_thresholds = torch.tensor(iou_thresholds, device=device)
+        self.rec_thresholds = torch.linspace(0, 1, 101, device=device)
+        super(MeanAveragePrecision, self).__init__(output_transform=output_transform, device=device)
+
+    def reset(self) -> None:
+        self._cm = {}
```
vfdev-5, line 42: How about using `defaultdict(defaultdict(list))` instead ?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,113 @@
...
+    def update(self, output: Tuple[torch.Tensor, torch.Tensor]) -> None:
+        """
+        Args:
+            output: a tuple of 2 tensors in which the first one is the truth and the second one is the prediction.
+                the shape of the output[0](truth) is (N, 5) where N stands for the number of ground truth boxes and 5 is
+                (x1, x2, y1, y2, class_number); the shape of the output[1](prediction) is (M, 6) where M stands for the
+                number of predicted boxes and 6 is (x1, x2, y1, y2, confidence, class_number).
+        """
+        y, y_pred = output[0].detach(), output[1].detach()
+        iou = box_iou(y_pred[:, :4], y[:, :4])
```
vfdev-5, line 53: I think we should handle explicitly the cases like:
1) no preds and no GT (both are empty)
2) no preds (preds are empty) and non-empty GT
3) non-empty preds and empty GT

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def update(self, output: Tuple[torch.Tensor, torch.Tensor]) -> None:
+        """
+        Args:
+            output: a tuple of 2 tensors in which the first one is the truth and the second one is the prediction.
+                the shape of the output[0](truth) is (N, 5) where N stands for the number of ground truth boxes and 5 is
+                (x1, x2, y1, y2, class_number); the shape of the output[1](prediction) is (M, 6) where M stands for the
+                number of predicted boxes and 6 is (x1, x2, y1, y2, confidence, class_number).
+        """
+        y, y_pred = output[0].detach(), output[1].detach()
```
vfdev-5, line 52: Let's add checks if shapes are correct for provided `y` and `y_pred`

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,113 @@
...
+        iou = box_iou(y_pred[:, :4], y[:, :4])
+        for iou_thres in self.iou_thresholds:
```
vfdev-5, line 54: @louis-she what's the reference for your implementation ?

louis-she: Mainly referenced from the official pycocotools.

louis-she:
Removed the WIP status.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,133 @@
...
+                form expected by the metric. This can be useful if, for example, you have a multi-output model and
+                you want to compute the metric with respect to one of the outputs.
+                By default, metrics require the output as ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y}``.
+            device: specifies which device updates are accumulated on. Setting the
+                metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
+                non-blocking. By default, CPU.
+        """
+        global box_iou
+        try:
+            from torchvision.ops import box_iou
```
vfdev-5, line 35: Let's make it in the following way:
```suggestion     
            from torchvision.ops import box_iou
            self.box_iou = box_iou
```

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,133 @@
...
+                the shape of the output[0](truth) is (N, 5) where N stands for the number of ground truth boxes and 5 is
+                (x1, x2, y1, y2, class_number); the shape of the output[1](prediction) is (M, 6) where M stands for the
+                number of predicted boxes and 6 is (x1, x2, y1, y2, confidence, class_number).
+        """
+        y, y_pred = output[0].detach(), output[1].detach()
+
+        assert y.shape[1] == 5, f"Provided y with a wrong shape, expected (N, 5), got {y.shape}"
+        assert y_pred.shape[1] == 6, f"Provided y_hat with a wrong shape, expected (N, 6), got {y.shape}"
+
+        iou = box_iou(y_pred[:, :4], y[:, :4])
```
vfdev-5, line 61:
```suggestion
        iou = self.box_iou(y_pred[:, :4], y[:, :4])
```

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,133 @@
+__all__ = ["MeanAveragePrecision"]
+
+from collections import defaultdict
+from typing import Callable, List, Optional, Tuple, Union
+
+import torch
+
+from ignite.metrics.metric import Metric
+
+box_iou = None
```
vfdev-5, line 10:
```suggestion
```

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,133 @@
...
+            output_transform: a callable that is used to transform the
+                :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
+                form expected by the metric. This can be useful if, for example, you have a multi-output model and
+                you want to compute the metric with respect to one of the outputs.
+                By default, metrics require the output as ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y}``.
+            device: specifies which device updates are accumulated on. Setting the
+                metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
+                non-blocking. By default, CPU.
+        """
+        global box_iou
```
vfdev-5, line 33:
```suggestion
```

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,133 @@
...
+            raise RuntimeError("This module requires torchvision to be installed.")
+
+        if iou_thresholds is None:
+            iou_thresholds = torch.range(0.5, 0.99, 0.05)
+        self.iou_thresholds = torch.tensor(iou_thresholds, device=device)
+        self.rec_thresholds = torch.linspace(0, 1, 101, device=device)
+        super(MeanAveragePrecision, self).__init__(output_transform=output_transform, device=device)
+
+    def reset(self) -> None:
+        self._cm = defaultdict(lambda: defaultdict(lambda: defaultdict(lambda: [])))
```
vfdev-5, line 46:
```suggestion
        self._cm = defaultdict(lambda: defaultdict(list))
```

louis-she:
The implementation of the metric is referenced from the official pycocotools. The average precision is computed dataset-wise ( not sample-wise) so I think it's not possible to use distribution. 

So is there any other things to do with the PR?

@vfdev-5

vfdev-5:
@louis-she Please check : https://github.com/pytorch/ignite/pull/2546/files#r853542677

> The average precision is computed dataset-wise ( not sample-wise) so I think it's not possible to use distribution.

Torchvision does DDP with pycocotools, have you check how they do ?

Not supporting DDP would be very large limitation. We have to supported it.

From my side I plan to play around and check the results of the metric vs other packages (I cited previously) to see the diffs etc...

louis-she:
> Having a wrapper of pycocotools is not so interesting.

@sdesrozis I don't think this implementation is **a wrapper of pycocotools**, the codes are written from scratch and are different from pycocotools. My original goal here is to match the result of `pycocotools`.

I have had some testing with distributed computing. Due to the different shapes of `Tensor` in the `self._cm`, I can only use the `torch.distributed.gather_object` to gather the results. But since the `torch.distributed.gather_object` uses `pickle` to transfer the data, I met a performance issue here.

I tested the code with a 6 CPU core machine, computing without distribution is 2x faster than with distribution. I addressed the performance bottleneck at the pickling process.

sadra-barikbin:
@vfdev-5, I finished the implementation and added some tests. Failure in XLA and Horovod is due to usage of  `dist.gather`.

sadra-barikbin:
Some other failures is due to a deprecated type in Numpy being used in cocoeval code. There's already an issue for that in their repo but yet to fix it. I made the PR for them.

vfdev-5:
@sadra-barikbin can you provide a link to your PR ?

sadra-barikbin:
I added commits to this same PR

sadra-barikbin:
Two notes:
- @louis-she 's matching implementation did not align with COCO, although was written smartly. I've described coco matching in `test_matching` docstring.
- PR asumes that the input to the `update` is not batched maybe because samples do not have the same number of prediction or ground truth boxes. I followed this assumption.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,363 @@
...
+                0.9500000000000001,
+                0.96,
+                0.97,
+                0.98,
+                0.99,
+                1.0,
+            ],
+            dtype=torch.double,
+            device=device,
+        )  # torch.linspace(0, 1, 101, device=device, dtype=torch.double)
```
vfdev-5, line 178: Let's use torch.linspace and figure out either test tolerences or metric computation accuracy

sadra-barikbin:
@louis-she , thanks for this PR. Why did you decide to find number of categories dynamically and not by an argument like `num_categories`?

louis-she:
> @louis-she , thanks for this PR. Why did you decide to find number of categories dynamically and not by an argument like `num_categories`?

I'm sure there is a reason for that, but I can not remember exactly what it is... It could be a compromise to match the results of the official coco implementation.

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,252 @@
...
+        """
+        try:
+            from torchvision.ops import box_iou
+
+            self.box_iou = box_iou
+        except ImportError:
+            raise ModuleNotFoundError("This module requires torchvision to be installed.")
+
+        if iou_thresholds is None:
+            self.iou_thresholds = [
```
vfdev-5, line 41: Can we replace this with `torch.linspace(0.5, 0.95, 10, device=device, dtype=torch.double)` ?

ignite/metrics/mean_average_precision.py
```diff
@@ -0,0 +1,252 @@
...
+                0.8999999999999999,
+                0.95,
+            ]  # torch.linspace(0.5, 0.95, 10).tolist()
+        elif isinstance(iou_thresholds, torch.Tensor):
+            if iou_thresholds.ndim != 1:
+                raise ValueError(
+                    "`iou_thresholds` should be a one-dimensional tensor or a list of floats"
+                    f", given a {iou_thresholds.ndim}-dimensional tensor."
+                )
+            self.iou_thresholds = iou_thresholds.sort().values.tolist()
```
vfdev-5, line 59: Please add a test for this case

sadra-barikbin: For which case exactly? We have `test_wrong_input` now.

ignite/metrics/mean_average_precision.py
sadra-barikbin: Yes. I did and even I could decrease `tol` from `1e-3` to `1e-4`.

PR #2547, opened 2022-04-13
Feature weights hist handler issue 2328 - Fix docstring

### LINKED ISSUE #2328: WeightsHistHandler should plot all weights (incl those without grad)

sadra-barikbin:
Fixes #2328 

Description:
Fix bug in docstring


ignite/contrib/handlers/tensorboard_logger.py
```diff
@@ -459,6 +459,9 @@ class WeightsHistHandler(BaseWeightsHistHandler):
                 event_name=Events.ITERATION_COMPLETED,
                 log_handler=WeightsHistHandler(model, whitelist=weight_selector)
             )
+
+    ..  versionchanged:: 0.5.0
+        accepts an optional `whitelist` parameter
```
vfdev-5, line 464:
```suggestion
        optional argument `whitelist` added.
```

vfdev-5 (approved):
Thanks @sadra-barikbin, LGTM 
I rephrased the message a bit.

PR #2555, opened 2022-04-19
Fix zero_grad place resulting in zero logs

### LINKED ISSUE #2459: GradsScalarHandler logs 0 gradients if default update function is used

sadra-barikbin:
Fixes #2459 

Description:
Based on PR #2470 . Tests are added.

Check list:


tests/ignite/engine/test_create_supervised.py
```diff
@@ -124,6 +124,37 @@ def _():
...
+        data = [(_x, _y) for _x, _y in zip(x, y)]
+
+        @trainer.on(Events.ITERATION_COMPLETED)
+        def _():
+            assert model.weight.data.grad != 0
+
+        trainer.run(data)
+
+    _test(1)
+    _test(3)
```
vfdev-5, line 155: Can you please use `@pytest.mark.parametrize` for that directly instead of calling `_test(1)`, `_test(3)`

tests/ignite/contrib/handlers/conftest.py
```diff
@@ -1,6 +1,5 @@
 from unittest.mock import Mock
 
-import numpy as np
```
vfdev-5, line 3: This should be reverted

vfdev-5 (commented):
Thanks for the PR @sadra-barikbin 
Few comments to make it better

tests/ignite/engine/test_create_supervised.py
```diff
@@ -340,25 +369,39 @@ def _test_create_evaluation_step(
     assert output_transform_mock.called
 
 
-def test_create_supervised_trainer():
+@pytest.mark.parametrize("gradient_accumulation_steps", [1, 3])
```
vfdev-5, line 372: I do not understand why you changed this ?

tests/ignite/engine/test_create_supervised.py
```diff
@@ -340,25 +369,39 @@ def _test_create_evaluation_step(
...
+def test_create_supervised_trainer(gradient_accumulation_steps):
     _test_create_supervised_trainer_wrong_accumulation()
-    _test_create_supervised_trainer(gradient_accumulation_steps=1)
-    _test_create_supervised_trainer(gradient_accumulation_steps=3)
+    _test_create_supervised_trainer(gradient_accumulation_steps=gradient_accumulation_steps)
+    _test_create_supervised_trainer(gradient_accumulation_steps=gradient_accumulation_steps)
```
vfdev-5, line 376: Here you call the same test twice ?

tests/ignite/engine/test_create_supervised.py
sadra-barikbin: You mean there is no need to `_test_create_supervised_trainer_have_grad_after_iteration` with different `gradient_accumulation_steps`?

sadra-barikbin: @vfdev-5 You mean there is no need to _test_create_supervised_trainer_have_grad_after_iteration with different gradient_accumulation_steps?

vfdev-5: @sadra-barikbin try to execute these tests and see what this code do.
It will generate 2 tests like 
`test_create_supervised_trainer[gradient_accumulation_steps-1]` and `test_create_supervised_trainer[gradient_accumulation_steps-3]` where 
`_test_create_supervised_trainer_wrong_accumulation()` and `_test_create_mocked_supervised_trainer()` will be executed twice which is useless.

I suggest to revert the change and just call 
```
_test_create_supervised_trainer_have_grad_after_iteration(1)
_test_create_supervised_trainer_have_grad_after_iteration(3)
```

Another solution could be split `test_create_supervised_trainer` into 2 tests such that the first test is executing once `_test_create_supervised_trainer_wrong_accumulation` and `_test_create_mocked_supervised_trainer` and the second is parametrizing `gradient_accumulation_steps` and calling others

vfdev-5: > Another solution could be split test_create_supervised_trainer into 2 tests such that the first test is executing once _test_create_supervised_trainer_wrong_accumulation and _test_create_mocked_supervised_trainer and the second is parametrizing gradient_accumulation_steps and calling others

@sadra-barikbin can you make suggested changes please ?

sadra-barikbin: Excuse me for delay. Sure.

sadra-barikbin: @vfdev-5 , I did the first one but also did some refactors. Does it look good?

vfdev-5: It's difficult to say as this adds more lines than it removes :)
Can you pull from your branch locally and push an empty commit to restart the CI. Autofix pep does not launch everything somehow...

sadra-barikbin: It's because of applying autopep(It has put each parameter of `pytest.param`s in a separate line). Now there is no redundancy.

sadra-barikbin: @vfdev-5 , what else should I do?

vfdev-5 (approved):
Thanks @sadra-barikbin , lgtm

vfdev-5:
@sadra-barikbin your PR is failing on CUDA:
- https://app.circleci.com/pipelines/github/pytorch/ignite/2573/workflows/1e1e6a3c-95fb-45af-808a-2191b36788ab/jobs/7784
or
- https://github.com/pytorch/ignite/runs/6248177618?check_suite_focus=true#step:10:186

PR #2560, opened 2022-05-02
Fix test on gpu

### LINKED ISSUE #2459: GradsScalarHandler logs 0 gradients if default update function is used

sadra-barikbin:
Fixes #2459 

Description:
There was an issue in `test_create_supervised` tests causing device mismatch error.


tests/ignite/engine/test_create_supervised.py
```diff
@@ -106,7 +106,7 @@ def _():
         assert pytest.approx(trainer.state.output[-1], abs=1e-5) == loss[0]
         accumulation[0] = loss[0] = 0.0
 
-    if model_device == trainer_device or ((model_device == "cpu") ^ (trainer_device == "cpu")):
+    if model_device == trainer_device or ((model_device in ["cpu", None]) and (trainer_device in ["cpu", None])):
```
sadra-barikbin, line 109: I changed this because in the case `trainer_device` and `model_device` were `cuda` and `cpu` or vice versa, we would not catch device mismatch error.

tests/ignite/engine/test_create_supervised.py
```diff
@@ -124,6 +124,40 @@ def _():
...
+
+    x = torch.tensor([[1.0], [1.0], [1.0], [1.0], [1.0]])
+    y = torch.tensor([[2.0], [3.0], [4.0], [5.0], [6.0]])
+    data = [(_x, _y) for _x, _y in zip(x, y)]
+
+    @trainer.on(Events.ITERATION_COMPLETED)
+    def _():
+        assert model.weight.grad != 0
+
+    if model_device == trainer_device or ((model_device in ["cpu", None]) and (trainer_device in ["cpu", None])):
```
sadra-barikbin, line 153: I copied this statement from `_test_create_supervised_trainer`.

vfdev-5 (approved):
Thanks @sadra-barikbin

PR #2573, opened 2022-05-14
Improve precision recall metric issue 2571

### LINKED ISSUE #2571: Add new options to Precision and Recall metrics
## 🚀 Feature

I suggest to add new options regarding `average` parameter of `Precision` and `Recall` metrics.
Currently they have two options which are `average==False` and `average==True`. I propose to change `average` options to the following:

- `False`: return per class/label metric for multiclass/label input and simple precision for binary input (default option). This is equivalent with `average=None` of Scikit-learn for multiclass/label input and `average=binary` for binary input.
- `micro`: for multilabel, computes metric using all samples of all labels altogether(equivalent with `micro` of Scikit-learn). Incompatible with binary and multiclass inputs, as it would be equivalent with `Accuracy` and user can use that metric.
- `weighted`: It computes metric for every class/label then returns weighted average of them using number of samples in each class/label. Recall does not have this option because weighted recall is equivalent with accuracy. Equivalent with `weighted` of Scikit-Learn.
- `samples`: just for multilabel input. It computes metric for every sample then returns their average. Equivalent with `samples` of Scikit-Learn.

I did not put `macro` option because user can instantiate metric with `average=False` then call `mean` on it to become macro metric.

sadra-barikbin:
Fixes #2571 

Check list:

### DIFF
--- a/ignite/metrics/classification_report.py
+++ b/ignite/metrics/classification_report.py
@@ -4,7 +4,6 @@ from typing import Callable, Collection, Dict, List, Optional, Union
 import torch
 
 from ignite.metrics.fbeta import Fbeta
-from ignite.metrics.metric import Metric
 from ignite.metrics.metrics_lambda import MetricsLambda
 from ignite.metrics.precision import Precision
 from ignite.metrics.recall import Recall
@@ -85,14 +84,14 @@ def ClassificationReport(
                 [0, 0, 0],
                 [1, 0, 0],
                 [0, 1, 1],
-            ]).unsqueeze(0)
+            ])
             y_pred = torch.tensor([
                 [1, 1, 0],
                 [1, 0, 1],
                 [1, 0, 0],
                 [1, 0, 1],
                 [1, 1, 0],
-            ]).unsqueeze(0)
+            ])
             state = default_evaluator.run([[y_pred, y_true]])
             print(state.metrics["cr"].keys())
             print(state.metrics["cr"]["0"])
@@ -119,25 +118,24 @@ def ClassificationReport(
     averaged_fbeta = fbeta.mean()
 
     def _wrapper(
-        recall_metric: Metric, precision_metric: Metric, f: Metric, a_recall: Metric, a_precision: Metric, a_f: Metric
+        re: torch.Tensor, pr: torch.Tensor, f: torch.Tensor, a_re: torch.Tensor, a_pr: torch.Tensor, a_f: torch.Tensor
     ) -> Union[Collection[str], Dict]:
-        p_tensor, r_tensor, f_tensor = precision_metric, recall_metric, f
-        if p_tensor.shape != r_tensor.shape:
+        if pr.shape != re.shape:
             raise ValueError(
                 "Internal error: Precision and Recall have mismatched shapes: "
-                f"{p_tensor.shape} vs {r_tensor.shape}. Please, open an issue "
+                f"{pr.shape} vs {re.shape}. Please, open an issue "
                 "with a reference on this error. Thank you!"
             )
         dict_obj = {}
-        for idx, p_label in enumerate(p_tensor):
+        for idx, p_label in enumerate(pr):
             dict_obj[_get_label_for_class(idx)] = {
                 "precision": p_label.item(),
-                "recall": r_tensor[idx].item(),
-                "f{0}-score".format(beta): f_tensor[idx].item(),
+                "recall": re[idx].item(),
+                "f{0}-score".format(beta): f[idx].item(),
             }
         dict_obj["macro avg"] = {
-            "precision": a_precision.item(),
-            "recall": a_recall.item(),
+            "precision": a_pr.item(),
+            "recall": a_re.item(),
             "f{0}-score".format(beta): a_f.item(),
         }
         return dict_obj if output_dict else json.dumps(dict_obj)
--- a/ignite/metrics/fbeta.py
+++ b/ignite/metrics/fbeta.py
@@ -56,8 +56,8 @@ def Fbeta(
 
         .. testcode:: 1
 
-            P = Precision(average=False)
-            R = Recall(average=False)
+            P = Precision(average=False)[1] # `[1]` is to select the value for class 1 that is TP/(TP+FP)
+            R = Recall(average=False)[1] # `[1]` is to select the value for class 1 that is TP/(TP+FN)
             metric = Fbeta(beta=1.0, precision=P, recall=R)
             metric.attach(default_evaluator, "f-beta")
             y_true = torch.tensor([1, 0, 1, 1, 0, 1])
@@ -127,8 +127,8 @@ def Fbeta(
                 y_pred = torch.round(y_pred)
                 return y_pred, y
 
-            P = Precision(average=False, output_transform=thresholded_output_transform)
-            R = Recall(average=False, output_transform=thresholded_output_transform)
+            P = Precision(average=False, output_transform=thresholded_output_transform)[1]
+            R = Recall(average=False, output_transform=thresholded_output_transform)[1]
             metric = Fbeta(beta=1.0, precision=P, recall=R)
             metric.attach(default_evaluator, "f-beta")
             y_true = torch.tensor([1, 0, 1, 1, 0, 1])
--- a/ignite/metrics/fbeta.py
+++ b/ignite/metrics/fbeta.py
@@ -56,8 +56,8 @@ def Fbeta(
 
         .. testcode:: 1
 
-            P = Precision(average=False)[1] # `[1]` is to select the value for class 1 that is TP/(TP+FP)
-            R = Recall(average=False)[1] # `[1]` is to select the value for class 1 that is TP/(TP+FN)
+            P = Precision(average=False)
+            R = Recall(average=False)
             metric = Fbeta(beta=1.0, precision=P, recall=R)
             metric.attach(default_evaluator, "f-beta")
             y_true = torch.tensor([1, 0, 1, 1, 0, 1])
@@ -127,8 +127,8 @@ def Fbeta(
                 y_pred = torch.round(y_pred)
                 return y_pred, y
 
-            P = Precision(average=False, output_transform=thresholded_output_transform)[1]
-            R = Recall(average=False, output_transform=thresholded_output_transform)[1]
+            P = Precision(average=False, output_transform=thresholded_output_transform)
+            R = Recall(average=False, output_transform=thresholded_output_transform)
             metric = Fbeta(beta=1.0, precision=P, recall=R)
             metric.attach(default_evaluator, "f-beta")
             y_true = torch.tensor([1, 0, 1, 1, 0, 1])
--- a/ignite/metrics/precision.py
+++ b/ignite/metrics/precision.py
@@ -19,7 +19,7 @@ class _BasePrecisionRecall(_BaseClassification):
         is_multilabel: bool = False,
         device: Union[str, torch.device] = torch.device("cpu"),
     ):
-        
+
         self._average = average
         self.eps = 1e-20
         self._updated = False
@@ -30,24 +30,24 @@ class _BasePrecisionRecall(_BaseClassification):
     def _check_type(self, output: Sequence[torch.Tensor]) -> None:
         super()._check_type(output)
 
-        if self._average == 'micro' and self._type in ['binary', 'multiclass']:
+        if self._average == "micro" and self._type in ["binary", "multiclass"]:
             raise ValueError(
                 "`Precision` and `Recall` with average=='micro' and binary or multiclass "
                 "input data are equivalent with `Accuracy`, so use this metric."
             )
-        if self._type in ['binary', 'multiclass'] and self._average == 'samples':
+        if self._type in ["binary", "multiclass"] and self._average == "samples":
             raise ValueError("Average == 'samples' is incompatible with binary and multiclass input data.")
 
     @reinit__is_reduced
     def reset(self) -> None:
-        if self._average == 'samples':
+        if self._average == "samples":
             self._sum_samples_metric = 0
             self._samples_cnt = 0
         else:
             self._true_positives = 0
             self._positives = 0
 
-        if self._average == 'weighted':
+        if self._average == "weighted":
             self._actual_positives = 0
         self._updated = False
 
@@ -59,26 +59,26 @@ class _BasePrecisionRecall(_BaseClassification):
                 f"{self.__class__.__name__} must have at least one example before it can be computed."
             )
         if not self._is_reduced:
-            if self._average == 'samples':
+            if self._average == "samples":
                 self._sum_samples_metric = idist.all_reduce(self._sum_samples_metric)
                 self._samples_cnt = idist.all_reduce(self._samples_cnt)
             else:
                 self._true_positives = idist.all_reduce(self._true_positives)
                 self._positives = idist.all_reduce(self._positives)
-            if self._average == 'weighted':
+            if self._average == "weighted":
                 self._actual_positives = idist.all_reduce(self._actual_positives)
             self._is_reduced = True
 
-        if self._average == 'samples':
+        if self._average == "samples":
             return (self._sum_samples_metric / self._samples_cnt).item()
 
         result = self._true_positives / (self._positives + self.eps)
-        if self._average == 'weighted':
+        if self._average == "weighted":
             return ((result @ self._actual_positives) / (self._actual_positives.sum() + self.eps)).item()
-        elif self._average == 'micro':
+        elif self._average == "micro":
             return result.item()
         else:
-            return result if self._type != 'binary' else result.item()
+            return result if self._type != "binary" else result.item()
 
 
 class Precision(_BasePrecisionRecall):
@@ -254,11 +254,8 @@ class Precision(_BasePrecisionRecall):
         device: Union[str, torch.device] = torch.device("cpu"),
     ):
 
-        if average not in [False, 'micro', 'weighted', 'samples']:
-            raise ValueError(
-                "Argument average should be one of values "
-                "False, 'micro', 'weighted' and 'samples'."
-            )
+        if average not in [False, "micro", "weighted", "samples"]:
+            raise ValueError("Argument average should be one of values " "False, 'micro', 'weighted' and 'samples'.")
         super(Precision, self).__init__(
             output_transform=output_transform, average=average, is_multilabel=is_multilabel, device=device
         )
@@ -274,7 +271,7 @@ class Precision(_BasePrecisionRecall):
             y_pred = y_pred.view(-1)
             y = y.view(-1)
 
-            if self._average == 'weighted':
+            if self._average == "weighted":
                 y = to_onehot(y, num_classes=2)
                 y_pred = to_onehot(y_pred.long(), num_classes=2)
         elif self._type == "multiclass":
@@ -299,22 +296,22 @@ class Precision(_BasePrecisionRecall):
         y = y.to(dtype=torch.float64, device=self._device)
         correct = y * y_pred
 
-        if self._average == 'samples':
+        if self._average == "samples":
 
             all_positives = y_pred.sum(dim=1)
             true_positives = correct.sum(dim=1)
             self._sum_samples_metric += torch.sum(true_positives / (all_positives + self.eps))
             self._samples_cnt += y.size(0)
-        elif self._average == 'micro':
+        elif self._average == "micro":
 
             self._positives += y_pred.sum()
             self._true_positives += correct.sum()
-        else: # _average in [False, 'weighted']
+        else:  # _average in [False, 'weighted']
 
             self._positives += y_pred.sum(dim=0)
             self._true_positives += correct.sum(dim=0)
 
-            if self._average == 'weighted':
+            if self._average == "weighted":
                 self._actual_positives += y.sum(dim=0)
 
         self._updated = True
--- a/ignite/metrics/precision.py
+++ b/ignite/metrics/precision.py
@@ -120,7 +120,7 @@ class Precision(_BasePrecisionRecall):
             'samples'
               for multilabel input, at first, precision is computed on a
               per sample basis and then average across samples is returned.
-              
+
               .. math::
                   \text{Sample-averaged Precision} = \frac{\sum_{n=1}^N \frac{TP_n}{TP_n+FP_n}}{N}
 
--- a/ignite/metrics/precision.py
+++ b/ignite/metrics/precision.py
@@ -20,9 +20,9 @@ class _BasePrecisionRecall(_BaseClassification):
         device: Union[str, torch.device] = torch.device("cpu"),
     ):
 
-        if type(average) != bool and average not in ["macro", "micro", "weighted", "samples"]:
+        if not (isinstance(average, bool) or average in ["macro", "micro", "weighted", "samples"]):
             raise ValueError(
-                "Argument average should be a boolean or one of values" " 'macro', 'micro', 'weighted' and 'samples'."
+                "Argument average should be a boolean or one of values 'macro', 'micro', 'weighted' and 'samples'."
             )
 
         if average is True:
--- a/ignite/metrics/precision.py
+++ b/ignite/metrics/precision.py
@@ -102,30 +102,58 @@ class Precision(_BasePrecisionRecall):
 
             False
               default option. For multicalss and multilabel
-              inputs, per class and per label metric is returned. By calling `mean()` on the
-              metric instance, the `macro` setting (which is unweighted average across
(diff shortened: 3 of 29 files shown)


ignite/metrics/recall.py
```diff
@@ -171,6 +167,9 @@ def __init__(
         is_multilabel: bool = False,
         device: Union[str, torch.device] = torch.device("cpu"),
     ):
+
+        if average not in [False, "micro", "samples"]:
```
vfdev-5, line 171: Why Recall can't be weighted ?

ignite/metrics/precision.py
```diff
@@ -93,139 +126,137 @@ class Precision(_BasePrecisionRecall):
...
     def __init__(
         self,
         output_transform: Callable = lambda x: x,
-        average: bool = False,
+        average: Union[bool, str] = False,
         is_multilabel: bool = False,
         device: Union[str, torch.device] = torch.device("cpu"),
     ):
+
+        if average not in [False, "micro", "weighted", "samples"]:
```
vfdev-5, line 258: We should still support average True and translate it to appropriate str value

vfdev-5 (commented):
@sadra-barikbin thanks a lot for the PR ! 
I left few comments to clear out the API. I haven't checked yet the details but the feature looks cool !
Please check also docs build failure.

ignite/metrics/precision.py
```diff
@@ -15,7 +15,7 @@ class _BasePrecisionRecall(_BaseClassification):
     def __init__(
         self,
         output_transform: Callable = lambda x: x,
-        average: bool = False,
+        average: Union[bool, str] = False,
```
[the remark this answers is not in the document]
sadra-barikbin: For the first point, I've done it in `Precision` and `Recall` themselves because their set of `average` options are not the same. I didn't get the second point. You mean if we got `False`, set `average` to `'false'` to be like other options?

ignite/metrics/recall.py
sadra-barikbin: Weighted recall is equivalent with accuracy. For example, in binary case we have:
image

in which 
image

ignite/metrics/precision.py
sadra-barikbin: Which option do you think is better to translate to? For precision, I suggest `weighted`, as it's `macro` plus considering classes/labels imbalance.

ignite/metrics/recall.py
sadra-barikbin: This is stated [here](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.recall_score.html) as well.

vfdev-5: I think it is not a problem that Weighted Recall == Accuracy. IMO we could support it as scikit-learn does.

ignite/metrics/classification_report.py
```diff
@@ -4,7 +4,6 @@
 import torch
```
vfdev-5, line 4: Let's keep classification report updates to another PR

ignite/metrics/precision.py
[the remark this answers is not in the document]
vfdev-5: You are right we are missing `None` value as scikit-learn which would match previous `average=False`. We should also deprecate `is_multilabel` as it becomes unused. What do you think ?

> I've done it in Precision and Recall themselves because their set of average options are not the same.

Can you detail a bit more this point ? I haven't checked but IMO they should be same.

ignite/metrics/precision.py
vfdev-5: Translation should be without any BC-breaking changes, so I'd say `macro`.

ignite/metrics/recall.py
sadra-barikbin: So, supported options for `Precision` and `Recall` would be equal.

ignite/metrics/precision.py
sadra-barikbin: Got it

ignite/metrics/classification_report.py
sadra-barikbin: Its tests woulf fail, then. Shall I mark them as `xfail`?

ignite/metrics/precision.py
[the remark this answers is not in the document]
sadra-barikbin: Yes, we could remove `multilabel`. It could be infered from data.
`Recall` has not the `weighted` option, but regarding the next comment, I'll add it so they become the same.

ignite/metrics/classification_report.py
vfdev-5: This is curious why the tests are failing then. As right now tests are passing and we introducing new features trying to keep BC, I would expect that tests won't break...

ignite/metrics/precision.py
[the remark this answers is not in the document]
vfdev-5: I was thinking that `average="samples"` is equivalent to `is_multilabel=True, average=True`. Can we use `average="samples"` for binary case ?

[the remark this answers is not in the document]
sadra-barikbin: `average='samples'` for binary case is equivalent with the term below which is not a known metric:
image
According to Scikit, `average='samples'` is "only meaningful for multilabel classification".
Anyway, without `is_multilabel`, we cannot differentiate binary data from multilabel data because we do not have a differentiator assumption on their dimensions. Scikit can, because it assumes binary data is 1-dimensional and multilabel data is 2-dimensional.

ignite/metrics/classification_report.py
sadra-barikbin: You're right. They had been marked as `xfail` before.

tests/ignite/metrics/test_precision.py
```diff
@@ -345,58 +357,40 @@ def get_test_cases():
...
-        assert pr._updated is False
+    with pytest.raises(ValueError):
+        pr.update((y_pred, y))
 
-    _test(average=True)
-    _test(average=False)
+    assert pr._updated is False
 
 
 def _test_distrib_integration_multiclass(device):
```
sadra-barikbin, line 396: Is there any reason that this test has not been parametrized?

sadra-barikbin:
Scikit-Learn has another data type called `multilabel-multioutput` which is like multilabel but its labels are not binary. It has yet to be implemented. We can do it if you will.

tests/ignite/metrics/test_precision.py
vfdev-5: This test and other similar ones are used in test_distrib_hvd, test_distrib_gloo_cpu_or_gpu, test_distrib_nccl_gpu tests. IMO, it is not obvious how to properly parametrize them. However, there was a task to rewrite distributed tests such that we do not need to write all: `test_distrib_hvd, test_distrib_gloo_cpu_or_gpu, test_distrib_nccl_gpu` but only `_test_distrib_integration_multiclass` and they could be executed with hvd, gloo, nccl confs...

vfdev-5:
> Scikit-Learn has another data type called multilabel-multioutput which is like multilabel but its labels are not binary. It has yet to be implemented. We can do it if you will.

Let's make it in a separate PR.

ignite/metrics/precision.py
```diff
@@ -15,28 +15,39 @@ class _BasePrecisionRecall(_BaseClassification):
...
         is_multilabel: bool = False,
         device: Union[str, torch.device] = torch.device("cpu"),
     ):
 
+        if type(average) != bool and average not in ["micro", "weighted", "samples"]:
```
vfdev-5, line 23: How about doing this like that ?
```suggestion
        if not (isinstance(average, bool) or average in ["micro", "weighted", "samples"]):
```

ignite/metrics/precision.py
vfdev-5: So, we need to add "macro" option

ignite/metrics/precision.py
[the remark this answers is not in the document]
vfdev-5: I may be misunderstand your point, but I was thinking about new API like that:
```python
pr = Precision(average="samples")  # old: Precision(is_multilabel=True, average=True)
pr.update(multilabel_data)

pr = Precision()  # old: Precision()
pr.update(binary_data)

pr = Precision(average="samples")  # old: Precision(is_multilabel=True, average=True)
pr.update(binary_data)  # try to interpret binary_data as multilabel
```

> Anyway, without is_multilabel, we cannot differentiate binary data from multilabel data because we do not have a differentiator assumption on their dimensions. Scikit can, because it assumes binary data is 1-dimensional and multilabel data is 2-dimensional.

By the way, scikit learn has "binary" option as well. If we stuck with that, we can add it as well...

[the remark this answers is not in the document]
sadra-barikbin: Do you still think we can remove `is_multilabel`?
About `average='binary'`, Scikit uses it only for binary data. When average is binary, it returns the precision for class `pos_label` which is another parameter. So if `pos_label` is 1 (default), the simple precision is returned, and if the `pos_label` is 0, `TN/(TN+FN)` (precision for class 0) is returned.
I think we don't need this option. If user wants precision for class 0, he/she can simply do `1-y` and `1-y_pred` in the `output_transform`.
In our current implementation, when average is `True`, we return the simple precision (precision for class 1) for binary data.

ignite/metrics/precision.py
sadra-barikbin: Didn't we assign `average=True` to macro?

ignite/metrics/precision.py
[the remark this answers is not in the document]
vfdev-5: > Do you still think we can remove is_multilabel?

We do not remove it but just deprecate for removing later. For me if average can take str values, e.g. "samples", `is_multilabel` is redundant and can be deprecated. What do you think ?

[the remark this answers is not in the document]
sadra-barikbin: If we depracate `is_multilabel`, how could we differentiate binary from multilabel in the case average is anything but 'samples'?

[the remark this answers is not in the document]
vfdev-5: If average is not 'samples' then input data is not multilabel: binary or multiclass.

ignite/metrics/precision.py
```diff
@@ -15,28 +15,39 @@ class _BasePrecisionRecall(_BaseClassification):
     def __init__(
         self,
         output_transform: Callable = lambda x: x,
-        average: bool = False,
+        average: Union[bool, str] = False,
         is_multilabel: bool = False,
         device: Union[str, torch.device] = torch.device("cpu"),
     ):
 
+        if not (isinstance(average, bool) or average in ["micro", "weighted", "samples"]):
```
vfdev-5, line 23: When I was asking to add "macro", I wanted to add it to acceptable values `average in ["micro", "macro", "weighted", "samples"]`. I agree that `average=True` is the same as `average="macro"`. But it is not bad. We are gradually deprecating previous api to the new one with average as str similar to scikit-learn.
By the way, we have to add `average=None` as well which maps to `average=False`. 


What do you think about trying to match Scikit-learn API for average arg ?

ignite/metrics/precision.py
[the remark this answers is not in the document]
sadra-barikbin: Other options of average are also compatible with multilabel data, so if average is not 'samples', intput data can still be multilabel.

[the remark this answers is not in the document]
vfdev-5: > Other options of average are also compatible with multilabel data, so if average is not 'samples', intput data can still be multilabel.

Can you give a concrete example ? Even if it is so, is output still meaningful and can be compared to scikit learn ?

ignite/metrics/precision.py
sadra-barikbin: Useful for users familiar with Scikit. If so, shall I add `pos_label` as well? (for `average='binary'`) As I said above, `pos_label` is another parameter which is used to determine precision/recall of which class be returned.

vfdev-5: Yes, it can be useful, but let's do that in another PR, not here

sadra-barikbin:
The formulas I just added to docstrings are for multi label as well. As an example, suppose we have exam result of some students and  exam questions are YES/NO questions. Exam results is multilabel data. If we measure precision considering questions altogether, it would be micro precision. If we calculate  precision for each question then measure (weighted) average of them, it would be (weighted) macro precision. 

Yes. I adopted weighted, samples and micro from Scikit, and basically used it in tests as well.

ignite/metrics/precision.py
sadra-barikbin: On the other hand , it decreases our api's elegance.

ignite/metrics/precision.py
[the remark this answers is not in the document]
sadra-barikbin: The formulas I just added to docstrings are for multi label as well. As an example, suppose we have exam result of some students and  exam questions are YES/NO questions. Exam results is multilabel data. If we measure precision considering questions altogether, it would be micro precision. If we calculate  precision for each question then measure (weighted) average of them, it would be (weighted) macro precision.

Yes. I adopted weighted, samples and micro from Scikit, and basically used it in tests as well.

ignite/metrics/precision.py
```diff
@@ -15,28 +15,47 @@ class _BasePrecisionRecall(_BaseClassification):
...
+        if self._type in ["binary", "multiclass"] and self._average == "samples":
+            raise ValueError("Argument average='samples' is incompatible with binary and multiclass input data.")
+
     @reinit__is_reduced
     def reset(self) -> None:
-        self._true_positives = 0  # type: Union[int, torch.Tensor]
-        self._positives = 0  # type: Union[int, torch.Tensor]
-        self._updated = False
+        if self._average == "samples":
+            self._sum_samples_metric = 0  # type: Union[int, torch.Tensor]
```
vfdev-5, line 50: @sadra-barikbin you introduce here new accumulators: `_sum_samples_metric`, `_samples_cnt`, `_actual_positives` in addition to existing `_true_positives`, `_positives`. Can we just keep 3 instead, we can rename them to something abstract if names like `_true_positives` and `_positives` are inappropriate.

ignite/metrics/precision.py
```diff
@@ -250,31 +322,32 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
             indices = torch.argmax(y_pred, dim=1).view(-1)
             y_pred = to_onehot(indices, num_classes=num_classes)
         elif self._type == "multilabel":
-            # if y, y_pred shape is (N, C, ...) -> (C, N x ...)
-            num_classes = y_pred.size(1)
-            y_pred = torch.transpose(y_pred, 1, 0).reshape(num_classes, -1)
-            y = torch.transpose(y, 1, 0).reshape(num_classes, -1)
+
```
vfdev-5, line 325: Can you keep and update the comment you removed in this line

ignite/metrics/fbeta.py
```diff
@@ -56,8 +56,8 @@ def Fbeta(
 
         .. testcode:: 1
 
-            P = Precision(average=False)
-            R = Recall(average=False)
+            P = Precision(average=False)[1] # `[1]` is to select the value for class 1 that is TP/(TP+FP)
+            R = Recall(average=False)[1] # `[1]` is to select the value for class 1 that is TP/(TP+FN)
```
vfdev-5, line 60: @sadra-barikbin What we changed exactly in this PR (please provide the line of code) such that we have to update this example. I'm trying to understand if we are doing BC-breaking update or not

ignite/metrics/precision.py
```diff
@@ -46,24 +65,33 @@ def compute(self) -> Union[torch.Tensor, float]:
...
-        if self._average:
-            return cast(torch.Tensor, result).mean().item()
+        result = self._true_positives / (self._positives + self.eps)
+        if self._average == "weighted":
+            denominator = self._actual_positives.sum() + self.eps  # type: ignore
+            return ((result @ self._actual_positives) / denominator).item()  # type: ignore
+        elif self._average == "micro":
+            return result.item()  # type: ignore
+        elif self._average == "macro":
+            return result.mean().item()  # type: ignore
```
vfdev-5, line 88: Why we have to `type: ignore` ?

ignite/metrics/precision.py
```diff
@@ -93,153 +179,139 @@ class Precision(_BasePrecisionRecall):
...
 
         .. testcode:: 1
 
-            metric = Precision(average=False)
+            metric = Precision()
+            weighted_metric = Precision(average='weighted')
             metric.attach(default_evaluator, "precision")
-            y_true = torch.Tensor([1, 0, 1, 1, 0, 1])
+            weighted_metric.attach(default_evaluator, "weighted precision")
+            y_true = torch.Tensor([1, 0, 1, 1, 0, 1]).long()
```
vfdev-5, line 191: @sadra-barikbin please can you check and replace everywhere in the docstrings (Precision, Recall) that we create tensors like `torch.tensor` and not `torch.Tensor`. Thanks !

ignite/metrics/precision.py
```diff
@@ -250,31 +322,32 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
...
             y_pred = to_onehot(indices, num_classes=num_classes)
         elif self._type == "multilabel":
-            # if y, y_pred shape is (N, C, ...) -> (C, N x ...)
-            num_classes = y_pred.size(1)
-            y_pred = torch.transpose(y_pred, 1, 0).reshape(num_classes, -1)
-            y = torch.transpose(y, 1, 0).reshape(num_classes, -1)
+
+            num_labels = y_pred.size(1)
+            y_pred = torch.transpose(y_pred, 1, -1).reshape(-1, num_labels)
+            y = torch.transpose(y, 1, -1).reshape(-1, num_labels)
```
vfdev-5, line 328: Why it is better to make it (N * ..., C) instead of (C, N * ...) ?

ignite/metrics/recall.py
```diff
@@ -40,151 +101,127 @@ class Recall(_BasePrecisionRecall):
...
     def update(self, output: Sequence[torch.Tensor]) -> None:
         self._check_shape(output)
         self._check_type(output)
         y_pred, y = output[0].detach(), output[1].detach()
 
         if self._type == "binary":
-            y_pred = y_pred.view(-1)
-            y = y.view(-1)
+
+            y = to_onehot(y.view(-1), num_classes=2)
```
vfdev-5, line 221: Do you think we can refactor this code to avoid code duplication ?

tests/ignite/metrics/test_metrics_lambda.py
```diff
@@ -149,7 +149,7 @@ def test_metrics_lambda_update():
     recall = Recall(average=False)
 
     def Fbeta(r, p, beta):
-        return torch.mean((1 + beta ** 2) * p * r / (beta ** 2 * p + r)).item()
+        return (1 + beta ** 2) * p * r / (beta ** 2 * p + r)
```
vfdev-5, line 152: Why this change ?

ignite/metrics/precision.py
```diff
@@ -93,153 +179,139 @@ class Precision(_BasePrecisionRecall):
...
     def update(self, output: Sequence[torch.Tensor]) -> None:
         self._check_shape(output)
         self._check_type(output)
         y_pred, y = output[0].detach(), output[1].detach()
 
         if self._type == "binary":
-            y_pred = y_pred.view(-1)
-            y = y.view(-1)
+
+            y = to_onehot(y.view(-1), num_classes=2)
```
vfdev-5, line 311: OK, I see, you pass binary case into multiclass (2-classes) case. I think we have to keep computed result for the single class (=1) instead of both. Our public API and usage should not change and this PR shouldn't be BC-breaking for this case.

ignite/metrics/precision.py
sadra-barikbin: Yes. For example `nominator`, `denominator` and `weight`. It hurts readability, though. Not only the type but also the semantic of `_sum_samples_metric` is different with  that of `_true_positives` or `_positives`. The former is a float value representing the summation of computed metric over samples and the latter is the number of samples.

tests/ignite/metrics/test_metrics_lambda.py
sadra-barikbin: As we agreed, `average=False` results in precision and recall becoming an array with length equal to number of classes. Had I not changed  that line, it would have returned the average of Fbeta for class zero and Fbeta for class one. Now it returns Fbeta array and I pick Fbeta for class one in line 215

ignite/metrics/fbeta.py
sadra-barikbin: https://github.com/sadra-barikbin/ignite/blob/214088b94c21471e4e287cefc411bc2f45bab3bb/ignite/metrics/precision.py#L311-L312

sadra-barikbin: Alternatively I can do this:
```python
P = Precision(average=False)
R = Recall(average=False)
metric = Fbeta(beta=1.0, precision=P, recall=R, average=False)[1]
```

sadra-barikbin: This change is because beforehand, Precision or recall with `average=False` returned a single float on binary input, but now they return an array with length 2.

ignite/metrics/precision.py
sadra-barikbin: Because we have
```python
result = self._true_positives / (self._positives + self.eps)
```
and `_true_positives` and `_positives` are of typehint `Union[int, torch.Tensor]` . So `result`'s type would implicitly be `Union[float, torch.Tensor]`. In line 88 we call `mean()` on `result`. On the other hand each feasible type of `result` should have `mean` method, but float has not. So MyPy raises error if I do not add the hint

ignite/metrics/precision.py
sadra-barikbin: Because in binary and multilabel case, y and y_pred finally become of size N*C. I did the same thing to multilabel to reuse the logic in following lines.

ignite/metrics/precision.py
sadra-barikbin: I added comment to `reset` and `compute` functions which are also common between precision and recall. Is it ok?

tests/ignite/metrics/test_metrics_lambda.py
vfdev-5: > As we agreed, average=False results in precision and recall becoming an array with length equal to number of classes.

It was the same behaviour previously as well. That's why `torch.mean(...).item()` was applied to compute an average. I think we should keep tests without much changes to ensure BC.

tests/ignite/metrics/test_metrics_lambda.py
```diff
@@ -195,24 +195,24 @@ def Fbeta(r, p, beta):
...
+    assert recall._nominator[1] == true_positives
+    assert recall._denominator[1] == positives
 
     """
     Test compute
     """
     F1.reset()
     F1.update((y_pred1, y1))
-    F1_metrics_lambda = F1.compute()
+    F1_metrics_lambda = F1.compute()[1]
```
vfdev-5, line 215: Following previous comment above. Let's revert this and above changes.

tests/ignite/metrics/test_metrics_lambda.py
```diff
@@ -400,8 +400,8 @@ def data(y_pred, y):
         np_y = y.numpy().ravel()
         assert state.metrics[metric_name] == approx(compute_true_value_fn(np_y_pred, np_y))
 
-    precision_1 = Precision()
-    precision_2 = Precision()
+    precision_1 = Precision()[1]
```
vfdev-5, line 403: Same here, I'd prefer to revert these changes as well

ignite/metrics/precision.py
```diff
@@ -15,55 +15,111 @@ class _BasePrecisionRecall(_BaseClassification):
...
-                self._true_positives = cast(torch.Tensor, idist.all_gather(self._true_positives))
-                self._positives = cast(torch.Tensor, idist.all_gather(self._positives))
+            self._nominator = idist.all_reduce(self._nominator)  # type: ignore[assignment]
+            self._denominator = idist.all_reduce(self._denominator)  # type: ignore[assignment]
+            if self._average == "weighted":
+                self._weight = idist.all_reduce(self._weight)  # type: ignore[assignment]
             self._is_reduced = True  # type: bool
 
-        result = self._true_positives / (self._positives + self.eps)
+        fraction = self._nominator / (self._denominator + (self.eps if self._average != "samples" else 0))
```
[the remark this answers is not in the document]
sadra-barikbin: You mean to use eps for average=samples as well?

vfdev-5 (approved):
Thanks a lot for this great PR @sadra-barikbin !
I'm checking it on GPU and if CI is green let's merge it.
In a follow-up PR, let's clean-up docstring about 
```diff
- torch.Tensor(...)
+ torch.tensor(...)
```

PR #2574, opened 2022-05-19
add NoImprovementHandler

### LINKED ISSUE #2314: Add no improvement handler (similar to early stopping handler)
## 🚀 Feature

To add a 'no-improvement' handler. It's similar to early stopping, except that early stopping calls the training to terminate, 'no-improvement' handler would've calls a function that user pass in. User could pass in different functions that they want to execute if they find no improvements is made. They may want to try various things before terminate the training.

Ishan-Kumar2:
Fixes #2314 

Description:
As described in the issue, added a generalized version of EarlyStopping where the `stop_function` and `pass_function` can be user-defined. 
@sdesrozis let me know if the approach is correct, I'll start working on the tests.

Check list:

### DIFF
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -98,7 +98,7 @@ class NoImprovementHandler(Serializable):
         else:
             self.pass_function(self.trainer)
 
-    def _update_state(self, score):
+    def _update_state(self, score: int) -> None:
 
         if self.best_score is None:
             self.best_score = score
@@ -177,12 +177,12 @@ class EarlyStopping(NoImprovementHandler):
 
         self.logger = setup_logger(__name__ + "." + self.__class__.__name__)
 
-    def __call__(self, engine):
+    def __call__(self, engine: Engine) -> None:
         super(EarlyStopping, self).__call__(engine)
 
-    def pass_function(self, trainer):
+    def pass_function(self, trainer: Engine) -> None:
         pass
 
-    def stop_function(self, trainer):
+    def stop_function(self, trainer: Engine) -> None:
         self.logger.info("EarlyStopping: Stop training")
         trainer.terminate()
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -93,6 +93,7 @@ class NoImprovementHandler(Serializable):
         self.best_score = None  # type: Optional[float]
         self.min_delta = min_delta
         self.cumulative_delta = cumulative_delta
+        self.logger = setup_logger(__name__ + "." + self.__class__.__name__)
 
     def __call__(self, engine: Engine) -> None:
         score = self.score_function(engine)
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -5,47 +5,57 @@ from ignite.base import Serializable
 from ignite.engine import Engine
 from ignite.utils import setup_logger
 
-__all__ = ["EarlyStopping"]
+__all__ = ["NoImprovementHandler", "EarlyStopping"]
 
 
-class EarlyStopping(Serializable):
-    """EarlyStopping handler can be used to stop the training if no improvement after a given number of events.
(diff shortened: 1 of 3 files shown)


ignite/handlers/early_stopping.py
```diff
@@ -54,37 +64,50 @@ def __init__(
...
         self.patience = patience
-        self.min_delta = min_delta
-        self.cumulative_delta = cumulative_delta
+        self.score_function = score_function
+        self.pass_function = pass_function
+        self.stop_function = stop_function
         self.trainer = trainer
         self.counter = 0
         self.best_score = None  # type: Optional[float]
-        self.logger = setup_logger(__name__ + "." + self.__class__.__name__)
```
vfdev-5, line 73: We may want also the logger for `NoImprovementHandler` ?

ignite/handlers/early_stopping.py
```diff
@@ -5,47 +5,57 @@
...
+            h = NoImprovementHandler(patience=3, score_function=score_function, pass_function=pass_function,
+                                     stop_function=stop_function, trainer=trainer)
 
-            handler = EarlyStopping(patience=10, score_function=score_function, trainer=trainer)
             # Note: the handler is attached to an *Evaluator* (runs one epoch on validation dataset).
             evaluator.add_event_handler(Events.COMPLETED, handler)
-
     """
 
-    _state_dict_all_req_keys = (
```
vfdev-5, line 40: If we want to save NoImprovementHandler's internal state we have to keep `_state_dict_all_req_keys`, IMO

vfdev-5 (commented):
Thanks for the PR @Ishan-Kumar2 !
I left few minor comments to address.

ignite/handlers/early_stopping.py
```diff
@@ -54,37 +64,50 @@ def __init__(
...
         self.counter = 0
         self.best_score = None  # type: Optional[float]
-        self.logger = setup_logger(__name__ + "." + self.__class__.__name__)
+        self.min_delta = min_delta
+        self.cumulative_delta = cumulative_delta
 
     def __call__(self, engine: Engine) -> None:
         score = self.score_function(engine)
 
+        self._update_state(score)
```
sdesrozis, line 95: If we would have liked been even more general, the test would have been wrapped in a user function too. Anyway, that’s a detail.

sdesrozis:
@Ishan-Kumar2 It looks very good, thanks ! Please go ahead with the tests !

ignite/handlers/early_stopping.py
Ishan-Kumar2: I was also considering that, but don't you think that it should be fixed since it's the No Improvement handler? 
The condition that there have been no improvements to the score in the last k epochs remains constant right otherwise it may become a little complicated.
Let me know what you think :)

Ishan-Kumar2:
@sdesrozis @vfdev-5 I have added the tests, a lot of them are similar to the EarlyStopping tests and might not be even needed since they both use the NoImprovementHandler functions. For instance, `test_state_dict_no_improvement_handler` and `test_state_dict` both use the same function, and having both may be redundant, should I remove one in this case?

ignite/handlers/early_stopping.py
sdesrozis: Let's do it like that :)

Ishan-Kumar2:
@sdesrozis I was thinking about the redundancies, while it's true that they both call the same function do you think it is still worth having retaining the `test_state_dict` as it ensures that the `EarlyStopping` is inheriting it correctly and it works for its set of stop and pass functions, just as a precaution. The same goes for the others. What do you think?

sdesrozis:
@Ishan-Kumar2 the tests are both useful especially if we consider the evolution of the library. I was thinking about having an unique and generic test code for the both. It would imply introducing an abstraction for the test result. As the tests are quite small and clear, I'm finally not sure about that. So let's keep it simple.

I will review asap. Thanks again.

PR #2592, opened 2022-06-06
Add a `handle_buffers` option for EMAHandler

### LINKED ISSUE #2590: EMAHandler (Buffer update) does lead to bad results for Mean-Teacher experiments
## 🚀 Feature

Hey there,
thanks for the increased effort in Ignite, the package just gets better by each update!

So my feature request ( or bug report ) is concerned with the recently developed [EMAHAndler](https://pytorch.org/ignite/_modules/ignite/handlers/ema_handler.html#EMAHandler).

Specifically I recently ran a lot of experiments using Ignite for Sound event classification and noticed that my results far underperform ( F1 Score of ~40 with Ignite and ~60 without ) models that do not use the included EMAHandler class.
I investigated this issue and came to the conclusion that the implementation of Ignites EMAHandler has some minor issues which lead to this behaviour.

Specifically, the lines responsible are:

https://github.com/pytorch/ignite/blob/80c6bb4c9b7b8bacfa5f39557c16065267a849ce/ignite/handlers/ema_handler.py#L174

and

https://github.com/pytorch/ignite/blob/80c6bb4c9b7b8bacfa5f39557c16065267a849ce/ignite/handlers/ema_handler.py#L182

## First problem, evaluation mode

First, I'd like to discuss the problem with ```self.ema_model.eval()```.
While this default behavior makes sense for doing EMA updates, it leads to some major problems regarding non-parameter (buffer) updates of the EMA model.
A particular example is the use of BatchNorm, which will not update its mean and variance at all during EMA training.
However BatchNorm **will** update its affine transformations during training, which might lead to a discrepancy between the EMA model and the original (student) model.

## Second problem, updating buffers

The second problem is directly connected to the first one.
Since the non-parameter updates of the EMA model are in-fact frozen in the Ignite implementation, the `buffer` variables are therefore simply copied from the Student (Original model) to the EMAModel using:

```python
        for ema_b, model_b in zip(self.ema_model.buffers(), self.model.buffers()):
            ema_b.data = model_b.data
```

While again this makes perfect sense, for variables such as `num_tracked_batches` in a BatchNorm layer, it does not make sense for their corresponding means and variances.
For example, after some EMA updates, the weights of the EMA model might not be similar to the ones of the student model. But the current implementation **forces** the EMA model to use the exact same mean's and variances from the student model, which might not be the "True" means and variances.
This then will likely lead to a problem when training for longer, since the EMA model cannot properly normalize the input data and thus somewhat diverges.

## Solutions

I suggest two solutions:

1. (Not recommended ) Also average floating point buffers, such as means and variances.

```python
for ema_b, model_b in zip(self.ema_model.buffers(),                                                                                                                                  
                          self.model.buffers()):                                                                                                                                     
    #Update buffers like running mean and variance                                                                                                                          
    if torch.is_floating_point(ema_b):                                                                                                                                      
        ema_b.mul_(1.0 - momentum).add_(model_b.data, alpha=momentum)                                                                                                       
    else:                                                                                                                                                                   
        ema_b.data = model_b.data
```

I tried this solution already in my experiments and at least from a "loss" perspective, it does outperform the standard EMA model by a large margin.
However the final results were still worse than the ones proposed in 2.:

2. Do not force `.eval()` mode and remove the buffer updates.

I'd recommend currently to follow the implementation of most mean-teacher approaches, which actually require the EMA model to be in `train()` mode, thus updating the batchnorm buffers as well as adding some noise (dropout) during training.
Thus, I'd recommend just removing the forced `self.ema_model.eval()` statement and removing the buffer updates.

What is your point of view about this issue? Did anybody maybe also encounter a similar problem?

.

sandylaker:
Fixes #2590 

Description:

Add a `handle_buffers` option for EMAHandler. 

cc @vfdev-5 @RicherMans

Check list:

### DIFF
--- a/ignite/handlers/ema_handler.py
+++ b/ignite/handlers/ema_handler.py
@@ -42,12 +42,14 @@ class EMAHandler:
           momentum: the update momentum after warmup phase, should be float in range :math:`\left(0, 1 \right)`.
           momentum_warmup: the initial update momentum during warmup phase.
           warmup_iters: iterations of warmup.
+          use_buffers: if True, compute the running averages for both the parameters and the buffers of the model.
 
     Attributes:
           ema_model: the exponential moving averaged model.
           model: the online model that is tracked by EMAHandler. It is ``model.module`` if ``model`` in
               the initialization method is an instance of ``DistributedDataParallel``.
           momentum: the update momentum.
+          use_buffers: if True, compute the running averages for both the parameters and the buffers of the model.
 
     Note:
           The EMA model is already in ``eval`` mode. If model in the arguments is an ``nn.Module`` or
@@ -149,6 +151,7 @@ class EMAHandler:
         momentum: float = 0.0002,
         momentum_warmup: Optional[float] = None,
         warmup_iters: Optional[int] = None,
+        use_buffers: bool = False,
     ) -> None:
         if not 0 < momentum < 1:
             raise ValueError(f"Invalid momentum: {momentum}")
@@ -173,14 +176,27 @@ class EMAHandler:
             param.detach_()
         self.ema_model.eval()
 
+        self.use_buffers = use_buffers
+
     def _update_ema_model(self, engine: Engine, name: str) -> None:
         """Update weights of ema model"""
         momentum = getattr(engine.state, name)
         for ema_p, model_p in zip(self.ema_model.parameters(), self.model.parameters()):
             ema_p.mul_(1.0 - momentum).add_(model_p.data, alpha=momentum)
-        # assign the buffers
-        for ema_b, model_b in zip(self.ema_model.buffers(), self.model.buffers()):
-            ema_b.data = model_b.data
+
+        if self.use_buffers:
+            for ema_b, model_b in zip(self.ema_model.buffers(), self.model.buffers()):
+                try:
+                    ema_b.mul_(1.0 - momentum).add_(model_b.data, alpha=momentum)
+                except RuntimeError:
+                    # Handle the case where ema_b is torch.int64, torch.int32 etc.,
+                    # where a runtime error will be thrown when performing the in-place operations with floats.
+                    # In this case, just copy the data
+                    ema_b.data = model_b.data
+        else:
+            # assign the buffers
+            for ema_b, model_b in zip(self.ema_model.buffers(), self.model.buffers()):
+                ema_b.data = model_b.data
 
     def attach(
         self,
--- a/ignite/handlers/ema_handler.py
+++ b/ignite/handlers/ema_handler.py
@@ -151,7 +151,7 @@ class EMAHandler:
         momentum: float = 0.0002,
         momentum_warmup: Optional[float] = None,
         warmup_iters: Optional[int] = None,
-        use_buffers: bool = False,
+        use_buffers: bool = True,
     ) -> None:
         if not 0 < momentum < 1:
             raise ValueError(f"Invalid momentum: {momentum}")
--- a/ignite/handlers/ema_handler.py
+++ b/ignite/handlers/ema_handler.py
@@ -42,19 +42,22 @@ class EMAHandler:
           momentum: the update momentum after warmup phase, should be float in range :math:`\left(0, 1 \right)`.
           momentum_warmup: the initial update momentum during warmup phase.
           warmup_iters: iterations of warmup.
-          use_buffers: if True, compute the running averages for both the parameters and the buffers of the model.
+          handle_buffers: how to handle model buffers during training. There are three options: 1. "copy" means
+              copying the buffers of the online model; 2. "update" means applying EMA to the buffers of the online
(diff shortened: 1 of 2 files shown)


vfdev-5:
@sandylaker thanks for a quick PR! 

As for the problem about integral params in buffers, does it mean that in pytorch they should here a similar issue here: https://github.com/pytorch/pytorch/blob/080cf84bed46c6c118c37fc2fa6fbd484fd9b4cd/torch/optim/swa_utils.py#L124-L132 ?

RicherMans:
@vfdev-5 

> As for the problem about integral params in buffers, does it mean that in pytorch they should here a similar issue here: https://github.com/pytorch/pytorch/blob/080cf84bed46c6c118c37fc2fa6fbd484fd9b4cd/torch/optim/swa_utils.py#L124-L132 ?

Yeah they should, but the default usage for the SWA model is to update the batchnorm after training and not during via https://github.com/pytorch/pytorch/blob/080cf84bed46c6c118c37fc2fa6fbd484fd9b4cd/torch/optim/swa_utils.py#L136 .

Also as I would hypothesize, that most code and examples usually set the model to .train() before transfer, thus the problem is less likely to occur.
Also (at least for me), its not that the model itself is throwing exceptions or is completely "useless", but rather that it just converges to a less optimal solution due to the "wrong" means and variances within the batch-norm.

Actually, I once trained a common Mean-teacher recipe and checked the absolute difference between the online (student) model and the EMA model.
For a lower layer like the first batch-norm, the batch-norm differences for mean/variance are neglectable and usually in range < 1e-1
But for higher layers, the differences get very noticeable at about ~ += 2 for means and += 200 for variance.

vfdev-5:
@sandylaker if you check again "Solutions" of https://github.com/pytorch/ignite/issues/2590 it looks like that RicherMans already tried to update floating buffers and copy integers (see his code snippet).
Looks like 2nd option in the solutions is better than 1st.

For this PR I think we could add an arg with 3 options: 1) "copy" = to keep current (copy) behaviour, 2) "update_buffers" = update buffers according to Richer's option 1 and 3) "ema_train" = Richer's option 2

What do you think ?

RicherMans:
@sandylaker 

> In the current implementation, the running_mean and running_var in a _BatchNorm will be updated with momentum, just like other params. But the num_batches_tracked will be directly copied. Do you think that the performence degradation issue will still present? Could you please test it in your codebase? Thanks a lot.

So I actually did already check during my investigations all the proposed methods. So to be more precise for my experiments, I do sound event classification for the [DCASE 2022 ](https://github.com/DCASE-REPO/DESED_task/tree/ffb3d49feb740da789fff80c082bdde6b98e7cb4/recipes/dcase2022_task4_baseline) task, where the baseline uses a mean-teacher approach that I have reimplemented using EMAHandler.
For the baseline, they report two values (PSDS-1 and PSDS-2 scores) for the experiment. Higher is better.

The results are as follows:

| Model | PSDS1  | PSDS2 |
| ------------- | ------------- | ------------- |
| Baseline| 33.6  |  53.60 |
| EMAHAndler v0.4.9 (Old) | 16.95 | 30.86 |
| `use_buffers=True` (New) |  27.53 | 43.99  |
| set model to .train() and do not update buffers (suggestion 2) | 33.9 |  54.84 |

So as I originally stated, the momentum update of buffers such as the mean and variance works obviously better than the simple copying mechanism.
However, all results pale in comparison to just simply let the model itself compute it's mean and variance.

Of course this is a mean-teacher learning case, and my results might not be representative for other cases where one might want to "freeze" the buffers.

What do you guys think?

vfdev-5:
@RicherMans thanks for sharing details !

I think we could provide all 3 options such that depending on a use-case user could pick appropriate configuration. We have to ensure that the 3rd option `ema_model.train()` does not break some assumptions in the flow.

vfdev-5:
@sandylaker can we move forward with this PR and implement 3 options described in https://github.com/pytorch/ignite/pull/2592#issuecomment-1147317265

sandylaker:
@vfdev-5 Hi. Sorry for the late reply, I have been busy during the working days. I will implement it this weekend.

vfdev-5 (approved):
Looks great, thanks a lot @sandylaker !

vfdev-5:
@RicherMans could you please check this PR if it solves the issue you have

vfdev-5:
I merged this PR and if we need more changes to it, we can do that it in a follow-up PR.
Thanks again @sandylaker !

RicherMans:
Sorry @vfdev-5 for my late response.
Yeah that commit looks pretty good and I tested it, looks alright!

Thanks guys for the good work! ( Even though I hope I can also one time commit myself )

vfdev-5:
> Even though I hope I can also one time commit myself

That would be awesome ! By the way, if you need some help to start with that, feel free to join our discord: https://pytorch-ignite.ai/chat

PR #2620, opened 2022-07-19
Parametrized tests for tests/ignite/metrics/test_accumulation

### LINKED ISSUE #2522: Improve tests using pytest parametrize

kamalojasv181:
Related #2522

Description: Parametrized tests for tests/ignite/metrics/test_accumulation using pytest.

### DIFF
--- a/tests/ignite/metrics/test_accumulation.py
+++ b/tests/ignite/metrics/test_accumulation.py
@@ -143,12 +143,10 @@ def test_integration(metric_cls, true_result_fn, shape):
     custom_variable = 10.0 + 5.0 * torch.rand(shape)
 
     def update_fn(engine, batch):
-        return (
-            0,
-            custom_variable[engine.state.iteration - 1]
-            if custom_variable[engine.state.iteration - 1].shape != torch.Size([])
-            else custom_variable[engine.state.iteration - 1].item(),
-        )
+
+        output = custom_variable[engine.state.iteration - 1]
+        output = output.item() if output.ndimension() < 1 else output
+        return 0, output
 
     engine = Engine(update_fn)
 
@@ -157,12 +155,9 @@ def test_integration(metric_cls, true_result_fn, shape):
 
     state = engine.run([0] * shape[0])
 
-    if len(shape) > 1:
-        np.testing.assert_almost_equal(
-            state.metrics["agg_custom_var"].numpy(), true_result_fn(custom_variable), decimal=5
-        )
-    else:
-        assert state.metrics["agg_custom_var"] == pytest.approx(true_result_fn(custom_variable))
+    np.testing.assert_almost_equal(
+        np.array(state.metrics["agg_custom_var"]), true_result_fn(custom_variable), decimal=5
+    )
 
 
 def test_compute_mean_std():
--- a/tests/ignite/metrics/test_accumulation.py
+++ b/tests/ignite/metrics/test_accumulation.py
@@ -133,36 +133,36 @@ def test_geom_average():
     np.testing.assert_almost_equal(m.numpy(), _geom_mean(y_true.reshape(-1, 10)), decimal=5)
 
 
-@pytest.mark.parametrize("metric_cls, true_result_fn", [(Average, _mean), (GeometricAverage, _geom_mean)])
-def test_integration(metric_cls, true_result_fn):
+@pytest.mark.parametrize(
+    "metric_cls, true_result_fn, shape", [(Average, _mean, [100, 12]), (GeometricAverage, _geom_mean, [100])]
+)
+def test_integration(metric_cls, true_result_fn, shape):
 
-    size = 100
-    custom_variable = 10.0 + 5.0 * torch.rand(size, 12)
+    assert len(shape) > 0 and len(shape) < 3
+
+    custom_variable = 10.0 + 5.0 * torch.rand(shape)
 
     def update_fn(engine, batch):
-        return 0, custom_variable[engine.state.iteration - 1]
+        return (
+            0,
+            custom_variable[engine.state.iteration - 1]
+            if custom_variable[engine.state.iteration - 1].shape != torch.Size([])
+            else custom_variable[engine.state.iteration - 1].item(),
+        )
 
     engine = Engine(update_fn)
 
     custom_var_mean = metric_cls(output_transform=lambda output: output[1])
     custom_var_mean.attach(engine, "agg_custom_var")
 
-    state = engine.run([0] * size)
-    np.testing.assert_almost_equal(state.metrics["agg_custom_var"].numpy(), true_result_fn(custom_variable), decimal=5)
+    state = engine.run([0] * shape[0])
 
-    size = 100
-    custom_variable = 10.0 + 5.0 * torch.rand(size)
-
-    def update_fn(engine, batch):
-        return 0, custom_variable[engine.state.iteration - 1].item()
-
-    engine = Engine(update_fn)
-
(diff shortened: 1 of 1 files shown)


vfdev-5:
@kamalojasv181, please change `Fixes` -> `Related`

tests/ignite/metrics/test_accumulation.py
```diff
@@ -129,44 +133,36 @@ def test_geom_average():
...
 
-def test_integration():
-    def _test(metric_cls, true_result_fn):
+@pytest.mark.parametrize("metric_cls, true_result_fn", [(Average, _mean), (GeometricAverage, _geom_mean)])
+def test_integration(metric_cls, true_result_fn):
 
-        size = 100
-        custom_variable = 10.0 + 5.0 * torch.rand(size, 12)
+    size = 100
+    custom_variable = 10.0 + 5.0 * torch.rand(size, 12)
```
vfdev-5, line 140: @kamalojasv181 can you also parametrize `custom_variable` shape ?

tests/ignite/metrics/test_accumulation.py
```diff
@@ -129,44 +133,36 @@ def test_geom_average():
...
+    np.testing.assert_almost_equal(state.metrics["agg_custom_var"].numpy(), true_result_fn(custom_variable), decimal=5)
 
-        engine = Engine(update_fn)
+    size = 100
+    custom_variable = 10.0 + 5.0 * torch.rand(size)
 
-        custom_var_mean = metric_cls(output_transform=lambda output: output[1])
-        custom_var_mean.attach(engine, "agg_custom_var")
+    def update_fn(engine, batch):
+        return 0, custom_variable[engine.state.iteration - 1].item()
```
vfdev-5, line 157: Here you can conditionally apply `.item()` if `custom_variable[engine.state.iteration - 1]` is a scalar

vfdev-5 (commented):
Thanks for the PR @kamalojasv181 
Please address my comments to improve it further

tests/ignite/metrics/test_accumulation.py
kamalojasv181: Sure but won't this always be a scalar? Since we only defined the test?

vfdev-5: I think no as there are 2 types of input we test on

kamalojasv181: Oh so you want me to handle both cases with one function

vfdev-5: Yes, I want to handle input size as a parameter, https://github.com/pytorch/ignite/pull/2620#discussion_r924135827

vfdev-5: The point is to reduce the number of lines of code. We have 4 lines of duplicated code: 
```python
    size = 100  # duplicated
    custom_variable = 10.0 + 5.0 * torch.rand(size)

    def update_fn(engine, batch):
        return 0, custom_variable[engine.state.iteration - 1].item()

    engine = Engine(update_fn)  # duplicated

    custom_var_mean = metric_cls(output_transform=lambda output: output[1])  # duplicated
    custom_var_mean.attach(engine, "agg_custom_var")  # duplicated

    state = engine.run([0] * size)  # duplicated
    assert state.metrics["agg_custom_var"] == pytest.approx(true_result_fn(custom_variable))
```

kamalojasv181: Also one more thing. device is defied separately as in every function 
image
except
image.

We have already put a check in this function to execute this only when we have GPU. So can I just declare device as `device = idist.device()` globally? This will help us parameterize a lot of functions.

vfdev-5: Test functions where we define `device = idist.device()` are specific for distributed tests. We can talk on discord about them and how to improve these tests.
I do not think we can define `device` globally as we may alter test setup...

tests/ignite/metrics/test_accumulation.py
```diff
@@ -129,45 +133,37 @@ def test_geom_average():
...
-        custom_var_mean = metric_cls(output_transform=lambda output: output[1])
-        custom_var_mean.attach(engine, "agg_custom_var")
+    state = engine.run([0] * shape[0])
 
-        state = engine.run([0] * size)
+    if len(shape) > 1:
+        np.testing.assert_almost_equal(
+            state.metrics["agg_custom_var"].numpy(), true_result_fn(custom_variable), decimal=5
+        )
+    else:
```
vfdev-5, line 164: Can we use `np.testing.assert_almost_equal` for both tests ?
```suggestion
    np.testing.assert_almost_equal(
        state.metrics["agg_custom_var"].numpy(), true_result_fn(custom_variable), decimal=5
    )
```
If not let's try to update the outputs to cover both cases with `np.testing.assert_almost_equal`

vfdev-5 (commented):
Thanks for the update @kamalojasv181

tests/ignite/metrics/test_accumulation.py
```diff
@@ -129,45 +133,37 @@ def test_geom_average():
...
-    def _test(metric_cls, true_result_fn):
-
-        size = 100
-        custom_variable = 10.0 + 5.0 * torch.rand(size, 12)
-
-        def update_fn(engine, batch):
-            return 0, custom_variable[engine.state.iteration - 1]
+@pytest.mark.parametrize(
+    "metric_cls, true_result_fn, shape", [(Average, _mean, [100, 12]), (GeometricAverage, _geom_mean, [100])]
+)
```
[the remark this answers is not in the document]
kamalojasv181: But I was asked to parametrize size in the conversation above.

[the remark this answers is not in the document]
vfdev-5: Sorry I was not clear with the previous request, but I meant this kind of parametrization. This way we keep same number of tests.

vfdev-5 (approved):
LGTM, thanks @kamalojasv181 !

PR #2633, opened 2022-07-27
Add rank param to Checkpoint

### LINKED ISSUE #2623: Consolidate ZeRO state before checkpoint saving
Is your feature request related to a problem? Please describe.
When checkpoint saving occurs there should be a check if the object that state_dict() is called on is a ZeroRedundancyOptimizer instance

Describe the solution you'd like
Prior to the call state_dict() call a consolidate_state_dict() call should be issued. This call needs to be issued on all ranks and point toward the same consolidating rank. 

Here there are two design solutions, you instantiate a Checkpoint handler on all ranks and only on the designated rank does it save the checkpoint or do you create a Handler that needs to run before the Checkpoint handler in order to consolidate.

Special care must be addressed to the PyTorch version as the naming scheme of the ZeRO's method arguments has changed between recent versions.

Describe alternatives you've considered
The only alternative is for the users to write themselves Handlers to do that, segregating the checkpoint-saving logic. And it can be written as:

```
class ConsolidateZeROHandler:
    """
    Handler that consolidated the Zero Redundancy Optimizer prior to the checkpoint Saving.

    Args:
        zero_optimizer (ZeroRedundancyOptimizer): The optimizer to be consolidated.
        recipient_rank (int): The rank on which the consolidation will happen. Defaults to 0.
        epoch_level (bool): Call every N epochs or every N iterations. `True` is epoch level, `False` is iteration
            level. Defaults to True.
    """

    def __init__(
        self,
        zero_optimizer: ZeroRedundancyOptimizer,
        call_every: int,
        recipient_rank: int = 0,
        epoch_level=True,
    ):
        self.zero_optimizer = zero_optimizer
        self.recipient_rank = recipient_rank
        self.epoch_level = epoch_level
        self.call_every = call_every

    def __call__(self, engine: Engine):
        self.zero_optimizer.consolidate_state_dict(to=self.recipient_rank)

        if self.epoch_level:
            engine.add_event_handler(Events.EPOCH_COMPLETED(every=self.call_every), self)
        else:
            engine.add_event_handler(Events.ITERATION_COMPLETED(every=self.call_every), self)
```

sadra-barikbin:
Fixes #2623 

Description:

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -172,7 +172,7 @@ class Checkpoint(Serializable):
 
     .. warning::
 
-        When running on XLA devices or using :class:`~torch.distributed.ZeroRedundancyOptimizer`, it
+        When running on XLA devices or using :class:`~torch.distributed.optim.ZeroRedundancyOptimizer`, it
         should be run in all processes, otherwise application can get stuck on saving the checkpoint.
 
         .. code-block:: python
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -7,10 +7,11 @@ import warnings
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
 from pathlib import Path
-from typing import Any, Callable, Dict, IO, List, Mapping, NamedTuple, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Mapping, NamedTuple, Optional, Tuple, Union
 
 import torch
 import torch.nn as nn
+from torch.distributed.optim import ZeroRedundancyOptimizer
 
 import ignite.distributed as idist
 from ignite.base import Serializable
@@ -101,6 +102,8 @@ class Checkpoint(Serializable):
             there must not be another object in ``to_save`` with key ``checkpointer``.
         greater_or_equal: if `True`, the latest equally scored model is stored. Otherwise, the first model.
             Default, `False`.
+        save_on_rank: Which rank to save the objects on, in the distributed configuration. If ``save_handler`` is
+            string or :class:`~pathlib.Path`, this is also used to instantiate a :class:`~ignite.handlers.DiskSaver`.
 
     .. _DistributedDataParallel: https://pytorch.org/docs/stable/generated/
         torch.nn.parallel.DistributedDataParallel.html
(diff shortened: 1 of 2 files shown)


ignite/handlers/checkpoint.py
```diff
@@ -801,52 +816,40 @@ def _check_and_setup(dirname: Path, create_dir: bool, require_empty: bool) -> No
...
-        else:
-            self._save_native(checkpoint, path)
+        if self.save_on_rank == idist.get_rank():
+            path = self.dirname / filename
 
-    @idist.one_rank_only()
-    def _save_native(self, checkpoint: Mapping, path: Path) -> None:
-        self._save_func(checkpoint, path, torch.save)
+            if idist.has_xla_support:
+                import orch_xla.core.xla_model as xm
```
vfdev-5, line 823:
```suggestion
                import torch_xla.core.xla_model as xm
```

ignite/handlers/checkpoint.py
```diff
@@ -801,52 +816,40 @@ def _check_and_setup(dirname: Path, create_dir: bool, require_empty: bool) -> No
...
 
-    def _save_xla(self, checkpoint: Mapping, path: Path) -> None:
-        import torch_xla.core.xla_model as xm
-
-        # all tpu procs should enter here as internally performs sync across device
-        self._save_func(checkpoint, path, xm.save, rank=idist.get_rank())
+                # all tpu procs should enter here as internally performs sync across device
+                self._save_func(checkpoint, path, xm.save)
```
vfdev-5, line 826: For XLA all procs should enter xm.save, but now you added `if self.save_on_rank == idist.get_rank():` on the top, so only one proc will enter this function. Am I missing something?

ignite/handlers/checkpoint.py
```diff
@@ -165,13 +168,12 @@ class Checkpoint(Serializable):
...
-        process. This class supports automatically distributed configuration and if used with
-        :class:`~ignite.handlers.DiskSaver`, checkpoint is stored by rank 0 process.
+        process.
 
     .. warning::
 
-        When running on XLA devices, it should be run in all processes, otherwise application can get stuck on
-        saving the checkpoint.
+        When running on XLA devices or using :class:`~torch.distributed.optim.ZeroRedundancyOptimizer`, it
+        should be run in all processes, otherwise application can get stuck on saving the checkpoint.
```
vfdev-5, line 176: We may need to rephrase this sentence...

vfdev-5 (commented):
I haven't checked everything and will continue the review later

ignite/handlers/checkpoint.py
sadra-barikbin: How exactly do you mean?

ignite/handlers/checkpoint.py
sadra-barikbin: I corrected it

ignite/handlers/checkpoint.py
```diff
@@ -7,10 +7,11 @@
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
 from pathlib import Path
-from typing import Any, Callable, Dict, IO, List, Mapping, NamedTuple, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Mapping, NamedTuple, Optional, Tuple, Union
 
 import torch
 import torch.nn as nn
+from torch.distributed.optim import ZeroRedundancyOptimizer
```
vfdev-5, line 14: Ignite is supposed to work with pytorch 1.3.1 where `ZeroRedundancyOptimizer` is absent. We have to deal with this import differently.

PR #2642, opened 2022-08-13
Feature add ZeRO support to Checkpoint in a distributed configuration

### LINKED ISSUE #2623: Consolidate ZeRO state before checkpoint saving

sadra-barikbin:
Fixes #2623 

Description: To call `consolidate_state_dict` method of ZeRO optimizer on all ranks to before calling `state_dict` on target rank, the one responsible for saving the checkpoint.

Check list:

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -16,6 +16,10 @@ from packaging.version import Version
 if Version(torch.__version__) >= Version("1.9.0"):
     from torch.distributed.optim import ZeroRedundancyOptimizer
 
+    HAVE_ZERO = True
+else:
+    HAVE_ZERO = False
+
 import ignite.distributed as idist
 from ignite.base import Serializable
 from ignite.engine import Engine, Events
@@ -287,7 +291,7 @@ class Checkpoint(Serializable):
         filename_pattern: Optional[str] = None,
         include_self: bool = False,
         greater_or_equal: bool = False,
-        save_on_rank: Optional[int] = 0,
+        save_on_rank: int = 0,
     ):
 
         if not isinstance(to_save, collections.Mapping):
@@ -471,7 +475,7 @@ class Checkpoint(Serializable):
             for k, obj in self.to_save.items():
                 if isinstance(obj, (nn.DataParallel, nn.parallel.DistributedDataParallel)):
                     obj = obj.module
-                elif Version(torch.__version__) >= Version("1.9.0") and isinstance(obj, ZeroRedundancyOptimizer):
+                elif HAVE_ZERO and isinstance(obj, ZeroRedundancyOptimizer):
                     obj.consolidate_state_dict(to=self.save_on_rank)
                     if self.save_on_rank != idist.get_rank():
                         continue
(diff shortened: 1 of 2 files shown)


ignite/handlers/checkpoint.py
```diff
@@ -466,6 +471,10 @@ def _setup_checkpoint(self) -> Dict[str, Dict[Any, Any]]:
             for k, obj in self.to_save.items():
                 if isinstance(obj, (nn.DataParallel, nn.parallel.DistributedDataParallel)):
                     obj = obj.module
+                elif Version(torch.__version__) >= Version("1.9.0") and isinstance(obj, ZeroRedundancyOptimizer):
```
vfdev-5, line 474: Let's make version check once during the import and use a flag or something else instead of here

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -1254,6 +1282,9 @@ def test_distrib_gloo_cpu_or_gpu(distributed_context_single_node_gloo, get_rank_
     _test_checkpoint_with_ddp(device)
     _test_checkpoint_load_objects_ddp(device)
 
+    if Version(torch.__version__) >= Version("1.9.0"):
```
vfdev-5, line 1285: Here we can import the flag (or what we decided from `checkpoint.py`)

vfdev-5 (commented):
@sadra-barikbin thanks for the update, I left few comments to address

ignite/handlers/checkpoint.py
sadra-barikbin: Is it OK now?

vfdev-5 (approved):
Thanks @sadra-barikbin

vfdev-5:
@sadra-barikbin GPU tests are broken : https://github.com/pytorch/ignite/runs/7914418582?check_suite_focus=true

PR #2645, opened 2022-08-15
Rewritten Engine's terminate and terminate_epoch logic

vfdev-5:
Description:
- terminate() work on all events, called on catched _EngineTerminateException
- terminate_epoch() work on iteration-based events, called on catched  _EngineTerminateSingleEpochExpection
- Fixed issue when attaching handlers on Events.TERMINATE_SINGLE_EPOCH
- Added tests

Check list:

@sadra-barikbin can you please review and comment out what do you think ?

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -438,14 +438,28 @@ class Engine(Serializable):
         return self._fire_event(event_name)
 
     def terminate(self) -> None:
-        """Sends terminate signal to the engine, so that it terminates completely the run after
-        the current iteration."""
+        """Sends terminate signal to the engine, so that it terminates completely the run. The run is
+        terminated after the event on which ``terminate`` method was called. The following events are triggered:
+
+        - ...
+        - Terminating event
+        - :attr:`~ignite.engine.events.Events.TERMINATE`
+        - :attr:`~ignite.engine.events.Events.COMPLETED`
+        """
         self.logger.info("Terminate signaled. Engine will stop after current iteration is finished.")
         self.should_terminate = True
 
     def terminate_epoch(self) -> None:
-        """Sends terminate signal to the engine, so that it terminates the current epoch
-        after the current iteration."""
+        """Sends terminate signal to the engine, so that it terminates the current epoch. The run
+        continues from the next epoch. The following events are triggered:
+
+        - ...
+        - Event on which ``terminate_epoch`` method is called
+        - :attr:`~ignite.engine.events.Events.TERMINATE_SINGLE_EPOCH`
+        - :attr:`~ignite.engine.events.Events.EPOCH_COMPLETED`
+        - :attr:`~ignite.engine.events.Events.EPOCH_STARTED`
+        - ...
+        """
         self.logger.info(
             "Terminate current epoch is signaled. "
             "Current epoch iteration will stop after current iteration is finished."
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -757,31 +757,39 @@ class Engine(Serializable):
         try:
             start_time = time.time()
             self._fire_event(Events.STARTED)
+            if self.should_terminate:
+                self._fire_event(Events.TERMINATE)
+
             while not self._is_done(self.state) and not self.should_terminate:
                 self.state.epoch += 1
+                handlers_start_time = time.time()
                 self._fire_event(Events.EPOCH_STARTED)
+                epoch_time_taken = time.time() - handlers_start_time
 
-                if self._dataloader_iter is None:
-                    self._setup_engine()
+                if not self.should_terminate:
+                    if self._dataloader_iter is None:
+                        self._setup_engine()
 
-                time_taken = self._run_once_on_dataset()
-                # time is available for handlers but must be update after fire
-                self.state.times[Events.EPOCH_COMPLETED.name] = time_taken
+                    epoch_time_taken += self._run_once_on_dataset()
+
+                # time is available for handlers but must be updated after fire
+                self.state.times[Events.EPOCH_COMPLETED.name] = epoch_time_taken
+
                 handlers_start_time = time.time()
                 if self.should_terminate:
                     self._fire_event(Events.TERMINATE)
                 else:
                     self._fire_event(Events.EPOCH_COMPLETED)
-                time_taken += time.time() - handlers_start_time
+                epoch_time_taken += time.time() - handlers_start_time
                 # update time wrt handlers
-                self.state.times[Events.EPOCH_COMPLETED.name] = time_taken
-                hours, mins, secs = _to_hours_mins_secs(time_taken)
+                self.state.times[Events.EPOCH_COMPLETED.name] = epoch_time_taken
+                hours, mins, secs = _to_hours_mins_secs(epoch_time_taken)
                 self.logger.info(f"Epoch[{self.state.epoch}] Complete. Time taken: {hours:02d}:{mins:02d}:{secs:02d}")
                 if self.should_terminate:
                     break
 
             time_taken = time.time() - start_time
-            # time is available for handlers but must be update after fire
+            # time is available for handlers but must be updated after fire
             self.state.times[Events.COMPLETED.name] = time_taken
             handlers_start_time = time.time()
             self._fire_event(Events.COMPLETED)
@@ -856,8 +864,10 @@ class Engine(Serializable):
 
                 self.state.iteration += 1
                 self._fire_event(Events.ITERATION_STARTED)
-                self.state.output = self._process_function(self, self.state.batch)
-                self._fire_event(Events.ITERATION_COMPLETED)
+
+                if not (self.should_terminate or self.should_terminate_single_epoch):
+                    self.state.output = self._process_function(self, self.state.batch)
+                    self._fire_event(Events.ITERATION_COMPLETED)
 
                 if self.should_terminate or self.should_terminate_single_epoch:
                     self._fire_event(Events.TERMINATE_SINGLE_EPOCH, iter_counter=iter_counter)
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -301,7 +301,12 @@ class Engine(Serializable):
 
         self._assert_allowed_event(event_name)
 
-        event_args = (Exception(),) if event_name == Events.EXCEPTION_RAISED else ()
+        event_args = ()  # type: Tuple[Any, ...]
+        if event_name == Events.EXCEPTION_RAISED:
+            event_args += (Exception(),)
+        elif event_name == Events.TERMINATE_SINGLE_EPOCH:
+            event_args += (0,)
+
         try:
             _check_signature(handler, "handler", self, *(event_args + args), **kwargs)
             self._event_handlers[event_name].append((handler, (self,) + args, kwargs))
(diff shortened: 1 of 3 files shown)


ignite/engine/engine.py
```diff
@@ -298,7 +298,12 @@ def execute_something():
 
         self._assert_allowed_event(event_name)
 
-        event_args = (Exception(),) if event_name == Events.EXCEPTION_RAISED else ()
+        event_args = ()  # type: Tuple[Any, ...]
+        if event_name == Events.EXCEPTION_RAISED:
+            event_args += (Exception(),)
+        elif event_name == Events.TERMINATE_SINGLE_EPOCH:
+            event_args += (0,)
```
vfdev-5, line 305: As this was missing, users couldn't previously attach any handler to `Events.TERMINATE_SINGLE_EPOCH`

ignite/engine/engine.py
```diff
@@ -298,7 +298,12 @@ def execute_something():
 
         self._assert_allowed_event(event_name)
 
-        event_args = (Exception(),) if event_name == Events.EXCEPTION_RAISED else ()
+        event_args = ()  # type: Tuple[Any, ...]
+        if event_name == Events.EXCEPTION_RAISED:
+            event_args += (Exception(),)
+        elif event_name == Events.TERMINATE_SINGLE_EPOCH:
```
sadra-barikbin, line 304: What is the engine supposed to do after receiving  `terminate_single_epoch` command?

ignite/engine/engine.py
```diff
@@ -835,29 +861,32 @@ def _run_once_on_dataset(self) -> float:
...
                     self.should_terminate = True
-                    break
+                    raise _EngineTerminateException()
+
+        except _EngineTerminateSingleEpochException:
+            self._fire_event(Events.TERMINATE_SINGLE_EPOCH, iter_counter=iter_counter)
+            self.should_terminate_single_epoch = False
+            self._setup_dataloader_iter()
 
         except Exception as e:
```
sadra-barikbin, line 891: Here general exception class is captured which includes `_EngineTerminateException`, while there is an `except` clause for that in line 782. This results in that clause not be able to capture the exception.

vfdev-5: Thanks for catching that ! Fixed

ignite/engine/engine.py
```diff
@@ -835,29 +861,37 @@ def _run_once_on_dataset(self) -> float:
...
+                    self._maybe_terminate()
 
+                    self._setup_dataloader_iter()
                     should_exit = True
 
                     continue
 
                 self.state.iteration += 1
                 self._fire_event(Events.ITERATION_STARTED)
+                self._maybe_terminate()
```
sadra-barikbin, line 873: Doc says about `terminate_epoch`: "it terminates the current epoch after the current iteration.", but here it could terminate while iteration uncompleted.

ignite/engine/engine.py
```diff
@@ -742,33 +747,43 @@ def _internal_run(self) -> State:
...
+                    if self._dataloader_iter is None:
+                        self._setup_engine()
+
+                    epoch_time_taken += self._run_once_on_dataset()
+
+                    # time is available for handlers but must be updated after fire
+                    self.state.times[Events.EPOCH_COMPLETED.name] = epoch_time_taken
+
+                    handlers_start_time = time.time()
                     self._fire_event(Events.EPOCH_COMPLETED)
```
sadra-barikbin, line 771: If user calls `terminate_epoch` during `_run_once_on_dataset`, Do we still want to emit `EPOCH_COMPLETED` event while run is terminated after iteration and moreover `TERMINATE_SINGLE_EPOCH` has been emitted?

ignite/engine/engine.py
vfdev-5: How would you rephrase the doc in this case ? 
I think the current code is doing what we would expect. If we call `terminate_epoch` in a handler on ITERATION_STARTED we expect engine to end the epoch after ITERATION_STARTED event and not after ITERATION_COMPLETED.

ignite/engine/engine.py
vfdev-5: Yes, I think it is still makes sense to call EPOCH_COMPLETED even if terminate_epoch() was called.
Check this example: https://github.com/pytorch/ignite/blob/d8af03c2e34e8dff22e17570a32aedd0b755ff82/examples/reinforcement_learning/reinforce.py#L94-L116

vfdev-5:
@sadra-barikbin do you have any other remarks for this PR that could be blocking it for being merged ?

sadra-barikbin:
> @sadra-barikbin do you have any other remarks for this PR that could be blocking it for being merged ?

I finalize my review in a few minutes.

vfdev-5:
> I finalize my review in a few minutes.

@sadra-barikbin have you submitted your review ?

tests/ignite/engine/test_engine.py
```diff
@@ -103,11 +103,11 @@ def end_of_epoch_handler(engine):
     assert engine.should_terminate
 
 
-@pytest.mark.parametrize("data", [None, [1, 2, 3]])
-def test_terminate_at_start_of_epoch_stops_run_after_completing_iteration(data):
+@pytest.mark.parametrize("data, epoch_length", [(None, 10), (range(10), None)])
+def test_terminate_at_start_of_epoch_stops_run_after_completing_iteration(data, epoch_length):
```
sadra-barikbin, line 107: Is the test name correct? According to the code, termination after `EPOCH_STARTED` terminates the run immediately.

tests/ignite/engine/test_engine.py
```diff
@@ -1014,3 +1168,66 @@ def train_step(engine, batch):
...
+
+    # error
+    with pytest.raises(ValueError, match="Argument max_epochs should be larger than the start epoch"):
+        engine.run(data, max_epochs=4, epoch_length=epoch_length)
+
+    # restart from 0 to 7 (As state.epoch == max_epochs(=7),
+    # this should be like that as we always do: evaluator.run(data) without any other instructions)
+    first_epoch_iter[0], first_epoch_iter[1] = 0, 0
+    # Engine run starting with max_epochs=7 => state.epoch=7
+    engine.run(data, max_epochs=7, epoch_length=epoch_length)
```
sadra-barikbin, line 1208: Here `max_epoch` is equal to `state.max_epoch` but in the engine code we have `Argument max_epochs should be larger than the start epoch defined in the state`.

tests/ignite/engine/test_engine.py
vfdev-5: Yes, this test name looks weird. I'll rename it to `test_terminate_at_start_of_epoch`. Any other suggestions if it does not work for you ?

tests/ignite/engine/test_engine.py
vfdev-5: So, what is the issue with that. If provided `max_epochs  == engine.state.max_epochs` then engine will restart. Similarly, to `evaluator.run(val_data)` where max_epochs=1 and after the first validation run we have `evaluator.state.max_epochs==1`.

vfdev-5:
@sadra-barikbin I updated the docs and tests etc. Please review it again. I want to merge it by tomorrow. Thanks

tests/ignite/engine/test_engine.py
sadra-barikbin: So in docstring we should say `Argument max_epochs should be greater than or equal to the start epoch defined in the state`.

vfdev-5: Yes, "Argument max_epochs should be greater or equal to the start epoch defined in the state"
if you would like to send a PR with the change, feel free.

tests/ignite/engine/test_engine.py
```diff
@@ -124,32 +124,147 @@ def start_of_epoch_handler(engine):
...
+            expected_iter += 1
+            assert engine.state.iteration == expected_iter
+            assert engine.state.batch == data[(expected_iter - first_epoch_iter[1] - 1) % len(data)]
+
+    first_epoch_iter[0], first_epoch_iter[1] = state.epoch, state.iteration
+    state = engine.run(data, max_epochs=max_epochs, epoch_length=epoch_length)
+
+    assert state.epoch == max_epochs
+    assert not engine.should_terminate
+    assert state.iteration == real_epoch_length * (max_epochs - 1)
```
sadra-barikbin, line 209: This shouldn't be `state.iteration == real_epoch_length * (max_epochs - 1) + (iteration_to_stop % real_epoch_length)` ? Since engine has run a few iterations from the epoch in which it stopped.

tests/ignite/engine/test_engine.py
```diff
@@ -124,32 +124,147 @@ def start_of_epoch_handler(engine):
...
+
     engine.add_event_handler(Events.ITERATION_STARTED, start_of_iteration_handler)
-    state = engine.run(data, max_epochs=3, epoch_length=num_iterations_per_epoch)
+    state = engine.run(data, max_epochs=max_epochs, epoch_length=epoch_length)
     # completes the iteration but doesn't increment counter (this happens just before a new iteration starts)
     assert state.iteration == iteration_to_stop
-    assert state.epoch == np.ceil(iteration_to_stop / num_iterations_per_epoch)  # it starts from 0
+    assert state.epoch == np.ceil(iteration_to_stop / real_epoch_length)  # it starts from 0
+
+    # Engine continue from epoch_to_terminate_on until max_epochs
```
sadra-barikbin, line 185: This shouldn't be `continue from iteration_to_stop until max_epochs` or something like that?

tests/ignite/engine/test_engine.py
```diff
@@ -159,10 +274,58 @@ def start_of_iteration_handler(engine):
...
+        (Events.ITERATION_STARTED(once=14), 14),
+        (Events.ITERATION_COMPLETED(once=14), 14),
+    ],
+)
+def test_terminate_epoch_events_sequence(terminate_epoch_event, i):
+    engine = RecordedEngine(MagicMock(return_value=1))
+    data = range(10)
+    max_epochs = 3
+
+    # TODO: Bug: Events.GET_BATCH_STARTED(once=12) is called twice !
```
sadra-barikbin, line 298: This will be solved when engine is refactored. Yes?
Because now event filter decides on `engine.iteration` (which is not incremented due to termination), but then decides on the actual count of `GET_BATCH_STARTED`.

vfdev-5: yes, something like that

tests/ignite/engine/test_engine.py
vfdev-5: As this test is passing then no, it is `state.iteration == real_epoch_length * (max_epochs - 1)`. 
However, I agree that we may expect iteration to be `state.iteration == real_epoch_length * (max_epochs - 1) + (iteration_to_stop % real_epoch_length)`...

Right now if we inspect epoch/iteration and data attributes we have the following:
```
- (epoch, iter, data): 1 1  |  0
- (epoch, iter, data): 1 2  |  1
- (epoch, iter, data): 1 3  |  2
- (epoch, iter, data): 1 4  |  3
- (epoch, iter, data): 1 5  |  4
- (epoch, iter, data): 1 6  |  5
- (epoch, iter, data): 1 7  |  6
- (epoch, iter, data): 1 8  |  7
- (epoch, iter, data): 1 9  |  8
- (epoch, iter, data): 1 10  |  9
- (epoch, iter, data): 2 11  |  0
- (epoch, iter, data): 2 12  |  1
- (epoch, iter, data): 2 13  |  2
-> terminate
Continue the run: 2 13
- (epoch, iter, data): 3 14  |  0
- (epoch, iter, data): 3 15  |  1
- (epoch, iter, data): 3 16  |  2
- (epoch, iter, data): 3 17  |  3
- (epoch, iter, data): 3 18  |  4
- (epoch, iter, data): 3 19  |  5
- (epoch, iter, data): 3 20  |  6
- (epoch, iter, data): 4 21  |  7
- (epoch, iter, data): 4 22  |  8
- (epoch, iter, data): 4 23  |  9
- (epoch, iter, data): 4 24  |  0
- (epoch, iter, data): 4 25  |  1
- (epoch, iter, data): 4 26  |  2
- (epoch, iter, data): 4 27  |  3
- (epoch, iter, data): 4 28  |  4
- (epoch, iter, data): 4 29  |  5
- (epoch, iter, data): 4 30  |  6
- (epoch, iter, data): 5 31  |  7
- (epoch, iter, data): 5 32  |  8
- (epoch, iter, data): 5 33  |  9
- (epoch, iter, data): 5 34  |  0
- (epoch, iter, data): 5 35  |  1
- (epoch, iter, data): 5 36  |  2
- (epoch, iter, data): 5 37  |  3
- (epoch, iter, data): 5 38  |  4
- (epoch, iter, data): 5 39  |  5
- (epoch, iter, data): 5 40  |  6
```

Does it look strange or weird to you ?

tests/ignite/engine/test_engine.py
sadra-barikbin: Couldn't we do in this same PR?

tests/ignite/engine/test_engine.py
sadra-barikbin: Epoch 3 is supposed to be run over all data not the first 7 batches:
image

sadra-barikbin: It's due to `iter_counter` local variable (in `run_once_on_dataset`) which has been initialized as `state.iteration % epoch_length` (3)  and causes engine to stop 7 batches later when `iter_counter == epoch_length`.

Hence, `iter_counter` should start with zero.

vfdev-5: Let's move the discussions in this issue: https://github.com/pytorch/ignite/issues/2669

tests/ignite/engine/test_engine.py
vfdev-5: Let's fix that in a follow-up PR

sadra-barikbin (approved):

PR #2646, opened 2022-08-15
Update test_accuracy.py

puhuk:
Description: Change `_test_distrib_integration_multilabel` to test with different rank

Check list:


vfdev-5 (approved):
Thanks !

PR #2650, opened 2022-08-16
Improved logging time taken message showing milliseconds

vfdev-5:
Description:
- Improved logging time taken message showing milliseconds in Engine, utils and Cifar10 example
- removed eval pbar from cifar10

Check list:

### DIFF
--- a/examples/contrib/cifar10/main.py
+++ b/examples/contrib/cifar10/main.py
@@ -260,7 +260,7 @@ def initialize(config):
 
 def log_metrics(logger, epoch, elapsed, tag, metrics):
     metrics_output = "\n".join([f"\t{k}: {v}" for k, v in metrics.items()])
-    logger.info(f"\nEpoch {epoch} - Evaluation time (seconds): {elapsed:.2f} - {tag} metrics:\n {metrics_output}")
+    logger.info(f"Epoch[{epoch}] - Evaluation time (seconds): {elapsed:.3f}\n - {tag} metrics:\n {metrics_output}")
 
 
 def log_basic_info(logger, config):
@@ -379,9 +379,6 @@ def create_evaluator(model, metrics, config, tag="val"):
 
     for name, metric in metrics.items():
         metric.attach(evaluator, name)
-
-    if idist.get_rank() == 0 and (not config["with_clearml"]):
-        common.ProgressBar(desc=f"Evaluation ({tag})", persist=False).attach(evaluator)
 
     return evaluator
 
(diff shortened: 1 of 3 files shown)


examples/contrib/cifar10/main.py
```diff
@@ -380,9 +380,6 @@ def evaluate_step(engine: Engine, batch):
     for name, metric in metrics.items():
         metric.attach(evaluator, name)
 
-    if idist.get_rank() == 0 and (not config["with_clearml"]):
```
sadra-barikbin, line 383: Why did you this pbar?

vfdev-5: I does not look helpful and also adds useless lines if logs are written into a file

ignite/engine/engine.py
```diff
@@ -776,7 +776,7 @@ def _internal_run(self) -> State:
             # update time wrt handlers
             self.state.times[Events.COMPLETED.name] = time_taken
             hours, mins, secs = _to_hours_mins_secs(time_taken)
-            self.logger.info(f"Engine run complete. Time taken: {hours:02d}:{mins:02d}:{secs:02d}")
+            self.logger.info(f"Engine run complete. Time taken: {hours:02d}:{mins:02d}:{secs:06.3f}")
```
sadra-barikbin, line 779: I do not get `6` in `06.3f`. seconds does not outnumber 60 which has two digits. why have you assumed 6 digits for it?

vfdev-5: Yes, `06.3f` means 6 digits in total for float number of seconds. Check for example:
```python
x = 7.890
print(f"{x:06.3f}")
```
>
```
07.890
```

I need the first padding zero to have uniform representation + milliseconds

examples/contrib/cifar10/main.py
```diff
@@ -260,7 +260,7 @@ def initialize(config):
 
 def log_metrics(logger, epoch, elapsed, tag, metrics):
     metrics_output = "\n".join([f"\t{k}: {v}" for k, v in metrics.items()])
-    logger.info(f"\nEpoch {epoch} - Evaluation time (seconds): {elapsed:.2f} - {tag} metrics:\n {metrics_output}")
+    logger.info(f"Epoch[{epoch}] - Evaluation time (seconds): {elapsed:.3f}\n - {tag} metrics:\n {metrics_output}")
```
sadra-barikbin, line 263: If evaluation time is long, isn't more understandable to have:
```suggestion
    minutes, seconds = divmod(elapsed, 60)
    logger.info(f"Epoch[{epoch}] - Evaluation time : {int(minutes)} min and {int(seconds)} s\n - {tag} metrics:\n {metrics_output}")
```

sadra-barikbin (approved):

examples/contrib/cifar10/main.py
vfdev-5: For cifar10 it wont be that long. It is around 2-3 seconds for resnet18. But in any case `elapsed` refers to `engine.state.times[Events.COMPLETED]` and where the time is seconds no matter how much time it took.

PR #2652, opened 2022-08-17
Check torch version once in fid.py

vfdev-5:
Description:
- Check torch version once in fid.py

Check list:

### DIFF
--- a/ignite/metrics/gan/fid.py
+++ b/ignite/metrics/gan/fid.py
@@ -12,6 +12,12 @@ __all__ = [
 ]
 
 
+if Version(torch.__version__) <= Version("1.7.0"):
+    torch_outer = torch.ger
+else:
+    torch_outer = torch.outer
+
+
 def fid_score(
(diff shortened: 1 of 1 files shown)


ignite/metrics/gan/fid.py
```diff
@@ -12,6 +12,12 @@
 ]
 
 
+if Version(torch.__version__) <= Version("1.7.0"):
+    torch_outer = torch.ger
```
sadra-barikbin, line 16: Is codecov warning OK?

sadra-barikbin (approved):

ignite/metrics/gan/fid.py
vfdev-5: Yes, we do not run the CI on older pth that's why we do not enter it here. I can also run manually https://github.com/pytorch/ignite/actions/workflows/pytorch-version-tests.yml to ensure that everything is good.

PR #2658, opened 2022-08-19
Fix device in checkpoint ZeRO test

sadra-barikbin:
Fix a bug in that caused failure in GPU tests.

Check list:


tests/ignite/handlers/test_checkpoint.py
```diff
@@ -1252,7 +1252,7 @@ def _test_checkpoint_with_ZeRO(device, dirname, local_rank):
     mocked_opt = MagicMock(ZeroRedundancyOptimizer, wraps=opt)
 
     # A `step` should be called to optimizer state get populated.
-    out = model(torch.Tensor([1.0]))
+    out = model(torch.Tensor([1.0], device=device))
```
vfdev-5, line 1255: Tensor creation should be with `torch.tensor` . Can you fix it here and everywhere in this file

vfdev-5 (approved):
Thanks for the fix

PR #2663, opened 2022-08-21
Check if score is bigger than self.best_score

Noobtoss:
Added a check if score is bigger than self.best_score.

Otherwise, when self.min_delta>0, self.best_score could decrease each call.

I am terribly sorry if my proposal is incorrect.


sadra-barikbin:
Hi @Noobtoss ,
Thank you for the PR.
In the `else` statement in which you've suggested this change, it is guaranteed that `score` is bigger than `self.best_score` so there's no need to do this change. Am I correct?

Noobtoss:
Yes you are correct.

My apologies.

PR #2666, opened 2022-08-22
Update r2_score and wave_hedges_distance in test for generating data with different rank

puhuk:
Description: Update `r2_score`, `wave_hedges_distance` to generate with different rank

Check list:


vfdev-5 (approved):
Thanks @puhuk

PR #2670, opened 2022-08-23
Update pairwise_dist, mse and lambda in test for generating data with different rank

puhuk:
Description: Update `mean_pairwise_distance`, `mean_squared_error`, `metrics_lambda`

Check list:

- [x] New tests are added (if a new feature is added)
- [ ] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #2685, opened 2022-09-01
To fix errors from PR #2670

puhuk:
Description: To fix errors from PR #2670

Check list:


vfdev-5:
@puhuk can you provide only one commit with the fix instead of the list of already merged updates?

PR #2691, opened 2022-09-02
Fix a little typo in engine docstring

### LINKED ISSUE #2689: Fix a few bugs in engine interrupt/resume feature
## 🐛 Bug description

- [ ] A typo in [docstrings](https://github.com/pytorch/ignite/blob/c12ad57e26aa8ec2b191ff4bc0e01e97ef24f8f7/ignite/engine/engine.py#L34). 0.4.10 or 0.5.0?
- [ ] ~~`_run_once_on_dataset_as_gen` is restarted (because it is called) whenever we get into `_internal_run_as_gen`~~
- [ ] ~~Some statements before `_run_once_on_dataset_as_gen` should be skipped on resume and some other after that should be skipped on interrupt~~
- [ ] ~~Once running after a terminated state, `_run_once_on_dataset_as_gen` should be set to None before entering `_internal_run`.~~
- [ ] ~~If engine was interrupted in the middle of the last epoch, `is_done` incorrectly returns True on resume thus preventing it.~~
- [ ] ~~`_run_once_on_dataset_as_gen` yields the state on interrupt, not the time, so [this](https://github.com/pytorch/ignite/blob/c12ad57e26aa8ec2b191ff4bc0e01e97ef24f8f7/ignite/engine/engine.py#L965) shoud be changed.~~
- [ ] Since `_run_once_on_dataset_as_gen` has both return value and yield value with different types, its annotation could be updated to reflect this e.g. become `Generator[State, None, float]`

sadra-barikbin:
Closes #2689

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -31,7 +31,7 @@ class Engine(Serializable):
         last_event_name: last event name triggered by the engine.
 
     Note:
-        :class:`~ignite.engine.engine.Engine` implementation has changed in v0.4.10 with "interrupt/resume" feature.
+        :class:`~ignite.engine.engine.Engine` implementation has changed in v0.5.0 with "interrupt/resume" feature.
         Engine may behave differently on certain corner cases compared to the one from v0.4.9 and before.
         In such case, you can set ``Engine.interrupt_resume_enabled = False`` to restore previous behaviour.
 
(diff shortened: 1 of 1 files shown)


vfdev-5 (approved):
Thanks @sadra-barikbin

PR #2700, opened 2022-09-05
Fixed wrong device type for test_psnr on hvd

vfdev-5:
Description:
- Fixed wrong device type for test_psnr on hvd

Check list:


vfdev-5:
cc @puhuk

PR #2712, opened 2022-09-16
Modify all_reduce to take `group`

puhuk:
Description: Modify all_reduce to take `group`

Check list:

- [ ] New tests are added (if a new feature is added)
- [ ] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #2726, opened 2022-09-29
To resolve version fails from PR #2712 (all_reduce with group)

puhuk:
Description: To resolve version fails from PR #2712 (all_reduce with group)

Check list:


vfdev-5 (approved):
Thanks @puhuk

PR #2727, opened 2022-09-30
add before and after events filter

louis-she:
Description:

add before and after events filter

```python
@trainer.on(Events.EPOCH_COMPLETED(before=10))
def foo1():
    pass

@trainer.on(Events.EPOCH_COMPLETED(after=10))
def foo2():
    pass
```

Check list:

### DIFF
--- a/tests/ignite/engine/test_custom_events.py
+++ b/tests/ignite/engine/test_custom_events.py
@@ -583,4 +583,5 @@ def test_list_of_events():
 
     _test(Events.ITERATION_STARTED(once=1) | Events.ITERATION_STARTED(once=1), [1, 1])
     _test(Events.ITERATION_STARTED(once=1) | Events.ITERATION_STARTED(once=10), [1, 10])
-    _test(Events.ITERATION_STARTED(once=1) | Events.ITERATION_STARTED(every=3), [1, 3, 6, 9, 12, 15])
+    _test(Events.ITERATION_STARTED(once=8) | Events.ITERATION_STARTED(before=3), [1, 2, 8])
+    _test(Events.ITERATION_STARTED(once=1) | Events.ITERATION_STARTED(after=12), [1, 13, 14, 15])
--- a/tests/ignite/engine/test_custom_events.py
+++ b/tests/ignite/engine/test_custom_events.py
@@ -130,11 +130,22 @@ def test_custom_events_with_events_list():
 
 def test_callable_events_with_wrong_inputs():
 
-    with pytest.raises(ValueError, match=r"Only one of the input arguments should be specified"):
+    with pytest.raises(
+        ValueError, match=r"Only one of the input arguments should be specified except before and after"
+    ):
         Events.ITERATION_STARTED()
 
-    with pytest.raises(ValueError, match=r"Only one of the input arguments should be specified"):
+    with pytest.raises(
+        ValueError, match=r"Only one of the input arguments should be specified except before and after"
+    ):
         Events.ITERATION_STARTED(event_filter="123", every=12)
+
+    with pytest.raises(
+        ValueError, match=r"Only one of the input arguments should be specified except before and after"
+    ):
+        Events.ITERATION_STARTED(after=10, before=30, once=1)
+
+    assert Events.ITERATION_STARTED(after=10, before=30)
 
     with pytest.raises(TypeError, match=r"Argument event_filter should be a callable"):
         Events.ITERATION_STARTED(event_filter="123")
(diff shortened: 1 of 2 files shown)


louis-she:
I removed the `WIP` in title. But one thing left is that should we add a `__and__` to the Event? Like this:

```python
@trainer.on(Events.EPOCH_COMPLETED(after=10) & Events.EPOCH_COMPLETED(before=30))
def foo1():
    # called on epoch > 10 and < 30
    pass
```

Currently with `|` or `__or__`, I think we can not achieve this with event filter only

tests/ignite/engine/test_custom_events.py
```diff
@@ -531,4 +583,5 @@ def execute_some_handler(e):
 
     _test(Events.ITERATION_STARTED(once=1) | Events.ITERATION_STARTED(once=1), [1, 1])
     _test(Events.ITERATION_STARTED(once=1) | Events.ITERATION_STARTED(once=10), [1, 10])
-    _test(Events.ITERATION_STARTED(once=1) | Events.ITERATION_STARTED(every=3), [1, 3, 6, 9, 12, 15])
```
sadra-barikbin, line 534: Why have you deleted the statement testing `every`?

sadra-barikbin (changes requested):

sadra-barikbin:
> I think it's ready to be reviewed. One thing to be discussed is that should we add a `__and__` operator to the Event? Like this:
> 
> ```python
> @trainer.on(Events.EPOCH_COMPLETED(after=10) & Events.EPOCH_COMPLETED(before=30))
> def foo1():
>     # called on epoch > 10 and < 30
>     pass
> ```
> 
> Currently with `|` or `__or__`, I think we can not achieve this with event filter only

We could accept both `before` and `after` in the `CallableEventWithFilter.__call__` and use `self.before_and_after_event_filter(before: Optional[int], after: Optional[int])` internally.

louis-she:
> We could accept both before and after in the CallableEventWithFilter.__call__ and use self.before_and_after_event_filter(before: Optional[int], after: Optional[int]) internally.

hi @sadra-barikbin , you mean if a user gives `before=30 | after=10`, the internal will do the `before_and_after_event_filter`? I'm not sure about this as it could lead to some misunderstanding. Still, I think adding a `&` operator is the way to go, what do you think?

tests/ignite/engine/test_custom_events.py
louis-she: Sorry, it's an accident, fixed.

sadra-barikbin:
> > We could accept both before and after in the CallableEventWithFilter.**call** and use self.before_and_after_event_filter(before: Optional[int], after: Optional[int]) internally.
> 
> For the `called on epoch > 10 and < 30` scenario, it'll work. Another possible scenario is `called every x epochs that after y epochs`. What do you think?

I think introducing `&` operator should be discussed in terms of its usefulness, but just in the case of mixing `before` and `after` we can do it as follows:
```python
@trainer.on(Events.EPOCH_COMPLETED(after=10, before=30))
def foo1():
    # called on epoch > 10 and < 30
    pass
```
@vfdev-5 , What's your thought?

vfdev-5:
> 

Yes, I agree. We can just allow after and before args together and thus do not need to introduce complicated & operator.

louis-she:
comments fixed and rebased, ready to review again @sadra-barikbin

vfdev-5:
@louis-she could you please fix mypy issues: https://github.com/pytorch/ignite/actions/runs/3180145927/jobs/5183390359#step:12:12

vfdev-5 (approved):
Thanks @louis-she , looks good to me !

tests/ignite/engine/test_custom_events.py
```diff
@@ -296,6 +313,79 @@ def test_every_event_filter_with_engine():
...
+        def _before_and_after_event():
+            nonlocal num_calls
+            num_calls += 1
+            assert getattr(engine.state, event_attr) > after
+
+        engine.run(data, max_epochs=5)
+        assert num_calls == expect_calls
+
+    _test(Events.ITERATION_STARTED, "iteration", 300, 100, 199)
+    _test(Events.EPOCH_COMPLETED, "epoch", 4, 1, 2)
```
sadra-barikbin, line 386: Thanks @louis-she , You can improve this test by using PyTest parameterization feature as follows:
```suggestion
@pytest.mark.parametrize(
    "event_name, before, after, expected_calls",
    [
        (Events.ITERATION_STARTED, 300, 100, 199),
        (Events.EPOCH_COMPLETED, 4, 1, 2)
    ],
)
def test_before_and_after_event_filter_with_engine(event_name, before, after, expected_calls):

    data = range(100)
    engine = Engine(lambda e, b: 1)
    num_calls = 0

   @engine.on(event_name(before=before, after=after))
   def _before_and_after_event():
        nonlocal num_calls
        num_calls += 1
        assert engine.state.get_event_attrib_value(event_name) > after

    engine.run(data, max_epochs=5)
    assert num_calls == expected_calls
```

tests/ignite/engine/test_custom_events.py
```diff
@@ -296,6 +313,79 @@ def test_every_event_filter_with_engine():
...
+    _test(Events.ITERATION_COMPLETED, "iteration", 0, 0)
+    _test(Events.ITERATION_COMPLETED, "iteration", 300, 299)
+    _test(Events.ITERATION_COMPLETED, "iteration", 501, 500)
+
+    _test(Events.EPOCH_COMPLETED, "epoch", 0, 0)
+    _test(Events.EPOCH_COMPLETED, "epoch", 3, 2)
+    _test(Events.EPOCH_COMPLETED, "epoch", 6, 5)
+
+
+def test_after_event_filter_with_engine():
```
sadra-barikbin, line 342: An also this test.

tests/ignite/engine/test_custom_events.py
```diff
@@ -296,6 +313,79 @@ def test_every_event_filter_with_engine():
     _test_every_event_filter_with_engine()
 
 
+def test_before_event_filter_with_engine():
```
sadra-barikbin, line 316: And this test

sadra-barikbin (approved):

PR #2730, opened 2022-10-03
Gather native version

puhuk:
Description: To correct GPU fails

Check list:


vfdev-5 (approved):
Thanks for the fix

PR #2737, opened 2022-10-09
FIX: accumulation should not effect value of loss

### LINKED ISSUE #2707: gradient_accumulation_steps influences scale of the loss
## 🐛 Bug description

[Because of the division here](https://github.com/pytorch/ignite/blob/master/ignite/engine/__init__.py#L185), and similar functions, the loss has the wrong scale when using `gradient_accumulation_steps`. This makes it confusingly low in comparison to the valid loss.

One option is to only divide on the backward call, i.e. doing this:
```
scaler.scale(loss / gradient_accumulation_steps).backward()
```
or one could multiply the loss again with `gradient_accumulation_steps` before returning it.

louis-she:
Fixes #2707 2707

Description:

Check list:

### DIFF
--- a/ignite/engine/__init__.py
+++ b/ignite/engine/__init__.py
@@ -106,7 +106,7 @@ def supervised_training_step(
         loss.backward()
         if engine.state.iteration % gradient_accumulation_steps == 0:
             optimizer.step()
-        return output_transform(x, y, y_pred, loss)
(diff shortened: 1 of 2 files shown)


louis-she:
`codecov` seems to be not going right here. The code just changes some value comparisons.

vfdev-5 (approved):
Thanks @louis-she , looks good to me!

PR #2742, opened 2022-10-17
Dropper TrainsLoger and TrainsSaver also removed the backward compati…

soma2000-lang:
…bility

Fixes #{issue number}

Description:

Check list:


vfdev-5:
@louis-she could you please review as you wanted to do that initially :)

louis-she:
@vfdev-5 of course

soma2000-lang:
@vfdev-5 sure

ignite/contrib/handlers/clearml_logger.py
```diff
@@ -115,15 +115,10 @@ def __init__(self, **kwargs: Any):
...
-            try:
-                # Backwards-compatibility for legacy Trains SDK
-                from trains import Task
-                from trains.binding.frameworks.tensorflow_bind import WeightsGradientHistHelper
-            except ImportError:
-                raise RuntimeError(
-                    "This contrib module requires clearml to be installed. "
-                    "You may install clearml using: \n pip install clearml \n"
-                )
+            raise RuntimeError(
```
louis-she, line 118: I think we can use `ModuleNotFoundError` to replace `RuntimeError` here.

vfdev-5: Yes, it makes sense, thanks for the suggestion! Then it would be better to consistently raise that in all contrib handlers and classes where we check for available 3rd party library. @louis-she can you please open a new issue, I'll label it as good first issue/hacktoberfest

vfdev-5 (approved):
lgtm, thanks @soma2000-lang !

Thanks for the review @louis-she

PR #2744, opened 2022-10-18
Replaced RuntimeError with ModuleNotFoundError in contribute handlers

### LINKED ISSUE #2743: Replace RuntimeError with ModuleNotFoundError in contribute handlers
Some contribute handlers rely on third-party tools not installed alone with ignite. So there usually will be an import error when users use them for the first time. Some of the import errors raised are `RuntimeError`, instead, we want it to be a more specific error type like `ModuleNotFoundError`.

We should replace all the manually raised `RuntimeError` with `ModuleNotFoundError` when some package can not be imported in third-party contribute handlers.

reference: https://github.com/pytorch/ignite/pull/2742#discussion_r997187850

aryanlath:
Fixes #2743

Description:

Check list:


ignite/contrib/handlers/clearml_logger.py
```diff
@@ -115,7 +115,7 @@ def __init__(self, **kwargs: Any):
             from clearml import Task
             from clearml.binding.frameworks.tensorflow_bind import WeightsGradientHistHelper
         except ImportError:
-            raise RuntimeError(
+            raise ModuleNotFoundError(
```
sadra-barikbin, line 118: @aryanlath , thanks for the PR. Could you please do the same thing to the `mlflow_logger`, `tensorboard_logger`, `wandb_logger`, `tqdm_logger`, `visdom_logger`, `polyaxon_logger`  and `neptune_logger` in `ignite/contrib/handlers` folder?

sadra-barikbin (changes requested):

ignite/contrib/handlers/clearml_logger.py
aryanlath: Sure

vfdev-5: I would also suggest to search for all `except ImportError:` and see which error type is raised and replaced there as well.

vfdev-5:
@aryanlath do you plan to work on the PR ?

aryanlath:
Sorry I have some work going on so could not update the PR

vfdev-5:
Unfortunately, I have to close your PR as you do not work on the issue such that others could help with. Thanks for your contribution.

PR #2750, opened 2022-10-23
Put `ModuleNotFoundError` instead of `RuntimeError` in proper places

### LINKED ISSUE #2743: Replace RuntimeError with ModuleNotFoundError in contribute handlers

sadra-barikbin:
Fixes #2743


examples/notebooks/CycleGAN_with_nvidia_apex.ipynb
```diff
@@ -1,32 +1,4 @@
 {
```
vfdev-5, line 1: @sadra-barikbin let's revert the change for this file

examples/notebooks/EfficientNet_Cifar100_finetuning.ipynb
```diff
@@ -1,30 +1,4 @@
 {
```
vfdev-5, line 1: Same here, @sadra-barikbin let's revert the change for this file

vfdev-5 (commented):
Looks good to me, thanks @sadra-barikbin 
Let's just revert changes in jupyter notebooks. Currently there is a lot unrelated changes, so I think we can just update python code.

vfdev-5:
@sadra-barikbin ipynb are still in the PR

vfdev-5 (approved):
Thanks @sadra-barikbin !

Issue #2754, opened 2022-10-28
DeterministicEngine improvement


louis-she:
* remove `torch.cuda.manual_seed_all(seed)` [here](https://github.com/pytorch/ignite/blob/1800feec9b4f70b2a6cd8f409ac76da06015efb1/ignite/utils.py#L237
) as `torch.manual_seed(seed)` will seed CUDA if it has CUDA enabled.
* try to use `torch.use_deterministic_algorithms(True)` if torch has it instead of `torch.backends.cudnn.deterministic = True`, as the later only makes convolutional deterministic.

vfdev-5:
Solved by #2756

PR #2756, opened 2022-10-29
improve deterministic engine

louis-she:
#2754 

The `neptune-client` has make some APIs to their legacy package, see https://github.com/neptune-ai/neptune-client/pull/1039

### DIFF
--- a/ignite/engine/deterministic.py
+++ b/ignite/engine/deterministic.py
@@ -194,11 +194,7 @@ class DeterministicEngine(Engine):
         self.state.seed = int(torch.randint(0, int(1e9), (1,)).item())
         if torch.cuda.is_available():
             if hasattr(torch, "use_deterministic_algorithms"):
-                # according to https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility
-                # CUBLAS_WORKSPACE_CONFIG must be set to let cuBLAS behave deterministic.
-                # **the behavior is expected to change in a future release of cuBLAS**.
-                os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
-                torch.use_deterministic_algorithms(True)
+                torch.use_deterministic_algorithms(True, warn_only=True)
             else:
                 torch.backends.cudnn.deterministic = True
             torch.backends.cudnn.benchmark = False
--- a/ignite/engine/deterministic.py
+++ b/ignite/engine/deterministic.py
@@ -1,4 +1,3 @@
-import os
 import random
 import warnings
 from collections import OrderedDict
--- a/ignite/engine/deterministic.py
+++ b/ignite/engine/deterministic.py
@@ -1,3 +1,4 @@
+import os
 import random
 import warnings
 from collections import OrderedDict
@@ -192,7 +193,14 @@ class DeterministicEngine(Engine):
     def _init_run(self) -> None:
         self.state.seed = int(torch.randint(0, int(1e9), (1,)).item())
         if torch.cuda.is_available():
-            torch.backends.cudnn.deterministic = True
+            if hasattr(torch, "use_deterministic_algorithms"):
+                # according to https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility
+                # CUBLAS_WORKSPACE_CONFIG must be set to let cuBLAS behave deterministic.
+                # **the behavior is expected to change in a future release of cuBLAS**.
+                os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
+                torch.use_deterministic_algorithms(True)
+            else:
+                torch.backends.cudnn.deterministic = True
             torch.backends.cudnn.benchmark = False
 
     def _setup_engine(self) -> None:
--- a/ignite/utils.py
+++ b/ignite/utils.py
@@ -236,9 +236,6 @@ def manual_seed(seed: int) -> None:
     random.seed(seed)
     torch.manual_seed(seed)
 
-    if torch.cuda.is_available():
-        torch.cuda.manual_seed_all(seed)
-
     try:
         import torch_xla.core.xla_model as xm
 


louis-she:
@vfdev-5 could you take a look at the failed tests?

For the RL failed examples, I can not reproduce them on my own machine.

vfdev-5:
Thanks for the PR, @louis-she !
I'll check in details a bit later the failures. Seeing one job RL failure complaining about seed arg, it seems that this could be related to gym version...

louis-she:
> Thanks for the PR, @louis-she ! I'll check in details a bit later the failures. Seeing one job RL failure complaining about seed arg, it seems that this could be related to gym version...

yes, I think it should be related to the version. I upgraded to the latest version of gym `0.26.2` but still can not reproduce the error.

vfdev-5:
Do you check with python 3.7 ?
Can you please see in the CI which version we are installing? Maybe they simply dropped support for py37

louis-she:
I'm using `3.7.10`, the CI uses `3.7.15`. I can't see the version of gym CI use cause the `pip install ... -qq` in the workflow. https://github.com/pytorch/ignite/blob/master/.github/workflows/unit-tests.yml#L90

Can we remove the `-qq` of `pip install` for easily debug?

vfdev-5:
Yes, let's remove -qq . I propose to create a separate PR for CI fix. For Neptune logger fix we can ping one of their folks for review

louis-she:
OK, then I'll create another PR to remove the `-qq` in CI. It'll be great if neptune guys could review the codes.

ignite/engine/deterministic.py
```diff
@@ -192,7 +193,14 @@ def state_dict(self) -> OrderedDict:
     def _init_run(self) -> None:
         self.state.seed = int(torch.randint(0, int(1e9), (1,)).item())
         if torch.cuda.is_available():
-            torch.backends.cudnn.deterministic = True
+            if hasattr(torch, "use_deterministic_algorithms"):
+                # according to https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility
+                # CUBLAS_WORKSPACE_CONFIG must be set to let cuBLAS behave deterministic.
+                # **the behavior is expected to change in a future release of cuBLAS**.
+                os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
```
vfdev-5, line 200: I'm not fan of doing that in ignite. If this call is necessary, it should be done by pytorch...
Reading the docs:

> set the debug environment variable CUBLAS_WORKSPACE_CONFIG to ":16:8" (may limit overall performance) or ":4096:8" (will increase library footprint in GPU memory by approximately 24MiB).

I do not think that we want to set a debug env variable. 

@louis-she let's remove that.

ignite/engine/deterministic.py
```diff
@@ -192,7 +193,14 @@ def state_dict(self) -> OrderedDict:
...
+                torch.use_deterministic_algorithms(True)
```
vfdev-5, line 201: Maybe, we should set `warn_only=True` such that we do not break previous code but only raise warnings about non-deterministic implementation.

vfdev-5 (approved):
LGTM, thanks @louis-she

vfdev-5:
@louis-she you were right, gpu tests are failing for deterministic engine:
```
>       return torch.matmul(kernel_x.t(), kernel_y)  # (kernel_size, 1) * (1, kernel_size)
E       RuntimeError: Deterministic behavior was enabled with either `torch.use_deterministic_algorithms(True)` or `at::Context::setDeterministicAlgorithms(true)`, but this operation is not deterministic because it uses CuBLAS and you have CUDA >= 10.2. To enable deterministic behavior in this case, you must set an environment variable before running your PyTorch application: CUBLAS_WORKSPACE_CONFIG=:4096:8 or CUBLAS_WORKSPACE_CONFIG=:16:8. For more information, go to https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility

=========================== short test summary info ============================
FAILED tests/ignite/engine/test_deterministic.py::test_gradients_on_resume_on_cuda - RuntimeError: Deterministic behavior was enabled with either `torch.use_det...
FAILED tests/ignite/metrics/test_ssim.py::test_ssim[shape0-7-False-True-cuda] - RuntimeError: Deterministic behavior was enabled with either `torch.use_det...
FAILED tests/ignite/metrics/test_ssim.py::test_ssim[shape1-11-True-False-cuda] - RuntimeError: Deterministic behavior was enabled with either `torch.use_det...
FAILED tests/ignite/metrics/gan/test_utils.py::test_device_mismatch_cuda - RuntimeError: Deterministic behavior was enabled with either `torch.use_det...
===== 4 failed, 20 passed, 5 skipped, 1920 deselected, 1 warning in 22.48s =====

Exited with code exit status 1
```
- https://app.circleci.com/pipelines/github/pytorch/ignite/2847/workflows/3e5bf1e8-de29-4dba-a70c-bd8bd1e0f139/jobs/8032
- https://github.com/pytorch/ignite/actions/runs/3354755842/jobs/5558524355

louis-she:
looks like it was related to multiple GPU nodes. Let me look at this.

louis-she:
I'm not sure if this is a bug of `PyTorch`, but I can reproduce this with this very straightforward snippet:

```python
import torch
torch.use_deterministic_algorithms(True, warn_only=True)
assert torch.is_deterministic_algorithms_warn_only_enabled()
torch.nn.Linear(10, 10, device="cuda")((torch.rand(1, 10, device="cuda")))
```

raise `RuntimeError` ,

```
  File "test.py", line 4, in <module>
    torch.nn.Linear(10, 10, device="cuda")((torch.rand(1, 10, device="cuda")))
  File "/environment/miniconda3/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/environment/miniconda3/lib/python3.7/site-packages/torch/nn/modules/linear.py", line 114, in forward
    return F.linear(input, self.weight, self.bias)
RuntimeError: Deterministic behavior was enabled with either `torch.use_deterministic_algorithms(True)` or `at::Context::setDeterministicAlgorithms(true)`, but this operation is not deterministic because it uses CuBLAS and you have CUDA >= 10.2. To enable deterministic behavior in this case, you must set an environment variable before running your PyTorch application: CUBLAS_WORKSPACE_CONFIG=:4096:8 or CUBLAS_WORKSPACE_CONFIG=:16:8. For more information, go to https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility
```

The error will not be raised with `torch==1.13.0`, I have had a look at the source of torch, and it was different of the 2 versions.

v1.12.1 https://github.com/pytorch/pytorch/blob/v1.12.1/aten/src/ATen/Context.cpp#L126
v1.13.0 https://github.com/pytorch/pytorch/blob/v1.13.0/aten/src/ATen/Context.cpp#L142

For v1.12.1, there is no `if` statement about the warning flag, but v1.13.0 did has it.

vfdev-5:
@louis-she does your code sample show a warning with 1.13.0 and cuda 11.6 ?
In my case, it does not show anything ... Seems like it is also related to cuda version as 1.12.1 with cuda 11.6 does not raise any error...

louis-she:
Hmm seems like NVIDIA makes `torch.nn.Linear` deterministic in cu116. Another test snippet

```python
import torch
import torchvision

warn_only = False
torch.use_deterministic_algorithms(True, warn_only=warn_only)
if warn_only:
    assert torch.is_deterministic_algorithms_warn_only_enabled()

model = torchvision.models.swin_s().cuda()
model(torch.rand(2, 3, 224, 224, device="cuda"))
```

Here are some experiments result:

| torch version | warn_only | behavior |
| ------------- | ----- | --------- |
| `torch==1.12.1+cu113` | True | RuntimeError |
| `torch==1.12.1+cu113` | False | RuntimeError|
| `torch==1.12.1+cu116` | True | RuntimeError|
| `torch==1.12.1+cu116` | False | RuntimeError|
| `torch==1.13.0+cu116` | True | UserWarning | 
| `torch==1.13.0+cu116` | False | RuntimeError|

PR #2757, opened 2022-10-29
Fix failing CI

louis-she:
https://github.com/pytorch/ignite/pull/2756#issuecomment-1295771750

### DIFF
--- a/ignite/contrib/handlers/neptune_logger.py
+++ b/ignite/contrib/handlers/neptune_logger.py
@@ -187,7 +187,10 @@ class NeptuneLogger(BaseLogger):
 
         if kwargs.get("offline_mode", False):
             self.mode = "offline"
-            neptune.init(project_qualified_name="dry-run/project", backend=neptune.OfflineBackend())
+            neptune.init(
+                project_qualified_name="dry-run/project",
+                backend=neptune.legacy.internal.api_clients.offline_backend.OfflineBackend(),
+            )
         else:
             self.mode = "online"
             neptune.init(api_token=kwargs.get("api_token"), project_qualified_name=kwargs.get("project_name"))
--- a/ignite/contrib/handlers/neptune_logger.py
+++ b/ignite/contrib/handlers/neptune_logger.py
@@ -189,7 +189,7 @@ class NeptuneLogger(BaseLogger):
             self.mode = "offline"
             neptune.init(
                 project_qualified_name="dry-run/project",
-                backend=neptune.legacy.internal.api_clients.offline_backend.OfflineBackend(),
+                backend=neptune.OfflineBackend(),
             )
         else:
             self.mode = "online"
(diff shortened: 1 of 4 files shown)


vfdev-5:
Let's put Neptune fix here and try to fix RL example as well once we understand the root of the issue.

ignite/contrib/handlers/neptune_logger.py
```diff
@@ -187,7 +187,10 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
 
         if kwargs.get("offline_mode", False):
             self.mode = "offline"
-            neptune.init(project_qualified_name="dry-run/project", backend=neptune.OfflineBackend())
+            neptune.init(
```
vfdev-5, line 190: @Blaizzy can you check if this is a correct fix ?

vfdev-5:
@louis-she can you please disable neptune tests such that we could merge this PR. 

Do you think this version pin is necessary ?
> gym==0.26.2

louis-she:
the root cause of the 2 errors, one from `neptune-client` and one from `gym`, is the version resolve conflicts of the 2 packages.
If install with `neptune-client==0.16.11` `gym==0.26.2` , error will raised:

```
ERROR: Cannot install gym==0.26.2 and neptune-client==0.16.11 because these package versions have conflicting dependencies.

The conflict is caused by:
    neptune-client 0.16.11 depends on importlib-metadata<4; python_version < "3.8.0"
    gym 0.26.2 depends on importlib-metadata>=4.8.0; python_version < "3.10"

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

The pip install will resolve the conflicts by get the older version of `gym`, and that's why the CI failed with gym. The gym is resolved to version `0.20.0` (since we do not ping the version of gym). And that will lead to the `seed` error.

vfdev-5:
Thanks for the explanation !

I would suggest to temporarily comment out neptune dependency in requirements-dev.txt and add 
```python
pytest.skip("Temporarily skip neptune tests", allow_module_level=True)
```
into `https://github.com/pytorch/ignite/blob/master/tests/ignite/contrib/handlers/test_neptune_logger.py` to avoid failures. 

Once neptune folks updated our code to their latest API version and also fix the issue with `importlib-metadata` we can reenable the tests. What do you think ?

louis-she:
I fixed the version of `neptune` to the last version before they did the legacy refactoring and leave the neptune test as it was in the past. Thus `gym` can be automatically resolved to the latest version `0.26.2` and pass the CI.

vfdev-5 (approved):
Thanks @louis-she !

ignite/contrib/handlers/neptune_logger.py
Blaizzy: I will take a look at it tomorrow morning :)

PR #2778, opened 2022-11-16
Fix MyPy issue

sadra-barikbin:
Description:
Using new MyPy release(0.991) raises some errors. This PR attempts to resolve them.

### DIFF
--- a/ignite/contrib/handlers/base_logger.py
+++ b/ignite/contrib/handlers/base_logger.py
@@ -9,7 +9,7 @@ import torch
 import torch.nn as nn
 from torch.optim import Optimizer
 
-from ignite.engine import Engine, Events, EventsList, State
+from ignite.engine import Engine, EventEnum, Events, EventsList, State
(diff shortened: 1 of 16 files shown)


sadra-barikbin:
@vfdev-5 , in `ignite.handlers.base_logger.BaseHandler` and its children why is `Union[str, Events]` is being used as type annotation for the second argument in `__call__` method and not `Union[str, EventEnum]`?

vfdev-5:
Events type is a bit tricky and we need unify the typehints at some point. As a rule of thumb, we have 
- EventEnum = base class for events: built-in and custom
- Events = built-in event for the Engine
- EventsList = list of events
- str = string as event
 
Let's first fix mypy issues and in a follow-up PR we can tackle typehint for events.

vfdev-5 (approved):

vfdev-5:
As failing tests are unrelated, merging this PR

PR #2780, opened 2022-11-16
Fixed torch LRScheduler issue and fixed CI

### LINKED ISSUE #2773: Use LRScheduler instead of _LRScheduler
## Update code to be compatible with LRScheduler too

I recently exposed LRScheduler to be public (instead of _LRScheduler) in https://github.com/pytorch/pytorch/pull/88503. I see several instance checks in this current repo that use _LRScheduler, and I believe these should be updated to be LRScheduler.

## Why?
This would be a good step in the right direction. Also, downstream users who subscribe to new LRScheduler (and not _LRScheduler) now will run into issues.

## Alternatives
Maybe the best thing to do for forward and backward compatibility is to accept LRScheduler whenever there is _LRScheduler too.

vfdev-5:
Fixes #2773

Description:
- Fixed torch LRScheduler issue and fixed CI

Check list:

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -96,7 +96,9 @@ jobs:
           bash ./tests/run_code_style.sh lint
 
       - name: Run Mypy
-        if: ${{ matrix.os == 'ubuntu-latest'}}
+        # https://github.com/pytorch/ignite/pull/2780
+        # 
+        if: ${{ matrix.os == 'ubuntu-latest' && matrix.pytorch-channel == 'pytorch-nightly'}}
         run: |
           bash ./tests/run_code_style.sh mypy
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -342,6 +342,7 @@ nitpick_ignore = [
     ("py:class", "torch.utils.data.sampler.BatchSampler"),
     ("py:class", "torch.cuda.amp.grad_scaler.GradScaler"),
     ("py:class", "torch.optim.lr_scheduler._LRScheduler"),
+    ("py:class", "torch.optim.lr_scheduler.LRScheduler"),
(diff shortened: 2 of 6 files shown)


vfdev-5:
There is again mypy inconsistency between stable and nightly pytorch:
```
# STABLE PyTorch:

mypy --config-file mypy.ini
ignite/handlers/param_scheduler.py:18: error: Module "torch.optim.lr_scheduler" has no attribute "LRScheduler"; maybe "_LRScheduler"?  [attr-defined]
        from torch.optim.lr_scheduler import LRScheduler as PyTorchLRScheduler
        ^
ignite/handlers/param_scheduler.py:882: error: Unused "type: ignore" comment
            self.lr_scheduler._get_lr_called_within_step = True  # type: ignore[attr-defined]
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ignite/handlers/param_scheduler.py:884: error: Unused "type: ignore" comment
            self.lr_scheduler._get_lr_called_within_step = False  # type: ignore[attr-defined]
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ignite/contrib/engines/common.py:13: error: Module "torch.optim.lr_scheduler" has no attribute "LRScheduler"; maybe "_LRScheduler"?  [attr-defined]
        from torch.optim.lr_scheduler import LRScheduler as PyTorchLRScheduler
        ^
Found 4 errors in 2 files (checked 102 source files)
```

Will disable mypy check on stable version.

vfdev-5:
@sadra-barikbin I merged the PR to fix the CI. If you have any comment here, let's address them in a follow-up PR.

PR #2786, opened 2022-11-26
update set-output usage[WIP]

### LINKED ISSUE #2784: [CI] update set-output usage
Use
```
- name: Set output
  run: echo "{name}={value}" >> $GITHUB_OUTPUT
```
instead of
```
- name: Set output
  run: echo "::set-output name={name}::{value}"
```
in our GHA yml files.

- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

What to do:
- comment here to be assigned to the issue
- find all files where we use "::set-output"
- replace set-output by what GH suggests (see above)
- send a PR

guptaaryan16:
Fixes #2784 

Description: I updated all the old set-output cases with new syntax as specified in the issue and the blog.

Check list:

### DIFF
--- a/.github/workflows/binaries-nightly-release.yml
+++ b/.github/workflows/binaries-nightly-release.yml
@@ -10,7 +10,7 @@ jobs:
   build-publish:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Setup Miniconda
         uses: conda-incubator/setup-miniconda@v2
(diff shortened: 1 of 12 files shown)


vfdev-5:
Thanks for the PR @guptaaryan16 but I think it's not a correct way, see the error : https://github.com/pytorch/ignite/actions/runs/3553020030/jobs/5968378813#step:6:12

Please check the GitHub docs how to save / load output. Thanks

guptaaryan16:
Thanks you solved the issue I was stuck on

vfdev-5 (approved):
Thanks for the PR @guptaaryan16 , lgtm !

PR #2787, opened 2022-11-27
Fix Flake8 issues

sadra-barikbin:


vfdev-5 (approved):
Thanks for the fix @sadra-barikbin !

PR #2788, opened 2022-11-27
Fix some bugs in `conftest.py`, one in `_destroy_dist_context` and the other in `distributed`

sadra-barikbin:
Description:
1. One step in NCCL and Gloo single node distributed settings is to setup a port shared by the processes. This is done using a temporary file in which the rank zero writes the port number and the others read it. Currently this file isn't deleted at the end, so in the case a process other than rank 0 opens the file first, it reads the port being used in the past which results in timeout error in all ranks. This bug is not encountered in GitHub CI because disk resources are newly allocated, so the temporary file does not exist.
2. To reduce setup and teardown time, I had set the scope of `distributed` fixture to `module` so as to have just one distributed setup and teardown for all distributed tests in a module but this caused some non-distributed tests to fail. It turned out that the teardown of the last instance of the fixture (we have an instance for `nccl`, one for `gloo`, one for `gloo_cpu` and...) takes place after running the last test of the module, even if it has not requested the fixture. Running the snippet below which is originally from [pytest website](https://docs.pytest.org/en/7.0.x/how-to/fixtures.html#automatic-grouping-of-tests-by-fixture-instances) shows this fact:

```python
import pytest


@pytest.fixture(scope="module", params=["mod1", "mod2"])
def modarg(request):
    param = request.param
    print("  SETUP modarg", param)
    yield param
    print("  TEARDOWN modarg", param)


@pytest.fixture(scope="function", params=[1, 2])
def otherarg(request):
    param = request.param
    print("  SETUP otherarg", param)
    yield param
    print("  TEARDOWN otherarg", param)


def test_0(otherarg):
    print("  RUN test0 with otherarg", otherarg)


def test_1(modarg):
    print("  RUN test1 with modarg", modarg)

def test_11():
    print("hi")

def test_2(otherarg, modarg):
    print("  RUN test2 with otherarg {} and modarg {}".format(otherarg, modarg))

def test_3():
    print("there!")
```
Output:
image

So I changed the scope of the fixture from `module` to `function` to solve the issue.

### DIFF
--- a/tests/ignite/conftest.py
+++ b/tests/ignite/conftest.py
@@ -457,7 +457,6 @@ def pytest_pyfunc_call(pyfuncitem: pytest.Function) -> None:
 
                 test_func(**kwargs)
 
-
                 hvd.shutdown()
 
             try:
--- a/tests/ignite/conftest.py
+++ b/tests/ignite/conftest.py
@@ -457,11 +457,6 @@ def pytest_pyfunc_call(pyfuncitem: pytest.Function) -> None:
 
                 test_func(**kwargs)
 
-                # Added a sleep to avoid flaky failures on circle ci
-                # Sometimes a rank is terminated before final collective
-                # op is finished.
-                # https://github.com/pytorch/ignite/pull/2357
-                time.sleep(2)
 
                 hvd.shutdown()
 
--- a/tests/ignite/conftest.py
+++ b/tests/ignite/conftest.py
@@ -364,7 +364,6 @@ is_xla_single_device_stash_key = pytest.StashKey[bool]()
 
 
 @pytest.fixture(
-    scope="module",
     params=[
         pytest.param("nccl", marks=[pytest.mark.distributed, skip_if_has_not_native_dist_support, skip_if_no_gpu]),
         pytest.param("gloo_cpu", marks=[pytest.mark.distributed, skip_if_has_not_native_dist_support]),
--- a/tests/ignite/conftest.py
+++ b/tests/ignite/conftest.py
@@ -106,10 +106,14 @@ def _create_dist_context(dist_info, lrank):
 
 def _destroy_dist_context():
 
-    dist.barrier()
-
     if dist.get_rank() == 0:
-        Path("/tmp/free_port").unlink(True)
+        # To support Python 3.7; Otherwise we could do `.unlink(missing_ok=True)`
+        try:
+            Path("/tmp/free_port").unlink()
+        except FileNotFoundError:
+            pass
+
+    dist.barrier()
 
     dist.destroy_process_group()
 
--- a/tests/ignite/conftest.py
+++ b/tests/ignite/conftest.py
@@ -107,6 +107,10 @@ def _create_dist_context(dist_info, lrank):
 def _destroy_dist_context():
 
     dist.barrier()
+
+    if idist.get_rank() == 0:
+        Path("/tmp/free_port").unlink(True)
+
     dist.destroy_process_group()
 
     from ignite.distributed.utils import _SerialModel, _set_model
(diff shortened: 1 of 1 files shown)


tests/ignite/conftest.py
```diff
@@ -107,6 +107,10 @@ def _create_dist_context(dist_info, lrank):
 def _destroy_dist_context():
 
     dist.barrier()
+
+    if idist.get_rank() == 0:
```
vfdev-5, line 111:
```suggestion
    if dist.get_rank() == 0:
```

vfdev-5:
@sadra-barikbin do you think it would make sense and can help if we were using a class with fixture, something like
```python

@pytest.mark.usefixtures("cleandir")
class TestDirectoryInit:
    def test_cwd_starts_empty(self):
        assert os.listdir(os.getcwd()) == []
        with open("myfile", "w") as f:
            f.write("hello")

    def test_cwd_again_starts_empty(self):
        assert os.listdir(os.getcwd()) == []
```
(in this example fixture is called for each function, but maybe there is a way to set it up once per class)

sadra-barikbin:
> @sadra-barikbin do you think it would make sense and can help if we were using a class with fixture, something like
> 
> ```python
> @pytest.mark.usefixtures("cleandir")
> class TestDirectoryInit:
>     def test_cwd_starts_empty(self):
>         assert os.listdir(os.getcwd()) == []
>         with open("myfile", "w") as f:
>             f.write("hello")
> 
>     def test_cwd_again_starts_empty(self):
>         assert os.listdir(os.getcwd()) == []
> ```
> 
> (in this example fixture is called for each function, but maybe there is a way to set it up once per class)

Yes we could and it helps. Teardown takes place exactly after the last test of the class:
```python
import pytest


@pytest.fixture(scope="class", params=[1, 2])
def cls_fixt(request):
    param = request.param
    print("  SETUP cls_fixt", param)
    yield param
    print("  TEARDOWN cls_fixt", param)


@pytest.fixture(scope="function", params=[1, 2])
def fun_fixt(request):
    param = request.param
    print("  SETUP fun_fixt", param)
    yield param
    print("  TEARDOWN fun_fixt", param)

@pytest.mark.usefixtures("cls_fixt")
class Test2:
    def test_1(self):
        print("hello")
    def test_4(self, fun_fixt):
        print("world")

def test_2():
    print("hi")

def test_3(fun_fixt):
    print("there!")
```
Output:
image

By the way, current behavior of `module` scope does not seem to be correct or consistent, considering the behavior for `class` and `function` scopes. Am I wrong?

sadra-barikbin:
In pytest they've considered this behavior, an inefficiency:
https://github.com/pytest-dev/pytest/issues/3806

sadra-barikbin:
The latest change that I introduced to distributed fixture is due to an error that came up when I removed `scope='module'`.

vfdev-5:
@sadra-barikbin CI is still failing somehow

tests/ignite/conftest.py
```diff
@@ -430,17 +437,65 @@ def distributed(request, local_rank, world_size):
...
+                if torch.cuda.is_available():
+                    torch.cuda.set_device(lrank)
+
+                test_func(**kwargs)
+
+                # Added a sleep to avoid flaky failures on circle ci
+                # Sometimes a rank is terminated before final collective
+                # op is finished.
+                # https://github.com/pytorch/ignite/pull/2357
+                time.sleep(2)
```
vfdev-5, line 464: Maybe, we can remove this sleep which will reduce CI time.
```suggestion
```

tests/ignite/conftest.py
```diff
@@ -430,17 +437,60 @@ def distributed(request, local_rank, world_size):
...
+            def hvd_worker():
+                import horovod.torch as hvd
+
+                hvd.init()
+                lrank = hvd.local_rank()
+                if torch.cuda.is_available():
+                    torch.cuda.set_device(lrank)
+
+                test_func(**kwargs)
+
```
vfdev-5, line 459:
```suggestion
```

vfdev-5 (approved):
Thanks @sadra-barikbin , lgtm !

sadra-barikbin:
Won't you merge it?

PR #2794, opened 2022-12-07
[BC-breaking] Align ssim with psnr, metric output is tensor

crj1998:
Description:
For the current version, the output of psnr and ssim are differnt type, where the type of psnr.comute() is `torch.Tensor` but the ssim.compute() is `float`
```
psnr = PSNR(data_range=1.0)
psnr.update((torch.rand([4, 3, 16, 16]),torch.rand([4, 3, 16, 16])))
psnr = psnr.compute()
print(psnr, type(psnr))

ssim = SSIM(data_range=1.0, gaussian=True, sigma=1.5)
ssim.update((torch.rand([4, 3, 16, 16]), torch.rand([4, 3, 16, 16])))
ssim = ssim.compute()

print(ssim, type(ssim))
```
 This is because in the ssim.py, there used `.item()`

https://github.com/pytorch/ignite/blob/cdf7b47802c8f8f65bd1bfef933e9154d696e317/ignite/metrics/psnr.py#L120-L124

while
https://github.com/pytorch/ignite/blob/cdf7b47802c8f8f65bd1bfef933e9154d696e317/ignite/metrics/ssim.py#L182-L186

The two metrics PSNR and SSIM are always used together. So I think it's better for the same input to have the same type of output. This is intuitive.

Check list:
I dont modify any unit test.

### DIFF
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -180,7 +180,7 @@ class SSIM(Metric):
         self._num_examples += y.shape[0]
 
     @sync_all_reduce("_sum_of_ssim", "_num_examples")
-    def compute(self) -> float:
+    def compute(self) -> torch.Tensor:
         if self._num_examples == 0:
             raise NotComputableError("SSIM must have at least one example before it can be computed.")
-        return (self._sum_of_ssim / self._num_examples).item()
+        return self._sum_of_ssim / self._num_examples


vfdev-5:
@crj1998 thanks a lot for the PR ! I wonder about the common output type: tensor or float
Currently we have : 
- PSNR -> tensor
- SSIM -> float
and this PR makes:
- PSNR -> tensor
- SSIM -> tensor
and I wonder if it would be better to output both floats :
- PSNR -> float
- SSIM -> float

@ydcjeff do you remember if there was a reason to output float value for SSIM and tensor for PSNR ?

crj1998:
> @crj1998 thanks a lot for the PR ! I wonder about the common output type: tensor or float Currently we have :
> 
> * PSNR -> tensor
> * SSIM -> float
>   and this PR makes:
> * PSNR -> tensor
> * SSIM -> tensor
>   and I wonder if it would be better to output both floats :
> * PSNR -> float
> * SSIM -> float
> 
> @ydcjeff do you remember if there was a reason to output float value for SSIM and tensor for PSNR ?
> 
> cc @sadra-barikbin what do you think ?


I think that since **ignite** is an extension of pytorch, it's best to return the `torch.Tensor` type , because converting `torch.Tensor` to `float`  only require  an `item()` operation.

vfdev-5 (approved):
LGTM, thanks @crj1998 !

vfdev-5:
@crj1998 I merged your PR however running GPU tests there are few related failures : https://app.circleci.com/pipelines/github/pytorch/ignite/2900/workflows/45b4d299-599c-47e7-b0f6-121b0e4b5988/jobs/8072/parallel-runs/0/steps/0-107

Can you please check and fix them ?

crj1998:
> @crj1998 I merged your PR however running GPU tests there are few related failures : https://app.circleci.com/pipelines/github/pytorch/ignite/2900/workflows/45b4d299-599c-47e7-b0f6-121b0e4b5988/jobs/8072/parallel-runs/0/steps/0-107
> 
> Can you please check and fix them ?

Yeah, as i said before, I do noting about unit test, so the previous test examples which return float will be failed. I will update unit test in a new pr.

PR #2795, opened 2022-12-08
Update ssim unit test

crj1998:
Fixes  https://github.com/pytorch/ignite/issues/2794#issuecomment-1342236316

Description:
add unit test for ssim with return type `torch.Tensor`.

Check list:


vfdev-5:
@crj1998 thanks for the fix ! In addition to that there is a bug in how tests are scheduled, curiously this does not run on CPU but should run.

vfdev-5 (approved):
LGTM, thanks @crj1998 !

PR #2796, opened 2022-12-08
Fixed device issue with metric tests SSIM, updated PSNR

vfdev-5:
Description:
- Fixed device issue with metric tests SSIM. Context: https://github.com/pytorch/ignite/pull/2795#issuecomment-1342430076
- Updated PSNR

Check list:

### DIFF
--- a/tests/ignite/metrics/test_ssim.py
+++ b/tests/ignite/metrics/test_ssim.py
@@ -9,6 +9,8 @@ import ignite.distributed as idist
 from ignite.exceptions import NotComputableError
 from ignite.metrics import SSIM
 
+from tests.ignite import cpu_and_maybe_cuda
+
 
 def test_zero_div():
     ssim = SSIM(data_range=1.0)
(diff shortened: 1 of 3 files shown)


tests/ignite/metrics/test_ssim.py
```diff
@@ -9,6 +9,8 @@
 from ignite.exceptions import NotComputableError
 from ignite.metrics import SSIM
 
+from tests.ignite import cpu_and_maybe_cuda
```
sadra-barikbin, line 12: Isn't better to be something like `available_devices`?  This gives the feeling of actually importing cpu and maybe cuda!

vfdev-5: OK, let's rename it. Thanks for the feedback

sadra-barikbin: It could also be a fixture to be used just by declaring as argument instead of writing `@pytest.mark.parametrize...`.

PR #2790, opened 2022-12-01
updated to latest version of actions for set-python and codecov

RishiKumarRay:
Fixes #2789

Description: updated to latest version for actions/setup-python@v2 and codecov/codecov-action@v1

Check list:

- [ ] New tests are added (if a new feature is added)
- [ ] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #2798, opened 2022-12-09
Fixed issue with Horovod cpu tests

sadra-barikbin:
Description:

- Fixed issue with Horovod cpu tests

Context:
Since #2790 Horovod cpu job skips horovod test: 
```
tests/ignite/metrics/test_psnr.py::test_distrib_input_float[horovod] SKIPPED (Skip if no Horovod dist support) [ 86%]
```
https://github.com/pytorch/ignite/actions/runs/3654919549/jobs/6175755766#step:9:1770

### DIFF
--- a/.github/workflows/hvd-tests.yml
+++ b/.github/workflows/hvd-tests.yml
@@ -69,10 +69,6 @@ jobs:
         with:
           target_dir: /tmp
 
-      - name: Check if Horovod actually exists
-        shell: bash -l {0}
-        run: python -c "import horovod.torch as hvd"
-
       - name: Run Tests
         shell: bash -l {0}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/hvd-tests.yml
```diff
@@ -72,6 +72,10 @@ jobs:
         with:
           target_dir: /tmp
 
+      - name: Check if Horovod actually exists
+        shell: python
```
vfdev-5, line 76: Here you should keep `shell: bash -l {0}` otherwise conda env wont be activated.
Try to run `python -c "import horovod.torch as hvd"` to see what it tells

vfdev-5:
@sadra-barikbin Can you update PR's title and description

vfdev-5 (approved):
Thanks @sadra-barikbin

PR #2800, opened 2022-12-11
Refactor `EpochMetric` and make it idempotent

sadra-barikbin:
By adding an assertion to check recomputing metric does not change its result, `_test_distrib_integration` fails.


vfdev-5:
Seems like there is a bug here: https://github.com/pytorch/ignite/blob/7c8d801b68c2ed5bc0c030e124df6bb4c027197a/ignite/metrics/epoch_metric.py#L148-L152

`_prediction_tensor` and `_target_tensor` are temporary tensors but `self._is_reduced` is set to True as for other metrics where `@sync_all_reduce` decorator updates metric's attributes and thus we avoid duplicated all_reduce calls.

ignite/metrics/epoch_metric.py
```diff
@@ -136,31 +135,32 @@ def update(self, output: Tuple[torch.Tensor, torch.Tensor]) -> None:
...
+                # All gather across all processes
+                _prediction_tensor = cast(torch.Tensor, idist.all_gather(_prediction_tensor))
+                _target_tensor = cast(torch.Tensor, idist.all_gather(_target_tensor))
 
-        if ws > 1 and not self._is_reduced:
-            # All gather across all processes
-            _prediction_tensor = cast(torch.Tensor, idist.all_gather(_prediction_tensor))
-            _target_tensor = cast(torch.Tensor, idist.all_gather(_target_tensor))
-        self._is_reduced = True
+            self._result = 0.0
```
vfdev-5, line 152: I think we should define `self._result = None` in `reset` method and use `if self._result is None` to run result computations. Also `reinit__is_reduced` should set `self._result = None` if attribute exists.

sadra-barikbin: Why not to use `self._is_reduced` in the `if` statement in `EpochMetric`? Anyway we have to set it to `True` at the end of the statement.

vfdev-5: IMO, `self._is_reduced` is not applicable here and everywhere once https://github.com/pytorch/ignite/pull/2803 is merged. The role of `self._is_reduced` was to avoid unnecessary (even faulty) dist reductions as we were overwriting internal counters. In #2803 we find out that this way of computing metrics has a bug. I think we should avoid using `self._is_reduced` at all

ignite/metrics/epoch_metric.py
```diff
@@ -136,31 +136,32 @@ def update(self, output: Tuple[torch.Tensor, torch.Tensor]) -> None:
...
                 warnings.warn(f"Probably, there can be a problem with `compute_fn`:\n {e}.", EpochMetricWarning)
 
-    def compute(self) -> Any:
+    def compute(self) -> float:
         if len(self._predictions) < 1 or len(self._targets) < 1:
             raise NotComputableError("EpochMetric must have at least one example before it can be computed.")
 
-        _prediction_tensor = torch.cat(self._predictions, dim=0)
-        _target_tensor = torch.cat(self._targets, dim=0)
+        if self._result is None:
```
vfdev-5, line 143: You have to also reset `self._result` in `@reinit__is_reduced` otherwise on futher `update`s we will output previously computed result instead of new one

ignite/metrics/metric.py
```diff
@@ -595,6 +595,8 @@ def reinit__is_reduced(func: Callable) -> Callable:
     def wrapper(self: Metric, *args: Any, **kwargs: Any) -> None:
         func(self, *args, **kwargs)
         self._is_reduced = False
+        if "_result" in self.__dict__:
+            self._result = None  # type: ignore[attr-defined]
```
sadra-barikbin, line 599: I've reset `_result` here.

ignite/contrib/metrics/precision_recall_curve.py
```diff
@@ -78,41 +78,44 @@ def __init__(
...
+                precision, recall, thresholds = None, None, None
+
+            if ws > 1:
+                # broadcast result to all processes
+                precision = idist.broadcast(precision, src=0, safe_mode=True)
+                recall = idist.broadcast(recall, src=0, safe_mode=True)
+                thresholds = idist.broadcast(thresholds, src=0, safe_mode=True)
+
+            self._result = (precision, recall, thresholds)  # type: ignore[assignment]
+            self._is_reduced = True
```
vfdev-5, line 119:
```suggestion
```

ignite/contrib/metrics/precision_recall_curve.py
```diff
@@ -78,41 +78,44 @@ def __init__(
...
+                _target_tensor = cast(torch.Tensor, idist.all_gather(_target_tensor))
+
+            if idist.get_rank() == 0:
+                # Run compute_fn on zero rank only
+                precision, recall, thresholds = cast(Tuple, self.compute_fn(_prediction_tensor, _target_tensor))
+                precision = torch.tensor(precision)
+                recall = torch.tensor(recall)
+                # thresholds can have negative strides, not compatible with torch tensors
+                # https://discuss.pytorch.org/t/negative-strides-in-tensor-error/134287/2
+                thresholds = torch.tensor(thresholds.copy())
```
vfdev-5, line 108: @sadra-barikbin set proper devices here as it is don for `RocCurve`. I'll start GPU tests for this PR

ignite/metrics/epoch_metric.py
```diff
@@ -136,31 +136,32 @@ def update(self, output: Tuple[torch.Tensor, torch.Tensor]) -> None:
...
-            # Run compute_fn on zero rank only
-            result = self.compute_fn(_prediction_tensor, _target_tensor)
+            if ws > 1:
+                # broadcast result to all processes
+                self._result = cast(float, idist.broadcast(self._result, src=0))
 
-        if ws > 1:
-            # broadcast result to all processes
-            result = cast(float, idist.broadcast(result, src=0))
+            self._is_reduced = True
```
vfdev-5, line 162:
```suggestion
```

ignite/contrib/metrics/precision_recall_curve.py
```diff
@@ -101,11 +101,11 @@ def compute(self) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:  # type: i
...
                 # Run compute_fn on zero rank only
                 precision, recall, thresholds = cast(Tuple, self.compute_fn(_prediction_tensor, _target_tensor))
-                precision = torch.tensor(precision)
-                recall = torch.tensor(recall)
+                precision = torch.tensor(precision, device=_prediction_tensor.device)
+                recall = torch.tensor(recall, device=_prediction_tensor.device)
                 # thresholds can have negative strides, not compatible with torch tensors
                 # https://discuss.pytorch.org/t/negative-strides-in-tensor-error/134287/2
-                thresholds = torch.tensor(thresholds.copy())
+                thresholds = torch.tensor(thresholds, device=_prediction_tensor.device)
```
vfdev-5, line 108: This will fail

vfdev-5 (approved):
LGTM, thanks @sadra-barikbin

PR #2803, opened 2022-12-21
Fix `sync_all_reduce` to consider update->compute->update case

sadra-barikbin:
Problem Description:
We get wrong result in the situation below, given that we are in a distributed config and `SomeMetric` has decorated its `compute` method with `sync_all_reduce`:
```python
metric = SomeMetric() 
metric.update(y_pred, y)
correct_result = metric.compute()
metric.update(y_pred, y)
wrong_result = metric.compute()
```
This is the case because `sync_all_reduce` changes the state attributes to the accumulated values in all ranks, making them ready to call `compute` and calling `update` again doesn't make sense.

Solution (By @vfdev-5 ):
In `sync_all_reduce`, store the state attributes in temporary variables. Then do the collective ops on the state attributes. After that, call the decorated function. Finally, restore the state attributes and return the result of the decorated function.


vfdev-5 (commented):
@sadra-barikbin let's write a test that fails on master and is fixed with this PR. In the test we have to check Metric attributes of type: tensor and scalar

vfdev-5:
@sadra-barikbin the code is definitely broken, no point to update the branch every time

tests/ignite/metrics/test_accuracy.py
```diff
@@ -275,16 +275,17 @@ def _test(metric_device):
...
         # gather y_pred, y
         y_pred = idist.all_gather(y_pred)
         y = idist.all_gather(y)
 
         np_y_pred = to_numpy_multilabel(y_pred.cpu())  # (N, C, H, W, ...) -> (N * H * W ..., C)
         np_y = to_numpy_multilabel(y.cpu())  # (N, C, H, W, ...) -> (N * H * W ..., C)
         assert acc._type == "multilabel"
-        n = acc._num_examples
         res = acc.compute()
-        assert n * idist.get_world_size() == acc._num_examples
```
vfdev-5, line 287: I think you should keep this test with the following change:
```python
assert n == acc._num_examples
```

tests/ignite/metrics/test_accuracy.py
```diff
@@ -306,14 +310,7 @@ def _test(metric_device):
...
 
         assert acc._type == "multilabel"
-        n = acc._num_examples
         res = acc.compute()
-        assert n * idist.get_world_size() == acc._num_examples
-        assert isinstance(res, float)
-        assert accuracy_score(np_y, np_y_pred) == pytest.approx(res)
-        # check that result is not changed
-        res = acc.compute()
-        assert n * idist.get_world_size() == acc._num_examples
```
vfdev-5, line 316: Revert this change and adapt to the new behaviour

tests/ignite/metrics/test_accuracy.py
```diff
@@ -342,9 +342,7 @@ def _test(metric_device):
         np_y = to_numpy_multilabel(y.cpu())  # (N, C, L, ...) -> (N * L ..., C)
 
         assert acc._type == "multilabel"
-        n = acc._num_examples
         res = acc.compute()
-        assert n * idist.get_world_size() == acc._num_examples
```
vfdev-5, line 347: Same here

vfdev-5:
@sadra-barikbin let's also remove `self._is_reduced` as now we do not need it anymore.

sadra-barikbin:
Shall I remove `reinit__is_reduced` as well?

sadra-barikbin:
Or you keep it for BC and just set `_result` to none if it exists?

vfdev-5:
> Or you keep it for BC and just set `_result` to none if it exists?

Yes, we should keep it for BC and also for reseting cached value. Later we can rename it with deprecation cycles etc.

ignite/metrics/precision.py
```diff
@@ -138,18 +139,13 @@ def compute(self) -> Union[torch.Tensor, float]:
...
-            if self._average == "weighted":
-                self._weight = idist.all_reduce(self._weight)  # type: ignore[assignment]
-            self._is_reduced: bool = True
 
         fraction = self._numerator / (self._denominator + (self.eps if self._average != "samples" else 0))
 
         if self._average == "weighted":
-            sum_of_weights = cast(torch.Tensor, self._weight).sum() + self.eps
-            return ((fraction @ self._weight) / sum_of_weights).item()  # type: ignore
+            _weight = idist.all_reduce(self._weight)
```
vfdev-5, line 146: maybe should clone here:
```suggestion
            _weight = idist.all_reduce(self._weight.clone())
```

ignite/metrics/metric.py
```diff
@@ -574,8 +574,10 @@ def another_wrapper(self: Metric, *args: Any, **kwargs: Any) -> Callable:
                             f"number or tensor but `{attr}` has type {type(t)}"
                         )
                     unreduced_attrs[attr] = t
-                    # Here `clone` is necessary since `idist.all_reduce` modifies `t` inplace in the case
+                    # Here `clone` is necessary since `idist.all_reduce` modifies `t` inp-lace in the case
```
vfdev-5, line 577:
```suggestion
                    # Here `clone` is necessary since `idist.all_reduce` modifies `t` inplace in the case
```

tests/ignite/metrics/test_accuracy.py
```diff
@@ -306,16 +311,14 @@ def _test(metric_device):
...
+        assert n == acc._num_examples
         assert isinstance(res, float)
         assert accuracy_score(np_y, np_y_pred) == pytest.approx(res)
         # check that result is not changed
         res = acc.compute()
-        assert n * idist.get_world_size() == acc._num_examples
+        assert n == acc._num_examples
         assert isinstance(res, float)
-        assert accuracy_score(np_y, np_y_pred) == pytest.approx(res)
```
vfdev-5, line 318: @sadra-barikbin revert this as well

vfdev-5:
@sadra-barikbin please resolve the conflict for this PR

tests/ignite/metrics/test_accuracy.py
vfdev-5: and 2 above lines as well

sadra-barikbin: I added them below

ignite/metrics/metric.py
```diff
@@ -556,25 +555,37 @@ def another_wrapper(self: Metric, *args: Any, **kwargs: Any) -> Callable:
...
+                    if not isinstance(t, (Number, torch.Tensor)):
+                        raise TypeError(
+                            "Attribute provided to sync_all_reduce should be a "
+                            f"number or tensor but `{attr}` has type {type(t)}"
+                        )
+                    unreduced_attrs[attr] = t
+                    # Here `clone` is necessary since `idist.all_reduce` modifies `t` inplace in the case
+                    # `t` is a tensor and its `device` is same as that of the process.
+                    # TODO: Remove this dual behavior of `all_reduce` to always either return a new tensor or
+                    #       modify it in-place.
```
vfdev-5, line 580: We could hide `.clone` into `idist.all_reduce` but this can have a perf impact

ignite/metrics/metric.py
```diff
@@ -556,25 +555,37 @@ def another_wrapper(self: Metric, *args: Any, **kwargs: Any) -> Callable:
...
+                    t_reduced = idist.all_reduce(cast(float, t) if isinstance(t, Number) else t.clone(), **op_kwargs)
+                    setattr(self, attr, t_reduced)
+
+            result = func(self, *args, **kwargs)
+
+            for attr, value in unreduced_attrs.items():
+                setattr(self, attr, value)
```
vfdev-5, line 587: We can also write `result` into `self._result` for caching

ignite/metrics/metric.py
sadra-barikbin: Yes. It's better not to do that.

ignite/metrics/metric.py
sadra-barikbin: Yes we can. You'd said we'd do it later.

sadra-barikbin: So shall I do it now?

vfdev-5: Yes, let's add it now

vfdev-5: Finally we decided to postpone

vfdev-5 (approved):

PR #2805, opened 2022-12-22
Fix a typo in `distributed.utils::broadcast`

sadra-barikbin:


ignite/distributed/utils.py
```diff
@@ -386,8 +386,8 @@ def broadcast(
         src: source rank. Default, 0.
         safe_mode: if True, non source input data can be ``None`` or anything (will be discarded), otherwise data
             type of the input ``tensor`` should be respected for all processes. Please, keep in mind, this mode is
-            working only for dense tensors as source input if a tensor is provided. There are additional collective
-            ops are performed before doing the broadcast and, thus, can be slower than without using this mode.
+            working only for dense tensors as source input if a tensor is provided. With ``safe_mode=False``, some
```
vfdev-5, line 389: @sadra-barikbin the way how you updated the docstring does not seem good. With safe mode false, there are no additional collective ops performed... can you explain what's wrong with original text ?

sadra-barikbin: The original typo was `There are additional collective ops are performed`, but I wanted to rephrase the sentence as well.

vfdev-5: And what's wrong with this ?

sadra-barikbin: It is written `There are something are performed` that's grammatically wrong.

sadra-barikbin: I can keep the sentence as it was but just omit the second `are`.

ignite/distributed/utils.py
```diff
@@ -386,8 +386,8 @@ def broadcast(
...
+            additional collective ops are performed before doing the broadcast so it can be faster to enable this mode.
```
vfdev-5, line 390: RIght now the sentence is not logically correct as `safe_mode` is `False` by default, so no need to enable it.
The whole point of the sentence about performance impact with `safe_mode=True` is to say that with `safe_mode=True` brings simplicity for the code but has a small runtime impact (a tradeoff). The way how you write the sentence leads to a confusion IMO

sadra-barikbin: How is it to say:
```
if True, non source input data can be ``None`` or anything (will be discarded), otherwise data type of the input 
``tensor`` should be respected for all processes. Please, keep in mind, this mode is working only for dense tensors as 
source input if a tensor is provided. It also leads to some additional collectives before the broadcast, making it slower 
than without using this mode. 
```

vfdev-5: Looks ok to me, thanks !

sadra-barikbin: Or we can keep everything unchanged but remove the second `are` in the original code.

vfdev-5 (approved):
Thanks @sadra-barikbin

PR #2765, opened 2022-11-08
fix(search): bump @docsearch/js to 3.3 & add `appId` config

ydcjeff:
Same as https://github.com/pytorch-ignite/pytorch-ignite.ai/pull/159

PR #2807, opened 2022-12-30
Update @docsearch/css

ydcjeff:
Follow up of #2765 I forgot to update the css

Description:

Check list:


vfdev-5:
Thanks for the PR @ydcjeff, right now for me it looks like 
image
but it was 
image

can we keep either "buttons" and also change the font for "K", in the PR it looks uglier with serif IMO

ydcjeff:
Ohhh, I forgot that my browser has setup sans font as default so it looks good on my end. will push a fix.

vfdev-5 (approved):
LGTM, thanks @ydcjeff !

PR #2836, opened 2023-01-30
Drop python 3.7 from CI

### LINKED ISSUE #2834: [CI] Drop support for python 3.7
https://github.com/pytorch/pytorch/issues/80513

abhi-glitchhg:
Fixes #2834 

I used `grep` command to look for 3.7 in workflow files; and also removed the target version `py37` from `pyproject.toml` file. 

Opening as a draft to check CI is happy with current state.

### DIFF
--- a/.github/workflows/pytorch-version-tests.yml
+++ b/.github/workflows/pytorch-version-tests.yml
@@ -14,7 +14,7 @@ jobs:
       max-parallel: 10
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: [ 3.8, 3.9, "3.10"]
         pytorch-version:
           [1.12.1, 1.11.0, 1.10.0, 1.9.1, 1.8.1, 1.7.1, 1.6.0, 1.5.1, 1.4.0, 1.3.1]
         exclude:
--- a/.github/workflows/pytorch-version-tests.yml
+++ b/.github/workflows/pytorch-version-tests.yml
@@ -14,7 +14,7 @@ jobs:
       max-parallel: 10
       fail-fast: false
       matrix:
-        python-version: [ 3.8, 3.9, "3.10"]
+        python-version: [3.8, 3.9, "3.10"]
         pytorch-version:
           [1.12.1, 1.11.0, 1.10.0, 1.9.1, 1.8.1, 1.7.1, 1.6.0, 1.5.1, 1.4.0, 1.3.1]
         exclude:
--- a/.github/workflows/tpu-tests.yml
+++ b/.github/workflows/tpu-tests.yml
@@ -62,9 +62,10 @@ jobs:
           # mkl version fixed due to https://github.com/pytorch/ignite/issues/2350
           pip install mkl==2021.4.0 requests gsutil
 
-          ## Install torch & xla
-          curl https://raw.githubusercontent.com/abhi-glitchhg/ignite_tpu_script/main/env_setup.py -o pytorch-xla-env-setup.py
-          python pytorch-xla-env-setup.py --version "nightly"
+          ## Install torch & xla and torchvision
+          pip install --pre  https://storage.googleapis.com/tpu-pytorch/wheels/colab/torch-nightly-cp38-cp38-linux_x86_64.whl
+          pip install --pre  https://storage.googleapis.com/tpu-pytorch/wheels/colab/torch_xla-nightly-cp38-cp38-linux_x86_64.whl
+          pip install --pre  https://storage.googleapis.com/tpu-pytorch/wheels/colab/torchvision-nightly-cp38-cp38-linux_x86_64.whl
 
           ## Install test deps and Ignite
           pip install -r requirements-dev.txt
--- a/.github/workflows/tpu-tests.yml
+++ b/.github/workflows/tpu-tests.yml
@@ -28,10 +28,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python 3.7
+      - name: Set up Python 3.8
         uses: actions/setup-python@v4
         with:
-          python-version: "3.7"
+          python-version: "3.8"
           architecture: "x64"
 
       - name: Get year & week number
--- a/.github/workflows/tpu-tests.yml
+++ b/.github/workflows/tpu-tests.yml
@@ -63,7 +63,7 @@ jobs:
           pip install mkl==2021.4.0 requests gsutil
 
           ## Install torch & xla
-          curl https://raw.githubusercontent.com/pytorch/xla/master/contrib/scripts/env-setup.py -o pytorch-xla-env-setup.py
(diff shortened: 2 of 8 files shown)


vfdev-5:
@abhi-glitchhg thanks for the PR, seems like torch XLA is still on python 3.7

abhi-glitchhg:
> @abhi-glitchhg thanks for the PR, seems like torch XLA is still on python 3.7

yeah; what should we do; Should i make an issue for this on XLA repo?

vfdev-5:
It may be possible that we are not correctly installing it for python 3.8. In colab they seem to pick python 3.8 : https://colab.research.google.com/github/pytorch/xla/blob/master/contrib/colab/pytorch-xla-profiling-colab.ipynb

abhi-glitchhg:
Hmm,

I went through the [error](https://github.com/pytorch/ignite/actions/runs/4043190671/jobs/6951819670#step:7:264)


 
```
WARNING: Skipping torch as it is not installed.
WARNING: Skipping torchvision as it is not installed.
CommandException: No URLs matched: gs://tpu-pytorch/wheels/colab/torch-nightly-cp38-cp38m-linux_x86_64.whl
CommandException: No URLs matched: gs://tpu-pytorch/wheels/colab/torch_xla-nightly-cp38-cp38m-linux_x86_64.whl
CommandException: No URLs matched: gs://tpu-pytorch/wheels/colab/torchvision-nightly-cp38-cp38m-linux_x86_64.whl
WARNING: Requirement 'torch-nightly-cp38-cp38m-linux_x86_64.whl' looks like a filename, but the file does not exist
ERROR: torch-nightly-cp38-cp38m-linux_x86_64.whl is not a supported wheel on this platform.
```


We are running [this](https://raw.githubusercontent.com/pytorch/xla/master/contrib/scripts/env-setup.py) set up script to download the wheels. From the error we can see that 3.8 python version is being used by the script but URL is not matching(found?). 

@vfdev-5, could it be possible that nightly versions for 3.8 are not up there? or can we get the list of all the available xla wheels for 3.8 version? 



Thanks

vfdev-5:
@abhi-glitchhg I do not know but I just tried in colab without any accelerators to run the following: 
```
pip install cloud-tpu-client==0.10 --pre torch https://storage.googleapis.com/tpu-pytorch/wheels/colab/torch_xla-nightly-cp38-cp38-linux_x86_64.whl
```
and it seems to work...

You can try either on your fork to enable GH actions and figure out how to install torch xla or either you can do things in docker container...

.github/workflows/tpu-tests.yml
```diff
@@ -63,7 +63,7 @@ jobs:
           pip install mkl==2021.4.0 requests gsutil
 
           ## Install torch & xla
-          curl https://raw.githubusercontent.com/pytorch/xla/master/contrib/scripts/env-setup.py -o pytorch-xla-env-setup.py
+          curl https://raw.githubusercontent.com/abhi-glitchhg/ignite_tpu_script/main/env_setup.py -o pytorch-xla-env-setup.py
```
vfdev-5, line 66: @abhi-glitchhg we can't use your copy of `env_setup.py` as a solution. What's the difference between two files?

vfdev-5: OK, I get the difference, you removed `m` from the wheel name: `cp38-cp38m` -> `cp38-cp38`.
One hacky solution we can do is download original file and update (remove `m`) in locally with `sed`. Another solution could be to change the installation process to the one recommended in colab. I think the second approach would be a bit better

abhi-glitchhg: Hello @vfdev-5 , removing `m` from the links is working, but I wanted to know the significance of m in the earlier link and why it has been dropped for 3.8 wheels! To make sure that we are not doing anything wrong here. 

Regarding the env_setup.py script, yes, we shouldn't rely on personal repos :). 
I was checking if this way is working or not.

The first approach is fine with me, or we could have the env_setup.py script inside the ignite repo; this way, we would have more control over the installation in the future. 

> Another solution could be to change the installation process to the one recommended in colab.

If I understand correctly, we would be directly installing wheels with the `pip install` command in the second approach, right? (correct me if my understanding is wrong) . (IMO, the current env_setup.py script is overkill for nightly installs.)  so this approach is fine with me. 

Let me know. 
Thanks

vfdev-5: Concerning `m` flag I could find the following info: 

> Per [PEP 3149](https://www.python.org/dev/peps/pep-3149/), m indicates that the ABI in question is using [the pymalloc allocator](https://docs.python.org/3/c-api/memory.html#pymalloc):

- https://stackoverflow.com/questions/54097033/what-does-the-m-in-a-python-abi-tag-mean
- https://peps.python.org/pep-3149/

and that in py38 `m` flag is no longer appended (ref: https://github.com/pypa/pip/pull/6874)

> If I understand correctly, we would be directly installing wheels with the pip install command in the second approach, right? (correct me if my understanding is wrong) . (IMO, the current env_setup.py script is overkill for nightly installs.) so this approach is fine with me.

yeah, let's take this approach and simplify the installation process. So, let's check it in the CI and hope that it works.

abhi-glitchhg: Thanks for the info; I should have searched about this on google. sorry for the trouble. 

>let's take this approach and simplify the installation process. So, let's check it in the CI and hope that it works.

 I will try this.

.github/workflows/pytorch-version-tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       max-parallel: 10
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: [ 3.8, 3.9, "3.10"]
```
vfdev-5, line 17:
```suggestion
        python-version: [3.8, 3.9, "3.10"]
```

.github/workflows/unit-tests.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: [ 3.8, 3.9, "3.10"]
```
vfdev-5, line 39:
```suggestion
        python-version: [3.8, 3.9, "3.10"]
```

vfdev-5 (approved):
Thanks for the PR, @abhi-glitchhg 
LGTM

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.black]
 line-length = 120
-target-version = ['py37', 'py38', 'py39']
+target-version = ['py38', 'py39']
```
abhi-glitchhg, line 3: i think we should add 3.10 as target version in here as well.

vfdev-5: I was also thinking about that. But let's keep it like this

PR #2838, opened 2023-01-31
Fix the broken link for a paper `A Systematic Comparison of Smoothing Techniques for Sentence-Level BLEU`

abhi-glitchhg:
i think http://acl2014.org/acl2014 is down. So updating the link with aclanthology link.


vfdev-5 (approved):
Thanks @abhi-glitchhg !

PR #2849, opened 2023-02-05
Removed wrong docstrings for Metric update method's input type

### LINKED ISSUE #2054: Wrong docstrings for Metric `update` method's input type
## 📚 Documentation

For example, [accuracy documentation](https://pytorch.org/ignite/metrics.html#ignite.metrics.Accuracy) says that 

> update must receive output of the form `(y_pred, y)` or `{'y_pred': y_pred, 'y': y}`.

which is incorrect as it can only takes `(y_pred, y)` but `iteration_completed` can handle both types.

sallycaoyu:
Fixes #2054 

Description:
1. Removed input form ``{'y_pred': y_pred, 'y': y, ...}`` from docstrings and from ``docs/source/metrics.rst`` for the ``ignite.Metric.update`` methods as #2054 suggested, since the code does not support this input form.

2. Added descriptions of update method's accepted input form and ``output_transform``'s usage in docstrings for ``ignite.Metric.PSNR`` and ``ignite.Metric.SSIM``.

3. Fixed a bug in ignite.metric.metric.py's CustomMetric example (line 172).

4. Question: There exist two types of function signatures for update methods that take ``(y_pred, y)`` as their inputs:
`def update(self, output: Sequence[torch.Tensor]) -> None:`, and
`def update(self, output: Tuple[torch.Tensor, torch.Tensor]) -> None:` (in EpochMetric)
Should this be made consistent somehow? I think the first option would allow users to pass either a `tuple` or a `list` as the input, whereas the second option will ask the input to stick to the `tuple` type.

Check list:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,6 @@ wheels/
 *.egg
 MANIFEST
 **/.DS_Store
-runs/
 
 # Unit test / coverage reports
 htmlcov/
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ wheels/
 *.egg
 MANIFEST
 **/.DS_Store
+runs/
 
 # Unit test / coverage reports
 htmlcov/
--- a/docs/source/metrics.rst
+++ b/docs/source/metrics.rst
@@ -35,7 +35,7 @@ event. It computes the metric based on its accumulated states. The metric value 
     print(f"Accuracy: {state.metrics['accuracy']}")
 
 
-If the engine's output is not in the format ``(y_pred, y)``, the user can
+If the engine's output is not in the format ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y, ...}``, the user can
 use the ``output_transform`` argument to transform it:
 
 .. code-block:: python
--- a/docs/source/metrics.rst
+++ b/docs/source/metrics.rst
@@ -35,7 +35,7 @@ event. It computes the metric based on its accumulated states. The metric value 
     print(f"Accuracy: {state.metrics['accuracy']}")
 
 
-If the engine's output is not in the format ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y, ...}``, the user can
+If the engine's output is not in the format ``(y_pred, y)``, the user can
 use the ``output_transform`` argument to transform it:
 
 .. code-block:: python
--- a/ignite/metrics/mean_absolute_error.py
+++ b/ignite/metrics/mean_absolute_error.py
@@ -15,14 +15,14 @@ class MeanAbsoluteError(Metric):
 
     where :math:`y_{i}` is the prediction tensor and :math:`x_{i}` is ground true tensor.
 
-    - ``update`` must receive output of the form ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y}``.
+    - ``update`` must receive output of the form ``(y_pred, y)``.
 
     Args:
         output_transform: a callable that is used to transform the
             :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
-            By default, metrics require the output as ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y}``.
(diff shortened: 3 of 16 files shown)


.gitignore
```diff
@@ -23,6 +23,7 @@ wheels/
 *.egg
 MANIFEST
 **/.DS_Store
+runs/
```
vfdev-5, line 26: Please revert this change

docs/source/metrics.rst
```diff
@@ -35,7 +35,7 @@ event. It computes the metric based on its accumulated states. The metric value
     print(f"Accuracy: {state.metrics['accuracy']}")
 
 
-If the engine's output is not in the format ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y, ...}``, the user can
+If the engine's output is not in the format ``(y_pred, y)``, the user can
```
vfdev-5, line 38: Here, it is fine to output dict of y_pred and y. Please revert this as well

ignite/metrics/mean_absolute_error.py
```diff
@@ -15,22 +15,22 @@ class MeanAbsoluteError(Metric):
...
-    - ``update`` must receive output of the form ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y}``.
+    - ``update`` must receive output of the form ``(y_pred, y)``.
 
     Args:
         output_transform: a callable that is used to transform the
             :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
-            By default, metrics require the output as ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y}``.
+            By default, metrics require the output as ``(y_pred, y)``.
```
vfdev-5, line 25: It is only `update` function that can't handle dicts, but engine's output is routed at first into `Metric.iteration_completed` which can handle dicts. So, this update is not correct.

ignite/metrics/metric.py
```diff
@@ -169,7 +169,7 @@ def __init__(self, *args, **kwargs):
                     super().__init__(*args, **kwargs)
 
                 def update(self, output):
-                    y_pred, y, x = output
+                    y_pred, y, x = output["y_pred"], output["y"], output["x"]
```
vfdev-5, line 172: Seems like this example is a bit confusing. It was correct as `update` function receives a tuple.

vfdev-5 (commented):
Thanks for the PR @sallycaoyu 
I left few comments. There is a difference in supported types between `Metric.update` and `Metric.iteration_completed`. As Engine's output goes first into `Metric.iteration_completed` which can handle dicts into tuple and pass this tuple into `Metric.update`. Seems like this is the source of the confusion... I understand that this is not ideal but I'm not what would be the best to fix that.

sallycaoyu:
Thank you @vfdev-5 for the explanation! For now, I have made another commit to revert the changes except for the ones only for ``ignite.Metric.update``. 

For future improvements to this issue, I think one way would be to keep the code as-is and add more details for the docstring

> `- ``update`` must receive output of the form ``(y_pred, y)``.`

...to explain how ``Metric.iteration_completed`` can help handle the dicts for ``ignite.Metric.update``, and that ``ignite.Metric.update`` needs to take ``(y_pred, y)`` eventually. 

Another way would be to change the code implementation, which I am still considering how to do. One way I could think of is to handle the dicts in the base Metric class instead of ``Metric.iteration_completed``, so that this statement will still be true:

> `- ``update`` must receive output of the form ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y}``.`

and other child classes can either run `super(Metric, self).update()` to handle the dicts as the base method defines, or ignore and override it if users want to build a custom method. But I don't think this implementation will look as elegant as the current one...

docs/source/metrics.rst
sallycaoyu: I think the code example right below it suggest that even if the engine's output is already a dict``{'y_pred': y_pred, 'y_true': y, ...}`` returned by process_function, it is still needed to pass output_transform to Accuracy and convert this dict into a tuple``(y_pred, y)``. Please let me know and my apologies if I were wrong.

vfdev-5 (approved):
LGTM, thanks @sallycaoyu !

PR #2851, opened 2023-02-06
Added debug mode for Engine

### LINKED ISSUE #1992: Debug mode for ignite
## 🚀 Feature

The idea is to provide a debug mode as discussed here: https://github.com/pytorch/ignite/issues/1989#issuecomment-833643158.

All the possible options can be discussed here. We could add an helper to track the workflow of handlers from the engine run. On the other hand, debug tools are needed to track the values of PyTorch objects (LR, loss, grads, etc.). Note that the preferred way to extend ignite is handlers' design so let's create a new set of tools for the debugging.

puhuk:
Fixes #1992

Description: Update engine.py to resolve #1992

Check list:


ignite/engine/engine.py
```diff
@@ -125,6 +125,10 @@ def compute_mean_std(engine, batch):
     _state_dict_all_req_keys = ("epoch_length", "max_epochs")
     _state_dict_one_of_opt_keys = ("iteration", "epoch")
 
+    DEBUG_EVENTS = 1
```
vfdev-5, line 128: @puhuk one of the requirements is to be able to combine debug flags: https://github.com/pytorch/ignite/issues/1992#issuecomment-1075132649

Please also add tests for this feature

tests/ignite/engine/test_engine.py
```diff
@@ -1388,3 +1388,72 @@ def check_iter_epoch():
...
+        def log_training_debug_events(engine):
+            trainer.debug(level=DEBUG_EVENTS)
+
+        @trainer.on(Events.ITERATION_COMPLETED(every=log_interval))
+        def log_training_debug_outputs(engine):
+            trainer.debug(level=DEBUG_OUTPUT, optimizer=optimizer)
+
+        @trainer.on(Events.ITERATION_COMPLETED(every=log_interval))
+        def log_training_debug_grads(engine):
+            trainer.debug(level=DEBUG_GRADS, optimizer=optimizer, layer=model.fc2)
```
vfdev-5, line 1455: Usage is a bit different:
```python
trainer = Engine(...)

trainer.debug(level=Engine.DEBUG_EVENTS | Engine.DEBUG_OUTPUT | Engine.DEBUG_GRADS)

trainer.run(data, max_epochs=max_epochs)
```

vfdev-5 (changes requested):

PR #2854, opened 2023-02-10
Fix mypy Type Error

theory-in-progress:
Fixes mypy issue in workflow

Description:
* Fixed the mypy type error because of tensorboard_logger.py for the nightly version
Check list:


vfdev-5 (approved):
LGTM, thanks @theory-in-progress !

PR #2880, opened 2023-02-22
Auto dataloader enable persistent workers

vfdev-5:
Description:
- Auto dataloader enable persistent workers

Check list:


vfdev-5:
Test timings for cifar10 with different configs looks a bit unfortunate:

This PR
image

Master
image


To investigate more on this

PR #2882, opened 2023-02-22
Issue#2878: Adds Siamese Network example

DeepC004:
Fixes #{issue number}

Description:
I have added an example on Siamese Networks in the examples folder along with requirements.txt and README.md.
I have tested the network locally with different parameters and set the default parameters to ones which perform the best.
Note: I am also working on Google Colab Notebook for the same and will add it to [pytorch/examples](https://github.com/pytorch/examples) soon.

Check list:

### DIFF
--- a/examples/siamese_network/README.md
+++ b/examples/siamese_network/README.md
@@ -1,0 +1,10 @@
+# Siamese Network example on MNIST dataset
+
+This example is ported over from [pytorch/examples](https://github.com/pytorch/examples)
+
+Usage:
+
+```
+pip install -r requirements.txt
+python siamese.py
+```
--- a/examples/siamese_network/siamese_network.py
+++ b/examples/siamese_network/siamese_network.py
@@ -1,0 +1,312 @@
+from __future__ import print_function
+import argparse
+import numpy as np
+import random
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+import torchvision 
+from ignite.engine import Engine, Events
+from ignite.contrib.handlers import ProgressBar
+from ignite.handlers.param_scheduler import LRScheduler
+from torch.optim.lr_scheduler import StepLR
+from torch.utils.data import DataLoader
+from torchvision import datasets
+
+class SiameseNetwork(nn.Module):
+    # update Siamese Network implementation in accordance with the dataset
+    """
+        Siamese network for image similarity estimation.
+        The network is composed of two identical networks, one for each input.
+        The output of each network is concatenated and passed to a linear layer. 
+        The output of the linear layer passed through a sigmoid function.
+        `"FaceNet" <https://arxiv.org/pdf/1503.03832.pdf>`_ is a variant of the Siamese network.
+        This implementation varies from FaceNet as we use the `ResNet-18` model from
+        `"Deep Residual Learning for Image Recognition" <https://arxiv.org/pdf/1512.03385.pdf>` 
+        as our feature extractor.
+        In addition, we aren't using `TripletLoss` as the MNIST dataset is simple, so `BCELoss` can do the trick.
+    """
+    def __init__(self):
+        super(SiameseNetwork, self).__init__()
+        # get resnet model
+        self.resnet = torchvision.models.resnet18(weights=None)
+
+        # over-write the first conv layer to be able to read MNIST images
+        # as resnet18 reads (3,x,x) where 3 is RGB channels
+        # whereas MNIST has (1,x,x) where 1 is a gray-scale channel
+        self.resnet.conv1 = nn.Conv2d(1, 64, kernel_size=(7, 7), stride=(2, 2), padding=(3, 3), bias=False)
+        self.fc_in_features = self.resnet.fc.in_features
+
+        # remove the last layer of resnet18 (linear layer which is before avgpool layer)
+        self.resnet = torch.nn.Sequential(*(list(self.resnet.children())[:-1]))
+
+        # add linear layers to compare between the features of the two images
+        self.fc = nn.Sequential(
+            nn.Linear(self.fc_in_features * 2, 256),
+            nn.ReLU(inplace=True),
+            nn.Linear(256, 1),
+        )
+
+        self.sigmoid = nn.Sigmoid()
+
+        # initialize the weights
+        self.resnet.apply(self.init_weights)
+        self.fc.apply(self.init_weights)
+
+    def init_weights(self, m):
+        if isinstance(m, nn.Linear):
+            torch.nn.init.xavier_uniform_(m.weight)
+            m.bias.data.fill_(0.01)
+
+    def forward_once(self, x):
+        output = self.resnet(x)
+        output = output.view(output.size()[0], -1)
+        return output
+
+    def forward(self, input1, input2):
+        # get two images' features
+        output1 = self.forward_once(input1)
+        output2 = self.forward_once(input2)
+
+        # concatenate both images' feature
+        output = torch.cat((output1, output2), 1)
+
+        # pass the concatenation to the linear layers
+        output = self.fc(output)
+
+        # pass the out of the linear layers to sigmoid layer
+        output = self.sigmoid(output)
+
+        return output
+
+class APP_MATCHER():
+    # following class implements data downloading and handles preprocessing
+    def __init__(self, root, train, download=False):
+        super(APP_MATCHER, self).__init__()
+
+        # get MNIST dataset
+        self.dataset = datasets.MNIST(root, train=train, download=download)
+
+        # as `self.dataset.data`'s shape is (Nx28x28), where N is the number of
+        # examples in MNIST dataset, a single example has the dimensions of
+        # (28x28) for (WxH), where W and H are the width and the height of the image. 
+        # However, every example should have (CxWxH) dimensions where C is the number 
+        # of channels to be passed to the network. As MNIST contains gray-scale images, 
+        # we add an additional dimension to corresponds to the number of channels.
+        self.data = self.dataset.data.unsqueeze(1).clone()
+
+        self.group_examples()
+
+    def group_examples(self):
+        """
+            To ease the accessibility of data based on the class, we will use `group_examples` to group
+            examples based on class. 
+            
+            Every key in `grouped_examples` corresponds to a class in MNIST dataset. For every key in 
+            `grouped_examples`, every value will conform to all of the indices for the MNIST 
+            dataset examples that correspond to that key.
+        """
+
+        # get the targets from MNIST dataset
+        np_arr = np.array(self.dataset.targets.clone())
+
+        # group examples based on class
+        self.grouped_examples = {}
+        for i in range(0,10):
+            self.grouped_examples[i]=np.where((np_arr==i))[0]
+
+    def __len__(self):
+        return self.data.shape[0]
+
+    def __getitem__(self, index):
+        """
+            For every example, we will select two images. There are two cases, 
+            positive and negative examples. For positive examples, we will have two 
+            images from the same class. For negative examples, we will have two images 
+            from different classes.
+
+            Given an index, if the index is even, we will pick the second image from the same class, 
+            but it won't be the same image we chose for the first class. This is used to ensure the positive
+            example isn't trivial as the network would easily distinguish the similarity between same images. However,
+            if the network were given two different images from the same class, the network will need to learn 
+            the similarity between two different images representing the same class. If the index is odd, we will 
+            pick the second image from a different class than the first image.
+        """
+
+        # pick some random class for the first image
+        selected_class = random.randint(0, 9)
+
+        # pick a random index for the first image in the grouped indices based of the label
+        # of the class
+        random_index_1 = random.randint(0, self.grouped_examples[selected_class].shape[0]-1)
+
+        # pick the index to get the first image
+        index_1 = self.grouped_examples[selected_class][random_index_1]
+
+        # get the first image
+        image_1 = self.data[index_1].clone().float()
+
+        # same class
+        if index % 2 == 0:
+            # pick a random index for the second image
+            random_index_2 = random.randint(0, self.grouped_examples[selected_class].shape[0]-1)
+
+            # ensure that the index of the second image isn't the same as the first image
+            while random_index_2 == random_index_1:
+                random_index_2 = random.randint(0, self.grouped_examples[selected_class].shape[0]-1)
+
+            # pick the index to get the second image
+            index_2 = self.grouped_examples[selected_class][random_index_2]
+
+            # get the second image
+            image_2 = self.data[index_2].clone().float()
+
+            # set the label for this example to be positive (1)
+            target = torch.tensor(1, dtype=torch.float)
+
+        # different class
+        else:
+            # pick a random class
+            other_selected_class = random.randint(0, 9)
+
+            # ensure that the class of the second image isn't the same as the first image
+            while other_selected_class == selected_class:
+                other_selected_class = random.randint(0, 9)
+
+            # pick a random index for the second image in the grouped indices based of the label
+            # of the class
+            random_index_2 = random.randint(0, self.grouped_examples[other_selected_class].shape[0]-1)
+
+            # pick the index to get the second image
+            index_2 = self.grouped_examples[other_selected_class][random_index_2]
+
+            # get the second image
+            image_2 = self.data[index_2].clone().float()
+
+            # set the label for this example to be negative (0)
+            target = torch.tensor(0, dtype=torch.float)
+
+        return image_1, image_2, target
+
+def train(model, device, optimizer, train_loader, lr_scheduler, log_interval, max_epochs):
+
+    # we aren't using `TripletLoss` as the MNIST dataset is simple, so `BCELoss` can do the trick.
+    criterion = nn.BCELoss()
+
+    # define model training step
+    def train_step(engine, batch):
+        model.train()
+        image_1, image_2, target = batch
+        image_1, image_2, target = image_1.to(device), image_2.to(device), target.to(device)
+        optimizer.zero_grad()
+        outputs = model(image_1, image_2,).squeeze()
+        loss = criterion(outputs, target)
+        loss.backward()
+        optimizer.step()
+        return loss
+
+    # create a trainer engine and attach train_step
+    trainer = Engine(train_step)
+
+    # attach progress bar to trainer
+    pbar = ProgressBar()
+    pbar.attach(trainer)
+
+    # attach various handlers to trainer engine
+    @trainer.on(Events.ITERATION_COMPLETED(every=log_interval))
+    def log_training_results(engine):
+        print(f'Train Epoch: {engine.state.epoch}, Train Loss: {engine.state.output: .5f}')
+
+    trainer.add_event_handler(Events.ITERATION_STARTED, lr_scheduler)
+
+    # run trainer engine
+    trainer.run(train_loader, max_epochs=max_epochs)
+
+def test(model, device, test_loader, lr_scheduler, log_interval):
+
+    # we aren't using `TripletLoss` as the MNIST dataset is simple, so `BCELoss` can do the trick.
+    criterion = nn.BCELoss()
+    average_test_loss = 0
+    # define model testing step
+    def test_step(engine, batch):
+        model.eval()
+        image_1, image_2, target = batch
+        image_1, image_2, target = image_1.to(device), image_2.to(device), target.to(device)
+        outputs = model(image_1, image_2).squeeze()
+        test_loss = criterion(outputs, target)
+        return test_loss
+
+    # create evaluator engine and attach test step
+    evaluator = Engine(test_step)
+
+    # attach progress bar to evaluator
+    pbar = ProgressBar()
+    pbar.attach(evaluator)
+
+    # attach various handlers to evaluator engine
+    @evaluator.on(Events.ITERATION_COMPLETED(every=log_interval))
+    def log_testing_results(engine):
+        nonlocal average_test_loss
+        average_test_loss += engine.state.output
+        print(f'Test Epoch: {engine.state.epoch} Test Loss: {engine.state.output: .5f}')
+
+    evaluator.add_event_handler(Events.ITERATION_STARTED, lr_scheduler)
+
+    # run evaluator engine 
+    evaluator.run(test_loader)
+
+    # print average loss over test dataset
+    print(f'Average Test Loss: {average_test_loss/len(test_loader.dataset): .7f}')
+
+def main():
+    # adds training defaults and support for terminal arguments
+    parser = argparse.ArgumentParser(description='PyTorch Siamese network Example')
+    parser.add_argument('--batch-size', type=int, default=64, metavar='N',
+                        help='input batch size for training (default: 64)')
+    parser.add_argument('--test-batch-size', type=int, default=200, metavar='N',
+                        help='input batch size for testing (default: 1000)')
+    parser.add_argument('--epochs', type=int, default=10, metavar='N',
+                        help='number of epochs to train (default: 14)')
+    parser.add_argument('--lr', type=float, default=1.0, metavar='LR',
+                        help='learning rate (default: 1.0)')
+    parser.add_argument('--gamma', type=float, default=0.95, metavar='M',
+                        help='Learning rate step gamma (default: 0.7)')
+    parser.add_argument('--no-cuda', action='store_true', default=False,
+                        help='disables CUDA training')
+    parser.add_argument('--no-mps', action='store_true', default=False,
+                        help='disables macOS GPU training')
+    parser.add_argument('--dry-run', action='store_true', default=False,
+                        help='quickly check a single pass')
+    parser.add_argument('--seed', type=int, default=1, metavar='S',
+                        help='random seed (default: 1)')
+    parser.add_argument('--log-interval', type=int, default=10, metavar='N',
+                        help='how many batches to wait before logging training status')
+    parser.add_argument('--save-model', action='store_true', default=False,
+                        help='For Saving the current Model')
+    args = parser.parse_args()
+
+    # set device
+    device = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+
+    # data loading
+    train_dataset = APP_MATCHER('../data', train=True, download=True)
+    test_dataset = APP_MATCHER('../data', train=False)
+    train_loader = DataLoader(train_dataset, shuffle=True, batch_size=args.batch_size)
+    test_loader = DataLoader(test_dataset, batch_size=args.test_batch_size)
+
+    # set model parameters
+    model = SiameseNetwork().to(device)
+    optimizer = optim.Adadelta(model.parameters(), lr=args.lr)
+    scheduler = StepLR(optimizer, step_size=15, gamma=args.gamma)
+    lr_scheduler = LRScheduler(scheduler)
+    
+    # call train function 
+    train(model, device, optimizer, train_loader, lr_scheduler, 
+    log_interval=args.log_interval, max_epochs=args.epochs)
+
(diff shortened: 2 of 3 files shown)


DeepC004:
@vfdev-5 please check the PR. I am unable to understand the origin of error. Is the code formatter adding the following line, ✨ 1 file would be formatted, 267 files already formatted ✨, by itself since it does not show up when formatter is run locally

vfdev-5:
@DeepC004 sure, I'll check the PR as soon as possible. To reformat please do
```
```

DeepC004:
Thank you. I was running the code formatter in the tests folder earlier, hence the code formatter was not picking up all the files for formatting.

vfdev-5:
> Note: I am also working on Google Colab Notebook for the same and will add it to [pytorch/examples](https://github.com/pytorch/examples) soon.

@DeepC004 it should be added to https://github.com/pytorch-ignite/examples not pytorch/examples

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,325 @@
+from __future__ import print_function
```
vfdev-5, line 1: Do we really need this ?

examples/siamese_network/README.md
```diff
@@ -0,0 +1,10 @@
+# Siamese Network example on MNIST dataset
+
+This example is ported over from [pytorch/examples](https://github.com/pytorch/examples)
```
vfdev-5, line 3:
```suggestion
This example is ported over from [pytorch/examples/siamese_network](https://github.com/pytorch/examples/tree/main/siamese_network)
```

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,325 @@
...
+        # pass the concatenation to the linear layers
+        output = self.fc(output)
+
+        # pass the out of the linear layers to sigmoid layer
+        output = self.sigmoid(output)
+
+        return output
+
+
+class APP_MATCHER:
```
vfdev-5, line 88: Let's rename this class to 

```suggestion
class MatcherDataset(Dataset):
```
You also forgot to inherit from `Dataset`

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,325 @@
...
+    # following class implements data downloading and handles preprocessing
+    def __init__(self, root, train, download=False):
+        super(APP_MATCHER, self).__init__()
+
+        # get MNIST dataset
+        self.dataset = datasets.MNIST(root, train=train, download=download)
```
vfdev-5, line 94: Let's make it more interesting and use cifar10 instead of MNIST.

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,325 @@
...
+            # set the label for this example to be negative (0)
+            target = torch.tensor(0, dtype=torch.float)
+
+        return image_1, image_2, target
+
+
+def train(model, device, optimizer, train_loader, lr_scheduler, log_interval, max_epochs):
+
+    # we aren't using `TripletLoss` as the MNIST dataset is simple, so `BCELoss` can do the trick.
+    criterion = nn.BCELoss()
```
vfdev-5, line 201: Let's use https://pytorch.org/docs/stable/generated/torch.nn.TripletMarginLoss.html#torch.nn.TripletMarginLoss instead of BCELoss

vfdev-5 (commented):
Thanks for the PR @DeepC004 !
I left few comments on improving the code and also suggested few changes. Let me know what you think.

examples/siamese_network/siamese_network.py
DeepC004: Now that I think of it, since python version used is > than 3.0 hence it will not be needed

examples/siamese_network/siamese_network.py
DeepC004: Sure!

DeepC004:
@vfdev-5 I have made the changes. May you please review them

vfdev-5:
@DeepC004 Thanks ! Can you share here training results (logs, metrics), does model converge etc ?

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,284 @@
...
+    trainer = Engine(train_step)
+
+    # attach progress bar to trainer
+    pbar = ProgressBar()
+    pbar.attach(trainer)
+
+    # attach various handlers to trainer engine
+    @trainer.on(Events.ITERATION_COMPLETED(every=log_interval))
+    def log_training_results(engine):
+        print(f"Train Epoch: {engine.state.epoch}, Train Loss: {engine.state.output: .5f}")
```
vfdev-5, line 184: If you are using progress bar, I think there is no need to print loss. You can do the following:
```python
    pbar = ProgressBar()
    pbar.attach(trainer, metric_names=['loss'])
```

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,284 @@
...
+    parser.add_argument("--save-model", action="store_true", default=False, help="For Saving the current Model")
+    args = parser.parse_args()
+
+    # set device
+    device = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+
+    # data loading
+    train_dataset = MatcherDataset("../data", train=True, download=True)
+    test_dataset = MatcherDataset("../data", train=False)
+    train_loader = DataLoader(train_dataset, shuffle=True, batch_size=args.batch_size)
```
vfdev-5, line 267: Please add `num_workers` as argparse argument and set default value to 4 (num_workers=4)

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,284 @@
...
+    model = SiameseNetwork().to(device)
+    optimizer = optim.Adadelta(model.parameters(), lr=args.lr)
+    scheduler = StepLR(optimizer, step_size=15, gamma=args.gamma)
+    lr_scheduler = LRScheduler(scheduler)
+
+    # call train function
+    train(model, device, optimizer, train_loader, lr_scheduler, log_interval=args.log_interval, max_epochs=args.epochs)
+
+    # call test function
+    test(model, device, test_loader, lr_scheduler, log_interval=args.log_interval)
```
vfdev-5, line 280: Instead of separating train and test like that we can mix them: train and run evaluation every N epochs

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,284 @@
...
+    )
+    parser.add_argument("--epochs", type=int, default=10, metavar="N", help="number of epochs to train (default: 14)")
+    parser.add_argument("--lr", type=float, default=1.0, metavar="LR", help="learning rate (default: 1.0)")
+    parser.add_argument(
+        "--gamma", type=float, default=0.95, metavar="M", help="Learning rate step gamma (default: 0.7)"
+    )
+    parser.add_argument("--no-cuda", action="store_true", default=False, help="disables CUDA training")
+    parser.add_argument("--no-mps", action="store_true", default=False, help="disables macOS GPU training")
+    parser.add_argument("--dry-run", action="store_true", default=False, help="quickly check a single pass")
+    parser.add_argument("--seed", type=int, default=1, metavar="S", help="random seed (default: 1)")
```
vfdev-5, line 250: Seed is not used anywhere. Please use `manual_seed` from `ignite.utils`

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,284 @@
...
+    pbar.attach(evaluator)
+
+    # attach various handlers to evaluator engine
+    @evaluator.on(Events.ITERATION_COMPLETED(every=log_interval))
+    def log_testing_results(engine):
+        nonlocal average_test_loss
+        average_test_loss += engine.state.output
+        print(f"Test Epoch: {engine.state.epoch} Test Loss: {engine.state.output: .5f}")
+
+    evaluator.add_event_handler(Events.ITERATION_STARTED, lr_scheduler)
```
vfdev-5, line 224: Why do you update lr_scheduler during the testing ?

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,284 @@
...
+    def train_step(engine, batch):
+        model.train()
+        anchor_image, positive_image, negative_image, anchor_label = batch
+        anchor_image = anchor_image.to(device)
+        positive_image, negative_image = positive_image.to(device), negative_image.to(device)
+        anchor_label = anchor_label.to(device)
+        optimizer.zero_grad()
+        anchor_out = model(anchor_image)
+        positive_out = model(positive_image)
+        negative_out = model(negative_image)
```
vfdev-5, line 168: Why we are doing 3 forward passes here ? Can't we group them into a single one ?

examples/siamese_network/siamese_network.py
DeepC004: Oh alright, I will add it

examples/siamese_network/siamese_network.py
DeepC004: Yes that is a good idea since we will get to see the progress of the model on unseen data with time. I will make the change

examples/siamese_network/siamese_network.py
DeepC004: I overlooked that detail, will remove it now.

DeepC004:
Yes, even I ran it locally before submitting and it took me a couple hours to complete the training (with cuda enabled). The `loss` drops down in ranges of `0.02` after the first epoch itself, and the average test loss came out to be around `0.00006` after 2 epochs of training. I will add time profiling to check for bottlenecks.

vfdev-5:
Sounds good.

> I will add time profiling to check for bottlenecks.

The point is not to add profiling into the code but just understand why GPU is almost idle during the run

DeepC004:
I trained the model for `2` epochs and these are the results. Around `98.5%` of time is spent in data flow tasks. I tried optimizations like using `num_workers` and setting `pin_memory` to `true` when creating `DataLoader` object, but there isn't much improvement. Is there any workaround for this problem?
image

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,284 @@
...
+        self.dataset = datasets.CIFAR10(root, train=train, download=download)
+
+        # convert data from numpy array to Tensor
+        self.data = torch.from_numpy(self.dataset.data)
+
+        # shift the dimensions of dataset to match the initial input layer dimensions
+        self.data = torch.movedim(self.data, (0, 1, 2, 3), (0, 2, 3, 1))
+
+        # convert targets list to torch Tensor
+        self.dataset.targets = torch.Tensor(self.dataset.targets)
```
vfdev-5, line 79: Minor fix:
```suggestion
        self.dataset.targets = torch.tensor(self.dataset.targets)
```

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,284 @@
...
+        anchor image.
+
+        Now we select two random images, one belonging to the same class as that of the
+        anchor image (named as positive_image) and the other belonging to a different class
+        than that of the anchor image (named as negative_image). We return the anchor image,
+        positive image, negative image and anchor label.
+        """
+
+        # obtain the anchor image
+        anchor_image = self.data[index].clone().float()
```
vfdev-5, line 118: Do we need to copy data here ?

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,284 @@
...
+
+        # obtain the class label of the anchor image
+        anchor_label = self.dataset.targets[index].clone().float()
```
vfdev-5, line 121: Same question here about `clone`

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,284 @@
...
+
+        # pick some random class for the first image
+        selected_class = random.randint(0, 9)
+
+        # pick a random index for the first image in the grouped indices based of the label
+        # of the class
+        random_index_1 = random.randint(0, self.grouped_examples[selected_class].shape[0] - 1)
+
+        while random_index_1 != anchor_label:
+            random_index_1 = random.randint(0, self.grouped_examples[selected_class].shape[0] - 1)
```
vfdev-5, line 131: Maybe, we can use `torch.rand*` here and rewrite/optimize this code

examples/siamese_network/siamese_network.py
DeepC004: I believe it might be the reason for taking access time. I will remove it and check

examples/siamese_network/siamese_network.py
DeepC004: I believe `torch.rand` used for generating a random tensor of given size. May you please guide me on how it can be used here? In the above code I am generating a random number within the range of number of samples in each class following which I am sampling a random image using the generated number.

DeepC004: I found the error in code. Fixing it asap!

DeepC004:
@vfdev-5 I have made the changes. The run time is much better now!

examples/siamese_network/siamese_network.py
vfdev-5: @DeepC004 this comment is not addressed

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,270 @@
...
+        output1 = self.forward_once(input1)
+        output2 = self.forward_once(input2)
+        output3 = self.forward_once(input3)
+
+        # pass the output of resnet to sigmoid layer
+        output1 = self.sigmoid(output1)
+        output2 = self.sigmoid(output2)
+        output3 = self.sigmoid(output3)
+
+        return output1, output2, output3
```
vfdev-5, line 67: I'm not 100% sure about this. Please check FaceNet paper to see what exactly model should do.

examples/siamese_network/siamese_network.py
vfdev-5: Same here, unaddressed comment

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,270 @@
...
+
+        # obtain the anchor image
+        anchor_image = self.data[index].clone().float()
+
+        # obtain the class label of the anchor image
+        anchor_label = self.dataset.targets[index]
+        anchor_label = int(anchor_label.item())
+
+        # find a label which is different from anchor_label
+        neg_label = random.randint(0, 9)
```
vfdev-5, line 131: Let's replace `random.randint` with `torch.randint(0, 10, size=())` and below

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,270 @@
...
+    # define model testing step
+    def test_step(engine, batch):
+        model.eval()
+        anchor_image, positive_image, negative_image, anchor_label = batch
+        anchor_image = anchor_image.to(device)
+        positive_image, negative_image = positive_image.to(device), negative_image.to(device)
+        anchor_label = anchor_label.to(device)
+        anchor_out, positive_out, negative_out = model(anchor_image, positive_image, negative_image)
+        test_loss = criterion(anchor_out, positive_out, negative_out)
+        return test_loss.item()
```
vfdev-5, line 179: Testing step and metric should be something like accuracy: given two images of the same class -> model accuracy that model correctly predicts that they belong to the same class. Same for samples from different classes.

vfdev-5 (commented):
@DeepC004 I left few more comments.
By the way have you committed the changes, because I see a lot of unaddressed comments

DeepC004:
> @DeepC004 I left few more comments. By the way have you committed the changes, because I see a lot of unaddressed comments

Oh, I believe I commited the wrong file with incomplete changes. I will update it.

examples/siamese_network/siamese_network.py
DeepC004: I will look into the original model architecture.

examples/siamese_network/siamese_network.py
DeepC004: Yes, that will be a better measure of testing the model. In the original paper they have described as follows, "given a pair of two face images a squared L2 distance threshold D(xi, xj) is used to determine the classification of same and different." I will work on the changes!

DeepC004: @vfdev-5 I read the FaceNet paper and tried tweaking the model structure and threshold values for classifying between same and different images. However, the accuracy which I am getting is not at par with the best models on CIFAR10 datasets. 
The problem with the Triplet Loss approach is that in the same category, let's say cars, there are many different types and colours of cars, hence the model is not able to `learn` the features properly. Even so, the loss function is trying to minimize the squared Euclidean distance between two images of the same category and maximize it between two images belonging to different category but there are some common features in between the images of different categories, for example `cars` and `trucks` categories are having similar features like tyres, which might the cause of poor performance. 
I believe implementing the Siamese Network and Triplet Margin Loss on the dataset used in the original FaceNet paper will be a better option.

DeepC004: In the original implementation in [pytorch/examples/siamese_network](https://github.com/pytorch/examples/blob/main/siamese_network/main.py), the implementation seems wrong to me, since in the model is using a sigmoid output after passing through the network and finally the predictions are made using the threshold value of 0.5 (i.e. if the output value is greater than 0.5 then the prediction is correct, incorrect otherwise). The problem in this approach is that the output of the model is there is no classification as such and we are just checking the sigmoid passed value of the final output of the model which is coming out to be near 1 for most of the pixels and hence the increased accuracy!

vfdev-5: @DeepC004 Thanks for the updates !

Concerning learning on CIFAR10 dataset, we try few other options: 1) try triplet loss with another distance, https://pytorch.org/docs/stable/generated/torch.nn.TripletMarginWithDistanceLoss.html, 2) improve feature extraction part in the model. 
I agree that loss function choice is very important. However, I think it's still possible to learn on cifar10, for example ultimately a pretrained for 10 labels classification resnet18 will be able to provide "features" (one-hot encoded vector for predicted class) such that computed similarity distance would make sense.

> I believe implementing the Siamese Network and Triplet Margin Loss on the dataset used in the original FaceNet paper will be a better option.

I'm not sure about something like FaceNet dataset if it is available on torchvision easily.

> Only the sigmoid passed value of the final output of the model are being checked. These values come out to be almost 1 for most of the pixels and hence the increased accuracy!

I'm not sure to fully understand your point here. Can you give an example and also say how to fix the issue in the example.

DeepC004: Thanks for the prompt response! In regards to the points raised:
1) I will try running the model with the mentioned loss function
2) I will tweak the model layers/use the more recent models which have shown better results
Regarding the loss function part, the model is not directly giving out the one hot encoded vector for the 10 classes, rather it is using the label to select an image and then using that anchor label to find positive and negative images accordingly. The model I recently tried running is similar to the one described in the original FaceNet paper (I updated it to include the FC layers in the end and changed the activation functions to ReLU). 
The LFW dataset (used in the FaceNet paper) is open source and is easily available.

My plan of action is as follows:
1) I will tweak the loss function, and the model architecture, both standalone and simultaneously, and check if I can get any better results.
2) Since LFW dataset is available, I will implement the same model on the dataset and check how it performs. Incase option 1 does not work then we can go with this example.

This piece of [code](https://github.com/pytorch/examples/blob/7ec911c46c00ac98d3adfd996ee2a011bbb9fdba/siamese_network/main.py#L70-L78) is just concatenating the anchor image and a randomly chosen positive or negative image followed by passing through a FC layer which is followed by a sigmoid layer. I am not completely sure if I understand the reason for doing this in the mentioned example.

vfdev-5: @DeepC004 thinking more about all these things: siamese networks and FaceNet. Seems like we are more trying to correctly implement FaceNet paper. In this case, it would be better to check most prominent repos from "paper with code": https://paperswithcode.com/paper/facenet-a-unified-embedding-for-face

In such case, this PR with an example is not just about siamese networks but FaceNet as image recognition and clustering.

I think it would be better to rescope the purpose of the example and also implement something more recent (paper by 2021, 2022, 2023).

I agree with you that pytorch/example code can be doing something different from they claim to do. I think it is correctly labelled as siamese networks example than facenet.

If we want to stick with siamese network idea, we can : 

1) try to revert this code to pytorch/example one: use BCELoss, model does concat of some resnet features as in the example and outputs similarity of 2 images. But let's still use that on CIFAR10. We should however complement the code with more relevant metrics as discussed in this PR. Maybe, also add some visualizations. 

2) let's pick something exciting from : https://paperswithcode.com/method/siamese-network

DeepC004: It will be great to complete this example and then move on to a more recent implementation as a separate example. I will make the mentioned changes, test the model and update you on that. Post this I can look into other interesting papers!

vfdev-5: Sounds good to me, thanks !

DeepC004: Hi, sorry for the late response since I had my mid terms the past week. I tried making a few different changes to the model architecture and tested them locally. Results are as follows:
1) I updated the original code taken from [pytorch/examples/siamese_network](https://github.com/pytorch/examples/blob/main/siamese_network/main.py) to make it work on CIFAR10 dataset. The model gave an accuracy of around 71%. I checked other one vs all classifier models (multi class classification using BCE loss) on CIFAR10 dataset and they have accuracy results in similar ranges.
2) Another network architecture which makes use of Triplet Margin Loss and random triplet selection gave an accuracy of around 75%. A good classification depends on appropriate choice of threshold for the pairwise distances. These results are using ResNet34 as base network.

DeepC004:
@vfdev-5 May you please review the changes.

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,322 @@
...
+        anchor_image = self.data[index].float()
+
+        # obtain the class label of the anchor image
+        anchor_label = self.dataset.targets[index]
+        anchor_label = int(anchor_label.item())
+
+        # find a label which is different from anchor_label
+        neg_label = torch.randint(0, 10, (1,)).item()
+        while neg_label == anchor_label:
+            neg_label = random.randint(0, 9)
```
vfdev-5, line 146: First, you are still using python `random.randint` instead of `torch.randint`.
Second, instead of this `while` loop, you can do better. For example
```python
t = torch.randint(0, 10, size=(200, ))
t = t[t != i]
t[0].item()
```
It is just an example, please find something appropriate.

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,322 @@
...
+        return correct
+
+    # create engines for trainer and evaluator
+    trainer = Engine(train_step)
+    evaluator = Engine(test_step)
+    evaluator.run(test_loader)
+
+    # attach Running Average Loss metric to trainer and evaluator engines
+    RunningAverage(output_transform=lambda x: x).attach(trainer, "loss")
+    RunningAverage(output_transform=lambda x: x).attach(evaluator, "loss")
```
vfdev-5, line 238: Does it make sense to compute running average loss for evaluator ?

vfdev-5 (commented):
@DeepC004 thanks for pinging. I left other comments to address

examples/siamese_network/siamese_network.py
DeepC004: I changed it to the following code. Is this a good way to implement?
```
labels = list(range(0, 10))
labels.remove(anchor_label)
neg_index = torch.randint(0, 9, (1,)).item()
neg_label = labels[neg_index]  
```

vfdev-5: Yes, looks good

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,322 @@
...
+        # find a label which is different from anchor_label
+        neg_label = torch.randint(0, 10, (1,)).item()
+        while neg_label == anchor_label:
+            neg_label = random.randint(0, 9)
+
+        # find the index of image with positive label
+        positive_index = random.choice(self.grouped_examples[anchor_label])
+
+        # find the index of image with negative label
+        negative_index = random.choice(self.grouped_examples[neg_label])
```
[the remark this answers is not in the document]
DeepC004: There is no alternative to random.choice as of now, according to the discussion [here](https://discuss.pytorch.org/t/torch-equivalent-of-numpy-random-choice/16146). An alternate way to implement it can be to use `torch.randint` in the range of incides of grouped_examples to generate a random index and choose a value using this index. Implementation:

```
random_index = torch.randint(0, len(self.grouped_examples[anchor_label]), (1, )).item()
positive_index = self.grouped_examples[anchor_label][random_index]
positive_image = self.data[positive_index].float()
random_index = torch.randint(0, len(self.grouped_examples[neg_label]), (1, )).item()
negative_index = self.grouped_examples[neg_label][random_index]
negative_image = self.data[negative_index].float()
```

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,322 @@
...
+        total_correct += engine.state.output
+
+    # event handler triggers evauator at end of every epoch
+    @trainer.on(Events.EPOCH_COMPLETED(every=args.log_interval))
+    def test(engine):
+        nonlocal total_correct
+        total_correct = 0
+        evaluator.run(test_loader)
+        print(f"Accurate predictions: {total_correct}/{len(test_loader.dataset)} ")
+        print(f"Average Accuracy: {(total_correct*100)/(len(test_loader.dataset )): .5f}")
```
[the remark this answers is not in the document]
DeepC004: I have made the requested changes, added ignite.metrics.Accuracy

examples/siamese_network/siamese_network.py
DeepC004: Sorry I realize it now that the `RunningAverage` is not a feasible metric since the function was returning the number of `correct` predictions. I have changed the metric to accuracy for the `evaluator` engine.

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,311 @@
...
+
+    # attach Running Average Loss metric to trainer and evaluator engines
+    RunningAverage(output_transform=lambda x: x).attach(trainer, "loss")
+    Accuracy(output_transform=lambda x: x).attach(evaluator, "accuracy")
+
+    # attach progress bar to trainer with loss
+    pbar1 = ProgressBar()
+    pbar1.attach(trainer, metric_names=["loss"])
+
+    # attach progress bar to evaluator with loss
```
vfdev-5, line 242:
```suggestion
    # attach progress bar to evaluator
```

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+
+    def __init__(self):
+        super(SiameseNetwork, self).__init__()
+        # get resnet model
+        self.resnet = torchvision.models.resnet34(weights=None)
+        self.fc_in_features = self.resnet.fc.in_features
+
+        # changing the FC layer of resnet model to a linear layer
+        self.resnet.fc = nn.Identity(self.fc_in_features)
```
vfdev-5, line 40:
```suggestion
        self.resnet.fc = nn.Identity()
```

examples/siamese_network/siamese_network.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # add linear layers to compare between the features of the two images
+        self.fc = nn.Sequential(
+            nn.Linear(self.fc_in_features, 256),
```
vfdev-5, line 44:
```suggestion
            nn.Linear(fc_in_features, 256),
```

vfdev-5 (commented):
Few other nits and it will be good to go

vfdev-5 (approved):
LGTM, thanks @DeepC004 
Next step is to integrate this example into CI. 
Please, check this PR as example: https://github.com/pytorch/ignite/pull/2899

DeepC004:
I will look into it. Thank you!

Issue #1288, opened 2020-09-14
Customization of reduction/gathering ops for Metric


vfdev-5:
## 🚀 Feature

Idea is to make configurable `Metric`'s reduction/gathering ops. By default, we are using our code, but user can globally override those functions. For example, if uses a custom unsupported distributed framework, or deals with asymmetry like [here](https://github.com/pytorch/ignite/issues/1242) etc

sdesrozis:
A really nice feature

PR #2895, opened 2023-03-16
Add knob (IGNITE_DISABLE_DISTRIBUTED_METRICS=1) to disable distributed metrics reduction

iXce:
This is useful for setups where distributed training is used, but evaluation is only performed on a single node (or independently over multiple nodes).

Description: In some setups one may want to have two different fleets for distributed training and (distributed) validation, however ignite currently conflates the two world sizes. Using this PR as a RFC before adding tests and extra documentation (if needed).

Check list:

- [?] New tests are added (if a new feature is added)
- [?] New doc strings: description and/or example code are in RST format
- [?] Documentation is updated (if required)

### DIFF
--- a/ignite/contrib/metrics/precision_recall_curve.py
+++ b/ignite/contrib/metrics/precision_recall_curve.py
@@ -92,7 +92,7 @@ class PrecisionRecallCurve(EpochMetric):
             _prediction_tensor = torch.cat(self._predictions, dim=0)
             _target_tensor = torch.cat(self._targets, dim=0)
 
-            ws = idist.get_world_size()
+            ws = idist.get_metrics_computation_world_size()
             if ws > 1:
                 # All gather across all processes
                 _prediction_tensor = cast(torch.Tensor, idist.all_gather(_prediction_tensor))
--- a/ignite/distributed/utils.py
+++ b/ignite/distributed/utils.py
@@ -1,3 +1,4 @@
+import os
 import socket
 from functools import wraps
 from typing import Any, Callable, List, Mapping, Optional, Tuple, Union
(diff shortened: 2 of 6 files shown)


sadra-barikbin:
Thanks @iXce for the suggestion. We could also add an boolean attribute to `Metric` which all metrics descend from, called `compute_per_rank`  and give it to the constructor of a metric.
But could you please explain why do you need to evaluate each rank separately?

iXce:
> Thanks @iXce for the suggestion. We could also add an boolean attribute to `Metric` which all metrics descend from, called `compute_per_rank` and give it to the constructor of a metric. But could you please explain why do you need to evaluate each rank separately?

The most typical use case would be that we run distributed training but only run validation on the chief worker (because the validation set is small, or uses a different procedure than training that doesn't necessarily work distributed training).

I was a bit puzzled initially to see that the ignite metrics automatically react to the usage of distributed training (practically speaking we were seeing them hang as they were waiting to allreduce() but none of the other workers would participate).

Regarding where to place the knob for this, it feels like a tough question: in a sense, one might want to be able to switch from single-worker to distributed-validation&metrics configuration without having to reconfigure the metrics, I think? So a constructor parameter to `Metric` may be somewhat cumbersome?

vfdev-5:
@iXce thanks for the RFC! We also wanted to be able to override metrics decorators responsible for reducing data: https://github.com/pytorch/ignite/issues/1288. So, you could override it and do a no-op for your use cases. Or provide a DDP subgroup and reduce over a subgroup.

iXce:
> @iXce thanks for the RFC! We also wanted to be able to override metrics decorators responsible for reducing data: #1288. So, you could override it and do a no-op for your use cases. Or provide a DDP subgroup and reduce over a subgroup. What do you think ?

Oh yeah that looks super relevant indeed!

PR #2898, opened 2023-03-22
Create get_clearml_task for Ignite.contrib.handlers.ClearMLLogger

### LINKED ISSUE #2888: ClearML support for configuration
## ❓ Questions/Help/Support

Hello! I'd like to know how to transfer the ClearML **configuration** logs from pure PyTorch to PyTorch-Ignite, that is:
```
task = Task.init("project-name", task_name="task-name")
task.connect_configuration(config) # or task.connect({k: v for k, v in config.items()})
```
In official docs, I only found `ClearMLLogger` class which implemented `Task.init()` inside, but with no support for `Task.connect()`. Is there any good solution to this problem in current verison of release? 

Thanks!

sallycaoyu:
Fixes #2888 

Description:
- Added a new method `get_clearml_task` for the `Ignite.contrib.handlers.ClearMLLogger` class.
- Wrote and passed test cases locally.

Check list:


sallycaoyu:
Hello @vfdev-5, for test cases `test_clearml_logger_get_clearml_task_not_bypass_create_task` and  `test_clearml_logger_get_clearml_task_not_bypass_task_already_exists` I added, I set ClearMLLogger's bypass mode to false. This requires ClearML to be configured on the machine before running the unit tests. 

I ran both tests locally and they worked, but they failed the Github checks. The error message says ClearML is not configured on the machine for running GitHub checks. And there doesn't seem to be any unit test that sets bypass mode to False except for the two I mentioned above. But still, I wonder if there is a way to test this scenario by bypassing the ClearML configuration? Any advice is appreciated!

ignite/contrib/handlers/clearml_logger.py
```diff
@@ -178,6 +180,20 @@ def bypass_mode(cls) -> bool:
         """
         return getattr(cls, "_bypass", bool(os.environ.get("CI")))
 
+    def get_clearml_task(self) -> Union[None, Task]:
```
vfdev-5, line 183:
```suggestion
    def get_task(self) -> Union[None, Task]:
```

ignite/contrib/handlers/clearml_logger.py
```diff
@@ -178,6 +180,20 @@ def bypass_mode(cls) -> bool:
...
+
+        Return:
+            If bypass mode state is ``True``, all outside communication is skipped,
+            and ``get_task`` will return ``None``.
+            If bypass mode state is ``False``, this method will return the ClearML Task.
+        """
+        if self.bypass_mode():
+            return None
+
+        return self._task
```
vfdev-5, line 195:
```suggestion
        return self._task
```
Let's just return `self._task`

ignite/contrib/handlers/clearml_logger.py
sallycaoyu: I have changed this as suggested in the new commit, but the reason I initially did not use `get_task` is because clearml.Task has a classmethod that has the same name: `Task.get_task()`.

sallycaoyu:
@vfdev-5 Thank you for your comments! I have addressed point 1 & 2 in #2888. I have changed the method name to `get_task`, but in fact clearml.Task has `Task.get_task()` with the same method name. Please let me know if this needs to be further changed to avoid duplicates.

Also, if to test `get_task`'s return value when bypass is True, `self._task` does not return a `Task` object, but returns a `_Stub` object instead. The implementation of the _Stub class makes it unrelated to the Task type, so I cannot call any `Task` method on `self._task` to verify it. Are there any possible ways to test it when bypass is True?

vfdev-5:
Thanks for the updates @sallycaoyu !

> Please let me know if this needs to be further changed to avoid duplicates.

I think it is OK with `ClearML.get_task`.

> Also, if to test get_task's return value when bypass is True, self._task does not return a Task object, but returns a _Stub object instead. The implementation of the _Stub class makes it unrelated to the Task type, so I cannot call any Task method on self._task to verify it. I also tried building mock objects but they did not work either. Are there any possible ways to test it when bypass is True?

It is OK that `get_task()` returns `_Stub`. It was a way to implement a testable ClearMLLogger class by ClearML folks. This way there is no external communication to their server and creating a real task. You can just if returned instance is `_Stab`.

sallycaoyu:
@vfdev-5 Thanks for the explanation! Have added a test case for `get_task`.

vfdev-5:
@sallycaoyu I just checked clearml docs and there is built-in "offline" option: https://clear.ml/docs/latest/docs/clearml_sdk/task_sdk#offline-mode
Maybe, we can use that instead of `_Stub`  and write into a local temp storage. cc @jkhenning

ignite/contrib/handlers/clearml_logger.py
```diff
@@ -178,6 +178,24 @@ def bypass_mode(cls) -> bool:
         """
         return getattr(cls, "_bypass", bool(os.environ.get("CI")))
 
+    def __getattr__(self, attr: Any) -> Any:
+        """
+        Calls the corresponding method of ``ClearML.Logger``.
```
vfdev-5, line 183:
```suggestion
        Calls the corresponding method of ``clearml.Logger``.
```

Class is exactly `clearml.Logger`: https://github.com/allegroai/clearml/blob/master/clearml/__init__.py

ignite/contrib/handlers/clearml_logger.py
```diff
@@ -178,6 +178,24 @@ def bypass_mode(cls) -> bool:
...
+
+        Args:
+            attr: methods of the ClearML.Logger class.
+        """
+        return getattr(self.clearml_logger, attr)
+
+    def get_task(self) -> Callable:
```
vfdev-5, line 190: Return type is incorrect here, let's make it `Any`

ignite/contrib/handlers/clearml_logger.py
sallycaoyu: Fixed.

ignite/contrib/handlers/clearml_logger.py
sallycaoyu: Fixed.

tests/ignite/contrib/handlers/test_clearml_logger.py
sallycaoyu, line 1: On line 721-722 and line 749-750, I changed them this way because `clearml.Logger.report_scalar` accepts float rather than string.

sallycaoyu:
@vfdev-5 Thanks for the suggestion! I have changed the task in bypass mode from a `_Stub` to an offline task, for which all data and logs are stored in a temp local folder.

vfdev-5 (commented):
Thanks for the update @sallycaoyu !
Just a minor comment to address and it seems to be good to go

sallycaoyu:
@vfdev-5 Thanks for the review! I have removed the "return".

vfdev-5 (approved):
LGTM, thanks for the PR @sallycaoyu !

jkhenning:
@vfdev-5 tests look good to me

sallycaoyu:
@vfdev-5 In `tests/ignite/contrib/handlers/test_clearml_logger.py`, I changed the Task object initialization in some unit tests that were
> clearml.Task.current_task = Mock(return_value=object())

to
> clearml.Task.current_task = MagicMock(spec=clearml.Task)

After the change, running the command you provided passed all tests locally.

PR #2885, opened 2023-03-03
Add the example of super_resolution

guptaaryan16:
Fixes #2878 

Description: I have added the example for the super resolution in the `PyTorch/examples`

Check list:

- [x] New tests are added (if a new feature is added)
- [ ] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #2899, opened 2023-03-22
Add CI for Super Resolution example and tqdm bar to the example

guptaaryan16:
Fixes #2885 

Description: I have added the SR example to CI and it works on the unit tests

Check list:


examples/super_resolution/main.py
```diff
@@ -45,7 +47,7 @@
 
 
 class SRDataset(torch.utils.data.Dataset):
-    def __init__(self, dataset, scale_factor, crop_size=256):
+    def __init__(self, dataset, scale_factor, crop_size=180):
```
vfdev-5, line 50: If you tweak this value for CI, let's make it as a parameter and keep default as 256

guptaaryan16: ok let me do that

guptaaryan16: Hey @vfdev-5 , Actually I tested it again and it may not work for `crop_size=256` always as some images may have a lower size than 200 pixels. So I have made it an argument of `main.py` that can be variable depending upon the choice of dataset for training and the `upscale_factor`

examples/super_resolution/main.py
```diff
@@ -145,4 +149,6 @@ def checkpoint():
     print("Checkpoint saved to {}".format(model_out_path))
 
 
+ProgressBar().attach(trainer)
```
vfdev-5, line 152: Let's remove `log_training_loss` handler and update `ProgressBar` options to display similar things.

vfdev-5 (commented):
Thanks for the PR @guptaaryan16 
Can you retrain a bit longer a mode and add to the example README a matplotlib image of super-resolved CIFAR10 image vs bicubic upsampled image (similar to Colab).

examples/super_resolution/main.py
```diff
@@ -145,4 +149,6 @@ def checkpoint():
...
+
```
guptaaryan16, line 153:
```
if opt.debug:
    @trainer.on(Events.ITERATION_COMPLETED(every=log_interval))
    def terminate():
        trainer.terminate_epoch()
```
Hey @vfdev-5 I tried to terminate the current epoch using this(which gets stuck) or the `trainer.terminate()` for the debug mode in this example but the second method fails to create a checkpoint for the model. Can you give me a possible solution for this

vfdev-5: Simplest solution is to use `trainer.run(..., epoch_length=10)`, so the epoch will be equal to 10 iters. However, on iteration 11 (epoch 2) the datapoint will be the next one from the dataloader and wont be the same as at iteration 1.

guptaaryan16:
Hey @vfdev-5 please review

guptaaryan16:
Hey @vfdev-5 I have added the BasicProfileHandler and reduced the time of `debug` option to 11 seconds

examples/super_resolution/README.md
```diff
@@ -30,8 +32,20 @@ This example trains a super-resolution network on the [Caltech101 dataset](https
 
 ### Train
 
-`python main.py --upscale_factor 3 --batch_size 4 --test_batch_size 100 --n_epochs 30 --lr 0.001`
+`python main.py --upscale_factor 3 --crop_size 180 --batch_size 4 --test_batch_size 100 --n_epochs 30 --lr 0.001`
 
 ### Super Resolve
```
vfdev-5, line 37:
```suggestion
### Super-Resolve
```

vfdev-5 (commented):
Looks good, thanks @guptaaryan16 
Just few improvements in the readme and good to go

PR #2900, opened 2023-03-23
[fix] make sure neptune saves the checkpoint

kshitij12345:
Neptune logger doesn't actually save the checkpoint.

Simple reproducer:
```python
import torch
import tempfile
import neptune
import time


class Net(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = torch.nn.Linear(8192, 8192)

    def forward(self, x):
        return self.linear(x)


model = Net()

run = neptune.init_run()

start = time.time()
with tempfile.NamedTemporaryFile() as tmp:
    torch.save(model.state_dict(), tmp.file)
    run['model'].upload(tmp.name)  # Fails because (upload is async and once outside of tmp context the file is deleted).

print(time.time() - start)


start = time.time()
with tempfile.NamedTemporaryFile() as tmp:
    torch.save(model.state_dict(), tmp.file)
    # Uploads in chunks.
    run['model_stream'].upload(neptune.types.File.from_stream(tmp.file))  # Works!

print(time.time() - start)

```

Description:

Check list:

### DIFF
--- a/ignite/contrib/handlers/neptune_logger.py
+++ b/ignite/contrib/handlers/neptune_logger.py
@@ -1,7 +1,7 @@
 """Neptune logger and its helper handlers."""
 import tempfile
-from typing import Any, Callable, List, Mapping, Optional, Union
 import warnings
+from typing import Any, Callable, List, Mapping, Optional, Union
 
 import torch
 from torch.optim import Optimizer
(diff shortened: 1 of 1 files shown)


vfdev-5 (approved):
Thanks @kshitij12345 !

Can we add a unit test for that somehow ?

kshitij12345:
> Can we add a unit test for that somehow ?

I think this can be verified only by actually running a valid `run` and verifying that checkpoint was uploaded correctly. Right now, there is mock test which verifies that `upload` was called but that doesn't guarantee that upload was successful.

vfdev-5:
@kshitij12345 Kshiteej, can you please run the following to fix code formatting issue:
```
bash tests/run_code_style.sh install
bash tests/run_code_style.sh fmt
```

kshitij12345:
Ah, sorry. I missed that. Updated now. Thanks!

PR #2902, opened 2023-03-29
[follow-up] NeptuneSaver fixes

kshitij12345:
Fixes 

* Correctly rewind the temporary files buffer.
* Correctly remove the file from a run. `delete_files` is not supported on `Run` object.
* Update the saver doc to reflect that Windows is not supported (as `TemporaryFile` and `NamedTemporaryFile` don't work correctly on Windows)

Description:

Check list:


vfdev-5 (approved):
Thanks Kshiteej!

PR #2315, opened 2021-11-09
[WIP] added PyTorch Profiler

Ishan-Kumar2:
Fixes #1917 

Description:
Added a minimum implementation of PyTorch profiler as a handler with engine. A lot of other features can be added, please let me know if you have any suggestions. Also I haven't added tests yet, if the initial code looks good, I can get started on those :)

Check list:

- [ ] New tests are added (if a new feature is added)
- [X] New doc strings: description and/or example code are in RST format
- [X] Documentation is updated (if required)

PR #2906, opened 2023-04-01
PyTorch Profiler [WIP]

### LINKED ISSUE #1917: Introduce an handler to use new profiler
## 🚀 Feature

As discussed #1912, a new profiler tool was recently introduced and it would be nice to have a specific handler in ignite.

guptaaryan16:
Fixes #1917 
(uses #2315)

Description: Added check for minimum version of PyTorch for import. Now need to add tests for the Profiler

Check list:


guptaaryan16:
Hey @vfdev-5 @sdesrozis I have added a check for minimum PyTorch version, Now I wanted to ask whether we should keep this profiler in the main module or in the contrib module as most of the handlers introduced afterwards were put there. Also what should be the additional tests that should be present in this handler?

PR #2914, opened 2023-04-12
fixed output_list | now it has only 5 elements instead of 5*batchsize

### LINKED ISSUE #2913: Minor Bug | metrics.ssim, produces empty tensors in update function
## 🐛 Bug description

In line 165
```
output_list = [outputs[x * y_pred.size(0) : (x + 1) * y_pred.size(0)] for x in range(len(outputs))]
```
the list comprehension produces a list with `len  = (Batch_size * 5)`, where only the first 5 elements are valid and correspond to stacked `[y_pred, y, y_pred * y_pred, y * y, y_pred * y]` of all the batches, in cases where the batch size is greater than one the elements with index>4 are empty `torch.Tensors` with shape `(0, C, H, W)` .

### Solution
This bug neither affects the output, nor consumes a lot of RAM, but I thought I should point it out.
The fix for this is pretty simple and you only need to divide the len of outputs by the batch size.
```
output_list = [outputs[x * y_pred.size(0) : (x + 1) * y_pred.size(0)] for x in range(int(len(outputs)/y_pred.size(0)))] # len(outputs) is B*5 so we need to divide it by B so it's only 5 -> [y_pred, y, y_pred * y_pred, y * y, y_pred * y]
```

moienr:
Fixes #2913

Description:
---
This pull request updates the code in `ignite/metrics/ssim.py` so it won't create extra `torch.Tensor`s.
The changes involve modifying the list comprehension that creates the  `output_list`, which has the wrong length of `BatchSize*5` instead of it being `5` (the same as `len([y_pred, y, y_pred * y_pred, y * y, y_pred * y])`). The elements with index>4 are empty tensors with shape `(0,C,H1,W1)` in the  `output_list`

Instead of concatenating the tensor inputs in `torch.cat([y_pred, y, y_pred * y_pred, y * y, y_pred * y])`, I have created a list of the input tensors as `input_list= [y_pred, y, y_pred * y_pred, y * y, y_pred * y]`, which I believe was the actual intention of the author.
Now we have access to the Length of this list, which we will use to retrieve `input_list` elements after passing them through a `conv2d` layer.
I then passed the concatenated input tensor using `torch.cat(input_list)` which has a size of `(B*5, C, H1, W1)` to the `F.conv2d` function call.

After that, the `outputs` which is a tensor with shape `(B*5, C, H, W)` is sliced for every `Batch` element to return a list of `5` elements. this is where the previous code had a problem and instead of a list of `5` (or `len(input_list)`) tensors, it was returning a list of  `B*5` elements, where only its first five elements were actual tensors and the rest where empty when the `Batch_size>1`.
(the previous code was calling `tesnor[index1:index2]` where `index1` and `index2` were larger than `tensor.size(0)` which was returning empty tensors)

We can see in lines `167-173` that the elements with indices `0-4` are being used, which proves this issue.

I have tested these changes on my local machine and confirmed that they do not introduce any new issues. Please let me know if there are any concerns or feedback regarding these changes.

Thanks for considering this pull request!

Check list:

### DIFF
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -159,10 +159,10 @@ class SSIM(Metric):
         y_pred = F.pad(y_pred, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
         y = F.pad(y, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
 
-        input_list = torch.cat([y_pred, y, y_pred * y_pred, y * y, y_pred * y])
-        outputs = F.conv2d(input_list, self._kernel, groups=channel)
-
-        output_list = [outputs[x * y_pred.size(0) : (x + 1) * y_pred.size(0)] for x in range(len(outputs))]
+        input_list= [y_pred, y, y_pred * y_pred, y * y, y_pred * y]
+        outputs = F.conv2d(torch.cat(input_list), self._kernel, groups=channel)
+        
+        output_list = [outputs[x * y_pred.size(0) : (x + 1) * y_pred.size(0)] for x in range(len(input_list))]
 
         mu_pred_sq = output_list[0].pow(2)
         mu_target_sq = output_list[1].pow(2)
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -159,10 +159,10 @@ class SSIM(Metric):
         y_pred = F.pad(y_pred, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
         y = F.pad(y, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
 
-        input_list= [y_pred, y, y_pred * y_pred, y * y, y_pred * y]
+        input_list = [y_pred, y, y_pred * y_pred, y * y, y_pred * y]
         outputs = F.conv2d(torch.cat(input_list), self._kernel, groups=channel)
-        
-        output_list = [outputs[x * y_pred.size(0) : (x + 1) * y_pred.size(0)] for x in range(len(input_list))]
+        batch_size = y_pred.size(0)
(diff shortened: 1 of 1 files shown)


ignite/metrics/ssim.py
```diff
@@ -159,10 +159,10 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
...
         y = F.pad(y, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
 
-        input_list = torch.cat([y_pred, y, y_pred * y_pred, y * y, y_pred * y])
-        outputs = F.conv2d(input_list, self._kernel, groups=channel)
-
-        output_list = [outputs[x * y_pred.size(0) : (x + 1) * y_pred.size(0)] for x in range(len(outputs))]
+        input_list= [y_pred, y, y_pred * y_pred, y * y, y_pred * y]
+        outputs = F.conv2d(torch.cat(input_list), self._kernel, groups=channel)
+        
+        output_list = [outputs[x * y_pred.size(0) : (x + 1) * y_pred.size(0)] for x in range(len(input_list))]
```
vfdev-5, line 165:
```suggestion
        input_list = [y_pred, y, y_pred * y_pred, y * y, y_pred * y]
        outputs = F.conv2d(torch.cat(input_list), self._kernel, groups=channel)
        batch_size = y_pred.size(0)
        output_list = [outputs[x * batch_size : (x + 1) * batch_size] for x in range(len(input_list))]
```

vfdev-5:
@moienr thanks for the explanation and the PR! Right now I see the problem, it should be basically : 
```python
output_list = [outputs[x * batch_size : (x + 1) * batch_size] for x in range(5)]
```
where 5 refers to the size of `input_list`.

I made a suggestion how to fix code style and introduced batch size variable. Please update your PR and make sure to run code style auto formatting to avoid formatting issues:
```
bash tests/run_code_style.sh
```

moienr:
Thank you dear @vfdev-5 I committed your code, and ran the style.
Also, the branch was behind the new updates, so I updated the branch as well.
Thanks again for all your time.

PR #2950, opened 2023-05-20
Improve RunningAverage reset when epoch_bound=False

sadra-barikbin:
Description:
When `epoch_bound=False`, an explicit `metric.reset()` was required before
another `engine.run`. This PR gets `RunningAverage` to reset itself automatically.

Check list:

### DIFF
--- a/ignite/metrics/running_average.py
+++ b/ignite/metrics/running_average.py
@@ -143,6 +143,8 @@ class RunningAverage(Metric):
         if self.epoch_bound:
             # restart average every epoch
             engine.add_event_handler(Events.EPOCH_STARTED, self.started)
+        else:
+            engine.add_event_handler(Events.EPOCH_STARTED(once=1), self.started)
         # compute metric
         engine.add_event_handler(Events.ITERATION_COMPLETED, self.iteration_completed)
         # apply running average
--- a/ignite/metrics/running_average.py
+++ b/ignite/metrics/running_average.py
@@ -144,7 +144,7 @@ class RunningAverage(Metric):
             # restart average every epoch
             engine.add_event_handler(Events.EPOCH_STARTED, self.started)
         else:
-            engine.add_event_handler(Events.EPOCH_STARTED(once=1), self.started)
+            engine.add_event_handler(Events.STARTED, self.started)
         # compute metric
         engine.add_event_handler(Events.ITERATION_COMPLETED, self.iteration_completed)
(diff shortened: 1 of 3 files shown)


ignite/metrics/running_average.py
```diff
@@ -143,6 +143,8 @@ def attach(self, engine: Engine, name: str, _usage: Union[str, MetricUsage] = Ep
         if self.epoch_bound:
             # restart average every epoch
             engine.add_event_handler(Events.EPOCH_STARTED, self.started)
+        else:
+            engine.add_event_handler(Events.EPOCH_STARTED(once=1), self.started)
```
vfdev-5, line 147: Shouldn't it be just 
```suggestion
            engine.add_event_handler(Events.STARTED, self.started)
```
?

tests/ignite/metrics/test_running_average.py
```diff
@@ -187,6 +187,9 @@ def assert_equal_running_avg_output_values(engine):
 
     trainer.run(data, max_epochs=3)
 
+    running_avg_acc[:] = [None]
```
vfdev-5, line 190:
```suggestion
    running_avg_acc = [None]
```
?

tests/ignite/metrics/test_running_average.py
```diff
@@ -187,6 +187,9 @@ def assert_equal_running_avg_output_values(engine):
...
+    trainer.run([1] * n_iters, max_epochs=3)
```
vfdev-5, line 191: and why not reuse previous `data`?
```suggestion
    trainer.run(data, max_epochs=3)
```

vfdev-5:
Thanks for the PR @sadra-barikbin ! I left few comments and please update PR's description as it states : 
> This PR removes that requirement ...

but you add more code, this looks a bit misleading

tests/ignite/metrics/test_running_average.py
vfdev-5: OK, I see, so to reset it we can just do `running_avg_acc[0] = None`

vfdev-5:
by the way, let's simplify this code as well:
```diff
-    @trainer.on(Events.STARTED)
-    def running_avg_output_init(engine):
-        engine.state.running_avg_output = None
+   trainer.state.running_avg_output = None
```

vfdev-5 (approved):
Thanks, lgtm!

PR #2959, opened 2023-05-29
Improve collective ops on numeric input

sadra-barikbin:
Description:
When `ComputationModel._collective_op_dtype` is `None` and the input to a
collective op is a float number, the constructed intermediate tensor is of
dtype `torch.float32` rather than `torch.double`, hence we would have a loss
in precision.

Check list:

### DIFF
--- a/ignite/distributed/comp_models/base.py
+++ b/ignite/distributed/comp_models/base.py
@@ -187,10 +187,9 @@ class ComputationModel(metaclass=ABCMeta):
         device = self.device()
         if isinstance(tensor, (Number, float)):
             tensor_to_number = True
-            if self._collective_op_dtype is None:
-                dtype = torch.double if isinstance(tensor, float) else torch.int64
-            else:
-                dtype = self._collective_op_dtype
+            dtype = self._collective_op_dtype
+            if dtype is None and isinstance(tensor, float):
+                dtype = torch.double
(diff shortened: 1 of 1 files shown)


ignite/distributed/comp_models/base.py
```diff
@@ -187,7 +187,11 @@ def _collective_op(
         device = self.device()
         if isinstance(tensor, (Number, float)):
             tensor_to_number = True
-            tensor = torch.tensor(tensor, device=device, dtype=self._collective_op_dtype)
+            if self._collective_op_dtype is None:
+                dtype = torch.double if isinstance(tensor, float) else torch.int64
+            else:
+                dtype = self._collective_op_dtype
```
vfdev-5, line 193:
```suggestion
            dtype = self._collective_op_dtype
            if dtype is None and isinstance(tensor, float):
                dtype = torch.double
```

vfdev-5:
@sadra-barikbin where do you spot such precision loss ?

sadra-barikbin:
> @sadra-barikbin where do you spot such precision loss ?

When I attempted to remove the need to use `pytest.approx` in `_test_distrib_on_output`, I ended up here.
After this PR I saw that there was no need to `approx` and `==` worked even without this change, however this change is correct.

vfdev-5 (approved):
LGTM, thanks for the PR @sadra-barikbin

PR #2982, opened 2023-07-03
[Follow-up] Fix a tiny bug in tests of Improvement in RunningAverage

sadra-barikbin:


vfdev-5:
@sadra-barikbin can you check what happens with Pillow on failed CI ?

vfdev-5 (approved):
Thanks @sadra-barikbin

PR #3001, opened 2023-07-15
Make metrics serializable

### LINKED ISSUE #2999: Add `state_dict` and `load_state_dict` to metrics
## 🚀 Feature

To add `state_dict` and `load_state_dict` to metrics to be able to save them to checkpoints and update them later.

sadra-barikbin:
Fixes #2999


sadra-barikbin:
Shall I add `_state_dict_all_req_keys` attribute to all of metrics?
Shall I add tests for `state_dict` and `load_state_dict` for each metric?

ignite/metrics/metric.py
```diff
@@ -546,6 +550,41 @@ def is_attached(self, engine: Engine, usage: Union[str, MetricUsage] = EpochWise
...
+        for attr_name in self._state_dict_all_req_keys:
+            attr = getattr(self, attr_name)
+            if idist.get_world_size() == 1:
+                state[attr_name] = [attr]
+            else:
+                if isinstance(attr, (float, torch.Tensor)):
+                    state[attr_name] = cast(List[Any], idist.all_gather(attr))
+                else:
+                    state[attr_name] = [None] * idist.get_world_size()
+                    torch_dist.all_gather_object(state[attr_name], attr)
```
vfdev-5, line 570: Let's raise an error here now. We can add a support for picklable objects later with `idist.all_gather`.

ignite/metrics/metric.py
```diff
@@ -546,6 +550,41 @@ def is_attached(self, engine: Engine, usage: Union[str, MetricUsage] = EpochWise
...
+    def state_dict(self) -> OrderedDict:
+        """Method returns state dict with attributes of the metric specified in its
+        `_state_dict_all_req_keys` attribute. Can be used to save internal state of the class.
+
+        If there's an active distributed configuration, some collective operations is done and
+        the list of values across ranks is saved under each attribute's name in the dict.
+        """
+        state = OrderedDict()
+        for attr_name in self._state_dict_all_req_keys:
+            attr = getattr(self, attr_name)
```
vfdev-5, line 562: Let's add a check if attr_name does not exist.

ignite/metrics/metric.py
```diff
@@ -546,6 +550,41 @@ def is_attached(self, engine: Engine, usage: Union[str, MetricUsage] = EpochWise
...
+
+        If there's an active distributed configuration, the process uses its rank to pick the proper value from
+        the list of values saved under each attribute's name in the dict.
+
+        Args:
+            state_dict: a dict containing attributes of the metric specified in its `_state_dict_all_req_keys`
+                attribute.
+        """
+        super().load_state_dict(state_dict)
+        rank = idist.get_local_rank()
```
vfdev-5, line 584: It should be rank and not local_rank here, right ?

vfdev-5 (commented):
The approach looks OK to me, let's update all metrics.

docs/source/engine.rst
```diff
@@ -82,8 +82,9 @@ store the trainer and then resume the training. For example:
     optimizer = ...
     lr_scheduler = ...
     data_loader = ...
+    metric = ...
 
-    to_save = {'trainer': trainer, 'model': model, 'optimizer': optimizer, 'lr_scheduler': lr_scheduler}
+    to_save = {'trainer': trainer, 'model': model, 'optimizer': optimizer, 'lr_scheduler': lr_scheduler, 'metric': metric}
```
vfdev-5, line 87: Ideally, we would like to get that in a more automatic way. Also if we want to save multiple metrics adding them one by one to save/load can be a pain.

vfdev-5 (approved):
LGTM, thanks @sadra-barikbin

ignite/metrics/running_average.py
```diff
@@ -87,6 +87,9 @@ def log_running_avg_metrics():
     """
 
     required_output_keys = None
+    # TODO Shall we put `src` here? Then we should add a new branch for metric-typed attributes in `state_dict`
+    # and `load_state_dict`. Examples; This class; `Rouge` which has a `List[_BaseRouge]`.
+    _state_dict_all_req_keys = ("_value",)
```
sadra-barikbin, line 92: @vfdev-5 , What's your thoughts on this?

vfdev-5: Let's add `src` and in state_dict/load_state_dict put an explicit check for `isinstance(attr, Metric)` and call `state[attr_name].update(attr.state_dict)` or something like that.
Do you have any other options ?

docs/source/engine.rst
sadra-barikbin: Could you please reiterate what you mean? If we have multiple metrics, is it cumbersome to do:
```python
to_save = {..., 'metric1': metric1, 'metric2': metric2, 'metric3': metric3}
```
?

vfdev-5: Yes, it is too verbose, let's try to find a way to simplify that.

PR #3014, opened 2023-07-31
Fix for wrong output of Discord Extension and Workflow

guptaaryan16:
Fixes : Discord Workflow

Description:

Check list:


.github/workflows/discord_issues.yml
```diff
@@ -1,4 +1,4 @@
-name: Discuss on Discord
+name: Discuss on Discord (Issues)
```
vfdev-5, line 1:
```suggestion
name: Discuss "help-wanted" issue on Discord
```

.github/workflows/discord_pull_requests.yaml
```diff
@@ -0,0 +1,28 @@
+name: Discuss on Discord
```
vfdev-5, line 1:
```suggestion
name: Discuss "help-wanted" PR on Discord
```

.github/workflows/discord_issues.yml
```diff
@@ -23,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: "Discuss on Discord-Issues"
-        if: github.event.issue && contains(github.event.issue.labels.*.name, 'help wanted')
+        if: ${{ github.event.label.name == 'help wanted' }}
```
vfdev-5, line 25: Why can't we have exact match like here and have two steps (one for issue and one for PR) as before ?

vfdev-5 (commented):
Thanks for the PR @guptaaryan16 
I have few questions about this approach

.github/workflows/discord_issues.yml
guptaaryan16: @vfdev-5 So basically I tried the same but it seems difficult to control the label type when we try to do it in a single file with both issue and PR, so I think this is more predictable and reliable as before the change

vfdev-5 (approved):
Thanks for the fix @guptaaryan16

PR #3034, opened 2023-08-21
Fix: change the dtype of self._kernel when input args have a different dtype

MarcBresson:
When the SSIM metric is given a float16 (Half) or float64 (Double), it fails because self._kernel will be of the default type.

There is two ways around that:
1. set the default dtype to Half or Double: `torch.set_default_dtype(torch.float16)`.
2. change self._kernel to take the dtype of the input tensor.

This pull request put the second option in action. The result is now transparent to the user who can use float16 or float64 **on GPU** directly.

A new test has been added to ensure of the right behaviour of this change.

Check list:

### DIFF
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -13,7 +13,10 @@ class SSIM(Metric):
     """
     Computes Structural Similarity Index Measure
 
-    - ``update`` must receive output of the form ``(y_pred, y)``.
+    - ``update`` must receive output of the form ``(y_pred, y)``. They have to be of the same type.
+        Valid :class:`torch.dtype` are the following:
+        - on CPU: torch.float32.
+        - on cuda: torch.float16, torch.float32, torch.float64.
 
     Args:
         data_range: Range of the image. Typically, ``1.0`` or ``255``.
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -161,7 +161,7 @@ class SSIM(Metric):
         y_pred = F.pad(y_pred, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
         y = F.pad(y, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
 
-        if y_pred.dtype != torch.get_default_dtype():
+        if y_pred.dtype != self._kernel.dtype:
             self._kernel = self._kernel.to(dtype=y_pred.dtype)
 
         input_list = [y_pred, y, y_pred * y_pred, y * y, y_pred * y]
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -15,8 +15,8 @@ class SSIM(Metric):
 
     - ``update`` must receive output of the form ``(y_pred, y)``. They have to be of the same type.
         Valid :class:`torch.dtype` are the following:
-        - on CPU: torch.float32.
-        - on cuda: torch.float16, torch.float32, torch.float64.
+        - on CPU: `torch.float32`, `torch.float64`.
+        - on cuda: `torch.float16`, `torch.float32`, `torch.float64`.
 
     Args:
         data_range: Range of the image. Typically, ``1.0`` or ``255``.
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -16,7 +16,7 @@ class SSIM(Metric):
     - ``update`` must receive output of the form ``(y_pred, y)``. They have to be of the same type.
         Valid :class:`torch.dtype` are the following:
         - on CPU: `torch.float32`, `torch.float64`.
-        - on cuda: `torch.float16`, `torch.float32`, `torch.float64`.
+        - on CUDA: `torch.float16`, `torch.bfloat16`, `torch.float32`, `torch.float64`.
 
     Args:
         data_range: Range of the image. Typically, ``1.0`` or ``255``.
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -161,6 +161,9 @@ class SSIM(Metric):
         y_pred = F.pad(y_pred, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
         y = F.pad(y, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
 
+        if y_pred.dtype != torch.get_default_dtype():
+            self._kernel = self._kernel.to(dtype=y_pred.dtype)
+
         input_list = [y_pred, y, y_pred * y_pred, y * y, y_pred * y]
         outputs = F.conv2d(torch.cat(input_list), self._kernel, groups=channel)
         batch_size = y_pred.size(0)
--- a/tests/ignite/metrics/test_ssim.py
+++ b/tests/ignite/metrics/test_ssim.py
@@ -80,6 +80,9 @@ def test_ssim(available_device, shape, kernel_size, gaussian, use_sample_covaria
     ssim.update((y_pred, y))
     ignite_ssim = ssim.compute()
 
+    if y_pred.dtype == torch.bfloat16:
+        y_pred = y_pred.to(dtype=torch.float16)
+
     skimg_pred = y_pred.cpu().numpy()
     skimg_y = skimg_pred * 0.8
     skimg_ssim = ski_ssim(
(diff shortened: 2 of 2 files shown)


ignite/metrics/ssim.py
```diff
@@ -161,6 +161,9 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
         y_pred = F.pad(y_pred, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
         y = F.pad(y, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
 
+        if y_pred.dtype != torch.get_default_dtype():
+            self._kernel = self._kernel.to(dtype=y_pred.dtype)
```
vfdev-5, line 165: Should not we compare to `self._kernel.dtype` here ?
```suggestion
        if y_pred.dtype != self._kernel.dtype:
            self._kernel = self._kernel.to(dtype=y_pred.dtype)
```

MarcBresson: we can, but I think they always are going to be equal

```python
self._kernel.dtype == torch.get_default_dtype()
>>> True
```

MarcBresson: but that would make the change more clear to others, so maybe it is a good idea

vfdev-5: If user explicitly define y and y_pred with torch.float64 dtype and `torch.get_default_dtype()` will be still float32, so the error will remain.

vfdev-5:
@MarcBresson thanks for a fix, can you also write a test here: https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/test_ssim.py, let's create a separate test function like `test_ssim_variable_batchsize`.

MarcBresson:
Nice idea to write the test ! I just discovered that a few operation are not available in float16 on CPU.

Working on a fix.

MarcBresson:
Just need a little help, how to exclude a test from being tested on CPU?

I don't see any exclusion pattern in `run_cpu_tests.sh`

vfdev-5:
> Just need a little help, how to exclude a test from being tested on CPU? I don't see any exclusion pattern in run_cpu_tests.sh

I think the simpliest way to do that is just call pytest:
```
pytest -vvv tests/ignite/metrics/test_ssim.py -k <pattern>
```

> And by the way, how is it possible that some tests failed while I haven't touched a thing yet in this remote repo?

GH CI is rather flaky so some jobs are failing due to issues with distributed configs for the tests. Do not worry about that.

tests/ignite/metrics/test_ssim.py
```diff
@@ -123,6 +123,25 @@ def test_ssim_variable_batchsize(available_device):
     assert np.allclose(out, expected)
 
 
+def test_cuda_ssim_dtypes(available_device):
+    # Checks https://github.com/pytorch/ignite/pull/3034
+    # this test should not be run on CPU
```
vfdev-5, line 128: Why this test should not run on CPU ?

tests/ignite/metrics/test_ssim.py
```diff
@@ -123,6 +123,25 @@ def test_ssim_variable_batchsize(available_device):
...
+    if available_device == "cpu":
+        assert True
+        return
+
+    for dtype in [torch.float16, torch.float32, torch.float64]:
```
vfdev-5, line 132: We can use pytest.mark.parametrize for this for-loop:
```
@pytest.mark.parametrize("dtype", [torch.float16, torch.float32, torch.float64])
def test_ssim_dtypes(available_device, dtype):
    test_ssim(available_device, ..., dtype=dtype)
```

tests/ignite/metrics/test_ssim.py
MarcBresson: Because the F.pad operation does not exist for Half type on CPU

MarcBresson: I could check on CPU if the code effectively gives a RuntimeError for Half and Double type.

vfdev-5: I think float32 and double should be supported on cpu, so let's run on it

MarcBresson: You are right ! I just tested. Thank you for your help :)

ignite/metrics/ssim.py
```diff
@@ -13,7 +13,10 @@ class SSIM(Metric):
     """
     Computes Structural Similarity Index Measure
 
-    - ``update`` must receive output of the form ``(y_pred, y)``.
+    - ``update`` must receive output of the form ``(y_pred, y)``. They have to be of the same type.
+        Valid :class:`torch.dtype` are the following:
+        - on CPU: `torch.float32`, `torch.float64`.
```
vfdev-5, line 18: Can you please check if torch.bfloat16 is supported  ? 
https://pytorch.org/docs/stable/tensors.html

ignite/metrics/ssim.py
```diff
@@ -13,7 +13,10 @@ class SSIM(Metric):
...
+        - on cuda: `torch.float16`, `torch.float32`, `torch.float64`.
```
vfdev-5, line 19:
```suggestion
        - on CUDA: `torch.float16`, `torch.float32`, `torch.float64`.
```

tests/ignite/metrics/test_ssim.py
vfdev-5: Cool, sounds good, can you please also remove above comment as it is less relevant:
```
    # Checks https://github.com/pytorch/ignite/pull/3034
    # this test should not be run on CPU
```

ignite/metrics/ssim.py
MarcBresson: I already tried on GPU it didn't work, will try on CPU :)

MarcBresson: On CPU:
- RuntimeError: "reflection_pad2d" not implemented for 'Half'
- RuntimeError: "reflection_pad2d" not implemented for 'BFloat16

On GPU:
- numpy doesn't support Bfloat16 (yet), so there is no baseline comparison. Below are the distribution of the difference between float16 or bloat16 in torch and float16 in scikit-image. So it is possible to create a special case for bfloat16 where it is tested against a numpy float16 array.

image
image

vfdev-5: Thanks for the analysis, yes, it makes sense to raise atol value to ~1e-3 for float16/bfloat16 dtypes due to their low-precision.

MarcBresson: Here is the scale comparison against all types on CPU and GPU.

image

I will add a parameter in my test function for the precision that will be linked to the dtype.

vfdev-5: Would it make more sense to plot absolute difference instead of "difference value" ?

vfdev-5:
@MarcBresson can you run the following to fix formatting issue: 
```
```

vfdev-5 (approved):
Looks good to me, thanks for the PR @MarcBresson !

MarcBresson:
@vfdev-5 you are welcome! Happy to help.

I think that `self._kernel` is already on the right device. It originates either from https://github.com/MarcBresson/ignite/blob/master/ignite/metrics/ssim.py#L127 or https://github.com/MarcBresson/ignite/blob/master/ignite/metrics/ssim.py#L116 and they are both created directly on the device.

The `.to(device)` (https://github.com/MarcBresson/ignite/blob/master/ignite/metrics/ssim.py#L162) originates from the initial SSIM commit (https://github.com/MarcBresson/ignite/commit/4a7428cbcce71cc16a5c1aa37fe67f2005b8817d). I think it has just been forgotten to be removed on this commit https://github.com/MarcBresson/ignite/commit/f86d42e7176ca5e32bab9a7fa3c372ce8d53c3a3.

vfdev-5:
Well, this code:
```
        if len(self._kernel.shape) < 4:
```
seems a bit weird as only kernel with ndims < 4 will have y_pred.device and do we need to change `self._kernel` device from `self._device` to y_pred.device...
I think we should handle the device separately from the shape expand...

MarcBresson:
indeed, it only happens when `kernel_size` is equal to 3. If I have time I will open a new PR for that matter.

MarcBresson:
Just a new analysis on dtype. These are the comparisons of the precision between float32 and float16, bfloat16 and float64. The error is not absolute so we can see if a dtype overshoots compared to float32.

image
image
image

```python
import tqdm
import torch
from ignite.metrics import SSIM
```
```python
def test_ssim_dtype_error(available_device, shape, dtype):
    y_pred_float32 = torch.rand(shape, device=available_device, dtype=torch.float32)
    y_float32 = y_pred_float32 * 0.8

    y_pred_typed = torch.clone(y_pred_float32)
    y_pred_typed = y_pred_typed.to(dtype=dtype)
    y_typed = y_pred_typed * 0.8

    sigma = 1.5
    data_range = 1.0
    ssim = SSIM(data_range=data_range, sigma=sigma, device=available_device)
    ssim.update((y_pred_float32, y_float32))
    ssim_float32 = ssim.compute()

    ssim_ty = SSIM(data_range=data_range, sigma=sigma, device=available_device)
    ssim_ty.update((y_pred_typed, y_typed))
    ssim_typed = ssim_ty.compute()

    difference = ssim_float32 - ssim_typed
    return difference

diff = {}

for device in [torch.device("cuda")]:
    available_dtypes = [torch.float32, torch.float64]
    if device == torch.device("cuda"):
        available_dtypes.extend((torch.bfloat16, torch.float16))

    for dtype in available_dtypes:

        diff[str(dtype) + " on " + str(device)] = []

        for _ in tqdm.tqdm(range(1000)):
            difference_ssim = test_ssim_dtype_error(device, (12, 3, 28, 28), dtype=dtype)
            diff[str(dtype) + " on " + str(device)].append(difference_ssim)
```
```python
for dtype in diff:
    plt.hist(diff[dtype], bins=50)
    axes = plt.gca()
    axes.ticklabel_format(axis='x', style="scientific", scilimits=(0, 0))

    plt.title(f"SSIM difference between float32 and a {dtype}")

    plt.xlabel("difference value")
    plt.ylabel("frequency")

    plt.show()
```

PR #3035, opened 2023-08-21
[Follow-up] Make `RunningAverage` and `Rouge` serializable

sadra-barikbin:
Now `state_dict` could serialize metric attributes that themselves are metric or simple collections of metrics.(For `Rouge`). If an attribute is `None`, `state_dict` sets the same in the dict (for `_value` in `RunningAverage`)

### DIFF
--- a/ignite/metrics/metric.py
+++ b/ignite/metrics/metric.py
@@ -554,31 +554,40 @@ class Metric(Serializable, metaclass=ABCMeta):
         `_state_dict_all_req_keys` attribute. Can be used to save internal state of the class.
 
         If there's an active distributed configuration, some collective operations is done and
-        the list of values across ranks is saved under each attribute's name in the dict.
+        the list of values across ranks is saved under each attribute's name in the dict, for numeric
+        and tensor values.
         """
-        state = OrderedDict()
+        state: Dict[str, Union[torch.Tensor, List, Dict, None]] = OrderedDict()
         for attr_name in self._state_dict_all_req_keys:
             if attr_name not in self.__dict__:
                 raise ValueError(
                     f"Found a value in _state_dict_all_req_keys that is not among metric attributes: {attr_name}"
                 )
             attr = getattr(self, attr_name)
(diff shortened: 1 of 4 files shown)


ignite/metrics/metric.py
```diff
@@ -554,32 +554,41 @@ def state_dict(self) -> OrderedDict:
...
+                    state[attr_name] = [attr]
+                else:
+                    if isinstance(attr, (int, float)):
+                        attr_type = type(attr)
+                        attr = float(attr)
+                    gathered_attr = idist.all_gather(attr)
+                    if isinstance(attr, float):
+                        gathered_attr = [attr_type(process_attr) for process_attr in cast(torch.Tensor, gathered_attr)]
+                    state[attr_name] = cast(Union[torch.Tensor, List], gathered_attr)
+            elif attr is None:
```
vfdev-5, line 584: What is the usage for None ?

vfdev-5 (approved):
Thanks @sadra-barikbin, PR looks good to me!
Can you please check and report here how it works on MNIST while saving and resuming.
As a test example, please use: 
https://github.com/pytorch/ignite/blob/11a1fba0594f6ac0db9e8bc5f21305dc338bb7c1/examples/mnist/mnist_save_resume_engine.py
Here are few instructions on https://github.com/pytorch/ignite/blob/master/examples/mnist/README.md#training-save--resume
Finally, we can also update `mnist_save_resume_engine.py` code adding RunningAverage metric for batch loss to the trainer and plotting it.

sadra-barikbin:
Here is the running loss until stop and after resume:
image

image

sadra-barikbin:
Regarding your [comment](https://github.com/pytorch/ignite/pull/3001#discussion_r1289662406) in the main PR, I introduced a new if branch in `Checkpoint` to accept a dictionary of serializable objects as an item in the `to_save`. This way we could do
```python
metrics = {'acc': Accuracy(), 'nll': Loss()}
to_save={'model':model, 'trainer':trainer, 'metrics': metrics}
```
How is this?

PR #3043, opened 2023-08-23
feat: improve how device switch is handled between the metric device and the input tensors device

MarcBresson:
@vfdev-5 I investigated on the weird code, and as it turns out, kernel could never be of size > 2 (it is only computed in _uniform() or _gaussian() which both output 2 dim tensors).

I wrote this little fix with a warning if the update tensors are not on the device device than the metric.

Do you know if calling .to(device) on a tensor that is already on the device will cause slow downs ?

Check list:

### DIFF
--- a/tests/ignite/metrics/test_ssim.py
+++ b/tests/ignite/metrics/test_ssim.py
@@ -296,7 +296,10 @@ def test_distrib_accumulator_device(distributed, metric_device):
 
     ssim = SSIM(data_range=1.0, device=metric_device)
 
-    for dev in [ssim._device, ssim._kernel.device]:
+    assert ssim._kernel is None
+    assert isinstance(ssim._kernel_2d, torch.Tensor)
+
+    for dev in [ssim._device, ssim._kernel_2d.device]:
         assert dev == metric_device, f"{type(dev)}:{dev} vs {type(metric_device)}:{metric_device}"
 
     y_pred = torch.rand(2, 3, 28, 28, dtype=torch.float, device=device)


vfdev-5:
Thanks a lot the PR @MarcBresson !


>  I investigated on the weird code, and as it turns out, kernel could never be of size > 2 (it is only computed in _uniform() or _gaussian() which both output 2 dim tensors).

Yes, that's correct and as the code overwrites `self._kernel`:
```
```
and usually the data has consistently the same device, the actual code is working.


Concerning this PR's code, I think it would be good to follow below logic:

kernel | y, y_pred | what to do
---|---|---
cpu | cpu | nothing
cuda | cuda | nothing
cpu | cuda | set kernel.to("cuda") for a temp var and use cuda for F.conv2d op
cuda | cpu | set y_pred, y to cuda

What do you think ?

> Do you know if calling .to(device) on a tensor that is already on the device will cause slow downs ?

There was a slow down in early pytorch version, we should measure that now. Please use torch utils benchmark to get some numbers : https://github.com/vfdev-5/pth-inductor-dev/blob/eb01fa071a2337c7037e8a7e961b2147c5fc8b42/perf_flip.py#L52-L61

MarcBresson:
```python
import torch
import torch.utils.benchmark as benchmark

def rand_to_device(shape: tuple, created_on_device, to_device):
    on_device = torch.rand(shape, device=created_on_device)

    if to_device is not None:
        t = on_device.to(device=to_device)

results = []
min_run_time = 5
shape = (12, 4, 256, 256)
```

```python
available_devices = [torch.device("cuda:0"), torch.device("cpu")]
for from_device in available_devices:
    for to_device in available_devices + [None]:
        print(f"{from_device} to {to_device} measurements")
        results.append(
            benchmark.Timer(
                stmt=f"fn({shape}, created_on_device, to_device)",
                globals={
                    "fn": rand_to_device,
                    "created_on_device": from_device,
                    "to_device": to_device,
                },
                num_threads=torch.get_num_threads(),
                label=f"{from_device} to {to_device} measurements"
            ).blocked_autorange(min_run_time=min_run_time)
        )
```

I had an error when calling `compare = benchmark.Compare(results); compare.print()` so I just printed out the objects:
```python
results
>>> cuda:0 to cuda:0 measurements
  Median: 113.61 us
  IQR:    0.70 us (113.34 to 114.04)
  430 measurements, 100 runs per measurement, 6 threads
>>> cuda:0 to cpu measurements
  Median: 4.95 ms
  IQR:    0.14 ms (4.89 to 5.03)
  11 measurements, 100 runs per measurement, 6 threads
>>> cuda:0 to None measurements
  Median: 110.68 us
  IQR:    0.30 us (110.45 to 110.75)
  46 measurements, 1000 runs per measurement, 6 threads
>>> cpu to cuda:0 measurements
  Median: 23.79 ms
  IQR:    6.15 ms (22.10 to 28.24)
  20 measurements, 10 runs per measurement, 6 threads
  WARNING: Interquartile range is 25.8% of the median measurement.
           This suggests significant environmental influence.
>>> cpu to cpu measurements
  Median: 21.80 ms
  IQR:    3.26 ms (21.11 to 24.37)
  21 measurements, 10 runs per measurement, 6 threads
  WARNING: Interquartile range is 14.9% of the median measurement.
           This could indicate system fluctuation.
>>> cpu to None measurements
  Median: 20.86 ms
  IQR:    5.49 ms (19.88 to 25.37)
  21 measurements, 10 runs per measurement, 6 threads
  WARNING: Interquartile range is 26.3% of the median measurement.
           This suggests significant environmental influence.
```

It seems like there is a really slight slow down (we must compare `cuda:0 to cuda:0` with `cuda:0 to None`).

MarcBresson:
I think that is a great suggestion, always performing the computation on GPU if either one of the kernel or input tensor in on it.

vfdev-5:
Here is how you can make it. I changed a bit the perf code to measure `.to` op only.
```python
import torch
import torch.utils.benchmark as benchmark


results = []
min_run_time = 5
shape = (12, 4, 256, 256)


available_devices = [torch.device("cuda"), torch.device("cpu")]
for shape in [(12, 4, 256, 256), (8, 3, 512, 512)]:
    for from_device in available_devices:
        data = torch.rand(shape, device=from_device)
        for to_device in available_devices:
            print(f"{shape} -> {from_device} to {to_device} measurements")
            results.append(
                benchmark.Timer(
                    stmt=f"data.to(to_device, non_blocking=False)",
                    globals={
                        "data": data,
                        "to_device": to_device,
                    },
                    description=f"{from_device} to {to_device}",
                    num_threads=torch.get_num_threads(),
                    label="Device to device",
                    sub_label=f"{tuple(shape)}",
                ).blocked_autorange(min_run_time=min_run_time)
            )

compare = benchmark.Compare(results)
compare.print()
```

On my infra it gives:
```
[---------------------------------- Device to device ---------------------------------]                                                        │| Processes:                                                                            |
                         |  cuda to cuda  |  cuda to cpu  |  cpu to cuda  |  cpu to cpu                                                        │|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
6 threads: ----------------------------------------------------------------------------                                                        │|        ID   ID                                                             Usage      |
      (12, 4, 256, 256)  |     844.8      |   1046673.6   |   1165780.2   |    731.3                                                           │|=======================================================================================|
      (8, 3, 512, 512)   |     868.6      |   2018511.1   |   2439948.9   |    721.6                                                           │|  No running processes found                                                           |
                                                                                                                                               │+---------------------------------------------------------------------------------------+
Times are in nanoseconds (ns). 
```

I agree that it is not a big deal but we also have to think that in the real application `.to` in the code can be repeated N times, where N depends on the dataset times number of epochs, thus overall slowdown may be visible.

vfdev-5:
> Now I wonder what .to(None) does ahahah.

according to https://pytorch.org/docs/stable/generated/torch.Tensor.to.html#torch.Tensor.to
it should go as 
```
torch.to(device=None, dtype=None, non_blocking=False, copy=False, memory_format=torch.preserve_format) → Tensor
```
and probably do nothing. So, we are observing just calling a python function for 250 ns (which is reasonable)

MarcBresson:
(there are bugs, I am writing test and will correct them)

MarcBresson:
I fixed everything. One thing to note is the (assumed) performance hit when self._kernel is not on the same device as y_pred as it will need to be moved onto y_pred's device every time update() is called.

ignite/metrics/ssim.py
```diff
@@ -158,17 +159,31 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
...
-            self._kernel = self._kernel.expand(channel, 1, -1, -1).to(device=y_pred.device)
+
+        self._kernel = self._kernel.expand(channel, 1, -1, -1)
+        kernel = self._kernel
+
+        if y_pred.device != self._kernel.device:
+            warnings.warn(
+                "The metric's device should be the same than your update tensors. See SSIM() device argument.",
+                RuntimeWarning,
+            )
```
vfdev-5, line 170: I have an impression that we can skip the warning.
I imagine the following situation, user create SSIM with device=cpu and does NN computations on cuda, so that y and y_pred would be on cuda. When eval part would be executed he will see this warning but it is legit to have the setup they have.

ignite/metrics/ssim.py
```diff
@@ -158,17 +159,31 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
...
+                y_pred = y_pred.to(device=self._kernel.device)
+                y = y.to(device=self._kernel.device)
 
         y_pred = F.pad(y_pred, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
         y = F.pad(y, [self.pad_w, self.pad_w, self.pad_h, self.pad_h], mode="reflect")
 
-        if y_pred.dtype != self._kernel.dtype:
-            self._kernel = self._kernel.to(dtype=y_pred.dtype)
+        if y_pred.dtype != kernel.dtype:
+            kernel = kernel.to(dtype=y_pred.dtype)
```
vfdev-5, line 183: I was also thinking about dtype and was wondering if we could get predicted image as uint8 dtype ?

ignite/metrics/ssim.py
MarcBresson: What would be the benefits of having the metric on CPU now that if y_pred is on GPU and kernel on CPU, kernel will be pushed to the GPU to perform the computation?

I could rephrase the warning to underline the lower performance when the two tensors are on different devices.

ignite/metrics/ssim.py
MarcBresson: that is absolutely right, scikit-image supports it.

ignite/metrics/ssim.py
vfdev-5: device arg for the metric in general is the device on which we store intermediate results and it is not computation device. In SSIM it happens that kernel is stored on metric device as well. But for conv2d computations it would be better to have it on the faster device (or cpu for cpu-only hardware).
Right now I think that actually, previous "weird" code was doing right thing about converting `self._kernel` to `y_pred` device once vs right now we do that on each iteration. What do you think ?
In general we also would like that user do not bother about the best config and by default it should be the best.

ignite/metrics/ssim.py
MarcBresson: https://github.com/scikit-image/scikit-image/blob/v0.21.0/skimage/metrics/_structural_similarity.py#L219 images are converted to float from what I can see.

So maybe just putting
```python
if y_pred.dtype == torch.int8:
    y_pred = y_pred / 255

# float to float conversion
if y_pred.dtype != kernel.dtype:
            y_pred = y_pred.to(dtype=kernel.dtype)
```

vfdev-5: Yes, this makes sense from skimage side. Also it would be rare to get uint8 output from an NN.
Also conv2d probably wont work on uint8 input neither...

ignite/metrics/ssim.py
MarcBresson: That was my commit comment. But now, because we are trying to do the "right thing" by putting all the tensors on GPU, there is the edge case where (y_pred1, y1) are on GPU but (y_pred2, y2) are on CPU.
If we move self._kernel (which is typically way lighter than y_pred) on GPU, then the second iteration will move y_pred on GPU (which will be longer, but maybe not as long as computing the SSIM metric entirely on CPU).

MarcBresson: I guess now we just have to define what's the best config !

vfdev-5: We can add also `.to(device, non_blocking=True)`, to hide a bit the device sync.

> here is the edge case where (y_pred1, y1) are on GPU but (y_pred2, y2) are on CPU.

I think this is very edge case and IMO it is ok to put y_pred and y from cpu to cuda in this case.


> I guess now we just have to define what's the best config !

I would say let's compute all on cuda if available even if metric device is cpu

vfdev-5: > device("cuda") Vs device("cuda:0").

this can be done as 
```
x = torch.tensor(0.0, device="cuda:0")
assert x.device.type == "cuda"
```

ignite/metrics/ssim.py
```diff
@@ -158,8 +159,21 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
             )
 
         channel = y_pred.size(1)
-        if len(self._kernel.shape) < 4:
-            self._kernel = self._kernel.expand(channel, 1, -1, -1).to(device=y_pred.device)
+
+        self._kernel = self._kernel.expand(channel, 1, -1, -1)
```
vfdev-5, line 163: I have some doubts about expanding all the times. I'm not sure what would be the best way of doing that especially taking into account that `channel` can potentially vary across the dataset.

MarcBresson: My last commit passed the tests and solve this issue

ignite/metrics/ssim.py
```diff
@@ -157,18 +159,34 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
                 f"Expected y_pred and y to have BxCxHxW shape. Got y_pred: {y_pred.shape} and y: {y.shape}."
             )
 
-        channel = y_pred.size(1)
-        if len(self._kernel.shape) < 4:
-            self._kernel = self._kernel.expand(channel, 1, -1, -1).to(device=y_pred.device)
+        if y_pred.device != self._kernel.device:
```
vfdev-5, line 162: Thanks for trying this approach with _kernel and _expanded_kernel !
Few remarks, with now 2 kernel attributes, we can easily get them unsynced. We would like to avoid the situations, so let's code this part as following:
- we should we define _expanded_kernel in the very beginning
```
        nb_channels = y_pred.size(1)
        if self._expanded_kernel is None or self._expanded_kernel.shape[0] != nb_channels:
            self._expanded_kernel = self._kernel.expand(nb_channels, 1, -1, -1)
```

- all device and dtype management should happen on _expanded_kernel only
- we should not use _kernel anywhere except when we define  _expanded_kernel

Nit remarks: I would rename  _kernel -> _kernel_2d and _expanded_kernel -> _kernel

ignite/metrics/ssim.py
```diff
@@ -157,18 +159,34 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
...
+            if self._kernel.device == torch.device("cpu"):
+                self._kernel = self._kernel.to(device=y_pred.device)
+                self._sum_of_ssim = self._sum_of_ssim.to(device=y_pred.device)
```
vfdev-5, line 165: I do not think that we should move _sum_of_ssim to another device. It should always stay on self._device

MarcBresson: What would be the benefit of keeping the sum on self._device ? That would mean moving the current value on the right device on each update call, just to spare 64 bits on GPU.

"Because the user wants it" is a valid answer :)

vfdev-5: > What would be the benefit of keeping the sum on self._device ?

It is a feature we have for all metrics that user can choose where to store internal intermediate parameters.
So, yes, user picks it.

MarcBresson:
I am writing new tests for the variable channel size, will push soon

MarcBresson:
oh no conflicts, what have I done ?

ignite/metrics/ssim.py
```diff
@@ -102,12 +103,13 @@ def __init__(
...
         self.pad_h = (self.kernel_size[0] - 1) // 2
         self.pad_w = (self.kernel_size[1] - 1) // 2
-        self._kernel = self._gaussian_or_uniform_kernel(kernel_size=self.kernel_size, sigma=self.sigma)
+        self._kernel_2d = self._gaussian_or_uniform_kernel(kernel_size=self.kernel_size, sigma=self.sigma)
 
     @reinit__is_reduced
     def reset(self) -> None:
         self._sum_of_ssim = torch.tensor(0.0, dtype=torch.float64, device=self._device)
         self._num_examples = 0
+        self._kernel = torch.empty(0)
```
vfdev-5, line 112: I liked the previous code with `self._kernel = None` in the ctor.

MarcBresson: Me too, but mypy didn't and was throwing errors :(

ignite/metrics/ssim.py
```diff
@@ -156,8 +158,25 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
...
+            if self._kernel.device == torch.device("cpu"):
+                self._kernel = self._kernel.to(device=y_pred.device)
+
+            if y_pred.device == torch.device("cpu"):
+                warnings.warn(
+                    "The metric device or one of the previous update tensor was set on another device than this "
+                    "update tensor, which is on CPU. To avoid having a performance hit, ensure that your metric "
+                    "device and all of your update tensors are on the same device.",
+                    RuntimeWarning,
+                )
```
vfdev-5, line 177: I think we can do here with `elif` and let's improve the warning message
```suggestion
            elif y_pred.device == torch.device("cpu"):
                warnings.warn(
                    f"y_pred tensor is on cpu device but previous computation was on another device: {self._kernel.device}. "
                    "To avoid having a performance hit, please ensure that all y and y_pred tensors "
                    "are on the same device.",
                )
```
Let's use UserWarning type instead of RuntimeWarning.

vfdev-5 (approved):
LGTM, thanks @MarcBresson !

PR #3044, opened 2023-08-24
refactor: remove redundant line as .reset() is call in Metrics().__init()

MarcBresson:
The same line was in reset() and init() which is unecessary.

Check list:


vfdev-5 (approved):
LGTM, if mypy check wont complain

MarcBresson:
Do we even know why the kernel is created in the reset method ?

I think I removed it on the wrong line. It should only be called in init().

vfdev-5:
Yes, you are right, we do not need to recreate the kernel on reset.

vfdev-5 (approved):
LGTM, thanks for the fix @MarcBresson !

PR #3045, opened 2023-08-24
Added compatibility with uint8 to SSIM metric

MarcBresson:
Add compatibility and tests for when y_pred and y are in uint8 dtype

Check list:


ignite/metrics/ssim.py
```diff
@@ -157,6 +159,15 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
                 f"Expected y_pred and y to have BxCxHxW shape. Got y_pred: {y_pred.shape} and y: {y.shape}."
             )
 
+        if y_pred.dtype == torch.uint8 or y.dtype == torch.uint8:
+            if self.data_range != 255:
+                warnings.warn(
+                    "dtypes of the input tensors are torch.uint8 but data range is not set to 255.", RuntimeWarning
+                )
```
vfdev-5, line 166: I'm not a big fan of showing a warning in this case. I would follow the same strategy as skimage where they just transform to floating point. So, I would do the following:
```
if not y.is_floating_point():
    y = y.float()
if not y_pred.is_floating_point():
    y_pred = y_pred.float()
```

vfdev-5 (commented):
Thanks for working on this PR @MarcBresson !
I left a comment on how I would tackle this dtype problem.
I also label this PR with "help wanted" label to create a thread on our discord, where we could discuss about this PR in a more fluent way

ignite/metrics/ssim.py
MarcBresson: Maybe we should put a default value to the data_range argument too? Most of the time we are dealing with [0; 1] images.

vfdev-5: I do not have a strong opinion on this. We have to figure out if there wont be any negative impact for the users in this case.

MarcBresson: I think the skiimg approach is quite good: default to data_range = 1.0, and warn if the detected type is of integer and that data range is still set to 1

vfdev-5:
@MarcBresson can you please resolve the conflict and let's apply what we discussed up to here: https://github.com/pytorch/ignite/pull/3045#discussion_r1305430207
Concerning default to data_range = 1.0, let's keep the public API the same for now.

vfdev-5 (approved):
Thanks @MarcBresson , LGTM!

MarcBresson:
Sorry for the delay btw! Got caught up in a job seeking activity ahah

vfdev-5:
No worries, hope you can get something exciting as a new job role!

PR #3009, opened 2023-07-27
Discord Workflow for Issues and pull-requests(non-maintainers-only)

guptaaryan16:
Fixes: Discuss in Discord for PRs and issues from non-maintainers

Description:

Check list:

- [ ] New tests are added (if a new feature is added)
- [ ] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #3046, opened 2023-08-28
Fix Discord PR extension workflow output

guptaaryan16:
Fixes #3014

Description:
The PR is to fix the wrong output of the Discord PR workflow, which now requires a pull request number to be a triggered. Also the issue workflow can't be triggered now when PR is labelled. This will fix the output of the comments in this extension. This PR also serves as a fix for the #3009.

Check list:

### DIFF
--- a/.github/workflows/discord_pull_requests.yaml
+++ b/.github/workflows/discord_pull_requests.yaml
@@ -25,5 +25,5 @@ jobs:
           destination: ${{ secrets.DISCORD_BOT_DESTINATION }}
           issue_number: ${{ github.event.inputs.pull_request_number || github.event.pull_request.number }}
           issue_comment: Hey 👋, I've just created a [thread]($THREAD_LINK$) for this pull request on [PyTorch-Ignite Discord](https://pytorch-ignite.ai/chat) where you can quickly talk to the community on the topic.
-          discord_message: New PR created in `${{ github.repository }}`:<https://github.com/${{ github.repository }}/pull/${{ github.event.pull_request.number || github.event.pull_request.number }}>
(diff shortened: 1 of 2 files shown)


.github/workflows/discord_pull_requests.yaml
```diff
@@ -24,5 +25,5 @@ jobs:
           destination: ${{ secrets.DISCORD_BOT_DESTINATION }}
           issue_number: ${{ github.event.inputs.pull_request_number || github.event.pull_request.number }}
           issue_comment: Hey 👋, I've just created a [thread]($THREAD_LINK$) for this pull request on [PyTorch-Ignite Discord](https://pytorch-ignite.ai/chat) where you can quickly talk to the community on the topic.
-          discord_message: New PR created in `${{ github.repository }}`:<https://github.com/${{ github.repository }}/pull/${{ github.event.pull_request.number }}>
+          discord_message: New PR created in `${{ github.repository }}`:<https://github.com/${{ github.repository }}/pull/${{ github.event.pull_request.number || github.event.pull_request.number }}>
```
vfdev-5, line 28: `github.event.pull_request.number || github.event.pull_request.number` ??

guptaaryan16: @vfdev-5 Since both values mostly give a same output it is better to have this structure as I am a bit unsure about the Github API requests in the actions. It is similar to fetching of the `issue_number: ${{ github.event.inputs.pull_request_number || github.event.pull_request.number }}` in line 26

vfdev-5: Correct me if I'm wrong but I see the same variable twice, like `${{ x || x}}`. If this is expected, I do not understand why can't you do `${{ x }}` ?

PR #3059, opened 2023-09-06
Fix attribute error in `Checkpoint::reload_objects` when `save_handler` is not of type `DiskSaver`

### LINKED ISSUE #2998: AttributeError in `Checkpoint::reload_objects` when `save_handler` is a callable object
## 🐛 Bug description

When using `Checkpoint` with a callable object as `save_handler`, calling `reload_objects` causes `AttributeError`. To reproduce:

```python
def test_checkpoint_simple_recovery():
    from pickle import dump
    class SaveHandler:
        def __call__(self, checkpoint, filename):
            with open(filename, 'wb') as f:
                dump(checkpoint, f)
    to_save = {'model': DummyModel()}
    h = Checkpoint(to_save, SaveHandler())
    engine = Engine(lambda e, b: None)
    engine.state = State(epoch=0, iteration=1)
    h(engine)
    to_load = {"model": DummyModel()}
    h.reload_objects(to_load, global_step=1)
```

result:
```
>       path = self.save_handler.dirname / checkpoint_fp
E       AttributeError: 'SaveHandler' object has no attribute 'dirname'
```

It attempts to access `dirname` of the handler while `dirname` is only the attribute of the `DiskSaver` not even `BaseSaveHandler`.

Also in the `Checkpoint` docstring, it states `String, method or callable class` (by callable class it means callable object?) for  `save_handler` type but in type checks we allow a callable function:

```python
if not (
    isinstance(save_handler, str)
    or isinstance(save_handler, Path)
    or callable(save_handler)
    or isinstance(save_handler, BaseSaveHandler)
):
    raise TypeError(
        "Argument `save_handler` should be a string or Path object or callable or inherit from BaseSaveHandler"
    )
```

<details>
<summary>Checklist</summary>

- [ ] `ignite/handlers/checkpoint.py`
> • In the `reload_objects` method of the `Checkpoint` class, before the line `path = self.save_handler.dirname / checkpoint_fp`, add a condition to check if `self.save_handler` is an instance of `DiskSaver`.
> • If `self.save_handler` is an instance of `DiskSaver`, keep the existing logic to get the `path`.
> • If `self.save_handler` is not an instance of `DiskSaver`, handle the checkpoint reloading for the callable object. This might involve calling the `save_handler` with the appropriate arguments to reload the checkpoint.

</details>

sadra-barikbin:
Fixes #2998

### DIFF
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -670,7 +670,8 @@ class Checkpoint(Serializable):
             `DataParallel`_, method ``load_state_dict`` will applied to their internal wrapped model (``obj.module``).
 
         Note:
-            This method works only when the ``save_handler`` is of type :class:`~ignite.handlers.checkpoint.DiskSaver`.
+            This method works only when the ``save_handler`` is of types string,
+            :class:`~pathlib.Path` or :class:`~ignite.handlers.checkpoint.DiskSaver`.
 
         .. _DistributedDataParallel: https://pytorch.org/docs/stable/generated/
             torch.nn.parallel.DistributedDataParallel.html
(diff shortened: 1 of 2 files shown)


vfdev-5 (approved):

PR #3069, opened 2023-09-23
Use class to organize distributed tests in modules

sadra-barikbin:
As already suggested by @vfdev-5

Description:

The gola of the PR is to reduce the number of distributed context setup/destroy calls.

We can either use `distributed` fixture per method or write a class like in the PR

For simple test metthods with `distributed` fixture, they are instantiated once per method then. It falls back to scope=function for those test items.


vfdev-5 (approved):
Thanks @sadra-barikbin , LGTM

PR #3132, opened 2023-11-11
Added The BInary Expected_Calibration_Error (ECE) Metric 

### LINKED ISSUE #1009: Add Expected Calibration Error metrics and derived
## 🚀 Feature

Idea is to provide Expected Calibration Error metrics as coded in Baal : https://github.com/ElementAI/baal/blob/master/src/baal/utils/metrics.py#L88

cc @Dref360

Zekrom-7780:
Fixes #1009 

## Description:
Added the binary `ECE(Expected_Calibration_Error)` metric.

## Reviewers 

cc. @vfdev-5 

P.S. 
- Do I have to add new tests, and documentation for the new metric? I don't know whether I had to add them or not.
- Will be doing the `MultiClassECE` in a new pull request, since it feels that it is the better thing to do.


ignite/contrib/metrics/ExpectedCalibrationError.py
```diff
@@ -0,0 +1,57 @@
+import torch
```
vfdev-5, line 1: First, let's put it into `ignite/metrics/ExpectedCalibrationError.py` instead of `ignite/contrib/metrics/ExpectedCalibrationError.py`

ignite/contrib/metrics/ExpectedCalibrationError.py
```diff
@@ -0,0 +1,57 @@
...
+        self.reset()
+
+    def reset(self):
+        self.confidences = torch.tensor([], device=self.device)
+        self.corrects = torch.tensor([], device=self.device)
+
+    def update(self, output):
+        y_pred, y = output
+
+        assert y_pred.dim() == 2 and y_pred.shape[1] == 2, "This metric is for binary classification."
```
vfdev-5, line 21: Let's use the following way to raise errors instead of `assert`:
```python
if not (y_pred.dim() == 2 and y_pred.shape[1] == 2):
    raise ValueError("This metric is for binary classification")
```

To assert if the input is binary we were doing previously something like here:
https://github.com/pytorch/ignite/blob/4dc4e04b901c682e3e6f0b193a18c50c56eb3c41/ignite/metrics/accuracy.py#L51-L67

ignite/contrib/metrics/ExpectedCalibrationError.py
```diff
@@ -0,0 +1,57 @@
...
+        self.num_bins = num_bins
+        self.device = device
+        self.reset()
+
+    def reset(self):
+        self.confidences = torch.tensor([], device=self.device)
+        self.corrects = torch.tensor([], device=self.device)
+
+    def update(self, output):
+        y_pred, y = output
```
vfdev-5, line 19: We usually call `.detach` on both to stop grad computation like here: https://github.com/pytorch/ignite/blob/4dc4e04b901c682e3e6f0b193a18c50c56eb3c41/ignite/metrics/accuracy.py#L231

vfdev-5 (commented):
Thanks for the PR @Zekrom-7780 !

I left few comments but I haven't yet fully checked the code and correctness.

> Do I have to add new tests, and documentation for the new metric? I don't know whether I had to add them or not.

Yes, we need to add new tests and provide a docstring and include the metric to the docs (i'll guide how to do that once we are at this point, nothing complicated).

Main question about tests will be what would be the reference for our implementation. For example, in most of the cases, we are using sklearn to compute expected value.

> Will be doing the MultiClassECE in a new pull request, since it feels that it is the better thing to do.

Won't be possible to have a single class for binary and multi-class ECE ?

Zekrom-7780:
@vfdev-5 made the changes, 

## Ready for comments

> Won't be possible to have a single class for binary and multi-class ECE ?

Will do it in the next commits if you want, actually I just wanted to understand how does the `binaryclassECE` work, and by completing that, the other one would be a piece of cake to do, hence just focusing on this one rn

vfdev-5:
@Zekrom-7780 any updates on this PR ?

Zekrom-7780:
@vfdev-5 I'll create the `Multi-Class-ECE` in the next 3-4 hours, and push the changes

Then i'll move to writing the unit-tests, where I'll need your help, since I haven't written any as such

Zekrom-7780:
@vfdev-5 , made the changes

PR #3186, opened 2024-01-11
Change LinearCyclicalScheduler to triangle wave to sawtooth wave

### LINKED ISSUE #3185: Changing LinearCyclicalScheduler triangle wave to sawtooth wave
## 🚀 Feature
I applied feature about warmup each cycle on CosineAnnealingScheduler -> #3064
and We had additional discussion to change LinearCyclicalScheduler triangle wave to sawtooth wave
you can see discussion in [here](https://discord.com/channels/831462531327328276/1143465439553392680/1154900705094684702)

sihyeong671:
Fixes #3185 

Description:

Add monotonic variable in LinearCyclicalScheduler
LinearCyclicalScheduler's get_param method is modified

Add Test case LinearCyclicalScheduler ValueError
Add Test case LinearCyclicalScheduler with warmup duration

Check list:

### DIFF
--- a/ignite/handlers/param_scheduler.py
+++ b/ignite/handlers/param_scheduler.py
@@ -348,11 +348,8 @@ class CyclicalScheduler(ParamScheduler):
             self.start_value *= self.start_value_mult
         if self.event_index != 0 and self.event_index == self.total_cycle_size:
             self.event_index = 0
-            # self.cycle_size = math.ceil(self.cycle_size * self.cycle_mult)
-            # self.warmup_duration = math.ceil(self.warmup_duration * self.cycle_mult)
             self.cycle_size = int(self.cycle_size * self.cycle_mult)
             self.warmup_duration = int(self.warmup_duration * self.cycle_mult)
-
             self.total_cycle_size = self.warmup_duration + self.cycle_size
             self.cycle += 1
             self.end_value *= self.end_value_mult
@@ -468,6 +465,9 @@ class LinearCyclicalScheduler(CyclicalScheduler):
 
     .. versionchanged:: 0.4.13
         Added cyclic warm-up to the scheduler using ``warmup_duration``.
+
+    .. versionchanged:: 0.5.0
+        Added monotonic argument ....
     """
 
     def __init__(self, *args, monotonic=False, **kwagrs):
--- a/ignite/handlers/param_scheduler.py
+++ b/ignite/handlers/param_scheduler.py
@@ -475,7 +475,8 @@ class LinearCyclicalScheduler(CyclicalScheduler):
         self.monotonic = monotonic
         if self.warmup_duration > 0 and not self.monotonic:
             raise ValueError(
-                "Invalid combination when warmup_duration > 0 and monotonic=False, please use either set warmup_duration=0 or monotonic=True"  # noqa E501
+                "Invalid combination when warmup_duration > 0 and monotonic=False, "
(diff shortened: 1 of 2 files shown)


vfdev-5:
Thanks for the PR and opening a corresponding feature request, @sihyeong671 !
The only problem I see here is BC-breaking change, so let's implement this feature as an optional flag with default value keeping current behavior, such that ignite users won't have their code broken after this PR get landed. What do you think?

sihyeong671:
> Thanks for the PR and opening a corresponding feature request, @sihyeong671 ! The only problem I see here is BC-breaking change, so let's implement this feature as an optional flag with default value keeping current behavior, such that ignite users won't have their code broken after this PR get landed. What do you think?

got it! I think that is the good idea.I will try add optional flag

ignite/handlers/param_scheduler.py
```diff
@@ -348,8 +348,11 @@ def __call__(self, engine: Optional[Engine], name: Optional[str] = None) -> None
             self.start_value *= self.start_value_mult
         if self.event_index != 0 and self.event_index == self.total_cycle_size:
             self.event_index = 0
+            # self.cycle_size = math.ceil(self.cycle_size * self.cycle_mult)
+            # self.warmup_duration = math.ceil(self.warmup_duration * self.cycle_mult)
             self.cycle_size = int(self.cycle_size * self.cycle_mult)
             self.warmup_duration = int(self.warmup_duration * self.cycle_mult)
+
```
vfdev-5, line 355: Let's remove commented lines and new empty line here

ignite/handlers/param_scheduler.py
```diff
@@ -467,10 +470,23 @@ def print_lr():
         Added cyclic warm-up to the scheduler using ``warmup_duration``.
     """
 
+    def __init__(self, *args, monotonic=False, **kwagrs):
```
vfdev-5, line 473: As we are adding new arg, please update the docstring and also please add
```
.. versionchanged:: 0.5.0
        Added monotonic argument ....
```

ignite/handlers/param_scheduler.py
```diff
@@ -465,12 +465,28 @@ def print_lr():
 
     .. versionchanged:: 0.4.13
         Added cyclic warm-up to the scheduler using ``warmup_duration``.
+
+    .. versionchanged:: 0.5.0
+        Added monotonic argument ....
```
vfdev-5, line 470: Here instead of `....` we need to write a proper sentence

sihyeong671: Actually, I was curious about meaning of `....`. I am gonna add proper docstring :)

vfdev-5 (approved):
Looks good to me, thanks a lot for the contribution, @sihyeong671 !

ignite/handlers/param_scheduler.py
```diff
@@ -465,12 +468,28 @@ def print_lr():
 
     .. versionchanged:: 0.4.13
         Added cyclic warm-up to the scheduler using ``warmup_duration``.
+
+    .. versionchanged:: 0.5.0
+        Added monotonic argument.
     """
 
+    def __init__(self, *args, monotonic=False, **kwagrs):
```
vfdev-5, line 476:
```suggestion
    def __init__(self, *args, monotonic: bool = False, **kwagrs):
```

PR #3199, opened 2024-02-14
docs: add MAPE description to MARE [skip ci]

### LINKED ISSUE #3198: Mean Absolute Percentage Error (MAPE)
## 🚀 Feature

I'd like to implement the mean absolute percentage error [(MAPE)](https://en.wikipedia.org/wiki/Mean_absolute_percentage_error) in `ignite/metrics`.

It is a commonly used metric for regression problems and it would be really convenient to be able to use it directly with ignite evaluators.

For that, I would write a custom Metric class in a new file `mean_absolute_percentage_error.py` inheriting from the base `Metric` class in `ignite/metrics/metric.py`.

aalbayrak394:
Fixes #3198

Description: Small adjustment to the Docstring of `ignite/contrib/metrics/regression/mean_absolute_relative_error.py`

Check list:

- [x] Added alias "MAPE" to docstring

### DIFF
--- a/ignite/contrib/metrics/regression/mean_absolute_relative_error.py
+++ b/ignite/contrib/metrics/regression/mean_absolute_relative_error.py
@@ -8,7 +8,7 @@ from ignite.metrics.metric import reinit__is_reduced, sync_all_reduce
 
 
 class MeanAbsoluteRelativeError(_BaseRegression):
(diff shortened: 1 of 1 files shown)


vfdev-5 (approved):
Thanks @aalbayrak394 , LGTM!

PR #3210, opened 2024-03-20
Add Entropy metric

kzkadc:
Description: adds Entropy metric.
Entropy is often used when evaluating uncertainty of classification predictions.
It is computed as $H=\sum_c - p_c \log p_c$, where $p_c$ is the predicted probability for $c$-th class.

Check list:


ignite/metrics/entropy.py
```diff
@@ -0,0 +1,82 @@
...
+class Entropy(Metric):
+    r"""Calculates the mean of `entropy <https://en.wikipedia.org/wiki/Entropy_(information_theory)>`_.
+
+    .. math:: H = \frac{1}{N} \sum_{i=1}^N \sum_{c=1}^C -p_{i,c} \log p_{i,c},
+       \quad p_{i,c} = \frac{\exp(z_{i,c})}{\sum_{c'=1}^C \exp(z_{i,c'})}
+
+    where :math:`p_{i,c}` is the prediction probability of :math:`i`-th data belonging to the class :math:`c`.
+
+    - ``update`` must receive output of the form ``(y_pred, y)`` while ``y`` is not used in this metric.
+    - ``y_pred`` is expected to be the unnormalized logits for each class.
```
vfdev-5, line 21: Do we have any limitations on the shape of the predictions ? For example, if y_pred is [B, C, H, W] as for image segmentation task, can we still compute the value ?

ignite/metrics/entropy.py
```diff
@@ -0,0 +1,82 @@
...
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._sum_of_entropies = torch.tensor(0.0, device=self._device)
+        self._num_examples = 0
+
+    @reinit__is_reduced
+    def update(self, output: Sequence[torch.Tensor]) -> None:
+        y_pred = output[0].detach()
+        prob = F.softmax(y_pred, dim=1)
+        log_prob = F.log_softmax(y_pred, dim=1)
```
vfdev-5, line 73: I think as `y_pred` is already a logit we may not need to reapply log_softmax ?

kzkadc: Thank you for your review.
log_softmax is necessary as $\log p_c = \log \text{softmax}(y_\text{pred}) =  y_{\text{pred},c} - \log \sum_i \exp (y_{\text{pred},i}) \neq y_{\text{pred},c}$

> Maybe, we could also reuse torch built-in cross entropy function and provide softmaxed y_pred as target?

Could you please explain in more detail?

ignite/metrics/entropy.py
kzkadc: It works with only [B, C] shape now, but supporting segmentation would be nice!
I'd like to modify the code to support [B, C, H, W].

ignite/metrics/entropy.py
vfdev-5: Yes, you are right there is additional log sum term.

> Could you please explain in more detail?

I was thinking that cross entropy mplementation could fit what we want to compute here. In details, actually, it is not exactly usable for our purpose as cross-entropy takes log_p at places when target == 1.

https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html#torch.nn.CrossEntropyLoss

ignite/metrics/entropy.py
kzkadc: I modified the code to support [B, C, ...] shape.
For instance, [B, C, H, W] tensors are reshaped into [B\*H\*W, C] and then entropy is computed over B\*H\*W predictions.

tests/ignite/metrics/test_entropy.py
```diff
@@ -0,0 +1,206 @@
...
+from ignite.exceptions import NotComputableError
+from ignite.metrics import Entropy
+
+
+def np_entropy(np_y_pred: np.ndarray):
+    np_y_pred = np_y_pred - np_y_pred.max(axis=1, keepdims=True)
+    prob = np.exp(np_y_pred) / np.sum(np.exp(np_y_pred), axis=1, keepdims=True)
+    log_prob = np_y_pred - np.log(np.sum(np.exp(np_y_pred), axis=1, keepdims=True))
+    np_ent = -np.sum(prob * log_prob) / np_y_pred.shape[0]
+    return np_ent
```
vfdev-5, line 17: I wonder whether we could use scipy entropy as ref code ?
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.entropy.html
We have scipy for dev dependencies: https://github.com/pytorch/ignite/blob/5fe74432ed5813d803b5c40641cd47d0cb9964df/requirements-dev.txt#L9
so if it is possible to reuse scipy I would prefer to use it instead of manually computing it here.
what do you think @kzkadc ?

tests/ignite/metrics/test_entropy.py
```diff
@@ -0,0 +1,206 @@
...
+    ent = Entropy()
+
+    y_pred = torch.tensor([[2.0, 3.0], [-2.0, -1.0]], requires_grad=True)
+    y = torch.zeros(2)
+    ent.update((y_pred, y))
+
+    assert not ent._sum_of_entropies.requires_grad
+
+
+@pytest.mark.distributed
```
vfdev-5, line 140: For distributed config tests, could you please rewrite them using new testing formalism that we are trying to adopt. Here is an example of the code to inspire of:
https://github.com/pytorch/ignite/blob/5fe74432ed5813d803b5c40641cd47d0cb9964df/tests/ignite/metrics/test_recall.py#L422-L426

Here is a PR showing how to pass from old code to the new one:
https://github.com/pytorch/ignite/pull/3208/files#diff-c56c264ef288f88e5738e9ad22de66dffd4c58d2e656eb62e8dbaa678672317d

Thanks!

ignite/metrics/entropy.py
```diff
@@ -0,0 +1,90 @@
...
+class Entropy(Metric):
+    r"""Calculates the mean of `entropy <https://en.wikipedia.org/wiki/Entropy_(information_theory)>`_.
+
+    .. math:: H = \frac{1}{N} \sum_{i=1}^N \sum_{c=1}^C -p_{i,c} \log p_{i,c},
+       \quad p_{i,c} = \frac{\exp(z_{i,c})}{\sum_{c'=1}^C \exp(z_{i,c'})}
+
+    where :math:`p_{i,c}` is the prediction probability of :math:`i`-th data belonging to the class :math:`c`.
+
+    - ``update`` must receive output of the form ``(y_pred, y)`` while ``y`` is not used in this metric.
+    - ``y_pred`` is expected to be the unnormalized logits for each class. :math:`(B, C)` (classification) or :math:`(B, C, ...)` (e.g., image segmentation) shapes are allowed.
```
vfdev-5, line 21: Code formatting tests are not passing: https://github.com/pytorch/ignite/actions/runs/8375820095/job/22974497359?pr=3210#step:11:89 due to this long line. 

Maybe we can split it like this:
```suggestion
    - ``y_pred`` is expected to be the unnormalized logits for each class. :math:`(B, C)` (classification) 
      or :math:`(B, C, ...)` (e.g., image segmentation) shapes are allowed.
```
two spaces on the next line after above `-`.

kzkadc: Thank you for your correction!

tests/ignite/metrics/test_entropy.py
kzkadc: Thanks for your suggestion!
Indeed, scipy has softmax and entropy.
I updated `np_entropy` to use them.

ignite/metrics/entropy.py
```diff
@@ -0,0 +1,91 @@
...
+class Entropy(Metric):
+    r"""Calculates the mean of `entropy <https://en.wikipedia.org/wiki/Entropy_(information_theory)>`_.
+
+    .. math:: H = \frac{1}{N} \sum_{i=1}^N \sum_{c=1}^C -p_{i,c} \log p_{i,c},
+       \quad p_{i,c} = \frac{\exp(z_{i,c})}{\sum_{c'=1}^C \exp(z_{i,c'})}
+
+    where :math:`p_{i,c}` is the prediction probability of :math:`i`-th data belonging to the class :math:`c`.
+
+    - ``update`` must receive output of the form ``(y_pred, y)`` while ``y`` is not used in this metric.
+    - ``y_pred`` is expected to be the unnormalized logits for each class. :math:`(B, C)` (classification)
```
vfdev-5, line 21:
```suggestion
    - ``y_pred`` is expected to be the unnormalized logits for each class. :math:`(B, C)` (classification)
```

vfdev-5:
@kzkadc can you please run `bash tests/run_code_style.sh fmt` to fix all code formatting issues, thanks!
https://github.com/pytorch/ignite/actions/runs/8390870373/job/22979934843?pr=3210

vfdev-5:
@kzkadc this failure is related: https://github.com/pytorch/ignite/actions/runs/8391979913/job/22983553399?pr=3210#step:9:345

```
         y_preds = idist.all_gather(y_preds)
        y_true = idist.all_gather(y_true)
    
        assert "entropy" in engine.state.metrics
        res = engine.state.metrics["entropy"]
    
>       true_res = np_entropy(y_preds.numpy())
E       TypeError: can't convert xla:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.

tests/ignite/metrics/test_entropy.py:102: TypeError
```

PR #3203, opened 2024-03-14
Add cosine similarity metric

kzkadc:
Fixes #3202

Description:

Check list:

- [x] New tests are added (if a new feature is added)
- [x] New doc strings: description and/or example code are in RST format
- [x] Documentation is updated (if required)

PR #3213, opened 2024-03-20
Fix error when testing xla in cosine similarity metric

kzkadc:
Fixes https://github.com/pytorch/ignite/pull/3203#issuecomment-2009094503

Description:

Check list:


vfdev-5 (approved):
Thanks @kzkadc , lgtm

PR #3214, opened 2024-03-22
Update references to handlers

leej3:
Follows on from #3204 by adjusting imports in examples etc. to match the migrated content.


vfdev-5 (approved):
Thanks, John!

PR #3215, opened 2024-03-22
add fbresearch_logger.py

leej3:
Description:
Add FBResearchLogger class from unmerged branch object-detection-example

Check list:

### DIFF
--- a/ignite/handlers/fbresearch_logger.py
+++ b/ignite/handlers/fbresearch_logger.py
@@ -152,8 +152,8 @@ class FBResearchLogger:
         epoch_time = engine.state.times[Events.EPOCH_COMPLETED.name]
         epoch_info = (
             f"Epoch [{engine.state.epoch}/{engine.state.max_epochs}]"
-            if engine.state.max_epochs > 1
-            else ""  # type: ignore
+            if engine.state.max_epochs > 1  # type: ignore
+            else ""
         )
         msg = self.delimiter.join(
             [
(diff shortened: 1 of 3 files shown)


ignite/handlers/fbresearch_logger.py
```diff
@@ -0,0 +1,185 @@
...
+    """Logs training and validation metrics for research purposes.
+
+    This logger is designed to attach to an Ignite Engine and log various metrics
+    and system stats at configurable intervals, including learning rates, iteration
+    times, and GPU memory usage.
+
+    Args:
+        logger (logging.Logger): The logger to use for output.
+        delimiter (str): The delimiter to use between metrics in the log output.
+        show_output (bool): Flag to enable logging of the output from the engine's process function.
```
vfdev-5, line 27: We are using some sphinx plug-in such that we do not need to provide types in the docstring:
```suggestion
        logger: The logger to use for output.
        delimiter: The delimiter to use between metrics in the log output.
        show_output: Flag to enable logging of the output from the engine's process function.
```

ignite/handlers/fbresearch_logger.py
```diff
@@ -0,0 +1,185 @@
...
+            step=Events.GET_BATCH_COMPLETED,
+        )
+
+    def log_every(self, engine: Engine, optimizer: Optional[torch.optim.Optimizer] = None) -> None:
+        """
+        Logs the training progress at regular intervals.
+
+        Args:
+            engine (Engine): The training engine.
+            optimizer (torch.optim.Optimizer): The optimizer used for training. Defaults to None.
```
vfdev-5, line 85: Let's remove types everywhere in the docstring

vfdev-5 (approved):
Thanks, John, looks good to me

PR #3224, opened 2024-03-31
Try to update actions/checkout and codecov/codecov-action to v4

vfdev-5:
We can't yet update due to the following errors on pytorch-infra:
```
Run actions/checkout@v4
/home/ec2-user/actions-runner/externals/node20/bin/node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /home/ec2-user/actions-runner/externals/node20/bin/node)
/home/ec2-user/actions-runner/externals/node20/bin/node: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /home/ec2-user/actions-runner/externals/node20/bin/node)
```
- https://github.com/pytorch/ignite/actions/runs/8500656796/job/23282754276

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -66,7 +66,7 @@ jobs:
           echo "::endgroup::"
 
       - name: Checkout repository (pytorch/test-infra)
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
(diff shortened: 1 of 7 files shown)


leej3:
There is a description of this error is [here](https://stackoverflow.com/a/74896209). Sounds like the solution is update the base image to use newer glibc etc. or downgrade the version of node.

PR #3228, opened 2024-04-02
fix misspecified version reference

leej3:
@vfdev-5, a quick fix to align release version with version in docs.


vfdev-5 (approved):
Thanks, John!

PR #3230, opened 2024-04-03
Add MutualInformation Metric

kzkadc:
Description: added MutualInformation metric, which is often used in representation learning.

It measures the mutual information between input and output for classification models.

Check list:


ignite/metrics/mutual_information.py
```diff
@@ -0,0 +1,110 @@
...
+            # regarding as B*... predictions
+            y_pred = y_pred.movedim(1, -1).reshape(-1, num_classes)
+        elif y_pred.ndim == 1:
+            raise ValueError(f"y_pred must be in the shape of (B, C) or (B, C, ...), got {y_pred.shape}.")
+
+        prob = F.softmax(y_pred, dim=1)
+        log_prob = F.log_softmax(y_pred, dim=1)
+        ent_sum = -(prob * log_prob).sum()
+
+        self._sum_of_probabilities = self._sum_of_probabilities + prob.sum(dim=0).to(self._device)
```
vfdev-5, line 95:
```suggestion
        self._sum_of_probabilities += prob.sum(dim=0).to(self._device)
```

tests/ignite/metrics/test_mutual_information.py
```diff
@@ -0,0 +1,147 @@
...
+from ignite.engine import Engine
+from ignite.exceptions import NotComputableError
+from ignite.metrics import MutualInformation
+
+
+def np_mutual_information(np_y_pred: np.ndarray) -> float:
+    prob = softmax(np_y_pred, axis=1)
+    marginal_ent = entropy(np.mean(prob, axis=0))
+    conditional_ent = np.mean(entropy(prob, axis=1))
+    return max(0.0, marginal_ent - conditional_ent)
```
vfdev-5, line 22: @kzkadc can we use scikit-learn mutual info metric as a reference ? 
https://scikit-learn.org/stable/modules/clustering.html#mutual-information-based-scores

vfdev-5:
@kzkadc thanks for the PR adding more metrics ! 
I left few comments to check

ignite/metrics/mutual_information.py
```diff
@@ -0,0 +1,110 @@
...
+
+class MutualInformation(Metric):
+    r"""Calculates the `mutual information <https://en.wikipedia.org/wiki/Mutual_information>`_
+    between input :math:`X` and prediction :math:`Y`.
+
+    .. math::
+
+       I(X;Y) = H(Y) - H(Y|X)
+       = H \left( \frac{1}{N}\sum_{i=1}^N \hat{\mathbf{p}}_i \right)
+       - \frac{1}{N}\sum_{i=1}^N H(\hat{\mathbf{p}}_i),
```
vfdev-5, line 20: Let's add new-line tag here. Otherwise it is rendered as a single line. See here: https://deploy-preview-3230--pytorch-ignite-preview.netlify.app/generated/ignite.metrics.mutualinformation#ignite.metrics.MutualInformation

ignite/metrics/mutual_information.py
kzkadc: Thank you for your suggestion!
But += produces an error here in the first call because the left term is tensor(0.0) while the right term prob.sum(dim=0) is a vector.

```python
import torch
a = torch.tensor(0.0)
a += torch.ones(10)
# RuntimeError: output with shape [] doesn't match the broadcast shape [10]
```

ignite/metrics/mutual_information.py
```diff
@@ -0,0 +1,109 @@
...
+            # regarding as B*... predictions
+            y_pred = y_pred.movedim(1, -1).reshape(-1, num_classes)
+        elif y_pred.ndim == 1:
+            raise ValueError(f"y_pred must be in the shape of (B, C) or (B, C, ...), got {y_pred.shape}.")
+
+        prob = F.softmax(y_pred, dim=1)
+        log_prob = F.log_softmax(y_pred, dim=1)
+        ent_sum = -(prob * log_prob).sum()
+
+        self._sum_of_probabilities = self._sum_of_probabilities + prob.sum(dim=0).to(self._device)
```
vfdev-5, line 94: How much this update function differs from Entropy.update and can't we inherit from Entropy to reduce redundant code?

ignite/metrics/mutual_information.py
```diff
@@ -0,0 +1,109 @@
...
+
+from ignite.exceptions import NotComputableError
+from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
+
+__all__ = ["MutualInformation"]
+
+
+class MutualInformation(Metric):
+    r"""Calculates the `mutual information <https://en.wikipedia.org/wiki/Mutual_information>`_
+    between input :math:`X` and prediction :math:`Y`.
```
vfdev-5, line 14: What is X here ? I'm a bit confused about the definition of MutualInformation here in the docstring, the implementation that just uses predictions and the fact that we can't match sklearn mutual info score: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mutual_info_score.html#sklearn.metrics.mutual_info_score

It would be better to provide metrics that match sklearn metrics instead of something else

You are saying in the description that : 
>  added MutualInformation metric, which is often used in representation learning.

can you please provide any links where it is used and the definition they have ? Thanks!

kzkadc: X is an input and model predictions correspond to the posterior distribution $P(Y|X)$, which includes information of X.
Thus, we can compute the mutual information between X and Y just using the predictions.

This mutual information is also sometimes called infomax loss.
For instance, it is used in the followings:
- https://proceedings.neurips.cc/paper/1991/hash/a8abb4bb284b5b27aa7cb790dc20f80b-Abstract.html (NeurIPS1991) Eqs. (7) - (9)
- https://papers.nips.cc/paper_files/paper/2010/hash/42998cf32d552343bc8e460416382dca-Abstract.html (NeurIPS2010) Eq. (1)
- https://arxiv.org/abs/1206.6438 (ICML2012) Eq. (5)
- https://proceedings.mlr.press/v70/hu17b.html (ICML2017) Eqs. (1) and (7)

vfdev-5: Thanks for the links @kzkadc !
I see now, docstring definitions for H(Y) and H(Y|X) are coming from (8), (9) of this paper: http://proceedings.mlr.press/v70/hu17b/hu17b.pdf "Learning Discrete Representations via Information Maximizing Self-Augmented Training".

Can you please put a ref to it into the docstring, otherwise I can't find a similar defs on the wikipedia page ?

ignite/metrics/entropy.py
```diff
@@ -80,9 +79,14 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
 
         prob = F.softmax(y_pred, dim=1)
         log_prob = F.log_softmax(y_pred, dim=1)
+
+        self._update(prob, log_prob)
+
+    @reinit__is_reduced
```
vfdev-5, line 85: Let's keep @reinit__is_reduced on `update` method.

ignite/metrics/mutual_information.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    _state_dict_all_req_keys = ("_sum_of_probabilities",)
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        super().reset()
+        self._sum_of_probabilities = torch.tensor(0.0, device=self._device)
+
+    @reinit__is_reduced
```
vfdev-5, line 78: This decorator is not needed anymore if we keep it on Entropy.update

vfdev-5 (commented):
Thanks for the update @kzkadc !
I left few other comments on how to use reinit__is_reduced decorator

ignite/metrics/mutual_information.py
vfdev-5: Can you please put a quick comment above this line why we can't use += op here. Something like
"We can't use += below as _sum_of_probabilities can be a scalar and prob.sum(dim=0) is a vector"

ignite/metrics/mutual_information.py
```diff
@@ -0,0 +1,94 @@
...
+    def reset(self) -> None:
+        super().reset()
+        self._sum_of_probabilities = torch.tensor(0.0, device=self._device)
+
+    def _update(self, prob: torch.Tensor, log_prob: torch.Tensor) -> None:
+        super()._update(prob, log_prob)
+        # We can't use += below as _sum_of_probabilities can be a scalar and prob.sum(dim=0) is a vector
+        self._sum_of_probabilities = self._sum_of_probabilities + prob.sum(dim=0).to(self._device)
+
+    @sync_all_reduce("_sum_of_probabilities")
```
vfdev-5, line 83: Here we need to reduce all intermediate vars: _sum_of_entropies, _sum_of_probabilities, _num_examples

kzkadc: Thanks for your review!
I fixed it.

PR #3231, opened 2024-04-04
Retry with lf

leej3:
Potentially supercedes #3229. Just checking some potential fixes on CI


leej3:
I think the "--lfnf none" behaviour described in the reproducer from the original PR will prevent this approach from working.

PR #3232, opened 2024-04-13
Add divergence metrics

kzkadc:
Description: added KL and JS divergences between two categorical predictions

Check list:

### DIFF
--- a/ignite/metrics/js_divergence.py
+++ b/ignite/metrics/js_divergence.py
@@ -17,7 +17,7 @@ class JSDivergence(KLDivergence):
            D_\text{JS}(\mathbf{p}_i \| \mathbf{q}_i) &= \frac{1}{2} D_\text{KL}(\mathbf{p}_i \| \mathbf{m}_i)
            + \frac{1}{2} D_\text{KL}(\mathbf{q}_i \| \mathbf{m}_i), \\
            \mathbf{m}_i &= \frac{1}{2}(\mathbf{p}_i + \mathbf{q}_i), \\
-           D_\text{KL}(\mathbf{p}_i, \mathbf{q}_i) &= \sum_{c=1}^C p_{i,c} \log \frac{p_{i,c}}{q_{i,c}}.
+           D_\text{KL}(\mathbf{p}_i \| \mathbf{q}_i) &= \sum_{c=1}^C p_{i,c} \log \frac{p_{i,c}}{q_{i,c}}.
        \end{align*}
 
     where :math:`\mathbf{p}_i` and :math:`\mathbf{q}_i` are the prediction and ground true probability tensors,
--- a/ignite/metrics/js_divergence.py
+++ b/ignite/metrics/js_divergence.py
@@ -20,7 +20,7 @@ class JSDivergence(KLDivergence):
            D_\text{KL}(\mathbf{p}_i \| \mathbf{q}_i) &= \sum_{c=1}^C p_{i,c} \log \frac{p_{i,c}}{q_{i,c}}.
        \end{align*}
 
-    where :math:`\mathbf{p}_i` and :math:`\mathbf{q}_i` are the prediction and ground true probability tensors,
+    where :math:`\mathbf{p}_i` and :math:`\mathbf{q}_i` are the prediction and ground truth probability tensors,
     and :math:`D_\text{KL}` is the KL-divergence.
 
     - ``update`` must receive output of the form ``(y_pred, y)``.
--- a/ignite/metrics/js_divergence.py
+++ b/ignite/metrics/js_divergence.py
@@ -1,0 +1,87 @@
+import torch
+import torch.nn.functional as F
+
+from ignite.exceptions import NotComputableError
+from ignite.metrics.kl_divergence import KLDivergence
+from ignite.metrics.metric import sync_all_reduce
+
+__all__ = ["JSDivergence"]
+
+
+class JSDivergence(KLDivergence):
+    r"""Calculates the mean of `Jensen-Shannon (JS) divergence
+    <https://en.wikipedia.org/wiki/Jensen%E2%80%93Shannon_divergence>`_.
+
+    .. math::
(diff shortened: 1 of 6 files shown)


ignite/metrics/js_divergence.py
```diff
@@ -0,0 +1,87 @@
...
+
+    .. math::
+       \begin{align*}
+           D_\text{JS}(\mathbf{p}_i \| \mathbf{q}_i) &= \frac{1}{2} D_\text{KL}(\mathbf{p}_i \| \mathbf{m}_i)
+           + \frac{1}{2} D_\text{KL}(\mathbf{q}_i \| \mathbf{m}_i), \\
+           \mathbf{m}_i &= \frac{1}{2}(\mathbf{p}_i + \mathbf{q}_i), \\
+           D_\text{KL}(\mathbf{p}_i \| \mathbf{q}_i) &= \sum_{c=1}^C p_{i,c} \log \frac{p_{i,c}}{q_{i,c}}.
+       \end{align*}
+
+    where :math:`\mathbf{p}_i` and :math:`\mathbf{q}_i` are the prediction and ground true probability tensors,
```
vfdev-5, line 23:
```suggestion
    where :math:`\mathbf{p}_i` and :math:`\mathbf{q}_i` are the prediction and ground truth probability tensors,
```

ignite/metrics/kl_divergence.py
```diff
@@ -0,0 +1,102 @@
...
+__all__ = ["KLDivergence"]
+
+
+class KLDivergence(Metric):
+    r"""Calculates the mean of `Kullback-Leibler (KL) divergence
+    <https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence>`_.
+
+    .. math:: D_\text{KL}(\mathbf{p}_i \| \mathbf{q}_i) = \sum_{c=1}^C p_{i,c} \log \frac{p_{i,c}}{q_{i,c}}
+
+    where :math:`\mathbf{p}_i` and :math:`\mathbf{q}_i` are the prediction and ground true probability tensors.
```
vfdev-5, line 18:
```suggestion
    where :math:`\mathbf{p}_i` and :math:`\mathbf{q}_i` are the prediction and ground truth probability tensors.
```

vfdev-5: According to the wiki page: https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence and also pytorch note on kl_div: https://pytorch.org/docs/stable/generated/torch.nn.KLDivLoss.html#torch.nn.KLDivLoss, in $D_{KL}(P \| Q)$, $P$ represents the data and $Q$ represents the model. Here in the docstring you have the opposite P for predictions and Q for targets. Should we use wiki's definitions?

ignite/metrics/kl_divergence.py
```diff
@@ -0,0 +1,102 @@
...
+        elif y_pred.ndim == 1:
+            raise ValueError(f"y_pred must be in the shape of (B, C) or (B, C, ...), got {y_pred.shape}.")
+
+        self._num_examples += y_pred.shape[0]
+        self._update(y_pred, y)
+
+    def _update(self, y_pred: torch.Tensor, y: torch.Tensor) -> None:
+        y_pred = F.log_softmax(y_pred, dim=1)
+        y = F.log_softmax(y, dim=1)
+        kl_sum = F.kl_div(y, y_pred, log_target=True, reduction="sum")
```
vfdev-5, line 95: Same question here, should we have instead: `F.kl_div(y_pred, y)`?

ignite/metrics/kl_divergence.py
vfdev-5: @kzkadc can you check above comment ?

kzkadc: Indeed, I think so too.

kzkadc: @vfdev-5 I swapped the order of the arguments to align with the KLDivLoss of pytorch. Could you check the change?

ignite/metrics/js_divergence.py
```diff
@@ -0,0 +1,87 @@
...
+
+    .. math::
+       \begin{align*}
+           D_\text{JS}(\mathbf{p}_i \| \mathbf{q}_i) &= \frac{1}{2} D_\text{KL}(\mathbf{p}_i \| \mathbf{m}_i)
+           + \frac{1}{2} D_\text{KL}(\mathbf{q}_i \| \mathbf{m}_i), \\
+           \mathbf{m}_i &= \frac{1}{2}(\mathbf{p}_i + \mathbf{q}_i), \\
+           D_\text{KL}(\mathbf{p}_i \| \mathbf{q}_i) &= \sum_{c=1}^C p_{i,c} \log \frac{p_{i,c}}{q_{i,c}}.
+       \end{align*}
+
+    where :math:`\mathbf{p}_i` and :math:`\mathbf{q}_i` are the prediction and ground truth probability tensors,
```
vfdev-5, line 23: Should we use same convention about which variable (p or q) corresponds to prediction and which to the ground truth? Here we have p = predictions and q = GT and for KLDivergence, it is the opposite.

kzkadc: Thanks for your review! I also swapped the definitions of p and q here.

vfdev-5 (approved):
Thanks a lot adding new metrics @kzkadc !

vfdev-5:
@kzkadc can you please check this failure with pytorch 1.5.1: https://github.com/pytorch/ignite/actions/runs/8730344048/job/23953994151

> TypeError: kl_div() got an unexpected keyword argument 'log_target'

Looks like `log_target` arg was not added in 1.5.1, can you see what would be the effort to support 1.5.1.
We have two options: a) fix ignite code if this is a minimal fix, b) drop support for 1.5.1 in our pytorch versions tests and replace it with a minimal later version. Option a) would be preferable but the fix should not be a large update of the code.

PR #3238, opened 2024-04-22
improve type support for fbrlogger

leej3:


vfdev-5:
@leej3 do you think it is possible to fix the CI here?

ignite/handlers/fbresearch_logger.py
```diff
@@ -1,18 +1,29 @@
...
 from ignite.handlers import Timer
+from ignite.handlers.utils import global_step_from_engine  # noqa
 
 
 MB = 1024.0 * 1024.0
 
+__all__ = ["FBResearchLogger", "global_step_from_engine"]
+
+
+def is_iterable(obj):
```
vfdev-5, line 20: We can remove this function

vfdev-5:
@leej3 can you please reopen this PR and change the merging branch to master?

ignite/utils.py
```diff
@@ -90,6 +92,130 @@ def _tree_map(
...
+            return f"{prefix}{item:.4f}"
+        elif torch.is_tensor(item):
+            if item.dim() == 0:
+                return f"{prefix}{item.item():.4f}"  # Format scalar tensor without brackets
+            elif item.dim() == 1 and item.size(0) > 10:
+                return f"{prefix}[" + ", ".join(f"{x.item():.4f}" for x in item[:10]) + ", ...]"
+            elif item.dim() == 1:
+                return f"{prefix}[" + ", ".join(f"{x.item():.4f}" for x in item) + "]"
+            else:
+                return f"{prefix}Shape: {item.shape}"
```
vfdev-5, line 191:
```suggestion
                return f"{prefix}Shape: {list(item.shape)}"
```
to avoid unnecessary `torch.Size(` substrings. Maybe, even "Shape:" can be omitted to save space.

vfdev-5: But we say "Shape" before providing the shape values, so this is clear. Actually, I wonder if this finally makes sense even to log the shape in the loop...

```
2024-04-22 12:05:47,843 trainer INFO: Train: start epoch [1/4]
... Epoch [1/4]  [20/100]:  ETA: 0:00:00  lr: 0.00100  ...  Shape: [32, 3, 256, 256]
... Epoch [1/4]  [40/100]:  ETA: 0:00:00  lr: 0.00100  ...  Shape: [32, 3, 256, 256]
... Epoch [1/4]  [60/100]:  ETA: 0:00:00  lr: 0.00100  ...  Shape: [32, 3, 256, 256]
... Epoch [1/4]  [80/100]:  ETA: 0:00:00 lr: 0.00100   ...  Shape: [32, 3, 256, 256]
```

docs/source/utils.rst
```diff
@@ -7,7 +7,6 @@ Module with helper methods
 
 .. autosummary::
     :nosignatures:
-    :autolist:
```
vfdev-5, line 10: Why do you remove this, John?

ignite/handlers/fbresearch_logger.py
```diff
@@ -1,18 +1,20 @@
...
-
-# from typing import Any, Dict, Optional, Union
+from typing import Any, Callable, List, Optional
 
 import torch
 
+from ignite import utils
 from ignite.engine import Engine, Events
 from ignite.handlers import Timer
+from ignite.handlers.utils import global_step_from_engine  # noqa
```
vfdev-5, line 11: nit: we can remove this as there is no way of using `global_step_from_engine` with this logger

vfdev-5 (approved):
LGTM, just few questions to clarify and we can land it.
Failing tpu jobs is unrelated

docs/source/utils.rst
leej3: `make clean; make html` fails if that line is added back in and succeeds when it is removed. I had another look at it there and have pushed a fix for the AutolistAutosummary class so I don't need to remove this line any more.

vfdev-5:
Thanks for the fixes, John!

PR #3242, opened 2024-04-26
Make testing improvements

leej3:
Description:
Uses xdist loadgroups to improve performance of tests when running them on multiple workers. Setting `--dist=loadgroups` when calling pytest should now send all tests with the "distributed" mark to the same worker. This  helps to avoids some of the errors where tests are trying to use the resources made available by the distributed fixture. A similar solution is now implemented for tests using the visdom_server fixture.

Also avoids the warnings for unregistered pytest marks and fixes some stray mypy failures.


vfdev-5 (approved):
Looks good, thanks John!

PR #3243, opened 2024-05-01
Add MaximumMeanDiscrepancy metric

kzkadc:
Description: added maximum mean discrepancy (MMD) metric, which measures the discrepancy of two distributions.
https://www.onurtunali.com/ml/2019/03/08/maximum-mean-discrepancy-in-machine-learning.html

Check list:

### DIFF
--- a/ignite/metrics/maximum_mean_discrepancy.py
+++ b/ignite/metrics/maximum_mean_discrepancy.py
@@ -16,9 +16,9 @@ class MaximumMeanDiscrepancy(Metric):
        \begin{align*}
            \text{MMD}^2 (P,Q) &= \underset{\| f \| \leq 1}{\text{sup}} | \mathbb{E}_{X\sim P}[f(X)]
            - \mathbb{E}_{Y\sim Q}[f(Y)] |^2 \\
-           &\approx \frac{1}{B(B-1)} \sum_{i} \sum_{j\neq i}k(\mathbf{x}_i,\mathbf{x}_j)
-           -\frac{2}{B^2}\sum_{i} \sum_{j} k(\mathbf{x}_i,\mathbf{y}_j)
-           + \frac{1}{B(B-1)} \sum_{i} \sum_{j\neq i}k(\mathbf{y}_i,\mathbf{y}_j)
+           &\approx \frac{1}{B(B-1)} \sum_{i=1}^B \sum_{\substack{j=1 \\ j\neq i}}^B k(\mathbf{x}_i,\mathbf{x}_j)
+           -\frac{2}{B^2}\sum_{i=1}^B \sum_{j=1}^B k(\mathbf{x}_i,\mathbf{y}_j)
+           + \frac{1}{B(B-1)} \sum_{i=1}^B \sum_{\substack{j=1 \\ j\neq i}}^B k(\mathbf{y}_i,\mathbf{y}_j)
        \end{align*}
 
     where :math:`B` is the batch size, and :math:`\mathbf{x}_i` and :math:`\mathbf{y}_j` are
--- a/ignite/metrics/maximum_mean_discrepancy.py
+++ b/ignite/metrics/maximum_mean_discrepancy.py
@@ -74,7 +74,7 @@ class MaximumMeanDiscrepancy(Metric):
            1.0726975202560425
     """
 
-    _state_dict_all_req_keys = ("_sum_of_mmd", "_num_batches")
+    _state_dict_all_req_keys = ("_sum_of_mmd2", "_num_batches")
 
     def __init__(
         self, var: float = 1.0, output_transform: Callable = lambda x: x, device: torch.device = torch.device("cpu")
@@ -84,7 +84,7 @@ class MaximumMeanDiscrepancy(Metric):
 
     @reinit__is_reduced
     def reset(self) -> None:
-        self._sum_of_mmd = torch.tensor(0.0, device=self._device)
+        self._sum_of_mmd2 = torch.tensor(0.0, device=self._device)
         self._num_batches = 0
 
     @reinit__is_reduced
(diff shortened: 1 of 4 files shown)


vfdev-5 (commented):
Thanks for the PR, @kzkadc 
Right now CI is failing with related broken tests, can you please check them ?

ignite/metrics/maximum_mean_discrepancy.py
```diff
@@ -0,0 +1,131 @@
...
+
+class MaximumMeanDiscrepancy(Metric):
+    r"""Calculates the mean of `maximum mean discrepancy (MMD)
+    <https://www.onurtunali.com/ml/2019/03/08/maximum-mean-discrepancy-in-machine-learning.html>`_.
+
+    .. math::
+       \begin{align*}
+           \text{MMD}^2 (P,Q) &= \underset{\| f \| \leq 1}{\text{sup}} | \mathbb{E}_{X\sim P}[f(X)]
+           - \mathbb{E}_{Y\sim Q}[f(Y)] |^2 \\
+           &\approx \frac{1}{B(B-1)} \sum_{i} \sum_{j\neq i}k(\mathbf{x}_i,\mathbf{x}_j)
```
vfdev-5, line 19: Is it possible to specify i, j ranges?

ignite/metrics/maximum_mean_discrepancy.py
```diff
@@ -0,0 +1,131 @@
...
+        n = x.shape[0]
+        XX = (XX.sum() - n) / (n * (n - 1))
+        YY = (YY.sum() - n) / (n * (n - 1))
+        XY = XY.sum() / (n * n)
+
+        # mmd cannot be negative
+        mmd2 = (XX - 2.0 * XY + YY).clamp(min=0.0)
+
+        self._sum_of_mmd += mmd2.sqrt().to(self._device)
+        self._num_batches += 1
```
vfdev-5, line 125: will resulting metric value practically depend on the batch size?

kzkadc: Thanks for your review.
Yes, it depends on the batch size.
But mmd2 above is an unbiased estimator of MMD^2 ( https://jmlr.org/papers/volume13/gretton12a/gretton12a.pdf , Lemma 6 ).

I will modify the code to accumulate mmd2 and take the sqrt in compute().
Then, we can expect that the computed mmd will converge with a sufficient number of batches.

ignite/metrics/maximum_mean_discrepancy.py
```diff
@@ -0,0 +1,131 @@
...
+    .. math::
+       \begin{align*}
+           \text{MMD}^2 (P,Q) &= \underset{\| f \| \leq 1}{\text{sup}} | \mathbb{E}_{X\sim P}[f(X)]
+           - \mathbb{E}_{Y\sim Q}[f(Y)] |^2 \\
+           &\approx \frac{1}{B(B-1)} \sum_{i=1}^B \sum_{\substack{j=1 \\ j\neq i}}^B k(\mathbf{x}_i,\mathbf{x}_j)
+           -\frac{2}{B^2}\sum_{i=1}^B \sum_{j=1}^B k(\mathbf{x}_i,\mathbf{y}_j)
+           + \frac{1}{B(B-1)} \sum_{i=1}^B \sum_{\substack{j=1 \\ j\neq i}}^B k(\mathbf{y}_i,\mathbf{y}_j)
+       \end{align*}
+
+    where :math:`B` is the batch size, and :math:`\mathbf{x}_i` and :math:`\mathbf{y}_j` are
```
vfdev-5, line 24: Sorry to coming back to this question, actually, I'm not quite sure about that B is the batch size.
According to https://www.onurtunali.com/ml/2019/03/08/maximum-mean-discrepancy-in-machine-learning.html, their `m` is the num of samples.
Also, https://jmlr.org/papers/volume13/gretton12a/gretton12a.pdf sets the problem as 
> Let x and y be random variables defined on a topological space X, with respective
Borel probability measures p and q . Given observations X := {x1, . . . , xm} and Y := {y1, . . . , yn},
independently and identically distributed (i.i.d.) from p and q, respectively, can we decide whether
p != q?

where m and n are num of samples of the observations.

Should not we use the same idea and replace B per N = num of samples ?

ignite/metrics/maximum_mean_discrepancy.py
```diff
@@ -0,0 +1,131 @@
...
+
+from ignite.exceptions import NotComputableError
+from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
+
+__all__ = ["MaximumMeanDiscrepancy"]
+
+
+class MaximumMeanDiscrepancy(Metric):
+    r"""Calculates the mean of `maximum mean discrepancy (MMD)
+    <https://www.onurtunali.com/ml/2019/03/08/maximum-mean-discrepancy-in-machine-learning.html>`_.
```
vfdev-5, line 13: Let's also put this link: https://jmlr.org/papers/volume13/gretton12a/gretton12a.pdf to the docstring

ignite/metrics/maximum_mean_discrepancy.py
```diff
@@ -0,0 +1,131 @@
...
+        v = self.var
+        XX = torch.exp(-0.5 * dxx / v)
+        YY = torch.exp(-0.5 * dyy / v)
+        XY = torch.exp(-0.5 * dxy / v)
+
+        # unbiased
+        n = x.shape[0]
+        XX = (XX.sum() - n) / (n * (n - 1))
+        YY = (YY.sum() - n) / (n * (n - 1))
+        XY = XY.sum() / (n * n)
```
vfdev-5, line 119: If we were to implement mmd as kernel average over the full set, we can accumulate xx_sum, yy_sum, xy_sum and num_samples on `update` (e.g. `xx_sum += xx.sum()`) and in `compute`, run mmd computation:
```
mmd2 = (xx_sum - 2.0 * xy_sum + yy_sum).clamp(min=0.0) / num_samples
```
What do you think ?

kzkadc: If we were to compute MMD over all samples, we would had to consider all $N^2$ combinations of k(x, y) from different batches.
I think it would be better to XX, YY, and XY within each batch and accumulate them.

ignite/metrics/maximum_mean_discrepancy.py
kzkadc: Thank you for the suggestion.

Although it is ideal to compute MMD using all $N$ samples, doing so requires to compute $N^2$ combinations of k(x, y), where we have to pool all inputs.
So, this metric computes MMD for each batch and takes the average.

I think using $B$ would be better so that users can be aware of this point.

ignite/metrics/maximum_mean_discrepancy.py
vfdev-5: OK, I think I get now your point about $N^2$ combinations. So, this metrics is computing average MMD over batch. Larger the batch size, closer "average MMD over batch" would be to MMD on N samples.

vfdev-5 (approved):
LGTM, thanks a lot for the PR @kzkadc !

PR #3249, opened 2024-05-21
Fix mps not functional

leej3:
Tests on mac-os-latest (macos-14-arm64) failing. MPS is not available on this as far as I can tell (see [here](https://github.com/actions/runner-images/issues/9254)

### DIFF
--- a/tests/ignite/__init__.py
+++ b/tests/ignite/__init__.py
@@ -3,3 +3,14 @@ import torch
 
 def cpu_and_maybe_cuda():
     return ("cpu",) + (("cuda",) if torch.cuda.is_available() else ())
+
+
+def is_mps_available_and_functional():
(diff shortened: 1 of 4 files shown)


tests/utils_for_tests.py
```diff
@@ -0,0 +1,12 @@
+import torch
```
vfdev-5, line 1: Can you put this check into `tests/ignite/__init__.py` and import it as `from tests.ignite import is_mps_available_and_functional`

vfdev-5 (approved):
LGTM, thanks!

vfdev-5:
Still failing on macos: https://github.com/pytorch/ignite/actions/runs/9175447451/job/25228464761?pr=3249#step:15:2267

PR #3251, opened 2024-05-23
downgrade retry action for gpu tests

leej3:


.github/workflows/gpu-tests.yml
```diff
@@ -121,7 +121,7 @@ jobs:
 
       - name: Run GPU Unit Tests
         continue-on-error: false
-        uses: nick-fields/retry@v3
+        uses: nick-fields/retry@14672906e672a08bd6eeb15720e9ed3ce869cdd4
```
vfdev-5, line 124: Can we specify a version like `v2` or v3 with minor and patch instead of a commit ?

leej3: yep, looks like 2.9.0 is a sufficient downgrade.

PR #3253, opened 2024-05-29
fix scheduled tests

leej3:
This should fix the failing scheduled tests. Pytest isn't found for them. These tests use the setup-miniconda action for python env installation. The action needs a bash login shell in subsequent steps for the env to be active.

Here's a description of the problem: https://stackoverflow.com/questions/72703363/how-to-activate-conda-environment-in-github-actions

Also torch.backends.mps does not exist for older pytorch versions and "filelock" should be listed explicitly as a dependency.


vfdev-5:
https://github.com/pytorch/ignite/actions/runs/9283545101

PR #3254, opened 2024-05-29
Fix scheduled tests

leej3:
supercedes #3253

See:
- https://github.com/pytorch/ignite/actions/runs/9284541602
- https://github.com/pytorch/ignite/actions/runs/9286709894

### DIFF
--- a/ignite/distributed/comp_models/base.py
+++ b/ignite/distributed/comp_models/base.py
@@ -5,7 +5,7 @@ from typing import Any, Callable, cast, List, Optional, Union
 import torch
 from packaging.version import Version
 
-_torch_version_ge_112 = Version(torch.__version__) > Version("1.12.0")
+_torch_version_gt_112 = Version(torch.__version__) > Version("1.12.0")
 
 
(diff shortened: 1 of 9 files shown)


ignite/distributed/comp_models/base.py
```diff
@@ -5,7 +5,7 @@
 import torch
 from packaging.version import Version
 
-_torch_version_le_112 = Version(torch.__version__) > Version("1.12.0")
+_torch_version_ge_112 = Version(torch.__version__) > Version("1.12.0")
```
vfdev-5, line 8:
```suggestion
_torch_version_gt_112 = Version(torch.__version__) > Version("1.12.0")
```

leej3: Touché! To be more precise, based on [this](https://pytorch.org/docs/stable/notes/mps.html) it looks like we want ge 1.12.3, shall I change it to that? Otherwise, yes, gt_112

vfdev-5 (approved):
Thanks for the fix, @leej3 !

PR #3255, opened 2024-06-03
Rename common-test-functionality.sh to common_test_functionality.sh

vfdev-5:
Description:
- Rename common-test-functionality.sh to common_test_functionality.sh


leej3 (commented):
underscores noted

PR #3258, opened 2024-06-25
Introduce a variable skip_unrolling in class Metric

### LINKED ISSUE #2940: Metric with multiple input runs in an unexpected way.
## ❓ Questions/Help/Support

My customized loss requires two pairs of input:
```python
class MyLoss(nn.Module):
    def __init__(self, ca: float = 1.0, cb: float = 1.0) -> None:
        super().__init__()
        self.ca = ca
        self.cb = cb

    def forward(self, y_pred: Tuple[torch.Tensor, torch.Tensor], y_true: Tuple[torch.Tensor, torch.Tensor]) -> torch.Tensor:
        a_true, b_true = y_true
        a_pred, b_pred = y_pred
        return self.ca * F.mse_loss(a_pred, a_true) + self.cb * F.cross_entropy(b_pred, b_true)

```

When I try to log the loss with `Loss` metric:
```python
loss = MyLoss(0.5, 1.0)
metrics = {
    "Loss": Loss(loss)
}
train_evaluator = create_supervised_evaluator(model, metrics, device, prepare_batch=prepare_batch)
```

It will crash on line:
https://github.com/pytorch/ignite/blob/4825bb6ecd5f432f0423bf3d587fbce04b9d2d4f/ignite/metrics/metric.py#L308
because it treats all inputs as independent pair of  y_pred and y, which is not what `MyLoss` need.

I dug into the source code I found https://github.com/pytorch/ignite/pull/2055 introduces a new feature, which causes this issue.
So, what are the best practices for dealing with multiple input losses?

simeetnayan81:
Fixes #2940

Description:
Introduce a variable skip_unrolling in class Metric as discussed here https://discord.com/channels/831462531327328276/1110662056622964860/1253769540710567977

Check list:

### DIFF
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -29,6 +29,9 @@ class Loss(Metric):
         device: specifies which device updates are accumulated on. Setting the
             metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
             non-blocking. By default, CPU.
+        skip_unrolling: specifies whether input should be unrolled or not before it is passed to to loss_fn.
+            Should be true for multi-output model, for example, if ``y_pred`` contains multi-ouput as
+            ``(y_pred_a, y_pred_b)``
 
     Attributes:
         required_output_keys: dictionary defines required keys to be found in ``engine.state.output`` if the
@@ -62,6 +65,8 @@ class Loss(Metric):
 
             -0.3499999...
 
+    .. versionchanged:: 0.5.1
+        ``skip_unrolling`` attribute is added.
     """
 
     required_output_keys = ("y_pred", "y", "criterion_kwargs")
@@ -73,8 +78,9 @@ class Loss(Metric):
         output_transform: Callable = lambda x: x,
         batch_size: Callable = len,
         device: Union[str, torch.device] = torch.device("cpu"),
+        skip_unrolling: bool = False,
     ):
-        super(Loss, self).__init__(output_transform, device=device)
+        super(Loss, self).__init__(output_transform, device=device, skip_unrolling=skip_unrolling)
         self._loss_fn = loss_fn
         self._batch_size = batch_size
 
--- a/ignite/metrics/loss.py
+++ b/ignite/metrics/loss.py
@@ -66,7 +66,7 @@ class Loss(Metric):
             -0.3499999...
 
     .. versionchanged:: 0.5.1
-        ``skip_unrolling`` attribute is added.
+        ``skip_unrolling`` argument is added.
     """
 
     required_output_keys = ("y_pred", "y", "criterion_kwargs")
--- a/ignite/metrics/metric.py
+++ b/ignite/metrics/metric.py
@@ -236,7 +236,7 @@ class Metric(Serializable, metaclass=ABCMeta):
         skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
             true for multi-output model, for example, if ``y_pred`` contains multi-ouput as ``(y_pred_a, y_pred_b)``
 
-            Example:
+            Examples:
                 The following example shows a custom loss metric that expects input from a multi-output model.
 
                 .. code-block:: python
--- a/ignite/metrics/metric.py
+++ b/ignite/metrics/metric.py
@@ -300,7 +300,7 @@ class Metric(Serializable, metaclass=ABCMeta):
     _required_output_keys = required_output_keys
 
     def __init__(
-        self, output_transform: Callable = lambda x: x, device: Union[str, torch.device] = torch.device("cpu")
+        self, output_transform: Callable = lambda x: x, device: Union[str, torch.device] = torch.device("cpu"), skip_unrolling : bool =False
     ):
         self._output_transform = output_transform
 
@@ -309,6 +309,7 @@ class Metric(Serializable, metaclass=ABCMeta):
             raise ValueError("Cannot create metric on an XLA device. Use device='cpu' instead.")
 
         self._device = torch.device(device)
+        self._skip_unrolling = skip_unrolling
         self.reset()
 
     @abstractmethod
@@ -377,30 +378,33 @@ class Metric(Serializable, metaclass=ABCMeta):
         """
 
         output = self._output_transform(engine.state.output)
-        if isinstance(output, Mapping):
-            if self.required_output_keys is None:
-                raise TypeError(
-                    f"Transformed engine output for {self.__class__.__name__} metric should be a tuple/list, "
-                    f"but given {type(output)}"
-                )
-            if not all([k in output for k in self.required_output_keys]):
-                raise ValueError(
-                    "When transformed engine's output is a mapping, "
-                    f"it should contain {self.required_output_keys} keys, but given {list(output.keys())}"
-                )
-            output = tuple(output[k] for k in self.required_output_keys)
+        if self._skip_unrolling is False:
+            if isinstance(output, Mapping):
+                if self.required_output_keys is None:
+                    raise TypeError(
+                        f"Transformed engine output for {self.__class__.__name__} metric should be a tuple/list, "
+                        f"but given {type(output)}"
+                    )
+                if not all([k in output for k in self.required_output_keys]):
+                    raise ValueError(
+                        "When transformed engine's output is a mapping, "
+                        f"it should contain {self.required_output_keys} keys, but given {list(output.keys())}"
+                    )
+                output = tuple(output[k] for k in self.required_output_keys)
 
-        if isinstance(output, Sequence) and all([_is_list_of_tensors_or_numbers(o) for o in output]):
-            if not (len(output) == 2 and len(output[0]) == len(output[1])):
-                raise ValueError(
-                    f"Output should have 2 items of the same length, "
-                    f"got {len(output)} and {len(output[0])}, {len(output[1])}"
-                )
-            for o1, o2 in zip(output[0], output[1]):
-                # o1 and o2 are list of tensors or numbers
-                tensor_o1 = _to_batched_tensor(o1)
-                tensor_o2 = _to_batched_tensor(o2, device=tensor_o1.device)
-                self.update((tensor_o1, tensor_o2))
+            if isinstance(output, Sequence) and all([_is_list_of_tensors_or_numbers(o) for o in output]):
+                if not (len(output) == 2 and len(output[0]) == len(output[1])):
+                    raise ValueError(
+                        f"Output should have 2 items of the same length, "
+                        f"got {len(output)} and {len(output[0])}, {len(output[1])}"
+                    )
+                for o1, o2 in zip(output[0], output[1]):
+                    # o1 and o2 are list of tensors or numbers
+                    tensor_o1 = _to_batched_tensor(o1)
+                    tensor_o2 = _to_batched_tensor(o2, device=tensor_o1.device)
+                    self.update((tensor_o1, tensor_o2))
+            else:
+                self.update(output)
         else:
             self.update(output)
 
--- a/ignite/metrics/metric.py
+++ b/ignite/metrics/metric.py
@@ -234,7 +234,7 @@ class Metric(Serializable, metaclass=ABCMeta):
             metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
             non-blocking. By default, CPU.
         skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
-            true for multi-output model, for example, if``y_pred`` contains multi-ouput as ``(y_pred_a, y_pred_b)``
+            true for multi-output model, for example, if ``y_pred`` contains multi-ouput as ``(y_pred_a, y_pred_b)``
             Alternatively, output_transform can be used to handle this.
 
     Attributes:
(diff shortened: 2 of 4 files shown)


ignite/metrics/metric.py
```diff
@@ -377,30 +378,33 @@ def iteration_completed(self, engine: Engine) -> None:
...
+                        f"but given {type(output)}"
+                    )
+                if not all([k in output for k in self.required_output_keys]):
+                    raise ValueError(
+                        "When transformed engine's output is a mapping, "
+                        f"it should contain {self.required_output_keys} keys, but given {list(output.keys())}"
+                    )
+                output = tuple(output[k] for k in self.required_output_keys)
+
+            if isinstance(output, Sequence) and all([_is_list_of_tensors_or_numbers(o) for o in output]):
```
vfdev-5, line 395: `self._skip_unrolling` should skip this if-clause but keep `if isinstance(output, Mapping):`

vfdev-5 (commented):
Thanks for the PR @simeetnayan81 
Left a comment about the implementation

ignite/metrics/metric.py
```diff
@@ -300,7 +300,7 @@ def compute(self):
     _required_output_keys = required_output_keys
 
     def __init__(
-        self, output_transform: Callable = lambda x: x, device: Union[str, torch.device] = torch.device("cpu")
+        self, output_transform: Callable = lambda x: x, device: Union[str, torch.device] = torch.device("cpu"), skip_unrolling : bool =False
```
vfdev-5, line 303: Let's also update above docstring by adding the new argument. Please also check CONTRIBUTING guideline the part about how to add `.. versionadded::` tag in the bottom of the docstring. Version to put should be `0.5.1`

ignite/metrics/metric.py
simeetnayan81: Okay. You mean if skip unrolling is True, then only ``` if isinstance(output, Mapping):  ``` clause should run. If it is false then the ``` if isinstance(output, Sequence) ``` clause should run ?

simeetnayan81:
```       if isinstance(output, Mapping):
                if self.required_output_keys is None:
                    raise TypeError(
                        f"Transformed engine output for {self.__class__.__name__} metric should be a tuple/list, "
                        f"but given {type(output)}"
                    )
                if not all([k in output for k in self.required_output_keys]):
                    raise ValueError(
                        "When transformed engine's output is a mapping, "
                        f"it should contain {self.required_output_keys} keys, but given {list(output.keys())}"
                    )
                output = tuple(output[k] for k in self.required_output_keys)

        if self._skip_unrolling is False:
            if isinstance(output, Sequence) and all([_is_list_of_tensors_or_numbers(o) for o in output]):
                if not (len(output) == 2 and len(output[0]) == len(output[1])):
                    raise ValueError(
                        f"Output should have 2 items of the same length, "
                        f"got {len(output)} and {len(output[0])}, {len(output[1])}"
                    )
                for o1, o2 in zip(output[0], output[1]):
                    # o1 and o2 are list of tensors or numbers
                    tensor_o1 = _to_batched_tensor(o1)
                    tensor_o2 = _to_batched_tensor(o2, device=tensor_o1.device)
                    self.update((tensor_o1, tensor_o2))
            else:
                self.update(output)
        else:
            self.update(output) ```
            Sort of something like this?

Comment (vfdev-5) [reply ignite/metrics/metric.py:395]: You join together two if-clauses:
```
if self._skip_unrolling is False:
        if isinstance(output, Sequence) and all([_is_list_of_tensors_or_numbers(o) for o in output]):
```

Comment (simeetnayan81) [reply ignite/metrics/metric.py:395]:
```
    if (not self._skip_unrolling) and isinstance(output, Sequence) and all([_is_list_of_tensors_or_numbers(o) for o in output])  :
            if not (len(output) == 2 and len(output[0]) == len(output[1])):
                raise ValueError(
                    f"Output should have 2 items of the same length, "
                    f"got {len(output)} and {len(output[0])}, {len(output[1])}"
                )
            for o1, o2 in zip(output[0], output[1]):
                # o1 and o2 are list of tensors or numbers
                tensor_o1 = _to_batched_tensor(o1)
                tensor_o2 = _to_batched_tensor(o2, device=tensor_o1.device)
                self.update((tensor_o1, tensor_o2))

    else:
        self.update(output) 
 ```
 
Is it better now?

ignite/metrics/metric.py
simeetnayan81: Should I add ``` .. versionchanged::``` instead?

vfdev-5: Yes, correct, it should be versionchanged tag, https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md#writing-documentation

ignite/metrics/metric.py
vfdev-5: yes, it is better now. Once you updated the code, please do not forget to reformat the code using ufmt: https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md#formatting-code to have compliant code

ignite/metrics/metric.py
simeetnayan81: On it. Thanks.

simeetnayan81: Have made the changes. Kindly review.

ignite/metrics/metric.py
```diff
@@ -233,6 +233,9 @@ class Metric(Serializable, metaclass=ABCMeta):
         device: specifies which device updates are accumulated on. Setting the
             metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
             non-blocking. By default, CPU.
+        skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
+            true for multi-output model, for example, if``y_pred`` contains multi-ouput as ``(y_pred_a, y_pred_b)``
```
vfdev-5, line 237:
```suggestion
            true for multi-output model, for example, if ``y_pred`` contains multi-ouput as ``(y_pred_a, y_pred_b)``
```

ignite/metrics/metric.py
```diff
@@ -233,6 +233,9 @@ class Metric(Serializable, metaclass=ABCMeta):
...
+            Alternatively, output_transform can be used to handle this.
```
vfdev-5, line 238:
```suggestion
            Alternatively, ``output_transform`` can be used to handle this.
```

ignite/metrics/metric.py
```diff
@@ -290,8 +293,8 @@ def compute(self):
 
             res = evaluator.run(data)
 
-    .. versionchanged:: 0.4.2
-        ``required_output_keys`` became public attribute.
+    .. versionchanged:: 0.5.1
+        ``skip_unrolling`` attribute is introduced for Metric class.
```
vfdev-5, line 297:
```suggestion
        ``skip_unrolling`` attribute is added.
```

vfdev-5 (commented):
Thanks for the update! I added few minor comments. Let's add new feature tests and run the CI to see if any failures

ignite/metrics/metric.py
```diff
@@ -290,8 +293,8 @@ def compute(self):
 
             res = evaluator.run(data)
 
-    .. versionchanged:: 0.4.2
```
vfdev-5, line 293: Actually, we need to keep this and add the new `versionchanged` tag:
```
.. versionchanged:: 0.4.2
    ``required_output_keys`` became public attribute.

.. versionchanged:: 0.5.1
    ``skip_unrolling`` attribute is added.
```
So, let's revert this suppression

simeetnayan81:
Tests should be added to end of the test_metric.py file?

vfdev-5:
Yes, you can add it in the end of the file

simeetnayan81:
```skip_unrolling = False ``` is already covered by all the prior tests. I have added a test for when ``` skip_unrolling = True ```. Kindly review and let me know the changes.

**Review by vfdev-5 — COMMENTED** [review]
Looks good @simeetnayan81 , thanks, let's just add an example of usage of the flag in the docstring and it will be good to go, once CI is green (except unrelated failures)

**Comment by simeetnayan81** [comment]
@vfdev-5 Before adding the docstring, I wanted to confirm, to make skip_unrolling effective for the loss function, we might also need to change this.
[https://github.com/pytorch/ignite/blob/master/ignite/metrics/loss.py#L77](url)
Prev:
```    
def __init__(
        self,
        loss_fn: Callable,
        batch_size: Callable = len,
    ):
        super(Loss, self).__init__(output_transform, device=device)
```
Change to:
```    
def __init__(
        self,
        loss_fn: Callable,
        batch_size: Callable = len,
        skip_unrolling=False
    ):
        super(Loss, self).__init__(output_transform, device=device, skip_unrolling=skip_unrolling)
```

**Comment by vfdev-5** [comment]
@simeetnayan81 yes, you are right, we need to add this new arg to all metrics defining a constructor. Let's update Loss metric here and update other metrics in a follow-up PR.

**Comment by simeetnayan81** [comment]
Things to do in a follow-up PR.
- [ ] Add test for updated Loss class
- [ ] Update other sub-classes of Metric with ```skip_unrolling``` arg as required, add tests and docstring

**Comment by vfdev-5** [comment]
Thanks for the updates and the TODO. Can we do this point here ?

> Add test for updated Loss class

**Comment by simeetnayan81** [comment]
Alright @vfdev-5

File: ignite/metrics/metric.py [hunk 7]
```diff
@@ -233,6 +233,58 @@ class Metric(Serializable, metaclass=ABCMeta):
...
             non-blocking. By default, CPU.
+        skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
+            true for multi-output model, for example, if ``y_pred`` contains multi-ouput as ``(y_pred_a, y_pred_b)``
+
+            Example usage:
+                The following example shows a custom loss metric that expects input from a multi-output model.
+                .. code-block:: python
+                    import torch
+                    import torch.nn as nn
+                    import torch.nn.functional as F
```
vfdev-5, line 244: I think we should reformat here to fix docs tests:
```suggestion
            Examples:
            
                The following example shows a custom loss metric that expects input from a multi-output model.
            
                .. code-block:: python

                    import torch
                    import torch.nn as nn
                    import torch.nn.functional as F
```

See for example here: https://github.com/pytorch/ignite/blob/5a66d9e09fad42edb1fdda88a426a8658876fedd/ignite/engine/engine.py#L38-L46

simeetnayan81:
Have made the changes, the new test works locally.

ignite/metrics/loss.py
```diff
@@ -62,6 +65,8 @@ class Loss(Metric):
 
             -0.3499999...
 
+    .. versionchanged:: 0.5.1
+        ``skip_unrolling`` attribute is added.
```
vfdev-5, line 69:
```suggestion
        ``skip_unrolling`` argument is added.
```

ignite/metrics/metric.py
```diff
@@ -233,6 +233,60 @@ class Metric(Serializable, metaclass=ABCMeta):
         device: specifies which device updates are accumulated on. Setting the
             metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
             non-blocking. By default, CPU.
+        skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
+            true for multi-output model, for example, if ``y_pred`` contains multi-ouput as ``(y_pred_a, y_pred_b)``
+
+            Example usage:
```
vfdev-5, line 239:
```suggestion
            Example:
```

ignite/metrics/metric.py
```diff
@@ -292,6 +346,9 @@ def compute(self):
 
     .. versionchanged:: 0.4.2
         ``required_output_keys`` became public attribute.
+
+    .. versionchanged:: 0.5.1
+        ``skip_unrolling`` attribute is added.
```
vfdev-5, line 351:
```suggestion
        ``skip_unrolling`` argument is added.
```

ignite/metrics/metric.py
```diff
@@ -233,6 +233,60 @@ class Metric(Serializable, metaclass=ABCMeta):
...
+                        "Loss": Loss(loss, skip_unrolling=True)
+                    }
+                    train_evaluator = create_supervised_evaluator(model, metrics, device, prepare_batch=prepare_batch)
+
+
+                    data = range(10)
+                    train_evaluator.run(data)
+                    train_evaluator.state.metrics["Loss"]
+
+            Alternatively, ``output_transform`` can be used to handle this.
```
vfdev-5, line 289: This line is from the `skip_unrolling` docstring, please move it there

ignite/metrics/metric.py
```diff
@@ -233,6 +233,60 @@ class Metric(Serializable, metaclass=ABCMeta):
         device: specifies which device updates are accumulated on. Setting the
             metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
             non-blocking. By default, CPU.
+        skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
+            true for multi-output model, for example, if ``y_pred`` contains multi-ouput as ``(y_pred_a, y_pred_b)``
+
+            Example:
```
vfdev-5, line 239:
```suggestion
            Examples:
```

simeetnayan81:
The test is failing because ```list[torch.Tensor, torch.Tensor]``` is supported on python 3.9 and above. Let me modify it a bit.

vfdev-5 (approved):
Thanks @simeetnayan81 , lgtm

PR #3263, opened 2024-07-16
Exclude special attributes from causing `MetricsLambda` creation

### LINKED ISSUE #3262: To exclude special attributes from causing `MetricsLambda` creation
Hi @vfdev-5 ! Long time no see! Happy to be here again!

It might be useful to exclude python special attributes from being subject to creating `MetricsLambda` in `Metric::__getattr__`.
This is the case I saw the current behavior might make trouble. I was using an Ignite metric as the `compute_metrics` argument to HuggingFace Trainer API. There, they do a check on the argument signature using `inspect.signature` which itself accesses `__signature__` attribute of the object which raises an error:

```python
class CustomMetric(Metric):
    def reset(self):
        ...
    def update(self, output):
        ...
    def compute(self):
        ...
    def __call__(self, eval_pred, compute_result=False): # HF trainer makes sure this has a `compute_result` arg otherwise raises an error
        self.update(eval_pred)
        if compute_result:
            return self.compute()
```
The check takes place [here](https://github.com/huggingface/transformers/blob/25e5e3fa56d73d07d6a1c46306a5f3f1fd862463/src/transformers/trainer.py#L395-L401):
```python
        if args.batch_eval_metrics and compute_metrics is not None:
            if "compute_result" not in inspect.signature(compute_metrics).parameters.keys():
                raise ValueError(
                    "When using `batch_eval_metrics`, your `compute_metrics` function must take a `compute_result`"
                    " boolean argument which will be triggered after the last batch of the eval set to signal that the"
                    " summary statistics should be returned by the function."
                )
```

```python
trainer = Trainer(
    model=model,
    tokenizer=tokenizer,
    args=args,
    data_collator=data_collator,
    train_dataset=dataset,
    eval_dataset=dataset,
    compute_metrics=CustomMetric(),
)
```

```
TypeError: unexpected object <function Metric.__getattr__.<locals>.wrapper at 0x7eda2cf82200> in __signature__ attribute
```

sadra-barikbin:
Fixes #3262

Description:
To exclude special attributes from causing `MetricsLambda` creation when they're accessed.

Check list:

### DIFF
--- a/ignite/metrics/metric.py
+++ b/ignite/metrics/metric.py
@@ -768,6 +768,9 @@ class Metric(Serializable, metaclass=ABCMeta):
     def __getattr__(self, attr: str) -> Callable:
         from ignite.metrics.metrics_lambda import MetricsLambda
 
+        if attr.startswith("__") and attr.endswith("__"):
+            return object.__getattribute__(self, attr)
+
         def fn(x: Metric, *args: Any, **kwargs: Any) -> Any:
             return getattr(x, attr)(*args, **kwargs)
 
--- a/tests/ignite/metrics/test_metric.py
+++ b/tests/ignite/metrics/test_metric.py
@@ -1446,3 +1446,25 @@ def test_skip_unrolling():
     state = State(output=(y_pred, y_true))
     engine = MagicMock(state=state)
     metric.iteration_completed(engine)
+
+
+class DummyMetric6(Metric):
+    def reset(self):
+        pass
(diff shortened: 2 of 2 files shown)


vfdev-5 (approved):
Hey Sadra, sounds reasonable, thanks for the fix!
And welcome back!

vfdev-5:
Let's add a simple test for that

PR #3273, opened 2024-08-02
Fix WandBLogger init

### LINKED ISSUE #3272: Initialization of WandBLogger with `init=True` fails
`wandb.init` function does not have 'init' keyword argument.
Therefore, [this code](https://github.com/pytorch/ignite/blob/4c93282385353b585b4536c4553589af7dd58af0/ignite/handlers/wandb_logger.py#L136-L137) fails:
```
        if kwargs.get("init", True):
            wandb.init(*args, **kwargs)
```
Solution is to pop 'init' key from kwargs.

zubatyuk:
Fixes #3272

### DIFF
--- a/ignite/handlers/wandb_logger.py
+++ b/ignite/handlers/wandb_logger.py
@@ -134,7 +134,7 @@ class WandBLogger(BaseLogger):
                 "You man install wandb with the command:\n pip install wandb\n"
             )
         if kwargs.get("init", True):
-            kwargs.pop('init')
+            kwargs.pop('init', None)
             wandb.init(*args, **kwargs)
 
     def __getattr__(self, attr: Any) -> Any:
--- a/ignite/handlers/wandb_logger.py
+++ b/ignite/handlers/wandb_logger.py
@@ -134,6 +134,7 @@ class WandBLogger(BaseLogger):
                 "You man install wandb with the command:\n pip install wandb\n"
             )
         if kwargs.get("init", True):
+            kwargs.pop('init')
             wandb.init(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


ignite/handlers/wandb_logger.py
```diff
@@ -134,6 +134,7 @@ def __init__(self, *args: Any, **kwargs: Any):
                 "You man install wandb with the command:\n pip install wandb\n"
             )
         if kwargs.get("init", True):
+            kwargs.pop('init')
```
vfdev-5, line 137: This may not always work, by default this code will give the error:
```python
KeyError: 'init'
```
as `kwargs` may not have "init" kwargs.

Let's do that in the following way:
```python
if kwargs.get("init", True):
    kwargs.pop("init", None)
    wandb.init(*args, **kwargs)
```

zubatyuk: agree. updated.

ignite/handlers/wandb_logger.py
```diff
@@ -134,6 +134,7 @@ def __init__(self, *args: Any, **kwargs: Any):
                 "You man install wandb with the command:\n pip install wandb\n"
             )
         if kwargs.get("init", True):
+            kwargs.pop('init', None)
```
vfdev-5, line 137: Black formatting requires double quotes instead of single quotes, please change it

zubatyuk: ok

vfdev-5 (approved):
LGTM, thanks for the PR, @zubatyuk !

PR #3282, opened 2024-09-08
add HSIC metric

kzkadc:
Description: added Hilbert-Schmidt Independence Criterion (HSIC).

HSIC is a kernel-based method to measure the statistical independence between two feature vectors:

https://papers.nips.cc/paper_files/paper/2007/hash/d5cfead94f5350c12c322b5b664544c1-Abstract.html
https://jejjohnson.github.io/research_journal/appendix/similarity/hsic/

This metric computes the unbiased estimator of the HSIC proposed in the following:

https://jmlr.csail.mit.edu/papers/v13/song12a.html

Check list:

- [x] New tests are added (if a new feature is added)
- [x] New doc strings: description and/or example code are in RST format
- [x] Documentation is updated (if required)

PR #3284, opened 2024-09-10
Modify docstring and argument for rank correlation metrics

kzkadc:
Fixes https://github.com/pytorch/ignite/pull/3282#discussion_r1749765351

Description:

Check list:

### DIFF
--- a/ignite/metrics/regression/spearman_correlation.py
+++ b/ignite/metrics/regression/spearman_correlation.py
@@ -80,7 +80,6 @@ class SpearmanRankCorrelation(EpochMetric):
     def __init__(
         self,
         output_transform: Callable[..., Any] = lambda x: x,
-        check_compute_fn: bool = True,
         device: Union[str, torch.device] = torch.device("cpu"),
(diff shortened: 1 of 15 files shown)


ignite/metrics/regression/spearman_correlation.py
```diff
@@ -84,7 +88,7 @@ def __init__(
         except ImportError:
             raise ModuleNotFoundError("This module requires scipy to be installed.")
 
-        super().__init__(_get_spearman_r(), output_transform, check_compute_fn, device, skip_unrolling)
+        super().__init__(_get_spearman_r(), output_transform, True, device, skip_unrolling)
```
vfdev-5, line 91: It might be better to add `check_compute_fn` to the docstring than hard-coding the value.
Alternatively, we could replace in the docstring and the constructor all the inherited args with `*args, **kwargs` and say that they are from EpochMetric...

vfdev-5 (approved):
Thanks @kzkadc !

PR #3285, opened 2024-09-13
Use the default runner type

ZainRizvi:
During the Linux Amazon 2023 AMI (base OS) migration these jobs had been temporarily shifted to use the new experimental AMI (see https://github.com/pytorch/ignite/pull/3265)

Now that the new AMI has become the base AMI, shifting these back to use the default runners


vfdev-5 (approved):
Thanks @ZainRizvi , lgtm!

PR #3290, opened 2024-10-12
Add clustering metrics

kzkadc:
Description: added metrics for evaluating clustering results implemented in scikit-learn:

- [silhouette score](https://en.wikipedia.org/wiki/Silhouette_(clustering))
- [Davies-Bouldin score](https://en.wikipedia.org/wiki/Davies%E2%80%93Bouldin_index)
- [Calinski-Harabasz score](https://en.wikipedia.org/wiki/Calinski%E2%80%93Harabasz_index)

Check list:

### DIFF
--- a/ignite/metrics/clustering/calinski_harabasz_score.py
+++ b/ignite/metrics/clustering/calinski_harabasz_score.py
@@ -35,7 +35,7 @@ class CalinskiHarabaszScore(_ClusteringMetricBase):
     <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.calinski_harabasz_score.html>`_.
 
     - ``update`` must receive output of the form ``(features, labels)``
-       or ``{'features': features, 'labels': labels}``.
+      or ``{'features': features, 'labels': labels}``.
     - `features` and `labels` must be of same shape `(B, D)` and `(B,)`.
 
     Parameters are inherited from ``EpochMetric.__init__``.
@@ -47,6 +47,9 @@ class CalinskiHarabaszScore(_ClusteringMetricBase):
             you want to compute the metric with respect to one of the outputs.
             By default, metrics require the output as ``(features, labels)``
             or ``{'features': features, 'labels': labels}``.
+        check_compute_fn: if True, ``compute_fn`` is run on the first batch of data to ensure there are no
+            issues. If issues exist, user is warned that there might be an issue with the ``compute_fn``.
+            Default, True.
         device: specifies which device updates are accumulated on. Setting the
             metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
             non-blocking. By default, CPU.
--- a/ignite/metrics/clustering/calinski_harabasz_score.py
+++ b/ignite/metrics/clustering/calinski_harabasz_score.py
@@ -5,6 +5,8 @@ from torch import Tensor
 
 from ignite.metrics.clustering._base import _ClusteringMetricBase
 
+__all__ = ["CalinskiHarabaszScore"]
+
 
 def _calinski_harabasz_score(features: Tensor, labels: Tensor) -> float:
     from sklearn.metrics import calinski_harabasz_score
--- a/ignite/metrics/clustering/calinski_harabasz_score.py
+++ b/ignite/metrics/clustering/calinski_harabasz_score.py
@@ -6,13 +6,13 @@ from torch import Tensor
 from ignite.metrics.clustering._base import _ClusteringMetricBase
 
 
-def _get_calinski_harabasz_score(**kwargs: Any) -> Callable[[Tensor, Tensor], float]:
+def _get_calinski_harabasz_score() -> Callable[[Tensor, Tensor], float]:
     from sklearn.metrics import calinski_harabasz_score
 
     def _calinski_harabasz_score(features: Tensor, labels: Tensor) -> float:
         np_features = features.numpy()
         np_labels = labels.numpy()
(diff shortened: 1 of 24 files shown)


ignite/metrics/clustering/calinski_harabasz_score.py
```diff
@@ -0,0 +1,107 @@
...
+        self,
+        output_transform: Callable[..., Any] = lambda x: x,
+        check_compute_fn: bool = True,
+        device: Union[str, torch.device] = torch.device("cpu"),
+        skip_unrolling: bool = False,
+    ) -> None:
+        try:
+            from sklearn.metrics import calinski_harabasz_score  # noqa: F401
+        except ImportError:
+            raise ModuleNotFoundError("This module requires scikit-learn to be installed.")
```
vfdev-5, line 105: Checking for the required package in init is a good idea!
Especially, that we may need to do the same for other epoch-like metrics: https://github.com/pytorch/ignite/blob/master/ignite/metrics/precision_recall_curve.py etc

ignite/metrics/clustering/calinski_harabasz_score.py
```diff
@@ -0,0 +1,107 @@
...
+def _get_calinski_harabasz_score(**kwargs: Any) -> Callable[[Tensor, Tensor], float]:
+    from sklearn.metrics import calinski_harabasz_score
+
+    def _calinski_harabasz_score(features: Tensor, labels: Tensor) -> float:
+        np_features = features.numpy()
+        np_labels = labels.numpy()
+        score = calinski_harabasz_score(np_features, np_labels, **kwargs)
+        return score
+
+    return _calinski_harabasz_score
```
vfdev-5, line 18: Concerning this way of importing once and returning the computation inner method, I understand the purpose but we can follow : https://github.com/pytorch/ignite/blob/302c707f79dbef50a3920baf11b76eba5ee4200e/ignite/metrics/average_precision.py#L8-L13

I think it is not a big deal with the import called multiple times. What do you think?

ignite/metrics/clustering/calinski_harabasz_score.py
vfdev-5: By the way, I saw that in Kohen Kappa implementation you were doing: 
https://github.com/pytorch/ignite/blob/302c707f79dbef50a3920baf11b76eba5ee4200e/ignite/metrics/cohen_kappa.py#L74C9-L93

Can we pick a single pattern for example like in https://github.com/pytorch/ignite/blob/master/ignite/metrics/average_precision.py and apply it everywhere?
If we agree with that can you please send a separate PR for Kohen Kappa fix please

vfdev-5 (commented):
Thanks for the PR, @kzkadc !
I left few comments, let me know what you think

ignite/metrics/clustering/calinski_harabasz_score.py
kzkadc: I think so too.

I compared the two ways of import and found that the difference is negligible:

```python
import time

import numpy as np


def get_metric():
    from sklearn.metrics import accuracy_score

    def _accuracy_score(y_pred, y):
        score = accuracy_score(y, y_pred)
        return score

    return _accuracy_score


start = time.perf_counter()

acc = get_metric()

for _ in range(100000):
    y_pred = np.random.randint(10, size=100)
    y = np.random.randint(10, size=100)
    acc(y_pred, y)

end = time.perf_counter()
print(end - start)
```

```python
import time

import numpy as np


def acc_score(y_pred, y):
    from sklearn.metrics import accuracy_score

    score = accuracy_score(y, y_pred)
    return score


start = time.perf_counter()

for _ in range(100000):
    y_pred = np.random.randint(10, size=100)
    y = np.random.randint(10, size=100)
    acc_score(y_pred, y)

end = time.perf_counter()
print(end - start)
```

ignite/metrics/clustering/silhouette_score.py
```diff
@@ -0,0 +1,121 @@
...
+
+import torch
+from torch import Tensor
+
+from ignite.metrics.clustering._base import _ClusteringMetricBase
+
+__all__ = ["SilhouetteScore"]
+
+
+def _get_silhouette_score(**kwargs: Any) -> Callable[[Tensor, Tensor], float]:
```
vfdev-5, line 11: Can you please change this function as well

kzkadc: Thanks for review.

I left this function in the previous style for retaining extra kwargs for sklearn's silhouette score function.

Should we make kwargs as a member of the metric class as in [Cohen Kappa's weight argument](https://github.com/pytorch/ignite/blob/a5d3464231f37b86fe4e5ae272fe14bbcb33b27d/ignite/metrics/cohen_kappa.py#L71-L87)?

vfdev-5: I see, thanks for pointing out this. Let's use one approach everywhere, I let you choose either this one or CohenKappa's one

kzkadc: I'll take the CohenKappa's style for simplicity.

vfdev-5:
@kzkadc can you please also update this metric implementation https://github.com/pytorch/ignite/blob/master/ignite/metrics/precision_recall_curve.py according to the way we are now doing everywhere else. Thanks!

kzkadc:
@vfdev-5 could you recheck this PR again?

PR #3295, opened 2024-10-17
Add logging of mappings in BaseLogger

### LINKED ISSUE #3294: Add the logging of dict metrics
## 🚀 Feature

The request would be to add the logging of `Mapping` metrics in the logging framework.

The `ignite.metrics.Metric` class supports the use of `Mapping` metrics as we can see below. However, the `BaseOutputHandler` does not support dictionary metrics and warns about them.

https://github.com/pytorch/ignite/blob/edd5025e7d597a6e5fe45c5173487c37d3f9d1df/ignite/metrics/metric.py#L488-L494

Ones can simply ask the logger to report the metric names produced by the `Metric` directly as those will be store in the metric state no matter which name was used for the metric. But I feel like it breaks the kind of "namespaces" that seems to be used in loggers.

I would find it practical if the logger could handle mappings and log their content as sub values of the metric itself.

This could be achieved by editing the `BaseOutputHandler` which would fix this issue in any existing logger. There should not be any side effect as the logger was warning users if using mappings, so I imagine very few users were already having a mapping metrics logged that would now appear if they upgraded to a version with this feature.

nowtryz:
Fixes #3294

Description:
Adds support for `Mapping` metrics in loggers

Check list:

### DIFF
--- a/ignite/handlers/base_logger.py
+++ b/ignite/handlers/base_logger.py
@@ -161,11 +161,20 @@ class BaseOutputHandler(BaseHandler):
             elif isinstance(value, torch.Tensor) and value.ndimension() == 1:
                 for i, v in enumerate(value):
                     metrics_state_attrs_dict[key_tf(self.tag, name, str(i))] = v.item()
+            elif isinstance(value, Mapping):
+                for key, sub_value in value.items():
+                    if isinstance(value, numbers.Number):
+                        metrics_state_attrs_dict[key_tf(self.tag, name, key)] = value
+                    elif isinstance(value, torch.Tensor) and value.ndimension() == 0:
+                        metrics_state_attrs_dict[key_tf(self.tag, name, key)] = value.item()
+                    elif isinstance(value, torch.Tensor) and value.ndimension() == 1:
+                        for i, v in enumerate(value):
+                            metrics_state_attrs_dict[
+                                key_tf(self.tag, name, key, str(i))] = v.item()
+            elif isinstance(value, str) and log_text:
+                metrics_state_attrs_dict[key_tf(self.tag, name)] = value
             else:
-                if isinstance(value, str) and log_text:
-                    metrics_state_attrs_dict[key_tf(self.tag, name)] = value
-                else:
-                    warnings.warn(f"Logger output_handler can not log metrics value type {type(value)}")
+                warnings.warn(f"Logger output_handler can not log metrics value type {type(value)}")
         return metrics_state_attrs_dict
 
 
--- a/ignite/handlers/base_logger.py
+++ b/ignite/handlers/base_logger.py
@@ -4,7 +4,9 @@ import numbers
 import warnings
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
-from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
+from collections.abc import Mapping, Collection
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union, TypeVar, \
+    Iterable
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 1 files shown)


nowtryz:
Before testing this, I would like to be sure the feature is of interest.

ignite/handlers/base_logger.py
```diff
@@ -161,11 +161,20 @@ def key_str_tf(tag: str, name: str, *args: str) -> str:
...
+                    if isinstance(value, numbers.Number):
+                        metrics_state_attrs_dict[key_tf(self.tag, name, key)] = value
+                    elif isinstance(value, torch.Tensor) and value.ndimension() == 0:
+                        metrics_state_attrs_dict[key_tf(self.tag, name, key)] = value.item()
+                    elif isinstance(value, torch.Tensor) and value.ndimension() == 1:
+                        for i, v in enumerate(value):
+                            metrics_state_attrs_dict[
+                                key_tf(self.tag, name, key, str(i))] = v.item()
+            elif isinstance(value, str) and log_text:
+                metrics_state_attrs_dict[key_tf(self.tag, name)] = value
```
vfdev-5, line 175: Yes, this makes sense to add the support, thanks!
I would implement that however a bit differently using a recursive function call, something like here: 
https://github.com/pytorch/ignite/blob/8782ca45204a083b3150271895751fd8fe0046d6/ignite/utils.py#L58-L80
This way we would avoid duplicating the code.

nowtryz: Yes, I considered a form of recursive behavior, but I don't know how deep you would allow such feature? If there is no limit, no problem!

vfdev-5: Mainly, the idea is to remove duplicated code: 
```python
                    if isinstance(value, numbers.Number):
                        metrics_state_attrs_dict[key_tf(self.tag, name, key)] = value
                    elif isinstance(value, torch.Tensor) and value.ndimension() == 0:
                        metrics_state_attrs_dict[key_tf(self.tag, name, key)] = value.item()
                    elif isinstance(value, torch.Tensor) and value.ndimension() == 1:
                        for i, v in enumerate(value):
                            metrics_state_attrs_dict[
                                key_tf(self.tag, name, key, str(i))] = v.item()
```

nowtryz: I'll look into that

nowtryz:
@vfdev-5 is it ok for you like that? If so, I'll fix the linting and adapt the tests

ignite/handlers/base_logger.py
```diff
@@ -130,43 +132,72 @@ def _setup_output_metrics_state_attrs(
...
+        self._compute_tags(
+            concat_tf,
+            log_text,
+            node=state_metrics.items(),
+            parent_tag=(self.tag, ) if key_tuple else self.tag,
+            dest_dict=metrics)
+        return metrics
+
+    @classmethod
+    def _compute_tags(
```
vfdev-5, line 167: You can make this just a private helper method in this submodule, not necessarily a classmethod.

ignite/handlers/base_logger.py
```diff
@@ -130,43 +132,72 @@ def _setup_output_metrics_state_attrs(
...
+            cls,
+            concat_tf: Callable[..., Union[str, Tuple[str, ...]]],
+            log_text: bool, node: Iterable[Tuple[str, Any]],
+            parent_tag: Union[str, Tuple[str, ...]],
+            dest_dict: Dict[Any, Union[str, float, numbers.Number]]
```
vfdev-5, line 172: I'm not keen of passing `dest_dict` as the argument, but maybe we can't do anything better here...

vfdev-5 (commented):
Thanks @nowtryz , this is going in a good direction, let's try to improve a bit more the code.

PR #3302, opened 2024-11-07
More idist allreduce/gather test fixes

vfdev-5:
Description:
- fixed all_reduce with group implementation
- fixed and rewrote tests: _test_distrib_all_reduce_group, _test_distrib_all_gather_group, etc

Based on https://github.com/pytorch/ignite/pull/3301

### DIFF
--- a/ignite/distributed/comp_models/native.py
+++ b/ignite/distributed/comp_models/native.py
@@ -418,6 +418,9 @@ if has_native_dist_support:
         }
 
         def _do_all_reduce(self, tensor: torch.Tensor, op: str = "SUM", group: Optional[Any] = None) -> torch.Tensor:
+            if group == dist.GroupMember.NON_GROUP_MEMBER:
+                return tensor
+
             if op not in self._reduce_op_map:
                 raise ValueError(f"Unsupported reduction operation: '{op}'")
             if group is not None and not isinstance(group, dist.ProcessGroup):
--- a/ignite/metrics/clustering/calinski_harabasz_score.py
+++ b/ignite/metrics/clustering/calinski_harabasz_score.py
@@ -11,8 +11,8 @@ __all__ = ["CalinskiHarabaszScore"]
 def _calinski_harabasz_score(features: Tensor, labels: Tensor) -> float:
     from sklearn.metrics import calinski_harabasz_score
 
-    np_features = features.numpy()
-    np_labels = labels.numpy()
+    np_features = features.cpu().numpy()
+    np_labels = labels.cpu().numpy()
     score = calinski_harabasz_score(np_features, np_labels)
     return score
 
--- a/ignite/metrics/clustering/davies_bouldin_score.py
+++ b/ignite/metrics/clustering/davies_bouldin_score.py
@@ -11,8 +11,8 @@ __all__ = ["DaviesBouldinScore"]
 def _davies_bouldin_score(features: Tensor, labels: Tensor) -> float:
     from sklearn.metrics import davies_bouldin_score
 
-    np_features = features.numpy()
-    np_labels = labels.numpy()
+    np_features = features.cpu().numpy()
+    np_labels = labels.cpu().numpy()
     score = davies_bouldin_score(np_features, np_labels)
     return score
 
--- a/ignite/metrics/clustering/silhouette_score.py
+++ b/ignite/metrics/clustering/silhouette_score.py
@@ -111,7 +111,7 @@ class SilhouetteScore(_ClusteringMetricBase):
     def _silhouette_score(self, features: Tensor, labels: Tensor) -> float:
         from sklearn.metrics import silhouette_score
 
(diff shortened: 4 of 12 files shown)


sadra-barikbin:
@vfdev-5 , it seems that HVD test `` fails because `new_group()` does not return a `ProcessGroup.NON_GROUP_MEMBER` in HVD model in case the rank does not belong to the group, although `dist.new_group` does.

https://github.com/pytorch/ignite/blob/ed934ef0ddb6efd93aac89e6b5b823d481f76306/ignite/distributed/utils.py#L387-L388

https://github.com/pytorch/ignite/blob/ed934ef0ddb6efd93aac89e6b5b823d481f76306/ignite/distributed/comp_models/horovod.py#L201-L202

By the way doesn't HVD really support `all_gather` with process groups?

https://github.com/horovod/horovod/blob/80780e10d97271472fc6f08cab630ad78d9c5aeb/horovod/torch/mpi_ops.py#L652

vfdev-5:
> By the way doesn't HVD really support all_gather with process groups?

We raise an error for now if a group is passed as arg: 
https://github.com/pytorch/ignite/blob/8255a0fbc501c0c2e99da9c335880cc68a968d4b/ignite/distributed/comp_models/horovod.py#L167-L168


One error in the CI is : 
```
[0]<stderr>:  File "/work/tests/ignite/distributed/utils/__init__.py", line 347, in _test_idist_all_gather_tensors_with_shapes_group
[0]<stderr>:    tensors = all_gather_tensors_with_shapes(rank_tensor, [[r + 2, r + 3, r + 4] for r in ranks], ranks)
[0]<stderr>:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[0]<stderr>:  File "/work/ignite/distributed/utils.py", line 395, in all_gather_tensors_with_shapes
[0]<stderr>:    padded_tensor = torch.nn.functional.pad(
[0]<stderr>:                    ^^^^^^^^^^^^^^^^^^^^^^^^
[0]<stderr>:  File "/opt/conda/lib/python3.11/site-packages/torch/nn/functional.py", line 5096, in pad
[0]<stderr>:    return torch._C._nn.pad(input, pad, mode, value)
[0]<stderr>:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[0]<stderr>:RuntimeError: Padding length should be less than or equal to two times the input dimension but got padding length 6 and input of dimension 1
```
and we check it with 
```
            with pytest.raises(NotImplementedError, match=r"all_gather with group for horovod is not implemented"):
                tensors = all_gather_tensors_with_shapes(rank_tensor, [[r + 2, r + 3, r + 4] for r in ranks], ranks)
```
Maybe there is something else is going on ?

sadra-barikbin:
In `all_gather_tensors_with_shapes`, first a group is created, if the arg is given, which is equal to `ProcessGroup.NON_GROUP_MEMBER` for non-participating processes. But for HVD, the resulting group is a `ProcessSet` anyway. When padding is being done, which takes places before `allgather`, it tries to pad the 1-dim input tensor which is `torch.tensor([NON_PARITIPATING_RANK])` with the padding size argument of 3 dimensions (3 * 2 (for left and right padding)) and raises error.

sadra-barikbin:
I mean the below  if statement is not enough for horovod:
https://github.com/pytorch/ignite/blob/ed934ef0ddb6efd93aac89e6b5b823d481f76306/ignite/distributed/utils.py#L390-L391

vfdev-5:
> I mean the below if statement is not enough for horovod as the resulting group is always a `ProcessSet`
> 
> https://github.com/pytorch/ignite/blob/ed934ef0ddb6efd93aac89e6b5b823d481f76306/ignite/distributed/utils.py#L390-L391

actually, this is weird why we use specific `dist.GroupMember.NON_GROUP_MEMBER` in the generic `all_gather_tensors_with_shapes` method...

sadra-barikbin:
So we could replace it with something like this maybe:

```python
# In idist:

def is_in_group(group):
    if group is dist.GroupMember.NON_GROUP_MEMBER:
        return False
    if isinstance(group, hvd.ProcessSet):
        return group.included()
    return True
```

vfdev-5:
> So we could replace it with something like this maybe:
> 
> ```python
> # In idist:
> 
> def is_in_group(group):
>     if group is dist.GroupMember.NON_GROUP_MEMBER:
>         return False
>     if isinstance(group, hvd.ProcessSet):
>         return group.included()
>     return True
> ```

Yes, this makes sense. I would just make it private and call like in pytorch: `_rank_not_in_group`. 
Sadra, can you please send a PR with this update?

vfdev-5:
I think we can close this PR

Issue #3308, opened 2024-12-02
Add flag to `Engine.terminate(skip_complete_event: bool = False)`


vfdev-5:
## 🚀 Feature

Add flag to `Engine.terminate(skip_complete_event: bool = False)` to optionally skip `Events.COMPLETE` when terminate was called.

Context:
- https://discuss.pytorch.org/t/average-time-per-epoch-calculation-on-pytorch-ignite/82506/10?u=vfdev-5

bonassifabio:
Hi, I tried to implement this in the pull request #3309!

vfdev-5:
This issue can be closed as completed by #3309 PR

PR #3309, opened 2024-12-02
Give the option to terminate the engine without firing Events.COMPLET…

bonassifabio:
As per [this Pytorch discussion post](https://discuss.pytorch.org/t/average-time-per-epoch-calculation-on-pytorch-ignite/82506/11?u=gnufabio), This pull request aims to allow the user to terminate an engine (via `engine.terminate()`) without consequently firing `Events.COMPLETED`. This is done by passing the optional argument `skip_event_completed=True` to the method `terminate()`.

While the default behavior is still to fire `Events.COMPLETED` after termination, it is useful to allow the user to do otherwise. This might be useful, for example, when one of handlers of `Events.COMPLETED` needs to be executed only upon graceful termination of the Engine.

Alternative solutions to the one proposed in this PR could be:
-  include a flag `has_been_terminated` that is set to True when `Events.TERMINATE` is triggered
-  include a new filter to `Events.COMPLETED`, like `Events.COMPLETED(terminated=True)`, which would be perhaps the most "ignitic" way to add this functionality.

### Other changes:

Fixed the table in the [Events docs](https://pytorch.org/ignite/generated/ignite.engine.events.Events.html).

### Related issues:

#3308 

### Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -618,7 +618,11 @@ class Engine(Serializable):
                     4 34  |  9
                     2 Engine ended the run at  4 34
 
-        .. versionchanged:: 0.5.1
+        .. versionchanged:: 0.4.10
+            Behaviour changed, for details see https://github.com/pytorch/ignite/issues/2669
+
+        .. versionchanged:: 0.5.2
+            Added `skip_event_completed` flag
         """
         self.logger.info("Terminate signaled. Engine will stop after current iteration is finished.")
         self.should_terminate = True
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -550,7 +550,7 @@ class Engine(Serializable):
 
         Args:
             skip_event_completed: if True, the event `~ignite.engine.events.Events.COMPLETED` is not fired after
-            `~ignite.engine.events.Events.COMPLETED`. Default is False.
+            `~ignite.engine.events.Events.TERMINATE`. Default is False.
 
         Examples:
             .. testcode::
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -4,7 +4,7 @@ import math
 import time
 import warnings
 import weakref
-from collections import defaultdict, OrderedDict
+from collections import OrderedDict, defaultdict
 from collections.abc import Mapping
 from typing import Any, Callable, Dict, Generator, Iterable, Iterator, List, Optional, Tuple, Union
 
@@ -140,6 +140,7 @@ class Engine(Serializable):
         self._process_function = process_function
         self.last_event_name: Optional[Events] = None
         self.should_terminate = False
+        self.skip_completed_after_termination = False
         self.should_terminate_single_epoch = False
         self.should_interrupt = False
         self.state = State()
@@ -538,7 +539,7 @@ class Engine(Serializable):
         self.logger.info("interrupt signaled. Engine will interrupt the run after current iteration is finished.")
         self.should_interrupt = True
 
-    def terminate(self) -> None:
+    def terminate(self, skip_event_completed: bool = False) -> None:
         """Sends terminate signal to the engine, so that it terminates completely the run. The run is
         terminated after the event on which ``terminate`` method was called. The following events are triggered:
 
@@ -547,6 +548,9 @@ class Engine(Serializable):
         - :attr:`~ignite.engine.events.Events.TERMINATE`
         - :attr:`~ignite.engine.events.Events.COMPLETED`
 
+        Args:
+            skip_event_completed: if True, the event `~ignite.engine.events.Events.COMPLETED` is not fired after
+            `~ignite.engine.events.Events.COMPLETED`. Default is False.
 
         Examples:
             .. testcode::
@@ -614,12 +618,11 @@ class Engine(Serializable):
                     4 34  |  9
                     2 Engine ended the run at  4 34
 
-        .. versionchanged:: 0.4.10
-            Behaviour changed, for details see https://github.com/pytorch/ignite/issues/2669
-
+        .. versionchanged:: 0.5.1
         """
         self.logger.info("Terminate signaled. Engine will stop after current iteration is finished.")
         self.should_terminate = True
+        self.skip_completed_after_termination = skip_event_completed
 
     def terminate_epoch(self) -> None:
         """Sends terminate signal to the engine, so that it terminates the current epoch. The run
@@ -898,6 +901,8 @@ class Engine(Serializable):
                 # If engine was terminated and now is resuming from terminated state
                 # we need to initialize iter_counter as 0
                 self._init_iter = 0
+                # And we reset the skip_completed_after_termination to its default value
+                self.skip_completed_after_termination = False
 
         if self._dataloader_iter is None:
             self.state.dataloader = data
@@ -994,7 +999,11 @@ class Engine(Serializable):
             # time is available for handlers but must be updated after fire
             self.state.times[Events.COMPLETED.name] = time_taken
             handlers_start_time = time.time()
-            self._fire_event(Events.COMPLETED)
+
+            # do not fire Events.COMPLETED if we terminated the run with flag `skip_completed_after_termination=True`
+            if self.should_terminate and not self.skip_completed_after_termination or not self.should_terminate:
+                self._fire_event(Events.COMPLETED)
+
             time_taken += time.time() - handlers_start_time
             # update time wrt handlers
             self.state.times[Events.COMPLETED.name] = time_taken
@@ -1175,7 +1184,11 @@ class Engine(Serializable):
             # time is available for handlers but must be updated after fire
             self.state.times[Events.COMPLETED.name] = time_taken
             handlers_start_time = time.time()
-            self._fire_event(Events.COMPLETED)
+
+            # do not fire Events.COMPLETED if we terminated the run with flag `skip_completed_after_termination=True`
+            if self.should_terminate and not self.skip_completed_after_termination or not self.should_terminate:
+                self._fire_event(Events.COMPLETED)
+                
             time_taken += time.time() - handlers_start_time
             # update time wrt handlers
             self.state.times[Events.COMPLETED.name] = time_taken
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -550,7 +550,7 @@ class Engine(Serializable):
 
         Args:
             skip_completed: if True, the event :attr:`~ignite.engine.events.Events.COMPLETED` is not fired after
-            :attr:`~ignite.engine.events.Events.TERMINATE`. Default is False.
+                :attr:`~ignite.engine.events.Events.TERMINATE`. Default is False.
 
         Examples:
             .. testcode::
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -1001,18 +1001,16 @@ class Engine(Serializable):
             # time is available for handlers but must be updated after fire
             self.state.times[Events.COMPLETED.name] = time_taken
 
-            if self.should_terminate and self.skip_completed_after_termination:
-                # do not fire Events.COMPLETED if we terminated the run with flag `skip_completed_after_termination=True`
-                hours, mins, secs = _to_hours_mins_secs(time_taken)
-                self.logger.info(f"Engine run terminated. Time taken: {hours:02d}:{mins:02d}:{secs:06.3f}")
-            else:
+            # do not fire Events.COMPLETED if we terminated the run with flag `skip_completed=True`
+            if not (self.should_terminate and self.skip_completed_after_termination):
                 handlers_start_time = time.time()
                 self._fire_event(Events.COMPLETED)
                 time_taken += time.time() - handlers_start_time
                 # update time wrt handlers
                 self.state.times[Events.COMPLETED.name] = time_taken
-                hours, mins, secs = _to_hours_mins_secs(time_taken)
-                self.logger.info(f"Engine run completed. Time taken: {hours:02d}:{mins:02d}:{secs:06.3f}")
+
+            hours, mins, secs = _to_hours_mins_secs(time_taken)
+            self.logger.info(f"Engine run finished. Time taken: {hours:02d}:{mins:02d}:{secs:06.3f}")
 
         except BaseException as e:
             self._dataloader_iter = None
@@ -1187,17 +1185,17 @@ class Engine(Serializable):
             time_taken = time.time() - start_time
             # time is available for handlers but must be updated after fire
             self.state.times[Events.COMPLETED.name] = time_taken
-            handlers_start_time = time.time()
 
-            # do not fire Events.COMPLETED if we terminated the run with flag `skip_completed_after_termination=True`
-            if self.should_terminate and not self.skip_completed_after_termination or not self.should_terminate:
+            # do not fire Events.COMPLETED if we terminated the run with flag `skip_completed=True`
+            if not (self.should_terminate and self.skip_completed_after_termination):
+                handlers_start_time = time.time()
                 self._fire_event(Events.COMPLETED)
+                time_taken += time.time() - handlers_start_time
+                # update time wrt handlers
+                self.state.times[Events.COMPLETED.name] = time_taken
 
-            time_taken += time.time() - handlers_start_time
-            # update time wrt handlers
-            self.state.times[Events.COMPLETED.name] = time_taken
             hours, mins, secs = _to_hours_mins_secs(time_taken)
-            self.logger.info(f"Engine run complete. Time taken: {hours:02d}:{mins:02d}:{secs:06.3f}")
+            self.logger.info(f"Engine run finished. Time taken: {hours:02d}:{mins:02d}:{secs:06.3f}")
 
         except BaseException as e:
             self._dataloader_iter = None
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -626,7 +626,7 @@ class Engine(Serializable):
         """
         self.logger.info("Terminate signaled. Engine will stop after current iteration is finished.")
         self.should_terminate = True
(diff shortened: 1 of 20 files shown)


bonassifabio:
This PR addresses issue #3308

ignite/engine/engine.py
```diff
@@ -547,6 +548,9 @@ def terminate(self) -> None:
         - :attr:`~ignite.engine.events.Events.TERMINATE`
         - :attr:`~ignite.engine.events.Events.COMPLETED`
 
+        Args:
+            skip_event_completed: if True, the event `~ignite.engine.events.Events.COMPLETED` is not fired after
+            `~ignite.engine.events.Events.COMPLETED`. Default is False.
```
vfdev-5, line 553:
```suggestion
            `~ignite.engine.events.Events.TERMINATE`. Default is False.
```
You meant terminate event, right ?

ignite/engine/engine.py
```diff
@@ -614,12 +618,11 @@ def terminate():
                     4 34  |  9
                     2 Engine ended the run at  4 34
 
-        .. versionchanged:: 0.4.10
-            Behaviour changed, for details see https://github.com/pytorch/ignite/issues/2669
-
+        .. versionchanged:: 0.5.1
```
vfdev-5, line 621: We should keep the previous tag and add the new one.
```suggestion
        .. versionchanged:: 0.4.10
            Behaviour changed, for details see https://github.com/pytorch/ignite/issues/2669

        .. versionchanged:: 0.5.2
            Added `skip_event_completed` flag
```

ignite/engine/engine.py
```diff
@@ -538,7 +539,7 @@ def call_interrupt():
         self.logger.info("interrupt signaled. Engine will interrupt the run after current iteration is finished.")
         self.should_interrupt = True
 
-    def terminate(self) -> None:
+    def terminate(self, skip_event_completed: bool = False) -> None:
```
vfdev-5, line 542: I like the name of the `self.skip_completed_after_termination`, let's call this flag in the same way:
```suggestion
    def terminate(self, skip_completed: bool = False) -> None:
```

ignite/engine/engine.py
```diff
@@ -614,12 +618,11 @@ def terminate():
...
                     2 Engine ended the run at  4 34
 
-        .. versionchanged:: 0.4.10
-            Behaviour changed, for details see https://github.com/pytorch/ignite/issues/2669
-
+        .. versionchanged:: 0.5.1
         """
         self.logger.info("Terminate signaled. Engine will stop after current iteration is finished.")
         self.should_terminate = True
+        self.skip_completed_after_termination = skip_event_completed
```
vfdev-5, line 625:
```suggestion
        self.skip_completed_after_termination = skip_completed
```

ignite/engine/engine.py
```diff
@@ -898,6 +901,8 @@ def switch_batch(engine):
                 # If engine was terminated and now is resuming from terminated state
                 # we need to initialize iter_counter as 0
                 self._init_iter = 0
+                # And we reset the skip_completed_after_termination to its default value
+                self.skip_completed_after_termination = False
```
vfdev-5, line 905: I wonder whether we really need to reset the flag here?

ignite/engine/engine.py
```diff
@@ -994,7 +999,11 @@ def _internal_run_as_gen(self) -> Generator[Any, None, State]:
             # time is available for handlers but must be updated after fire
             self.state.times[Events.COMPLETED.name] = time_taken
             handlers_start_time = time.time()
-            self._fire_event(Events.COMPLETED)
+
+            # do not fire Events.COMPLETED if we terminated the run with flag `skip_completed_after_termination=True`
+            if self.should_terminate and not self.skip_completed_after_termination or not self.should_terminate:
+                self._fire_event(Events.COMPLETED)
```
vfdev-5, line 1005: I think this condition should be:
```suggestion
            if not (self.should_terminate and self.skip_completed_after_termination):
                self._fire_event(Events.COMPLETED)
```

Maybe, all the code below that updates `self.state.times[Events.COMPLETED.name] = time_taken` can be inside the if clause

ignite/engine/events.py
```diff
@@ -265,7 +265,8 @@ class Events(EventEnum):
 
     - EPOCH_COMPLETED : triggered when the epoch is ended. Note that this is triggered even
       when :meth:`~ignite.engine.engine.Engine.terminate_epoch()` is called.
-    - COMPLETED : triggered when engine's run is completed
+    - COMPLETED : triggered when engine's run is completed or terminated :meth:`~ignite.engine.engine.Engine.terminate()`,
```
vfdev-5, line 268:
```suggestion
    - COMPLETED : triggered when engine's run is completed or terminated with :meth:`~ignite.engine.engine.Engine.terminate()`,
```

vfdev-5 (commented):
Thanks for the PR @bonassifabio !
I made the first pass on the code and left some comments, let me know what you think.

ignite/engine/engine.py
bonassifabio: I guess it would be useful in case one first terminates the training with `engine.terminate(True)`,  then for some reason resumes the training and wants to call `engine.terminate()` assuming `Events.COMPLETED` will be triggered. Do you think this is a realistic scenario?

ignite/engine/engine.py
bonassifabio: Agree on the reformulation of the if condition. 

Concerning your second point, could this create compatibility issues with existing handlers accessing `self.state.times[Events.COMPLETED.name]`? If we move the code below line 1100 into the if, in case of `Events.TERMINATE` the state `self.state.times[Events.COMPLETED.name]` would not be written...

ignite/engine/engine.py
vfdev-5: `self.skip_completed_after_termination` is always defined by the value passed in `engine.terminate(...)`, so if they resume and do not pass anything we'll internally overwrite `self.skip_completed_after_termination`. So, I do not see any problems in this case or missed something?

bonassifabio: Well, you are very right! Unless one manually fires `Events.TERMINATE` from an attached handler, I guess.

ignite/engine/engine.py
```diff
@@ -547,6 +548,9 @@ def terminate(self) -> None:
         - :attr:`~ignite.engine.events.Events.TERMINATE`
         - :attr:`~ignite.engine.events.Events.COMPLETED`
 
+        Args:
+            skip_completed: if True, the event `~ignite.engine.events.Events.COMPLETED` is not fired after
+            `~ignite.engine.events.Events.TERMINATE`. Default is False.
```
vfdev-5, line 553: There is a warning when building docs here: https://app.netlify.com/sites/pytorch-ignite-preview/deploys/674e251e13e2fa000828b048

Links to the events should be actually:
```
:attr:`~ignite.engine.events.Events.COMPLETED`
:attr:`~ignite.engine.events.Events.TERMINATE`
```

bonassifabio:
I just pushed [5f2cd88](https://github.com/pytorch/ignite/pull/3309/commits/5f2cd887328762288c1554f22228a5f04c908b8e) where I fixed the unit test and hopefully the docs. I also followed your suggestion concerning the update of `self.state.times[Events.COMPLETED.name]`

ignite/engine/engine.py
```diff
@@ -1002,17 +1000,19 @@ def _internal_run_as_gen(self) -> Generator[Any, None, State]:
...
             # time is available for handlers but must be updated after fire
             self.state.times[Events.COMPLETED.name] = time_taken
-            handlers_start_time = time.time()
 
-            # do not fire Events.COMPLETED if we terminated the run with flag `skip_completed_after_termination=True`
-            if self.should_terminate and not self.skip_completed_after_termination or not self.should_terminate:
+            if self.should_terminate and self.skip_completed_after_termination:
+                # do not fire Events.COMPLETED if we terminated the run with flag `skip_completed_after_termination=True`
+                hours, mins, secs = _to_hours_mins_secs(time_taken)
+                self.logger.info(f"Engine run terminated. Time taken: {hours:02d}:{mins:02d}:{secs:06.3f}")
```
vfdev-5, line 1007: Let's refactor this to avoid code duplication and finally use `not (...)`

vfdev-5:
Can you also please run the following to fix code formatting problems:
```bash
```

ignite/engine/engine.py
```diff
@@ -547,6 +548,9 @@ def terminate(self) -> None:
         - :attr:`~ignite.engine.events.Events.TERMINATE`
         - :attr:`~ignite.engine.events.Events.COMPLETED`
 
+        Args:
+            skip_completed: if True, the event :attr:`~ignite.engine.events.Events.COMPLETED` is not fired after
+            :attr:`~ignite.engine.events.Events.TERMINATE`. Default is False.
```
vfdev-5, line 553:
```suggestion
            skip_completed: if True, the event :attr:`~ignite.engine.events.Events.COMPLETED` is not fired after
                :attr:`~ignite.engine.events.Events.TERMINATE`. Default is False.
```

ignite/engine/engine.py
bonassifabio: Solved by using "finished" in the log message in place of "terminated" and "completed".

ignite/engine/events.py
```diff
@@ -265,7 +265,8 @@ class Events(EventEnum):
 
     - EPOCH_COMPLETED : triggered when the epoch is ended. Note that this is triggered even
       when :meth:`~ignite.engine.engine.Engine.terminate_epoch()` is called.
-    - COMPLETED : triggered when engine's run is completed
+    - COMPLETED : triggered when engine's run is completed or terminated with :meth:`~ignite.engine.engine.Engine.terminate()`,
+      unless the flag `skip_completed` is set to True.
```
vfdev-5, line 269: This line is too long: https://github.com/pytorch/ignite/actions/runs/12130079499/job/33819707813?pr=3309#step:11:80

```suggestion
    - COMPLETED : triggered when engine's run is completed or terminated with
      :meth:`~ignite.engine.engine.Engine.terminate()`, unless the flag
      `skip_completed` is set to True.
```

ignite/engine/events.py
```diff
@@ -265,7 +265,9 @@ class Events(EventEnum):
 
     - EPOCH_COMPLETED : triggered when the epoch is ended. Note that this is triggered even
       when :meth:`~ignite.engine.engine.Engine.terminate_epoch()` is called.
-    - COMPLETED : triggered when engine's run is completed
+    - COMPLETED : triggered when engine's run is completed or terminated with
+      :meth:`~ignite.engine.engine.Engine.terminate()`, unless the flag
+      `skip_completed` is set to True.
 
     The table below illustrates which events are triggered when various termination methods are called.
```
vfdev-5, line 272: @bonassifabio actually, I find myself as well this table very misleading and hard to understand.
I think we can improve it in the following way:
- EVENT_COMPLETED -> EPOCH_COMPLETED
- Let's add new column on the right after "TERMINATE" and call it "COMPLETED"
- Check symbol for `terminate()` line on `TERMINATE_SINGLE_EPOCH` column is wrong actually and should be replaced with `x`.



By the way, here is how it is generated now: https://deploy-preview-3309--pytorch-ignite-preview.netlify.app/generated/ignite.engine.events.events#ignite.engine.events.Events

bonassifabio: Actually, I was interpreting the first column as Events.COMPLETED, but from your comment it would seem that's not the case. 

What if we change the table to something like this?

| Method      |  `TERMINATE_SINGLE_EPOCH` | `EPOCH_COMPLETED` | `TERMINATE` | `COMPLETED` 
| ----------- | ----------- | ----------- | ----------- | ----------- |
| No termination |  x | ✔ | x | ✔
| `terminate_epoch()`   | ✔ | x | x |  x |
| `terminate()`   | x | x  |  ✔ |  ✔ |
| `terminate(skip_completed=True)`   | x  |  x |  ✔ |  x |

Few comments:

1. To my understanding (please correct me if I'm wrong) if `terminate()` is called, the epoch is not necessarily completed. If this statement is true, it would perhaps make more sense to move  `EPOCH_COMPLETED` before `TERMINATE` in the list of Events, as I did in the table, since `EPOCH_COMPLETED` would almost never be fired after `EPOCH_COMPLETED`.
2. The columns would thus be  ordered so that the first two are "epoch wise", while the last two happen at the end of the engine run. Not sure I
3. I included a new row for `terminate(skip_completed=True)`. Not sure if that's really necessary.

vfdev-5: Thanks for the new table, looks good!
Few corrections however about its content:

The line `terminate_epoch()`, column "EPOCH_COMPLETED" and "COMPLETED" should be both checked as `v`.


About your comments, I'm OK with all your suggestions, let's keep this order and the new line.


Checking code:
```python
from ignite.utils import setup_logger, logging


train_data = range(10)
max_epochs = 5


    pass

trainer = Engine(train_step)

# Enable trainer logger for a debug mode
trainer.logger = setup_logger("trainer", level=logging.DEBUG)

@trainer.on(Events.ITERATION_COMPLETED(once=12))
def call_terminate_epoch():
    trainer.terminate_epoch()

trainer.run(train_data, max_epochs=max_epochs)
```
Output:
```
...
2024-12-03 09:06:33,056 trainer INFO: Terminate current epoch is signaled. Current epoch iteration will stop after current iteration is finished.
2024-12-03 09:06:33,058 trainer DEBUG: 2 | 12, Firing handlers for event Events.TERMINATE_SINGLE_EPOCH
2024-12-03 09:06:33,060 trainer DEBUG: 2 | 12, Firing handlers for event Events.EPOCH_COMPLETED
2024-12-03 09:06:33,061 trainer INFO: Epoch[2] Complete. Time taken: 00:00:00.023
2024-12-03 09:06:33,064 trainer DEBUG: 3 | 12, Firing handlers for event Events.EPOCH_STARTED
2024-12-03 09:06:33,066 trainer DEBUG: 3 | 12, Firing handlers for event Events.GET_BATCH_STARTED
...
```


You can run it in https://pytorch-ignite.ai/playground

bonassifabio: Thank you for pointing this out! 

1. Do you think we should add an argument  `skip_completed` to `terminate_epoch()` with the same behavior of suppressing the firing of `Events.EPOCH_COMPLETED`? That might be useful, for example, to avoid running the evaluator, the checkpointer, and the LR scheduler if an epoch has been terminated. 
2. I agree on checking  `EPOCH_COMPLETED` for `terminate_epoch()`. However, I think that checking `COMPLETED` might seem to imply that `Events.COMPLETED` is also fired by `terminate_epoch()` right after `Events.EPOCH_COMPLETED`. What if we leave that cell empty?

vfdev-5:
@bonassifabio we have now the CI completely broken due to yesterday change of [pynvml package](https://pypi.org/project/pynvml/) what we should replace (https://github.com/pytorch/ignite/pull/3310). By the way, if you have suggestions on a similar package to be used, we are open to suggestions :)

ignite/engine/events.py
bonassifabio: > Good idea! Let's make this in a separate PR such that this one does not become too large

Cool, I'll create a new PR when we're done with this one

> I did not understand v mark on COMPLETED for terminate_epoch as the sequence of triggered events, but just whether an event is triggered or not.

I see the point, but doesn't the firing of `TERMINATE` and `COMPLETED` independent of `terminate_epoch()`?
See for example this code.

```
from ignite.utils import setup_logger, logging

train_data = range(10)
max_epochs = 5


    pass

trainer = Engine(train_step)

# Enable trainer logger for a debug mode
trainer.logger = setup_logger("trainer", level=logging.DEBUG)

@trainer.on(Events.ITERATION_COMPLETED(once=12))
def call_terminate_epoch():
    trainer.terminate_epoch()

@trainer.on(Events.ITERATION_COMPLETED(once=15))
def call_terminate():
    trainer.terminate(skip_completed=True)

trainer.run(train_data, max_epochs=max_epochs)
```

This is probably an edge case. As a newbie of ignite, it would honestly make more sense to see no entry for `TERMINATE` and `COMPLETED` for `terminate_epoch()`. However, you are for sure more experienced than me, so we can go ahead with your solution if you prefer 🙂

**Solution 1**

| Method      |  `TERMINATE_SINGLE_EPOCH` | `EPOCH_COMPLETED` | | `TERMINATE` | `COMPLETED` |
| ----------- | ----------- | ----------- | -- | ----------- | ----------- |
| No termination |  x | ✔ |   | x  | ✔ |
| `terminate_epoch()`   | ✔ | x | |  |  |
| `terminate()`   | x | x   | |  ✔ |  ✔ |
| `terminate(skip_completed=True)`   | x  |  x | |  ✔ |  x |


**Solution 2**

| Method      |  `TERMINATE_SINGLE_EPOCH` | `EPOCH_COMPLETED` | | `TERMINATE` | `COMPLETED` |
| ----------- | ----------- | ----------- | -- | ----------- | ----------- |
| No termination |  x | ✔ |   | x  | ✔ |
| `terminate_epoch()`   | ✔ | x | | x | ✔ |
| `terminate()`   | x | x   | |  ✔ |  ✔ |
| `terminate(skip_completed=True)`   | x  |  x | |  ✔ |  x |

bonassifabio: I would propose to keep the table (either Table 1 or Table 2), and then to open a new issue to seek some input also from other developers/users and discuss the problem more in detail there.
We could for example think to a flow chart diagram or something like that...

bonassifabio:
@vfdev-5 I am not sure why some test is failing, seems unrelated to the changes in this PR. Any input?

vfdev-5:
> @vfdev-5 I am not sure why some test is failing, seems unrelated to the changes in this PR. Any input?

We can ignore failing hvd tests

vfdev-5 (approved):
Thanks @bonassifabio, LGTM!
We can either fix the table here or in a follow-up PR

PR #3339, opened 2025-02-28
Add `_rank_not_in_group` to idist

sadra-barikbin:
Hi there!

To add `_rank_not_in_group` to `idist.utils`.

### DIFF
--- a/ignite/distributed/comp_models/horovod.py
+++ b/ignite/distributed/comp_models/horovod.py
@@ -209,4 +209,4 @@ if has_hvd_support:
             hvd.allreduce(torch.tensor(0, device="cpu"), name="barrier")
 
         def _rank_not_in_group(self, group: Any) -> bool:
-            return group.included()
+            return not group.included()
--- a/ignite/distributed/comp_models/horovod.py
+++ b/ignite/distributed/comp_models/horovod.py
@@ -7,6 +7,7 @@ from ignite.distributed.comp_models.base import ComputationModel
 
 try:
     import horovod.torch as hvd
+    from horovod.common.process_sets import global_process_set
 
     try:
         # old API
@@ -186,11 +187,9 @@ if has_hvd_support:
             return reduced_res[0]
 
         def _do_all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None) -> torch.Tensor:
-            if group is not None:
-                raise NotImplementedError("all_gather with group for horovod is not implemented")
             if tensor.ndimension() == 0:
                 tensor = tensor.unsqueeze(0)
-            return hvd.allgather(tensor)
+            return hvd.allgather(tensor, process_set=group or global_process_set)
 
         def _do_all_gather_object(self, tensor: Any, group: Optional[Any] = None) -> List[Any]:
             if group is not None:
--- a/ignite/distributed/comp_models/horovod.py
+++ b/ignite/distributed/comp_models/horovod.py
@@ -1,3 +1,4 @@
+import os
 import warnings
 from typing import Any, Callable, cast, List, Mapping, Optional, Tuple
 
@@ -22,6 +23,9 @@ except ImportError:
 
 if has_hvd_support:
     HOROVOD = "horovod"
+
+    # Enables dynamic process sets: new_group methods and passing group into collective ops
+    os.environ["HOROVOD_DYNAMIC_PROCESS_SETS"] = "1"
 
     class _HorovodDistModel(ComputationModel):
         """Private class for `Horovod <https://horovod.readthedocs.io/en/stable/>`_ distributed computation model."""
@@ -155,6 +159,15 @@ if has_hvd_support:
                 **kwargs,
             )
 
+        def _setup_group(self, group: Any) -> hvd.ProcessSet:
+            if isinstance(group, list) and all(isinstance(item, int) for item in group):
+                group = self._do_new_group(group)
+            if not isinstance(group, hvd.ProcessSet):
+                raise ValueError(
+                    f"Argument group should be list of int or hvd.ProcessSet, got {type(group)}, group={group}"
+                )
+            return group
+
         _reduce_op_map = {
             "SUM": hvd.mpi_ops.Sum,
             "AVERAGE": hvd.mpi_ops.Average,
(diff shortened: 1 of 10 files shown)


ignite/distributed/utils.py
```diff
@@ -737,3 +740,11 @@ def download_dataset():
 
     if current_rank == rank:
         barrier()
+
+
+def _rank_not_in_group(group: Optional[dist.ProcessGroup | "hvd.ProcessSet"]) -> bool:
```
vfdev-5, line 745:
```suggestion
def _rank_not_in_group(group: Optional[Union[Any, List[int]]]) -> bool:
```
maybe this type hint instead?

ignite/distributed/utils.py
```diff
@@ -737,3 +740,11 @@ def download_dataset():
...
+    """Check if the current process's rank is not in a given group."""
+    if has_hvd_support and isinstance(group, hvd.common.process_sets.ProcessSet):
```
vfdev-5, line 747: As group can be None we should handle that as well

vfdev-5:

ignite/distributed/utils.py
sadra-barikbin: `dist._rank_not_in_group(group)` handles that. Doesn't it?

ignite/distributed/utils.py
```diff
@@ -16,6 +16,9 @@
 )
 from ignite.utils import setup_logger
 
+if has_hvd_support:
```
vfdev-5, line 19: Let's remove this as it was for typing only, right?

ignite/distributed/utils.py
vfdev-5: In case when we have HVD and group is None, we should not enter `dist._rank_not_in_group(group)` ?

ignite/distributed/utils.py
```diff
@@ -16,6 +16,7 @@
 )
 from ignite.utils import setup_logger
 
+
```
vfdev-5, line 19:
```suggestion
```

ignite/distributed/utils.py
```diff
@@ -737,3 +738,17 @@ def download_dataset():
...
+    if group is None:
+        return False
+    if isinstance(group, list) and all(isinstance(item, int) for item in group):
+        group = new_group(group)
+    if has_hvd_support:
+        from hvd.common.process_sets import ProcessSet
+
+        if isinstance(group, ProcessSet):
+            return group.included()
+    return dist._rank_not_in_group(group)
```
vfdev-5, line 754: Ideally, we should implement `_rank_not_in_group` in all backends and call `_model._rank_not_in_group` method from here.

sadra-barikbin: Yes, you're right

sadra-barikbin:
@vfdev-5 , is XLA's `_do_new_group` correct?
```python
def _do_new_group(self, ranks: List[int], **kwargs: Any) -> Any:
            return [ranks] # Shouldn't it return `list(ranks)` ?
```

ignite/distributed/comp_models/horovod.py
```diff
@@ -208,3 +207,6 @@ def barrier(self) -> None:
             # https://github.com/horovod/horovod/issues/159#issuecomment-424834603
             # hvd.allreduce(torch.tensor(0, device=self.device()), name="barrier")
             hvd.allreduce(torch.tensor(0, device="cpu"), name="barrier")
+
+        def _rank_not_in_group(self, group: Any) -> bool:
+            return group.included()
```
vfdev-5, line 212: Should not this be the opposite:
```suggestion
            return not group.included()
```

ignite/distributed/comp_models/xla.py
```diff
@@ -175,3 +175,6 @@ def _check_group_type(self, group: Optional[Any]) -> bool:
             if isinstance(group, list) and all(isinstance(item, int) for item in group):
                 return True
             return False
+
+        def _rank_not_in_group(self, group: Any) -> bool:
+            return self.get_rank() in group
```
vfdev-5, line 180: same here: 
```suggestion
            return self.get_rank() not in group
```

ignite/distributed/utils.py
```diff
@@ -737,3 +737,12 @@ def download_dataset():
...
     if current_rank == rank:
         barrier()
+
+
+def _rank_not_in_group(group: Optional[Union[Any, List[int]]]) -> bool:
+    """Check if the current process's rank is not in a given group."""
+    if group is None:
+        return False
+    if isinstance(group, list) and all(isinstance(item, int) for item in group):
+        group = new_group(group)
```
vfdev-5, line 747: Can we hide this inside horovod implementation of _do_all_gather / _rank_not_in_group etc ?

vfdev-5:
> @vfdev-5 , is XLA's `_do_new_group` correct?
> 
> ```python
> def _do_new_group(self, ranks: List[int], **kwargs: Any) -> Any:
>             return [ranks] # Shouldn't it return `list(ranks)` ?
> ```

On xla `group` is a list of list of ranks, for example: https://pytorch.org/xla/release/r2.6/learn/api-guide.html#torch_xla.core.xla_model.all_gather

vfdev-5:
Failing test:
```
  >       gloo_hvd_executor(_test_distrib_all_gather_group, (device,), np=np, do_init=True)

   [0]<stderr>:  File "/home/runner/work/ignite/ignite/tests/ignite/distributed/utils/__init__.py", line 228, in _test_distrib_all_gather_group
  [0]<stderr>:    res = idist.all_gather(t, group=group)
  [0]<stderr>:          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [0]<stderr>:  File "/home/runner/work/ignite/ignite/ignite/distributed/utils.py", line 435, in all_gather
  [0]<stderr>:    return _model.all_gather(tensor, group=group)
  [0]<stderr>:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [0]<stderr>:  File "/home/runner/work/ignite/ignite/ignite/distributed/comp_models/base.py", line 224, in all_gather
  [0]<stderr>:    return self._collective_op(tensor, self._do_all_gather, group=group)
  [1]<stderr>:User function raise error: horovod_torch_allgather_async_torch_LongTensor(): incompatible function arguments. The following argument types are supported:
  [1]<stderr>:    1. (arg0: torch.Tensor, arg1: torch.Tensor, arg2: str, arg3: int) -> int
```

PR #3054, opened 2023-08-31
Distributed ndcg

ili0820:
Fixes #2632

This PR is based on https://github.com/pytorch/ignite/pull/2632 (thanks a lot to kamalojasv181!)

Description:

I ve tried to fix distributed NDCG tests like the [examples](https://github.com/pytorch/ignite/pull/2632#issuecomment-1699564400). I am not really sure whether this is the way you guys wanted, but still, I tried my best and let me know which parts should be fixed  (It's my first time contributing, so be kind plz 😯 )

Check list:
- [x] New tests are added (if a new feature is added)
- [ ] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #2632, opened 2022-07-23
Added the ndcg metric [WIP]

kamalojasv181:
Related #2631

Description: This is the implementation [WIP] for the NDCG metric.

Check list:

- [X] Create a basic implementation. 
- [x] Handle rating ties.
- [x] Write and parameterize tests.
- [ ] Add comments about its use
- [X] Add exponential implementation
- [ ] Add Docs

Issue #2631, opened 2022-07-23
Implementation of RecSys Metrics


kamalojasv181:
## 🚀 Feature

To add some more metrics used in Recommendation Systems with tests that can directly be used. Here is the list (@alykhantejani feel free to add more):
[] NGCD

shivance:
Can I contribute to this one ?

vfdev-5:
@shivance yes, sure. Do you have a recsys metric in mind to implement ?

shivance:
`FactorizedTopK` could be a place to start?

I'm not aware of many metrics, so am open to learn !

vfdev-5:
@shivance how about implementing hitrate, mean reciprocal rank etc ?
Some refs to check:
- https://github.com/catalyst-team/catalyst/blob/master/catalyst/metrics/functional/_hitrate.py
- https://github.com/catalyst-team/catalyst/blob/master/catalyst/metrics/functional/_mrr.py
- https://github.com/zuoxingdong/recsys_metrics/tree/master/recsys_metrics

shivance:
Sounds good !

rwtarpit:
> @shivance how about implementing hitrate, mean reciprocal rank etc ? Some refs to check:
> 
> * https://github.com/catalyst-team/catalyst/blob/master/catalyst/metrics/functional/_hitrate.py
> * https://github.com/catalyst-team/catalyst/blob/master/catalyst/metrics/functional/_mrr.py
> * https://github.com/zuoxingdong/recsys_metrics/tree/master/recsys_metrics

can i work on implementing hitrate?

vfdev-5:
Yes, you can do that

Koolvansh07:
i See this issues has been opened for a while i think i can work on this

aniketpandey05:
Hi @vfdev-5, I noticed that Precision and Recall are still
missing from ignite/metrics/rec_sys/.

Would it be a good idea to add these? I can implement both
following the same structure as the existing HitRate implementation.

vfdev-5:
There are already Recall and Precision in ignite.metrics, they can be adapted for recsys as well

aniketpandey05:
Thanks for the clarification!

Would you recommend opening an issue to track adding adapted versions for `ignite/metrics/rec_sys`, or would it be fine if I directly work on a PR that adapts the existing `Precision` and `Recall` implementations for recommender systems (similar to how `HitRate` is structured)?

rwtarpit:
[...]

PR #3346, opened 2025-03-19
Add NDCG metrics

exitflynn:
Adds NDCG metrics as a part of adding RecSys metrics in #2631  

This PR builds on the work done in #3054 and #2632 (much thanks to ili0820 and kamalojasv181!)

The primary changes left at this stage as per my understanding would be around vectorising parts of the DCG computation as previously discussed [here](https://github.com/pytorch/ignite/pull/2632/files#r930048810) and [here](https://github.com/pytorch/ignite/pull/3054/files#r1311816781), and then docs.

I've considered different ways to address that, namely using `torch.vmap` (but I think that will be better to pursue once https://github.com/pytorch/pytorch/issues/124423 is resolved); a batched vectorisation and a full one. Right now I think the batch-based vectorisation could be a good way to go since a full vectorisation approach could be a bit too complex (/overkill?) here, the main cause of complexity being different numbers of unique values across samples. Let me know what you think! I'm still in the early stages of working at this level and so open to feedback is all I am :cowboy_hat_face:


vfdev-5:
Thanks for the PR @exitflynn 
Concerning the vectorization, can you please check how others implement this, for example here:
- https://github.com/catalyst-team/catalyst/blob/master/catalyst/metrics/functional/_ndcg.py
- https://github.com/zuoxingdong/recsys_metrics
- maybe, some other libs

The part to try to vectorize is `_tie_averaged_dcg_batched`. I do think vmap will ever support getitem operator.
What we can do first is to run some benchmarks on the same data between `ignore_ties=True` vs `ignore_ties=False` and see how slow `_tie_averaged_dcg_batched` implementation.
Next, we can try to improve it, maybe padding with zeros can help?
I haven't yet checked the details.

vfdev-5:
@exitflynn if you have any questions, blockers, feel free to ask here or on our discord: https://pytorch-ignite.ai/chat

vfdev-5:
Closing in favor of https://github.com/pytorch/ignite/pull/3608

Issue #3335, opened 2025-02-27
Improve metrics tests


vfdev-5:
The idea is to add `available_device` pytest fixture into the appropriate metric tests. For example:

```diff
@pytest.mark.parametrize("n_times", range(5))
- def test_multiclass_input(n_times, test_data):
+ def test_multiclass_input(n_times, available_device, test_data):
    y_pred, y, num_classes, batch_size = test_data
-    cm = ConfusionMatrix(num_classes=num_classes)
+    cm = ConfusionMatrix(num_classes=num_classes, device=available_device)
+.   assert cm._device == torch.device(available_device)
    ...
```
This will generate 2 more tests and run `test_multiclass_input` on cpu, cuda if available, mps if available.

**No need** to add `available_device` fixture to the tests:
1) checking if an error is raised, for example `test_confusion_matrix.py::test_num_classes_wrong_input`:
```python
def test_num_classes_wrong_input():
    with pytest.raises(ValueError, match="Argument num_classes needs to be > 1"):
        ConfusionMatrix(num_classes=1)
```
2) running distributed tests, e.g. tests inside `TestDistributed` or `test_distrib_*`.

Files to update:
- [x] [test_average_precision.py](https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/test_average_precision.py)
- [x] test_classification_report.py
- [x] test_cohen_kappa.py
- [x] test_confusion_matrix.py
- [x] test_cosine_similarity.py
- [x] test_entropy.py
- [x] test_fbeta.py
- [x] test_frequency.py
- [x] test_hsic.py
- [x] test_js_divergence.py
- [x] test_kl_divergence.py
- [x] test_loss.py
- [x] test_maximum_mean_discrepancy.py
- [x] test_mean_absolute_error.py
- [x] test_mean_pairwise_distance.py
- [x] test_mean_squared_error.py
- [x] test_multilabel_confusion_matrix.py
- [x] test_mutual_information.py
- [ ] test_precision_recall_curve.py
- [x] test_roc_auc.py
- [x] test_roc_curve.py
- [x] test_root_mean_squared_error.py
- [x] test_top_k_categorical_accuracy.py
- [x] [test_object_detection_map.py](https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/vision/test_object_detection_map.py)

Additional lot of metrics:
- [x] clustering / test_calinski_harabasz_score.py ... test_silhouette_score.py
[...]

PR #3351, opened 2025-03-25
adds argument name for device in test_mean_squared_error #3335

BanzaiTokyo:
missing argument name for device when instantiating MeanSquaredError

### DIFF
--- a/tests/ignite/metrics/test_mean_squared_error.py
+++ b/tests/ignite/metrics/test_mean_squared_error.py
@@ -30,7 +30,7 @@ def test_case(request):
 
 @pytest.mark.parametrize("n_times", range(5))
 def test_compute(n_times, available_device, test_case):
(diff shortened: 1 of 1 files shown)


vfdev-5 (approved):
Thanks for catching this @BanzaiTokyo !

tests/ignite/metrics/test_mean_squared_error.py
```diff
@@ -48,6 +48,7 @@ def test_compute(n_times, available_device, test_case):
 
     np_res = np.power((np_y - np_y_pred), 2.0).sum() / np_y.shape[0]
 
+    assert mse._device == torch.device(available_device)
```
vfdev-5, line 51: Can you put this just after the metric creating to fail it earlier.

vfdev-5 (approved):
LGTM

PR #3360, opened 2025-03-27
Test fbeta available device #3335

BanzaiTokyo:


tests/ignite/metrics/test_fbeta.py
```diff
@@ -37,26 +37,40 @@ def _output_transform(output):
...
+    p, r = precision_and_recall
 
     n_iters = 10
     batch_size = 10
     n_classes = 10
 
-    y_true = np.arange(0, n_iters * batch_size, dtype="int64") % n_classes
-    y_pred = 0.2 * np.random.rand(n_iters * batch_size, n_classes)
+    y_true = torch.arange(n_iters * batch_size, dtype=torch.long).to(available_device) % n_classes
+    y_pred = 0.2 * torch.rand(n_iters * batch_size, n_classes).to(available_device)
```
vfdev-5, line 73:
```suggestion
    y_true = torch.arange(n_iters * batch_size, dtype=torch.long, device=available_device) % n_classes
    y_pred = 0.2 * torch.rand(n_iters * batch_size, n_classes, device=available_device)
```

tests/ignite/metrics/test_fbeta.py
```diff
@@ -37,26 +37,40 @@ def _output_transform(output):
...
-        (None, None, False, _output_transform),
-        (None, None, True, _output_transform),
-        (Precision(average=False), Recall(average=False), False, None),
-        (Precision(average=False), Recall(average=False), True, None),
+        ("none", False, None),
+        ("none", True, None),
+        ("none", False, _output_transform),
+        ("none", True, _output_transform),
+        ("custom", False, None),
+        ("custom", True, None),
```
vfdev-5, line 61: Let's make it a bit simplier:
```suggestion
"precision_cls, recall_cls, average, output_transform",
[
        (None, None, False, None),
        ... (keep previous code) ...
        
        (lamdba device: Precision(average=False, device=device), lamdba device: Recall(average=False, device=device), True, None),

...
p = precision_cls(available_device)
p._device == torch.device(available_device)

r = recall_cls(available_device)
r._device == torch.device(available_device)

```

BanzaiTokyo: please check the results

vfdev-5:
Let's also check inside `Fbeta` implementation that if provided precision, recall metrics, their device corresponds to the provided device. Also let's change the default value here to None: https://github.com/pytorch/ignite/blob/795dccb4282dc2880791fb784ff9bda762ad8bf2/ignite/metrics/fbeta.py#L18
If precision, recall are None and device is None then set device to cpu.

ignite/metrics/fbeta.py
```diff
@@ -149,6 +149,9 @@ def thresholded_output_transform(output):
     if recall is not None and output_transform is not None:
         raise ValueError("If recall argument is provided, output_transform should be None")
 
+    if recall is not None and precision is not None and device is not None:
```
vfdev-5, line 152: Make the same check per metric as for output_transform, see above code

tests/ignite/metrics/test_fbeta.py
```diff
@@ -22,7 +22,7 @@ def test_wrong_inputs():
 
     with pytest.raises(ValueError, match=r"Input recall metric should have average=False"):
         r = Recall(average="samples")
-        Fbeta(1.0, recall=r)
+        Fbeta(1.0, recall=r, device=torch.device("cpu"))
```
vfdev-5, line 25: Revert this change and write its own "error raised" check

BanzaiTokyo: like this?
```python
    with pytest.raises(TypeError, match=r"device\(\) received an invalid combination of arguments"):
        r = Recall(average="samples")
        Fbeta(1.0, recall=r)
```

tests/ignite/metrics/test_fbeta.py
```diff
@@ -20,7 +20,7 @@ def test_wrong_inputs():
         p = Precision(average="micro")
         Fbeta(1.0, precision=p)
 
-    with pytest.raises(ValueError, match=r"Input recall metric should have average=False"):
+    with pytest.raises(TypeError, match=r"device\(\) received an invalid combination of arguments"):
```
vfdev-5, line 23: Revert this and add new code like below:
```python
    with pytest.raises(ValueError, match=r"If precision argument is provided, output_transform should be None"):
        p = Precision(average=False)
        Fbeta(1.0, precision=p, output_transform=lambda x: x)
```
but adapted to device

BanzaiTokyo: I am missing something... If I revert to the original check:
```python
    with pytest.raises(ValueError, match=r"Input recall metric should have average=False"):
        r = Recall(average="samples")
        Fbeta(1.0, recall=r)
```
it fails with:
```
self = <ignite.metrics.precision.Precision object at 0x7256b39409b0>
output_transform = <function Fbeta.<locals>.<lambda> at 0x7256b1bca700>
device = None, skip_unrolling = False

    def __init__(
        self,
        skip_unrolling: bool = False,
    ):
        self._output_transform = output_transform
    
        # Some metrics have a large performance regression when run on XLA devices, so for now, we disallow it.
>       if torch.device(device).type == "xla":
E       TypeError: device() received an invalid combination of arguments - got (NoneType), but expected one of:
E        * (torch.device device)
E             didn't match because some of the arguments have invalid types: (!NoneType!)
E        * (str type, int index = -1)
```
if I understand it correctly in Fbeta precision is instantiated with device=None and we are not addressing this problem.

vfdev-5: I think you haven't implemented this part of the comment: https://github.com/pytorch/ignite/pull/3360#issuecomment-2760913003 :

> If precision, recall are None and device is None then set device to cpu.

In details it should be something like this:
1) both precision and recall are None and device is None => set device to cpu
2) if one of precision or recall is defined and other is None, fetch _device from the defined metric and set device to it.
3) if both precision and recall are defined and device is None => keep device as None.

ignite/metrics/fbeta.py
```diff
@@ -143,17 +143,26 @@ def thresholded_output_transform(output):
     if not (beta > 0):
         raise ValueError(f"Beta should be a positive integer, but given {beta}")
 
-    if precision is not None and output_transform is not None:
-        raise ValueError("If precision argument is provided, output_transform should be None")
+    if precision is not None:
+        if output_transform is not None:
+            raise ValueError("If precision argument is provided, output_transform should be None")
+        if device is not None:
+            raise ValueError("If precision argument is provided, device should be None")
```
vfdev-5, line 150: We should write checks that this error and a similar the one below are properly raised in `test_wrong_inputs`, see:
```python
    with pytest.raises(ValueError, match=r"If precision argument is provided, output_transform should be None"):
        p = Precision(average=False)
        Fbeta(1.0, precision=p, output_transform=lambda x: x)
```

tests/ignite/metrics/test_fbeta.py
```diff
@@ -71,19 +99,31 @@ def update_fn(engine, batch):
...
+    f2 = Fbeta(beta=2.0, average=average, precision=p, recall=r, output_transform=output_transform, device=device)
+
     f2.attach(evaluator, "f2")
 
     data = list(range(n_iters))
     state = evaluator.run(data, max_epochs=1)
 
-    f2_true = fbeta_score(y_true, np.argmax(y_pred, axis=-1), average="macro" if average else None, beta=2.0)
-    np.testing.assert_allclose(np.array(f2_true), np.array(state.metrics["f2"]))
+    f2_pred = to_numpy_float32(state.metrics["f2"])
```
vfdev-5, line 115: Why we should convert to f32 for all devices: cpu, cuda, mps?

BanzaiTokyo: great observation. But to be honest, I don't see how to make conversions only on MPS without overly polluting the code with device checks. How would you like to do that?

tests/ignite/metrics/test_fbeta.py
```diff
@@ -24,6 +33,10 @@ def test_wrong_inputs():
         r = Recall(average="samples")
         Fbeta(1.0, recall=r)
 
+    with pytest.raises(ValueError, match=r"If precision argument is provided, device should be None"):
```
vfdev-5, line 36: Same for recall

tests/ignite/metrics/test_fbeta.py
vfdev-5: Fbeta should handle dtype internally, for cpu and cuda as f64 is available, the output should be f64 (to verify). For mps, precision/recall value should be f32. So, to compare output with the expected we can compare float values directly and have to convert output to numpy array and it should keep the dtype from torch tensor. What do you think?

tests/ignite/metrics/test_fbeta.py
```diff
@@ -71,19 +94,22 @@ def update_fn(engine, batch):
...
+
     f2.attach(evaluator, "f2")
 
     data = list(range(n_iters))
     state = evaluator.run(data, max_epochs=1)
 
-    f2_true = fbeta_score(y_true, np.argmax(y_pred, axis=-1), average="macro" if average else None, beta=2.0)
-    np.testing.assert_allclose(np.array(f2_true), np.array(state.metrics["f2"]))
+    f2_true = fbeta_score(y_true, torch.argmax(y_pred, dim=-1), average="macro" if average else None, beta=2.0)
+    f2_true = np.float32(f2_true) if available_device == "mps" else f2_true
```
vfdev-5, line 111: Let's explicitly convert y_true and torch.argmax(y_pred, dim=-1) to numpy passing by cpu.
And remove `np.float32(f2_true)`:
```suggestion
    f2_true = fbeta_score(y_true, torch.argmax(y_pred, dim=-1), average="macro" if average else None, beta=2.0)
```

BanzaiTokyo: something like that?

```python
    y_true_np = y_true.detach().cpu().numpy()
    y_pred_np = torch.argmax(y_pred, dim=-1).detach().cpu().numpy()
    f2_true = fbeta_score(y_true_np, y_pred_np, average="macro" if average else None, beta=2.0)
```

vfdev-5 (approved):
LGTM, we can merge as failures are unrelated to the PR

Issue #3352, opened 2025-03-25
Add output_transform type check in Metric class


vfdev-5:
Let's check `output_transform` input argument type:
```python
class Metric(...):

    def __init__(
        self,
        output_transform: Callable = lambda x: x,
        device: Union[str, torch.device] = torch.device("cpu"),
        skip_unrolling: bool = False,
    ):
        if not callable(output_transform):
            raise TypeError(
                "Argument output_transform should be callable, "
                f"got {type(output_transform)}"
            )
        self._output_transform = output_transform
        ...
```

TODO:
- [ ] Add above code
- [ ] Add a test checking that error is raised to https://github.com/pytorch/ignite/blob/master/tests/ignite/metrics/test_metric.py

PR #3371, opened 2025-04-04
Add output_transform type check in Metric class (#3352)

HyeSungP:
Implement #3352

Check list:

- [x] Add above code
- [x] Add a test checking that error is raised to

### DIFF
--- a/ignite/metrics/metric.py
+++ b/ignite/metrics/metric.py
@@ -361,6 +361,8 @@ class Metric(Serializable, metaclass=ABCMeta):
         device: Union[str, torch.device] = torch.device("cpu"),
         skip_unrolling: bool = False,
     ):
+        if not callable(output_transform):
+            raise TypeError("Argument output_transform should be callable, " f"got {type(output_transform)}")
         self._output_transform = output_transform
 
         # Some metrics have a large performance regression when run on XLA devices, so for now, we disallow it.
--- a/ignite/metrics/metric.py
+++ b/ignite/metrics/metric.py
@@ -362,7 +362,7 @@ class Metric(Serializable, metaclass=ABCMeta):
         skip_unrolling: bool = False,
     ):
         if not callable(output_transform):
-            raise TypeError("Argument output_transform should be callable, " f"got {type(output_transform)}")
+            raise TypeError(f"Argument output_transform should be callable, got {type(output_transform)}")
         self._output_transform = output_transform
 
         # Some metrics have a large performance regression when run on XLA devices, so for now, we disallow it.
--- a/tests/ignite/metrics/test_metric.py
+++ b/tests/ignite/metrics/test_metric.py
@@ -1470,17 +1470,8 @@ def test_access_to_metric_dunder_attributes():
     assert "value" in inspect.signature(metric).parameters.keys()
 
 
-class DummyMetric7(Metric):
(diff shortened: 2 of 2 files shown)


ignite/metrics/metric.py
```diff
@@ -361,6 +361,8 @@ def __init__(
         device: Union[str, torch.device] = torch.device("cpu"),
         skip_unrolling: bool = False,
     ):
+        if not callable(output_transform):
+            raise TypeError("Argument output_transform should be callable, " f"got {type(output_transform)}")
```
vfdev-5, line 365: Please fix the f-string here

tests/ignite/metrics/test_metric.py
```diff
@@ -1468,3 +1468,19 @@ def test_access_to_metric_dunder_attributes():
...
+    def update(self):
+        pass
+
+    def compute(self):
+        pass
+
+
+def test_output_transform_type_check():
+    with pytest.raises(TypeError, match="Argument output_transform should be callable"):
+        DummyMetric7(output_transform=1)
```
vfdev-5, line 1486: Let's remove the definition DummyMetric7 and DummyMetric1 for the test

ignite/metrics/metric.py
HyeSungP: Thanks! I've updated the f-string formatting accordingly.

tests/ignite/metrics/test_metric.py
HyeSungP: @vfdev-5 Thanks for the advice. Would it be better to define a single, shared DummyMetric class for these tests, or should we define the class within each test function, as is done in some other tests?

vfdev-5: Thanks for pointing out this! Maybe, later we can improve this part. I do not think we can define a single shared DummyMetric for all test cases. Probably, there are also some tests using some globally defined DummyMetric, so both cases can exist.
For our purposes I think we can reuse DummyMetric1

vfdev-5: @HyeSungP can you please update the test according to the last comment: https://github.com/pytorch/ignite/pull/3371#discussion_r2028357790

-> For our purposes I think we can reuse DummyMetric1 and remove DummyMetric7

HyeSungP: @vfdev-5 Sorry for the delay. I've updated the code based on your advice. Could you please review it?

tests/ignite/metrics/test_metric.py
```diff
@@ -1468,3 +1468,10 @@ def test_access_to_metric_dunder_attributes():
 
     # `inspect.signature` accesses `__signature__` attribute of the metric.
     assert "value" in inspect.signature(metric).parameters.keys()
+
+
+def test_output_transform_type_check():
+    y_pred = torch.tensor([[2.0], [-2.0]])
+    y = torch.zeros(2)
+    with pytest.raises(TypeError, match="Argument output_transform should be callable"):
+        DummyMetric1(true_output=(y_pred, y), output_transform=1)
```
vfdev-5, line 1477: You can probably set treu output to none. But ok to have it this way.

vfdev-5 (approved):
LGTM, thanks @HyeSungP !

PR #3373, opened 2025-04-05
At end of epoch on an IterableDataset ensure engine.state.output rema…

### LINKED ISSUE #3372: Engine object does not retain state.output when using an IterableDataset
## 🐛 Engine object does not retain state.output when using an IterableDataset

When using Pytorch ignite with an `IterableDataset`, The engine.state object at the end of an epoch, `state.output` is consistently `None`.  When using a map style dataset `state.output` contains the return value of the model's training function at the end of the epoch.

This can be reproduced with the following snippet of python
```
import torch
from torch.utils.data import DataLoader, IterableDataset
from torchvision.models import resnet18
from torchvision.transforms import Compose, Normalize, ToTensor

from ignite.engine import create_supervised_trainer

#device = torch.device("cuda")
device = torch.device("mps")
#device = torch.device("cpu")

#use_iterable = False
use_iterable = True

class Net(torch.nn.Module):
    def __init__(self):
        super(Net, self).__init__()
    
        self.model = resnet18(num_classes=10)

        self.model.conv1 = self.model.conv1 = nn.Conv2d(
            1, 64, kernel_size=3, padding=1, bias=False
        )

    def forward(self, x):
        return self.model(x)
model = Net().to(device)

class IterableMnist(IterableDataset):
    mnist = MNIST(download=True, root=".", transform=data_transform, train=False)
    def __iter__(self):
        for item in IterableMnist.mnist:
            yield item

mnist = IterableMnist() if use_iterable else MNIST(download=True, root=".", transform=data_transform, train=False)

train_loader = DataLoader(mnist, batch_size=128)

optimizer = torch.optim.RMSprop(model.parameters(), lr=0.005)

trainer = create_supervised_trainer(model, optimizer, criterion, device)

state = trainer.run(train_loader, max_epochs=1)
print(state.output)
```
To run, set your device appropriately at the top, then set `use_iterable` to True or False. When `use_iterable` is True, the MNIST dataset will be an `IterableDataset` and nothing will be printed. When `use_iterable` is False, the MNIST dataset will be the usual `Dataset` and the loss value from resnet18 will be printed.

I would expect that if an `IterableDataset` is used the engine `state.output` member would have the same behavior as when `Dataset` is provided. It seems to me that using map style datasets (`Dataset`) produces correct behavior and iterator style datasets (`IterableDataset`) does not.

I believe the fix is to stop resetting state.output in these two places:
https://github.com/pytorch/ignite/blob/f4f7d1cd228048f27437d38e726395f561ae68c7/ignite/engine/engine.py#L1072
https://github.com/pytorch/ignite/blob/f4f7d1cd228048f27437d38e726395f561ae68c7/ignite/engine/engine.py#L1257

It appears to me that when the loop in the `_run_once_on_dataset_*` functions is terminated due to a `StopIteration` Exception, self.state.output has been reset unnecessarily.

## Environment

- PyTorch Version (e.g., 1.4): 2.4.0
- Ignite Version (e.g., 0.3.0): 0.5.1
- OS (e.g., Linux): OSX
- How you installed Ignite (`conda`, `pip`, source): Installed from pip inside a conda environment
- Python version: 3.10

mtauraso:
…ins.

Fixes #3372 

Description: Implements the fix described in the bugreport

Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -1069,7 +1069,7 @@ class Engine(Serializable):
                 )
 
             while True:
-                self.state.batch = self.state.output = None
+                self.state.batch = None
 
                 try:
                     # Avoid Events.GET_BATCH_STARTED triggered twice when data iter is restarted
@@ -1254,7 +1254,7 @@ class Engine(Serializable):
                 )
 
             while True:
-                self.state.batch = self.state.output = None
+                self.state.batch = None
                 try:
                     # Avoid Events.GET_BATCH_STARTED triggered twice when data iter is restarted
                     if self.last_event_name != Events.DATALOADER_STOP_ITERATION:
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -1081,7 +1081,7 @@ class Engine(Serializable):
                             yield from self._maybe_terminate_or_interrupt()
 
                     self.state.batch = next(self._dataloader_iter)
-                    # We on purpose reset state.output here as for iterable dataloaders 
+                    # We on purpose reset state.output here as for iterable dataloaders
                     # we accidentally can remove it when one epoch is completed.
                     self.state.output = None
 
@@ -1268,10 +1268,10 @@ class Engine(Serializable):
                             self._maybe_terminate_legacy()
 
                     self.state.batch = next(self._dataloader_iter)
-                    # We on purpose reset state.output here as for iterable dataloaders 
+                    # We on purpose reset state.output here as for iterable dataloaders
                     # we accidentally can remove it when one epoch is completed.
                     self.state.output = None
-                    
+
                     # We should not trigger GET_BATCH_STARTED, GET_BATCH_COMPLETED, DATALOADER_STOP_ITERATION events
                     # if no data was provided to engine.run(data=None, ...)
                     if self.state.dataloader is not None:
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -1081,6 +1081,9 @@ class Engine(Serializable):
                             yield from self._maybe_terminate_or_interrupt()
 
                     self.state.batch = next(self._dataloader_iter)
+                    # We on purpose reset state.output here as for iterable dataloaders 
+                    # we accidentally can remove it when one epoch is completed.
+                    self.state.output = None
 
                     # We should not trigger GET_BATCH_STARTED, GET_BATCH_COMPLETED, DATALOADER_STOP_ITERATION events
                     # if no data was provided to engine.run(data=None, ...)
@@ -1265,6 +1268,10 @@ class Engine(Serializable):
                             self._maybe_terminate_legacy()
 
                     self.state.batch = next(self._dataloader_iter)
+                    # We on purpose reset state.output here as for iterable dataloaders 
+                    # we accidentally can remove it when one epoch is completed.
+                    self.state.output = None
+                    
                     # We should not trigger GET_BATCH_STARTED, GET_BATCH_COMPLETED, DATALOADER_STOP_ITERATION events
                     # if no data was provided to engine.run(data=None, ...)
                     if self.state.dataloader is not None:
--- a/tests/ignite/engine/test_engine.py
+++ b/tests/ignite/engine/test_engine.py
@@ -1279,6 +1279,44 @@ class TestEngine:
         assert engine.state.epoch == 10
         assert engine.state.iteration == 10 * real_epoch_length
 
+    def test_iterator_state_output(self):
+        torch.manual_seed(12)
+
+        def finite_iterator(length, batch_size):
+            for _ in range(length):
+                batch = torch.rand(batch_size, 3, 32, 32)
+                yield batch
+
+        def train_step(trainer, batch):
+            s = trainer.state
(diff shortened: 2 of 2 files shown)


ignite/engine/engine.py
```diff
@@ -1069,7 +1069,7 @@ def _run_once_on_dataset_as_gen(self) -> Generator[State, None, float]:
                 )
 
             while True:
-                self.state.batch = self.state.output = None
+                self.state.batch = None
```
vfdev-5, line 1072: Instead of removing `self.state.output = None`, let's move it below before the line `self.state.batch = next(self._dataloader_iter)`:
```python
self.state.batch = next(self._dataloader_iter)
# We on purpose reset state.output here as for iterable dataloaders 
# we accidentally can remove it when one epoch is completed.
self.state.output = None
```

vfdev-5 (commented):
Thanks for the PR @mtauraso , I have another suggestion here to approximatively keep the previous behavior.

ignite/engine/engine.py
mtauraso: Agree, I'll push that to the branch now.

vfdev-5:
@mtauraso can you please run code formatting script: `bash ./tests/run_code_style.sh install && bash ./tests/run_code_style.sh fmt` to fix:
```
+ flake8 ignite tests examples --config setup.cfg
ignite/engine/engine.py:1084:88: W291 trailing whitespace
ignite/engine/engine.py:1271:88: W291 trailing whitespace
ignite/engine/engine.py:1274:1: W293 blank line contains whitespace
```

mtauraso:
Ran the formatter and committed. My apologies for the churn, I see now this is all documented in the first time contributors guide which I didn't read earlier.

vfdev-5:
Thanks for the updates and no worries about the contributing guidelines. By the way, I almost forgot about that we also need to add a test for the feature we adding. Can you please create a test here checking that state.output is not None for the iterable input data, below this line: https://github.com/pytorch/ignite/blob/f4f7d1cd228048f27437d38e726395f561ae68c7/tests/ignite/engine/test_engine.py#L1051

Maybe something like this:
```python
def test_iterator_state_output(self):
    ...
```
you can take a look for inspiration this test: test_faq_inf_iterator_with_epoch_length

tests/ignite/engine/test_engine.py
```diff
@@ -1279,6 +1279,44 @@ def check_iter_epoch(first_epoch_iter):
...
+            s = trainer.state
+            print(f"{s.epoch}/{s.max_epochs} : {s.iteration} - {batch.norm():.3f}")
+            return "flag_value"
+
+        trainer = Engine(train_step)
+        trainer.run(finite_iterator(4, 4), max_epochs=2)
+
+        assert trainer.state.output == "flag_value"
+        assert trainer.state.epoch == 2
+        # assert trainer.state.iteration == 2*4
```
mtauraso, line 1300: Found another bug: In the iterable case, the engine stops 1 epoch early, when its terminating due to StopIteration. 

Due to this comment, the test passes right now, but I haven't fixed the bug yet on this branch.

vfdev-5, line 1300: This assert does not pass?

mtauraso (commented):
I found another bug while testing this 🙃 

If there's an obvious fix I'd like to know, and would be happy for someone else to move the diff forward, but I probably won't have time to look in detail for a couple days.

If I find a fix I'll add it to the branch and ping again.

tests/ignite/engine/test_engine.py
mtauraso: Yeah, see my comment. Feel free to play with it. I might have time to look again tonight.

vfdev-5:
@mtauraso thanks for adding the tests. Let me reproduce the bug and make a separate ticket for it. This PR looks good to me and solves the initial problem. Let's land it and maybe work on the other bug in separate PR.

vfdev-5 (approved):
LGTM

tests/ignite/engine/test_engine.py
vfdev-5: I checked the test case and actually it is not a bug but kind of expected behavior with a finite-size input data iterator that was not manually restarted:
- `finite_iterator(4, 4)` can give only 4 batches and then StopIteration. This also defines the epoch length (as not provided in `.run(...)`). So, we increment the epoch counter 1 -> 2.
- As we do not restart the iterator manually, engine stops as it can't provide any data and also shows the following warning:
https://github.com/pytorch/ignite/blob/8d850e1f750281a14b304661eb5ff17dfbdad34f/ignite/engine/engine.py#L1114-L1118

mtauraso: Thanks, I've played around with the test a bit, and I think I understand the explanation.  Ironically, had I fully understood that I'm supposed to reset the iterator in an event handler, I would have never written the code that found this bug in the first place!

I have a more readable version of the tests in a branch ( https://github.com/mtauraso/ignite/tree/issue/3372-test-followup ), with the correct checking on the commented-out line.

Should I make another PR since this one is merged?

mtauraso: Now having read all comments I've put a PR here: https://github.com/pytorch/ignite/pull/3375

PR #3391, opened 2025-04-22
Roc curve tests available device

BanzaiTokyo:


vfdev-5:
@BanzaiTokyo there is a conflict to resolve for this PR

PR #3393, opened 2025-04-23
temporarily skip SR example if Caltech101 dataset fails to download

BanzaiTokyo:


vfdev-5 (approved):
Thanks @BanzaiTokyo !

PR #3398, opened 2025-04-30
replace conda with uv in mps-tests.yml

BanzaiTokyo:


.github/workflows/mps-tests.yml
```diff
@@ -42,64 +42,29 @@ jobs:
     timeout-minutes: 60
 
     steps:
-      - name: Clean workspace
```
vfdev-5, line 45: @BanzaiTokyo I do not think we should do that. To have valid MPS support we should use the current runner. GHA m1 runner does not have MPS enabled in torch, we've already seen together.

.github/workflows/binaries-nightly-release.yml
```diff
@@ -12,22 +12,21 @@ jobs:
...
 
-      - name: Setup Miniconda
-        uses: conda-incubator/setup-miniconda@v2
+      - name: Setup Python
+        uses: actions/setup-python@v4
         with:
-          miniconda-version: "latest"
           python-version: "3.10"
-
+      - name: Install Astral UV
```
vfdev-5, line 19: Use this GHA instead of manual installation: https://docs.astral.sh/uv/guides/integration/github/

.github/workflows/mps-tests.yml
```diff
@@ -42,64 +42,29 @@ jobs:
...
-          conda shell.bash hook
-          conda activate $CONDA_ENV
-          pip install --pre -U torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
+      - name: Install uv using GitHub Action
+        uses: astral-sh/setup-uv@v1
 
-      - name: Install dependencies
-        shell: bash -l {0}
-        working-directory: ${{ github.repository }}
+      - name: Set up virtual environment and install dependencies
```
vfdev-5, line 61: Do we really need to install venv, can we skip it ?

Issue #3350, opened 2025-03-25
Try to replace conda/setup-python with astral-uv if faster


vfdev-5:
Let's try to use https://docs.astral.sh/uv instead of conda in this CI job:
- [ ] https://github.com/pytorch/ignite/blob/41d75522eb36d4b9c55afea941ec53f42430315e/.github/workflows/pytorch-version-tests.yml
- [ ] https://github.com/pytorch/ignite/blob/41d75522eb36d4b9c55afea941ec53f42430315e/.github/workflows/unit-tests.yml

We can use GHA: actions/setup-uv

- **Each task should be a separate PR!**
- Please also provide in the description of the PR the total CI time in the PR vs same CI job time on master.
  - To start CI jobs on the fork, please check your forks GHA settings

Context:
- https://hugovk.dev/blog/2024/speed-up-ci-with-uv/ (we do not want to use tox however).

DharunKumar04:
Hi @vfdev-5 ,
is this still open to contribute ?

oxqnd:
HI @vfdev-5 , could you please assign me to this issue?

Stu-ops:
Hey @vfdev-5, if this issue is open, then can you assign it to me i want to fix this.

vfdev-5:
Closing this issue as done

PR #3401, opened 2025-05-01
uv in unit-tests.yml #3350

BanzaiTokyo:


.github/workflows/unit-tests.yml
```diff
@@ -57,43 +57,35 @@ jobs:
...
 
-      - name: Get year & week number
-        id: get-date
-        run: |
-          echo "date=$(/bin/date "+%Y-%U")" >> $GITHUB_OUTPUT
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
 
-      - name: Get pip cache dir
+      - name: Cache pip/uv dependencies
```
vfdev-5, line 68: @BanzaiTokyo check astral uv docs, they have a built-in feature to cache deps, no need to add a new step

.github/workflows/unit-tests.yml
```diff
@@ -57,43 +57,35 @@ jobs:
...
           restore-keys: |
-            ${{ steps.get-date.outputs.date }}-pytorch-${{ runner.os }}-${{ matrix.python-version }}-${{ matrix.pytorch-channel }}-
-
-      - run: pip install pip wheel setuptools -Uqq
+            ${{ runner.os }}-${{ matrix.python-version }}-${{ matrix.pytorch-channel }}-
 
       - name: Install PyTorch
         if: ${{ matrix.pytorch-channel == 'pytorch' }}
-        run: pip install torch torchvision --index-url https://download.pytorch.org/whl/cpu
+        run: uv pip install --system torch torchvision --index-url https://download.pytorch.org/whl/cpu
```
vfdev-5, line 79: I think you can set equivalent env var once instead of adding everywhere `--system`

.github/workflows/unit-tests.yml
```diff
@@ -32,6 +32,12 @@ concurrency:
 jobs:
   cpu-tests:
     runs-on: ${{ matrix.os }}
+    env:
+      UV_SYSTEM_PIP: 1
+      CLEARML_NO_DEFAULT_SERVER: 1
+      CLEARML_API_HOST: "http://localhost"
+      CLEARML_WEB_HOST: "http://localhost"
+      CLEARML_FILES_HOST: "http://localhost"
```
vfdev-5, line 40: Why this vs `CLEARML_OFFLINE_MODE=1`, https://clear.ml/docs/latest/docs/guides/set_offline/#setting-task-to-offline-mode

https://github.com/pytorch/ignite/blob/5caecf25b467e4e2c246818ced24829aaca7b7d7/ignite/handlers/clearml_logger.py#L152-L167

BanzaiTokyo: I don't quite understand that, but without 
      CLEARML_API_HOST: "http://localhost"
      CLEARML_WEB_HOST: "http://localhost"
      CLEARML_FILES_HOST: "http://localhost"
the following error is produced:
```  FAILED tests/ignite/contrib/engines/test_common.py::test_setup_clearml_logging - ValueError: ClearML configuration could not be found (missing `~/clearml.conf` or Environment CLEARML_API_HOST)```

File: .github/workflows/unit-tests.yml [hunk 4]
```diff
@@ -57,57 +63,34 @@ jobs:
...
-          pip install .
-          pip list
+          uv pip install -r requirements-dev.txt
+          uv pip install -e .
+          uv pip list
 
       - name: Check code formatting
         run: |
-          bash ./tests/run_code_style.sh install
+          uv pip install --upgrade flake8 "black==24.10.0" "usort==1.0.8.post1" "ufmt==2.7.3" "mypy"
```
vfdev-5, line 88: This is bad, once we need to upgrade the version of one of these tools we need to do that in +1 place now.
In future I wanted to remove `./tests/run_code_style.sh` and use `pre-commit` such that we can have only one file showing the used tools and the versions. But this is for another PR.

BanzaiTokyo: I understand that this is not ideal. If you have ideas for an alternative solution, I would gladly explore it

BanzaiTokyo: @vfdev-5 redone with pre-commit. seems to be working fine

vfdev-5:
Let's close this PR as migration to uv is done in other PRs

PR #3403, opened 2025-05-03
1 regression tests available device #3335

BanzaiTokyo:
test_canberra_metric.py 
test_fractional_absolute_error.py 
test_fractional_bias.py 
test_geometric_mean_absolute_error.py

### DIFF
--- a/ignite/metrics/regression/pearson_correlation.py
+++ b/ignite/metrics/regression/pearson_correlation.py
@@ -119,6 +119,12 @@ class PearsonCorrelation(_BaseRegression):
         y_var = self._sum_of_y_squares / n - y_mean * y_mean
         y_var = torch.clamp(y_var, min=0.0)
 
-        denom = torch.sqrt(torch.clamp(y_pred_var * y_var, min=self.eps))
+        denom = y_pred_var * y_var
+        denom = torch.clamp(denom, min=self.eps)
+        denom = torch.sqrt(denom)
         r = cov / denom
+
+        if torch.isnan(r):
+            return 0.0
+
         return float(r.item())
--- a/ignite/metrics/regression/pearson_correlation.py
+++ b/ignite/metrics/regression/pearson_correlation.py
@@ -87,11 +87,11 @@ class PearsonCorrelation(_BaseRegression):
 
     def _update(self, output: Tuple[torch.Tensor, torch.Tensor]) -> None:
         y_pred, y = output[0].detach(), output[1].detach()
-        self._sum_of_y_preds += y_pred.sum().to(self._device)
-        self._sum_of_ys += y.sum().to(self._device)
-        self._sum_of_y_pred_squares += y_pred.square().sum().to(self._device)
-        self._sum_of_y_squares += y.square().sum().to(self._device)
-        self._sum_of_products += (y_pred * y).sum().to(self._device)
+        self._sum_of_y_preds += y_pred.sum().to(dtype=self._double_dtype, device=self._device)
+        self._sum_of_ys += y.sum().to(dtype=self._double_dtype, device=self._device)
+        self._sum_of_y_pred_squares += y_pred.square().sum().to(dtype=self._double_dtype, device=self._device)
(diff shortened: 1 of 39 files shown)


ignite/metrics/regression/pearson_correlation.py
```diff
@@ -119,5 +119,12 @@ def compute(self) -> float:
...
         y_var = torch.clamp(y_var, min=0.0)
 
-        r = cov / torch.clamp(torch.sqrt(y_pred_var * y_var), min=self.eps)
+        denom = y_pred_var * y_var
+        denom = torch.clamp(denom, min=self.eps)
+        denom = torch.sqrt(denom)
+        r = cov / denom
+
+        if torch.isnan(r):
+            return 0.0
```
vfdev-5, line 128: Revert this, I do not think this should be done right now. I also do not see a test updates for this metric

tests/ignite/metrics/regression/test_canberra_metric.py
```diff
@@ -20,83 +20,83 @@ def test_wrong_input_shapes():
...
 
-    m = CanberraMetric()
+    m = CanberraMetric(device=available_device)
+    assert m._device == torch.device(available_device)
 
     canberra = DistanceMetric.get_metric("canberra")
 
-    m.update((torch.from_numpy(a), torch.from_numpy(ground_truth)))
-    np_sum = (np.abs(ground_truth - a) / (np.abs(a) + np.abs(ground_truth))).sum()
+    m.update((a, ground_truth))
```
vfdev-5, line 35: let's do a for-loop as you do in tests/ignite/metrics/regression/test_fractional_absolute_error.py

tests/ignite/metrics/regression/test_fractional_absolute_error.py
```diff
@@ -28,77 +28,65 @@ def test_wrong_input_shapes():
...
+    data = list(range(y_pred.shape[0] // batch_size))
+    fab = engine.run(data, max_epochs=1).metrics["fab"]
 
-    def get_test_cases():
-        test_cases = [
-            (torch.rand(size=(100,)), torch.rand(size=(100,)), 10),
-            (torch.rand(size=(100, 1)), torch.rand(size=(100, 1)), 20),
-        ]
-        return test_cases
+    np_sum = (2 * np.abs(np_y_pred - np_y) / (np.abs(np_y_pred) + np.abs(np_y))).sum()
```
vfdev-5, line 86: Let's use torch everywhere such that we could also remove `import numpy as np`

tests/ignite/metrics/regression/test_geometric_mean_absolute_error.py
```diff
@@ -28,47 +28,48 @@ def test_wrong_input_shapes():
...
 
-    m = GeometricMeanAbsoluteError()
-    m.update((torch.from_numpy(a), torch.from_numpy(ground_truth)))
+    m = GeometricMeanAbsoluteError(device=available_device)
+    assert m._device == torch.device(available_device)
+    m.update((a, ground_truth))
 
-    errors = np.abs(ground_truth - a)
-    np_prod = np.multiply.reduce(errors) * np_prod
+    errors = torch.abs(ground_truth - a)
```
vfdev-5, line 43: Let's write this using a for-loop

tests/ignite/metrics/regression/test_fractional_absolute_error.py
BanzaiTokyo: done, but numpy is still used in _test_distrib_compute

vfdev-5: You can do the same there: `error = 2 * torch.abs(pred - ground_truth) / (torch.abs(pred) + torch.abs(ground_truth))`

BanzaiTokyo: sure, but wouldn't it make more sense to do a separate PR to do a pass on distributed tests?

PR #3413, opened 2025-06-11
Update to use pyproject.toml only

goanpeca:
Remove the use of `setup.py` and migrate all to `pyproject.toml` and hatch.

The source of truth for the version of the package remains on `ignite/__init__.py`.

```toml
[build-system]
requires = ["hatchling>=1.24.2"]
build-backend = "hatchling.build"

[project]
name = "pytorch-ignite"
authors = [
    {name = "PyTorch-Ignite Team", email = "contact@pytorch-ignite.ai"},
]
...
```

Depends on:
- #3414


goanpeca:
Need to update the `sed` part for the nighlys but in the meantime could you approve the pending checks @vfdev-5 :) ?

Thanks!

pyproject.toml
```diff
@@ -1,3 +1,32 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "pytorch-ignite"
+version = "0.6.0"
```
vfdev-5: Can we fetch the version from `__init__.py` as a single source?

vfdev-5:
We may need to change here the conditions when unit tests are triggered: https://github.com/pytorch/ignite/blob/adbb260bfb1756c106fed2e793e1b645918ece27/.github/workflows/unit-tests.yml#L7-L23

pyproject.toml
goanpeca: So using `__version__` has been common for decades now, but it is not really standard. If we still want to have that we could do something like for backwards compatibility:

```python
from importlib.metadata import version
...
__version__ = version('pytorch-ignite')
```

goanpeca: We could if we moved to hatchling https://hatch.pypa.io/latest/version/, so we could also do that :) 

That could also help with some errors I am seeing now for building the docs on CI.

goanpeca: Hatch works like a charm. Will update the release docs as needed!

goanpeca:
@vfdev-5 could we change the netlify command to use so we can check this PR runs ok?

`pip install .` instead of `python setup.py install` ?

```
npm install katex && katex --version && pip install -U "networkx<3.2" numpy && pip install torch --extra-index-url https://download.pytorch.org/whl/cpu && python setup.py install && cd docs && pip install -r requirements.txt && make html
```

---

Besides the nigthly builds, which seem to be unrelated to this PR and related to precision differences, this one is ready for review.

CONTRIBUTING.md
```diff
@@ -102,7 +102,7 @@ upstream        https://github.com/pytorch/ignite (push)
 
 ```bash
 git pull upstream master
-python setup.py develop
+pip install .
```
vfdev-5: For development, we may need to add `-e` ?

goanpeca: correct!

goanpeca: Fixed!

.github/workflows/binaries-nightly-release.yml
```diff
@@ -1,6 +1,8 @@
 name: Nightly Releases
 
 on:
+  # Allows manual triggering of the workflow
+  workflow_dispatch:
```
vfdev-5: Let's remove this finally

goanpeca: Fixed!

.github/workflows/binaries-nightly-release.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           # workaround to fix https://github.com/pytorch/ignite/issues/2373
           pip uninstall -y twine pkginfo
           pip install --upgrade --no-cache-dir twine 'pkginfo>=1.8.2'
-          python setup.py sdist bdist_wheel
+          hatch build
```
vfdev-5, line 50: We can use `pip wheel --wheel-dir dist .` for that ?

goanpeca: Hatch build creates both wheel and sdist by default. Is the expected command to use.

Screenshot 2025-06-12 at 1 15 56 PM

goanpeca:
I know setuptools was more convenient with respect to the conda build meta yaml, cause we could load the setup.py data. Conda build should provide something similar for pyproject.toml but it does not 🤦🏼 .

I could create a small pre-commit local hook so that we make sure the `meta.yaml` file is always up to date from the `pyproject.toml` (in a separate PR). Let me know what you think @vfdev-5 

---

I tested the rest of the conda and pypi package creation locally step by step, to make sure all was working as expected.

conda.recipe/meta.yaml
```diff
@@ -1,27 +1,29 @@
-{% set data = load_setup_py_data() %}
+{% set __version__ = "0.6.0" %}
```
vfdev-5, line 1: OK, I understand now what you were talking about when saying that conda recipe stuff support less pyproject. Can we write some bash or python script in this field `{% set variable = script() %}` to fetch the version from `__init__.py` ?
Maybe, we can provide this in a follow-up PR.

.gitignore
```diff
@@ -44,3 +44,4 @@ coverage.xml
 # Virtualenv
 .venv/
 .python-version
+conda_build/
```
vfdev-5:
```suggestion
conda_build/

```

vfdev-5 (approved):
Thanks a lot for the PR, Gonzalo!
Let's land it and see how it works

PR #3430, opened 2025-06-18
Update docker images

goanpeca:
Update docker images

### DIFF
--- a/docker/docker.cfg
+++ b/docker/docker.cfg
@@ -1,4 +1,4 @@
 [DEFAULT]
-build_docker_image_pytorch_version = 2.6.0-cuda12.4-cudnn9
+build_docker_image_pytorch_version = 2.7.0-cuda12.6-cudnn9
 build_docker_image_hvd_version = v0.28.1
 build_docker_image_msdp_version = v0.14.0


goanpeca:
Hi @vfdev-5, I think this one is ready for review :) 

---


Should I check if everything builds with cuda 12.8 and update to use that? 🤔

docker/README.md
```diff
@@ -64,6 +64,11 @@ Available Tensor Operations:
 docker run -it -v $PWD:/workspace/project --network=host --ipc=host pytorchignite/base:latest /bin/bash
 ```
 
+
+## Images
+
+You can see all available pytorch images on [docker hub](https://hub.docker.com/r/pytorch/pytorch).
+
```
vfdev-5: I do not think we need this addition

docker/README.md
```diff
@@ -72,3 +77,13 @@ Dockerfiles are supplied to build images with dependencies that ease the use of
...
+### Building the images for horovod
+
+You can locally build a docker image by running the `build.sh`.
+
+For runnning the horovod base image you can run:
+
+```bash
+PTH_VERSION="2.7.1-cuda12.6-cudnn9" HVD_VERSION="v0.28.1" bash build.sh hvd hvd-base
+```
```
vfdev-5: Same here, let's revert this.

docker/hvd/Dockerfile.hvd-apex
```diff
@@ -25,9 +25,12 @@ ARG HVD_VERSION
 
 # Build Horovod
 RUN apt-get update && apt-get install -y git && \
-        git clone --recursive --depth 1 --branch ${HVD_VERSION} https://github.com/horovod/horovod.git /horovod && \
-        conda install -y cmake nccl -c conda-forge && \
+        conda install -y cmake=3.* nccl scikit-build -c conda-forge && \
```
vfdev-5: I'm surprised a bit seeing scikit-build as dependency, do we really need it?

vfdev-5 (commented):
Thanks, Gonzalo! 
I left a comment and questions

docker/README.md
goanpeca: Reverting!

docker/README.md
goanpeca: Reverting!

docker/hvd/Dockerfile.hvd-apex
goanpeca: I was seeing a build error pecause that was missing on my initial tests, but it was probably to the branch no being the right one at checkout. 

Will revert!

docker/docker.cfg
```diff
@@ -1,4 +1,4 @@
 [DEFAULT]
-build_docker_image_pytorch_version = 2.6.0-cuda12.4-cudnn9
+build_docker_image_pytorch_version = 2.7.1-cuda12.8-cudnn9
```
vfdev-5: It was on purpose to keep not too recent cuda version as user's host machine driver may not be that recent and this would require to upgrade the drivers etc to use the image.
If we can set it to 12.4 it is good, if pytorch image has 12.6 as minimal version, let's use that.

docker/hvd/Dockerfile.hvd-base
```diff
@@ -3,15 +3,18 @@
...
 # Build Horovod
 RUN apt-get update && apt-get install -y git && \
-        git clone --recursive --depth 1 --branch ${HVD_VERSION} https://github.com/horovod/horovod.git /horovod && \
-        conda install -y cmake nccl -c conda-forge && \
+        conda install -y cmake=3.* nccl -c conda-forge && \
+        git clone --depth 1 https://github.com/horovod/horovod.git /horovod && \
         cd /horovod && \
+        git fetch --tags && \
+        git checkout ${HVD_VERSION} && \
+        git submodule update --init --recursive && \
```
vfdev-5, line 17: Can't we do that in one line as previously?

docker/docker.cfg
goanpeca: Ok, 6 it is !

docker/hvd/Dockerfile.hvd-base
goanpeca: The command was not working as expected I tried different variations locally and this is a bit longer, but it results in the correct branch being checked out.

docker/docker.cfg
goanpeca: @vfdev-5 for the 2.7.1 version we have images with

2.7.1-cuda11.8-cudnn9-...
2.7.1-cuda12.6-cudnn9-...
2.7.1-cuda12.8-cudnn9-...

So minimal version is 11.8 for the pytroch images.

vfdev-5: 11.8 is rather old, let's use 12.6

vfdev-5 (approved):
Thanks Gonzalo!

PR #3440, opened 2025-09-02
Add SAVED_CHECKPOINT event to Checkpoint handler

### LINKED ISSUE #934: Add new events
## 🚀 Feature

Idea is to add supplementary events to Checkpoint/ModelCheckpoint, EarlyStopping  and maybe some others:
```python
class Checkpoint:

    def __call__(self, engine):
        # ...
        self.save_handler(checkpoint, filename)
        # TRIGGER new EVENT
        engine.fire_event(Checkpoint.SAVED_CHECKPOINT)
```

JeevanChevula:
Fixes #934

This PR adds a "saved_checkpoint" event that fires after successful checkpoint saves.

**Usage:**
```python
# Users need to register the event first
engine.register_events("saved_checkpoint")

@trainer.on("saved_checkpoint")
def after_saving(engine):
    print("Checkpoint saved!")

### DIFF
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -37,7 +37,6 @@ Complete list of generic handlers
 Checkpoint Events
 -----------------
 
-.. versionadded:: 0.5.0
 
 The Checkpoint handler provides a ``SAVED_CHECKPOINT`` event that fires after successful checkpoint saves.
 This allows users to attach custom handlers that react to checkpoint operations without manual event registration.
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -34,6 +34,37 @@ Complete list of generic handlers
     state_param_scheduler.StateParamScheduler
 
 
+Checkpoint Events
+-----------------
+
+.. versionadded:: 0.5.0
+
+The Checkpoint handler provides a ``SAVED_CHECKPOINT`` event that fires after successful checkpoint saves.
+This allows users to attach custom handlers that react to checkpoint operations without manual event registration.
+
+**Usage:**
+
+.. code-block:: python
+
+    from ignite.handlers import Checkpoint
+    from ignite.engine import Engine, Events
+    
+    # Setup checkpoint handler
+    checkpoint_handler = Checkpoint(
+        {'model': model, 'optimizer': optimizer}, 
+        save_dir,
+        n_saved=2
+    )
+    
+    # Attach handler to checkpoint event (no manual registration needed)
+    @trainer.on(Checkpoint.SAVED_CHECKPOINT)
+    def on_checkpoint_saved(engine):
+        print(f"Checkpoint saved at epoch {engine.state.epoch}")
+        # Add custom logic: notifications, logging, etc.
+    
+    trainer.add_event_handler(Events.EPOCH_COMPLETED, checkpoint_handler)
+
+
 Loggers
 --------
 
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -11,6 +11,7 @@ Complete list of generic handlers
     :toctree: generated
 
     checkpoint.Checkpoint
+    checkpoint.CheckpointEvents
     DiskSaver
     checkpoint.ModelCheckpoint
     ema_handler.EMAHandler
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -35,36 +35,6 @@ Complete list of generic handlers
     state_param_scheduler.StateParamScheduler
 
 
-Checkpoint Events
------------------
-
-
-The Checkpoint handler provides a ``SAVED_CHECKPOINT`` event that fires after successful checkpoint saves.
-This allows users to attach custom handlers that react to checkpoint operations without manual event registration.
-
-**Usage:**
-
-.. code-block:: python
-
-    from ignite.handlers import Checkpoint
-    from ignite.engine import Engine, Events
-    
-    # Setup checkpoint handler
-    checkpoint_handler = Checkpoint(
-        {'model': model, 'optimizer': optimizer}, 
-        save_dir,
-        n_saved=2
-    )
-    
-    # Attach handler to checkpoint event (no manual registration needed)
-    @trainer.on(Checkpoint.SAVED_CHECKPOINT)
-    def on_checkpoint_saved(engine):
-        print(f"Checkpoint saved at epoch {engine.state.epoch}")
-        # Add custom logic: notifications, logging, etc.
-    
-    trainer.add_event_handler(Events.EPOCH_COMPLETED, checkpoint_handler)
-
-
 Loggers
 --------
 
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -28,8 +28,10 @@ __all__ = ["Checkpoint", "DiskSaver", "ModelCheckpoint", "BaseSaveHandler", "Che
 
 
 class CheckpointEvents(EventEnum):
-    """Events fired by Checkpoint handler
-
+    """Events fired by :class:`~ignite.handlers.checkpoint.Checkpoint`
+    
+    - SAVED_CHECKPOINT : triggered when checkpoint handler has saved objects
+    
     .. versionadded:: 0.5.3
     """
 
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -275,6 +275,29 @@ class Checkpoint(Serializable):
                 to_save, save_handler=DiskSaver('/tmp/models', create_dir=True, **kwargs), n_saved=2
             )
 
+        Respond to checkpoint events:
+
+        .. code-block:: python
+
+            from ignite.handlers import Checkpoint
+            from ignite.engine import Engine, Events
+
+            checkpoint_handler = Checkpoint(
+                {'model': model, 'optimizer': optimizer},
+                save_dir,
+                n_saved=2
+            )
+
+            @trainer.on(Checkpoint.SAVED_CHECKPOINT)
+            def on_checkpoint_saved(engine):
+                print(f"Checkpoint saved at epoch {engine.state.epoch}")
+
+            trainer.add_event_handler(Events.EPOCH_COMPLETED, checkpoint_handler)
+
+    Attributes:
+        SAVED_CHECKPOINT: Alias of ``SAVED_CHECKPOINT`` from
+            :class:`~ignite.handlers.checkpoint.CheckpointEvents`.
+
     .. versionchanged:: 0.4.3
 
         - Checkpoint can save model with same filename.
@@ -286,7 +309,7 @@ class Checkpoint(Serializable):
         - `save_handler` automatically saves to disk if path to directory is provided.
         - `save_on_rank` saves objects on this rank in a distributed configuration.
 
-    .. 
+    ..
     """
 
     SAVED_CHECKPOINT = CheckpointEvents.SAVED_CHECKPOINT
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -21,10 +21,16 @@ else:
 
 import ignite.distributed as idist
 from ignite.base import Serializable
-from ignite.engine import Engine, Events
+from ignite.engine import Engine, Events, EventEnum
 from ignite.utils import _tree_apply2, _tree_map
 
-__all__ = ["Checkpoint", "DiskSaver", "ModelCheckpoint", "BaseSaveHandler"]
+__all__ = ["Checkpoint", "DiskSaver", "ModelCheckpoint", "BaseSaveHandler", "CheckpointEvents"]
+
+
+class CheckpointEvents(EventEnum):
+    """Events fired by Checkpoint handler"""
+
+    SAVED_CHECKPOINT = "saved_checkpoint"
 
 
 class BaseSaveHandler(metaclass=ABCMeta):
@@ -276,6 +282,7 @@ class Checkpoint(Serializable):
         - `save_on_rank` saves objects on this rank in a distributed configuration.
     """
 
+    SAVED_CHECKPOINT = CheckpointEvents.SAVED_CHECKPOINT
     Item = NamedTuple("Item", [("priority", int), ("filename", str)])
     _state_dict_all_req_keys = ("_saved",)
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by vfdev-5** [comment]
@JeevanChevula thanks for the PR. However, let's rework the API of the new feature you are working on:

1) Let's avoid using string type as event name and do not ask users to register the event. This should be done internally by ignite.

2) Let's define the event in the Checkpoint as following:
```python
# checkpoint.py

class CheckpointEvents(EventEnum):
    SAVED_CHECKPOINT = "saved_checkpoint"


class Checkpoint(...):
    SAVED_CHECKPOINT = CheckpointEvents.SAVED_CHECKPOINT
    ...
```

3) finally the usage can be something like this:
```python
from ignite.handlers import Checkpoint, global_step_from_engine

trainer = ...
evaluator = ...
# Setup Accuracy metric computation on evaluator.
# evaluator.state.metrics contain 'accuracy',
# which will be used to define ``score_function`` automatically.
# Run evaluation on epoch completed event
# ...

to_save = {'model': model}
handler = Checkpoint(
    to_save, '/tmp/models',
    n_saved=2, filename_prefix='best',
    score_name="accuracy",
    global_step_transform=global_step_from_engine(trainer)
)

# ---- New API with Checkpoint.SAVED_CHECKPOINT event: -----

@evaluator.on(Checkpoint.SAVED_CHECKPOINT)
def notify_when_saved(eval_engine, chkpt_handler):   # we should pass to the attached handlers the engine and the checkpoint instance.
    assert eval_engine is engine
    assert chkpt_handler is handler
    print("Saved checkpoint:", chkpt_handler.last_checkpoint)
    
# ---- End of New API with Checkpoint.SAVED_CHECKPOINT event: -----

trainer.run(data_loader, max_epochs=10)
> ["best_model_9_accuracy=0.77.pt", "best_model_10_accuracy=0.78.pt", ]
```


Let me know what do you think?

JeevanChevula:
Thanks for the suggestion . I’ll try to work on updating the PR to follow the API approach you mentioned with

JeevanChevula:
**Implementation Note:**

Implemented EventEnum-based SAVED_CHECKPOINT event as requested. However, Ignite's event system only supports single-parameter handlers - the originally requested two-parameter signature (handler(engine, checkpoint_handler)) failed during event firing and registration.

Current implementation uses single parameter with checkpoint access via engine._current_checkpoint_handler. All 61 core tests pass, confirming functionality works without breaking existing features. The 3 distributed test errors are pre-existing infrastructure issues unrelated to this change.

ignite/handlers/checkpoint.py
```diff
@@ -460,11 +470,15 @@ def __call__(self, engine: Engine) -> None:
             if self.include_self:
                 # Now that we've updated _saved, we can add our own state_dict.
                 checkpoint["checkpointer"] = self.state_dict()
-
+            # Store reference to self in engine for event handlers to access    
+            engine._current_checkpoint_handler = self
```
vfdev-5: Can you explain this code? I do not understand why we need this.

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -1850,6 +1850,7 @@ def test_load_single_object(obj_to_save, dirname):
     Checkpoint.load_objects(to_load=to_save, checkpoint=str(checkpoint_fp))
 
 
+
```
vfdev-5, line 1885: Please add real tests

ignite/handlers/checkpoint.py
```diff
@@ -72,6 +75,7 @@ def remove(self, filename: str) -> None:
 
 
 class Checkpoint(Serializable):
+
```
vfdev-5: Let's revert code modifications like this one: new empty line etc.

vfdev-5 (commented):
Thanks for working on this PR @JeevanChevula 
I left few more comments to improve the PR

ignite/handlers/checkpoint.py
JeevanChevula: This is a workaround for Ignite's event system limitation. You originally requested a two-parameter handler signature `handler(engine, checkpoint_handler)`, but Ignite's `fire_event()` only supports single parameters and rejects handlers expecting additional arguments. This line stores the checkpoint reference in the engine so handlers can access it via `engine._current_checkpoint_handler`.

tests/ignite/handlers/test_checkpoint.py
JeevanChevula: Added comprehensive test coverage for SAVED_CHECKPOINT event functionality in test_checkpoint_saved_event()

ignite/handlers/checkpoint.py
JeevanChevula: Cleaned up unnecessary blank lines and formatting issues

ignite/handlers/checkpoint.py
vfdev-5: I have to check details about this limitation. Unfortunately `engine._current_checkpoint_handler` is not good enough for a public API usage.

Alternatively, we can pass the instance of the checkpointer as an arg when attaching:
```python
checkpoint = Checkpoint(...)

@trainer.on(Checkpoint. SAVED_CHECKPOINT, checkpoint)
def handler(engine, chkpt_handler):
    assert engine is trainer
    assert chkpt_handler is checkpoint
```
Maybe, we can skip automatic `chkpt_handler` arg injection:
````python
checkpoint = Checkpoint(...)

@trainer.on(Checkpoint. SAVED_CHECKPOINT)
def handler(engine):
    assert engine is trainer
```

tests/ignite/handlers/test_checkpoint.py
```diff
@@ -1850,6 +1850,46 @@ def test_load_single_object(obj_to_save, dirname):
...
+    save_handler = MagicMock(spec=BaseSaveHandler)
+    to_save = {"model": DummyModel()}
+
+    checkpointer = Checkpoint(to_save, save_handler=save_handler, n_saved=2)
+
+    trainer = Engine(lambda e, b: None)
+    trainer.state = State(epoch=0, iteration=0)
+
+    # Register the event first
+    trainer.register_events(CheckpointEvents.SAVED_CHECKPOINT)
```
vfdev-5: We should not register the event ourselves

ignite/handlers/checkpoint.py
```diff
@@ -400,6 +407,10 @@ def _compare_fn(self, new: Union[int, float]) -> bool:
             return new > self._saved[0].priority
 
     def __call__(self, engine: Engine) -> None:
+        # Register the custom event if not already registered
+        if not hasattr(engine, "_checkpoint_events_registered"):
+            engine.register_events(CheckpointEvents.SAVED_CHECKPOINT)
+            engine._checkpoint_events_registered = True
```
vfdev-5: I'm finally hesitating about adding this private attribute `_checkpoint_events_registered` to the engine. I think of separating concerns, Engine is self-contained class and by default does not know about Checkpoint class. On the other hand Checkpoint is aware of Engine class, so it could either query the Engine whether it has already the CheckpointEvents.SAVED_CHECKPOINT event or keep a table of mappings engine instance -> true/false.
Let's do the following, in a separate PR let's add a method to Engine to check whether it has an event: `Engine.has_registered_events(event: Any) -> bool`. Once done, in this PR we can use it.

tests/ignite/handlers/test_checkpoint.py
JeevanChevula: Understood. I'll remove the manual event registration from the test once the proper Engine method is available

ignite/handlers/checkpoint.py
JeevanChevula: I'll create a separate PR to add the Engine.has_registered_events(event) -> bool method. Once that's merged, I'll update this PR to use that method instead of the private attribute approach. Should I proceed with creating the separate PR?

JeevanChevula: I've created a separate PR to add the Engine.has_registered_events(event) -> bool method as you suggested. Once that's merged, I'll update this PR to use that method instead of the private attribute approach and remove the manual event registration from the test.

JeevanChevula:
Pushing current implementation with working SAVED_CHECKPOINT event functionality. Will add proper Google-style docstrings with version directives by Monday per contributing guidelines

vfdev-5:
@JeevanChevula please rebase your PR branch, you have now some extra commits

ignite/handlers/checkpoint.py
vfdev-5: @JeevanChevula please remove this private attribute

docs/source/handlers.rst
```diff
@@ -34,6 +34,37 @@ Complete list of generic handlers
     state_param_scheduler.StateParamScheduler
 
 
+Checkpoint Events
+-----------------
+
+.. versionadded:: 0.5.0
```
vfdev-5: This should be added to the docstrings. Can you locally render the docs and show how this look like?
```suggestion
.. versionadded:: 0.5.3
```

ignite/handlers/checkpoint.py
```diff
@@ -21,10 +21,16 @@
...
-from ignite.engine import Engine, Events
+from ignite.engine import Engine, Events, EventEnum
 from ignite.utils import _tree_apply2, _tree_map
 
-__all__ = ["Checkpoint", "DiskSaver", "ModelCheckpoint", "BaseSaveHandler"]
+__all__ = ["Checkpoint", "DiskSaver", "ModelCheckpoint", "BaseSaveHandler", "CheckpointEvents"]
+
+
+class CheckpointEvents(EventEnum):
+    """Events fired by Checkpoint handler"""
```
```suggestion
    """Events fired by Checkpoint handler
    
    .. versionadded:: 0.5.0
    """
```

JeevanChevula: Thanks for the feedback 🙏

I will update checkpoint.py so that the CheckpointEvents class docstring includes the version:

```
class CheckpointEvents(EventEnum):
    """Events fired by Checkpoint handler
    
    .. versionadded:: 0.5.3
    """
    SAVED_CHECKPOINT = `"saved_checkpoint"
````


One clarification: should I completely remove the Checkpoint Events section from docs/source/handlers.rst, or just update the version there from 0.5.0 → 0.5.3?

Comment (vfdev-5) [reply ignite/handlers/checkpoint.py]: > One clarification: should I completely remove the Checkpoint Events section from docs/source/handlers.rst, or just update the version there from 0.5.0 → 0.5.3?

Do not remove it. Change what is suggested and build and visualize the docs locally. I think `versionadded` part of docs:
```
Checkpoint Events
-----------------

.. versionadded:: 0.5.3
```
wont be rendered correctly. In this case just remove `.. versionadded:: 0.5.3` and keep everything else

Comment (JeevanChevula) [reply ignite/handlers/checkpoint.py]: I've updated the CheckpointEvents docstring with the version directive and kept the RST documentation without the versionadded line as suggested. The local docs build hits a Windows-specific environment issue at 97% completion (not related to our changes - the RST parsing completed successfully).However, the documentation syntax appears correct based on the successful parsing before the build failure.

File: ignite/handlers/checkpoint.py [hunk 8]
```diff
@@ -400,6 +410,8 @@ def _compare_fn(self, new: Union[int, float]) -> bool:
             return new > self._saved[0].priority
 
     def __call__(self, engine: Engine) -> None:
+        if not engine.has_registered_events(CheckpointEvents.SAVED_CHECKPOINT):
+            engine.register_events(CheckpointEvents.SAVED_CHECKPOINT)
```
```suggestion
            engine.register_events(*CheckpointEvents)
```

vfdev-5:
@JeevanChevula here is how docs are rendered for this PR: https://deploy-preview-3440--pytorch-ignite-preview.netlify.app/handlers

Thanks for all the updates you made recently. The last thing I think we still have to do is to reorganize a bit the docs on CheckpointEvents. 

The example you wrote in handlers.rst: https://github.com/pytorch/ignite/pull/3440/files#diff-fed20f17bf0c40747938f76730fe5cdc467c919bd185eeb9fe0870b861379681R38 should be moved into `Checkpoint` docstring. And this page: https://deploy-preview-3440--pytorch-ignite-preview.netlify.app/generated/ignite.handlers.checkpoint.checkpoint#ignite.handlers.checkpoint.Checkpoint should mention the attribute `Checkpoint.SAVED_CHECKPOINT`: e.g. alias of ``SAVED_CHECKPOINT`` from :class:`~ignite.handlers.checkpoint.CheckpointEvents`.

JeevanChevula:
Hi! Thank you for the detailed feedback on the documentation reorganization

I want to make sure I implement the changes exactly as you envision them. Three specific questions for clarification:

1. **Checkpoint class docstring placement** (class Checkpoint(Serializable):)
    The Checkpoint class already has a long docstring with multiple Examples: sections.
      Should I add the checkpoint event usage example:

      * After the last existing example (the “Customise the save_handler” section) but before the .. versionchanged:: notes?
      * Or in a different location within the docstring?
    
2. **handlers.rst reorganization**  
     For the Checkpoint Events section in handlers.rst, should I:
    
     * Remove the entire section completely (since the example will be moved to the docstring)?
     * Or keep the heading and basic description, but remove only the detailed usage example?

3.**SAVED_CHECKPOINT attribute documentation**
        For documenting "alias of SAVED_CHECKPOINT from CheckpointEvents", would you prefer that I:

     * Add it in an Attributes: section inside the Checkpoint docstring?
     * Or simply put a short comment/docstring next to the line SAVED_CHECKPOINT = CheckpointEvents.SAVED_CHECKPOINT?
     * Or another location entirely?
     


I’d like to follow your vision precisely, so confirming these points will help me make the changes correctly.

vfdev-5:
Here are details on your questions:

Point 1: -> After the last existing example (the “Customise the save_handler” section) but before the .. versionchanged:: notes

Point 2: -> Remove the entire section completely (since the example will be moved to the docstring)?

Point 3: -> Add it in an Attributes: section inside the Checkpoint docstring 
```
Attributes:
    SAVED_CHECKPOINT: alias of SAVED_CHECKPOINT from :class:~ignite.handlers.checkpoint.CheckpointEvents.
```

Make sure to check rendered docs (on netlify: https://deploy-preview-3440--pytorch-ignite-preview.netlify.app/)

JeevanChevula:
Hi!

I’ve applied the requested changes:

- Moved the checkpoint event usage example into the Checkpoint class docstring (after “Customise the save_handler” and before the .. versionchanged:: notes).

- Added an Attributes: section documenting SAVED_CHECKPOINT as an alias of    :class:~ignite.handlers.checkpoint.CheckpointEvents``.

- Removed the “Checkpoint Events” section from handlers.rst and kept only the API references.

On my Windows machine I wasn’t able to fully render the docs locally due to a Sphinx subprocess issue in my environment. I may be missing a dependency—I'll revisit this when I’m back. In the meantime, the Netlify preview should reflect these changes; if anything doesn’t look right, I’m happy to adjust.

I’ll be OOO  ( Sep 26 – Oct 5, IST]. Please feel free to leave comments or push small cleanups; I’ll pick up any remaining fixes as soon as I return. Thanks!

vfdev-5 (approved):
LGTM, thanks @JeevanChevula !

PR #3442, opened 2025-09-04
Add typing and style checks as separate workflows

goanpeca:
Add typing and style checks as separate workflows


vfdev-5:
@goanpeca should we also remove mypy check step from unit-tests.yml: https://github.com/pytorch/ignite/blob/400973d452a7d1e66da853811cd15b1fdde77621/.github/workflows/unit-tests.yml#L100-L105

goanpeca:
>should we also remove mypy check step from unit-tests.yml:

Yes I will remove that!

PR #3444, opened 2025-09-04
Fix MPS

goanpeca:
Fix MPS by detaching before moving to cpu


vfdev-5:
@goanpeca please post here the error you see without this PR to understand better why we need now to call detach

goanpeca:
@vfdev-5 it was a suggestion from copilot on the first PR to add the detach method before the CPU.

>Added .detach() calls before .cpu() for mu1, mu2, sigma1, and sigma2 tensors to prevent gradient tracking issues when 
moving from GPU to CPU

>Detach the tensor from the computational graph (if it requires gradients):
>If the tensor was created with requires_grad=True or is part of a computation where gradients are being tracked, you need to detach it from the computational graph using the .detach() method. This prevents errors related to attempting to convert a tensor that requires gradients to a NumPy array, as NumPy arrays do not have a concept of gradients.

Is this something that could apply to mu1, mu2, sigma1, sigma2?

goanpeca:
Also, with the fix you suggested, we still run into de same error, so I should revert to using my original fix?

```python
    def test_compute_fid_sqrtm():
        mu1 = torch.tensor([0, 0])
        mu2 = torch.tensor([0, 0])
    
        sigma1 = torch.tensor([[-1, 1], [1, 1]], dtype=torch.float64)
        sigma2 = torch.tensor([[1, 0], [0, 1]], dtype=torch.float64)
    
        with pytest.raises(ValueError, match=r"Imaginary component "):
>           fid_score(mu1, mu2, sigma1, sigma2)

tests/ignite/metrics/gan/test_fid.py:102: 
ignite/metrics/gan/fid.py:39: in fid_score
    covmean, _ = scipy.linalg.sqrtm(sigma1.mm(sigma2), disp=False)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

A = tensor([[-1.,  1.],
        [ 1.,  1.]], dtype=torch.float64), disp = False
blocksize = <object object at 0x1127520e0>

    def sqrtm(A, disp=_NoValue, blocksize=_NoValue):
        """
        Compute, if exists, the matrix square root.
    
        The matrix square root of ``A`` is a matrix ``X`` such that ``X @ X = A``.
        Every square matrix is not guaranteed to have a matrix square root, for
        example, the array ``[[0, 1], [0, 0]]`` does not have a square root.
    
        Moreover, not every real matrix has a real square root. Hence, for
        real-valued matrices the return type can be complex if, numerically, there
        is an eigenvalue on the negative real axis.
    
        Parameters
        ----------
        A : ndarray
            Input with last two dimensions are square ``(..., n, n)``.
        disp : bool, optional
            Print warning if error in the result is estimated large
            instead of returning estimated error. (Default: True)
    
            .. deprecated:: 1.16.0
                The `disp` argument is deprecated and will be
                removed in SciPy 1.18.0. The previously returned error estimate
                can be computed as ``norm(X @ X - A, 'fro')**2 / norm(A, 'fro')``
    
        blocksize : integer, optional
    
            .. deprecated:: 1.16.0
                The `blocksize` argument is deprecated as it is unused by the algorithm
                and will be removed in SciPy 1.18.0.
    
        Returns
        -------
        sqrtm : ndarray
            Computed matrix squareroot of `A` with same size ``(..., n, n)``.
    
        errest : float
            Frobenius norm of the estimated error, ||err||_F / ||A||_F. Only
            returned, if ``disp`` is set to ``False``. This return argument will be
            removed in version 1.20.0 and only the sqrtm result will be returned.
    
            .. deprecated:: 1.16.0
    
        Notes
        -----
        This function uses the Schur decomposition method to compute the matrix
        square root following [1]_ and for real matrices [2]_. Moreover, note
        that, there exist matrices that have square roots that are not polynomials
        in ``A``. For a classical example from [2]_, the matrix satisfies::
    
                [ a, a**2 + 1]**2     [-1,  0]
                [-1,       -a]     =  [ 0, -1]
    
        for any scalar ``a`` but it is not a polynomial in ``-I``. Thus, they will
        not be found by this function.
    
        References
        ----------
        .. [1] Edvin Deadman, Nicholas J. Higham, Rui Ralha (2013)
               "Blocked Schur Algorithms for Computing the Matrix Square Root,
               Lecture Notes in Computer Science, 7782. pp. 171-182.
               :doi:`10.1016/0024-3795(87)90118-2`
        .. [2] Nicholas J. Higham (1987) "Computing real square roots of a real
               matrix", Linear Algebra and its Applications, 88/89:405-430.
               :doi:`10.1016/0024-3795(87)90118-2`
    
        Examples
        --------
        >>> import numpy as np
        >>> from scipy.linalg import sqrtm
        >>> a = np.array([[1.0, 3.0], [1.0, 4.0]])
        >>> r = sqrtm(a)
        >>> r
        array([[ 0.75592895,  1.13389342],
               [ 0.37796447,  1.88982237]])
        >>> r.dot(r)
        array([[ 1.,  3.],
               [ 1.,  4.]])
    
        """
        if disp is _NoValue:
            disp = True
        else:
            warnings.warn("The `disp` argument is deprecated and will be removed in SciPy "
                          "1.18.0.",
                          DeprecationWarning, stacklevel=2)
        if blocksize is not _NoValue:
            warnings.warn("The `blocksize` argument is deprecated and will be removed in "
                          "SciPy 1.18.0.",
                          DeprecationWarning, stacklevel=2)
    
        a = np.asarray(A)
        if a.size == 1 and a.ndim < 2:
            return np.array([[np.exp(a.item())]])
    
        if a.ndim < 2:
            raise LinAlgError('The input array must be at least two-dimensional')
        if a.shape[-1] != a.shape[-2]:
            raise LinAlgError('Last 2 dimensions of the array must be square')
    
        # Empty array
        if min(*a.shape) == 0:
            dtype = sqrtm(np.eye(2, dtype=a.dtype)).dtype
            return np.empty_like(a, dtype=dtype)
    
        # Scalar case
        if a.shape[-2:] == (1, 1):
            return np.emath.sqrt(a)
    
        if not np.issubdtype(a.dtype, np.inexact):
            a = a.astype(np.float64)
        elif a.dtype == np.float16:
            a = a.astype(np.float32)
        elif a.dtype.char in 'G':
            a = a.astype(np.complex128)
        elif a.dtype.char in 'g':
            a = a.astype(np.float64)
    
        if a.dtype.char not in 'fdFD':
            raise TypeError("scipy.linalg.sqrtm is not supported for the data type"
                            f" {a.dtype}")
    
        res, isIllconditioned, isSingular, info = recursive_schur_sqrtm(a)
        if info < 0:
            raise LinAlgError(f"Internal error in scipy.linalg.sqrtm: {info}")
    
        if isSingular or isIllconditioned:
            if isSingular:
                msg = ("Matrix is singular. The result might be inaccurate or the"
                       " array might not have a square root.")
            else:
                msg = ("Matrix is ill-conditioned. The result might be inaccurate"
                       " or the array might not have a square root.")
            warnings.warn(msg, LinAlgWarning, stacklevel=2)
    
        if disp is False:
            try:
>               arg2 = norm(res @ res - A, 'fro')**2 / norm(A, 'fro')
                            ^^^^^^^^^^^^^
E               TypeError: unsupported operand type(s) for -: 'numpy.ndarray' and 'Tensor'
```

goanpeca:
@vfdev-5 the error was something on 
```
tests/ignite/metrics/gan/test_fid.py:51: 
ignite/metrics/gan/fid.py:40: in fid_score
    covmean, _ = scipy.linalg.sqrtm(sigma1.mm(sigma2).cpu(), disp=False)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

...
          if disp is False:
              try:
  >               arg2 = norm(res @ res - A, 'fro')**2 / norm(A, 'fro')
                              ^^^^^^^^^^^^^
  E               TypeError: unsupported operand type(s) for -: 'numpy.ndarray' and 'Tensor'
  
  .venv/lib/python3.11/site-packages/scipy/linalg/_matfuncs.py:547: TypeError
```

so that is why we had to add a .numpy() call. For the detach see previous message, it was a suggestion from copilot, but if it does not apply, I can revert it!

vfdev-5 (approved):
Thanks Gonzalo!

PR #3449, opened 2025-09-09
Add has_registered_events method to Engine class

JeevanChevula:
Adds a new method `has_registered_events(event) -> bool` to the Engine class to check whether a specific event is registered with the engine.

This method will be used by the Checkpoint handler to determine if custom events need to be registered, improving separation of concerns between Engine and Checkpoint classes.

Related to PR #3440 (checkpoint events feature)


ignite/engine/engine.py
```diff
@@ -248,6 +248,17 @@ class TBPTT_Events(EventEnum):
...
+    def has_registered_events(self, event: Any) -> bool:
+    """Check whether engine has a registered event.
+    
+    Args:
+        event: Event to check for registration.
+        
+    Returns:
+        bool: True if the event is registered, False otherwise.
+    """
+    return event in self._allowed_events
```
vfdev-5: Indent issue

vfdev-5 (commented):
Thanks for the quick PR @JeevanChevula 
Please also write tests with built-in events in [test_engine.py](https://github.com/pytorch/ignite/blob/master/tests/ignite/engine/test_engine.py) and custom events in [test_custom_events.py](https://github.com/pytorch/ignite/blob/master/tests/ignite/engine/test_custom_events.py)

vfdev-5:
@JeevanChevula can you please code styling issue: https://github.com/pytorch/ignite/actions/runs/17579433225/job/49932341505?pr=3449

JeevanChevula:
Fixed formatting issues

JeevanChevula:
I investigated the CI failures and found the error message 'Please remove this workaround and restore dataset check.' appearing in all failed runs. I'm unsure of the exact cause, but I believe this may be an error independent of the PR functionality, as it appears to be related to Caltech101 dataset handling rather than the Engine method implementation

PR #3454, opened 2025-10-05
Unit test uv Support

### LINKED ISSUE #3350: Try to replace conda/setup-python with astral-uv if faster

ved1beta:
Fixes #3350 

Description:
pr for UV support on unit tests 
little to no peformance gain just optimizations here and there 

Check list: NA

- [] New tests are added (if a new feature is added)
- [] New doc strings: description and/or example code are in RST format
- [] Documentation is updated (if required)

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -69,28 +69,24 @@ jobs:
         with:
           version: "latest"
           python-version: ${{ matrix.python-version }}
-          activate-environment: true
           enable-cache: true
           cache-suffix: "${{ steps.get-date.outputs.date }}-pytorch-${{ runner.os }}-${{ matrix.python-version }}-${{ matrix.pytorch-channel }}"
           cache-dependency-glob: |
             **/requirements-dev.txt
             **/pyproject.toml
 
-      - name: Install dev tools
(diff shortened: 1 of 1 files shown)


vfdev-5:
@ved1beta thanks for the PR, but I'm not sure to understand why you change the GHA workflow already using UV. Can you give a context?

vfdev-5:
In absense of the response to motivate this change, I prefer to close this PR as it does not do in the correct direction, IMO.

PR #3472, opened 2025-12-04
Removed sync arg from wandb.log

### LINKED ISSUE #3471: WandBLogger calls Run.log with sync argument
## 🐛 Bug description

The W&B Logger interface seems to have been updated -- it does not take the `sync` argument anymore in the `log()` method. (See l. 301 in `wandb_logger.py`) It seems to have been removed in v0.20 of the wandb package, the current version is 0.23. It doesn't have any effect.

## Environment
```
pytorch-ignite                     0.5.3
torch                              2.9.1
wandb                              0.23.0
```
On Linux, installed through `uv` (pip), python version 3.12.11.

Removing the `sync` argument from the log call resolves the issue. I'll make a PR.

DanielNobbe:
Fixes #3471 

Description:
Remove the `sync` argument from the `wandb.log()` calls in the OutputHandler and OptimizerParamsHandler, since the argument was removed. Kept the argument in the initialisation of the handlers, so that existing code that may include the argument keeps working. 

Check list:

### DIFF
--- a/tests/ignite/handlers/test_wandb_logger.py
+++ b/tests/ignite/handlers/test_wandb_logger.py
@@ -31,14 +31,14 @@ def test_optimizer_params():
     mock_engine.state.iteration = 123
 
     wrapper(mock_engine, mock_logger, Events.ITERATION_STARTED)
-    mock_logger.log.assert_called_once_with({"lr/group_0": 0.01}, step=123, sync=None)
+    mock_logger.log.assert_called_once_with({"lr/group_0": 0.01}, step=123)
 
     wrapper = OptimizerParamsHandler(optimizer, param_name="lr", tag="generator")
     mock_logger = MagicMock(spec=WandBLogger)
     mock_logger.log = MagicMock()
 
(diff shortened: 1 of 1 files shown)


vfdev-5 (approved):
LGTM, thanks @DanielNobbe !

vfdev-5:
@DanielNobbe can you also please remove `sync` arg usage from tests/ignite/handlers/test_wandb_logger.py. Thanks!
and also fix the code style checker error: https://github.com/pytorch/ignite/actions/runs/19937571924/job/57228045157?pr=3472

DanielNobbe:
Done. I didn't think of checking the tests, sorry about that.
 I removed the one test that checked for the functionality of the `sync` argument, since the argument does nothing anymore. We don't have a test verifying that the sync argument does not trigger any errors, but I don't think that's necessary. If you would like me to add that test, let me know.

vfdev-5:
Thanks @DanielNobbe ! That's fine as you did. Let's see if the CI is green and we can land it.

Issue #3473, opened 2025-12-05
Pyrefly Type Checking?


rchen152:
Hi! I'm one of the maintainers of [Pyrefly](https://github.com/facebook/pyrefly), which is a Python type checker and language server. We've recently migrated a couple projects in the pytorch ecosystem (https://github.com/pytorch/helion/pull/1143, https://github.com/pytorch/pytorch/pull/166197) to Pyrefly for faster type checking and better IDE support, and I was wondering if Ignite would be interested in migrating as well?

I've drafted a commit that would enable Pyrefly, so you can see what that would look like: https://github.com/rchen152/ignite/commit/5233b91801bccfd9f257ffe8368f65f03f8a224e. Pyrefly does a full check in 1s, compared to 35s for Mypy. It does require more error suppressions for additional things it checks for, like possibly unbound variable names.

If this seems interesting, I can open a couple PRs (one to add Pyrefly, then a follow-up to remove Mypy and clean up some of the easy-to-fix suppressions).

vfdev-5:
Hi @rchen152 , sounds interesting! When we first added mypy, it could reveal some flows in the code. I wonder whether ignored errors are real or limitations of purefly?
Yes, please go ahead with PRs to add purefly and then remove mypy and other as you proposed. Thanks!

rchen152:
Awesome, I've opened https://github.com/pytorch/ignite/pull/3474 to add Pyrefly.

> I wonder whether ignored errors are real or limitations of purefly?

A lot of them are real errors, but there are also some pyrefly issues. For example, [this line](https://github.com/rchen152/ignite/blob/5233b91801bccfd9f257ffe8368f65f03f8a224e/ignite/metrics/ssim.py#L174) and [this line](https://github.com/rchen152/ignite/blob/5233b91801bccfd9f257ffe8368f65f03f8a224e/ignite/contrib/engines/tbptt.py#L121) look to me like legitimate errors. For the first one, the program will crash if `ndims` is any value except `2` or `3`, since `kernel_z` will be undefined, and the second is due to an incorrect type annotation in `Engine`. But [this one](https://github.com/rchen152/ignite/blob/5233b91801bccfd9f257ffe8368f65f03f8a224e/ignite/metrics/vision/object_detection_average_precision_recall.py#L164) is due to a missing [exhaustiveness check](https://github.com/facebook/pyrefly/issues/1708) in Pyrefly.

PR #3474, opened 2025-12-06
Enable Pyrefly type checking

rchen152:
Ref #3473

Description:

Enables [Pyrefly](https://github.com/facebook/pyrefly), a fast type checker. I ported the `mypy.ini` file to a `pyrefly.toml` using the `pyrefly init` command and added `# pyrefly: ignore` comments to suppress new type errors.

Since this PR already touches a lot of files due to the suppressions, I opted to keep it focused on adding Pyrefly. If it's accepted, I'll follow up with another PR to remove Mypy, update documentation and other references, and clean up easy-to-fix suppressions.

### DIFF
--- a/ignite/engine/events.py
+++ b/ignite/engine/events.py
@@ -91,8 +91,7 @@ class CallableEventWithFilter:
             raise ValueError("Argument every should be integer and greater than zero")
 
         if once is not None:
-            # pyrefly: ignore [unsupported-operation]
-            c1 = isinstance(once, numbers.Integral) and once > 0
+            c1 = isinstance(once, int) and once > 0
             c2 = isinstance(once, Sequence) and len(once) > 0 and all(isinstance(e, int) and e > 0 for e in once)
             if not (c1 or c2):
                 raise ValueError(
--- a/ignite/metrics/ssim.py
+++ b/ignite/metrics/ssim.py
@@ -161,16 +161,19 @@ class SSIM(Metric):
             kernel_y = self._gaussian(kernel_size[1], sigma[1])
             if ndims == 3:
                 kernel_z = self._gaussian(kernel_size[2], sigma[2])
+            else:
+                kernel_z = None
         else:
             kernel_x = self._uniform(kernel_size[0])
             kernel_y = self._uniform(kernel_size[1])
             if ndims == 3:
                 kernel_z = self._uniform(kernel_size[2])
+            else:
+                kernel_z = None
 
         result = (
             torch.einsum("i,j->ij", kernel_x, kernel_y)
             if ndims == 2
-            # pyrefly: ignore [unbound-name]
             else torch.einsum("i,j,k->ijk", kernel_x, kernel_y, kernel_z)
         )
         return result
--- a/ignite/metrics/vision/object_detection_average_precision_recall.py
+++ b/ignite/metrics/vision/object_detection_average_precision_recall.py
@@ -160,7 +160,9 @@ class ObjectDetectionAvgPrecisionRecall(Metric, _BaseAveragePrecision):
         elif self._area_range == "large":
             min_area = 9216
             max_area = 1e10
-        # pyrefly: ignore [unbound-name]
+        else:
+            min_area = 0
+            max_area = 1e10
         return torch.logical_and(areas >= min_area, areas <= max_area)
 
     def _check_matching_input(


.github/workflows/typing-checks.yml
```diff
@@ -64,6 +64,10 @@ jobs:
           uv pip install -r requirements-dev.txt
           uv pip install .
           uv pip install mypy
+          uv pip install pyrefly==0.44.2
```
vfdev-5: Can we install the latest (and avoid upgrading the version)?

ignite/contrib/engines/common.py
```diff
@@ -265,6 +265,7 @@ def _setup_common_distrib_training_handlers(
 
         @trainer.on(Events.EPOCH_STARTED)
         def distrib_set_epoch(engine: Engine) -> None:
+            # pyrefly: ignore [missing-attribute]
```
vfdev-5, line 268: @rchen152 do you know why there is an error here?

train_sampler is checked above to be a DistributedSampler, which has set_epoch method: https://github.com/pytorch/pytorch/blob/d38164a545b4a4e4e0cf73ce67173f70574890b6/torch/utils/data/distributed.py#L139

ignite/contrib/handlers/base_logger.py
```diff
@@ -9,6 +9,7 @@
 removed_in = "0.6.0"
 deprecation_warning = (
     f"{__file__} has been moved to /ignite/handlers/base_logger.py"
+    # pyrefly: ignore [redundant-condition]
```
vfdev-5: I agree with pyrefly here, can you please rewrite below line as 
```diff
- + (f" and will be removed in version {removed_in}" if removed_in else "")
+ + (f" and will be removed in version {removed_in}")
```
and remove the ignore line

ignite/distributed/auto.py
```diff
@@ -336,6 +336,7 @@ class _MpDeviceLoader:
         # From pytorch/xla if `torch_xla.distributed.parallel_loader.MpDeviceLoader` is not available
         def __init__(self, loader: Any, device: torch.device, **kwargs: Any) -> None:
             self._loader = loader
+            # pyrefly: ignore [read-only]
```
vfdev-5, line 339: @rchen152 can you please explain this error? I'm not totally sure what is read-only here.

ignite/distributed/comp_models/horovod.py
```diff
@@ -70,6 +70,7 @@ def __init__(self, backend: Optional[str] = None, **kwargs: Any) -> None:
                 self._init_from_context()
 
         def _create_from_backend(self, backend: str, **kwargs: Any) -> None:
+            # pyrefly: ignore [bad-override]
```
vfdev-5, line 73: Same here, rather unclear what is bad-override here and below.

ignite/engine/deterministic.py
```diff
@@ -89,6 +89,7 @@ def _get_rng_states() -> List[Any]:
     try:
         import numpy as np
 
+        # pyrefly: ignore [bad-argument-type]
```
vfdev-5: Do you know what's wrong here?

ignite/engine/engine.py
```diff
@@ -1038,6 +1039,7 @@ def _internal_run_as_gen(self) -> Generator[Any, None, State]:
                     "https://github.com/pytorch/ignite/issues/new/choose"
                 )
 
+            # pyrefly: ignore [unbound-name]
```
vfdev-5, line 1042: Here, I agree with pyrefly, in a follow-up PR we can put `start_time` before the try/except clauses.

ignite/engine/events.py
```diff
@@ -91,6 +91,7 @@ def __call__(
             raise ValueError("Argument every should be integer and greater than zero")
 
         if once is not None:
+            # pyrefly: ignore [unsupported-operation]
```
vfdev-5: A bit unclear what's wrong here

.github/workflows/typing-checks.yml
rchen152: Done. Fair warning that we release new pyrefly versions 1-2 times a week at the moment, so not pinning the pyrefly version may lead to ignite picking up new features (that may generate new type errors) rather often.

ignite/contrib/engines/common.py
rchen152: This error is due to pyrefly treating the type of `train_sampler` as `DistributedSampler | None` inside the nested function, to guard against problems like this:
```
def f(train_sampler: DistributedSampler | None):
    if train_sampler is not None:
        def g():
            train_sampler.set_epoch(...)
        train_sampler = None
        g()  # oops! This would crash
```

With that said, we may be able to make this check less conservative. We have https://github.com/facebook/pyrefly/issues/40 open to do some additional analysis to figure out when we can not emit this error.

ignite/distributed/auto.py
rchen152: This one took me a while to figure out! It's a bug in pyrefly (https://github.com/facebook/pyrefly/issues/1803) caused by treating `torch.device` as a descriptor.

ignite/distributed/comp_models/horovod.py
rchen152: The `bad-override` error is caused by setting the type of `_backend` (and the type signature of `spawn` below) to a different type than in the `ComputationModel` base class. Pyrefly errors on this to prevent problems like this:
```
class ComputationModel:
    def __init__(self):
        self._backend: str | None = ...
class _HorovodDistModel(ComputationModel):
    def _create_from_backend(self):
        self._backend: str = ...

def clear_backend(model: ComputationModel):
    # This is accepted by a type checker because `_backend` in `ComputationModel` is annotated as `str | None`.
    model._backend = None

# This is accepted by a type checker because `_HorovodDistModel` is a subclass of `ComputationModel`.
# But it would lead to `_backend` being set to `None`, and any code in `_HorovodDistModel` that assumes
# `_backend` to be `str` would crash.
clear_backend(_HorovodDistModel())
```

ignite/engine/deterministic.py
rchen152: This is down to a difference between mypy and pyrefly's type inference strategies. Mypy says that `output` has type `List[object]` because `random.getstate()` and `torch.get_rng_state()` return types with no common superclass except `object`, whereas Pyrefly sets the list's contained type to the union of the two types. I removed this suppression and added an explicit annotation, so that no inference is needed.

ignite/engine/events.py
rchen152: The text of the error message is:
```
`>` is not supported between `Integral` and `Literal[0]`
```
Interestingly, mypy emits the [same error](https://mypy-play.net/?mypy=latest&python=3.12&gist=664a0d12fbfb007db8c7b240e4e12408) if you make it check an operation between Integral and int directly. The reason it doesn't error here seems to be that it (incorrectly) treats the code as unreachable: [playground](https://mypy-play.net/?mypy=latest&python=3.12&gist=4e6941b586c5939504e4b4cd4f3f29b3).

rchen152:
@vfdev-5 Thanks for the review! I believe I've addressed all the comments.

ignite/metrics/vision/object_detection_average_precision_recall.py
```diff
@@ -160,6 +160,7 @@ def _match_area_range(self, bboxes: torch.Tensor) -> torch.Tensor:
         elif self._area_range == "large":
             min_area = 9216
             max_area = 1e10
+        # pyrefly: ignore [unbound-name]
```
vfdev-5: Can we do this here?
```suggestion
        else:
            min_area = 0
            max_area = 1e10
```

ignite/metrics/ssim.py
```diff
@@ -170,6 +170,7 @@ def _gaussian_or_uniform_kernel(
         result = (
             torch.einsum("i,j->ij", kernel_x, kernel_y)
             if ndims == 2
+            # pyrefly: ignore [unbound-name]
```
vfdev-5: Can we get rid of the ignore if we rewrite it as?
```python
result = (
    torch.einsum("i,j,k->ijk", kernel_x, kernel_y, kernel_z)
    if ndims == 3
    else torch.einsum("i,j->ij", kernel_x, kernel_y)
)
```

ignite/engine/events.py
vfdev-5: I see, thanks! It's true that previously we had also hard time with typing and numbers. Maybe we can safely replace the check as:
```python
c1 = isinstance(once, int) and once > 0
```

ignite/metrics/ssim.py
rchen152: That didn't work, unfortunately (I think the control flow was too complex for pyrefly to follow); I added `kernel_z = None` branches to make sure `kernel_z` is always defined instead.

PR #3480, opened 2026-01-08
[skip ci] fix type hints of OuptutHandlers based on its base class.

### LINKED ISSUE #3479: OutputHandler typing issue: metric_names should accept str as well as Optional[List[str]] in all logger handlers
## 🐛 Bug description

The typing for the `metric_names` parameter in OutputHandler-derived classes is too restrictive in several files. The signature currently expects `Optional[List[str]]`, but according to usage in documentation and code, it should also allow a `str` value (e.g., `"all"`).

- The **base class** in `ignite/handlers/base_logger. py` BaseOutputHandler is correct:
  ```python
  metric_names: Optional[Union[str, List[str]]] = None
  ```
- **But all the logger OutputHandler implementations** in these files override it with the more restrictive type:
  - `ignite/handlers/tensorboard_logger.py`
  - `ignite/handlers/wandb_logger.py`
  - `ignite/handlers/mlflow_logger.py`
  - `ignite/handlers/polyaxon_logger. py`
  - `ignite/handlers/visdom_logger.py`
  - `ignite/handlers/neptune_logger.py`
  - `ignite/handlers/clearml_logger.py`

  All of these should also use `Optional[Union[str, List[str]]]` for `metric_names` for full typing correctness and consistency with the base class and library usage.

**Current signature:**
```python
metric_names: Optional[List[str]] = None
```

**Expected signature (type hint):**
```python
metric_names:  Union[str, Optional[List[str]]] = None
# or, preferably, to match the base:
metric_names:  Optional[Union[str, List[str]]] = None
```

This would align the typing with examples and usage where a string like "all" is passed.

**Location**: All logger OutputHandler classes (see list above), parameter `metric_names`

**Steps to reproduce**:
- Try type checking with `metric_names="all"` for OutputHandler in any logger module; type checker (e.g. mypy) issues an error.
- Documentation also shows examples where a string is allowed.

**Expected behavior:**
- Typing should reflect that both string (e.g., "all") and List[str] are allowed for `metric_names` in all OutputHandler classes.

**Pull request**
If this change is approved, i would like to create a PR.

## Environment

- PyTorch Version: 
- Ignite Version: 
- OS:  
- Python version:

gabrielfruet:
Fixes #3479

Fix type hints for derived classes of BaseOutputHandler. They can receive `None | List[str] | str` , but their type hints only for `None | List[str]`.

### DIFF
--- a/ignite/handlers/lr_finder.py
+++ b/ignite/handlers/lr_finder.py
@@ -317,7 +317,7 @@ class FastaiLRFinder:
             lrs = lrs[skip_start:-skip_end]
             losses = losses[skip_start:-skip_end]
 
-        plt.legend(legends)
+        plt.legend(legends)  # pyrefly: ignore
         # Plot loss as a function of the learning rate
         ax.plot(lrs, losses)
         if log_lr:


vfdev-5:
Thanks for the PR @gabrielfruet !
If you are interested in further typing improvements we can also make the change: `Optional[Union[int, float]] = None` into `int | float = None` which should be OK starting from python 3.10

vfdev-5:
@rchen152 we have the following pyrefly issue on this PR:
```
ERROR Argument `list[str]` is not assignable to parameter `handles` with type `list[Artist | tuple[Artist, ...]] | None` in function `matplotlib.pyplot.legend` [bad-argument-type]
   --> ignite/handlers/lr_finder.py:320:20
    |
320 |         plt.legend(legends)
    |                    ^^^^^^^
    |
Error: Argument `list[str]` is not assignable to parameter `handles` with type `list[Artist | tuple[Artist, ...]] | None` in function `matplotlib.pyplot.legend`
 INFO 1 error (103 suppressed)
```
is it pyrefly related or fixed or should we add an ignore comment? Thanks!


I checked matplotlib API: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.legend.html
and using only labels (list of str) is possible but discouraged (`4. Labeling existing plot elements`)

gabrielfruet:
@vfdev-5 Thank you!

I could also make that improvement. I'll start with the BaseOutputHandler derived classes.

Are we supporting <3.10?

vfdev-5:
> I could also make that improvement.

Thanks @gabrielfruet! Let's do that in a follow-up PRs.

> Are we supporting <3.10?

python 3.9 is EOL, min supported python version is 3.10

gabrielfruet:
@vfdev-5 Also, if you could give some help. Didn't understand why these others GH actions like CPU tests ran with error. I didn't see any. Only on pyrefly.

vfdev-5:
We can ignore them for now. pytorch-nightly on 3.11+ started failing on some distrbuted metrics tests...

rchen152:
> @rchen152 we have the following pyrefly issue on this PR:
> 
> ```
> ERROR Argument `list[str]` is not assignable to parameter `handles` with type `list[Artist | tuple[Artist, ...]] | None` in function `matplotlib.pyplot.legend` [bad-argument-type]
>    --> ignite/handlers/lr_finder.py:320:20
>     |
> 320 |         plt.legend(legends)
>     |                    ^^^^^^^
>     |
> Error: Argument `list[str]` is not assignable to parameter `handles` with type `list[Artist | tuple[Artist, ...]] | None` in function `matplotlib.pyplot.legend`
>  INFO 1 error (103 suppressed)
> ```
> 
> is it pyrefly related or fixed or should we add an ignore comment? Thanks!
> 
> I checked matplotlib API: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.legend.html and using only labels (list of str) is possible but discouraged (`4. Labeling existing plot elements`)

The error is due to this type signature in `matplotlib-stubs`, which pyrefly bundles automatically to provide better type information for `matplotlib`: https://github.com/hoel-bagard/matplotlib-stubs/blob/59fdaf216331aa54996e5f1149e6ee096ca63557/src/matplotlib-stubs/pyplot.pyi#L690.

With that said, I checked the description of `matplotlib-stubs` (https://pypi.org/project/matplotlib-stubs/), and it looks like `matplotlib` itself ships with official stubs, so maybe we should stop bundling `matplotlib-stubs` if the stubs are causing false positives.

I'll raise this with the other pyrefly devs. In the meantime, I think you can safely `# pyrefly: ignore` this error.

Issue #3481, opened 2026-01-14
Replace typing hints with new version


vfdev-5:
Typing improvements we can make the change: `Optional[Union[int, float]] = None` into `int | float = None` which should be OK starting from python 3.10

We should replace all: 
- `Union[A, B]` with `A | B`
- `Optional[A]` with `A | None`
- `Tuple` -> `tuple`, `List` -> `list`, etc
in the function signature and also of the output type hint

python 3.9 is EOL, min supported python version is 3.10.

It is a large issue, multiple people can work on it.
If you plan to work on it. Make changes in a single file and submit a PR. Smaller PR == easy review == quicker merging.

maifeeulasad:
Can I get assigned to this. Willing to work on this one.

vfdev-5:
It is a large issue, multiple people can work on it. I wont assign anyone on it.

maifeeulasad:
@vfdev-5 excellent. Raised first MR for a single file as instructed. And I only made changes for `Optional[Union[...]] = None` into the new pattern.

We can also do `Union[int, float]` into `int | float`, do we track a seperate issue for that, or just use this one?

ref: https://github.com/pytorch/ignite/pull/3482

vfdev-5:
I updated the description, we should just replace all old type hints with the new one: Optional and Union in this issue.

omkar-334:
@vfdev-5 we can add a pre-commit hook to check PEP 585 compliance like we did in https://github.com/meta-pytorch/torchtune/pull/2720 and https://github.com/meta-pytorch/torchtune/issues/2734 . Do you think this should be done?

omkar-334:
Also, `pyproject.toml` still includes python 3.9 . I think the new typing hints will result in `TypeErrors` if someone uses python 3.9

vfdev-5:
> Also, `pyproject.toml` still includes python 3.9 . I think the new typing hints will result in `TypeErrors` if someone uses python 3.9

Let's update it, thanks!


>  we can add a pre-commit hook to check PEP 585 compliance like we did in https://github.com/meta-pytorch/torchtune/pull/2720 and https://github.com/meta-pytorch/torchtune/issues/2734 . Do you think this should be done?

Do you mean using ruff in the pre-commit hook like this PR: https://github.com/meta-pytorch/torchtune/pull/2737 ?

omkar-334:
> Let's update it, thanks!

Sure, i can open a PR for this...

[...]

PR #3482, opened 2026-01-14
refactor `Optional[Union[...]]` w `|` for `utils.py`;

maifeeulasad:
Related to #3481

Description:

- Removed un-used from `typing` imports
- `Optional[Union[str, torch.device]]` → `str | torch.device | None`
- `Optional[Union[int, str]]` → `int | str | None`  
- `Optional[str]` → `str | None`
- `Optional[TextIO]` → `TextIO | None`
- `Optional[int]` → `int | None`
- `Tuple` -> `tuple`, `List` -> `list`, `Dict` -> `dict`, `Type` -> `type`

Tested w `python -m pytest tests/ignite/test_utils.py -x --tb=line --disable-warnings -q`

NB: The `Union[Any, "_CollectionItem"]` was kept as-is since `Any` doesn't work with the `|` operator syntax.

Check list:

### DIFF
--- a/ignite/utils.py
+++ b/ignite/utils.py
@@ -9,7 +9,7 @@ import random
 import shutil
 import warnings
 from pathlib import Path
-from typing import Any, Callable, cast, Dict, List, TextIO, Tuple, Type, TypeVar
+from typing import Any, Callable, cast, TextIO, TypeVar
 
 import torch
 
@@ -59,7 +59,7 @@ def apply_to_tensor(
 
 def apply_to_type(
     x: Any | collections.Sequence | collections.Mapping | str | bytes,
-    input_type: Type | Tuple[Type[Any], Any],
+    input_type: type | tuple[type[Any], Any],
     func: Callable,
 ) -> Any | collections.Sequence | collections.Mapping | str | bytes:
     """Apply a function on an object of `input_type` or mapping, or sequence of objects of `input_type`.
@@ -94,7 +94,7 @@ def _tree_map(
     return func(x, key=key)
 
 
-def _to_str_list(data: Any) -> List[str]:
+def _to_str_list(data: Any) -> list[str]:
     """
     Recursively flattens and formats complex data structures, including keys for
     dictionaries, into a list of human-readable strings.
@@ -129,7 +129,7 @@ def _to_str_list(data: Any) -> List[str]:
         A list of formatted strings, each representing a part of the input data
         structure.
     """
-    formatted_items: List[str] = []
+    formatted_items: list[str] = []
 
     def format_item(item: Any, prefix: str = "") -> str | None:
         if isinstance(item, numbers.Number):
@@ -171,9 +171,9 @@ def _to_str_list(data: Any) -> List[str]:
 
 
 class _CollectionItem:
-    types_as_collection_item: Tuple = (int, float, torch.Tensor)
+    types_as_collection_item: tuple = (int, float, torch.Tensor)
 
-    def __init__(self, collection: Dict | List, key: int | str) -> None:
+    def __init__(self, collection: dict | list, key: int | str) -> None:
         if not isinstance(collection, (dict, list)):
             raise TypeError(
                 f"Input type is expected to be a mapping or list, but got {type(collection)} " f"for input key '{key}'."
@@ -191,7 +191,7 @@ class _CollectionItem:
         return self.collection[self.key]  # type: ignore[index]
 
     @staticmethod
-    def wrap(object: Dict | List, key: int | str, value: Any) -> Any | "_CollectionItem":
+    def wrap(object: dict | list, key: int | str, value: Any) -> Any | "_CollectionItem":
         return (
             _CollectionItem(object, key)
             if value is None or isinstance(value, _CollectionItem.types_as_collection_item)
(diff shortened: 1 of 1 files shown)


ignite/utils.py
```diff
@@ -25,7 +25,7 @@
 
 def convert_tensor(
     x: Union[torch.Tensor, collections.Sequence, collections.Mapping, str, bytes],
-    device: Optional[Union[str, torch.device]] = None,
+    device: str | torch.device | None = None,
```
vfdev-5, line 30: You can also change above type hint:
```diff
- x: Union[torch.Tensor, collections.Sequence, collections.Mapping, str, bytes],
+ x: torch.Tensor | collections.Sequence | collections.Mapping | str | bytes,
```

and the same for the output signature:
```diff
- Union[torch.Tensor, collections.Sequence, collections.Mapping, str, bytes]:
+ torch.Tensor | collections.Sequence | collections.Mapping | str | bytes:
```

ignite/utils.py
```diff
@@ -81,7 +81,7 @@ def apply_to_type(
 
 
 def _tree_map(
-    func: Callable, x: Union[Any, collections.Sequence, collections.Mapping], key: Optional[Union[int, str]] = None
+    func: Callable, x: Union[Any, collections.Sequence, collections.Mapping], key: int | str | None = None
```
vfdev-5: Same here and everywhere

maifeeulasad:
After 0f1d178:

 - Added `from __future__ import annotations`, explained later
 - Removed `Union` from imports
 - Replaced all `Union[A, B]` with `A | B`
 - Replaced all `Optional[A]` with `A | None`
 
 
 ### `from __future__ import annotations` explaination

We needed `from __future__ import annotations` because of **PEP 563 - Postponed Evaluation of Annotations**.

Without it, Python evaluates type annotations at **runtime** when the module is imported. This caused the error:

```python
def wrap(object: Dict | List, key: int | str, value: Any) -> Any | "_CollectionItem":
                                                             ^^^^^^^^^^^^^^^^^^^^^^^
TypeError: unsupported operand type(s) for |: '_AnyMeta' and 'str'
```

The problem: Python tried to execute `Any | "_CollectionItem"` at import time, but `Any` (from typing) doesn't support the `|` operator with strings.

**With `from __future__ import annotations`:**
- All annotations are stored as **strings** and not evaluated at import time
- They're only evaluated when explicitly needed (by type checkers, IDEs, or `get_type_hints()`)
- This allows:
  - Forward references (like `"_CollectionItem"` before the class is defined)
  - Complex type expressions that might not be valid at runtime
  - Faster import times (annotations aren't evaluated)

This becomes the default behavior in Python 3.11+, but for Python 3.10 we need the `__future__` import to enable it.

ignite/utils.py
```diff
@@ -422,7 +424,7 @@ def wrapper(*args: Any, **kwargs: Dict[str, Any]) -> Callable:
     return decorator
 
 
-def hash_checkpoint(checkpoint_path: Union[str, Path], output_dir: Union[str, Path]) -> Tuple[Path, str]:
+def hash_checkpoint(checkpoint_path: str | Path, output_dir: str | Path) -> Tuple[Path, str]:
```
vfdev-5:
```suggestion
def hash_checkpoint(checkpoint_path: str | Path, output_dir: str | Path) -> tuple[Path, str]:
```

ignite/utils.py
```diff
@@ -56,10 +58,10 @@ def apply_to_tensor(
 
 
 def apply_to_type(
-    x: Union[Any, collections.Sequence, collections.Mapping, str, bytes],
-    input_type: Union[Type, Tuple[Type[Any], Any]],
+    x: Any | collections.Sequence | collections.Mapping | str | bytes,
+    input_type: Type | Tuple[Type[Any], Any],
```
vfdev-5: Tuple -> tuple, List -> list, etc everywhere, please

ignite/utils.py
```diff
@@ -171,7 +173,7 @@ def format_item(item: Any, prefix: str = "") -> Optional[str]:
 class _CollectionItem:
     types_as_collection_item: Tuple = (int, float, torch.Tensor)
 
-    def __init__(self, collection: Union[Dict, List], key: Union[int, str]) -> None:
+    def __init__(self, collection: Dict | List, key: int | str) -> None:
```
vfdev-5: Above:
```diff
- types_as_collection_item: Tuple = (int, float, torch.Tensor)
+ types_as_collection_item: tuple = (int, float, torch.Tensor)
```

ignite/utils.py
```diff
@@ -189,7 +191,7 @@ def value(self) -> Any:
         return self.collection[self.key]  # type: ignore[index]
 
     @staticmethod
-    def wrap(object: Union[Dict, List], key: Union[int, str], value: Any) -> Union[Any, "_CollectionItem"]:
+    def wrap(object: Dict | List, key: int | str, value: Any) -> Any | "_CollectionItem":
```
vfdev-5:
```suggestion
    def wrap(object: dict | list, key: int | str, value: Any) -> Any | "_CollectionItem":
```

vfdev-5 (commented):
Thanks for working on the PR!

maifeeulasad:
@vfdev-5 thanks for giving me the opportunity to work on this one! And reviewing my work so many times.

PR #3483, opened 2026-01-14
refactor `Union[A, B]` w `A | B` and `Optional[A]` w `A | None` for `common.py`

maifeeulasad:
Related to #3481 

Description:
- Added `from __future__ import annotations` at the top
- Removed un-wanted types from `typing` imports
- Replaced all `Union[A, B]` with `A | B`
- Replaced all `Optional[A]` with `A | None`
- Replaced all `Dict` with `dict`
- validated w ` python -m pytest tests/ignite/contrib/engines/test_common.py -x --tb=line --disable-warnings -q` before and after

Check list:

### DIFF
--- a/.github/workflows/code-style-checks.yml
+++ b/.github/workflows/code-style-checks.yml
@@ -29,6 +29,7 @@ on:
       - "!docs/**"
       - "!conda.recipe"
   workflow_dispatch:
+  merge_group:
 
 concurrency:
   # <workflow_name>-<branch_name>-<true || commit_sha (if branch is protected)>
--- a/.github/workflows/hvd-tests.yml
+++ b/.github/workflows/hvd-tests.yml
@@ -16,6 +16,7 @@ on:
       - "tests/run_cpu_tests.sh"
       - ".github/workflows/hvd-tests.yml"
   workflow_dispatch:
+  merge_group:
 
 concurrency:
   # <workflow_name>-<branch_name>-<true || commit_sha (if branch is protected)>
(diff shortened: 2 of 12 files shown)


ignite/contrib/engines/common.py
```diff
@@ -286,8 +288,8 @@ def setup_any_logging(
     logger: BaseLogger,
     logger_module: Any,
     trainer: Engine,
-    optimizers: Optional[Union[Optimizer, Dict[str, Optimizer], Dict[None, Optimizer]]],
-    evaluators: Optional[Union[Engine, Dict[str, Engine]]],
+    optimizers: Optimizer | Dict[str, Optimizer] | Dict[None, Optimizer] | None,
+    evaluators: Engine | Dict[str, Engine] | None,
```
vfdev-5:
```suggestion
    optimizers: Optimizer | dict[str, Optimizer] | dict[None, Optimizer] | None,
    evaluators: Engine | dict[str, Engine] | None,
```

Dict -> dict everywhere please

ignite/contrib/engines/common.py
```diff
@@ -1,7 +1,9 @@
+from __future__ import annotations
```
vfdev-5: Is it needed in this file, we do not return any type with `|` here ?

maifeeulasad: you are god damn right, just pushed the resolution

maifeeulasad:
Thanks for all the support @vfdev-5 . Really appreciate it. Looking forward to work with you in the future as well.

PR #3484, opened 2026-01-15
Replace Union with | syntax in root_mean_squared_error.py

JonSnow1807:
## Summary
  - Replace `Union[torch.Tensor, float]` with `torch.Tensor | float`
  - Remove unused `from typing import Union` import

  Part of #3481

  ## Test plan
  - [x] Existing tests pass (25 passed)
  - [x] Import check verified


vfdev-5:
Thanks for the PR @JonSnow1807 
Let's see whether CI would be happy with this change

PR #3485, opened 2026-01-15
Replace Union with | syntax in mean_squared_error.py

JonSnow1807:
## Summary
  - Replace `Union[float, torch.Tensor]` with `float | torch.Tensor`
  - Remove `Union` from typing import (keep `Sequence`)

  Part of #3481

  ## Test plan
  - [x] Existing tests pass (42 passed)
  - [x] Import check verified


vfdev-5 (approved):
Thanks @JonSnow1807 !

PR #3489, opened 2026-01-15
Replace Union with | syntax in cosine_similarity.py

JonSnow1807:
## Summary
- Replace `Union[str, torch.device]` with `str | torch.device`
- Remove `Union` from typing import (keep `Callable`, `Sequence`)

Part of #3481

## Test plan
- [x] Import check verified


vfdev-5 (approved):
Thanks !

PR #3493, opened 2026-01-16
Use `Parallel.spawn_kwargs` as kwargs to the backend when appropriate

qwfy:
Fixes #{issue number}
None.

Description:

Allow passing extra arguments to the distributed backend, this enables the following usage pattern:

```
torchrun ... main.py ...

# Note that nproc_per_node is not specified.
with idist.Prallel(backend=foo, timeout=bar) as par:
    ....
```
Which is not previously possible.

Note that `kwargs` in `_NativeDistModel._create_from_backend` is unused, so there is still a chance that the args passed via spawn_args is still unused, but it works for explicitly listed arguments like timeout

Check list:

### DIFF
--- a/ignite/distributed/launcher.py
+++ b/ignite/distributed/launcher.py
@@ -242,8 +242,6 @@ class Parallel:
                 )
             else:
                 self._pg_init_kwargs = spawn_kwargs
-                if "init_method" in self._pg_init_kwargs:
-                    raise ValueError("Argument 'init_method' should not be specified in 'spawn_kwargs'")
 
 
         # The logger will be setup after the idist.initialize() call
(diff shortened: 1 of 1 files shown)


vfdev-5:
Thanks for the PR @qwfy 
It makes sense, but reusing `spawn_kwargs` as `kwargs` for initialization method can be misleading.

Let's rename `spawn_kwargs` to `kwargs` and update the docstring:
```diff
- spawn_kwargs: kwargs to ``idist.spawn`` function.
+ kwargs: kwargs to the function responsible to initialize a processing group, e.g. ``idist.spawn`` or ``idist.initialize``.
```

PR #3512, opened 2026-02-04
Update type hints to Python 3.10+ in ignite/metrics/roc_auc.py

davideticchiarelli01:
Fixes part of #3481.

Updated type hints in `ignite/metrics/roc_auc.py` to use Python 3.10+ syntax (e.g., "|" instead of "Union").


vfdev-5 (approved):
Thanks @davideticchiarelli01 !

PR #3514, opened 2026-02-05
Add ability to choose between flattening or using named for nested metrics.

### LINKED ISSUE #3513: Nested metrics are being duplicated when logging.
## 🚀 Feature

When we use nested metrics (i.e. ClassificationReport) we have a structure like this:

```python
{
   'macro_avg': {
         'precision': ...,
         'f1-score': ...,
         'recall': ...,
    }
}
```

https://github.com/pytorch/ignite/blob/7bb6cd70d3855588a4efec63ca24934ba7dee300/ignite/metrics/classification_report.py#L120-L141

Currently it goes through this code path when adding the metrics to the engine state metrics.

https://github.com/pytorch/ignite/blob/7bb6cd70d3855588a4efec63ca24934ba7dee300/ignite/metrics/metric.py#L497-L503

Which will fill the metrics dict in the next way:

```python
{
   '$ENGINE_METRIC_NAME': {
         macro_avg': {
           'precision': ...,
           'f1-score': ...,
           'recall': ...
          },
    },
    macro_avg': {
         'precision': ...,
         'f1-score': ...,
         'recall': ...
     },
}
```

`$ENGINE_METRIC_NAME` is the one defined here:

```python
    classification_report.attach(evaluator, "$ENGINE_METRIC_NAME")
```

The main problem in that is for when we want to log metrics using TensorBoard logger for example, we need to either specify what metrics we want or all.

*my code below* 

```python
def setup_tblogger(
...
    train_metric_names: list[str] | str = "all",
...
) -> TensorboardLogger:

    logger.attach(
        trainer,
        log_handler=OutputHandler(
            tag="train",
            metric_names=train_metric_names, # list or all
        ),
        event_name=Events.EPOCH_STARTED,
    )
```

And in the case of classification report, we end with something not much desirable, that is repeated metrics in the TensorBoard logger, since we will flatten and get something like this:

$ENGINE_METRIC_NAME/macro_avg/precision
macro_avg/precision

We could potentially list only the metrics we want, or a blacklist of metrics that we don't want (not supported), but in any way, would be just a workaround of the real problem.

gabrielfruet:
Fixes #3513

Description:

First of a few PRs to fix this. Enabling the user to choose if should use flattening, named or both.

Check list:

### DIFF
--- a/tests/ignite/metrics/test_metric.py
+++ b/tests/ignite/metrics/test_metric.py
@@ -1526,7 +1526,7 @@ def test_metrics_result_mode_parametrized(mode, expect_map, expect_keys):
 
 def test_metrics_result_mode_conflict_error():
     metric = MappingMetric({"map": 3}, metrics_result_mode="named")
-    engine = Engine(lambda x: 0)
+    engine = Engine(lambda e, x: 0)
     metric.attach(engine, "map")
     with pytest.raises(ValueError):
         engine.run([0], max_epochs=1)


tests/ignite/metrics/test_metric.py
```diff
@@ -1475,3 +1475,62 @@ def test_output_transform_type_check():
...
+    def process_function(x):
+        return 0
+
+    engine = Engine(process_function)
+    metric.attach(engine, name)
+    engine.run([0], max_epochs=1)
+    return engine.state.metrics
+
+
+def test_metrics_result_mode_flatten():
```
vfdev-5: I wonder if we can write these tests as a single parametrized test?
It seems a bit complicated, if there is no a bright idea of doing it in less code lines, but still readable, I'm ok keeping it as it is.

tests/ignite/metrics/test_metric.py
```diff
@@ -1475,3 +1475,58 @@ def test_output_transform_type_check():
...
+
+    def update(self, output):
+        pass
+
+    def compute(self):
+        return dict(self._mapping)
+
+
+def _run_and_get_metrics(metric: Metric, name: str):
+    engine = Engine(lambda x: 0)
```
vfdev-5:
```suggestion
    engine = Engine(lambda e, x: 0)
```

CI is failing here: https://github.com/pytorch/ignite/actions/runs/21711323355/job/62615450498

Issue #3411, opened 2025-06-04
Enhancements to EarlyStopping handler


bonassifabio:
## 🚀 Feature

I would propose several enhancements to the  [EarlyStopping handler](https://docs.pytorch.org/ignite/_modules/ignite/handlers/early_stopping.html#EarlyStopping).

1. Introduce a new parameter, `min_delta_mode`, mimicking the parameter `threshold_mode` of [ReduceLROnPlateau](https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.ReduceLROnPlateau.html). If `min_delta_mode="abs"` (default) everything works as of now, while if `min_delta_mode="rel"` the score comparison becomes `score >= best_score (1 + min_delta)`.
2. Introduce a new parameter `mode` that work's as in [ReduceLROnPlateau](https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.ReduceLROnPlateau.html), allowing to handle validation metrics to be minimized.

Other ideas I'd like to discuss are:

-  Natively implementing `min_epochs`/`min_evals`, preventing EarlyStopping from terminating until it has been evaluated at least X times. I know this could be addressed as in #3317 but this solution might be more compact? 
- It would be nice to have the same `threshold` and `threshold_mode` nomenclature as `ReduceLROnPlateau`, but I realize this would probably break older code.
- I personally find the parameter name `cumulative_delta` quite counter-intuitive. When I first read about that parameter, I understood it was taking the cumulative sum of the improvements and comparing this cumulative sum to `min_delta`.

vfdev-5:
Thanks for suggestions, @bonassifabio !

> I personally find the parameter name cumulative_delta quite counter-intuitive. When I first read about that parameter, I understood it was taking the cumulative sum of the improvements and comparing this cumulative sum to min_delta.

We can start deprecating it in favor of another argument. 

> Introduce a new parameter, min_delta_mode, mimicking the parameter threshold_mode of [ReduceLROnPlateau](https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.ReduceLROnPlateau.html). If min_delta_mode="abs" (default) everything works as of now, while if min_delta_mode="rel" the score comparison becomes score >= best_score (1 + min_delta).

Sounds good!


[...]

PR #3516, opened 2026-02-05
Add min_delta_mode parameter to EarlyStopping and update logic for sc…

aaishwarymishra:
This pull request enhances the `EarlyStopping` handler in `ignite/handlers/early_stopping.py` by introducing a new parameter to control how `min_delta` is interpreted, allowing users to specify whether the minimum improvement threshold should be considered as an absolute or relative increase. This makes early stopping more flexible and suitable for a wider range of training scenarios.

**Early stopping configuration improvements:**

* Added a new `min_delta_mode` parameter (with options `"abs"` or `"rel"`) to the `EarlyStopping` class, allowing users to specify whether `min_delta` is treated as an absolute or relative improvement. Default is `"rel"`. [[1]](diffhunk://#diff-90b35f252a4f1589bd64e544cc0fa83f556b0ae987d0f625cf66aa74df7ffcbfR23-R24) [[2]](diffhunk://#diff-90b35f252a4f1589bd64e544cc0fa83f556b0ae987d0f625cf66aa74df7ffcbfR54)
* Updated the constructor to validate the `min_delta_mode` argument and store it as an instance attribute. [[1]](diffhunk://#diff-90b35f252a4f1589bd64e544cc0fa83f556b0ae987d0f625cf66aa74df7ffcbfR68-R70) [[2]](diffhunk://#diff-90b35f252a4f1589bd64e544cc0fa83f556b0ae987d0f625cf66aa74df7ffcbfR79-R90)
* Modified the early stopping logic in `__call__` to compute the improvement threshold based on the selected `min_delta_mode`.

**Type annotation updates:**

* Updated type annotations to import `Literal` from `typing` for the new parameter.

Part of issue #3411

### DIFF
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -21,6 +21,10 @@ class EarlyStopping(Serializable):
         cumulative_delta: It True, `min_delta` defines an increase since the last `patience` reset, otherwise,
             it defines an increase after the last event. Default value is False.
         min_delta_mode: Determine whether `min_delta` is an absolute increase or a relative increase.
+            In 'abs' mode, the threshold is min_delta,
+            i.e. an increase of less than or equal to min_delta, will count as no improvement.
+            In 'rel' mode, the threshold is abs(best_score) * min_delta,
+            i.e. an increase of less than or equal to abs(best_score) * min_delta, will count as no improvement.
             Possible values are "abs" and "rel". Default value is "abs".
 
     Examples:


aaishwarymishra:
The pytest and tests look very scary. Anyways, I could also refactor type annotations for these file in this pr if its fine.

ignite/handlers/early_stopping.py
```diff
@@ -70,13 +76,18 @@ def __init__(
...
     def __call__(self, engine: Engine) -> None:
         score = self.score_function(engine)
 
         if self.best_score is None:
             self.best_score = score
-        elif score <= self.best_score + self.min_delta:
+            return
+        upper_bound = (
+            self.best_score + self.min_delta if self.min_delta_mode == "abs" else self.best_score + abs(self.best_score) * self.min_delta
+        )
```
vfdev-5, line 95: why do you use abs: `abs(self.best_score)`?
In the issue they proposed: `self.best_score * (1 + self.min_delta)`, I think this is reasonable. What do you think?

aaishwarymishra: The early stopping handlers assumes higher value as better so for loss we take negative, so loss of -0.1 is better than loss of -0.2.

For abs say our best was -0.4 and min_delta as 0.1 then the upper bound would be (-0.4 + 0.1) = -0.3.

But for rel the proposed formula was best*(1+min_delta) which works for the positive values but fails for negative.

Like (-0.4)*(1+0.1) = -0.44 which is incorrect.

So we altered the best*(1+min_delta)  to best+abs(best)*min_delta.

By this new formula (-0.4)+(0.4)*(0.1) = -0.36 Which is the correct result.

ignite/handlers/early_stopping.py
```diff
@@ -20,6 +20,8 @@ class EarlyStopping(Serializable):
             i.e. an increase of less than or equal to `min_delta`, will count as no improvement.
         cumulative_delta: It True, `min_delta` defines an increase since the last `patience` reset, otherwise,
             it defines an increase after the last event. Default value is False.
+        min_delta_mode: Determine whether `min_delta` is an absolute increase or a relative increase.
```
vfdev-5, line 23: Can you also update above sentence "i.e. an increase of less than or equal to `min_delta`, will count as no improvement." which assumes abs mode.

ignite/handlers/early_stopping.py
vfdev-5: OK, sounds good

vfdev-5:
Please fix : https://github.com/pytorch/ignite/actions/runs/21744233058/job/62741966345?pr=3516

ignite/handlers/early_stopping.py
aaishwarymishra: Does the new one seem fine ? or should I change it to something else.

ignite/handlers/early_stopping.py
```diff
@@ -17,9 +17,11 @@ class EarlyStopping(Serializable):
             object, and return a score `float`. An improvement is considered if the score is higher.
         trainer: Trainer engine to stop the run if no improvement.
         min_delta: A minimum increase in the score to qualify as an improvement,
-            i.e. an increase of less than or equal to `min_delta`, will count as no improvement.
+            i.e. an increase of less than or equal to the minimum delta threshold (as determined by min_delta and min_delta_mode), will count as no improvement.
```
vfdev-5, line 20:
```suggestion
            i.e. an increase of less than or equal to ``min_delta`` (assuming ``min_delta_mode="abs"``), will count as no improvement.
```

aaishwarymishra: Should we also mention formula for the abs and rel calculation in doc string for more clarity or Should I just commit this suggestion?

vfdev-5: Yes, good point, let's put the formula explaining what is "rel" mode in `min_delta_mode` arg description and give an example similar to "i.e. an increase of less than or equal to `min_delta`, will count as no improvement."

aaishwarymishra: I have updated the doc string, I think mentioning what count as no improvement should be explained in min_delta_mode as it shows the formula for calculating the threshold.

ignite/handlers/early_stopping.py
```diff
@@ -17,9 +17,15 @@ class EarlyStopping(Serializable):
...
         cumulative_delta: It True, `min_delta` defines an increase since the last `patience` reset, otherwise,
             it defines an increase after the last event. Default value is False.
+        min_delta_mode: Determine whether `min_delta` is an absolute increase or a relative increase.
```
vfdev-5, line 23: @aaishwarymishra I forgot, we should update the docstring with `.. versionchanged::` directlve for the new arg. For example: https://github.com/pytorch/ignite/blob/d2020e4e253ac1455a757c2db895c68ccfd2b958/ignite/metrics/metric.py#L281-L282

- https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md#writing-documentation

aaishwarymishra: I can add all the changes in docs together in next pr if it sounds good?

PR #3519, opened 2026-02-06
refactor: improve Checkpoint-Serialize consistency for state_dict

rwtarpit:
Fixes #{issue number}
NA
Description:
Refactored `Checkpoint.state_dict` and `Checkpoint.load_state_dict` to remove hard-coded `"_saved"` strings.
Uses base class `Serializable`'s validation mechanism to verify state_dict.
Check list:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -83,8 +83,11 @@ jobs:
         run: bash .github/workflows/install_docs_deps.sh
 
       - name: make linkcheck
-        working-directory: ./docs/
-        run: make linkcheck
+        uses: nick-fields/retry@v3.0.2
+        with:
+          max_attempts: 3
+          timeout_minutes: 10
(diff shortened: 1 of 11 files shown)


ignite/handlers/checkpoint.py
```diff
@@ -778,7 +778,8 @@ def load_state_dict(self, state_dict: Mapping) -> None:
             state_dict: a dict with "saved" key and list of ``(priority, filename)`` pairs as values.
         """
         super().load_state_dict(state_dict)
-        self._saved = [Checkpoint.Item(p, f) for p, f in state_dict["_saved"]]
+        key = self._state_dict_all_req_keys[0]
```
vfdev-5: I think it's fine to keep the previous code here.
The line `key = self._state_dict_all_req_keys[0]` is not unclear

PR #3525, opened 2026-02-09
Modernize type hints in time_profilers.py to Python 3.10+ syntax

dhawalc:
Migrates `ignite/handlers/time_profilers.py` from legacy typing syntax to Python 3.10+ union operators per #3481.

## Changes
- `Union[A, B]` → `A | B` (function signatures, local variables, return types)
- `Dict[K, V]` → `dict[K, V]` (annotations throughout)
- `List[T]` → `list[T]` (annotations throughout)
- `Tuple[A, B]` → `tuple[A, B]` (annotations throughout)
- Removed unused `Dict`, `List`, `Tuple`, `Union` imports from `typing` module
- Kept `Any`, `Callable`, `cast`, `Mapping`, `Sequence` as they have no builtin equivalent

Both `BasicTimeProfiler` and `HandlersTimeProfiler` classes are updated.

## Checklist
- [x] Single file change as requested in #3481
- [x] No functional changes — only type annotation syntax
- [x] Python syntax validated via `ast.parse()`

### DIFF
--- a/ignite/handlers/time_profilers.py
+++ b/ignite/handlers/time_profilers.py
@@ -1,6 +1,6 @@
 import functools
 from collections import OrderedDict
-from typing import Any, Callable, cast, Dict, List, Mapping, Sequence, Tuple, Union
+from typing import Any, Callable, cast, Mapping, Sequence
 
 import torch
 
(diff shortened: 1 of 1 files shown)


ignite/handlers/time_profilers.py
```diff
@@ -1,6 +1,6 @@
 import functools
 from collections import OrderedDict
-from typing import Any, Callable, cast, Dict, List, Mapping, Sequence, Tuple, Union
+from typing import Any, Callable, cast, Mapping, Sequence
```
vfdev-5, line 3: @copilot Let's also replace Mapping with dict in this file

vfdev-5:
We can close this PR as time_profilers.py has already updated type hints

PR #3530, opened 2026-02-11
[Feature] Implement HitRate metric for RecSys

rwtarpit:
Implements Hitrate metric as proposed in #2631 

Description:
Implements hitrate metric for recommendation systems. The hitrate is calculated for the ` top-k` list provided by the user and returns a list of `hitrate` in sorted order. A "hit" is considered when atleast one of  our model's top-k predictions for recommedations is actually correct.
Test cases and docstrings are also implemented.
 
 I followed the code structure of other metrics like `loss`, `KL divergence`, etc.
 
Reference : https://github.com/catalyst-team/catalyst/blob/master/catalyst/metrics/functional/_hitrate.py
Also i used Gemini initially for understanding how recommendation systems and its evaluation/metrics work, and for understanding disrtibuted tests.

Check list:

### DIFF
--- a/ignite/metrics/rec_sys/hitrate.py
+++ b/ignite/metrics/rec_sys/hitrate.py
@@ -80,8 +80,8 @@ class HitRate(Metric):
 
         .. testcode:: 2
 
-            metric = HitRate(top_k=[1, 2, 3, 4])
-            metric.attach(default_evaluator,"hit_rate", ignore_zero_hits=False)
+            metric = HitRate(top_k=[1, 2, 3, 4], ignore_zero_hits=False)
+            metric.attach(default_evaluator,"hit_rate")
             y_pred=torch.Tensor([
                 [4.0, 2.0, 3.0, 1.0],
                 [1.0, 2.0, 3.0, 4.0]


ignite/metrics/rec_sys/hitrate.py
```diff
@@ -0,0 +1,121 @@
...
+
+        self._num_examples += y.shape[0]
+
+    @sync_all_reduce("_hits_per_k", "_num_examples")
+    def compute(self) -> dict[int, float]:
+        if self._num_examples == 0:
+            raise NotComputableError("HitRate must have at least one example.")
+
+        rates = (self._hits_per_k / self._num_examples).tolist()
+        return {k: rate for k, rate in zip(self.top_k, rates)}
```
vfdev-5: Can we just return a list of float instead of the dict?

ignite/metrics/rec_sys/hitrate.py
```diff
@@ -0,0 +1,121 @@
...
+        self.top_k = sorted(top_k)
+        super(HitRate, self).__init__(output_transform, device=device, skip_unrolling=skip_unrolling)
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._hits_per_k = torch.zeros(len(self.top_k), device=self._device)
+        self._num_examples = 0
+
+    @reinit__is_reduced
+    def update(self, output: tuple[torch.Tensor, torch.Tensor]) -> None:
```
vfdev-5, line 127: In catalyst they specifically handle zero_division: https://github.com/catalyst-team/catalyst/blob/e99f90655d0efcf22559a46e928f0f98c9807ebf/catalyst/metrics/functional/_hitrate.py#L73
how do we handle that and where?

ignite/metrics/rec_sys/hitrate.py
```diff
@@ -0,0 +1,121 @@
...
+import torch
+
+from ignite.exceptions import NotComputableError
+from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
+
+__all__ = ["HitRate"]
+
+
+class HitRate(Metric):
+    """
```
vfdev-5: We may need to update the docstring according to the guidelines: https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md#writing-documentation
and add entry into [metrics.rst](https://github.com/pytorch/ignite/blob/master/docs/source/metrics.rst) to be seen in the documentation.

vfdev-5 (commented):
Thanks a lot for the PR @rwtarpit 
There are few things to handle and we should be good to go.

ignite/metrics/rec_sys/hitrate.py
rwtarpit: reason to return the `dict` was that since i used `max_k` to find the hits instead of looping though `top_k`, list won't return the in the same order as `top_k` if `top_k` is not sorted.
i'll add a check for `top_k` being sorted then.

ignite/metrics/rec_sys/hitrate.py
rwtarpit: i'll take a look at this

rwtarpit: i assume its about how to handle cold users(who don't pick any item) and we want to make a choice whether or  not to include them in total count (`_num_examples`).
we can add a intuitive parameter (likely `count_cold_users`) as a boolean flag, and can simply count the user in `_num_examples` if its set `True` and vice versa.

vfdev-5: Can you search in other python packages implementing this metric how they call this arg and we can use the same name.

rwtarpit: i checked a few, but most of them filter the batch for empty rows. couldn't find any similar implementation.
we can either just ignore such empty rows, or if we want to be explicit, `ignore_empty` flag sounds good

ignite/metrics/rec_sys/hitrate.py
```diff
@@ -0,0 +1,121 @@
...
+    Calculates the Hit Rate at `k` for Recommendation Systems.
+
+    The Hit Rate measures the fraction of users for which the model was able
+    to predict atleast one correct recommendation.
```
vfdev-5:
```suggestion
    to predict at least one correct recommendation.
```

ignite/metrics/rec_sys/hitrate.py
```diff
@@ -0,0 +1,121 @@
...
+    `Hit` for each user is either 0 or 1, irrespective of how many correct recommendations
+    our model was able to predict for that user.
```
vfdev-5:
```suggestion
    the model was able to predict for that user.
```

ignite/metrics/rec_sys/hitrate.py
vfdev-5: @rwtarpit please check above comment

ignite/metrics/rec_sys/hitrate.py
```diff
@@ -0,0 +1,121 @@
...
+
+    .. math:: \text{HR}@K = \frac{1}{N} \sum_{i=1}^{N} \mathbb{1}(\text{rank}_i \leq K)
+
+    where :math:`\text{rank}_i` is rank of the first relevant item in the predicted
+    tensor :math:`\mathbf{q}_i` that exists in the ground truth tensor :math:`\mathbf{p}_i`.
+
+    - ``update`` must receive output of the form ``(y_pred, y)``.
+    - ``y_pred`` is expected to be raw logits or probability score for each item in the catalog.
+    - ``y`` is expected to be binary (only 0s and 1s) values where `1` indicates relevant item.
+    - ``y_pred`` and ``y`` are only allowed shape :math:`(batch,num_items)`.
```
vfdev-5:
```suggestion
    - ``y_pred`` and ``y`` are only allowed shape :math:`(batch, num_items)`.
```

ignite/metrics/rec_sys/hitrate.py
```diff
@@ -0,0 +1,121 @@
...
+            of shape ``(batch, num_items)``.
+        device: specifies which device updates are accumulated on. Setting the
+            metric's device to be the same as your ``update`` arguments ensures the ``update`` method is
+            non-blocking. By default, CPU.
+        skip_unrolling: specifies whether input should be unrolled or not before being
+            processed. Should be true for multi-output models.
+
+    Attributes:
+        required_output_keys: dictionary defines required keys to be found in ``engine.state.output`` if the
+        latter is a dictionary. Default, ``("y_pred", "y")``.
```
vfdev-5:
```suggestion
```

ignite/metrics/rec_sys/hitrate.py
```diff
@@ -0,0 +1,121 @@
...
+        To use with ``Engine`` and ``process_function``, simply attach the metric instance to the engine.
+        The output of the engine's ``process_function`` needs to be in the format of
+        ``(y_pred, y)``. If not, ``output_tranform`` can be added
+        to the metric to transform the output into the form expected by the metric.
+
+        For more information on how metric works with :class:`~ignite.engine.engine.Engine`, visit :ref:`attach-engine`.
+
+        .. testcode::
+
+            metric = HitRate(top_k = [1,2,3,4])
```
vfdev-5:
```suggestion
            metric = HitRate(top_k=[1, 2, 3, 4])
```

ignite/metrics/rec_sys/hitrate.py
```diff
@@ -0,0 +1,121 @@
...
+            metric.attach(default_evaluator,'hit_rate')
+            y_pred=torch.Tensor([
+                [4.0, 2.0, 3.0, 1.0],
+                [1.0, 2.0, 3.0, 4.0]
+            ]),
+            y_true=torch.Tensor([
+                [0, 0, 1.0, 1.0],
+                [0, 0, 0.0, 0.0]
+            ])
```
vfdev-5, line 71: Why predicitons and ground truth are mixing int and float values?

ignite/metrics/rec_sys/hitrate.py
```diff
@@ -0,0 +1,121 @@
...
+        The output of the engine's ``process_function`` needs to be in the format of
+        ``(y_pred, y)``. If not, ``output_tranform`` can be added
+        to the metric to transform the output into the form expected by the metric.
+
+        For more information on how metric works with :class:`~ignite.engine.engine.Engine`, visit :ref:`attach-engine`.
+
+        .. testcode::
+
+            metric = HitRate(top_k = [1,2,3,4])
+            metric.attach(default_evaluator,'hit_rate')
```
vfdev-5:
```suggestion
            metric.attach(default_evaluator, "hit_rate")
```

ignite/metrics/rec_sys/hitrate.py
```diff
@@ -0,0 +1,121 @@
...
+            y_pred=torch.Tensor([
+                [4.0, 2.0, 3.0, 1.0],
+                [1.0, 2.0, 3.0, 4.0]
+            ]),
+            y_true=torch.Tensor([
+                [0, 0, 1.0, 1.0],
+                [0, 0, 0.0, 0.0]
+            ])
+            state = default_evaluator.run([[y_pred, y_true]])
+            print(state.metrics['hit_rate'])
```
vfdev-5:
```suggestion
            print(state.metrics["hit_rate"])
```

ignite/metrics/rec_sys/hitrate.py
rwtarpit: @vfdev-5 how should we approach this?

vfdev-5: I have to take time a bit to understand exactly the problem. If you can provide detailed description of it, it could be helpful.

rwtarpit: its basically how we handle the users with no picks / ground truth tensor for a user being all zeros. (`torch.zeros`)
ex: 
`preds = torch.tensor([[0.8, 0.2], # user_1`
                                 `                       [0.8, 0.5]]) # user_2 `
`targets = torch.tensor([[1.0, 0.0], # user_1`
                                   `                        [0.0, 0.0]]) # user_2`

`hit_rate = total_hits/total_users` 
for `top_k`=1, 
if we consider the user who doesn't choose any item, ie the cold user,
we would have 2 choices:
ignore that user : in this case hitrate would be `1.0`, cause effectively we only have one user 
count that user : in this case hitrate would be `0.5`, cause now we will count that  cold user but it won't have a hit.

now for this, we can ideally ignore such rows/tensors in the dataset, or we can let user handle this with a boolean flag (`ignore_empty`).
the python packages i checked that  were using explicit arg system didn't had a common name for this arg, 
while others of them just simply either filter the dataset with such empty rows or weren't counting them during metric calculation

vfdev-5: Thanks for details, @rwtarpit !
Let's add a flag: `ignore_zero_hits` and explain in the docstring what it is doing (similarly to your comment here).

rwtarpit:
next step after hitrate could be recall@top_k which has similar logic but instead of binary hits per user, it calculates fraction/percentage of hits per user.

vfdev-5:
> next step after hitrate could be recall@top_k which has similar logic but instead of binary hits per user, it calculates fraction/percentage of hits per user.

We already have Precision/Recall metrics, so if we could add top_k wrapper or arg to the existing metrics, it would be better than introducing new metric class

rwtarpit:
yes that sounds  better!!
i'll take a look and try to find the  best approach

PR #3531, opened 2026-02-11
MetricsLambda and ClassificationReport ability to use metrics_result_mode.

### LINKED ISSUE #3513: Nested metrics are being duplicated when logging.

gabrielfruet:
Fixes #3513

The main reason the issue #3513 was created particularly for me was due to the ClassificationReport behavior. So i'm creating this PR to add this functionality following up the #3514 PR.

### DIFF
--- a/ignite/metrics/metrics_lambda.py
+++ b/ignite/metrics/metrics_lambda.py
@@ -116,9 +116,7 @@ class MetricsLambda(Metric):
         self.kwargs = kwargs
         self.engine: Optional[Engine] = None
         self._updated = False
-        super(MetricsLambda, self).__init__(
-            device="cpu", metrics_result_mode=metrics_result_mode, skip_unrolling=skip_unrolling
-        )
+        super().__init__(device="cpu", metrics_result_mode=metrics_result_mode, skip_unrolling=skip_unrolling)
 
     @reinit__is_reduced
     def reset(self) -> None:
--- a/ignite/metrics/accumulation.py
+++ b/ignite/metrics/accumulation.py
@@ -57,9 +57,7 @@ class VariableAccumulation(Metric):
 
         self._op = op
 
-        super(VariableAccumulation, self).__init__(
-            output_transform=output_transform, device=device, skip_unrolling=skip_unrolling
-        )
+        super().__init__(output_transform=output_transform, device=device, skip_unrolling=skip_unrolling)
 
     @reinit__is_reduced
     def reset(self) -> None:
@@ -191,9 +189,7 @@ class Average(VariableAccumulation):
                 x = x.sum(dim=0)
             return a + x
 
-        super(Average, self).__init__(
-            op=_mean_op, output_transform=output_transform, device=device, skip_unrolling=skip_unrolling
-        )
+        super().__init__(op=_mean_op, output_transform=output_transform, device=device, skip_unrolling=skip_unrolling)
 
     @sync_all_reduce("accumulator", "num_examples")
     def compute(self) -> float | torch.Tensor:
@@ -308,9 +304,7 @@ class GeometricAverage(VariableAccumulation):
                 x = x.sum(dim=0)
             return a + x
 
-        super(GeometricAverage, self).__init__(
-            op=_geom_op, output_transform=output_transform, device=device, skip_unrolling=skip_unrolling
-        )
+        super().__init__(op=_geom_op, output_transform=output_transform, device=device, skip_unrolling=skip_unrolling)
 
     @sync_all_reduce("accumulator", "num_examples")
     def compute(self) -> float | torch.Tensor:
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -19,9 +19,7 @@ class _BaseClassification(Metric):
         self._is_multilabel = is_multilabel
         self._type: str | None = None
         self._num_classes: int | None = None
-        super(_BaseClassification, self).__init__(
-            output_transform=output_transform, device=device, skip_unrolling=skip_unrolling
-        )
+        super().__init__(output_transform=output_transform, device=device, skip_unrolling=skip_unrolling)
 
     def reset(self) -> None:
         self._type = None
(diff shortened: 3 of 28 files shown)


ignite/metrics/metrics_lambda.py
```diff
@@ -92,13 +92,15 @@ def Fbeta(r, p, beta):
 
     _state_dict_all_req_keys = ("_updated", "args", "kwargs")
 
-    def __init__(self, f: Callable, *args: Any, **kwargs: Any) -> None:
+    def __init__(
+        self, f: Callable, *args: Any, metrics_result_mode: Literal["flatten", "named", "both"] = "both", **kwargs: Any
```
vfdev-5: I'm not totally sure we should propagate `metrics_result_mode` here as we do not propagate other args like device etc

vfdev-5 (commented):
Thanks for the PR @gabrielfruet 
I left a comment

ignite/metrics/metrics_lambda.py
gabrielfruet: I'll just guarantee that it's working without the explicit KW. If so, then i'll change it.

Might take couple days since i'm on vacation.

vfdev-5: @gabrielfruet once you are back, can you please address this comment. I think we should not propagate the kwargs

gabrielfruet: I was addressing this and one thing that i noticed:

1. We cannot just pass the *kwargs ot the __init__, since we have kwargs that are related to the `MetricsLambda` and the super init don't gather all kwargs, so extra kwargs causes errors.

```
FAILED tests/ignite/metrics/test_metrics_lambda.py::test_metrics_lambda - TypeError: Metric.__init__() got an unexpected keyword argument 'other'
FAILED tests/ignite/metrics/test_metrics_lambda.py::test_metrics_lambda_reset - TypeError: Metric.__init__() got an unexpected keyword argument 'z'
```

2. What we can do is explicitly extracting the metrics_result_mode from the kwargs and forwarding it to the `super().__init__`

```python
metrics_result_mode = kwargs["metrics_result_mode"] if "metrics_result_mode" in kwargs else "both"
super(MetricsLambda, self).__init__(device="cpu", metrics_result_mode=metrics_result_mode)

```

vfdev-5: Thanks for the update! I need to check deeper this point from my side.

vfdev-5: I think we should propagate skip_unrolling and metrics_result_mode directly to the base class as following:
```python
    def __init__(
        self,
        f: Callable,
        *args: Any,
        skip_unrolling: bool = False,
        metrics_result_mode: Literal["flatten", "named", "both"] = "both",
        **kwargs: Any
    ) -> None:
        self.function = f
        self.args = list(args)  # we need args to be a list instead of a tuple for state_dict/load_state_dict feature
        self.kwargs = kwargs
        self.engine: Optional[Engine] = None
        self._updated = False
        super(MetricsLambda, self).__init__(device="cpu", metrics_result_mode=metrics_result_mode, skip_unrolling=skip_unrolling)
```

gabrielfruet: Thank you for the update. Just to ensure we're in the same page, 
this is basically the way i did initially, right?

Shouldn't we propagate device also?

vfdev-5: Yes, you are right, it should be like you did in the very beginning, sorry for going back and forth. I was thinking about other args and output_transform and device can't be used in MetricsLambda.

ignite/metrics/classification_report.py
```diff
@@ -18,6 +18,7 @@ def ClassificationReport(
     device: Union[str, torch.device] = torch.device("cpu"),
     is_multilabel: bool = False,
     labels: Optional[List[str]] = None,
+    metrics_result_mode: Literal["flatten", "named", "both"] = "both",
```
vfdev-5, line 21: Please add the new arg to the docstring and in the docstring we should add `versionchanged` tag (for details see CONTRIBUTING.md: https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md#writing-documentation) or MetricLambda my commit to your PR

gabrielfruet:
Thank you for this final polish! Really appreciate that.

PR #3538, opened 2026-02-17
Modernize type hints to Python 3.10+ in ignite/contrib/engines/tbptt.py

Ayush-Aditya:
Part of #3481

This PR updates type hints in `ignite/contrib/engines/tbptt.py` to Python 3.10+ syntax:

- Replaced `Union` with `|`
- Replaced `Optional[str]` with `str | None`
- Replaced `typing.Sequence` with `collections.abc.Sequence`
- Removed unused typing imports

No functional changes were made. All tests pass locally.

### DIFF
--- a/ignite/metrics/regression/canberra_metric.py
+++ b/ignite/metrics/regression/canberra_metric.py
@@ -36,7 +36,7 @@ class CanberraMetric(_BaseRegression):
             non-blocking. By default, CPU.
 
     .. _`Botchkarev 2018`:
-        https://arxiv.org/ftp/arxiv/papers/1809/1809.03006.pdf
+        https://arxiv.org/abs/1809.03006
 
     Examples:
         To use with ``Engine`` and ``process_function``, simply attach the metric instance to the engine.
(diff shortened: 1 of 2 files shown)


ignite/metrics/regression/canberra_metric.py
```diff
@@ -1,81 +1,81 @@
...
-
-    def _update(self, output: Tuple[torch.Tensor, torch.Tensor]) -> None:
-        y_pred, y = output[0].detach(), output[1].detach()
-        errors = torch.abs(y - y_pred) / (torch.abs(y_pred) + torch.abs(y) + 1e-15)
-        self._sum_of_errors += torch.sum(errors).to(self._device)
-
-    @sync_all_reduce("_sum_of_errors")
-    def compute(self) -> float:
-        return self._sum_of_errors.item()
+from typing import Tuple
```
vfdev-5: Thanks for fixing linkcheck test! However the diff looks strange, the URL is only changed but it shows complete file change. Can you revert the changes in both files: canberra and MARE metrics and apply only url change? Thanks!

Ayush-Aditya: Thanks for the feedback!
I've reverted the unintended formatting changes and applied only the URL fix in both files. The diff should now contain only the link updates. Please let me know if anything else needs adjustment.

vfdev-5 (approved):
Thanks!

PR #3540, opened 2026-02-18
Bug fix in object detection metric

obsessor-ak1:
**Bug**
In the code for `ObjectDetectionAvgPrecisionRecall` metric class constructor, the member `_iou_thresholds` initially was not on device specified in the `__init__` parameters unless the user explicitly passes `iou_thresholds` parameter as an instance of`torch.Tensor` with specified device same as `device` parameter.

**Effect**
This will produce a `RuntimeError` when calling the `update` method of this metric when the device for `torch.Tensor` instances nested within `output` parameter will live on a different device other than the CPU, e.g.:
```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```
This will also affect other metrics like `CommonObjectDetectionMetrics` which use one/multiple instances for this metric

**Solution**
Just turn the `iou_thresholds` parameter to a `torch.Tensor` instance with a specified device. So, I just turned that to a tensor and optionally for 'rec_thresholds` too.

**How to reproduce?**

Just run the following code snippet in a notebook or a script:
```python
import torch
from ignite.metrics import ObjectDetectionAvgPrecisionRecall

# Dummy targets/predictions for a batch of 8 images (300x300), SSD-style object detection
num_images = 8
num_classes = 21
image_size = 300

device = torch.device("cuda")

g = torch.Generator(device=device).manual_seed(42)

dummy_targets = []
dummy_predictions = []

# Generating some dummy predictions and targets

for _ in range(num_images):
    # Targets
    n_tgt = int(torch.randint(1, 5, (1,), generator=g, device=device).item())  # 1..4 objects
    x1 = torch.randint(0, image_size - 60, (n_tgt,), generator=g, device=device).float()
    y1 = torch.randint(0, image_size - 60, (n_tgt,), generator=g, device=device).float()
    w = torch.randint(20, 100, (n_tgt,), generator=g, device=device).float()
    h = torch.randint(20, 100, (n_tgt,), generator=g, device=device).float()
    x2 = torch.clamp(x1 + w, max=image_size - 1)
    y2 = torch.clamp(y1 + h, max=image_size - 1)
    tgt_boxes = torch.stack([x1, y1, x2, y2], dim=1)
    # In SSD Detection 0 represents background class so not included
    tgt_labels = torch.randint(1, num_classes-1, (n_tgt,), generator=g, device=device, dtype=torch.int64)

    dummy_targets.append({
        "bbox": tgt_boxes,
        "labels": tgt_labels
    })

    # Predictions
    n_pred = int(torch.randint(2, 7, (1,), generator=g, device=device).item())  # 2..6 detections
    x1p = torch.randint(0, image_size - 60, (n_pred,), generator=g, device=device).float()
    y1p = torch.randint(0, image_size - 60, (n_pred,), generator=g, device=device).float()
    wp = torch.randint(20, 120, (n_pred,), generator=g, device=device).float()
    hp = torch.randint(20, 120, (n_pred,), generator=g, device=device).float()
    x2p = torch.clamp(x1p + wp, max=image_size - 1)
    y2p = torch.clamp(y1p + hp, max=image_size - 1)
    pred_boxes = torch.stack([x1p, y1p, x2p, y2p], dim=1)
    # In SSD Detection 0 represents background class so not included
    pred_labels = torch.randint(1, num_classes-1, (n_pred,), generator=g, device=device, dtype=torch.int64)
    pred_scores = torch.rand((n_pred,), generator=g, device=device)

    dummy_predictions.append({
        "bbox": pred_boxes, 
        "labels": pred_labels,
        "scores": pred_scores
    })

print(len(dummy_targets), len(dummy_predictions))
print(dummy_targets[0].keys(), dummy_predictions[0].keys())


print("Dummy Target Example:", dummy_predictions[0])

metric = ObjectDetectionAvgPrecisionRecall(
    iou_thresholds=[0.5],
    num_classes=20,
    device=device
)
metric.reset()

metric.update((dummy_predictions, dummy_targets))
```
Output:
```
dict_keys(['bbox', 'labels']) dict_keys(['bbox', 'labels', 'scores'])
Dummy Target Example: {'bbox': tensor([[ 33.,  92.,  69., 185.],
        [127., 166., 226., 213.],
        [214., 125., 299., 184.],
        [ 64., 232., 147., 299.]], device='cuda:0'), 'labels': tensor([12, 16, 20, 13], device='cuda:0'), 'scores': tensor([0.3806, 0.7871, 0.7982, 0.8877], device='cuda:0')}
  File "D:\Code\Projects\torch-ignite\dummy_testing.py", line 65, in <module>
    metric.update((dummy_predictions, dummy_targets))
  File "D:\Code\Projects\torch-ignite\ignite\metrics\metric.py", line 888, in wrapper
    func(self, *args, **kwargs)
  File "D:\Code\Projects\torch-ignite\ignite\metrics\vision\object_detection_average_precision_recall.py", line 343, in update
    ious[ious < self._iou_thresholds] = NO_MATCH
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

### DIFF
--- a/ignite/metrics/mean_average_precision.py
+++ b/ignite/metrics/mean_average_precision.py
@@ -1,8 +1,8 @@
 import warnings
-from typing import Callable, cast, Sequence
+from collections.abc import Callable, Sequence
+from typing import cast, Literal
 
 import torch
-from typing_extensions import Literal
 
 import ignite.distributed as idist
 from ignite.distributed.utils import all_gather_tensors_with_shapes
--- a/ignite/metrics/vision/object_detection_average_precision_recall.py
+++ b/ignite/metrics/vision/object_detection_average_precision_recall.py
@@ -1,10 +1,10 @@
-from typing import Callable, cast, Sequence
+from collections.abc import Callable, Sequence
+from typing import cast
 
 import torch
 from typing_extensions import Literal
 
 from ignite.metrics import MetricGroup
-
 from ignite.metrics.mean_average_precision import _BaseAveragePrecision, _cat_and_agg_tensors
 from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
 
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -5,3 +5,5 @@
 fe5a29b6942829ff2870181b6c1a68ebd2af9add
 # Fix code style (#3423)
 f9e8ca106a0d3f386f26f3a9481fd2eb341999f0
+# Add nb-clean to pre-commit config and clean codebase (#3618)
+8755e2c13979b24736c6e3c6716cb41f95fcfcce
--- a/.github/workflows/code-style-checks.yml
+++ b/.github/workflows/code-style-checks.yml
@@ -48,12 +48,18 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Get year & week number
+        id: get-date
+        run: |
+          echo "date=$(/bin/date "+%Y-%U")" >> $GITHUB_OUTPUT
+
       - uses: astral-sh/setup-uv@v7
         with:
           version: "latest"
           python-version: ${{ matrix.python-version }}
           activate-environment: true
           enable-cache: true
+          cache-suffix: "${{ steps.get-date.outputs.date }}-style-${{ runner.os }}-${{ matrix.python-version }}"
           cache-dependency-glob: |
             **/requirements-dev.txt
             **/pyproject.toml
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,13 +26,19 @@ jobs:
         python-version: ["3.10"]
     steps:
       - uses: actions/checkout@v6
+
+      - name: Get year & week number
+        id: get-date
+        run: |
+          echo "date=$(/bin/date "+%Y-%U")" >> $GITHUB_OUTPUT
+
       - uses: astral-sh/setup-uv@v7
         with:
           version: "latest"
           python-version: "${{ matrix.python-version }}"
           activate-environment: true
           enable-cache: true
-          cache-suffix: "${{ hashFiles('requirements-dev.txt') }}-${{ hashFiles('docs/requirements.txt') }}"
+          cache-suffix: "${{ steps.get-date.outputs.date }}-${{ hashFiles('requirements-dev.txt') }}-${{ hashFiles('docs/requirements.txt') }}"
           cache-dependency-glob: |
             **/docs/requirements.txt
             **/requirements-dev.txt
@@ -67,13 +73,19 @@ jobs:
         python-version: ["3.10"]
     steps:
       - uses: actions/checkout@v6
+
+      - name: Get year & week number
+        id: get-date
+        run: |
+          echo "date=$(/bin/date "+%Y-%U")" >> $GITHUB_OUTPUT
+
       - uses: astral-sh/setup-uv@v7
         with:
           version: "latest"
           python-version: ${{ matrix.python-version }}
           activate-environment: true
           enable-cache: true
-          cache-suffix: "${{ hashFiles('requirements-dev.txt') }}-${{ hashFiles('docs/requirements.txt') }}"
+          cache-suffix: "${{ steps.get-date.outputs.date }}-${{ hashFiles('requirements-dev.txt') }}-${{ hashFiles('docs/requirements.txt') }}"
           cache-dependency-glob: |
             **/docs/requirements.txt
             **/requirements-dev.txt
@@ -100,13 +112,19 @@ jobs:
         python-version: ["3.10"]
     steps:
       - uses: actions/checkout@v6
+
+      - name: Get year & week number
+        id: get-date
+        run: |
+          echo "date=$(/bin/date "+%Y-%U")" >> $GITHUB_OUTPUT
+
       - uses: astral-sh/setup-uv@v7
         with:
           version: "latest"
           python-version: "${{ matrix.python-version }}"
           activate-environment: true
           enable-cache: true
-          cache-suffix: "${{ hashFiles('requirements-dev.txt') }}-${{ hashFiles('docs/requirements.txt') }}"
+          cache-suffix: "${{ steps.get-date.outputs.date }}-${{ hashFiles('requirements-dev.txt') }}-${{ hashFiles('docs/requirements.txt') }}"
           cache-dependency-glob: |
             **/docs/requirements.txt
             **/requirements-dev.txt
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -184,3 +184,8 @@ jobs:
           else
             echo "Caltech101 dataset failed to download. Skipping SR example test."
           fi
+      - name: Run CIFAR10 with Ray Tune and Ignite
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        run: |
+          uv pip install -r examples/cifar10_raytune/requirements.txt
+          python examples/cifar10_raytune/main.py --num_epochs 1 --num_trials 1 --cpus_per_trial 1 --gpus_per_trial 0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,15 @@
 exclude: "^conda.recipe|assets|docs"
+
 repos:
+  - repo: https://github.com/srstevenson/nb-clean
+    rev: 3.2.0
+    hooks:
+      - id: nb-clean
+        args:
+          - --preserve-cell-outputs
+          - --remove-empty-cells
+        types: [jupyter]
+
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
     hooks:
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,7 +1,7 @@
-sphinx<6
+sphinx==7
 git+https://github.com/pytorch/pytorch_sphinx_theme.git#egg=pytorch_sphinx_theme
 sphinxcontrib-katex
 sphinx-copybutton==0.4.0
-docutils<0.18
+docutils==0.19
 sphinx_togglebutton
 sphinx_design
--- a/docs/source/_templates/_static/css/ignite_theme.css
+++ b/docs/source/_templates/_static/css/ignite_theme.css
@@ -175,45 +175,10 @@ footer table {
 }
 
 /* dim out pytorch header
-algolia searchbox has z-index of 200
-so make sure not to go above that value */
+*/
 .header-holder {
   z-index: 100 !important;
 }
-
-:root {
-  --docsearch-primary-color: #ee4c2c !important;
-}
-
-.DocSearch-Button {
-  border-radius: 8px !important;
-  width: 100% !important;
-  height: 2.75rem !important;
-  margin: 0 !important;
-  background: var(--docsearch-searchbox-focus-background) !important;
-}
-
-.DocSearch-Button .DocSearch-Search-Icon {
-  color: var(--docsearch-primary-color) !important;
-}
-
-.DocSearch-Button-Key {
-  background: none !important;
-  box-shadow: none !important;
-  top: 0 !important;
-  padding-bottom: 0 !important;
-  color: inherit !important;
-  font-size: 1rem !important;
-  font-family: Arial !important;
-}
-
-@media (max-width: 750px) {
-  .DocSearch-Button-Key,
-  .DocSearch-Button-Placeholder {
-    display: block !important;
-  }
-}
-/* docsearch end */
 
 article.pytorch-article p.rubric {
   font-weight: bold;
--- a/docs/source/_templates/layout.html
+++ b/docs/source/_templates/layout.html
@@ -40,8 +40,6 @@
   <meta property="og:image:alt" content="PyTorch-Ignite logo">
   <meta property="og:description" content="High-level library to help with training and evaluating neural networks in PyTorch flexibly and transparently.">
 
-  <link rel="preconnect" href="https://BH4D9OD16A-dsn.algolia.net" crossorigin />
-
   {% block htmltitle %}
   <title>{{ title|striptags|e }}{{ titlesuffix }}</title>
   {% endblock %}
@@ -67,8 +65,10 @@
 
   {% endif %}
 
+  {%- if style is defined %}
   <link rel="preload" href="{{ pathto('_static/' + style, 1) }}" as="style" />
   <link rel="stylesheet" href="{{ pathto('_static/' + style, 1) }}" type="text/css" />
+  {%- endif %}
   <!-- <link rel="stylesheet" href="{{ pathto('_static/pygments.css', 1) }}" type="text/css" /> -->
   {%- for css in css_files %}
     {%- if css|attr("rel") %}
@@ -196,7 +196,13 @@
               {% endif %}
             {% endif %}
 
-            <div id="docsearch"></div>
+            <div role="search">
+              <form id="rtd-search-form" class="wy-form" action="{{ pathto('search') }}" method="get">
+                <input type="text" name="q" placeholder="{{ _('Search docs') }}" />
+                <input type="hidden" name="check_keywords" value="yes" />
+                <input type="hidden" name="area" value="default" />
+              </form>
+            </div>
 
             {% endblock %}
           </div>
@@ -515,25 +521,6 @@
         $(this).closest("a").addClass("has-code");
       });
     })
-  </script>
-  <script src="https://cdn.jsdelivr.net/npm/@docsearch/js@3"></script>
-  <script type="text/javascript">
-  let VERSION
-  if ('{{ version }}'.startsWith('v')) {
-    VERSION = '{{ version }}'
-  } else {
-    VERSION = 'master'
-  }
-  docsearch({
-    container: '#docsearch',
-    appId: '7EWYE1JCT3',
-    apiKey: '841e93e60c16975ba1bd8c7c716eed82',
-    indexName: 'pytorch-ignite',
-    placeholder: 'Search PyTorch-Ignite docs',
-    searchParameters: {
-      facetFilters: [`version:${VERSION}`, 'tags:API-reference'],
-    }
-  });
   </script>
 </body>
 </html>
--- a/examples/cifar10_raytune/.gitignore
+++ b/examples/cifar10_raytune/.gitignore
@@ -1,0 +1,4 @@
+/data/
+
+ray_tune_logs/
+ray_results/
--- a/examples/cifar10_raytune/README.md
+++ b/examples/cifar10_raytune/README.md
@@ -1,0 +1,105 @@
+# CIFAR10 Example with Ray Tune and Ignite
+
+This example demonstrates how to use **Ray Tune** for hyperparameter tuning with **PyTorch Ignite** for training and validation on the CIFAR10 dataset.
+
+In this example, we show how to use _Ignite_ to train a neural network with:
+
+- hyperparameter search using Ray Tune,
+- ASHA scheduler for early stopping, see also: https://docs.ray.io/en/latest/tune/api/doc/ray.tune.schedulers.ASHAScheduler.html
+
+## Requirements
+
+- pytorch-ignite: `pip install pytorch-ignite`
+- torch: `pip install torch`
+- torchvision: `pip install torchvision`
+- [Ray Tune](https://docs.ray.io/): `pip install ray[tune]`
+
+Alternatively, please see `requirements.txt`
+
+```bash
+pip install -r requirements.txt
+```
+
+## Usage
+
+Run with default settings (10 trials, 10 epochs):
+
+```bash
+python main.py
+```
+
+### Command-line arguments
+
+| Argument           | Default  | Description                   |
+| ------------------ | -------- | ----------------------------- |
+| `--num_trials`     | 10       | Number of trials for Ray Tune |
+| `--num_epochs`     | 10       | Number of epochs per trial    |
+| `--gpus_per_trial` | 1        | GPUs per trial (0 for CPU)    |
+| `--cpus_per_trial` | 4        | CPUs per trial                |
+| `--data_dir`       | `./data` | Path to CIFAR10 dataset       |
+
+### Examples
+
+Run with custom settings:
+
+```bash
+python main.py --num_trials 5 --num_epochs 5 --gpus_per_trial 1
+```
+
+Use a different data directory:
+
+```bash
+python main.py --data_dir /path/to/cifar10/
+```
+
+For more details on accepted arguments:
+
+```bash
+python main.py --help
+```
+
+## Search Space
+
+The example searches over the following hyperparameters:
+
+| Parameter    | Values                       |
+| ------------ | ---------------------------- |
+| `l1`         | [4, 8, 16, 32, 64, 128, 256] |
+| `l2`         | [4, 8, 16, 32, 64, 128, 256] |
+| `lr`         | [1e-4, 1e-1]                 |
+| `batch_size` | [256, 512, 1024]             |
+
+- `l1`: First fully connected layer size
+- `l2`: Second fully connected layer size
+- `lr`: Log-uniformly sampled learning rate in [1e-4, 1e-1] for SGD optimizer
+- `batch_size`: Training batch size
+
+## Output
+
+If we run main.py an example output could look like:
+
+```
+╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
+│ Trial name                      status         l1     l2            lr     batch_size     iter     total time (s)      loss     accuracy │
+├──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
+│ train_with_ignite_ab528_00000   TERMINATED      4    128   0.000202867           1024        1            4.28721   2.31388       0.0993 │
+│ train_with_ignite_ab528_00001   TERMINATED    128      4   0.0180925              256       10           20.296     1.24503       0.5711 │
+│ train_with_ignite_ab528_00002   TERMINATED      8    256   0.000386848           1024        1            4.01333   2.30241       0.1003 │
+│ train_with_ignite_ab528_00003   TERMINATED     16      4   0.00105969             256        1            3.59295   2.32562       0.0995 │
+│ train_with_ignite_ab528_00004   TERMINATED     16    128   0.000495626           1024        2            5.8557    2.30494       0.0876 │
+│ train_with_ignite_ab528_00005   TERMINATED     16    128   0.0379958              512       10           20.3416    1.20055       0.5765 │
+│ train_with_ignite_ab528_00006   TERMINATED      4     32   0.00186652            1024        1            4.06322   2.30845       0.1059 │
+│ train_with_ignite_ab528_00007   TERMINATED     16     16   0.00687211             512        2            5.40865   2.12892       0.2006 │
+│ train_with_ignite_ab528_00008   TERMINATED      4    256   0.000143329           1024        1            3.79915   2.30698       0.0908 │
+│ train_with_ignite_ab528_00009   TERMINATED     32    256   0.000630852            512        2            5.72633   2.29931       0.1221 │
+╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
+
+Best trial config: {'l1': 16, 'l2': 128, 'lr': 0.03799583796432589, 'batch_size': 512, 'device': 'cuda'}
+Best trial final validation loss: 1.200553125
+Best trial final validation accuracy: 0.5765
+Best trial test set accuracy: 0.5688
+```
+
+We see that most trials stop earlier than the number of maximum epochs (= 10, here). This is because the ASHAScheduler performs early stopping:
+after each reported epoch, it compares trial performance and prunes the worst-performing ones. So `num_epochs = 10` is the upper limit
+here, most trials stop earlier.
--- a/examples/cifar10_raytune/main.py
+++ b/examples/cifar10_raytune/main.py
@@ -1,0 +1,167 @@
+import argparse
+from pathlib import Path
+
+import torch
+import torch.nn as nn
+import torch.optim as optim
+
+import ray
+from ray import tune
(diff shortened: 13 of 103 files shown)


ignite/metrics/vision/object_detection_average_precision_recall.py
```diff
@@ -114,11 +114,13 @@ def box_iou(pred_boxes: torch.Tensor, gt_boxes: torch.Tensor, iscrowd: torch.Boo
 
         if iou_thresholds is None:
             iou_thresholds = torch.linspace(0.5, 0.95, 10, dtype=torch.double)
+        iou_thresholds = torch.as_tensor(iou_thresholds, dtype=torch.double, device=device)
 
         self._iou_thresholds = self._setup_thresholds(iou_thresholds, "iou_thresholds")
 
         if rec_thresholds is None:
             rec_thresholds = torch.linspace(0, 1, 101, dtype=torch.double)
+        rec_thresholds = torch.as_tensor(rec_thresholds, dtype=torch.double, device=device)
```
vfdev-5: This is done below, line 139:
```python
self.rec_thresholds = cast(torch.Tensor, self.rec_thresholds).to(device=device, dtype=precision)
```
so we can remove this line.

And if we fix the device issue in self._setup_thresholds we can even remove the lines 138 and 139.

ignite/metrics/vision/object_detection_average_precision_recall.py
```diff
@@ -114,11 +114,13 @@ def box_iou(pred_boxes: torch.Tensor, gt_boxes: torch.Tensor, iscrowd: torch.Boo
 
         if iou_thresholds is None:
             iou_thresholds = torch.linspace(0.5, 0.95, 10, dtype=torch.double)
+        iou_thresholds = torch.as_tensor(iou_thresholds, dtype=torch.double, device=device)
```
vfdev-5: I think this should be done in `self._setup_thresholds` method instead.

ignite/metrics/vision/object_detection_average_precision_recall.py
obsessor-ak1: Good idea!

ignite/metrics/vision/object_detection_average_precision_recall.py
obsessor-ak1: So, I am going to try it that way...and will also be posting a reproduced error to you... Also, although `_setup_thresholds` doesn't know the device so we need to pass an optional device parameter to it. Will notify after some progress...!

vfdev-5: Let's add device arg to `_BaseAveragePrecision` class

obsessor-ak1: Adding it,  Just updated the PR post on how to reproduce that behavior, check that if it is doing that.

obsessor-ak1: I have made some changes in the `_BaseAveragePrecision` class and `ObjectDetectionAvgPrecisionRecall`. It implements the device moving code in `_setup_thresholds`. Also, I had to use `__` both sides as `self.__device__`, because if I do `self._device`, then in `_setup_thresholds`, it will not be accessible that way. In that case, `self._device` will probably be a `Callable` even if I set it initially to a `torch.device` instance. This behavior is probably attributed to `__getattr__` being overloaded inside the `Metric` class from which `ObjectDetectionAvgPrecisionRecall` is inheriting. If you know a better way, please suggeset...

vfdev-5: I'm not sure why the same code would not work:
```python
class A:
    def __init__(self, a, b, device):
        print("Call A.__init__")
        self.a1 = a
        self.a2 = b
        self.device = device
        print("A: self.device=", self.device)


class B:
    def __init__(self, c, d, device):
        print("Call B.__init__")
        self.b1 = c
        self.b2 = d
        self.device = device
        print("B: self.device=", self.device)


class C(B, A):
    def __init__(self, a, b, c, d, device):
        super().__init__(c, d, device)
        A.__init__(self, a, b, device)

    def test(self):
        print(self.device)


a = C(1, 2, 3, 4, "cpu")
a.test()
```
Can you check ?

obsessor-ak1: Well, I think I made a mistake. I got your point, actually, in the initial code for `ObjectDetectionAvgPrecisionRecall.__init__`, `_setup_thresholds` was called before calling `_BaseAveragePrecision.__init__`. Due to this accessing something like `self.device` or `self._device` inside `_setup_thresholds` would trigger the `__getattr__` overloaded method for the `ObjectDetectionAvgPrecisionRecall` as a fallback which comes from `Metric` class as it is inherited by `ObjectDetectionAvgPrecisionRecall`.
* I initially set, the name of the attribute to `__device__` because for attributes which start and end with `__`, the `__getattr__` overloaded method by `Metric` class doesn't return a `MetricsLambda` instance which is callable but actual attribute with that name.
* Now, I am making call to `_setup_thresholds` after calling `_BaseAveragePrecision.__init__` inside `ObjectDetectionAvgPrecisionRecall.__init__`, this will make sure that `_device` is available and `__getattr__` would not be invoked.
Thanks , for the feedback, please review the new changes (also, sorry for redundant commits, I am a noob in deep learning and this is my first PR😅)

ignite/metrics/vision/object_detection_average_precision_recall.py
```diff
@@ -132,7 +129,9 @@ def box_iou(pred_boxes: torch.Tensor, gt_boxes: torch.Tensor, iscrowd: torch.Boo
         super(Metric, self).__init__(
             rec_thresholds=rec_thresholds,
             class_mean=None,
+            device=device
```
vfdev-5: Can you please also change `super(Metric, self).__init__(`  by explicit call of the base class
```
_BaseAveragePrecision.__init__(self, ...)
```

ignite/metrics/vision/object_detection_average_precision_recall.py
vfdev-5: Thanks for checking that and no worries about your experience, it is just a matter of practice

ignite/metrics/vision/object_detection_average_precision_recall.py
obsessor-ak1: sure if that's appropriate and accepted pattern....
Should I also change this?:
```python
super(ObjectDetectionAvgPrecisionRecall, self).__init__(
    output_transform=output_transform,
    device=device,
    skip_unrolling=skip_unrolling,
)
```
to:
```python
Metric.__init__(self, ...)
```

vfdev-5: No, this line is fine

obsessor-ak1: Did it, just check

vfdev-5:
@obsessor-ak1 thanks for the fix! Can you check existing tests and see why this bug was not picked before? Thanks!
Ideally, if we can add a new test case arg using parameterize to cover this case.

obsessor-ak1:
Okay, I am looking for the test cases...

obsessor-ak1:
I have actually, ran some test cases, here is the output:
```
(ml-env-pt) PS D:\Code\Projects\torch-ignite> pytest -n 2 -m "not distributed" .\tests\ignite\metrics\vision\test_object_detection_map.py
=================================================================== test session starts ===================================================================
platform win32 -- Python 3.12.11, pytest-9.0.2, pluggy-1.6.0
rootdir: D:\Code\Projects\torch-ignite
configfile: pyproject.toml
plugins: anyio-4.9.0, xdist-3.8.0
2 workers [71 items]
....s..s...s.ss..s...s.s..s..s..s..s...............ssssssssssssssss....                                                                              [100%]
==================================================================== warnings summary =====================================================================
tests/ignite/metrics/vision/test_object_detection_map.py: 18 warnings
  D:\Code\Projects\torch-ignite\ignite\distributed\utils.py:396: UserWarning: Using a non-tuple sequence for multidimensional indexing is deprecated and will be changed in pytorch 2.9; use x[tuple(seq)] instead of x[seq]. In pytorch 2.9 this will be interpreted as tensor index, x[torch.tensor(seq)], which will result either in an error or a different result (Triggered internally at C:\actions-runner\_work\pytorch\pytorch\pytorch\torch\csrc\autograd\python_variable_indexing.cpp:312.)
    all_padded_tensors[

tests/ignite/metrics/vision/test_object_detection_map.py: 300 warnings
  D:\Code\Projects\torch-ignite\tests\ignite\metrics\vision\test_object_detection_map.py:712: DeprecationWarning: __array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments. To learn more, see the migration guide https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword
    pyc_iou = pycoco_iou(pred_bbox.numpy(), tgt_bbox.numpy(), iscrowd.int())

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================ 43 passed, 28 skipped, 318 warnings in 115.71s (0:01:55) =================================================
```
I am intentionally not running the test cases related to distributed training because they are causing errors and always failing on my laptop (It's a Windows machine with, just a small RTX 3050 4GB VRAM GPU). 

What exactly should I do now? Do I also need to look at other test files? Should I just start working on designing a test case? Please guide me.

vfdev-5:
We may need to improve this test: https://github.com/pytorch/ignite/blob/e24c991d482e3edcc9080525a50dd7e1d6cd2475/tests/ignite/metrics/vision/test_object_detection_map.py#L718

obsessor-ak1:
I have improved the test case for device checking, but there are two caveats:
* First, I am comparing the type of devices as you could see `assert metric._iou_thresholds.device.type == current_device.type`, because passing a device argument to the tensor makes it more precise by adding index information as `torch.device("cuda", index=0) != torch.device("cuda") # Equality will give False`. It works but I doubt this will cause problems or leak some cases in distributed training.
* Second a lot of other test cases are failed because the tensors they are using as dummies for targets and predictors are always created on CPU by default, and don't take the device passed to `ObjectDetectionAvgPrecisionRecall.__init__` into account. So, their tensors also need to be moved...
Please tell me your thoughts on these, so that I can move further. As always thanks for your feedback and support!

vfdev-5:
The change looks ok to me. Please fix the code styling issues and let's see what CI tells us

tests/ignite/metrics/vision/test_object_detection_map.py
```diff
@@ -1,3 +1,13 @@
+from ignite.utils import manual_seed
```
vfdev-5: @obsessor-ak1 let's revert all unrelated changes in this file

obsessor-ak1:
I have made the changes...in styling....but again here is the output of test runs:
```
(ml-env-pt) PS D:\Code\Projects\torch-ignite> git status
On branch obj_detection_bugfix
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        dummy_testing.py

nothing added to commit but untracked files present (use "git add" to track)
(ml-env-pt) PS D:\Code\Projects\torch-ignite> pytest -n 2 -m "not distributed" .\tests\ignite\metrics\vision\test_object_detection_map.py
=================================================================== test session starts ===================================================================
platform win32 -- Python 3.12.11, pytest-9.0.2, pluggy-1.6.0
rootdir: D:\Code\Projects\torch-ignite
configfile: pyproject.toml
plugins: anyio-4.9.0, xdist-3.8.0
2 workers [71 items]
....s..s..s...s.s.s...s.s...s.sFF.s.FFs..FFFFFFFFFFFssssssssssssssssFF.                                                                              [100%]
======================================================================== FAILURES =========================================================================
__________________________________________________________________ test_compute[sample0] __________________________________________________________________
[gw1] win32 -- Python 3.12.11 D:\Miniconda3\envs\ml-env-pt\python.exe

sample = Sample(data=([{'bbox': tensor([[418., 260., 513., 310.],
        [175., 252., 208., 280.],
        [269., 256., 310., ....44059406, 0.35615276, 0.72425743, 0.4377193 , 0.56096491, 0.5754386 , 0.44      , 0.3547619 , 0.73589744]), length=10)

    def test_compute(sample):
        device = idist.device()

        if device == torch.device("mps"):
            pytest.skip("Due to MPS backend out of memory")

        # AP@.5...95, AP@.5, AP@.75, AP-S, AP-M, AP-L, AR-1, AR-10, AR-100, AR-S, AR-M, AR-L
        ap_50_95_ar_100 = ObjectDetectionAvgPrecisionRecall(
            num_classes=91, device=device)
        ap_50 = ObjectDetectionAvgPrecisionRecall(
            num_classes=91, iou_thresholds=[0.5], device=device)
        ap_75 = ObjectDetectionAvgPrecisionRecall(
            num_classes=91, iou_thresholds=[0.75], device=device)
        ap_ar_S = ObjectDetectionAvgPrecisionRecall(
            num_classes=91, device=device, area_range="small")
        ap_ar_M = ObjectDetectionAvgPrecisionRecall(
            num_classes=91, device=device, area_range="medium")
        ap_ar_L = ObjectDetectionAvgPrecisionRecall(
            num_classes=91, device=device, area_range="large")
        ar_1 = ObjectDetectionAvgPrecisionRecall(
            num_classes=91, device=device, max_detections_per_image_per_class=1)
        ar_10 = ObjectDetectionAvgPrecisionRecall(
            num_classes=91, device=device, max_detections_per_image_per_class=10)

        metrics = [ap_50_95_ar_100, ap_50, ap_75,
                   ap_ar_S, ap_ar_M, ap_ar_L, ar_1, ar_10]
        for metric in metrics:
>           metric.update(sample.data)

tests\ignite\metrics\vision\test_object_detection_map.py:939:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
ignite\metrics\metric.py:888: in wrapper
    func(self, *args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
....
```
These test cases are failing because, they still rely on old behavior of `_iou_thresholds` still being on CPU and not using the passed device, for example, here in this test case:
```
def test_empty_data(available_device):
    """
    Note that PyCOCO returns -1 when threre's no ground truth data.
    """

    metric = ObjectDetectionAvgPrecisionRecall(device=available_device)
    assert metric._device == torch.device(available_device)
    metric.update(
        (
            [{"bbox": torch.zeros((0, 4)), "scores": torch.zeros(
                (0,)), "labels": torch.zeros((0), dtype=torch.long)}],
            [{"bbox": torch.zeros((0, 4)), "iscrowd": torch.zeros(
                (0,)), "labels": torch.zeros((0), dtype=torch.long)}],
        )
    )
    assert len(metric._tps) == 0
    assert len(metric._fps) == 0
    assert metric.compute() == (-1, -1)
    metric.update(
        (
            [
                {
                    "bbox": torch.tensor([[0.0, 0.0, 100.0, 100.0]]),
                    "scores": torch.ones((1,)),
                    "labels": torch.ones((1,), dtype=torch.long),
                }
            ],
            [{"bbox": torch.zeros((0, 4)), "iscrowd": torch.zeros(
                (0,)), "labels": torch.zeros((0), dtype=torch.long)}],
        )
    )
    assert metric.compute() == (-1, -1)

    metric = ObjectDetectionAvgPrecisionRecall(device=available_device)
    assert metric._device == torch.device(available_device)
    metric.update(
        (
            [{"bbox": torch.zeros((0, 4)), "scores": torch.zeros(
                (0,)), "labels": torch.zeros((0), dtype=torch.long)}],
            [
                {
                    "bbox": torch.tensor([[0.0, 0.0, 100.0, 100.0]]),
                    "iscrowd": torch.zeros((1,)),
                    "labels": torch.ones((1,), dtype=torch.long),
                }
            ],
        )
    )
    assert len(metric._tps) == 0
    assert len(metric._fps) == 0
    assert metric._y_true_count[1] == 1
    assert metric.compute() == (0, 0)

    metric = ObjectDetectionAvgPrecisionRecall(device=available_device)
    assert metric._device == torch.device(available_device)
    pred = {
        "bbox": torch.tensor([[0.0, 0.0, 100.0, 100.0]]),
        "scores": torch.tensor([0.9]),
        "labels": torch.tensor([5]),
    }
    target = {"bbox": torch.zeros((0, 4)), "iscrowd": torch.zeros(
        (0,)), "labels": torch.zeros((0), dtype=torch.long)}
    metric.update(([pred], [target]))
    assert len(metric._tps) == len(metric._fps) == 1
    pycoco_result = pycoco_mAP([pred], [target])
    assert metric.compute() == (pycoco_result[0], pycoco_result[8])

```
The tensors for "bbox" and "scores" and "labels" etc. are still not taking device into account they are created on CPU by default and when passed to the `update` method, they produce that same `RuntimeError`. Hence they would fail for "cuda". We need to update them as well...

tests/ignite/metrics/vision/test_object_detection_map.py
obsessor-ak1: Hmm, well I just used the `VSCode` Auto Format Feature which did this...but let me try...

vfdev-5:
OK to me to set explicitly the device in the tests using `available_device` fixture

obsessor-ak1:
Exactly, each tensor in each of the test cases need to use the device specified by `available_device` fixture. We need to add `device=torch.device(available_device)` or  `X = X.to(torch.device(available_device))` to each of the tensors created or brought using other fixtures in these test cases. Should I do that?

vfdev-5:
> Exactly, each tensor in each of the test cases need to use the device specified by `available_device` fixture. We need to add `device=torch.device(available_device)` or `X = X.to(torch.device(available_device))` to each of the tensors created or brought using other fixtures in these test cases. Should I do that?

yes, `t = torch.tensor(..., device=available_device)` would be fine.

obsessor-ak1:
I have updated all the test cases to provide device argument to its tensors. I have also changed `sample` fixture to `get_sample` fixture factory, which sets all the tensors of `data` to the appropriate device after using the utilities from `pycocotools` (which only uses numpy arrays). As always I couldn't run those test cases which are marked "distributed"
Please check the code.

tests/ignite/metrics/vision/test_object_detection_map.py
```diff
@@ -1,7 +1,7 @@
 import sys
 from collections import namedtuple
 from math import ceil
-from typing import Dict, List, Tuple
+from typing import Dict, List, Optional, Tuple
```
vfdev-5: Let's use new typing syntax: `Optional[A] -> A | None`
And also replace Tuple with tuple and remove the import

vfdev-5 (commented):
Thanks for working on this PR!
I left a small comment.
I doubt that code style check CI job would pass (there are unrelated code style changes). Try to use pre-commit to reformat the code.

tests/ignite/metrics/vision/test_object_detection_map.py
obsessor-ak1: Why are unit tests also failing?

vfdev-5: https://github.com/pytorch/ignite/actions/runs/22275461371/job/64436933387?pr=3540#step:10:16286
```
  tests/ignite/metrics/vision/test_object_detection_map.py:641: 
  ignite/metrics/vision/object_detection_average_precision_recall.py:129: in __init__
      _BaseAveragePrecision.__init__(
  ignite/metrics/mean_average_precision.py:61: in __init__
      self.rec_thresholds: Optional[torch.Tensor] = self._setup_thresholds(rec_thresholds, "rec_thresholds")
                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  
  self = <ignite.metrics.vision.object_detection_average_precision_recall.ObjectDetectionAvgPrecisionRecall object at 0x1149d2bd0>
  thresholds = tensor([0.0000, 0.0100, 0.0200, 0.0300, 0.0400, 0.0500, 0.0600, 0.0700, 0.0800, 0.0900, 0.1000, 0.1100, 0.1200, 0.1300, 0.1400, 0.1500, 0.1600, 0.1700, 0.1800, 0.1900, 0.2000, 0.2100, 0.2200, 0.2300,
          0.2400, 0.2500, 0.2600, 0.2700, 0.2800, 0.2900, 0.3000, 0.3100, 0.3200, 0.3300, 0.3400, 0.3500, 0.3600, 0.3700, 0.3800, 0.3900, 0.4000, 0.4100, 0.4200, 0.4300, 0.4400, 0.4500, 0.4600, 0.4700,
          0.4800, 0.4900, 0.5000, 0.5100, 0.5200, 0.5300, 0.5400, 0.5500, 0.5600, 0.5700, 0.5800, 0.5900, 0.6000, 0.6100, 0.6200, 0.6300, 0.6400, 0.6500, 0.6600, 0.6700, 0.6800, 0.6900, 0.7000, 0.7100,
          0.7200, 0.7300, 0.7400, 0.7500, 0.7600, 0.7700, 0.7800, 0.7900, 0.8000, 0.8100, 0.8200, 0.8300, 0.8400, 0.8500, 0.8600, 0.8700, 0.8800, 0.8900, 0.9000, 0.9100, 0.9200, 0.9300, 0.9400, 0.9500,
          0.9600, 0.9700, 0.9800, 0.9900, 1.0000], dtype=torch.float64)
  threshold_type = 'rec_thresholds'
  
      def _setup_thresholds(self, thresholds: Union[Sequence[float], torch.Tensor], threshold_type: str) -> torch.Tensor:
          if isinstance(thresholds, Sequence):
              thresholds = torch.tensor(thresholds, dtype=torch.double)
      
          if isinstance(thresholds, torch.Tensor):
              if thresholds.ndim != 1:
                  raise ValueError(
                      f"{threshold_type} should be a one-dimensional tensor or a sequence of floats"
                      f", given a {thresholds.ndim}-dimensional tensor."
                  )
              thresholds = thresholds.sort().values
          else:
              raise TypeError(f"{threshold_type} should be a sequence of floats or a tensor, given {type(thresholds)}.")
      
          if min(thresholds) < 0 or max(thresholds) > 1:
              raise ValueError(f"{threshold_type} values should be between 0 and 1, given {thresholds}")
  >       return thresholds.to(self._device)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  E       TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead.
```


Keep them on f32

obsessor-ak1: Also, can I remove line 137 in `ignite\metrics\vision\object_detection_average_precision_recall.py` in `ObjectDetectionAvgPrecisionRecall.__init__` class:
```python
self.rec_thresholds = cast(torch.Tensor, self.rec_thresholds).to(device=device, dtype=precision)
```
It is just unnecessary after moving all the logic to `_setup_thresholds`.

obsessor-ak1:
Made some changes, please see...

vfdev-5:
Actually, there is already a dtype precision handling in the code:
```python
precision = torch.double if torch.device(device).type != "mps" else torch.float32
self.rec_thresholds = cast(torch.Tensor, self.rec_thresholds).to(device=device, dtype=precision)
```
There are two things we can do:
1) put precision into setup thresholds method
2) I can run tests on GPU to see if we really need f64 precision
```
# use f32 instead of f64 here:
iou_thresholds = torch.linspace(0, 1, 101, dtype=torch.float32)
rec_thresholds = torch.linspace(0, 1, 101, dtype=torch.float32)
```

tests/ignite/metrics/vision/test_object_detection_map.py
```diff
@@ -1,7 +1,7 @@
 import sys
 from collections import namedtuple
 from math import ceil
-from typing import Dict, List, Tuple
+from typing import Dict, List
```
vfdev-5: Replace here as well Dict -> dict, List -> list and remove unused imports

obsessor-ak1:
> 2\. I can run tests on GPU to see if we really need f64 precision

The recent change I made turns all thresholds passed to `_setup_thresholds` to f64. So, technically both `_iou_thresholds` and `rec_thresholds` have f64 type unless the device is MPS. Is that appropriate?

ignite/metrics/mean_average_precision.py
```diff
@@ -80,8 +82,8 @@ def _setup_thresholds(self, thresholds: Union[Sequence[float], torch.Tensor], th
 
         if min(thresholds) < 0 or max(thresholds) > 1:
             raise ValueError(f"{threshold_type} values should be between 0 and 1, given {thresholds}")
-
-        return thresholds
+        appropriate_type = torch.float32 if self._device.type == "mps" else torch.float64
+        return thresholds.to(self._device, dtype=appropriate_type)
```
vfdev-5:
```suggestion
        precision = torch.float32 if self._device.type == "mps" else torch.float64
        return thresholds.to(self._device, dtype=precision)
```

obsessor-ak1:
There is some problem with the unit tests I think, they are not failing on my laptop.
```
      data_pred = [
  >       {"bbox": d["bbox"].to(device), "scores": d["scores"].to(device), "labels": d["labels"].to(device)}
                   ^^^^^^^^^^^^^^^^^^^^
          for d in data_pred
      ]
  E   TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead.
```
The thing is that "bbox" tensor needs to be already of type "float32" and not "float64":
```
def get_sample(request) -> callable:
    data = coco_val2017_sample() if request.param[0] == "coco2017" else random_sample()
    if request.param[1] == "with_an_empty_pred":
        data[0][1] = {
            "bbox": torch.zeros(0, 4),
            "scores": torch.zeros(
                0,
            ),
            "labels": torch.zeros(0, dtype=torch.long),
        }
        ...
```
`torch.zeros` by default returns a tensor with dtype `float32`, see here:
```
In [8]: x = torch.zeros(0, 4)

In [9]: x.dtype
Out[9]: torch.float32
```
Could you please help?

vfdev-5:
It is set to double here:
https://github.com/obsessor-ak1/torch-ignite/blob/568d37785c50d246b3f674990e613982c877bb59/tests/ignite/metrics/vision/test_object_detection_map.py#L408-L410

ignite/metrics/vision/object_detection_average_precision_recall.py
```diff
@@ -11,10 +11,10 @@
 
 def coco_tensor_list_to_dict_list(
     output: Tuple[
-        Union[List[torch.Tensor], List[Dict[str, torch.Tensor]]],
-        Union[List[torch.Tensor], List[Dict[str, torch.Tensor]]],
+        Union[list[torch.Tensor], list[dict[str, torch.Tensor]]],
+        Union[list[torch.Tensor], list[dict[str, torch.Tensor]]],
     ]
-) -> Tuple[List[Dict[str, torch.Tensor]], List[Dict[str, torch.Tensor]]]:
+) -> Tuple[list[dict[str, torch.Tensor]], list[dict[str, torch.Tensor]]]:
```
vfdev-5: Ideally, let's replace also `Tuple` with `tuple` and `Union[A, B] -> A | B `

obsessor-ak1:
I have just fixed the `test_iou` test case to use `np.allclose` instead of absolute equality. Also, I have fixed the suggested type hint everywhere. Please check that.

vfdev-5:
@obsessor-ak1 thanks for working on this! This bug fix seems a bit tricky, but looks like we are on the good track.
Please fix the conflict with master branch

obsessor-ak1:
I have resolved some merge conflicts. Check if that would work. Also, can we remove the occurrences of `cast` function. As, it seems as if it serves no purpose, given the current type annotations....

vfdev-5:
> I have resolved some merge conflicts. Check if that would work. Also, can we remove the occurrences of `cast` function. As, it seems as if it serves no purpose, given the current type annotations....

Yes, sure we can remote them if they are useless. Maybe in another PR to avoid to much changes in this one.

CI for macosx is still failing.

obsessor-ak1:
I have fixed the test but before running CI, please look for other places where assertion may fail due to precision errors. And should we change all of that? Since come previous changes many parts of code are using f32 instead of f64. Also, I don't know from where did this `KeyboardInterrupt` came up in other tests.

vfdev-5:
I'll take a look. Thanks!

tests/ignite/metrics/vision/test_object_detection_map.py
```diff
@@ -607,7 +608,21 @@ def sample(request) -> Sample:
...
+        if device is None:
+            device = torch.device("cpu")
+        data_pred = [
+            {"bbox": d["bbox"].to(device), "scores": d["scores"].to(device), "labels": d["labels"].to(device)}
+            for d in data_pred
+        ]
+        data_true = [
+            {"bbox": d["bbox"].to(device), "labels": d["labels"].to(device), "iscrowd": d["iscrowd"].to(device)}
+            for d in data_true
+        ]
```
vfdev-5: @obsessor-ak1 Instead of manually recreating data_pred and data_true we can try using ignite.utils.convert_tensor

obsessor-ak1: I will do it but looking at the definition of that function it seems like that would have almost no practical effect. Why is that necessary? Why are MPS tests still failing? Is that memory error a random chance?

obsessor-ak1:
Please check...I have made some changes.

obsessor-ak1:
@vfdev-5, any updates?

vfdev-5:
@obsessor-ak1 can you first please fix the merge conflict. See on GH

obsessor-ak1:
@vfdev-5 Conflicts resolved and merged! Please review.

vfdev-5:
@obsessor-ak1 thanks again for your work on this PR! I appreciate that you reported it and tried your best to fix it! We had to move forward and make other changes in the file, especially that we need to make a new release, so it was done in #3667 PR. Thanks for your understanding and hope to see you around contributing or reporting other issues!

Issue #3547, opened 2026-02-19
Modernize `super()` calls


omkar-334:
Should we replace super(ClassName, self) with super() across the codebase?  
There is no need to pass the class and instance explicitly.  

Example - `super(DummyModel, self).__init__()` becomes `super().__init__()`  

There are over 130 occurrences of this, which can be simplified

vfdev-5:
Sounds good! if you would like to send a PR with that...

omkar-334:
will do, thanks!

omkar-334:
hi @vfdev-5 , since you specified the 1 file per PR policy, how should i frame this fix?  Is it fine if i make PR per directory - like `examples`, `tests`, `ignite/metrics`, `ignite/handlers`?

vfdev-5:
That policy has specifically for that issue under the question. Your suggestion with changes per directory is fine.

aaishwarymishra:
currently 'examples/notebooks` is remaining pr is welcomed :)

nightcityblade:
Hi, I noticed there are still ~15 remaining old-style super() calls across ignite/, tests/ that were missed or re-introduced after the earlier PRs. I'll submit a PR to clean these up.

Note: I've preserved the intentional grandparent super() calls in metric.py and mean_average_precision.py (e.g. `super(BatchWise, self).__init__`) as they deliberately skip intermediate classes in the MRO.

PR #3553, opened 2026-02-19
refactor `super()` in `examples/notebooks`

omkar-334:
Part of #3547


omkar-334:
I think I should revert all the notebook changes which happened

aaishwarymishra:
I think all the diffs happening is because you are using different editor, this notebook was made in colab and when you saved it in different jupyter environment it formatted differently.

1 work around could be you could upload notebook in colab make changes and re-download it.

@vfdev-5 is this diff fine?

I think we should add a pre-commit for the jupyter notebooks too to avoid these huge diffs.

omkar-334:
closed in favor of #3577

PR #3558, opened 2026-02-20
replace typing hints in `ignite/distributed/auto.py`

omkar-334:
Part of #3481


omkar-334:
nice, tests are passing. I think the cases where a class is used as a type hint string, we need to use `from __future__ import annotations` and convert the string to the class. 

For example, 
```
"_MpDeviceLoader"
```
becomes
```
_MpDeviceLoader
```

vfdev-5:
I'm also ok with "_MpDeviceLoader". Which python version it starts working with `_MpDeviceLoader` without quotes?

PR #3561, opened 2026-02-20
replace typing hints in `ignite/distributed/comp_models/__init__.py`

omkar-334:
Part of #3481

### DIFF
--- a/ignite/distributed/comp_models/__init__.py
+++ b/ignite/distributed/comp_models/__init__.py
@@ -1,4 +1,6 @@
-from typing import Type, TYPE_CHECKING
+from __future__ import annotations
+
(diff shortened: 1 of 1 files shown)


ignite/distributed/comp_models/__init__.py
```diff
@@ -1,4 +1,6 @@
-from typing import List, Tuple, Type, TYPE_CHECKING, Union
+from __future__ import annotations
```
vfdev-5, line 1: I think we do not need this. Let's remove it

omkar-334: check this [comment](https://github.com/pytorch/ignite/pull/3562/#discussion_r2844803932)

PR #3562, opened 2026-02-20
replace typing hints in `ignite/distributed/comp_models/base.py`

omkar-334:
Part of #3481

### DIFF
--- a/ignite/distributed/comp_models/base.py
+++ b/ignite/distributed/comp_models/base.py
@@ -311,7 +311,7 @@ class _SerialModel(ComputationModel):
     name = "serial"
     available_backends = ()
 
-    def __init__(self, _backend: Optional[str] = None, **kwargs: Any) -> None:
+    def __init__(self, _backend: str | None = None, **kwargs: Any) -> None:
         super().__init__()
 
     def get_local_rank(self) -> int:


ignite/distributed/comp_models/base.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
vfdev-5, line 1: Why do we need this?

vfdev-5: @omkar-334 can you check this comment please

omkar-334: hey @vfdev-5 , sorry for the delay.  
with this import, we can use a class inside its own type hints or use a class that is defined later in the file without getting a NameError. 

example (https://stackoverflow.com/a/61544901) - 
this fails (since annotations are processed at runtime) - 
```
class A:
    def f(self) -> A: # NameError: name 'A' is not defined
        pass
```
but this passes (annotations are deferred) - 
```
from __future__ import annotations

class A:
    def f(self) -> A:
        pass
```

omkar-334: I thought this is cleaner than using strings, so I added this.

omkar-334: what do you think? based on our decision, i can make the change to the other PRs where i've used this import

omkar-334: from python 3.14+, this will become the default behaviour of all imports - deferred evaluation and lazy imports.

vfdev-5: Let's use import as you coded it, thanks!

vfdev-5:
@omkar-334 please check the failing CI jobs

omkar-334:
i think it should pass now

Issue #3535, opened 2026-02-14
[Feature] Add reset_on_start to EarlyStopping


TahaZahid05:
**Motivation**

Currently, the `EarlyStopping` handler maintains its internal state (`counter` and `best_score`) across multiple calls of `engine.run()`. While this is ideal for resuming training from checkpoints, it creates a friction point for users in interactive environments (e.g., Jupyter Notebooks or Google Colab).

In these settings, a user might:

1. Run a training session that triggers early stopping.
2. Adjust a hyperparameter in the same notebook.
3. Call trainer.run() again using the same engine/handler instance.

Because the state is preserved, the engine immediately terminates because the `counter` is already at the limit. The user must manually re-instantiate the handler or manually reset the attributes, which is not intuitive for experiment-heavy workflows.

**Proposed Solution**

I propose adding an optional boolean parameter `reset_on_start` (defaulting to `False`) to the `EarlyStopping` constructor.

If set to `True`, the handler will attach a listener to `Events.STARTED` on the provided `trainer`. This listener calls a internal `reset()` method to clear `self.counter` and `self.best_score` at the start of every run.

The feature is also backwards compatible as the default is set to `False` which preserve existing behavior for distributed/long-running jobs where state must persist across crashes/resumes.

**Example**

```python
# The handler will now automatically "refresh" every time trainer.run() is called.
handler = EarlyStopping(
    patience=10, 
    score_function=score_function, 
    trainer=trainer, 
    reset_on_start=True
)

evaluator.add_event_handler(Events.COMPLETED, handler)
```

vfdev-5:
Yes, this suggestion makes sense!
I remember there was a discussion about that, I do not quite remember what we decided...
There are other similar handlers that require a reset between runs from start. Ideally we should update the API the same way for all of them

TahaZahid05:
Standardizing this across similar handlers would definitely improve the functionality for notebook users.

I can start by implementing it for `EarlyStopping` as a reference for implementation details. Then I can open a follow-up PR to bring the same functionality to other handlers.

Is this fine or would you prefer the bulk way of adding it to different handlers in the same PR?

vfdev-5:
First, we need to decide the API for all stateful handlers before implementing

aaishwarymishra:
I don't think adding the parameter is necessary as on rerun of trainer.run()  early stopping should start fresh by default. So resetting the internal variables of handlers should be default behaviour.

TahaZahid05:
[...]

PR #3564, opened 2026-02-20
Introduce `ResettableHandler` interface and `attach()` method

TahaZahid05:
Partially Addresses #3535 

**Description:**
Introduces a `ResettableHandler` abstract base class to establish a formalized interface for stateful handlers. `EarlyStopping` is the first adopter of this pattern, implementing both `reset()` and `attach()` methods. 

This mirrors the existing `Metric.attach()` lifecycle pattern, avoiding any Engine-level modifications while providing a clean, self-contained way to manage handler state across runs.

**API**

```python
from ignite.handlers import EarlyStopping

trainer = Engine(train_fn)
evaluator = Engine(eval_fn)

handler = EarlyStopping(patience=5, score_function=score_fn, trainer=trainer)

# Attaches the handler's check to the evaluator (Events.EPOCH_COMPLETED)
# AND automatically attaches the handler's `reset()` to the trainer (Events.STARTED)
handler.attach(
    engine=evaluator, 
    event=Events.EPOCH_COMPLETED, 
    reset_engine=trainer, 
    reset_event=Events.STARTED
)
```

Check list:

### DIFF
--- a/ignite/base/mixins.py
+++ b/ignite/base/mixins.py
@@ -1,5 +1,30 @@
+from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
 from collections.abc import Mapping
+from typing import Any, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from ignite.engine import Engine
+
+
+class ResettableHandler(metaclass=ABCMeta):
+    """Interface for handlers whose internal state can be reset.
+
+    Subclasses must implement the :meth:`reset` method to clear any accumulated
+    state, typically at the beginning of a training run.
+
+    .. versionadded:: 0.6.0
+    """
+
+    @abstractmethod
+    def reset(self) -> None:
+        """Reset the handler's internal state."""
+        pass
+
+    @abstractmethod
+    def attach(self, engine: "Engine", *args: Any, **kwargs: Any) -> None:
+        """Attach the handler to an engine."""
+        pass
 
 
 class Serializable:
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -1,14 +1,14 @@
 from collections import OrderedDict
-from typing import Callable, cast, Mapping, Literal
+from typing import Any, Callable, cast, Mapping, Literal
 
-from ignite.base import Serializable
-from ignite.engine import Engine
+from ignite.base import Serializable, ResettableHandler
+from ignite.engine import Engine, Events
 from ignite.utils import setup_logger
 
 __all__ = ["EarlyStopping"]
 
 
-class EarlyStopping(Serializable):
+class EarlyStopping(Serializable, ResettableHandler):
     """EarlyStopping handler can be used to stop the training if no improvement after a given number of events.
 
     Args:
@@ -127,6 +127,45 @@ class EarlyStopping(Serializable):
         else:
             self.best_score = score
             self.counter = 0
+
+    def reset(self) -> None:
+        """Reset the early stopping state, including the counter and best score.
+
+        .. versionadded:: 0.6.0
+        """
+        self.counter = 0
+        self.best_score = None
+
+    def attach(  # type: ignore[override]
+        self,
+        engine: Engine,
+        event: Any = Events.COMPLETED,
+        reset_engine: Engine | None = None,
+        reset_event: Any = Events.STARTED,
+        *args: Any,
+        **kwargs: Any,
+    ) -> None:
+        """Attaches the early stopping handler to an engine and registers its reset callback.
+
+        This method will:
+        1. Add the early stopping evaluation logic (``self``) to ``engine`` on the given ``event``.
+        2. Add the ``reset`` method to ``reset_engine`` (or ``engine`` if not provided) on the given ``reset_event``.
+
+        Args:
+            engine: The engine to attach the early stopping evaluation to (typically an evaluator).
+            event: The event on ``engine`` that triggers the early stopping check. Default is
+                :attr:`~ignite.engine.events.Events.COMPLETED`.
+            reset_engine: The engine to attach the reset callback to (typically the trainer).
+                If ``None``, defaults to ``engine``.
+            reset_event: The event on ``reset_engine`` that triggers the handler state reset.
+                Default is :attr:`~ignite.engine.events.Events.STARTED`.
+
+        .. versionadded:: 0.6.0
+        """
+        engine.add_event_handler(event, self)
+
+        target_reset_engine = reset_engine or engine
+        target_reset_engine.add_event_handler(reset_event, self.reset)
 
     def state_dict(self) -> "OrderedDict[str, float]":
         """Method returns state dict with ``counter`` and ``best_score``.
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -34,6 +34,17 @@ Complete list of generic handlers
     param_scheduler.ParamScheduler
     state_param_scheduler.StateParamScheduler
 
+Interfaces
+----------
+
+.. currentmodule:: ignite.base.mixins
+
+.. autosummary::
+    :nosignatures:
+    :toctree: generated
+
+    ResettableHandler
+
 
 Loggers
 --------
--- a/ignite/base/__init__.py
+++ b/ignite/base/__init__.py
@@ -1,1 +1,1 @@
-from ignite.base.mixins import Serializable
+from ignite.base.mixins import Serializable, ResettableHandler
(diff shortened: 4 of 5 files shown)


TahaZahid05:
@vfdev-5 I have refactored the PR to use the `Metric.attach()` pattern as we discussed.

**Key changes:**
- Reverted all modifications to `Engine` (`engine.py` and `events.py` are now untouched).
- Added the `ResettableHandler` ABC in `ignite/base/mixins.py` with an abstract `reset()` method to establish the formalized interface.
- Refactored `EarlyStopping` to inherit from the ABC, implementing `reset()` and `attach()`. 
- Added test coverage in `test_early_stopping.py` verifying that `attach()` registers both the evaluation event and the reset callback (including cross-engine scenarios).
- Sphinx documentation configuration updated to recognize the new ABC.

Let me know what you think!

**Future Direction:**
If you are happy with how this pattern turned out for `EarlyStopping`, I'd be more than happy to help migrate the other stateful handlers in follow-up PRs. Just let me know if that's something you'd want!

vfdev-5:
@TahaZahid05 can you please also update the PR description. Thanks a lot for working on this topic!

ignite/base/mixins.py
```diff
@@ -1,7 +1,21 @@
...
 
+class ResettableHandler(metaclass=ABCMeta):
+    """Interface for handlers whose internal state can be reset.
+
+    Subclasses must implement the :meth:`reset` method to clear any accumulated
+    state, typically at the beginning of a training run.
+    """
+
+    @abstractmethod
+    def reset(self) -> None:
```
vfdev-5, line 20: I wonder whether this interface should also have `attach` method? 
Should it be abstractclass or a [Protocol](https://typing.python.org/en/latest/spec/protocol.html)?

ignite/handlers/early_stopping.py
```diff
@@ -128,6 +128,38 @@ def __call__(self, engine: Engine) -> None:
...
+        """Reset the early stopping state, including the counter and best score."""
+        self.counter = 0
+        self.best_score = None
+
+    def attach(
+        self,
+        engine: Engine,
+        event: Any = Events.COMPLETED,
+        reset_engine: Engine | None = None,
+        reset_event: Any = Events.STARTED,
```
vfdev-5, line 144: Here we can also think about HandlerUsage similar to [MetricUsage](https://github.com/pytorch/ignite/blob/18c719344f0a4b9ce7198079803cecebd48d009b/ignite/metrics/metric.py#L30) and maybe even unify both into a single class...

Also, we can promote `attach` method as alternative to direct adding the handler to the engine in the sense that, it still works like `trainer.add_event_handler(event, handler)`, but if we care about handler's state lifecycle during the training we can use `attach` method. So attach does the reset by default according to the default usage.

ignite/base/mixins.py
TahaZahid05: I considered adding an `attach` method to the interface, but the method signature differs significantly across handlers. To enforce a universal signature on the abstract class would force us to use `def attach(self, engine: Engine, *args: Any, **kwargs: Any) -> None:` which may be less effective as it removes the type-hinting/auto-complete for the users. Do you think we should still add it?

Regarding using `Protocol` or `ABCMeta`, I chose `ABCMeta` because all major interfaces (`Metric`, `Checkpoint`, etc) in the repository are using it so I wanted to align the function with the rest of the repository. I didn't see `Protocol` being used anywhere in the codebase.

ignite/handlers/early_stopping.py
TahaZahid05: Yeah, that would be great. I can work on that. Creating a unifying single class like `Usage` (open to suggestions on the name)  would require modifying base files, changing docs, and imports across the codebase. Would you prefer I scope that unificiation into a dedicated follow-up PR, or should I attempt to refactor `MetricUsage` alongside `EarlyStopping` within this PR?

TahaZahid05: Hey @vfdev-5, following up on your feedback about generalizing `MetricUsage` to a shared `Usage`

I've opened a dependent PR here: #3579 

Since the new PR relies on the `attach()` logic we introduced in this one, we can review and merge this PR first. Once merged, I will rebase the new PR so it merges cleanly. Let me know what you think!

ignite/base/mixins.py
vfdev-5: Even if `def attach(self, engine: Engine, *args: Any, **kwargs: Any) -> None:` is less informative, it is still an interface asking to override `attach` method = essential thing we want to do. What do you think?

ignite/base/mixins.py
```diff
@@ -1,7 +1,21 @@
...
 from collections import OrderedDict
 from collections.abc import Mapping
 
 
+class ResettableHandler(metaclass=ABCMeta):
+    """Interface for handlers whose internal state can be reset.
+
+    Subclasses must implement the :meth:`reset` method to clear any accumulated
+    state, typically at the beginning of a training run.
+    """
```
vfdev-5, line 17: let's put versionadded tag in the docs. Please take a look at the contribution.md guidelines, docs writing part. I think this class should be also exposed in the docs for the users. let's make sure we can see it in the docs.

ignite/handlers/early_stopping.py
vfdev-5: Sounds good to me

ignite/base/mixins.py
TahaZahid05: Apologies, I'll make sure to be mindful of that from next time.

vfdev-5: No worries, it happens rather frequently to everyone

ignite/base/mixins.py
TahaZahid05: Yeah, makes sense. I'll change it.

ignite/base/mixins.py
TahaZahid05: @vfdev-5 i have added versionadded tag

vfdev-5 (approved):
LGTM, thanks @TahaZahid05 !

TahaZahid05:
@vfdev-5 can you guide me regarding the errors? I built it locally and ran fine, and here it seems to be some error regarding `Sampler`, `Dataset`, `DataLoader` but i didn't touch those.

vfdev-5:
"Build docs" errors are failing unrelated. Pyrefly are related to the change but we should avoid circular import of Engine

TahaZahid05:
@vfdev-5 

Thanks. Should I just use a `TYPE_CHECKING` block to import `Engine` and then use a string i.e `engine: "Engine"` in the attach method?

vfdev-5:
> @vfdev-5
> 
> Thanks. Should I just use a `TYPE_CHECKING` block to import `Engine` and then use a string i.e `engine: "Engine"` in the attach method?

Try that and see if it helps

TahaZahid05:
> > @vfdev-5
> > Thanks. Should I just use a `TYPE_CHECKING` block to import `Engine` and then use a string i.e `engine: "Engine"` in the attach method?
> 
> Try that and see if it helps

Done. Ran locally as well and it passed. I have also retained type hints for `attach()` by using `# type: ignore[override]` as i saw it had been implemented in other functions such as for `ProgressBar` in  `tqdm_logger.py` as well.

Thanks.

ignite/handlers/early_stopping.py
```diff
@@ -128,6 +128,45 @@ def __call__(self, engine: Engine) -> None:
...
 
+    def reset(self) -> None:
+        """Reset the early stopping state, including the counter and best score.
+
+        .. versionadded:: 0.6.0
+        """
+        self.counter = 0
+        self.best_score = None
+
+    def attach(  # type: ignore[override]
```
vfdev-5, line 139: pyrefly fails without `ignore[override]` ?

TahaZahid05: @vfdev-5 yes, if ran without `ignore[override]` it gives:

```
ERROR Class member `EarlyStopping.attach` overrides parent class `ResettableHandler` in an inconsistent manner [bad-override]
   --> ignite/handlers/early_stopping.py:139:9
    |
139 |     def attach(  
    |         ^^^^^^
    |
  `EarlyStopping.attach` has type `BoundMethod[EarlyStopping, (self: EarlyStopping, engine: Engine, event: Any = Events.COMPLETED, reset_engine: Engine | None = None, reset_event: Any = Events.STARTED, *args: Any, **kwargs: Any) -> None]`, which is not assignable to `BoundMethod[EarlyStopping, (self: EarlyStopping, engine: Engine, *args: Any, **kwargs: Any) -> None]`, the type of `ResettableHandler.attach`
 INFO 1 error (106 suppressed)
```

PR #3574, opened 2026-02-22
Implement Ray Tune example with Ignite

### LINKED ISSUE #1735: Port Ray example to Ignite
https://docs.ray.io/en/master/tune/tutorials/tune-pytorch-cifar.html

atharva7-g:
Fixes #1735.

Description:

Added an example to run Ray Tune with Ignite. Uses an `ASHAScheduler` to train a simple `CNN` on the CIFAR10 dataset as in the original example: https://docs.ray.io/en/master/tune/examples/tune-pytorch-cifar.html. Uses Ignite's inbuilt: `create_supervised_trainer` and `create_supervised_evaluator`.

Please let me know if this looks OK. 

Thanks!

Check list:

### DIFF
--- a/ignite/metrics/hsic.py
+++ b/ignite/metrics/hsic.py
@@ -1,4 +1,4 @@
-from typing import Callable, Sequence, Union
+from collections.abc import Callable, Sequence
 
 import torch
 from torch import Tensor
@@ -98,7 +98,7 @@ class HSIC(Metric):
         sigma_y: float = -1,
         ignore_invalid_batch: bool = True,
         output_transform: Callable = lambda x: x,
-        device: Union[str, torch.device] = torch.device("cpu"),
+        device: str | torch.device = torch.device("cpu"),
         skip_unrolling: bool = False,
     ):
         super().__init__(output_transform, device, skip_unrolling=skip_unrolling)
@@ -132,7 +132,7 @@ class HSIC(Metric):
         rx = xx.diag().unsqueeze(0).expand_as(xx)
         dxx = rx.T + rx - xx * 2
 
-        vx: Union[Tensor, float]
+        vx: Tensor | float
         if self.sigma_x < 0:
             # vx = torch.quantile(dxx, 0.5)
             vx = torch.quantile(dxx, 0.5)
@@ -144,7 +144,7 @@ class HSIC(Metric):
         ry = yy.diag().unsqueeze(0).expand_as(yy)
         dyy = ry.T + ry - yy * 2
 
-        vy: Union[Tensor, float]
+        vy: Tensor | float
         if self.sigma_y < 0:
             vy = torch.quantile(dyy, 0.5)
         else:
--- a/ignite/metrics/matthews_corrcoef.py
+++ b/ignite/metrics/matthews_corrcoef.py
@@ -1,4 +1,4 @@
-from typing import Callable, Union
+from collections.abc import Callable
 
 import torch
 
@@ -80,7 +80,7 @@ class MatthewsCorrCoef(EpochMetric):
         self,
         output_transform: Callable = lambda x: x,
         check_compute_fn: bool = False,
-        device: Union[str, torch.device] = torch.device("cpu"),
+        device: str | torch.device = torch.device("cpu"),
         skip_unrolling: bool = False,
     ):
         try:
--- a/ignite/metrics/metric.py
+++ b/ignite/metrics/metric.py
@@ -1,9 +1,9 @@
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
-from collections.abc import Mapping
+from collections.abc import Callable, Mapping, Sequence
 from functools import wraps
 from numbers import Number
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Literal, Optional, Sequence, Tuple, Union, cast
+from typing import TYPE_CHECKING, Any, Literal, cast
 
 import torch
 
@@ -356,14 +356,14 @@ class Metric(Serializable, metaclass=ABCMeta):
     """
 
     # public class attribute
-    required_output_keys: Optional[Tuple] = ("y_pred", "y")
+    required_output_keys: tuple | None = ("y_pred", "y")
     # for backward compatibility
     _required_output_keys = required_output_keys
 
     def __init__(
         self,
         output_transform: Callable = lambda x: x,
-        device: Union[str, torch.device] = torch.device("cpu"),
+        device: str | torch.device = torch.device("cpu"),
         skip_unrolling: bool = False,
         metrics_result_mode: Literal["flatten", "named", "both"] = "both",
     ):
@@ -525,7 +525,7 @@ class Metric(Serializable, metaclass=ABCMeta):
 
             engine.state.metrics[name] = result
 
-    def _check_usage(self, usage: Union[str, MetricUsage]) -> MetricUsage:
+    def _check_usage(self, usage: str | MetricUsage) -> MetricUsage:
         if isinstance(usage, str):
             usages = [EpochWise, RunningEpochWise, BatchWise, RunningBatchWise, SingleEpochRunningBatchWise]
             for usage_cls in usages:
@@ -541,7 +541,7 @@ class Metric(Serializable, metaclass=ABCMeta):
             raise TypeError(f"Unhandled usage type {type(usage)}")
         return usage
 
-    def attach(self, engine: Engine, name: str, usage: Union[str, MetricUsage] = EpochWise()) -> None:
+    def attach(self, engine: Engine, name: str, usage: str | MetricUsage = EpochWise()) -> None:
         """
         Attaches current metric to provided engine. On the end of engine's run, `engine.state.metrics` dictionary will
         contain computed metric's value under provided name.
@@ -582,7 +582,7 @@ class Metric(Serializable, metaclass=ABCMeta):
             engine.add_event_handler(usage.ITERATION_COMPLETED, self.iteration_completed)
         engine.add_event_handler(usage.COMPLETED, self.completed, name)
 
-    def detach(self, engine: Engine, usage: Union[str, MetricUsage] = EpochWise()) -> None:
+    def detach(self, engine: Engine, usage: str | MetricUsage = EpochWise()) -> None:
         """
         Detaches current metric from the engine and no metric's computation is done during the run.
         This method in conjunction with :meth:`~ignite.metrics.metric.Metric.attach` can be useful if several
@@ -625,7 +625,7 @@ class Metric(Serializable, metaclass=ABCMeta):
         if engine.has_event_handler(self.iteration_completed, usage.ITERATION_COMPLETED):
             engine.remove_event_handler(self.iteration_completed, usage.ITERATION_COMPLETED)
 
-    def is_attached(self, engine: Engine, usage: Union[str, MetricUsage] = EpochWise()) -> bool:
+    def is_attached(self, engine: Engine, usage: str | MetricUsage = EpochWise()) -> bool:
         """
         Checks if current metric is attached to provided engine. If attached, metric's computed
         value is written to `engine.state.metrics` dictionary.
@@ -639,9 +639,7 @@ class Metric(Serializable, metaclass=ABCMeta):
         return engine.has_event_handler(self.completed, usage.COMPLETED)
 
     def _state_dict_per_rank(self) -> OrderedDict:
-        def func(
-            x: Union[torch.Tensor, Metric, None, float], **kwargs: Any
-        ) -> Union[torch.Tensor, float, OrderedDict, None]:
+        def func(x: torch.Tensor | Metric | None | float, **kwargs: Any) -> torch.Tensor | float | OrderedDict | None:
             if isinstance(x, Metric):
                 return x._state_dict_per_rank()
             if x is None or isinstance(x, (int, float, torch.Tensor)):
@@ -652,7 +650,7 @@ class Metric(Serializable, metaclass=ABCMeta):
                     " numeric types, tensor, Metric or sequence/mapping of metrics."
                 )
 
-        state: OrderedDict[str, Union[torch.Tensor, List, Dict, None]] = OrderedDict()
+        state: OrderedDict[str, torch.Tensor | list | dict | None] = OrderedDict()
         for attr_name in self._state_dict_all_req_keys:
             if attr_name not in self.__dict__:
                 raise ValueError(
@@ -808,10 +806,10 @@ class Metric(Serializable, metaclass=ABCMeta):
 
         return MetricsLambda(lambda x: x[index], self)
 
-    def __getstate__(self) -> Dict:
+    def __getstate__(self) -> dict:
         return self.__dict__
 
-    def __setstate__(self, d: Dict) -> None:
+    def __setstate__(self, d: dict) -> None:
         self.__dict__.update(d)
 
 
@@ -893,11 +891,11 @@ def reinit__is_reduced(func: Callable) -> Callable:
     return wrapper
 
 
-def _is_list_of_tensors_or_numbers(x: Sequence[Union[torch.Tensor, float]]) -> bool:
+def _is_list_of_tensors_or_numbers(x: Sequence[torch.Tensor | float]) -> bool:
     return isinstance(x, Sequence) and all([isinstance(t, (torch.Tensor, Number)) for t in x])
 
 
-def _to_batched_tensor(x: Union[torch.Tensor, Number], device: Optional[torch.device] = None) -> torch.Tensor:
+def _to_batched_tensor(x: torch.Tensor | Number, device: torch.device | None = None) -> torch.Tensor:
     if isinstance(x, torch.Tensor):
         return x.unsqueeze(dim=0)
     return torch.tensor([x], device=device)
--- a/ignite/metrics/precision_recall_curve.py
+++ b/ignite/metrics/precision_recall_curve.py
@@ -1,4 +1,5 @@
-from typing import Any, Callable, cast, Tuple, Union
+from collections.abc import Callable
+from typing import Any, cast
 
 import torch
 
@@ -7,7 +8,7 @@ from ignite.exceptions import NotComputableError
 from ignite.metrics.epoch_metric import EpochMetric
 
 
-def precision_recall_curve_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> Tuple[Any, Any, Any]:
+def precision_recall_curve_compute_fn(y_preds: torch.Tensor, y_targets: torch.Tensor) -> tuple[Any, Any, Any]:
     from sklearn.metrics import precision_recall_curve
 
     y_true = y_targets.cpu().numpy()
@@ -77,7 +78,7 @@ class PrecisionRecallCurve(EpochMetric):
         self,
         output_transform: Callable = lambda x: x,
         check_compute_fn: bool = False,
-        device: Union[str, torch.device] = torch.device("cpu"),
+        device: str | torch.device = torch.device("cpu"),
         skip_unrolling: bool = False,
     ) -> None:
         try:
@@ -93,7 +94,7 @@ class PrecisionRecallCurve(EpochMetric):
             skip_unrolling=skip_unrolling,
         )
 
-    def compute(self) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:  # type: ignore[override]
+    def compute(self) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:  # type: ignore[override]
         if len(self._predictions) < 1 or len(self._targets) < 1:
             raise NotComputableError("PrecisionRecallCurve must have at least one example before it can be computed.")
 
@@ -109,7 +110,7 @@ class PrecisionRecallCurve(EpochMetric):
 
             if idist.get_rank() == 0:
                 # Run compute_fn on zero rank only
-                precision, recall, thresholds = cast(Tuple, self.compute_fn(_prediction_tensor, _target_tensor))
+                precision, recall, thresholds = cast(tuple, self.compute_fn(_prediction_tensor, _target_tensor))
                 precision = torch.tensor(precision, device=_prediction_tensor.device, dtype=self._double_dtype)
                 recall = torch.tensor(recall, device=_prediction_tensor.device, dtype=self._double_dtype)
                 # thresholds can have negative strides, not compatible with torch tensors
@@ -126,4 +127,4 @@ class PrecisionRecallCurve(EpochMetric):
 
             self._result = (precision, recall, thresholds)  # type: ignore[assignment]
 
-        return cast(Tuple[torch.Tensor, torch.Tensor, torch.Tensor], self._result)
+        return cast(tuple[torch.Tensor, torch.Tensor, torch.Tensor], self._result)


examples/cifar10raytune/requirements.txt
```diff
@@ -0,0 +1,4 @@
+ray[tune]
+torch
+torchvision
+ipywidgets
```
rwtarpit: we should add `pytorch-ignite` to requirements.txt.

examples/cifar10raytune/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    val_size = len(train_set) - train_size
+
+    train_subset, val_subset = torch.utils.data.random_split(train_set, [train_size, val_size])
+
+    train_loader = torch.utils.data.DataLoader(
+        train_subset, batch_size=batch_size, shuffle=True, num_workers=num_workers
+    )
+    val_loader = torch.utils.data.DataLoader(val_subset, batch_size=batch_size, shuffle=True, num_workers=num_workers)
+
+    return train_loader, val_loader
```
rwtarpit: `shuffle` should be `False` for val_loader just like `test_loader`

examples/cifar10raytune/main.py
```diff
@@ -0,0 +1,191 @@
...
+    if checkpoint_dir:
+        checkpoint_path = os.path.join(checkpoint_dir, "checkpoint.pt")
+        if os.path.exists(checkpoint_path):
+            checkpoint_state = torch.load(checkpoint_path)
+            start_epoch = checkpoint_state["epoch"]
+            net.load_state_dict(checkpoint_state["net_state_dict"])
+            optimizer.load_state_dict(checkpoint_state["optimizer_state_dict"])
+
+    train_loader, val_loader = get_data_loaders(int(config["batch_size"]), data_dir, num_workers)
+
```
rwtarpit: explicit cast to `int` seems unnecessary since we already have defined `batch_size` with `int` in config

examples/cifar10raytune/README.md
```diff
@@ -0,0 +1,85 @@
...
+- `lr`: Learning rate for SGD optimizer
+- `batch_size`: Training batch size
+
+## Output
+
+```
+Best trial config: {'l1': 128, 'l2': 64, 'lr': 0.0012, 'batch_size': 8, 'device': 'cuda'}
+Best trial final validation loss: 1.2345
+Best trial final validation accuracy: 0.5842
+Best trial test set accuracy: 0.5842
```
rwtarpit: results seems dummy data here.
i wonder if we could get a similar tabular representation as reference showing bad runs getting dropped early on. this could show ray-tune's compatibility with ignite.
what do you think?

examples/cifar10raytune/README.md
```diff
@@ -0,0 +1,85 @@
...
+| Parameter    | Values                             |
+| ------------ | ---------------------------------- |
+| `l1`         | [1, 2, 4, 8, 16, 32, 64, 128, 256] |
+| `l2`         | [1, 2, 4, 8, 16, 32, 64, 128, 256] |
+| `lr`         | [1e-4, 1e-1]                       |
+| `batch_size` | [2, 4, 8, 16]                      |
+
+- `l1`: First fully connected layer size
+- `l2`: Second fully connected layer size
+- `lr`: Learning rate for SGD optimizer
```
rwtarpit: this sounds much more intuitive about `lr` being a distribution rather than hardcoded values
```suggestion
- `l2`: Second fully connected layer size
- `lr`: Log-uniformly sampled learning rate for SGD optimizer
```

examples/cifar10raytune/main.py
```diff
@@ -0,0 +1,191 @@
...
+    return correct / total
+
+
+def tune_cifar(num_samples=10, num_epochs=10, gpus_per_trial=0, cpus_per_trial=2, data_dir="./data"):
+    data_dir = os.path.abspath(data_dir)
+
+    config = {
+        "l1": tune.choice([2**i for i in range(9)]),
+        "l2": tune.choice([2**i for i in range(9)]),
```
rwtarpit: i think we should start both `l1` and `l2` from 4 just as the reference, 

```suggestion
    config = {
        "l1": tune.choice([2**i for i in range(2,9)]),
        "l2": tune.choice([2**i for i in range(2,9)]),
```

examples/cifar10raytune/README.md
atharva7-g: I also added the range.

examples/cifar10raytune/README.md
atharva7-g: Yeah, this is dummy data. I think this will require editing the RunConfig within main.py. We can use Ray's `CLIReporter`. I'll add in actual results soon. 

But I am not sure what you mean by showing bad runs. Do you mean we should check the `TERMINATED` status from ray to determine which trials were completed? And then we could compare with num_epochs to determine if some trial were early stopped. Did I get you correctly? 

Or do you just mean that we should add something similar to the last table in the reference to the README? That I will do soon.

Thanks!

rwtarpit: [trial table](https://docs.ray.io/en/master/tune/examples/tune-pytorch-cifar.html#:~:text=%7C%20%20%20accuracy%20%7C-,training_iteration,-%7C%0A%20%20%7C%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2B%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2B%2D%2D%2D%2D%2D%2D%2D%2B%2D%2D%2D%2D%2D%2D%2B%2D%2D%2D%2D%2D%2D%2B%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2B%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2B%2D%2D%2D%2D%2D%2D%2D%2D%2D%2B%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2B%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D%2D)
yes, if you scroll across the table, you can see the trials getting terminated in early iterations. i think something similar would be great

examples/cifar10raytune/README.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+## Search Space
+
+The example searches over the following hyperparameters:
+
+| Parameter    | Values                             |
+| ------------ | ---------------------------------- |
+| `l1`         | [1, 2, 4, 8, 16, 32, 64, 128, 256] |
+| `l2`         | [1, 2, 4, 8, 16, 32, 64, 128, 256] |
```
rwtarpit: one last minor change, looks good otherwise!!
```suggestion
| Parameter    | Values                             |
| ------------ | ---------------------------------- |
| `l1`         | [4, 8, 16, 32, 64, 128, 256] |
| `l2`         | [4, 8, 16, 32, 64, 128, 256] |
```

atharva7-g: Fixed, thanks!

rwtarpit:
thanks for the PR @atharva7-g .
Looks good from my side @vfdev-5 , i tried to compare with other examples in ignite along with the ray-tune's reference.

examples/cifar10raytune/main.py
```diff
@@ -0,0 +1,197 @@
...
+from utils import Net, get_data_loaders, get_test_loader, load_data
+
+
+def train_with_ignite(config, data_dir=None, checkpoint_dir=None, num_epochs=10, num_workers=8):
+    device = config.get("device", "cpu")
+    net = Net(config["l1"], config["l2"])
+    net = net.to(device)
+
+    if torch.cuda.device_count() > 1:
+        net = nn.DataParallel(net)
```
vfdev-5: I do not think we should handle multiple devices in this example. If we handle it, I think we should do it differently

examples/cifar10raytune/main.py
```diff
@@ -0,0 +1,197 @@
...
+    optimizer = optim.SGD(net.parameters(), lr=config["lr"], momentum=0.9)
+
+    start_epoch = 0
+    if checkpoint_dir:
+        checkpoint_path = os.path.join(checkpoint_dir, "checkpoint.pt")
+        if os.path.exists(checkpoint_path):
+            checkpoint_state = torch.load(checkpoint_path)
+            start_epoch = checkpoint_state["epoch"]
+            net.load_state_dict(checkpoint_state["net_state_dict"])
+            optimizer.load_state_dict(checkpoint_state["optimizer_state_dict"])
```
vfdev-5: Let's use ignite Checkpoint to handle saving and loading checkpoints

examples/cifar10raytune/main.py
```diff
@@ -0,0 +1,197 @@
...
+            )
+
+            checkpoint = Checkpoint.from_directory(tmp_chkpt_dir)
+            tune.report(
+                {"loss": val_loss, "accuracy": val_accuracy},
+                checkpoint=checkpoint,
+            )
+
+    if num_epochs - start_epoch > 0:
+        trainer.run(train_loader, num_epochs - start_epoch)
```
vfdev-5: If we use ignite Checkpoint where we save the trainer state as well we do not need this if-clause

examples/cifar10raytune/main.py
```diff
@@ -0,0 +1,197 @@
...
+    with torch.no_grad():
+        for data in test_loader:
+            images, labels = data
+            images, labels = images.to(device), labels.to(device)
+            outputs = model(images)
+            _, predicted = torch.max(outputs.data, 1)
+            total += labels.size(0)
+            correct += (predicted == labels).sum().item()
+
+    return correct / total
```
vfdev-5: Why don't we use ignite Accuracy metric to compute it ?

examples/cifar10raytune/README.md
```diff
@@ -0,0 +1,104 @@
+# CIFAR10 Example with Ray Tune and Ignite
+
+This example demonstrates how to use **Ray Tune** for hyperparameter tuning with **PyTorch Ignite** for training and validation on the CIFAR10 dataset.
+
+In this example, we show how to use _Ignite_ to train a neural network with:
+
+- hyperparameter search using Ray Tune,
+- ASHA scheduler for early stopping
+- multi-GPU support
```
vfdev-5: For a proper multi gpu support we should use DDP, for example like here: 
- https://github.com/pytorch/ignite/tree/master/examples/cifar10#distributed-training (using function spawn inside the code)
- https://github.com/pytorch/ignite/blob/89b594fe5211048702b3b6098b890079619ba6e6/examples/cifar10/main.py#L201-L202
- auto_model, auto_dataloader etc

examples/cifar10raytune/requirements.txt
```diff
@@ -0,0 +1,5 @@
+pytorch-ignite
+ray[tune]
+torch
+torchvision
+ipywidgets
```
vfdev-5: Why do we need ipywidgets?

examples/cifar10raytune/main.py
rwtarpit: never crossed my mind, the whole point of the example!

examples/cifar10raytune/requirements.txt
atharva7-g: IPyWidgets seems to be common with Ray and Jupyter Notebooks. But I agree, it doesn't seem to be necessary. I'll remove it then?

vfdev-5: If it is Ray dependency it will be installed by ray. Let's remove it

examples/cifar10raytune/README.md
atharva7-g: So would you recommend keeping multi-GPU support or not?

examples/cifar10raytune/main.py
atharva7-g: Thanks very much! Is this what you meant by this:

https://github.com/pytorch/ignite/pull/3574/changes#diff-174b816bdd5c31aac14abc4becdd49d42461aa45ae0252ae03c2c8a631242e43R40-R48

examples/cifar10raytune/main.py
```diff
@@ -0,0 +1,185 @@
...
+import torch.optim as optim
+
+import ray
+from ray import tune
+from ray.tune import Checkpoint as RayCheckpoint, CLIReporter
+from ray.tune.schedulers import ASHAScheduler
+
+from ignite.engine import Events, create_supervised_trainer, create_supervised_evaluator
+from ignite.metrics import Loss, Accuracy
+from ignite.handlers import Checkpoint as IgniteCheckpoint, DiskSaver
```
vfdev-5: Let's use Checkpoint instead of IgniteCheckpoint
```suggestion
from ignite.handlers import Checkpoint, DiskSaver
```

examples/cifar10raytune/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Requirements
+
+- pytorch-ignite: `pip install pytorch-ignite`
+- torchvision: `pip install torchvision`
+- [Ray Tune](https://docs.ray.io/): `pip install ray[tune]`
+- ipywidgets: `pip install ipywidgets`
```
vfdev-5: Remove this line

examples/cifar10raytune/README.md
vfdev-5: Up to you, but let's make it into two steps: 1) remove all multi-gpu support mentions and DP model wrapping etc. 2) in a follow-up PR you can provide multi-gpu support using DDP and ignite.distributed helpers.

examples/cifar10raytune/utils.py
```diff
@@ -0,0 +1,66 @@
...
+
+    train_size = int(len(train_set) * 0.8)
+    val_size = len(train_set) - train_size
+
+    train_subset, val_subset = torch.utils.data.random_split(train_set, [train_size, val_size])
+
+    train_loader = torch.utils.data.DataLoader(
+        train_subset, batch_size=batch_size, shuffle=True, num_workers=num_workers
+    )
+    val_loader = torch.utils.data.DataLoader(val_subset, batch_size=batch_size, shuffle=False, num_workers=num_workers)
```
vfdev-5: Let's use ignite.distributed.auto_dataloader here, similar to original cifar10 example: https://github.com/pytorch/ignite/blob/5901ed499451cfd90de1d2923c52efdd17360732/examples/cifar10/main.py#L211

Same for get_test_loader method

examples/cifar10raytune/main.py
```diff
@@ -0,0 +1,185 @@
...
+        print(f"Epoch {epoch}: loss={val_loss:.4f}, accuracy={val_accuracy:.4f}")
+
+        with tempfile.TemporaryDirectory() as temp_checkpoint_dir:
+            to_save = {"net": net, "optimizer": optimizer, "trainer": trainer}
+            chkpt_handler = Checkpoint(
+                to_save,
+                save_handler=DiskSaver(temp_checkpoint_dir, create_dir=True),
+                filename_pattern="checkpoint.pt",
+            )
+            chkpt_handler(engine)
```
vfdev-5: Actually, this is a bit weird, you save checkpoints every epoch in a different tmp folder which is removed after the TemporaryDirectory context manager.
What does exactly tune.report with checkpoint arg, do they copy the content of `temp_checkpoint_dir` somewhere? If yes, where and how they handle num_epochs of checkpoints: do they store all of them or just the last one etc?

Anyway, to make this code a bit looking more like what ignite does in other examples, we should define the Checkpoint handler and save_handler once in the same level where trainer is defined. The output folder for save_handler is not removed and we would like to be able to use `RayCheckpoint.from_directory` without copying the data.

vfdev-5 (commented):
Thanks for updates @atharva7-g 
I have few more comments on the code, please check

examples/cifar10raytune/main.py
atharva7-g: We can avoid using a temporary directory by using https://docs.ray.io/en/latest/tune/api/doc/ray.tune.TuneContext.get_trial_dir.html. But this API is in public beta right now. 

https://github.com/pytorch/ignite/pull/3574/changes#diff-174b816bdd5c31aac14abc4becdd49d42461aa45ae0252ae03c2c8a631242e43R35-R43

Does this look OK? Thanks a lot!

vfdev-5: I'm not sure what happens exactly in the code with the checkpoints. Can you explain it in details please?

atharva7-g: So from what I understand a usual pattern of using Ray is loading checkpoints using `tune.get_checkpoint()` and saving checkpoints using `tune.report(..., checkpoint=tune.get_checkpoint())`.  

So here, yeah, `torch.save` saves it to a temporary directory and `tune.report` copies it to: `<storage_path>/<exp_name>/<trial_name>/checkpoint_<epoch>` as described [here](https://docs.ray.io/en/latest/tune/tutorials/tune-trial-checkpoints.html). 

In my latest commit I avoid using a tempdir by saving directly to Ray's checkpoint dir using https://docs.ray.io/en/latest/tune/api/doc/ray.tune.TuneContext.get_trial_dir.html. 

I suppose a reason to use temporary directories to write a checkpoint locally first is that you don't have to worry about your own custom directory to be overwritten when checkpointing is done multiple times. With a temporary directory, all your processes are atomic so you don't have to care about cleanup. 

I hope this made sense. Thanks!

vfdev-5: What kind of storages Ray can handle?

atharva7-g: Ray provides support for:
1. cloud storage, S3, GCP etc.
2. NFS
3. on a single node and local filesystem, Tune uses the local filesystem itself.

https://docs.ray.io/en/latest/tune/tutorials/tune-storage.html

If you meant the checkpoint format, Tune doesn't really read whatever checkpoint is stored. If we implement saving and loading logic and pass a config to Tune, Tune will sample params from this config at the start of a new trial and when said trial is created, Tune passes whatever config we had defined.

vfdev-5: what would give get_trial_dir if users set S3 for example as storage path?

It makes sense to use one temp storage during one trial run.

atharva7-g: > what would give get_trial_dir if users set S3 for example as storage path?

`get_trial_dir()` gives out directories such as `~/ray_results`. So even with S3 it will return local paths only AFAIK. 

> It makes sense to use one temp storage during one trial run. 

Hmm was that not what was happening in the first commit? Did you mean something different?

vfdev-5: OK, I confirm that get_trial_dir() returns something like: ` /tmp/ray/session_2026-03-01_22-24-17_720086_2222/artifacts/2026-03-01_22-24-20/train_func_2026-03-01_22-24-17/driver_artifacts/train_func_64a74_00000_0_2026-03-01_22-24-22` even with S3 like storage path.

In this case, the current code is ok.

examples/cifar10_raytune/main.py
```diff
@@ -0,0 +1,171 @@
+import argparse
```
vfdev-5, line 1: Let's rename the folder to `cifar10_raytune`:
```diff
- examples/cifar10raytune
+ examples/cifar10_raytune
```

examples/cifar10raytune/main.py
```diff
@@ -0,0 +1,171 @@
...
+
+    return results
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Ignite + Ray Tune CIFAR10 Example")
+    parser.add_argument("--num_samples", type=int, default=10, help="Number of trials for Ray Tune")
+    parser.add_argument("--num_epochs", type=int, default=10, help="Number of epochs per trial")
+    parser.add_argument("--gpus_per_trial", type=float, default=0, help="GPUs per trial")
+    parser.add_argument("--cpus_per_trial", type=int, default=2, help="CPUs per trial")
```
vfdev-5:
```suggestion
    parser.add_argument("--gpus_per_trial", type=float, default=1, help="GPUs per trial")
    parser.add_argument("--cpus_per_trial", type=int, default=4, help="CPUs per trial")
```

examples/cifar10raytune/main.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def tune_cifar(num_samples=10, num_epochs=10, gpus_per_trial=0, cpus_per_trial=2, data_dir="./data"):
+    data_dir = os.path.abspath(data_dir)
+
+    config = {
+        "l1": tune.choice([2**i for i in range(2, 9)]),
+        "l2": tune.choice([2**i for i in range(2, 9)]),
+        "lr": tune.loguniform(1e-4, 1e-1),
+        "batch_size": tune.choice([2, 4, 8, 16]),
```
vfdev-5:
```suggestion
        "batch_size": tune.choice([256, 512, 1024]),
```

examples/cifar10_raytune/main.py
```diff
@@ -0,0 +1,171 @@
...
+
+    checkpoint_dir = tune.get_context().get_trial_dir()
+    save_handler = DiskSaver(checkpoint_dir, create_dir=True)
+    to_save = {"net": net, "optimizer": optimizer, "trainer": trainer}
+    checkpoint_handler = Checkpoint(to_save, save_handler=save_handler, filename_pattern="checkpoint.pt")
+    trainer.add_event_handler(Events.EPOCH_COMPLETED, checkpoint_handler)
+
+    if checkpoint_dir and os.path.exists(os.path.join(checkpoint_dir, "checkpoint.pt")):
+        load_handler = Checkpoint(to_save, save_handler=DiskSaver(checkpoint_dir, create_dir=False))
+        load_handler.load_objects(checkpoint=os.path.join(checkpoint_dir, "checkpoint.pt"))
```
vfdev-5: Why do we have `checkpoint_dir` arg in the train_with_ignite function signature and we just override it with ` checkpoint_dir = tune.get_context().get_trial_dir()`

I'm assuming that all trials are independent and trial_dir should be unique, so I do not quite understand why we would like to load something from an empty checkpoint folder.
Am I missing something?

Also let's use `pathlib` to join and check if exists instead of `os.path`

examples/cifar10_raytune/README.md
```diff
@@ -0,0 +1,103 @@
...
+│ train_with_ignite_f740a_00003   TERMINATED     32     64   0.0898379               16        1            10.2487   2.32586       0.0985 │
+│ train_with_ignite_f740a_00004   TERMINATED     32     64   0.00243182               4        2            71.8534   1.52549       0.4728 │
+│ train_with_ignite_f740a_00005   TERMINATED    128     16   0.000177485              8        1            18.2198   1.96733       0.2845 │
+│ train_with_ignite_f740a_00006   TERMINATED    256     16   0.00147097               8       10           165.496    1.23385       0.6033 │
+│ train_with_ignite_f740a_00007   TERMINATED     16     16   0.0402147               16        1            10.5927   2.24759       0.1452 │
+│ train_with_ignite_f740a_00008   TERMINATED    256      8   0.000176101              4        2            71.1053   1.51049       0.4422 │
+│ train_with_ignite_f740a_00009   TERMINATED    128     32   0.000174279              4        4           139.638    1.31689       0.5304 │
+╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```
vfdev-5, line 95: I run it on 1 GPU and also see some runs stopped before they do 10 epochs:
```
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Trial name                      status         l1     l2            lr     batch_size     iter     total time (s)      loss     accuracy │
├──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ train_with_ignite_a8b05_00000   TERMINATED      8     64   0.00100238            1024       10           24.9241    2.23429       0.1584 │
│ train_with_ignite_a8b05_00001   TERMINATED     32     32   0.0676801              256       10           24.2626    1.31434       0.534  │
│ train_with_ignite_a8b05_00002   TERMINATED    128    256   0.0708455             1024        1            4.06854   1.98995       0.2522 │
│ train_with_ignite_a8b05_00003   TERMINATED    128     32   0.0224516              512       10           24.7738    1.11887       0.6046 │
│ train_with_ignite_a8b05_00004   TERMINATED      8      8   0.0108409             1024        1            3.9997    2.30114       0.0941 │
│ train_with_ignite_a8b05_00005   TERMINATED     64      8   0.000428377            256        1            4.03655   2.32653       0.098  │
│ train_with_ignite_a8b05_00006   TERMINATED    256      4   0.000215237           1024        1            4.06274   2.34084       0.0973 │
│ train_with_ignite_a8b05_00007   TERMINATED     16      4   0.0707152             1024        2            6.42573   2.19325       0.2103 │
│ train_with_ignite_a8b05_00008   TERMINATED      4    256   0.00340601             256        2            6.25761   1.89807       0.285  │
│ train_with_ignite_a8b05_00009   TERMINATED      8    128   0.000411553           1024        1            4.11832   2.30404       0.1166 │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Do you know why?

examples/cifar10_raytune/main.py
atharva7-g: I think I had kept it in when we were using TempDir, yes it is redundant right now.

examples/cifar10_raytune/README.md
atharva7-g: That is intentional, the example uses ASHAScheduler that doesn't perform all epochs when the model is know to not perform in the top fraction. 

Also, I might have accidentally committed engine.py so GitHub actions wrongly put a engine lable on the PR. Sorry about that.

vfdev-5: Please put a bit more details explaining that in the readme right after the sentence: "We see that most trials stop earlier than the number of maximum epochs (= 10, here)."

examples/cifar10_raytune/main.py
vfdev-5: Actually, in their code they also load existing checkpoint: https://docs.ray.io/en/master/tune/examples/tune-pytorch-cifar.html#the-train-function 
It worth understand why (e.g. if a trial crashed they can restore from the last checkpoint?)

examples/cifar10_raytune/main.py
```diff
@@ -0,0 +1,169 @@
...
+    metrics = {"accuracy": Accuracy(), "loss": Loss(criterion)}
+    evaluator = create_supervised_evaluator(net, metrics=metrics, device=device)
+
+    checkpoint_dir = tune.get_context().get_trial_dir()
+    save_handler = DiskSaver(checkpoint_dir, create_dir=True)
+    to_save = {"net": net, "optimizer": optimizer, "trainer": trainer}
+    checkpoint_handler = Checkpoint(to_save, save_handler=save_handler, filename_pattern="checkpoint.pt")
+    trainer.add_event_handler(Events.EPOCH_COMPLETED, checkpoint_handler)
+
+    checkpoint_path = Path(checkpoint_dir) / "checkpoint.pt"
```
vfdev-5: This is unused now and can be deleted?

examples/cifar10_raytune/README.md
```diff
@@ -0,0 +1,105 @@
...
+## Search Space
+
+The example searches over the following hyperparameters:
+
+| Parameter    | Values                       |
+| ------------ | ---------------------------- |
+| `l1`         | [4, 8, 16, 32, 64, 128, 256] |
+| `l2`         | [4, 8, 16, 32, 64, 128, 256] |
+| `lr`         | [1e-4, 1e-1]                 |
+| `batch_size` | [2, 4, 8, 16]                |
```
vfdev-5: Please update the search space to the code and other logged output

examples/cifar10_raytune/README.md
```diff
@@ -0,0 +1,105 @@
...
+python main.py
+```
+
+### Command-line arguments
+
+| Argument           | Default  | Description                   |
+| ------------------ | -------- | ----------------------------- |
+| `--num_samples`    | 10       | Number of trials for Ray Tune |
+| `--num_epochs`     | 10       | Number of epochs per trial    |
+| `--gpus_per_trial` | 0        | GPUs per trial (0 for CPU)    |
```
vfdev-5: This one as well

examples/cifar10_raytune/main.py
atharva7-g: Do you mean something like this?

```python
if tune.get_checkpoint():
    loaded_checkpoint = tune.get_checkpoint()

    with loaded_checkpoint.as_directory() as ckpt_dir:
        checkpoint_data = torch.load(
            Path(ckpt_dir) / "checkpoint.pt",
            map_location=device,
        )

        net.load_state_dict(checkpoint_data["net"])
        optimizer.load_state_dict(checkpoint_data["optimizer"])

        if "trainer" in checkpoint_data:
            trainer.load_state_dict(checkpoint_data["trainer"])
```

vfdev-5: I mean we should understand why they do that. No code update required right now.

examples/cifar10_raytune/README.md
```diff
@@ -0,0 +1,105 @@
...
+- `l2`: Second fully connected layer size
+- `lr`: Log-uniformly sampled learning rate in [1e-4, 1e-1] for SGD optimizer
+- `batch_size`: Training batch size
+
+## Output
+
+If we run main.py an example output could look like:
+
+```
+╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
```
vfdev-5, line 82: This one as well

examples/cifar10_raytune/main.py
atharva7-g: Ah, okay, I think they just do that if as you say training fails then it is easy to restore.

examples/cifar10_raytune/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+```bash
+python main.py
+```
+
+### Command-line arguments
+
+| Argument           | Default  | Description                   |
+| ------------------ |----------| ----------------------------- |
+| `--num_samples`    | 10       | Number of trials for Ray Tune |
```
vfdev-5: Nit: let's use num_trials instead:
```suggestion
| `--num_trials`    | 10       | Number of trials for Ray Tune |
```

.github/workflows/unit-tests.yml
```diff
@@ -184,3 +184,13 @@ jobs:
...
+      - name: Run CIFAR10 with Ray Tune and Ignite
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        run: |
+          set -e
+          echo "Testing the CIFAR10 with Ray Tune and Ignite example..."
+          if python -c "import torchvision; import torch; import ray; import ignite"; then
+            python examples/cifar10_raytune/main.py --num_epochs 1 --num_samples 1 --cpus_per_trial 1 --gpus_per_trial 0
+          else
+            echo "CIFAR10 with Ray Tune and Ignite example failed to download. Skipping Ray Tune and Ignite example test."
+          fi
```
vfdev-5:
```suggestion
          pip install -r examples/cifar10_raytune/requirements.txt
          python examples/cifar10_raytune/main.py --num_epochs 1 --num_trials 1 --cpus_per_trial 1 --gpus_per_trial 0
```

examples/cifar10_raytune/README.md
```diff
@@ -0,0 +1,105 @@
...
+| `--gpus_per_trial` | 1        | GPUs per trial (0 for CPU)    |
+| `--cpus_per_trial` | 4        | CPUs per trial                |
+| `--data_dir`       | `./data` | Path to CIFAR10 dataset       |
+
+### Examples
+
+Run with custom settings:
+
+```bash
+python main.py --num_samples 5 --num_epochs 5 --gpus_per_trial 1
```
vfdev-5: replace num_samples to num_trials everywhere

.github/workflows/unit-tests.yml
```diff
@@ -184,3 +184,8 @@ jobs:
           else
             echo "Caltech101 dataset failed to download. Skipping SR example test."
           fi
+      - name: Run CIFAR10 with Ray Tune and Ignite
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        run: |
+          pip install -r examples/cifar10_raytune/requirements.txt
```
vfdev-5:
```suggestion
          uv pip install -r examples/cifar10_raytune/requirements.txt
```

PR #3575, opened 2026-02-22
Add py-rouge implementation in tests and remove py-rouge/setuptools dependencies

omkar-334:
`py-rouge` is unmaintained (last release 2018) and difficult to use:
* It depends on `pkg_resources`, which required pinning `setuptools<82`
* The project has no active maintenance and may break as the ecosystem evolves

To remove this fragile dependency while preserving test behavior, this PR adds the minimal `Rouge` implementation used in tests.

---

### Changes

**Vendoring**

* Copied `py-rouge`'s `Rouge` class into `tests/ignite/metrics/nlp/_pyrouge.py`
* Removed `pkg_resources`; data files are resolved via `__file__`
* Bundled WordNet data files alongside the vendored code
* Removed `py-rouge`, `setuptools<82`, and related TODO from `requirements-dev.txt`

**Test fixes**
* Removed unused `worker_id` parameter from `download_nltk_punkt` fixture (caused failures without `pytest-xdist`)
* Added `nltk.download("punkt_tab")` to support newer NLTK versions
---

There are other implementations like `rouge-score` by google and `torchmetrics` by lighting-AI but I didn't find them suitable.  
Ignite’s tests rely on corpus-level multi-reference aggregation (summing raw n-gram counts before computing P/R).
`rouge-score` and `torchmetrics` instead average per-reference scores and do not expose raw counts, leading to different results for unequal-length references.  

All test but one passing
Screenshot 2026-02-23 at 12 56 47 AM


vfdev-5:
@omkar-334 thanks a lot for taking the initiative to fix py-rouge dependency issue, I appreciate that!
Concerning the error in the test, the test looks like a distributed one, but strange that it is failed on the setup...

About bundling the wordnet data, I think we can put it elsewhere, for example, I can create a special repo in the https://github.com/pytorch-ignite org to host the files.

omkar-334:
> Concerning the error in the test, the test looks like a distributed one, but strange that it is failed on the setup...

I realized it was because I didn't install `python-xdist`.  
All are passing now  
Screenshot 2026-02-23 at 10 58 57 AM  


> About bundling the wordnet data, I think we can put it elsewhere, for example, I can create a special repo in the https://github.com/pytorch-ignite org to host the files.

1. tests/ignite/metrics/nlp/wordnet_key_value.txt 
2. tests/ignite/metrics/nlp/wordnet_key_value_special_cases.txt

do you think that doint this would introduce a network dependency?

vfdev-5:
> do you think that doint this would introduce a network dependency?

We just download these files. I do not want to store them in ignite repo.

omkar-334:
sure, you can upload them to a repo and give me the URLs. i can change the current test to download from there and read instead.

vfdev-5:
For now use the URLs from py-rouge repo and we'll update the urls once we have them somewhere else. I wonder whether we could not find another already existing storage for these files

omkar-334:
Done, all tests are passing.

PR #3576, opened 2026-02-23
Replace typing hints with new version in `ignite/engine/__init__.py`

kyuksel:
## Description
- Part of https://github.com/pytorch/ignite/issues/3481
- Replaces legacy `typing` imports (`Dict`, `Optional`, `Sequence`, `Tuple`, `Union`) with the newer syntax (`X | Y`, `X | None`, `dict`, `tuple`) in `ignite/engine/__init__.py`.
- Moves `Sequence` to `collections.abc`

## Changes
- Moved `Sequence` import from `typing` to `collections.abc`
- Removed `Dict`, `Optional`, `Tuple`, `Union` from `typing` imports
- Replaced all `Optional[X]` → `X | None`
- Replaced all `Union[A, B]` → `A | B`
- Replaced `Dict` → `dict`, `Tuple` → `tuple`

## Checklist
- [X] I have read the [Contributing Guide](https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md)
- [X] My changes are consistent with the [Guidelines](https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md#guidelines)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -360,6 +360,7 @@ nitpick_ignore = [
     ("py:class", "torch.utils.data.distributed.DistributedSampler"),
     ("py:class", "torch.amp.grad_scaler.GradScaler"),
     ("py:class", "torch.amp.GradScaler"),
+    ("py:class", "GradScaler"),
     ("py:class", "torch.optim.lr_scheduler._LRScheduler"),
     ("py:class", "torch.optim.lr_scheduler.LRScheduler"),
     ("py:class", "torch.utils.data.dataloader.DataLoader"),


ignite/engine/__init__.py
```diff
@@ -1,5 +1,7 @@
-from collections.abc import Mapping
-from typing import Any, Callable, Dict, Optional, Sequence, Tuple, Union
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
```
aaishwarymishra:
```suggestion
```
Why here we using collections? I think typing provides both mapping and sequence we should be consistent.

ignite/engine/__init__.py
```diff
@@ -1,5 +1,7 @@
...
+from typing import Any, Callable
```
aaishwarymishra:
```suggestion
from typing import Any, Callable, Mapping, Sequence
```

ignite/engine/__init__.py
```diff
@@ -1,5 +1,7 @@
-from collections.abc import Mapping
-from typing import Any, Callable, Dict, Optional, Sequence, Tuple, Union
+from __future__ import annotations
```
vfdev-5, line 1: @kyuksel why exactly we need this import? Generally, we can skip it, but maybe there is something particular here.

ignite/engine/__init__.py
vfdev-5: Can we use `callable` instead of `Callable`? If yes, let's make the replacement. Thanks!

ignite/engine/__init__.py
kyuksel: @vfdev-5 It isn't necessary, given the repo requires Python 3.10+.

 I was referencing a previous PR which added that import ([here](https://github.com/pytorch/ignite/pull/3482/changes#diff-027c45d8fa66c16bb78ec1ea520d0dbdefac1006d3e8fcd256ee991ef01130aeR1)) and I wanted to be consistent, but upon a second look, that was the only other place where the import was added. Sorry about my oversight. 

The import could also have a performance benefit as it results in lazy evaluation but that doesn't matter in this case.

kyuksel: Changed in [c414bc9](https://github.com/pytorch/ignite/pull/3576/commits/c414bc96376e913dbcb40d05bb581a7c39d73a5b).

ignite/engine/__init__.py
kyuksel: `callable` can't be used for type hinting purposes, as far as I know. 

For Python 3.9+, `typing.Callable` is functionally equivalent to `collections.abc.Callable`. Was importing `Callable` from `collections.abc` instead of `typing` what you meant to ask to do by any chance?

ignite/engine/__init__.py
vfdev-5: CI is failing due to type hints, probably we should keep in this file `from __future__ import annotations`

docs/source/conf.py
```diff
@@ -355,6 +355,7 @@ def run(self):
     ("py:class", "torch.utils.data.dataset.Dataset"),
     ("py:class", "torch.utils.data.sampler.BatchSampler"),
     ("py:class", "torch.amp.grad_scaler.GradScaler"),
+    ("py:class", "torch.amp.GradScaler"),
```
kyuksel, line 362: @vfdev-5 This is to fix the CI failure due to doctest error caused by Sphinx treating `torch.amp.GradScaler` in annotation strings as an unresolvable cross-reference.

vfdev-5:
@kyuksel here is what we should do to fix doctest: https://github.com/pytorch/ignite/pull/3586
and I have to fix linkcheck on master as well (well, copilot will do that)

ignite/engine/__init__.py
```diff
@@ -1,5 +1,15 @@
...
+from collections.abc import Mapping, Sequence
+from typing import TYPE_CHECKING, Any, Callable
+
+if TYPE_CHECKING:
+    # GradScaler is imported here rather than used as a string literal ("torch.amp.GradScaler")
+    # because from __future__ import annotations causes Sphinx's autodoc to process annotations
+    # as raw strings. String literals inside | unions trigger unresolvable cross-reference warnings
+    # that the -W flag in docs/Makefile turns into errors. Importing under TYPE_CHECKING lets
+    # autodoc trace GradScaler back to torch.amp.GradScaler, which is in nitpick_ignore.
+    from torch.amp import GradScaler
```
vfdev-5:
```suggestion
    # Min pytorch version support is 2.2.2
    try:
        from torch.amp import GradScaler
    except ImportError:
        from torch.cuda.amp import GradScaler
```

vfdev-5:
Thanks again @kyuksel for your patience with this docs checking system!
Let's merge it finally :)

kyuksel:
> Thanks again @kyuksel for your patience with this docs checking system! Let's merge it finally :)

@vfdev-5 Thank you for the review and the suggestions!

PR #3579, opened 2026-02-23
Refactor: Unify `MetricUsage` into a library-wide `Usage` class and adopt in `EarlyStopping`

TahaZahid05:
**Overview**
This PR addresses maintainer feedback from #3564  by taking `MetricUsage` and generalizing it into a shared Usage class that can be used universally across both metrics and resettable handlers. 

*(Note: This branch currently builds on top of #3564  and includes its commits. I can rebase it once that PR is merged).*

**Key Changes:**
- **Moved Usage to `ignite.base`:** Extracted `MetricUsage` and its subclasses (like EpochWise, BatchWise) into a new `ignite.base.usage` module and renamed it to Usage.
- **Backward Compatibility:** `MetricUsage` is still exposed in `ignite.metrics` as an alias to Usage. It has been officially deprecated in the Sphinx documentation with a warning block.
- **Internal Import Refactoring:** Updated all internal `ignite/metrics` files to import the `Usage` classes directly from `ignite.base.usage`.
- **Updated EarlyStopping:** The attach() method now accepts a single Usage object (defaulting to RunWise()).
- **Graceful Legacy Handling:** If users pass the old `event=` kwarg to `EarlyStopping.attach()`, it throws a standard `DeprecationWarning` but constructs the Usage object under-the-hood so nothing breaks. (Note: `reset_event=` is handled silently without a warning since it was never in a public release).

**Testing**
Updated tests across `tests/ignite/metrics` and `tests/ignite/handlers` to verify the unified typing, deprecation warnings, and runtime stability.

Check list:


vfdev-5:
I haven't checked the details but please keep the backward compatibility (previous users code should not break). We can add a deprecation note, but we should have two classes MetricUsage and Usage.

TahaZahid05:
@vfdev-5 Yeah, I have refactored codebase to use the updated `Usage` class while also ensuring backwards compatibility by using alias. I have also added deprecation note in documentation.

ignite/handlers/early_stopping.py
```diff
@@ -139,33 +141,48 @@ def reset(self) -> None:
     def attach(  # type: ignore[override]
         self,
         engine: Engine,
-        event: Any = Events.COMPLETED,
+        usage: Usage | None = None,
         reset_engine: Engine | None = None,
-        reset_event: Any = Events.STARTED,
         *args: Any,
         **kwargs: Any,
```
vfdev-5, line 147: I think we should rethink a bit the API of attach and update Resettable.attach signature accordingly.

First, let's remove all now useless args from attach method, in the code do not keep any BC as there is no stable release with EarlyStopping.attach.

Now, unfortunately, I find the API more cumbersome than I was thinking. Specifying the usage instead simply an Event is weird:
```python
trainer = ...
evaluator = ...
train_handler = AStatefulHandler(...)
eval_handler = AStatefulHandler(...)

# this API
train_handler.attach(trainer, usage=???)
eval_handler.attach(evaluator, usage=???, reset_engine=trainer)

# previous API
train_handler.attach(
  trainer, 
  event=Events.ITERATION_COMPLETED(every=1000), 
  reset_event=Events.STARTED
)

eval_handler.attach(
  evaluator, 
  event=Events.EPOCH_COMPLETED, 
  reset_engine=trainer,
  reset_event=Events.STARTED,
)
```
Old API looks a bit more clear compared to the new API.
If you have other ideas I would be happy to discuss them.

TahaZahid05: @vfdev-5 

I'll remove BC logic and deprecation warnings for `EarlyStopping.attach()`.

As for your concerns regarding the old API being more clear, I do agree. I think forcing `Usage` onto handlers that really only care about specific trigger events rather than a three-step lifecycle makes the API obscure. So I think the best way forward would be to revert `EarlyStopping` to the old function definition and logic since it is not directly applicable to `Usage`. However, refactoring to `Usage` still makes sense as other handlers like `EpochOutputStore` can use it later down the line to move from hardcoded logic in `attach` to using `Usage`

Let me know if that sounds okay with you and I'll update it.

vfdev-5: Seems like we are trying to patch a bad design and missing a big picture. Let's make few more trials on the handlers attach API. If you have any ideas, please provide a code snippet to make it more clear.

TahaZahid05: @vfdev-5

I agree that the old API is more clear and my solution wasn't general. I propose the following instead:

1. Make `Usage` truly general by accepting flexible `events` rather than 3 hardcoded events:

```python
class Usage:

    def __init__(self, **events):
        self._events = events

    def get(self, role, default=None):
        return self._events.get(role, default)

    @property
    def STARTED(self):
        return self._events.get("started")

    @property
    def COMPLETED(self):
        return self._events.get("completed")

    @property
    def ITERATION_COMPLETED(self):
        return self._events.get("iteration_completed")
```

Presets can now define any number of events:
```python
class EpochWise(Usage):       
    def __init__(self):
        super().__init__(started=Events.EPOCH_STARTED,
                         iteration_completed=Events.ITERATION_COMPLETED,
                         completed=Events.EPOCH_COMPLETED)

class RunWise(Usage):          
    def __init__(self):
        super().__init__(started=Events.STARTED,
                         completed=Events.COMPLETED)
```

**2. Handler `attach` accepts both explicit events and a `Usage` preset** (mutually exclusive):

```python
class EarlyStopping:
    def attach(self, engine, event=None, reset_event=None,
               usage=None, reset_engine=None):

        if usage is not None and (event is not None or reset_event is not None):
            raise ValueError("Specify either 'usage' or explicit events, not both.")

        if usage is not None:
            event = usage.get("completed")
            reset_event = usage.get("started")
            if event is None or reset_event is None:
                raise ValueError(
                    "Usage preset must define 'completed' and 'started' events "
                    "for EarlyStopping."
                )
        else:
            event = event or Events.COMPLETED
            reset_event = reset_event or Events.STARTED

        engine.add_event_handler(event, self)
        (reset_engine or engine).add_event_handler(reset_event, self.reset)
```

```python
# Primary API — explicit, clear
handler.attach(evaluator, event=Events.COMPLETED,
               reset_engine=trainer, reset_event=Events.STARTED)

# Alternative — Usage preset for common patterns
handler.attach(evaluator, usage=RunWise(), reset_engine=trainer)
```

This way:
- The explicit event API acts as the primary API
- `Usage` is general meaning presets can define any number of event roles
- Each handler validates that the preset has the keys it needs
- `Metric.attach` remains unchanged

Does this direction make sense?

vfdev-5: I do not think we should provide two APIs.
Let's keep the first primary API without changing MetricUsage and introducing Usage. Seems like there is no real need.

We can make a structure to hold reset engine and event together (option 2):
```python
handler.attach(
  evaluator, 
  event=Events.COMPLETED,
  reset_options=handler.ResetOptions(engine=trainer, event=Events.STARTED)
)
```
vs option 1:
```
handler.attach(evaluator, event=Events.COMPLETED,
               reset_engine=trainer, reset_event=Events.STARTED)
```

What would you prefer: option 1 or option 2? By the way, let's also ask on discord the community for other thoughts.

TahaZahid05: Understood.

I personally feel Option 1 is better since users don't have to instantiate a new object just to attach a handler. It keeps the signature cleaner IMO.

I will drop a message in the Discord community to see if others have strong preferences between the two or any other ideas.

vfdev-5:
@TahaZahid05 I suggest to close this PR for now. We can get back it later if needed

PR #3584, opened 2026-02-24
fix: replace Optional typing hints with Python 3.10+ syntax in handlers

ramyars466:
## What does this PR do?
Updates typing hints in `ignite/handlers/time_limit.py` and 
`ignite/handlers/timing.py` to use modern Python 3.10+ syntax.

### `ignite/handlers/time_limit.py`
- Removed `from typing import Optional`
- Replaced `Optional[int]` with `int | None`

### `ignite/handlers/timing.py`
- Removed `Optional` from `from typing import Any, Optional`
- Replaced `Optional[Events]` with `Events | None` (lines 99, 100)

## Testing
- All `time_limit` tests passed locally ✅
  - `test_arg_validation` PASSED
  - `test_terminate_on_time_limit[20-10]` PASSED
  - `test_terminate_on_time_limit[5-10]` PASSED
- All `timing` tests passed locally ✅
  - `test_timer` PASSED

### DIFF
--- a/ignite/distributed/comp_models/horovod.py
+++ b/ignite/distributed/comp_models/horovod.py
@@ -1,6 +1,9 @@
+from __future__ import annotations
 import os
 import warnings
-from typing import Any, Callable, cast, List, Mapping, Optional, Tuple, TYPE_CHECKING
+from collections.abc import Callable, Mapping
+from typing import Any, cast, TYPE_CHECKING
+
 
 import torch
 
@@ -48,7 +51,7 @@ if has_hvd_support:
             return rank
 
         @staticmethod
-        def create_from_context() -> Optional["_HorovodDistModel"]:
+        def create_from_context() -> _HorovodDistModel | None:
             rank = _HorovodDistModel._get_hvd_rank()
             # hvd must be initialized
             if not rank > -1:
@@ -66,7 +69,7 @@ if has_hvd_support:
                 raise RuntimeError("Can not re-initialize Horovod if it is already initialized")
             return _HorovodDistModel(backend, **kwargs)
 
-        def __init__(self, backend: Optional[str] = None, **kwargs: Any) -> None:
+        def __init__(self, backend: str | None = None, **kwargs: Any) -> None:
             """This is a private method. Please, use `create_from_backend` or `create_from_context`"""
             super().__init__()
             if backend is not None:
@@ -126,7 +129,7 @@ if has_hvd_support:
             hvd.shutdown()
 
         @staticmethod
-        def _dist_worker_task_fn(backend: str, fn: Callable, args: Tuple, kwargs_dict: Mapping) -> None:
+        def _dist_worker_task_fn(backend: str, fn: Callable, args: tuple, kwargs_dict: Mapping) -> None:
             from ignite.distributed.utils import _set_model, finalize
 
             model = _HorovodDistModel.create_from_backend(backend)
@@ -138,10 +141,10 @@ if has_hvd_support:
         # pyrefly: ignore [bad-override]
         def spawn(
             fn: Callable,
-            args: Tuple,
-            kwargs_dict: Optional[Mapping] = None,
+            args: tuple,
+            kwargs_dict: Mapping | None = None,
             nproc_per_node: int = 1,
-            hosts: Optional[str] = None,
+            hosts: str | None = None,
             backend: str = HOROVOD,
             **kwargs: Any,
         ) -> None:
@@ -183,7 +186,7 @@ if has_hvd_support:
 
         _manual_reduce_op_map = {"MIN": torch.min, "MAX": torch.max, "PRODUCT": torch.prod}
 
-        def _do_all_reduce(self, tensor: torch.Tensor, op: str = "SUM", group: Optional[Any] = None) -> torch.Tensor:
+        def _do_all_reduce(self, tensor: torch.Tensor, op: str = "SUM", group: Any | None = None) -> torch.Tensor:
             if group is not None:
                 raise NotImplementedError("all_reduce with group for horovod is not implemented")
             if op in self._manual_reduce_op_map:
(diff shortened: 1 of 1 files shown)


tests/ignite/conftest.py
```diff
@@ -48,9 +50,12 @@ def pytest_configure(config):
...
     config.addinivalue_line("markers", "tpu: run on tpu")
-    if config.option.treat_unrun_as_failed:
-        unrun_tracker = UnrunTracker()
-        config.pluginmanager.register(unrun_tracker, "unrun_tracker_plugin")
+    try:
+        if config.option.treat_unrun_as_failed:
+            unrun_tracker = UnrunTracker()
+            config.pluginmanager.register(unrun_tracker, "unrun_tracker_plugin")
+    except AttributeError:
+        pass  # Option not available in this pytest version
```
omkar-334, line 54: why was this new `try-except` added?

tests/ignite/conftest.py
```diff
@@ -610,7 +587,9 @@ def pytest_sessionfinish(session, exitstatus):
     This is a pytest hook (due to its name) and is called after the whole test
     run finished, right before returning the exit status to the system.
     """
-    # If requested by the user, track all unrun tests and add them to the lastfailed cache
-    if session.config.option.treat_unrun_as_failed:
-        unrun_tracker = session.config.pluginmanager.get_plugin("unrun_tracker_plugin")
-        unrun_tracker.record_unrun_as_failed(session, exitstatus)
+    try:
```
omkar-334: https://github.com/pytorch/ignite/pull/3584/changes#r2849421828 - same here?

tests/ignite/conftest.py
```diff
@@ -568,10 +547,8 @@ def hvd_worker():
                 hvd.shutdown()
 
             try:
-                # old API
                 from horovod.run.runner import run
             except ImportError:
```
omkar-334, line 573: can you revert all the comments that you removed?  is there any reason for this?

tests/ignite/conftest.py
```diff
@@ -8,10 +8,12 @@
 import time
 from pathlib import Path
 
+
```
omkar-334: also make sure you run the pre-commit before pushing. these extra lines should be reverted.

omkar-334:
> * Wrapped `config.option.treat_unrun_as_failed` in `try/except AttributeError`
>   to fix pytest 9.0+ compatibility issue

is this issue documented anywhere? or did you face any failing tests which led you to fix this?  
if so, you should probably open an issue and a separate PR for this. 
thanks!

tests/ignite/conftest.py
ramyars466: Hi @omkar-334! 

When I ran the tests locally, I was getting this error during collection:

AttributeError: 'Namespace' object has no attribute 'treat_unrun_as_failed'

This was happening in `pytest_configure()` because `pytest 9.0+` changed 
how command-line options are registered, causing `config.option` to not 
have `treat_unrun_as_failed` available at configure time.

The `try-except AttributeError` prevents this crash when the option 
is not yet registered.

However, if this change is not needed or there's a better way to fix it, 
I'm happy to revert it and keep the PR focused only on the typing hints. 
Please let me know!

tests/ignite/conftest.py
ramyars466: Hi @omkar-334! You're right on both counts. I apologize for the 
unintended changes to conftest.py. 

I will revert conftest.py completely to its original state and keep 
this PR focused only on the typing hint changes in 
`time_limit.py` and `timing.py`.

Will push the fix shortly!

tests/ignite/conftest.py
ramyars466: Hi @omkar-334! You're right on both counts. I apologize for the 
unintended changes to conftest.py. 

I will revert conftest.py completely to its original state and keep 
this PR focused only on the typing hint changes in 
`time_limit.py` and `timing.py`.

Will push the fix shortly!

tests/ignite/conftest.py
ramyars466: Sure! I'll run pre-commit before pushing and make sure 
those extra blank lines are cleaned up. Thanks for pointing that out!

ramyars466:
> > * Wrapped `config.option.treat_unrun_as_failed` in `try/except AttributeError`
> >   to fix pytest 9.0+ compatibility issue
> 
> is this issue documented anywhere? or did you face any failing tests which led you to fix this? if so, you should probably open an issue and a separate PR for this. thanks!

ramyars466:
Hi @omkar-334! Yes, I did face a failing test locally — the entire 
test collection was crashing with:

AttributeError: 'Namespace' object has no attribute 'treat_unrun_as_failed'

I added the try-except as a quick fix to unblock myself. 
You're right that this deserves its own issue and PR. 

I'll revert the conftest.py changes from this PR completely and 
open a separate issue for the pytest 9.0+ compatibility problem. 
Thanks for the guidance!

omkar-334:
> Hi @omkar-334! Yes, I did face a failing test locally — the entire test collection was crashing with:
> 
> AttributeError: 'Namespace' object has no attribute 'treat_unrun_as_failed'
> 
> I added the try-except as a quick fix to unblock myself. You're right that this deserves its own issue and PR.
> 
> I'll revert the conftest.py changes from this PR completely and open a separate issue for the pytest 9.0+ compatibility problem. Thanks for the guidance!

interesting. I havent come across this yet. If you open an issue, we can look into this further.

ignite/handlers/time_limit.py
```diff
@@ -1,5 +1,5 @@
 import time
-from typing import Optional
+
```
vfdev-5, line 2:
```suggestion
```

ignite/handlers/time_limit.py
```diff
@@ -28,7 +28,8 @@ class TimeLimit:
     .. versionadded:: 0.4.3
     """
 
-    def __init__(self, limit_sec: Optional[int] = 28800):
+    def __init__(self, limit_sec: int | None = 28800):
+
```
vfdev-5:
```suggestion
    def __init__(self, limit_sec: int | None = 28800):
```

ignite/handlers/time_limit.py
ramyars466: @vfdev-5 Thanks for the review! Applied all suggested changes including removing extra blank lines. Please take another look

PR #3591, opened 2026-02-25
replace typing hints in ignite/handlers/tqdm_logger.py

roli-lpci:
Refs #3481

Description: Replace old typing hints with Python 3.10+ syntax in `ignite/handlers/tqdm_logger.py`.

- `Union[A, B]` -> `A | B`
- `Optional[A]` -> `A | None`
- `List` -> `list`
- Added `from __future__ import annotations`
- Removed unused typing imports (`List`, `Optional`, `Union`)

Check list:

### DIFF
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -34,6 +34,17 @@ Complete list of generic handlers
     param_scheduler.ParamScheduler
     state_param_scheduler.StateParamScheduler
 
+Interfaces
+----------
+
+.. currentmodule:: ignite.base.mixins
+
+.. autosummary::
+    :nosignatures:
+    :toctree: generated
+
+    ResettableHandler
+
 
 Loggers
 --------
--- a/ignite/base/__init__.py
+++ b/ignite/base/__init__.py
@@ -1,1 +1,1 @@
-from ignite.base.mixins import Serializable
+from ignite.base.mixins import Serializable, ResettableHandler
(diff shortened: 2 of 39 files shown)


ignite/handlers/tqdm_logger.py
```diff
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
 """TQDM logger."""
+
```
vfdev-5:
```suggestion
```

ignite/handlers/tqdm_logger.py
```diff
@@ -1,7 +1,10 @@
...
+from __future__ import annotations
```
omkar-334: we don't need this here since we're not referencing any classes defined later. you can remove it

omkar-334 (dismissed):

roli-lpci:
Thanks for the review! Removed the unnecessary `__future__` import and the extra blank line.

ignite/handlers/tqdm_logger.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """TQDM logger."""
 from collections import OrderedDict
-from typing import Any, Callable, List, Optional, Union
+from typing import Any, Callable
```
vfdev-5: I see in different PRs we have either from collections.abc import Callable, Mapping or from typing import Callable, Mapping. Let's make the code consistent and use from collections.abc import Callable, Mapping.

vfdev-5 (approved):
Thanks!

PR #3592, opened 2026-02-25
Replace old-style typing hints in handlers/state_param_scheduler.py

Manimaran-tech:
## Description

Modernize type hints in `ignite/handlers/state_param_scheduler.py` to use Python 3.10+ syntax.

Related to #3481

### Changes made:
- `Union[A, B]` → `A | B`
- `List[...]` → `list[...]`
- `Tuple[...]` → `tuple[...]`
- Removed unused imports (`List`, `Tuple`, `Union`) from `typing`

### Files changed:
- `ignite/handlers/state_param_scheduler.py`

All changes are limited to type annotations in function signatures and local variable declarations across the 6 scheduler classes (`StateParamScheduler`, `LambdaStateScheduler`, `PiecewiseLinearStateScheduler`, `ExpStateScheduler`, `StepStateScheduler`, `MultiStepStateScheduler`). No functional changes.

### DIFF
--- a/ignite/handlers/state_param_scheduler.py
+++ b/ignite/handlers/state_param_scheduler.py
@@ -1,7 +1,8 @@
 import numbers
 import warnings
 from bisect import bisect_right
-from typing import Any, Callable, Sequence
+from collections.abc import Callable, Sequence
+from typing import Any
 
 from ignite.engine import CallableEventWithFilter, Engine, Events, EventsList
 from ignite.handlers.param_scheduler import BaseParamScheduler


ignite/handlers/state_param_scheduler.py
```diff
@@ -1,7 +1,7 @@
 import numbers
 import warnings
 from bisect import bisect_right
-from typing import Any, Callable, List, Sequence, Tuple, Union
+from typing import Any, Callable, Sequence
```
vfdev-5: > I see in different PRs we have either from collections.abc import Callable, Mapping or from typing import Callable, Mapping. Let's make the code consistent and use from collections.abc import Callable, Mapping.

Manimaran-tech: Thanks for pointing that out. I’ll update the imports to use collections.abc to keep things consistent across the codebase.

vfdev-5 (approved):
Thanks!

PR #3594, opened 2026-02-26
Fix #3481: Replace typing hints in stores.py

### LINKED ISSUE #3481: Replace typing hints with new version

ramyars466:
Fixes #3481

Changes:
- `List[Any]` -> `list[Any]`
- `Optional[str]` -> `str | None` 
- `from collections.abc import Callable`

Tests pass ✅

### DIFF
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -34,6 +34,17 @@ Complete list of generic handlers
     param_scheduler.ParamScheduler
     state_param_scheduler.StateParamScheduler
 
+Interfaces
+----------
+
+.. currentmodule:: ignite.base.mixins
+
+.. autosummary::
+    :nosignatures:
+    :toctree: generated
+
+    ResettableHandler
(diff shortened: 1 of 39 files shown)


ramyars466:
Thank you for the review @vfdev-5! 😊 
Is the linkcheck failure related to my changes 
or a pre-existing issue? Happy to fix if needed!

PR #3609, opened 2026-02-27
Replace old typing in top_k_categorical_aaccuracy.py

rogueslasher:
part of  #3481

Replaced old typing hints with Python 3.10 syntax in
`top_k_categorical_accuracy.py`.

Changes include:
- Union[...] → | syntax
- Removed typing imports in favor of collections.abc 
- Updated return type hints


vfdev-5 (approved):
Thanks !

PR #3610, opened 2026-02-27
[Feature] Implement TopK wrapper

### LINKED ISSUE #3568: [Feature] Implement Recall@top_k for RecSys
## 🚀 Feature
(Follow up to #3530 and #2631)

After getting in the `HitRate@top_k` metric for recommendation systems, it makes sense to have `Recall@top_k` as well. To implement this metric, as discussed we can add a `top_k` argument to existing `Recall` metric instead of creating a new class:

The idea is to use the `_prepare_output` method of `_BasePrecision_Recall` to convert our data to the required format that can then be used directly with same logic of `if self._average == "samples":` of `Recall` class:
for the `true_positives`, fetch the top-k indices of `y_pred` and gather those elements from `y`.
`y` can then also be directly be passed as `actual_positives`.
https://github.com/pytorch/ignite/blob/24535a0bb078398d1374657ce6f8c1f583cd467a/ignite/metrics/precision.py#L60
<pre>
_, top_indices = torch.topk(y_pred, k=self._top_k, dim=1)
true_positives = torch.gather(y, 1, top_k_indices)
</pre>

then in https://github.com/pytorch/ignite/blob/24535a0bb078398d1374657ce6f8c1f583cd467a/ignite/metrics/recall.py#L229
we can use this logic for final metric calculation

rwtarpit:
Fixes #3568 

Description:

As discussed in the issue thread, this is a **draft PR** for implementing a base structure for the `TopK` wrapper class.
the idea is to skip redundant checks on input data by skipping `_check_shape` and `_check_type` `k` times.

the wrapper only instantiates a single object of the base class and maintains states for each `k` using a `dict`.

for each batch, we check the data shape and type only once before updating the metric for each `k` individually. for this we use `_skip_checks` flag that can only be set from the `TopK` itself so it ensures metric function as usual when called without wrapper.

`TopK` class provides `_wrap_prepare_output` that handles top-k logic and transforms data accordingly. then metric can be calculated by base metric's update and compute as usual by the means of `state dict` and `_state_dict_all_req_keys`

I have currently only implemented the wrapper for recall and precision currently to get an idea how it would look when extended to other metrics in future
Check list:

### DIFF
--- a/ignite/metrics/precision.py
+++ b/ignite/metrics/precision.py
@@ -88,10 +88,6 @@ class _BasePrecisionRecall(_BaseClassification):
 
         return y_pred, y, correct
 
-    @classmethod
-    def _topk_transform(cls, output: Sequence[torch.tensor], top_k: int) -> Sequence[torch.tensor]:
-        return output
-
     @reinit__is_reduced
     def reset(self) -> None:
         """
--- a/ignite/metrics/precision.py
+++ b/ignite/metrics/precision.py
@@ -88,6 +88,10 @@ class _BasePrecisionRecall(_BaseClassification):
 
         return y_pred, y, correct
 
+    @classmethod
+    def _topk_transform(cls, output: Sequence[torch.tensor], top_k: int) -> Sequence[torch.tensor]:
+        return output
+
     @reinit__is_reduced
     def reset(self) -> None:
         """
@@ -410,8 +414,9 @@ class Precision(_BasePrecisionRecall):
                 For binary and multilabel data, both y and y_pred should consist of 0's and 1's, but for multiclass
                 data, y_pred and y should consist of probabilities and integers respectively.
         """
-        self._check_shape(output)
-        self._check_type(output)
+        if not getattr(self, "_skip_checks", False):
+            self._check_shape(output)
+            self._check_type(output)
         y_pred, y, correct = self._prepare_output(output)
 
         if self._average == "samples":
--- a/ignite/metrics/top_k.py
+++ b/ignite/metrics/top_k.py
@@ -9,7 +9,8 @@ class TopK(Metric):
     """https://github.com/open-mmlab/mmengine/blob/main/mmengine/registry/registry.py
         https://github.com/open-mmlab/mmdetection/tree/main
 
-    the idea is to maintain tok_k transforms here for metrics.
+    the idea is to maintain top_k transforms here for metrics.
+    each topk_transform will be registered in TopK class.
     and user will have to pass output_transform to TopK instead of base_metric
     and the output transform must only unpack output into y_pred, y, and do no other transformations like binarisation
     """
@@ -46,8 +47,11 @@ class TopK(Metric):
         self._states = {k: self._base_metric.state_dict() for k in self._ks}
 
     def update(self, output):
-        self._base_metric._check_shape(output)
-        self._base_metric._check_type(output)
+        """run checks on output only once for highest `k` value and then skip checks
+        in base metric's update for each `k`"""
+        _output = self._transform(output, self._ks[-1])
+        self._base_metric._check_shape(_output)
+        self._base_metric._check_type(_output)
         self._base_metric._skip_checks = True
 
         for k in self._ks:
@@ -71,6 +75,7 @@ class TopK(Metric):
 
 
 def _precision_recall_topk_transform(output: Sequence[torch.Tensor], k: int):
+    """top_k transform for precision and recall"""
     y_pred, y = output[0], output[1]
     _, top_indices = torch.topk(y_pred, k=k, dim=-1)
     masked = torch.zeros_like(y_pred)
--- a/ignite/metrics/top_k.py
+++ b/ignite/metrics/top_k.py
@@ -1,0 +1,81 @@
+import torch
+from typing import Sequence
+
+from ignite.metrics import Metric
+from ignite.metrics.precision import _BasePrecisionRecall
+
+
+class TopK(Metric):
+    """https://github.com/open-mmlab/mmengine/blob/main/mmengine/registry/registry.py
+        https://github.com/open-mmlab/mmdetection/tree/main
+
+    the idea is to maintain tok_k transforms here for metrics.
+    and user will have to pass output_transform to TopK instead of base_metric
+    and the output transform must only unpack output into y_pred, y, and do no other transformations like binarisation
+    """
+
+    _output_transform_registry = {}
+
+    @classmethod
+    def register(cls, metric_type, k_transform):
+        cls._output_transform_registry[metric_type] = k_transform
+
+    def __init__(
+        self,
+        base_metric: Metric,
+        top_k: int | list[int],
+        output_transform=lambda x: x,
+        device: str | torch.device = torch.device("cpu"),
+        skip_unrolling: bool = False,
+    ):
+        transform = None
+        for metric_type, k_transform in self._output_transform_registry.items():
+            if isinstance(base_metric, metric_type):
+                transform = k_transform
+
+        if transform is None:
+            raise ValueError(f"{type(base_metric).__name__} does not support TopK.")
+
+        self._transform = transform
+        self._base_metric = base_metric
+        self._ks = sorted(top_k) if isinstance(top_k, list) else [top_k]
+        super().__init__(output_transform=output_transform, device=device, skip_unrolling=skip_unrolling)
+
+    def reset(self):
+        self._base_metric.reset()
+        self._states = {k: self._base_metric.state_dict() for k in self._ks}
+
+    def update(self, output):
+        self._base_metric._check_shape(output)
+        self._base_metric._check_type(output)
+        self._base_metric._skip_checks = True
+
+        for k in self._ks:
+            # restore state for this k
+            self._base_metric.load_state_dict(self._states[k])
+
+            k_output = self._transform(output, k)
+            self._base_metric.update(k_output)
+
+            # save state for this k
+            self._states[k] = self._base_metric.state_dict()
+
+        self._base_metric._skip_checks = False
+
+    def compute(self) -> list:
+        results = []
+        for k in self._ks:
+            self._base_metric.load_state_dict(self._states[k])
+            results.append(self._base_metric.compute())
+        return results
+
+
+def _precision_recall_topk_transform(output: Sequence[torch.Tensor], k: int):
+    y_pred, y = output[0], output[1]
+    _, top_indices = torch.topk(y_pred, k=k, dim=-1)
+    masked = torch.zeros_like(y_pred)
+    masked.scatter_(-1, top_indices, 1.0)
+    return (masked, y)
+
+
+TopK.register(_BasePrecisionRecall, _precision_recall_topk_transform)
--- a/tests/ignite/metrics/test_precision.py
+++ b/tests/ignite/metrics/test_precision.py
@@ -503,3 +503,34 @@ class TestDistributed:
             if average == "weighted":
                 assert pr._weight.device == metric_device, f"{type(pr._weight.device)}:{pr._weight.device} vs "
                 f"{type(metric_device)}:{metric_device}"
+
+
+def test_top_k_precision_multilabel_samples():
+    """dummy test for checking workflow of TopK wrapper and registry idea"""
+    import torch
+    from ignite.metrics import Precision
+    from ignite.metrics.top_k import TopK
+
+    y_pred = torch.tensor(
+        [
+            [0.9, 0.3, 0.8, 0.1],
+            [0.5, 0.8, 0.2, 0.9],
+            [0.3, 0.6, 0.9, 0.1],
+        ]
+    )
+    y_true = torch.tensor(
+        [
+            [1, 0, 1, 0],
+            [0, 1, 0, 1],
+            [0, 1, 1, 0],
+        ]
+    )
+
+    metric = TopK(Precision(average="samples", is_multilabel=True), top_k=[1, 2, 3])
+    metric.update((y_pred, y_true))
+    result = metric.compute()
+
+    assert len(result) == 3
+    assert pytest.approx(result[0], abs=1e-4) == 1.0
+    assert pytest.approx(result[1], abs=1e-4) == 1.0
+    assert pytest.approx(result[2], abs=1e-4) == 2 / 3


ignite/metrics/top_k.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def _wrap_prepare_output(self, output, k):
+        if isinstance(self._base_metric, _BasePrecisionRecall):
+            y_pred, y = output[0], output[1]
+            _, top_indices = torch.topk(y_pred, k=k, dim=-1)
+            masked = torch.zeros_like(y_pred)
+            masked.scatter_(-1, top_indices, 1.0)
+            return (masked, y)
+
+    @reinit__is_reduced
```
vfdev-5: I think, we do not need to decorate method with reinit__is_reduced and reinit__is_reduced as base metric will handle that.

ignite/metrics/top_k.py
```diff
@@ -0,0 +1,60 @@
...
+    def update(self, output):
+        self._base_metric._check_shape(output)
+        self._base_metric._check_type(output)
+        self._base_metric._skip_checks = True
+
+        for k in self._ks:
+            # restore state for this k
+            for attr in self._base_metric._state_dict_all_req_keys:
+                setattr(self._base_metric, attr, self._states[k].get(attr, getattr(self._base_metric, attr)))
```
vfdev-5: We can use Metric.state_dict and Metric.load_state_dict for that?

vfdev-5 (commented):
Thanks for the PR @rwtarpit !
I checked the code and thinking on how to make it a bit better.
We may need to figure out precisely as well on what kind of inputs we can compute top-k metric: binary, multi-class, multi-label? All of them or a set of them...

rwtarpit:
> Thanks for the PR @rwtarpit ! I checked the code and thinking on how to make it a bit better. We may need to figure out precisely as well on what kind of inputs we can compute top-k metric: binary, multi-class, multi-label? All of them or a set of them...

yes this would need some research before finalizing the metric.
also i was thinking if we go with current implementation of implementing `_wrap_prepare_output` in `TopK` itself, the method might grow too big with branching and all. we can create `_wrap_prepare_output` method in base metric instead of  `TopK` for all metrics as you proposed earlier for the metrics that can have TopK feature, and TopK instantiation will fail on metrics that don't have this method defined

ignite/metrics/top_k.py
rwtarpit: i'll take a look at this

rwtarpit:
@vfdev-5 
i checked the various supported input types and these are some of the distinctions i would like to flag:

TopK woudn't support `binary` input for any metric:
```python
y_pred = torch.tensor([0.9, 0.3, 0.7])  # (N,) 
y      = torch.tensor([1,   0,   1  ])
# only one element per sample
```
Multiclass case:
```python
#k=2
y_pred = torch.tensor([[0.1, 0.6, 0.2, 0.4],   # top2: class 1,3
                           [0.5, 0.2, 0.8, 0.3],   # top2: class 2,0
                           [0.3, 0.1, 0.2, 0.9]])  # top2: class 3,0
y = torch.tensor([1, 2, 1])    # true classes
# accuracy@2 = 2/3 = 0.66
# samples: recall@2 = (1/1 + 1/1 + 0/1)/3 = 0.66 = accuracy@2
# micro:   recall@2 = sum(TP)/sum(actual) = 2/3 = 0.66 = accuracy@2
# samples: precision@2 = (1/2 + 1/2 + 0/2)/3 = 0.333
# micro:   precision@2 = sum(TP)/N*k = 2/6 = 0.333
```
 - Multiclass `Accuracy@k` would be same as existing `TopKCategoricalAccuracy`
 - also for Recall@k for `average = samples or micro`, this would be same as TopKCategoricalAccuracy as `actual_positivies=1` for all samples.
  - For Recall@k with `average=macro and weighted`, it is logical and distinct from `Accuracy@k` since these averages compute recall per class : `TP_class / actual_class`, rather than collapsing to a simple per sample hit/miss, making it sensitive to class imbalance unlike accuracy.
  ```python
  #macro recall@2:
 class 1: actual= sample 1 and 3, predicted = sample 1; recall@class1 = 1/2=0.5
 class 2: actual= sample 2, predicted = sample 2; recall@class2 = 1/1=1.0
 macro recall@2 = (0.5+1.0)/2 = 0.75 
  ```
 - and `Precision@k` for `average=samples/micro` doesn't make sense too as again, since there is only one true class per sample, `precision@k <= 1/k` always.
 - Precision@k would make sense for `average = macro or weighted` as we will be calculating per class precision rather than per sample:
`TP_class / predicted_class`  
(how many times did we predict this class in top-k and were actually correct)

Multilabel case:
 - For multilabel, TopK should be able to expanded for all averaging strategies since each sample can have multiple true labels, ie both `actual_positives > 1 and TP > 1` are possible, making precision and recall denominators variable unlike multiclass where `actual_positives=1`.
 
Similary `F-score@k` should be valid where both precision@k and recall@k are valid

rwtarpit:
> also i was thinking if we go with current implementation of implementing `_wrap_prepare_output` in `TopK` itself, the method might grow too big with branching and all. we can create `_wrap_prepare_output` method in base metric instead of `TopK` for all metrics as you proposed earlier for the metrics that can have TopK feature, and TopK instantiation will fail on metrics that don't have this method defined

can you confirm this too, i can then proceed with the implementation

rwtarpit:
I have updated the draft PR with idea of keeping the `TopK` logic seperated and decoupled as much as possible from metrics(them being already stable):

 - The idea is to maintain the `topk_transform` registry for metrics we want to implement in single file itself. `TopK` wrapper will use these transforms to convert output accordingly before passing it to base_metric's update in a loop for each k. 
 - Also to skip the redundant input validation checks, base metrics will need to have a small addition of `skip_checks` logic that can only be fired from `TopK` itself (this keeps base_metrics stable with minor change)
 
 Advantages:
 - keeps code clean and modular by limiting TopK logic to itself.
 - user can register their own topk logic by implementing custom topk_transform for their custom metric
 - skip redundant checks and supports list[k]
 
 Disadvantages:
 - `base_metric` is  never attatched to engine, so its `output_transform` will never be called if user passes it. we would need to keep checks and clear documentation about this.
 - for `topk` logic to work, the `output_transform` passed to it should only unpack the output into `(y_pred, y)` and not do any other transformations like binarisation. again this would require some checks and warnings
 - slightly different than existing API in terms of UX

NOTE:
the registry idea is taken from :
https://github.com/open-mmlab/mmengine/blob/main/mmengine/registry/registry.py
https://github.com/open-mmlab/mmdetection/tree/main

i also checked how torch metrics implement TopK. they have topk logic for each metric where topk is passed as arg:
` precision = Precision(..., top_k=4)`
but thing is they pass topk as int.
if we want to have topk feature supporting list[int], i really can't think of other than having an additional wrapper class.


currently i have kept it to precision/recall and added a small testcase to check sanity of the idea. this is a rough sketch of the idea and we still need to think of adding of other metrics and edge cases/ exceptions before moving forward and finalising this idea

ignite/metrics/precision.py
```diff
@@ -410,8 +410,9 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
                 For binary and multilabel data, both y and y_pred should consist of 0's and 1's, but for multiclass
                 data, y_pred and y should consist of probabilities and integers respectively.
         """
-        self._check_shape(output)
-        self._check_type(output)
+        if not getattr(self, "_skip_checks", False):
```
steaphenai, line 413:
```suggestion
        if not self._skip_checks:
```
I think using `getattr` might implies the attribute might not exist when this method runs. So it is better to initialize `self._skip_checks = False`

ignite/metrics/top_k.py
```diff
@@ -0,0 +1,86 @@
+import torch
+from typing import Sequence
+
+from ignite.metrics import Metric
```
steaphenai, line 4:
```suggestion
from ignite.metrics import Metric
from ignite.metrics.metric import reinit__is_reduced
```

ignite/metrics/top_k.py
```diff
@@ -0,0 +1,86 @@
...
+        for metric_type, k_transform in self._output_transform_registry.items():
+            if isinstance(base_metric, metric_type):
+                transform = k_transform
+
+        if transform is None:
+            raise ValueError(f"{type(base_metric).__name__} does not support TopK.")
+
+        self._transform = transform
+        self._base_metric = base_metric
+        self._ks = sorted(top_k) if isinstance(top_k, list) else [top_k]
```
steaphenai, line 42:
```suggestion
        self._ks = sorted(top_k) if isinstance(top_k, list) else [top_k]
       identity = lambda x: x
       if base_metric.output_transform is not identity:
            import warnings
            warnings.warn(
                "base_metric's output_transform will never be called inside TopK. "
                "Pass output_transform to TopK directly instead.",
                UserWarning,
            )
```
`TopK` feeds the data directly to the base metric's `update` method, it completely bypasses the base metric's own `output_transform`. If a user passes a custom transform to the base metric, it will silently fail to run. So I think its better to add some warning.

ignite/metrics/top_k.py
```diff
@@ -0,0 +1,86 @@
...
+        super().__init__(output_transform=output_transform, device=device, skip_unrolling=skip_unrolling)
+
+    def reset(self):
```
steaphenai, line 45:
```suggestion
   @reinit__is_reduced
    def reset(self):
```
Its better to add this decorator because its resets the flag when new epoch starts

ignite/metrics/top_k.py
```diff
@@ -0,0 +1,86 @@
...
+        self._base_metric.reset()
+        self._states = {k: self._base_metric.state_dict() for k in self._ks}
+
+    def update(self, output):
```
steaphenai, line 49:
```suggestion
    @reinit__is_reduced
    def update(self, output):
```
I guess we need the decorator here too

ignite/metrics/top_k.py
```diff
@@ -0,0 +1,86 @@
...
+
+            k_output = self._transform(output, k)
+            self._base_metric.update(k_output)
+
+            # save state for this k
+            self._states[k] = self._base_metric.state_dict()
+
+        self._base_metric._skip_checks = False
+
+    def compute(self) -> list:
```
steaphenai, line 69: Distributed sync is missing here. The state tensors for each k inside `self._states` are never reduced across GPUs before `compute()` is called. I think we should add a docstring noting that distributed evaluation is not yet supported for TopK to avoid silent failures.

ignite/metrics/top_k.py
```diff
@@ -0,0 +1,86 @@
...
+        for k in self._ks:
+            self._base_metric.load_state_dict(self._states[k])
+            results.append(self._base_metric.compute())
+        return results
+
+
+def _precision_recall_topk_transform(output: Sequence[torch.Tensor], k: int):
+    """top_k transform for precision and recall"""
+    y_pred, y = output[0], output[1]
+    _, top_indices = torch.topk(y_pred, k=k, dim=-1)
```
steaphenai, line 80:
```suggestion
     actual_k = min(k, y_pred.shape[-1])
    _, top_indices = torch.topk(y_pred, k=actual_k, dim=-1)
```
I think we need to guard against `k` exceeding the total number of items. If a user asks for `k=10` but the test batch only has 5 items, `torch.topk` will throw a `RuntimeError` and leds to crash. So taking min prevents this.

tests/ignite/metrics/test_precision.py
```diff
@@ -503,3 +503,34 @@ def test_multilabel_accumulator_device(self, average):
...
+    )
+
+    metric = TopK(Precision(average="samples", is_multilabel=True), top_k=[1, 2, 3])
+    metric.update((y_pred, y_true))
+    result = metric.compute()
+
+    assert len(result) == 3
+    assert pytest.approx(result[0], abs=1e-4) == 1.0
+    assert pytest.approx(result[1], abs=1e-4) == 1.0
+    assert pytest.approx(result[2], abs=1e-4) == 2 / 3
```
steaphenai, line 536:
```suggestion
    assert pytest.approx(result[2], abs=1e-4) == 2 / 3
def test_top_k_k_exceeds_num_items():
    """torch.topk crashes if k > number of items — verify guard works."""
    from ignite.metrics import Precision
    from ignite.metrics.top_k import TopK
    import torch

    y_pred = torch.tensor([[0.9, 0.3, 0.8]])   # 3 items
    y_true = torch.tensor([[1, 0, 1]])

    metric = TopK(Precision(average="samples", is_multilabel=True), top_k=[5])  # k=5 > 3 items
    metric.update((y_pred, y_true))   # should NOT crash

    result = metric.compute()
    assert len(result) == 1
```
Lets add this unit test to verify the `min(k, total_items)` guard works and prevents a `RuntimeError` !

steaphenai (commented):
@rwtarpit Great work on this TopK wrapper! @vfdev-5 and @rwtarpit I have some few suggestions. Let me know what you think.

ignite/metrics/top_k.py
steaphenai: @vfdev-5 I just checked your previous comment that `reinit__is_reduced` flag automatically. If `self._base_metric.update() handles the flag reset for the wrapper, then we can safely ignore the decorator suggestions.

ignite/metrics/top_k.py
```diff
@@ -0,0 +1,86 @@
...
+        cls._output_transform_registry[metric_type] = k_transform
+
+    def __init__(
+        self,
+        base_metric: Metric,
+        top_k: int | list[int],
+        output_transform=lambda x: x,
+        device: str | torch.device = torch.device("cpu"),
+        skip_unrolling: bool = False,
+    ):
```
vfdev-5, line 31: @TahaZahid05 seems like we have a similar idea/problem as SubgroupMetric from your PR. Can you check this implementation and give the feedback, thanks!

vfdev-5:
> base_metric is never attatched to engine, so its output_transform will never be called if user passes it. we would need to keep checks and clear documentation about this.

@rwtarpit you can check https://github.com/pytorch/ignite/blob/6bfbdc488e74ef95250f77a8539e689fedeb7942/ignite/metrics/metrics_lambda.py#L148-L170  where lambda metric is contructed from inner metrics and when lambda metric is attached, inner metrics are also attached.

TahaZahid05:
@vfdev-5 Thanks for tagging.

Part of my approach has already been discussed in #3568 from what I can see. I'll list down my approach again for explanation purposes:

The main difference is in state management. This PR uses a single instance and swaps `state_dicts` across the different K's. in my PR (#3627), i used `copy.deepcopy` to create a seperate instance per key:
```python
self._metrics = {k: copy.deepcopy(base_metric) for k in keys}
```

This simplifies the `update()` function as well because there is no need for `_skip_checks` since each `k` has its own instance. Even though this PR approach is more optimized by skipping redundant checks, it requires modifying each base's metric `update()` to add the `_skip_checks` guard, which is less scalable as every future metric that wants TopK support would need the same change.

Following is how the update looks under my implementation:
```python
for k in self._ks:
    k_output = self._transform(output, k)
    self._metrics[k].update(k_output)

```

and `compute()` simply returns the results:
```python
return {k: self._metrics[k].compute() for k in self._ks}
```

This is better in my opinion as you don't need to add the `_skip_checks`, meaning no modification to the base metric itself, and also you don't suffer from the distributed sync issues of the current implementation, as @steaphenai mentioned which don't occur under my implementation due to simplicity of it.

@rwtarpit mentioned in #3568 that they considered the use of `deepcopy` but were concerned about it being costly. Since `Precision` contains just a few tensors, I think cloning them for a typical set of a few values of k would be negligible and a good tradeoff compared to simplicity in implementation.

ignite/metrics/top_k.py
TahaZahid05: @vfdev-5 i have made a general comment on the PR.

rwtarpit:
thanks for the explaination @TahaZahid05 , i was bit reluctant too with this approach due to touching stable API of metrics.
i'll let you guys(@TahaZahid05 and @vfdev-5 ) decide on this part, if to use deepcopy or single instance for state management.
overall is using the registry approach feasible??

vfdev-5:
> overall is using the registry approach feasible??

@rwtarpit I could not understand the usefulness of registry approach. If you can pitch again the idea with clear example, it would be helpful.

rwtarpit:
so i tried to interleave many issues we were facing with TopK:
 - for code maintainablity we don't need to change update method of TopK for future metrics, we would just need to add transform function for new metric and register it. otherwise we would have branching logic in update method with `is_instance(base_metric, ...)` for each new metric added.
  - this gives user ability to add a custom TopK if they want for their custom metric by just writing the transform method and registering it in TopK for their custom metric. (although there is a edge case: what if user adds a custom topk transform logic for already registered base metric)
   - TopK's logic remains seperated in single file from base metrics and we don't need to touch them.
   
   
```python
#TopK's update 
class TopK:
     # topk with branching logic
    def update(base_metric, top_k,...):
        if is_instance(base_metric, PrecisonRecall):
        ...
        if is_instance(base_metric, Accuracy):
        ...
        
    # with transform registry logic:
    def update(base_metric, top_k,...):
        transform = None
            for metric_type, k_transform in self._output_transform_registry.items():
                if isinstance(base_metric, metric_type):
                    transform = k_transform
        ...
 # registering metrics to TopK    
TopK.register(PrecisonRecall, _precision_recall_transform)     
TopK.register(PrecisonRecall, _accuracy_transform) 
...          
``` 
 - so the main goal is to keep it maintainable and provide option of custom TopK transform logic too. apparently if `TopK.register` doesn't sound aligned with customs of repo, maybe we can use `TopK.attatch`??

star11293:
Hey @vfdev-5, @rwtarpit, @TahaZahid05, @steaphenai,

been reading through #3568, this PR, and the related discussion and i have a few thoughts.

On input types, @rwtarpit's Mar 3 breakdown was useful. Binary doesnt really fit, multiclass overlaps a lot with `TopKCategoricalAccuracy` (especially with samples/micro averaging), and multilabel is where the wrapper actually adds something new. Starting with multilabel and leaving standard multiclass top k accuracy to `TopKCategoricalAccuracy` would cut down the design surface for a first version.

On state management, @TahaZahid05's deepcopy per k approach already shipped in #3627 for the fairness metrics, so there's some precedent there. The tradeoff comes down to deepcopy cost vs touching base metrics with `_skip_checks`. The actual cost depends on how big the base metric state is and how many k values you have, so it's pretty empirical.

I can put together a small benchmark comparing both approaches across a few state sizes and k list lengths if it would help move things forward. Not trying to take over the implementation, you both have put real thought into this.

@vfdev-5 would starting the first version with multilabel make this easier to land?

PR #3615, opened 2026-03-01
Replace old typing hints with Python 3.10 syntax in visdom_logger.py

rogueslasher:
Part of #3481

Replaced old typing hints with Python 3.10 syntax in
'ignite/handlers/visdom_logger.py`.

Changes include:

Union[A, B] → A | B syntax
Optional[A] → A | None syntax
List, Dict → list, dict
Moved Callable import from typing to collections.abc

### DIFF
--- a/ignite/handlers/visdom_logger.py
+++ b/ignite/handlers/visdom_logger.py
@@ -2,7 +2,8 @@
 
 import os
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Any
+from concurrent.futures import ThreadPoolExecutor
+from typing import Any
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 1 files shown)


ignite/handlers/visdom_logger.py
```diff
@@ -201,7 +202,7 @@ def __init__(
         if not self.vis.offline and not self.vis.check_connection():  # type: ignore[attr-defined]
             raise RuntimeError(f"Failed to connect to Visdom server at {server}. Did you run python -m visdom.server ?")
 
-        self.executor: Union[_DummyExecutor, "ThreadPoolExecutor"] = _DummyExecutor()
+        self.executor: _DummyExecutor | "ThreadPoolExecutor" = _DummyExecutor()
```
vfdev-5: This type hint is failing. To make it work we should remove quotes from ThreadPoolExecutor but in this file we have lazy import of ThreadPoolExecutor due to old support of python 2.7 and ThreadPoolExecutor was introduced to python from 3.2.
Please remove the lazy import and import ThreadPoolExecutor in the top of the file as we now support python 3.10+

PR #3612, opened 2026-02-28
fix: resolve doctest failures in FID and CalinskiHarabaszScore metrics

TahaZahid05:
Fixes #3611
Close #3616, #3620

Description:
Fixes 2 doctest failures found when running `make doctest`:

1. `ignite/metrics/gan/fid.py`: `scipy.linalg.sqrtm(disp=False)` no longer returns a tuple in newer scipy versions, causing `ValueError: not enough values to unpack`. Fixed by removing tuple unpacking and adding an `isinstance` check for backward compatibility.
2. `ignite/metrics/clustering/calinski_harabasz_score.py`: Replaced exact float expected output `5.733935832977295` with ellipsis pattern `5.73393...` to handle float variance.

Check list:

- [ ] New tests are added (if a new feature is added)
- [ ] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #3616, opened 2026-03-01
testing cache invalidation

TahaZahid05:
Tempoary PR to show doctest failure.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.10"]
+        python-version: ["3.12"]
     steps:
       - uses: actions/checkout@v6
 
@@ -70,7 +70,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.10"]
+        python-version: ["3.12"]
     steps:
       - uses: actions/checkout@v6
 
@@ -109,7 +109,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.10"]
+        python-version: ["3.12"]
     steps:
       - uses: actions/checkout@v6
 


vfdev-5:
I deleted some recent caches here: https://github.com/pytorch/ignite/actions/caches in case. Let me restart the doctest actions

vfdev-5:
I removed all caches and I also restarted a CI job unit tests on master: https://github.com/pytorch/ignite/actions/runs/22532482914
according to your message: 

> ignite/metrics/gan/fid.py: scipy.linalg.sqrtm(disp=False) no longer returns a tuple in newer scipy versions, causing ValueError: not enough values to unpack. Fixed by removing tuple unpacking and adding an isinstance check for backward compatibility.

it should fail with ValueError?

TahaZahid05:
@vfdev-5 

Sorry if this is getting a bit complicated!

- The `ValueError` for [fid.py](cci:7://file:///home/taha/ignite/ignite/ignite/metrics/gan/fid.py:0:0-0:0) would fail the unit tests, but only if `scipy >= 1.17.0` is installed. Your unit test run passed because it still resolved to an older version of `scipy` (reason mentioned below).

- I confirmed this by checking the logs of this PR and your units tests logs as well. `uv` still decided to install older versions (`scikit-learn==1.7.2` and `scipy==1.15.3`) and passed the CI test. I found this in the action logs here: https://github.com/pytorch/ignite/actions/runs/22548205766/job/65315111403?pr=3616#step:5:245 and https://github.com/pytorch/ignite/actions/runs/22532482914/job/65315371166#step:7:209 

The reason [docs.yml](cci:7://file:///home/taha/ignite/ignite/.github/workflows/docs.yml:0:0-0:0) capped the versions is because it uses `python-version: "3.10"`. Packages like Scikit-learn 1.8.0 officially dropped support for Python 3.10 and require 3.11+, preventing the newest versions from being installed on that CI runner.

We would have to bump [docs.yml](cci:7://file:///home/taha/ignite/ignite/.github/workflows/docs.yml:0:0-0:0) to Python 3.11+ to physically see `make doctest` fail in GitHub Actions.

vfdev-5:
https://github.com/pytorch/ignite/actions/runs/22532482914/job/65315365337#step:7:209

Shows:  scipy==1.17.1

However, the tests show a related warning:
```
2026-03-01T17:47:02.7726573Z   /home/runner/work/ignite/ignite/ignite/metrics/gan/fid.py:40: DeprecationWarning: The `disp` argument is deprecated and will be removed in SciPy 1.18.0.
2026-03-01T17:47:02.7727294Z     covmean, _ = scipy.linalg.sqrtm(sigma1.mm(sigma2).numpy(), disp=False)
2026-03-01T17:47:02.7727562Z 
2026-03-01T17:47:02.7727690Z tests/ignite/metrics/gan/test_fid.py::test_fid_function
2026-03-01T17:47:02.7728091Z tests/ignite/metrics/gan/test_fid.py::test_compute_fid_from_features[cpu]
2026-03-01T17:47:02.7728624Z tests/ignite/metrics/gan/test_fid.py::test_distrib_cpu
2026-03-01T17:47:02.7728957Z tests/ignite/metrics/gan/test_fid.py::test_distrib_cpu
2026-03-01T17:47:02.7729740Z   /home/runner/work/ignite/ignite/.venv/lib/python3.11/site-packages/pytorch_fid/fid_score.py:188: DeprecationWarning: The `disp` argument is deprecated and will be removed in SciPy 1.18.0.
2026-03-01T17:47:02.7730535Z     covmean, _ = linalg.sqrtm(sigma1.dot(sigma2), disp=False)
```

TahaZahid05:
@vfdev-5 Thank you for the observation regarding the `cpu-tests`. I looked into this deeply to figure out why the unit tests only show a `DeprecationWarning` while the `make doctest` hard-fails with `ValueError`.

It turns out we've hit a rather obscure bug inside SciPy 1.17+ regarding their handling of 1x1 matrices.

In SciPy 1.17+, `scipy.linalg.sqrtm(..., disp=False)` is indeed deprecated, but for almost all matrices, it still returns the expected `(result, errest)` tuple. However, they accidentally missed a check in their early-exit logic. If you pass an exact 1x1 matrix into `sqrtm(..., disp=False)`, it hits this specific block in their source code (`scipy.linalg._matfuncs`):

```python
    # Scalar case
    if a.shape[-2:] == (1, 1):
        return np.emath.sqrt(a)
```

Notice that it immediately returns the single scalar array and completely bypasses the `disp=False` logic at the bottom of the function that would normally append the `errest` value.

This perfectly explains the discrepancy we are seeing:
* **Unit Tests (`test_fid.py`)**: Most of these tests use larger dimensions like `torch.rand(10, 10)` or `num_features=10`. Since these are not 1x1, SciPy successfully returns the tuple, the unpacking `covmean, _ = ` works as intended, and Python just prints the `DeprecationWarning`.
* **Doctests (`fid.py`)**: The example in our documentation sets `metric = FID(num_features=1)`. This forces a 1x1 covariance matrix, causing SciPy to hit the buggy early-exit. It wrongly returns a single 1x1 array, and the unpacking `covmean, _ = ` instantly triggers the `ValueError: not enough values to unpack`.

The fix in my PR handles this safely by removing the strict tuple unpacking:
```python
covmean = scipy.linalg.sqrtm(..., disp=False)
if isinstance(covmean, tuple):
    covmean = covmean[0]
```

This acts as a solid adapter for all three scenarios:
1. **SciPy < 1.18 (Normal matrices)**: It correctly handles the tuple.
2. **SciPy < 1.18 (1x1 matrices / doctests)**: It safely handles the SciPy bug without crashing.
3. **SciPy 1.18+ (The Future)**: When `disp` is finalized for removal and `sqrtm` always returns a single matrix, Ignite will continue working smoothly without any code changes needed later on.

vfdev-5:
Thanks for investigations @TahaZahid05 !

> In SciPy 1.17+, scipy.linalg.sqrtm(..., disp=False) is indeed deprecated, but for almost all matrices, it still returns the expected (result, errest) tuple. However, they accidentally missed a check in their early-exit logic.  

We may open an issue in scipy about that? Can you please post a link on the code here.

> The fix in my PR handles this safely by removing the strict tuple unpacking:

This code `covmean = scipy.linalg.sqrtm(..., disp=False)` will fail with 1.18 as disp arg is deprecated and will be removed. We should remove it as well.

vfdev-5:
well the correct line is https://github.com/scipy/scipy/blob/8c75ae75176236f233824e9a0483c26a69e6dfec/scipy/linalg/_matfuncs.py#L526

For empty array they also return an empty array: https://github.com/scipy/scipy/blob/8c75ae75176236f233824e9a0483c26a69e6dfec/scipy/linalg/_matfuncs.py#L522

Not really sure whether this is a bug especially that they deprecate disp arg. Nothing to do on their repo I would say

TahaZahid05:
@vfdev-5 

I have updated PR #3612 to remove the `disp` argument.

Regarding opening an issue, yeah I think since `disp` is being removed anyways, I don't think there is a need to point it out on their repo.

PR #3621, opened 2026-03-02
[fix]: .gitignore to correctly ignore all __pycache__ directories

aaishwarymishra:
Updated the .gitignore to stop tracking empty pycache directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 # Byte-compiled / optimized / DLL files
 *.pyc
-__pycache__/
+*/__pycache__/
 *.py[cod]
 *$py.class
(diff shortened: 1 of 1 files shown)


aaishwarymishra:
There are many __pycache__ directories that needs to be removed from the project.

vfdev-5:
> There are many **pycache** directories that needs to be removed from the project.

Where are they :) 
This is kind of surprising, not sure if it is not a local problem with your setup

aaishwarymishra:
Oh sorry I just noticed it sorry for confusing you , I think it's my local setup .
Sorry again :(

PR #3624, opened 2026-03-03
Feat: implement GradMonitor handler for gradient norm tracking and batch skipping

### LINKED ISSUE #3563: [Feature] Implementation of a TrainingHealthMonitor for real-time gradient norm tracking.
# 🚀 Feature
## Description
While the current suite of Ignite handlers covers a wide range of training needs, monitoring loss alone can be insufficient for detecting early-stage instability in complex models, such as transformers or tasks involving curriculum-based learning.

I would like to propose the addition of a TrainingHealthMonitor utility. In my recent work with logic-based transformer prototypes, I found that tracking the L2 gradient norm at the iteration level was essential for identifying "entropy collapse" and optimization stress during curriculum stage transitions before the model diverged.

## Proposed Implementation
I suggest a handler that hooks into the ITERATION_COMPLETED event:
* Threshold Alerts: Automatically log a warning or trigger a user-defined action if the L2 gradient norm exceeds a specific threshold.

* Optimization Visibility: Provide researchers with a real-time heartbeat of the training process, which is particularly valuable when tuning learning rates or stabilizing curriculum jumps. 

* Ignite Native Integration: Designed to attach seamlessly to the Engine and work alongside existing handlers like the ProgressBar. 
 
## Technical Justification
While safety measures like gradient clipping (as discussed in #375) prevent catastrophic failures, they often mask underlying instability. A dedicated health monitor provides the transparency needed to debug optimization issues proactively.

I am very interested in contributing this feature to the library and would appreciate the maintainers' feedback on this proposal.

leelakrishnaRajasimha:
Closes #3563

# Overview
This PR introduces the GradMonitor handler, designed to track L2 gradient norms in real-time to detect training instability, such as "entropy collapse" or exploding gradients in Transformer models.

## Key Features
* Real-time Monitoring: Hooks into Events.ITERATION_COMPLETED to calculate and log L2 gradient norms.

* AMP & DDP Support: Fully compatible with Automatic Mixed Precision (unscaling via GradScaler) and Distributed Data Parallel (synchronization via idist.all_reduce).

* Reactive Training Logic: Sets an engine.state.unhealthy_spike flag, enabling users to implement automated batch skipping or custom callbacks when thresholds are breached.

* Flexible Thresholding: Supports both static L2 norm limits and dynamic spike detection using a moving average window.

## Performance & Benchmarking
The handler has been benchmarked on a model with 1 million parameters to ensure it does not significantly impact training throughput:

* Local Hardware (HP Victus / RTX 4050): For 100,000 (1 lakh) iterations, the total execution time typically falls within a 9s to 16s range, depending on background system overhead.

* High-Performance Environment (Google Colab GPU): In optimized cloud environments with dedicated high-performance GPUs, the same benchmark has achieved results as fast as 5.4s.

* Efficiency: This demonstrates that the per-iteration overhead remains negligible across various hardware tiers, making it highly suitable for large-scale production models.

## Testing Suite
* Verified mathematical accuracy of L2 norm calculations.

* Validated engine.state flag toggling and callback execution.

* Confirmed correct unscaling behavior when a GradScaler is present.

* ncluded input validation for all initialization parameters.

### DIFF
--- a/ignite/handlers/__init__.py
+++ b/ignite/handlers/__init__.py
@@ -5,6 +5,7 @@ from ignite.handlers.clearml_logger import ClearMLLogger
 from ignite.handlers.early_stopping import EarlyStopping
 from ignite.handlers.ema_handler import EMAHandler
 from ignite.handlers.fbresearch_logger import FBResearchLogger
+from ignite.handlers.grad_monitor import GradMonitor
 from ignite.handlers.lr_finder import FastaiLRFinder
 from ignite.handlers.mlflow_logger import MLflowLogger
 from ignite.handlers.neptune_logger import NeptuneLogger
@@ -50,6 +51,7 @@ __all__ = [
     "Timer",
     "EarlyStopping",
     "TerminateOnNan",
+    "GradMonitor",
     "global_step_from_engine",
     "TimeLimit",
     "EpochOutputStore",
--- a/ignite/handlers/grad_monitor.py
+++ b/ignite/handlers/grad_monitor.py
@@ -1,0 +1,274 @@
+import torch
+from typing import Any, Callable, Optional
+from ignite.engine import Engine, Events
+
+
+def _default_spike_detector(mean: float, m2: float, count: int, norm: float, k: float) -> bool:
+    """
+    Default spike detection rule: flags if norm > mean + k * std.
+
+    Args:
+        mean: Running mean of gradient norms.
+        m2: Running sum of squared deviations (Welford's algorithm).
+        count: Number of iterations seen so far.
+        norm: Current gradient norm.
+        k: Standard deviation multiplier.
+
+    Returns:
+        True if the current norm is a spike, False otherwise.
+    """
+    if count < 2:
+        return False
+    std = (m2 / (count - 1)) ** 0.5
+    return norm > mean + k * std + 1e-6
+
+
+class GradMonitor:
+    """
+    Monitors the L2 gradient norm each iteration to detect training instability
+    (e.g. exploding gradients, entropy collapse) using a dynamic threshold.
+
+    The handler attaches to ``Events.ITERATION_STARTED``, meaning it reads
+    gradients computed during the **previous** iteration. The spike flag is
+    set on ``engine.state.unhealthy_spike`` and can be checked at the start
+    of the next iteration's train step.
+
+    .. warning::
+        This handler requires that gradients are **not zeroed** at the end of
+        your train step. If you call ``optimizer.zero_grad()`` at the end of
+        ``train_step``, all gradients will be gone by the time this handler
+        runs and the norm will always be 0. Instead, call
+        ``optimizer.zero_grad()`` at the **start** of your train step, after
+        checking the spike flag.
+
+    .. warning::
+        The ``unhealthy_spike`` flag on ``engine.state`` reflects gradients
+        from the **previous** iteration, not the current one. Design your
+        train step accordingly.
+
+    .. warning::
+        Call ``attach`` only once per engine instance. Calling it multiple
+        times on the same engine will raise a ``RuntimeError`` to prevent
+        doubled stat updates and corrupted running statistics.
+
+    .. note::
+        If multiple handlers are registered to ``Events.ITERATION_STARTED``,
+        execution order depends on registration order. Register
+        ``GradMonitor`` before any handler that reads ``unhealthy_spike``
+        to ensure the flag is fresh when read.
+
+    Args:
+        model: The model whose gradient norms will be monitored.
+        k: Multiplier for standard deviation when using the default spike
+            detector. Ignored if ``spike_detector`` is provided. Default: 3.0.
+        scaler: Optional ``torch.cuda.amp.GradScaler`` instance for AMP
+            workflows. When provided, the raw gradient norm is divided by
+            the current scale factor to recover the true unscaled norm.
+        spike_detector: Optional callable with signature
+            ``(mean, m2, count, norm, k) -> bool``. If provided, replaces
+            the default ``mean + k * std`` rule entirely. Use this to
+            implement custom thresholding logic.
+
+    Example:
+        .. code-block:: python
+
+            import torch
+            from ignite.engine import Engine, Events
+            from ignite.handlers import GradMonitor
+
+            model = torch.nn.Linear(10, 1)
+            optimizer = torch.optim.SGD(model.parameters(), lr=0.01)
+
+            def train_step(engine, batch):
+                # Check the spike flag set by GradMonitor from the previous
+                # iteration. On iteration 1 this will always be False
+                # (not enough history yet).
+                if engine.state.unhealthy_spike:
+                    # Discard this batch entirely.
+                    # Do NOT zero grads here so GradMonitor can still
+                    # read them on the next iteration.
+                    return {"loss": None, "skipped": True}
+
+                optimizer.zero_grad()  # zero at START, not end
+                x, y = batch
+                loss = ((model(x) - y) ** 2).mean()
+                loss.backward()
+                optimizer.step()
+                # Do NOT call optimizer.zero_grad() here.
+                return {"loss": loss.item(), "skipped": False}
+
+            trainer = Engine(train_step)
+
+            # Attach GradMonitor BEFORE any handler that reads unhealthy_spike
+            # to guarantee correct ordering on Events.ITERATION_STARTED.
+            monitor = GradMonitor(model, k=3.0)
+            monitor.attach(trainer)
+
+            # This handler runs after GradMonitor because it is registered after.
+            @trainer.on(Events.ITERATION_STARTED)
+            def log_spike(engine):
+                if engine.state.unhealthy_spike:
+                    print(f"Spike at iteration {engine.state.iteration}!")
+
+    Example with AMP:
+        .. code-block:: python
+
+            scaler = torch.cuda.amp.GradScaler()
+            monitor = GradMonitor(model, k=3.0, scaler=scaler)
+            monitor.attach(trainer)
+
+    Example with a custom spike detector:
+        .. code-block:: python
+
+            def my_detector(mean, m2, count, norm, k):
+                # Flag only if norm exceeds an absolute limit of 100.
+                return norm > 100.0
+
+            monitor = GradMonitor(model, spike_detector=my_detector)
+            monitor.attach(trainer)
+
+    .. versionadded:: 0.6.0
+    """
+
+    def __init__(
+        self,
+        model: torch.nn.Module,
+        k: float = 3.0,
+        scaler: Optional[Any] = None,
+        spike_detector: Optional[Callable] = None,
+    ):
+        if not isinstance(model, torch.nn.Module):
+            raise TypeError(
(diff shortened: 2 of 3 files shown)


aaishwarymishra:
Hmm, I read your pr there are few suggestions:
1. Instead of using a brittle user defined threshold use dynamic threshold as default.
2. Instead of using brittle `mean*0.5` as threshold we can use `mean + (k*std) ` where `k` can be defined by user and `std` is the standard deviation.
3. Tracking history and window is kinda expensive I think using single variable for iteration_count, mean and variance will be much more efficient.
4. Attaching the handler on event `ITERATION.COMPLETED` is wrong as on iteration.completed the weights are already updated it should be `ITERATION.STARTED`
5. Also you are updating the list before the calculations for mean, it should be after.

Also by looking at code the handler doesn't actually skip the batch user has to write custom code in train step?

like :
```
if not engine.state.unhealthy_spike:
    optimizer.zero_grad()
    return None
```

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,133 @@
...
+        if not engine.has_event_handler(self, Events.ITERATION_COMPLETED):
+            engine.add_event_handler(Events.ITERATION_COMPLETED, self)
+
+    def _compute_grad_norm(self, engine: Optional[Engine] = None) -> float:
+        # Calculates the total L2 norm,Handling DDP synchronization and AMP scaling.
+        total_norm_sq = 0.0
+        for p in self.model.parameters():
+            if p.grad is not None:
+                # Uses .data.norm to get raw scalar value quickly.
+                param_norm = p.grad.detach().data.norm(2)
```
vfdev-5: Why do you need to detach and call data on grad tensor?

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,133 @@
...
+        monitor = GradMonitor(model, threshold=10.0)
+        monitor.attach(trainer)
+
+        # Implementation for batch skipping
+        @trainer.on(Events.ITERATION_COMPLETED)
+        def skip_batch_on_spike(engine):
+            if getattr(engine.state, "unhealthy_spike", False):
+                # Custom logic here, e.g., zeroing grads or logging
+                optimizer.zero_grad()
+                print(f"Spike detected at iteration {engine.state.iteration}, skipping.")
```
vfdev-5: This example is unclear. Let's not use getattr as the attribute should exist.
Second this should be put into train_step.

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,133 @@
...
+__all__ = ["GradMonitor"]
+
+class GradMonitor:
+    """
+    Monitors the L2 gradient norm to catch training instability early.
+    
+    If the norm exceeds the threshold, we log a warning and run an optional callback.
+    This is especially useful for transformers where gradients can explode suddenly.
+
+    .. versionadded:: 0.5.3
```
vfdev-5: This should be in the end of docstring and mention the next unreleased version

vfdev-5 (commented):
Thanks for the draft. I left few quick comments without going yet into details.

leelakrishnaRajasimha:
I have refactored the GradMonitor implementation to fully address the feedback:
Efficiency: Implemented Welford's Algorithm for O(1) memory-efficient running statistics (mean and variance).Performance: Optimized the L2 norm calculation using .pow(2).sum()(avoiding redundant square roots inside the lop) and minimized CPU-GPU synchronization by calling .item() only once per iteration.
Logic: Attached the handler to Events.ITERATION_STARTED to allow for proper batch skipping before weight updates.
Testing: Added unit tests to verify the statistical threshold logic (mean + k \times std) and confirmed that the k multiplier is functioning correctly.
Examples: Updated the docstrings and examples to clarify how users can implement custom batch-skipping logic in their train_step.
All tests are passing locally, and the implementation follows the repository's formatting standards.

ignite/handlers/__init__.py
```diff
@@ -1,5 +1,4 @@
-from collections.abc import Callable
-from typing import Any
+from typing import Any, Callable, Optional
```
aaishwarymishra: why are you removing these and replacing it with typing module and `Optional`?

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(self, model: torch.nn.Module, k: float = 3.0):
+        self.model = model
+        self.k = k
+        self.count = 0
+        self.mean = 0.0
+        self.m2 = 0.0 
+
+    def __call__(self, engine: Engine):
+        device = next(self.model.parameters()).device
```
aaishwarymishra: you could just save the device `self.device`

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,64 @@
...
+
+            monitor = GradMonitor(model)
+            monitor.attach(trainer)  # Attach first so it runs before the check
+
+            @trainer.on(Events.ITERATION_STARTED)
+            def check_spike(engine):
+                # The training step logic must check this flag to skip
+                if getattr(engine.state, "unhealthy_spike", False):
+                    optimizer.zero_grad()
+
```
aaishwarymishra, line 129: This example is not good, as all it does is apply `optimizer.zero_grad()` `train_step` will still run and weighs will get updated, in example there should be a custom train_step with uses the `engine.state.unhealthy_spike`

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,64 @@
...
+    .. versionadded:: 0.6.0
+    """
+
+    def __init__(self, model: torch.nn.Module, k: float = 3.0):
```
aaishwarymishra: didn't this handler  took an optional `callable` which ran in case of spike?

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,64 @@
...
+        total_norm_sq = torch.tensor(0.0, device=device)
+        
+        for p in self.model.parameters():
+            if p.grad is not None:
+                # Optimized: pow(2).sum() is faster than norm()**2.
+                total_norm_sq += p.grad.pow(2).sum()
+
+        total_norm = torch.sqrt(total_norm_sq).item()
+        
+        scaler = getattr(engine.state, "scaler", None)
```
aaishwarymishra: @vfdev-5  can you check it I am not that much familiar with distributed.

leelakrishnaRajasimha:
Thanks, @aaishwarymishra , I have updated the code based on your feedback. I optimized the device handling to avoid redundant calls and refactored the example to show a custom_train_step for proper batch skipping. I kept the typing module for now to stay consistent with the existing library style. Let me know if anything else needs adjusting.

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,76 @@
...
+                    # Skip the entire forward/backward pass for this batch
+                    return {"skipped": True}
+                
+                # Normal training logic.
+                model.train()
+                optimizer.zero_grad()
+                # ... forward/backward/step ...
+                return {"loss": loss.item()}
+
+            trainer = Engine(custom_train_step)
```
aaishwarymishra: You are basically creating the trainer with custom `train_step` after the monitor has been attached to the trainer.
The flow is create trainer -> attach.

ignite/handlers/__init__.py
```diff
@@ -1,10 +1,12 @@
...
```
vfdev-5: Why do we need this line here?

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,76 @@
...
+
+    .. versionadded:: 0.6.0
+    """
+
+    def __init__(self, model: torch.nn.Module, k: float = 3.0):
```
vfdev-5: @leelakrishnaRajasimha I suggest to make configurable the algorithm computing the threshold.
Right now the rule is hard-coded, we can code this as a helper function and users that would like to use their own rule can easily override the default rule.

ignite/handlers/grad_monitor.py
vfdev-5: it is fine. However, it is written nowhere that this handler can look as engine.state.scaler attribute. Do we really need it as dynamic? Can't we define it as a static const in the constructor? In any case, I prefer to recode this part

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,76 @@
...
+            engine.state.unhealthy_spike = total_norm > threshold
+        else:
+            engine.state.unhealthy_spike = False
+
+        self.count += 1
+        delta = total_norm - self.mean
+        self.mean += delta / self.count
+        self.m2 += delta * (total_norm - self.mean)
+
+    def attach(self, engine: Engine):
```
vfdev-5: if user's train step looks like that:
```python
    ...
    optimizer.step()
    optimizer.zero_grad()
    return loss_value
```
so, the handler wont see any grads in the model and silently wont do anything.

vfdev-5:
@leelakrishnaRajasimha will you still work on this PR?

leelakrishnaRajasimha:
@vfdev-5 
Yes, I am still working on this PR. I am currently refactoring the code to address your feedback regarding the configurable threshold algorithm and the handling of the engine.state.scaler attribute. I thought it would be better to update the code first before replying to the individual comments. I will push the changes soon.

leelakrishnaRajasimha:
Hi @vfdev-5, I have refactored the GradMonitor implementation to use Welford’s online algorithm for improved numerical stability. I also added comprehensive unit tests. The PR is now ready for a final review. Thank you.

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,274 @@
...
+            engine: The Ignite training engine.
+
+        Returns:
+            self, to allow fluent chaining:
+            ``GradMonitor(model).attach(trainer)``.
+        """
+        if self._attached:
+            raise RuntimeError(
+                "GradMonitor is already attached to an engine. "
+                "Create a new GradMonitor instance to attach to a different engine."
```
aaishwarymishra, line 262: I think instead of throwing error, giving user warning will be fine.

ignite/handlers/grad_monitor.py
```diff
@@ -0,0 +1,274 @@
...
+        self.mean = 0.0
+        self.m2 = 0.0
+
+    def _get_device(self) -> torch.device:
+        if self._device is None:
+            try:
+                self._device = next(self.model.parameters()).device
+            except StopIteration:
+                self._device = torch.device("cpu")
+        return self._device
```
aaishwarymishra, line 169: I thin this could have been simpler with just 
```device = model.device```
But its fine ig

aaishwarymishra:
@vfdev-5 can you check the distributed and scaler code I am not that much familiar with it.

Also I am not sure if its feasible for handler to do this task, like the handler can be attached at iteration start or iteration end, during iteration start we don't know gradients as its calculated at train step, at iteration end we have the gradients but it is for the previous step not current.

This functionality, should be after loss calculation and before the optimizer step, so the best place for it is inside the train step itself, not a handler.

what you think?

PR #3627, opened 2026-03-03
Add fairness and bias metrics

### LINKED ISSUE #3617: Add fairness and bias metrics
## 🚀 Feature

**Description**
I propose adding fairness metrics to `ignite.metrics` (or creating a dedicated `ignite.metrics.fairness` module). 

These metrics would accept [(y_pred, y, group_labels)](cci:1://file:///home/taha/ignite/ignite/ignite/metrics/metric.py:797:8-798:52) as input during the [update](cci:1://file:///home/taha/ignite/ignite/ignite/metrics/metric.py:402:4-412:12) step. A simple example could be `SubgroupAccuracyDifference`:

```python

class SubgroupAccuracyDifference(Metric):
    # Expects predictions, targets, and the subgroup identifier
    required_output_keys = ("y_pred", "y", "group_labels")

    def reset(self):
        self._correct_per_group = {}
        self._total_per_group = {}

    def update(self, output):
        y_pred, y, groups = output[0].detach(), output[1].detach(), output[2].detach()
        # Segment predictions into group buckets...

    def compute(self):
        # Calculate metric per bucket, then return max difference between any two groups...
```

Other metrics in this suite could include:
- `DemographicParityDifference`: Measures the difference in selection rates (positive predictions) between different demographic groups, regardless of the true label.
- `EqualizedOddsDifference`: Measures the maximum difference in true positive rates or false positive rates between different subgroups.
- `MaxSubgroupError`: Returns the worst-performing metric value across any monitored subgroup (useful for Minimax fairness).

**Motivation**
Currently, Ignite has a ton of great classification and regression metrics, but it doesn't seem to have any metrics designed for evaluating algorithmic fairness or subgroup bias (like measuring accuracy differences between demographic groups or data sources). 

In our recent project (SemEval 2026), my team had to track a complex metric called "content effect bias." This required evaluating our model across two overlapping subgroups simultaneously: the logical "validity" of a statement, and the real-world "plausibility" of that statement. 

To compute the final metric, we had to slice the predictions into four intersectional subgroups (Valid-Plausible, Valid-Implausible, Invalid-Plausible, Invalid-Implausible) and calculate the difference in both accuracy and confidence scores *between* those specific slices. 

Because we needed to dynamically map predictions to these specific subgroups during the validation epoch, we found that we needed to write custom tracking logic. 

Evaluating fairness and subgroup disparities like this is becoming a standard requirement for many ML pipelines. Currently, users have to log all predictions and sensitive features to evaluate disparity metrics using secondary libraries like `fairlearn` (e.g., `fairlearn.metrics.demographic_parity_difference`) after the epoch ends. Adding these directly to Ignite would be highly beneficial.

**Alternatives**
Calculating these offline with libraries like `fairlearn` (which specializes in metrics requiring a `sensitive_features` argument) or custom `scikit-learn` wrappers after pulling all the predictions out of the engine. However, this breaks the real-time logging (e.g., to TensorBoard/WandB) that makes Ignite's event system so useful.

**Additional Context**
I already have working PyTorch implementations for some of these from our project. I'd be happy to adapt them to the `ignite.metrics.Metric` standard, add the DDP support, and open a PR if interested!

TahaZahid05:
Fixes #3617 

Description:
Initial implementation of algorithmic fairness and subgroup bias metrics in `ignite.metrics.fairness`. This PR adds a base class for fairness evaluations and two concrete metrics:

1. `SubgroupAccuracyDifference`: Measures the maximum accuracy disparity across different subgroups.
2. `DemographicParityDifference`: Measures the maximum disparity in selection rates (positive predictions) across subgroups.

Examples:
```python
from ignite.metrics.fairness import SubgroupAccuracyDifference, DemographicParityDifference

metric_acc = SubgroupAccuracyDifference()
metric_acc.update((y_pred, y, groups))
disparity_acc = metric_acc.compute()

metric_dp = DemographicParityDifference()
# Targets 'y' are required by API but ignored for Parity
metric_dp.update((y_pred, y, groups))
disparity_dp = metric_dp.compute()

```

Check list:

### DIFF
--- a/ignite/metrics/fairness/base.py
+++ b/ignite/metrics/fairness/base.py
@@ -1,5 +1,6 @@
 import copy
-from collections.abc import Callable, Sequence
+from collections import OrderedDict
+from collections.abc import Callable, Mapping, Sequence
 from typing import Any
 
 import torch
@@ -51,6 +52,24 @@ class _SubgroupBase(Metric):
             except NotComputableError:
                 continue
         return results
+
+    def state_dict(self) -> OrderedDict:
+        state_dict = super().state_dict()
+        state_dict["_metrics"] = {str(g): m.state_dict() for g, m in self._metrics.items()}
+        return state_dict
+
+    def load_state_dict(self, state_dict: Mapping) -> None:
+        state_dict_dict = dict(state_dict)
+        metrics_state = state_dict_dict.pop("_metrics")
+        super().load_state_dict(state_dict_dict)
+        for g, m_state in metrics_state.items():
+            found_g = None
+            for original_g in self._metrics.keys():
+                if str(original_g) == g:
+                    found_g = original_g
+                    break
+            if found_g is not None:
+                self._metrics[found_g].load_state_dict(m_state)
 
 
 class SubgroupMetric(_SubgroupBase):
--- a/ignite/metrics/fairness/demographic_parity.py
+++ b/ignite/metrics/fairness/demographic_parity.py
@@ -3,6 +3,7 @@ from collections.abc import Callable, Sequence
 
 from ignite.exceptions import NotComputableError
 from ignite.metrics.accuracy import _BaseClassification
+from ignite.metrics.metric import sync_all_reduce
 from ignite.metrics.fairness.base import SubgroupDifference
 
 __all__ = ["DemographicParityDifference", "SelectionRate"]
@@ -41,6 +42,8 @@ class SelectionRate(_BaseClassification):
 
     .. versionadded:: 0.6.0
     """
+
+    _state_dict_all_req_keys = ("_num_positives", "_num_examples")
 
     def __init__(
         self,
@@ -88,6 +91,7 @@ class SelectionRate(_BaseClassification):
             self._num_positives = self._num_positives + positives.to(self._device)
         self._num_examples += total
 
+    @sync_all_reduce("_num_examples", "_num_positives")
     def compute(self) -> torch.Tensor:
         """Computes the selection rate.
 
--- a/tests/ignite/metrics/fairness/test_accuracy_difference.py
+++ b/tests/ignite/metrics/fairness/test_accuracy_difference.py
@@ -138,7 +138,10 @@ def test_compare_accuracy_difference_with_fairlearn(available_device) -> None:
     # Fairlearn computation
     # Fairlearn's MetricFrame takes numpy arrays
     mf = MetricFrame(
-        metrics=accuracy_score, y_true=y_true.numpy(), y_pred=y_pred.numpy(), sensitive_features=group_labels.numpy()
+        metrics=accuracy_score,
+        y_true=y_true.cpu().numpy(),
+        y_pred=y_pred.cpu().numpy(),
+        sensitive_features=group_labels.cpu().numpy(),
     )
     fairlearn_result = mf.difference()
 
--- a/tests/ignite/metrics/fairness/test_demographic_parity.py
+++ b/tests/ignite/metrics/fairness/test_demographic_parity.py
@@ -116,12 +116,12 @@ def test_compare_demographic_parity_with_fairlearn(available_device) -> None:
     # Fairlearn verification for multiclass
     # We calculate demographic parity difference for each class independently (one-vs-rest)
     # and take the max, which is how Ignite's SubgroupDifference handles vector outputs.
-    y_pred_classes = torch.argmax(y_pred_probs, dim=1).numpy()
+    y_pred_classes = torch.argmax(y_pred_probs, dim=1).cpu().numpy()
     fairlearn_max_diff = 0.0
     for c in range(3):
         y_pred_bin = (y_pred_classes == c).astype(int)
         diff = demographic_parity_difference(
-            y_true=y_true.numpy(), y_pred=y_pred_bin, sensitive_features=group_labels.numpy()
+            y_true=y_true.cpu().numpy(), y_pred=y_pred_bin, sensitive_features=group_labels.cpu().numpy()
         )
         fairlearn_max_diff = max(fairlearn_max_diff, diff)
 
@@ -136,7 +136,9 @@ def test_compare_demographic_parity_with_fairlearn(available_device) -> None:
     ignite_res_bin = ignite_metric.compute()
 
     fairlearn_res_bin = demographic_parity_difference(
-        y_true=y_true.numpy(), y_pred=y_pred_binary.numpy(), sensitive_features=group_labels_binary.numpy()
+        y_true=y_true.cpu().numpy(),
+        y_pred=y_pred_binary.cpu().numpy(),
+        sensitive_features=group_labels_binary.cpu().numpy(),
     )
 
     assert_close(ignite_res_bin, float(fairlearn_res_bin))
@@ -299,3 +301,33 @@ def test_demographic_parity_type_switch_validation() -> None:
     metric.update((torch.randn(4, 3), torch.randint(0, 3, (4,)), torch.zeros(4)))
     with pytest.raises(ValueError, match="Input data number of classes has changed"):
         metric.update((torch.randn(4, 5), torch.randint(0, 5, (4,)), torch.zeros(4)))
+
+
+def test_demographic_parity_difference_checkpointing() -> None:
+    """Verifies that the state of SubgroupDifference is correctly preserved across state_dict calls."""
+    metric = DemographicParityDifference(groups=[0, 1])
+
+    # 1. Update with some data for Group 0
+    y_pred_0 = torch.tensor([[0.1, 0.9], [0.1, 0.9]])
+    y_true_0 = torch.tensor([1, 1])
+    groups_0 = torch.tensor([0, 0])
+    metric.update((y_pred_0, y_true_0, groups_0))
+
+    # 2. Save the state
+    state = metric.state_dict()
+
+    # 3. Create a NEW metric instance and load the state
+    new_metric = DemographicParityDifference(groups=[0, 1])
+    new_metric.load_state_dict(state)
+
+    # 4. Update the NEW metric with data for Group 1 only
+    y_pred_1 = torch.tensor([[0.9, 0.1], [0.9, 0.1]])
+    y_true_1 = torch.tensor([1, 1])
+    groups_1 = torch.tensor([1, 1])
+    new_metric.update((y_pred_1, y_true_1, groups_1))
+
+    # 5. Compute the result on the new metric
+    # The new metric should have 1.0 selection rate for Group 0 (from state)
+    # and 0.0 selection rate for Group 1 (from new update).
+    # Result: 1.0 - 0.0 = 1.0
+    assert new_metric.compute() == 1.0


ignite/metrics/fairness/accuracy_difference.py
```diff
@@ -0,0 +1,101 @@
...
+            correct = torch.eq(indices, y).view(-1)
+        elif self._type == "multilabel":
+            # Multilabel: Each label for each example is treated as a separate task to provide
+            # a granular view of disparities across individual classification boundaries.
+            correct = torch.eq(y_pred.type_as(y), y)
+        else:
+            raise ValueError(f"Unexpected type: {self._type}")
+
+        correct_sum = torch.sum(correct).item()
+        total = correct.numel()
```
vfdev-5: Can we reuse Accuracy code here?

ignite/metrics/fairness/accuracy_difference.py
```diff
@@ -0,0 +1,101 @@
+import torch
+from typing import Callable
+from ignite.metrics.fairness.base import _BaseFairness
+
+__all__ = ["SubgroupAccuracyDifference"]
+
+
+class SubgroupAccuracyDifference(_BaseFairness):
+    r"""Calculates the Subgroup Accuracy Difference.
```
vfdev-5, line 10: Can you put a reference on a paper or a resource to read about this metric?
Also do we have a reference implementation for it in sklearn or other known package?

ignite/metrics/fairness/demographic_parity.py
```diff
@@ -0,0 +1,166 @@
...
+from typing import Callable
+
+from ignite.exceptions import NotComputableError
+from ignite.metrics.fairness.base import _BaseFairness
+
+__all__ = ["DemographicParityDifference"]
+
+
+class DemographicParityDifference(_BaseFairness):
+    r"""Calculates the Demographic Parity Difference.
```
vfdev-5, line 107: Same question to this metric as for accuracy difference

ignite/metrics/fairness/accuracy_difference.py
TahaZahid05: Done. I have refactored to reuse `Accuracy` metric instead.

ignite/metrics/fairness/accuracy_difference.py
TahaZahid05: I've included an inline reference in the docstring to Verma & Rubin (2018), *Fairness Definitions Explained*, where this is defined as **Overall Accuracy Equality**.

For a reference implementation, this code aligns with how **Fairlearn** approaches it using [`MetricFrame.difference()`](https://fairlearn.org/main/api_reference/generated/fairlearn.metrics.MetricFrame.html#fairlearn.metrics.MetricFrame.difference) paired with `accuracy_score`.

ignite/metrics/fairness/demographic_parity.py
TahaZahid05: I've referenced the same Verma & Rubin (2018) paper in the docstring, where this is defined as **Group Fairness / Statistical Parity**.

For a reference implementation, this logic corresponds directly to Fairlearn's [`demographic_parity_difference()`](https://fairlearn.org/main/api_reference/generated/fairlearn.metrics.demographic_parity_difference.html).

ignite/metrics/fairness/base.py
```diff
@@ -0,0 +1,237 @@
...
+        self._type: str | None = None
+        self._num_classes: int | None = None
+        super().__init__(output_transform=output_transform, device=device)
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        """Resets the metric state."""
+        self._y_preds: list[torch.Tensor] = []
+        self._ys: list[torch.Tensor] = []
+        self._groups: list[torch.Tensor] = []
```
vfdev-5: I have an impression that we can recode the computation of both metrics without storing the whole history. The set of unique groups is dynamic, but we could make it static by asking user to provide the set?  
And ideally if we could avoid code duplication it would be better.

ignite/metrics/fairness/accuracy_difference.py
```diff
@@ -0,0 +1,94 @@
...
+            y: targets for the specific subgroup.
+
+        Returns:
+            The computed accuracy for the subgroup.
+        """
+        acc = Accuracy(is_multilabel=self._is_multilabel, device=self._device)
+        acc._type = self._type
+        acc._num_classes = self._num_classes
+        acc.update((y_pred, y))
+        return acc.compute()
```
vfdev-5: This looks like a composition, we may also put also Accuracy instance into self.acc and reset on the call of compute_metric_for_group.
Feels like it could be nice to have functional accuracy method.

ignite/metrics/fairness/base.py
TahaZahid05: Done, I have replaced the computation to accumulate. I have also avoided code duplication in `demographic_parity.py` to make the interface class more generalized so that the class design only requires a new metric to define the `_update_group` function

ignite/metrics/fairness/accuracy_difference.py
TahaZahid05: Done. I defined `self.acc` instead.

ignite/metrics/fairness/accuracy_difference.py
Prathamesh8989, line 1: Nice implementation! I had a small question regarding the use of internal attributes from `Accuracy`.

In `_update_group`, the code accesses and modifies several private fields such as `_type`, `_num_classes`, `_num_correct`, and `_num_examples`. Since these are internal implementation details of `Accuracy`, there might be a risk of breakage if the `Accuracy` metric changes internally in the future.

Would it be safer to rely only on the public interface (e.g., using `compute()` or a functional accuracy helper) rather than accessing the private state directly? Just wondering if this could make the implementation more robust long-term.

TahaZahid05: Good point. We can't use `compute()` since we need absolute counts. There doesn't currently exist a helper function that can be used for metrics. We can implement this by extracting core correctness and accumulation logic found inside `Accuracy` into a helper function (like `_accuracy_compute_fn`) and update both `Accuracy.update()` and `SubgroupAccuracyDifference._update_group()` to call that function.

@vfdev-5 kindly let me know if this makes sense. I'll commit to the PR then.

vfdev-5: I think it is fine. If in future Accuracy change some private methods, we'll see it in the CI as this metrics will break, this way we'll be able to update both.

ignite/metrics/fairness/accuracy_difference.py
vfdev-5: Can we use fairlearn in the CI?

ignite/metrics/fairness/accuracy_difference.py
```diff
@@ -0,0 +1,91 @@
...
+    ) -> None:
+        self._is_multilabel = is_multilabel
+        super().__init__(groups=groups, output_transform=output_transform, device=device)
+        self.acc = Accuracy(is_multilabel=is_multilabel, device=device)
+
+    def _update_group(self, group: int, y_pred: torch.Tensor, y: torch.Tensor) -> None:
+        """Updates per-group accuracy accumulators by reusing :class:`~ignite.metrics.Accuracy`."""
+        self.acc.reset()
+        self.acc._type = self._type
+        self.acc._num_classes = self._num_classes
```
vfdev-5: Can this two lines be defined in the constructor?

ignite/metrics/fairness/accuracy_difference.py
```diff
@@ -0,0 +1,91 @@
...
+        self.acc.update((y_pred, y))
+        if group not in self._group_numerator:
+            self._group_numerator[group] = torch.tensor(0.0, device=self._device)
```
vfdev-5: Can this be defined in the SubgroupAccuracyDifference.reset or its base class?

ignite/metrics/fairness/base.py
```diff
@@ -0,0 +1,244 @@
...
+        super().__init__(output_transform=output_transform, device=device)
+
+    @reinit__is_reduced
+    def reset(self) -> None:
+        """Resets the metric state."""
+        self._type = None
+        self._num_classes = None
+        self._updated = False
+        self._group_numerator: dict[int, torch.Tensor] = {}
+        self._group_denominator: dict[int, int] = {g: 0 for g in self._user_groups}
```
vfdev-5: Why do we define differently numerator and denominator?

ignite/metrics/fairness/accuracy_difference.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(
+        self,
+        groups: Sequence[int],
+        is_multilabel: bool = False,
+        output_transform: Callable = lambda x: x,
+        device: torch.device | str = torch.device("cpu"),
+    ) -> None:
+        self._is_multilabel = is_multilabel
+        super().__init__(groups=groups, output_transform=output_transform, device=device)
+        self.acc = Accuracy(is_multilabel=is_multilabel, device=device)
```
vfdev-5: Seeing this implementation, I have another impression that we can define a sort metric per group class where we define what base metric is used to compute per group. For SubgroupAccuracyDifference it would be Accuracy. So, we can even have few classes like:

Instead of defining a dict of group numerator and group denominator we could define a dict of base metrics. On update we would update base metrics per group. This way we let the base metric its business: check shapes, dtypes etc and we do not need to handle distributed sync part neither. 

- SubgroupMetric : defining a dict of base metrics
- SubgroupDifferenceMetric derived from SubgroupMetric and doing the difference

To implement DemographicParityDifference, we may need to introduce a base metric computing the rate of positive predictions.

vfdev-5 (commented):
Thanks for the updates @TahaZahid05 !

ignite/metrics/fairness/accuracy_difference.py
TahaZahid05: Thanks for the direction. I changed a couple things. I added `_SubgroupBase` as the dict of base metrics. The reason for doing this was that `SubgroupMetric` itself has a compute function that returns a dictionary, which differs from `SubgroupDifference`. Therefore, it made more sense to break it down even further.

For `DemographicParityDifference`, I added `SelectionRate` as suggested.

ignite/metrics/fairness/base.py
TahaZahid05: Solved through a change in implementation

ignite/metrics/fairness/accuracy_difference.py
TahaZahid05: Solved through a change in implementation

ignite/metrics/fairness/accuracy_difference.py
TahaZahid05: Solved through a change in implementation

ignite/metrics/fairness/accuracy_difference.py
TahaZahid05: I added some test comparing with `fairlearn` to make sure our results match.

ignite/metrics/fairness/accuracy_difference.py
```diff
@@ -0,0 +1,91 @@
...
+        self,
+        groups: Sequence[int],
+        is_multilabel: bool = False,
+        output_transform: Callable = lambda x: x,
+        device: torch.device | str = torch.device("cpu"),
+    ) -> None:
+        acc = Accuracy(is_multilabel=is_multilabel, device=device)
+        super().__init__(base_metric=acc, groups=groups, output_transform=output_transform, device=device)
+
+    def compute(self) -> float:
```
vfdev-5: @TahaZahid05 was this method override just for the docstring? In this case why not exposing `update` method?

ignite/metrics/fairness/accuracy_difference.py
```diff
@@ -0,0 +1,91 @@
+import torch
+from typing import Callable, Sequence
```
vfdev-5: Let's use the modern typing in this PR as we did in other submodules

ignite/metrics/fairness/base.py
```diff
@@ -0,0 +1,120 @@
+import copy
+from typing import Any, Callable, Dict, Sequence
```
vfdev-5: Same here about the typing hints

ignite/metrics/fairness/demographic_parity.py
```diff
@@ -0,0 +1,166 @@
...
+        device: str | torch.device = torch.device("cpu"),
+        skip_unrolling: bool = False,
+    ):
+        self._num_positives: torch.Tensor | None = None
+        super().__init__(
+            output_transform=output_transform, is_multilabel=is_multilabel, device=device, skip_unrolling=skip_unrolling
+        )
+
+    def reset(self) -> None:
+        self._num_positives: torch.Tensor | None = None
```
vfdev-5: Why it can't be 0 at reset instead of None?

TahaZahid05: Right, fixed.

ignite/metrics/fairness/base.py
TahaZahid05: Fixed!

ignite/metrics/fairness/accuracy_difference.py
TahaZahid05: @vfdev-5 True, what if we remove the compte override method as well? It is not adding new information from the docstring. 

I have followed this approach in the new commit. let me know if you want me to rather add `update` instead.

tests/ignite/metrics/fairness/test_demographic_parity.py
```diff
@@ -0,0 +1,125 @@
...
+def test_distrib_cpu(distributed_context_single_node_gloo: Callable) -> None:
+    """Tests distributed integration on CPU."""
+    device = torch.device("cpu")
+    _test_distrib_integration(device)
+
+
+@pytest.mark.distributed
+@pytest.mark.skipif(not idist.has_native_dist_support, reason="Skip if no native dist support")
+@pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Skip if no GPU")
+def test_distrib_gpu(distributed_context_single_node_nccl: Callable) -> None:
```
vfdev-5: Let's write distributed tests following this new formalism: https://github.com/pytorch/ignite/blob/9ed171c062adccf98da691ecdcf59cd0a68043ed/tests/ignite/metrics/test_accuracy.py#L195-L198

tests/ignite/metrics/fairness/test_demographic_parity.py
```diff
@@ -0,0 +1,125 @@
+import pytest
```
vfdev-5, line 1: Should we have also SelectionRate metric tests as well?

tests/ignite/metrics/fairness/test_fairlearn_comparison.py
```diff
@@ -0,0 +1,89 @@
+import pytest
```
vfdev-5: Let's put these tests into appropriate ignite metric test files. The correctness checks should be put directly there instead of additional file

tests/ignite/metrics/fairness/test_fairlearn_comparison.py
```diff
@@ -0,0 +1,89 @@
...
+import torch
+from torch.testing import assert_close
+
+# Skip if fairlearn is not installed
+pytest.importorskip("fairlearn")
```
vfdev-5: This should not be done this way

TahaZahid05: Yeah, apologies. I removed it. we already have it in `requirements-dev.txt`.

tests/ignite/metrics/fairness/test_accuracy_difference.py
```diff
@@ -0,0 +1,272 @@
...
+    y = torch.tensor([0, 0])
+    group_labels = torch.tensor([0, 0])
+
+    metric.update((y_pred, y, group_labels))
+
+    with pytest.raises(NotComputableError, match="Fairness metrics require at least two unique subgroups"):
+        metric.compute()
+
+
+def test_subgroup_accuracy_difference_binary_labels() -> None:
```
vfdev-5: Can we add available_device fixture to check the computation code on different devices to ensure no bugs in all metric tests (except those checking errors or distributed)

TahaZahid05:
@vfdev-5 Done!

vfdev-5:
The failure however is real: https://github.com/pytorch/ignite/actions/runs/23442545839/job/68197243478?pr=3627#step:10:18193

TahaZahid05:
Yeah, @vfdev-5 i'll fix it.

PR #3628, opened 2026-03-03
Add optional `show_epoch` flag to ProgressBar for customizable epoch display

### LINKED ISSUE #3659: Feature Request: Option to disable automatic epoch display in ProgressBar
## Problem

Currently ignite.handlers.ProgressBar automatically appends epoch
information like [epoch/max_epochs] to the description when max_epochs > 1.

Example:

Training [1/4]
Training [2/4]

This behavior can make it difficult to align multiple progress bars
(e.g. trainer and evaluator) or customize the display layout.

## Current Workaround

Users can redefine the entire `bar_format`, but this requires rewriting
the full format string even if they only want to remove the epoch display.

Also, removing `{desc}` from `bar_format` disables the dynamic `desc`
argument from tqdm.

## Proposed Solution

Allow users to control how the description is formatted.

Possible approaches:

1. `show_epoch` boolean flag
2. `description_format` parameter

Example:

```python
ProgressBar(desc="Training", show_epoch=False)

### PR DESCRIPTION
**Pull request by SpandanBhoiIITM**
Closes #3659

## Description

This PR adds a `show_epoch` parameter to `ignite.handlers.tqdm_logger.ProgressBar`.

Currently, epoch information (e.g. "[5/10]") is automatically appended to the
progress bar description when `max_epochs > 1`. This behavior makes it
difficult to align multiple progress bars (e.g. trainer and evaluator) with
custom formatting.

The new `show_epoch` flag (default=True) allows users to disable automatic
epoch display while preserving backward compatibility.

All existing behavior remains unchanged by default.

## Checklist

- [ ] New tests are added (can be added if required)
- [x] New docstrings are updated
- [ ] Documentation update required

### DIFF
--- a/ignite/handlers/tqdm_logger.py
+++ b/ignite/handlers/tqdm_logger.py
@@ -117,13 +117,12 @@ class ProgressBar(BaseLogger):
     ]
 
     def __init__(
-        self,
-        persist: bool = False,
-        bar_format: (
-            str | None
-        ) = "{desc}[{n_fmt}/{total_fmt}] {percentage:3.0f}%|{bar}{postfix} [{elapsed}<{remaining}]",
-        **tqdm_kwargs: Any,
-    ):
+    self,
+    persist: bool = False,
+    bar_format: str | None = "{desc}[{n_fmt}/{total_fmt}] {percentage:3.0f}%|{bar}{postfix} [{elapsed}<{remaining}]",
+    show_epoch: bool = True,
+    **tqdm_kwargs: Any,
+):
         try:
             from tqdm.autonotebook import tqdm
         except ImportError:
@@ -136,6 +135,7 @@ class ProgressBar(BaseLogger):
         self.pbar = None
         self.persist = persist
         self.bar_format = bar_format
+        self.show_epoch = show_epoch
         self.tqdm_kwargs = tqdm_kwargs
 
     def _reset(self, pbar_total: int | None) -> None:
@@ -212,12 +212,13 @@ class ProgressBar(BaseLogger):
             raise ValueError(f"Logging event {event_name} should be called before closing event {closing_event_name}")
 
         log_handler = _OutputHandler(
-            desc,
-            metric_names,
-            output_transform,
-            closing_event_name=closing_event_name,
-            state_attributes=state_attributes,
-        )
+    desc,
+    metric_names,
+    output_transform,
+    closing_event_name=closing_event_name,
+    state_attributes=state_attributes,
+    show_epoch=self.show_epoch,
+)
 
         super(ProgressBar, self).attach(engine, log_handler, event_name)
         engine.add_event_handler(closing_event_name, self._close)
@@ -261,13 +262,14 @@ class _OutputHandler(BaseOutputHandler):
     """
 
     def __init__(
-        self,
-        description: str,
-        metric_names: str | list[str] | None = None,
-        output_transform: Callable | None = None,
-        closing_event_name: Events | CallableEventWithFilter = Events.EPOCH_COMPLETED,
-        state_attributes: list[str] | None = None,
-    ):
+    self,
+    description: str,
+    metric_names: str | list[str] | None = None,
+    output_transform: Callable | None = None,
+    closing_event_name: Events | CallableEventWithFilter = Events.EPOCH_COMPLETED,
+    state_attributes: list[str] | None = None,
+    show_epoch: bool = True,
+):
         if metric_names is None and output_transform is None:
             # This helps to avoid 'Either metric_names or output_transform should be defined' of BaseOutputHandler
             metric_names = []
@@ -275,6 +277,7 @@ class _OutputHandler(BaseOutputHandler):
             description, metric_names, output_transform, global_step_transform=None, state_attributes=state_attributes
         )
         self.closing_event_name = closing_event_name
+        self.show_epoch = show_epoch
 
     @staticmethod
     def get_max_number_events(event_name: str | Events | CallableEventWithFilter, engine: Engine) -> int | None:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by TahaZahid05** [comment]
@SpandanBhoiIITM Thanks for the contribution. 

1. Could you please ensure your PR aligns with the code style, documentation, and test requirements mentioned in our [CONTRIBUTING.md](https://github.com/pytorch/ignite/blob/master/CONTRIBUTING.md). This includes fixing the indentation (running `black` or `pre-commit`), adding docstrings, and including a unit test. Also, please replace {issue_number} in the description with the actual issue ID by opening an issue.
2. Instead of a specific show_epoch boolean, would it be possible to make this more general? We could use a `description_format` parameter (defaulting to `"{desc} [{epoch}/{max_epochs}]"`). This would allow users to remove the epoch by setting it to `"{desc}"`, while also supporting other custom formats without needing more boolean flags in the future.

**Comment by vfdev-5** [comment]
@SpandanBhoiIITM @TahaZahid05 I suggest to discuss the problem before reviewing the code change. We should not introduce alternative API if already exists a solution to the problem.

**Comment by TahaZahid05** [comment]
@vfdev-5.
Right now, the only way a user can hide the epoch is by creating a completely custom `bar_format` string and deleting the `{desc}` placeholder from it entirely.
  
  
However, doing this is flawed because it breaks the progress bar's ability to have a title. If a user removes the `{desc}` placeholder, `tqdm` completely ignores the `desc="..."` argument. Users are forced to permanently type their title directly into the formatting string itself (like `bar_format="Training | {bar}... "`). If they do this, standard features `ProgressBar(desc="Training")` will silently stop working, and they can no longer update the title while the script is running.
  
@SpandanBhoiIITM  Given this, rather than introducing a strict `show_epoch` boolean, I believe we should think about a `description_format` string parameter.
  
For instance, Users can pass `description_format="{desc}"` to drop the epoch string. This gives us customizability without needing more boolean flags down the line.

What do you think?

**Comment by vfdev-5** [comment]
> Currently, epoch information (e.g. "[5/10]") is automatically appended to the
progress bar description when max_epochs > 1. This behavior makes it
difficult to align multiple progress bars (e.g. trainer and evaluator) with
custom formatting.

@SpandanBhoiIITM can provide a small reproducer of what you are trying to achieve to see explicitly what does not work as desired. Thanks! 



For example, this code:
```python
from ignite.handlers import ProgressBar


train_data = range(100)
eval_data = range(40)
max_epochs = 10


    import time

    time.sleep(0.01)
    return None

trainer = Engine(train_step)
evaluator = Engine(lambda e, b: None)


@trainer.on(Events.EPOCH_COMPLETED(every=2))
def run_validation():
    evaluator.run(eval_data)

ProgressBar(persist=True, desc="Training").attach(trainer)
ProgressBar(persist=True, desc="Validation").attach(evaluator)

trainer.run(train_data, max_epochs=max_epochs)
```
gives the following representation in Jupyter notebook
image


and if we are bothered that progress bars are not aligned we can redefine the bar format as we would like:
```python
bformat = "{percentage:3.0f}%|{bar}{postfix} [{elapsed}<{remaining}] - {desc}[{n_fmt}/{total_fmt}]"
ProgressBar(persist=True, bar_format=bformat, desc="Training").attach(trainer)
ProgressBar(persist=True, bar_format=bformat, desc="Validation").attach(evaluator)
```
image

SpandanBhoiIITM:
@vfdev-5 @TahaZahid05 Thanks for the feedback!

My motivation for adding `show_epoch` was to make it easier to hide the epoch
information while still keeping the `desc` argument usable.

Right now the only way to remove the epoch is by rewriting the whole
`bar_format`. But if `{desc}` is removed from `bar_format`, tqdm ignores the
`desc` argument entirely, so the title can't be controlled dynamically.

The `show_epoch` flag was meant as a simple way to disable the epoch display
without rewriting the full format string.

I, however agree that a `description_format` parameter could be a more
flexible solution than a boolean flag. I'm happy to update the implementation
in that direction if you think that would be better.

TahaZahid05:
@SpandanBhoiIITM First, can you kindly reproduce some code examples of progress bar formats that currently do not work as desired? This would help us in understanding where the current implementation is lacking. You can refer to the conversation above for more guidance.

SpandanBhoiIITM:
I ran a small example to illustrate the current behavior and the motivation behind this change.



```python
from ignite.handlers import ProgressBar

train_data = range(100)
eval_data = range(40)
max_epochs = 4

    import time
    time.sleep(0.01)

trainer = Engine(train_step)
evaluator = Engine(lambda e, b: None)

@trainer.on(Events.EPOCH_COMPLETED(every=2))
def run_validation():
    evaluator.run(eval_data)

ProgressBar(persist=True, desc="Training").attach(trainer)
ProgressBar(persist=True, desc="Validation").attach(evaluator)

trainer.run(train_data, max_epochs=max_epochs)
```

---

### Current behavior

This produces the following representation in the terminal / Jupyter notebook:

image


Here the epoch information `[epoch/max_epochs]` is automatically appended to the progress bar description, for example:

```
Training [1/4]
Training [2/4]
Validation [40/40]
```

---

### Attempt using `bar_format`

As @vfdev-5 mentions , It is possible to modify the layout using `bar_format`, for example:

```python
bformat = "{percentage:3.0f}%|{bar}{postfix} [{elapsed}<{remaining}] - {desc}[{n_fmt}/{total_fmt}]"

ProgressBar(persist=True, bar_format=bformat, desc="Training").attach(trainer)
ProgressBar(persist=True, bar_format=bformat, desc="Validation").attach(evaluator)
```

image


However, if a user only wants to **remove the epoch display**, they currently need to redefine the entire `bar_format`.

---

### With the proposed change

Using the new parameter:

```python
from ignite.handlers import ProgressBar
import time

data = range(20)

    time.sleep(0.05)
    return batch


trainer = Engine(train_step)


ProgressBar(
    persist=True,
    desc="Training",
    show_epoch=False   
).attach(trainer)


trainer.run(data, max_epochs=3)
```



image


Here the `[epoch/max_epochs]` part is removed while still keeping the dynamic description provided through `desc`.

This allows users to hide the epoch indicator without rewriting the entire `bar_format`, which makes the API simpler for this use case.

vfdev-5:
@SpandanBhoiIITM what's the point to remove the appended epoch/max-epochs info ?

To move forward with this, I suggest to create an issue discussing this feature. If there's a support of the community on it, we can consider this PR. For now, personally I do not see the usefulness of this feature.

SpandanBhoiIITM:
@vfdev-5  Thanks for the suggestion. I have opened an issue to discuss the feature here: #3659.

PR #3637, opened 2026-03-05
Feat: implement get_sequence_transform for Accuracy metric

### LINKED ISSUE #3593: [Feature]:SequenceAccuracy Metric for PyTorch Ignite
## 🚀 Feature

I think  the addition of a ```SequenceAccuracy``` metric (or an enhancement to the existing Accuracy metric) to natively support sequence-to-sequence outputs common in NLP and Transformer-based models will be good.

## Motivation:

When working with sequence models, the model output is typically of shape ```(batch, sequence_length, num_classes)``` and the target is ```(batch, sequence_length)```. Currently, the standard ignite.metrics.Accuracy expects inputs to be already flattened or simple 1D/2D tensors.

I have to write a custom output_transform for every single project to flatten the tensors and create masking.

Current work around:
```
def get_metrics(loss_fn, pad_idx=0):
    def accuracy_transform(output):
        preds, y = output

        preds_flat = preds.reshape(-1, preds.size(-1))
        y_flat = y.reshape(-1)

        mask = y_flat != pad_idx

        preds_masked = preds_flat[mask]
        y_masked = y_flat[mask]

        return preds_masked, y_masked

    return {
        "accuracy": metrics.Accuracy(output_transform=accuracy_transform),
        "loss": metrics.Loss(loss_fn),
    }

```
As ```Accuracy``` is one of the most common nlp metric I think offering a more plug and play option will be godd.

We can create a new metric ```SequenceAccuracy``` which can be a wrapper around ```Accuracy``` and perform masking, alternatively we can modify the ```Accuracy``` itself to support sequences and masking.

Manimaran-tech:
## 🚀 Feature: Sequence Accuracy Transform
Resolves #3593 

This PR implements \Accuracy.get_sequence_transform()\, natively supporting sequence-to-sequence outputs (common in NLP and Transformer-based models) without requiring users to write custom flattening/masking boilerplate.

**Changes:**
- Added \get_sequence_transform(ignore_index, output_transform)\ to \ignite.metrics.Accuracy\.
- Automatically handles \(batch, seq_len, num_classes)\ and \(batch, num_classes, seq_len)\ shaped outputs by flattening sequence dimensions to \(batch*seq_len, num_classes)\.
- Applies an optional \ignore_index\ mask to filter out padding tokens prior to accuracy calculation.
- Added comprehensive unit tests in \	est_accuracy.py\ for padded models, binary sequences, and various shape orientations.


Prathamesh8989:
Nice implementation! This looks like a clean way to support sequence outputs without introducing a new metric class .

I had a couple of small questions/suggestions while reading through the code:

- Should we add validation for unsupported tensor shape combinations? Currently, if the shapes don't match any of the handled cases, the function will silently pass through without reshaping, which might lead to confusing errors later in `Accuracy.

- In the earlier discussion, there was a mention of possibly supporting multiple ignored indices. Would it make sense for `ignore_index` to also accept something like `Iterable[int]`, or do we want to keep it limited to a single value for now?

- The issue thread also briefly mentioned targets potentially being `PackedSequence`. Do we want this transform to support that case as well, or should we assume simple padding-based masking for now?

Overall the approach looks good and keeps the API simple by reusing `Accuracy`. Nice work!

ignite/metrics/accuracy.py
```diff
@@ -263,3 +263,39 @@ def compute(self) -> float:
         if self._num_examples == 0:
             raise NotComputableError("Accuracy must have at least one example before it can be computed.")
         return self._num_correct.item() / self._num_examples
+
+    @staticmethod
+    def get_sequence_transform(
```
vfdev-5: Actually, should this be specific to Accuracy only? Other metrics like Precision, Recall etc could benefit from it

aaishwarymishra:
Also there was suggestion for the packed sequence and reduction on sequence on the issue you can check them and give your thoughts.

Manimaran-tech:
I have addressed the comments by:
1. Moving `get_sequence_transform` out of the `Accuracy` metric and into a new `ignite/metrics/utils.py` file.
2. Adding explicit `ValueError` shape parameter validations.
3. Updating the logic to support passing an `Iterable[int]` to `ignore_index` to mask out multiple tokens simultaneously.
4. Adding tests for the generalized handler in `test_utils.py`.

Let me know if there are any other changes needed!

ignite/metrics/accuracy.py
```diff
@@ -263,3 +263,5 @@ def compute(self) -> float:
         if self._num_examples == 0:
             raise NotComputableError("Accuracy must have at least one example before it can be computed.")
         return self._num_correct.item() / self._num_examples
+
```
vfdev-5: Let's revert added empty lines

ignite/metrics/utils.py
```diff
@@ -0,0 +1,67 @@
+import torch
+from typing import Callable, Iterable, Sequence, Union
+
+
+def get_sequence_transform(
+    ignore_index: Union[int, Iterable[int], None] = None,
```
vfdev-5: use the new type hints

tests/ignite/metrics/test_accuracy.py
```diff
@@ -498,3 +498,5 @@ def update(self, output):
     state = State(output=(y_pred, y_true))
     engine = MagicMock(state=state)
     acc.iteration_completed(engine)
+
```
vfdev-5, line 501: Same here

Manimaran-tech: sry my bad done the changes..!!!

aaishwarymishra:
There is still no case for `y_pred.ndimension() == 3`  and  `y.ndimension() == 3`

tests/ignite/metrics/test_utils.py
```diff
@@ -0,0 +1,75 @@
...
+            [[0.1, 0.9], [0.8, 0.2], [0.3, 0.7], [0.5, 0.5]],
+            [[0.9, 0.1], [0.2, 0.8], [0.4, 0.6], [0.5, 0.5]],
+        ]
+    )  # shape: (2, 4, 2)
+    y = torch.tensor([[1, 0, 1, -1], [0, 1, 0, -1]])  # shape: (2, 4)
+
+    transform = get_sequence_transform(ignore_index=-1)
+    y_pred_t, y_t = transform((y_pred, y))
+
+    assert y_pred_t.shape == (6, 2)
```
aaishwarymishra: I think when you apply mask the returned tensor is 1D, can you double check tests and run them.

This test will fail.

aaishwarymishra:
The nested if-else statements can be improved, also I think it would be better if you make separate test for shape check and ignore index check.

aaishwarymishra:
@vfdev-5 what do you think about having a separate utils file? I am not sure if this is the approach.

tests/ignite/metrics/test_utils.py
aaishwarymishra: my bad its correct

ignite/metrics/utils.py
```diff
@@ -0,0 +1,77 @@
...
+            else:
+                raise ValueError(
+                    f"y_pred and y have incompatible sequence shapes: "
+                    f"y_pred={y_pred.shape} vs y={y.shape}"
+                )
+        elif y_pred.ndimension() == 3 and y.ndimension() == 3:
+            # y_pred is (N, L, C) or (N, C, L), y has the same shape
+            if y_pred.shape == y.shape:
+                y_pred = y_pred.reshape(-1)
+                y = y.reshape(-1)
```
aaishwarymishra: For this case the `y_pred`   should be in shape `(N,C)` this block is turning it to `(N*C)` and for `y` argmax should be taken because the Accuracy metric doesn't support one hot encoded labels.

vfdev-5:
> @vfdev-5 what do you think about having a separate utils file? I am not sure if this is the approach.

It can be one way of doing things. I would do it a bit differently and add the transform as static method to the base class for Accuracy, Precision, Recall such that users could get it quicker as `Accuracy.get_sequence_transform` or `Precision.get_sequence_transform` etc.
@aaishwarymishra how would you do it?

aaishwarymishra:
> > @vfdev-5 what do you think about having a separate utils file? I am not sure if this is the approach.
> 
> It can be one way of doing things. I would do it a bit differently and add the transform as static method to the base class for Accuracy, Precision, Recall such that users could get it quicker as `Accuracy.get_sequence_transform` or `Precision.get_sequence_transform` etc. @aaishwarymishra how would you do it?

Accuracy,precision and recall inherit from the `_BaseClassification` we can add it as a static method on that class.

vfdev-5:
@Manimaran-tech let's do the following:

> Accuracy,precision and recall inherit from the _BaseClassification we can add it as a static method on that class.

Manimaran-tech:
Sure @vfdev-5 
Thanks for the suggestion! @aaishwarymishra  That makes sense. Since **Accuracy, Precision, and Recall** inherit from `_BaseClassification`, adding the transform as a **static method in `_BaseClassification`** would keep the implementation centralized and make it easier to access from the derived classes.

I'll proceed with implementing it there so it can be reused directly by the child classes.

aaishwarymishra:
you forgot to remove the `metrics_utils.py` I forgot its name, and its test case.

Also that if else block is very big and hard to read can you try to make it simpler?

aaishwarymishra:
Here use this instead of that nested if-else block for shape check , its much cleaner.
@vfdev-5  what do you think should we make this shape check a static helper function? Or use it directly.

```python
incompat = f"y_pred and y have incompatible sequence shapes: y_pred={y_pred.shape} vs y={y.shape}"

match (ndp, ndy):
    # 1. 3D / 2D
    case (3, 2):
        if y_pred.shape[:2] == y.shape:  # (N, L, C)
            y_pred = y_pred.reshape(-1, y_pred.size(-1))
        elif (y_pred.shape[0], y_pred.shape[2]) == y.shape:  # (N, C, L)
            y_pred = y_pred.transpose(1, 2).reshape(-1, y_pred.size(1))
        else:
            raise ValueError(incompat)
        y = y.reshape(-1)

    # 2. 3D / 3D one-hot
    case (3, 3):
        if y_pred.shape != y.shape:
            raise ValueError(incompat)
        y_pred = y_pred.reshape(-1, y_pred.size(-1))
        y = y.reshape(-1, y.size(-1)).argmax(dim=-1)

    # 3. 2D / 2D binary
    case (2, 2):
        if y_pred.shape != y.shape:
            raise ValueError(incompat)
        y_pred, y = y_pred.reshape(-1), y.reshape(-1)

    # 4. Fallback for unsupported dimensions
    case _:
        raise ValueError(f"Unsupported dimensions: y_pred={ndp}D, y={ndy}D")
```

vfdev-5:
@aaishwarymishra I wonder if these checks should not be already done in the metric implementation?
Can't we just apply the mask and leave the metric implementation do the check?
Otherwise, this looks a bit redundant...

aaishwarymishra:
> @aaishwarymishra I wonder if these checks should not be already done in the metric implementation?
> Can't we just apply the mask and leave the metric implementation do the check?
> Otherwise, this looks a bit redundant...

Right but what about the case of when y_pred is in shape (N,C,L) in that case we would need to take transpose and in case y_labels in shape (N,L,C) ?

And If the shape or not aligned the mask will give error during the masking.

Also I have a doubt like what if we recieve a shape of say 4 dim in y_preds and 3_dim in the y_labels , the mask will just flatten them and the metric will accept it, I am not sure if masking should just support till 3 dim or higher dims too as sequences are 3 dim max

vfdev-5:
Actually, there is no mention of what is expected for y_pred and y. I think we should first decide what type of y_pred, y we want to support. I think we should support at most 2 most common cases and put that info explicitly in the docs.
This way we would avoid also these long if/else statements.

aaishwarymishra:
sounds good.

Manimaran-tech:
implemented the requested updates. I simplified get_sequence_transform to support only the documented common sequence cases, removed the 3D/3D one-hot branch, and updated tests/docs accordingly. I also validated locally with sequence-transform and non-distributed accuracy tests, and they pass

aaishwarymishra:
@vfdev-5 what shapes should we support? I think format (N,L,C) be default for sequences, and max 3 dim for this static method ?

Also should we specify that labels should be (N,L) or also one hot encoding (N,L,C)

vfdev-5:
@aaishwarymishra let's support the most natural formats. For me, they are y_preds is (N, S, E) and y_true is (N, S)

aaishwarymishra:
@Manimaran-tech  can you refactor your code for these shapes :)

aaishwarymishra:
@vfdev-5  I think this looks fine for now, we can add packing support and other features to this in separate pr.

aaishwarymishra (approved):

PR #3639, opened 2026-03-05
Fix torch.load calls to explicitly set weights_only=True

### LINKED ISSUE #3303: [BC breaking change in torch] weights_only default flip for torch.load
We've flipped the default for the weights_only argument in torch.load to True in pytorch/pytorch, see [here](https://dev-discuss.pytorch.org/t/bc-breaking-change-torch-load-is-being-flipped-to-use-weights-only-true-by-default-in-the-nightlies-after-137602/2573) for details + [documentation](https://pytorch.org/docs/main/notes/serialization.html#torch-load-with-weights-only-true) and this is coming in torch 2.6.

This is expected to be quite a BC-breaking change, especially if any torch.load calls are not loading state_dicts of plain tensors.

I see that `ignite` runs nightly CI as https://github.com/pytorch/ignite/pull/3299/files already addressed some of the `torch.load` calls in test cases.

However, I see a [couple other `torch.load` calls without weights_only explicitly set](https://github.com/search?q=repo%3Apytorch%2Fignite+%22torch.load%22&type=code&p=1) and wanted to double check that these are passing with the flip :) e.g.
https://github.com/pytorch/ignite/blob/0d8b8a1828eb00d40bb38018cca017ba3b13bc8b/ignite/handlers/lr_finder.py#L515
and 
https://github.com/pytorch/ignite/blob/0d8b8a1828eb00d40bb38018cca017ba3b13bc8b/ignite/handlers/param_scheduler.py#L848

rogueslasher:
Fixes #3303

`torch.load` calls in `lr_finder.py` and `param_scheduler.py` were missing  an explicit `weights_only` argument. torch 2.6 flips the default for  `weights_only` to `True`, which breaks loading of non-tensor objects from  cache. Explicitly set `weights_only=True` to fix potential issues.

### DIFF
--- a/ignite/metrics/precision.py
+++ b/ignite/metrics/precision.py
@@ -122,7 +122,7 @@ class _BasePrecisionRecall(_BaseClassification):
         self._weight: int | torch.Tensor = 0
         self._updated = False
 
-        super(_BasePrecisionRecall, self).reset()
+        super().reset()
 
     @sync_all_reduce("_numerator", "_denominator")
     def compute(self) -> torch.Tensor | float:
@@ -179,7 +179,7 @@ class Precision(_BasePrecisionRecall):
         average: available options are
 
             False
-              default option. For multicalss and multilabel inputs, per class and per label
+              default option. For multiclass and multilabel inputs, per class and per label
               metric is returned respectively.
 
             None
@@ -212,7 +212,7 @@ class Precision(_BasePrecisionRecall):
               Incompatible with binary and multiclass inputs.
 
             'weighted'
-              like macro precision but considers class/label imbalance. for binary and multiclass
+              like macro precision but considers class/label imbalance. For binary and multiclass
               input, it computes metric for each class then returns average of them weighted by
               support of classes (number of actual samples in each class). For multilabel input,
               it computes precision for each label then returns average of them weighted by support
--- a/ignite/metrics/recall.py
+++ b/ignite/metrics/recall.py
@@ -27,7 +27,7 @@ class Recall(_BasePrecisionRecall):
         average: available options are
 
             False
-              default option. For multicalss and multilabel inputs, per class and per label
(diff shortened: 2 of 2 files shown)


vfdev-5:
Thanks for the PR @rogueslasher ! Can you please check if pytorch version 2.2.2 has already `weights_only` arg. We support min pytorch version: 2.2.2

aniketpandey05:
Hi @vfdev-5! I verified by installing PyTorch 2.2.2 and checking the signature of `torch.load` `weights_only` is present with a default of `False`. So explicitly passing `weights_only=False` is fully backward compatible with the minimum supported version of 2.2.2.

vfdev-5:
Actually, let's do the opposite. Everywhere we use torch.load we add `weights_only=True`. Let's see if the CI is passing

aniketpandey05:
@vfdev-5  i have updated the weight_only to true

vfdev-5:
@rogueslasher let's add weights_only=True **everywhere** we use `torch.load`

aniketpandey05:
@vfdev-5 i have added weights_only=true to all torch.load calls to `checkpoint.py`, `param_scheduler.py`, and `lr_finder.py`

ignite/handlers/checkpoint.py
```diff
@@ -635,7 +635,7 @@ def load_objects(to_load: Mapping, checkpoint: str | Mapping | Path, **kwargs: A
 
                     to_load = to_save
                     checkpoint_fp = Path(tmpdirname) / 'myprefix_checkpoint_40.pt'
-                    checkpoint = torch.load(checkpoint_fp)
+                    checkpoint = torch.load(checkpoint_fp,weights_only=True)
```
```suggestion
                    checkpoint = torch.load(checkpoint_fp, weights_only=True)
```

aniketpandey05:
@vfdev-5 thank you for the fix and reviewing the pr.

PR #3651, opened 2026-03-06
replace process.wait() with process.communicate() in execute() to prevent deadlock

TahaZahid05:
Related: #3652

**Description**
Fixed potential deadlock in the `execute` function by replacing `process.wait()` with `process.communicate()`
following the note in Popen.wait doc: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

> Note This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use [Popen.communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) when using pipes to avoid that.

Check list:

### DIFF
--- a/tests/ignite/distributed/test_launcher.py
+++ b/tests/ignite/distributed/test_launcher.py
@@ -48,11 +48,14 @@ def execute(cmd, env=None):
     env = dict(os.environ) if env is None else env
     env["PYTHONPATH"] = f"{os.path.dirname(ignite.__path__[0])}"
     process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=env)
-    process.wait()
+    stdout, stderr = process.communicate()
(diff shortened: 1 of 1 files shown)


vfdev-5:
@TahaZahid05 thanks a lot for digging that! 
The fact that it was deadlocked on nightly only and not with stable version may think that its a pytorch issue and which will be fixed by them at some point...
let's split this PR into two: 
- fix for potential deadlock in execute by replacing process.wait() with process.communicate()
- tcp option skip in the test for nightly only

The first can land quickly. The second is a bit problematic as we do not want to remove tests. We should reenable it once it is fixed.

TahaZahid05:
@vfdev-5 Done! I have changed this pr to only merge `process.communicate()`. PR #3658 deals with the tcp change.

Issue #3655, opened 2026-03-07
Fix Edge Case Crashes in Clustering Metrics


TahaZahid05:
During early training epochs, models often output cluster distributions that are degenerate (e.g., assigning all samples to a single cluster, or assigning every single sample to its own unique cluster). 

Currently, metrics within the `ignite.metrics.clustering` module rely on underlying `scikit-learn` implementations that do not support these edge cases.  They throw a `ValueError` (such as `Number of labels is 1. Valid values are 2 to n_samples - 1`), which causes the entire Ignite `Engine` to crash during an evaluation run. 

**Identified Cases**

- SilhouetteScore (#3641)
- DaviesBouldinScore (#3642)
- CalinskiHarabaszScore (#3641)

Credits to @Prathamesh8989 for pointing these issues out!

This issue serves to track these issues under one consolidated issue. PRs can be submitted for any new metrics (if added in the future) that suffer from the same problem.

nightcityblade:
Hi, I'd like to work on the CalinskiHarabaszScore edge case (the remaining metric from this tracking issue not yet covered by #3654). I'll submit a PR shortly.

ChandraguptSharma07:
Hey! I believe that this issue is probably happening due to scikit learn raising a value error, I am able to recreate the issue, and I am close to fixing it. If possible, can I work on this issue as well?

TahaZahid05:
@ChandraguptSharma07 Hi, you can help reviewing the existing PRs once they are up for review.

ChandraguptSharma07:
@TahaZahid05 that's great, but I have actually started working on DaviesBouldinScore (sorry I am new to open sourcing, so I started working on it before discussing it here). I was about to run the tests, and if they pass, can I make a PR?

TahaZahid05:
@ChandraguptSharma07 Hi! No worries at all, and welcome to the project! To keep things organized and avoid duplicate work, we usually stick to one PR per issue. Since someone has already started on this one, it is best to proceed with it.

You can still contribute by reviewing the PR once it’s live. In the meantime, you can checkout other 'help wanted' issues,

ChandraguptSharma07:
@TahaZahid05 sure that sounds great, actually my tests passed, and @nightcityblade mentioned that they were working on CalinskiHarabaszScore edge case, so if they are okay, I can make PRs for the other two :D

TahaZahid05:
[...]

PR #3654, opened 2026-03-07
fix: return NaN for invalid cluster counts in SilhouetteScore and DaviesBouldinScore

lfshao:
Fixes #3641
Fixes #3642

## Description

`SilhouetteScore` and `DaviesBouldinScore` currently crash with a `ValueError` when passed labels with invalid cluster counts (e.g., all samples in a single cluster, or every sample being its own cluster). This commonly happens during early training epochs before the model has learned to separate data.

**Root cause:** Both metrics delegate to `sklearn`, which raises `ValueError` when the number of unique labels is outside its valid range:
- `davies_bouldin_score` requires 2 ≤ n_labels ≤ n_samples
- `silhouette_score` requires 2 ≤ n_labels ≤ n_samples - 1

**Fix:** Add a pre-validation guard in both compute functions. When the cluster count is invalid, return `float('nan')` instead of raising an exception. This allows an `Engine`-based training loop to continue without crashing.

## Changes

- `ignite/metrics/clustering/davies_bouldin_score.py`: Return `float('nan')` when fewer than 2 unique cluster labels are present.
- `ignite/metrics/clustering/silhouette_score.py`: Return `float('nan')` when the number of unique cluster labels is < 2 or ≥ n_samples.
- Added tests for both edge cases in the respective test files.

## Check list

- [x] New tests are added (if a new feature is added)
- [x] New doc strings: description and/or example code are in RST format
- [ ] Documentation is updated (if required)

PR #3656, opened 2026-03-07
fix: return NaN for invalid cluster counts in CalinskiHarabaszScore

nightcityblade:
Related to #3655

Description:

`CalinskiHarabaszScore` crashes with a `ValueError` when all samples belong to a single cluster, because `sklearn.metrics.calinski_harabasz_score` requires at least 2 unique labels. This is the same issue that was fixed for `SilhouetteScore` and `DaviesBouldinScore` in #3654.

This PR adds a guard that returns `float("nan")` when the number of unique labels is less than 2, preventing the engine from crashing during early training epochs.

Changes:
- Added cluster count validation in `_calinski_harabasz_score()` 
- Added docstring note about NaN return behavior
- Added `test_single_cluster_returns_nan` test

Check list:

### DIFF
--- a/tests/ignite/metrics/clustering/test_calinski_harabasz_score.py
+++ b/tests/ignite/metrics/clustering/test_calinski_harabasz_score.py
@@ -11,6 +11,7 @@ from ignite.exceptions import NotComputableError
 from ignite.metrics.clustering import CalinskiHarabaszScore
 
 
+@pytest.mark.parametrize("check_compute_fn", [True, False])
 @pytest.mark.parametrize(
     "features, labels, desc",
     [
@@ -18,11 +19,11 @@ from ignite.metrics.clustering import CalinskiHarabaszScore
         (torch.randn(5, 3), torch.arange(5, dtype=torch.long), "each sample own cluster"),
     ],
 )
-def test_invalid_cluster_count_returns_nan(features, labels, desc):
+def test_invalid_cluster_count_returns_nan(features, labels, desc, check_compute_fn):
     """CalinskiHarabaszScore should return NaN for invalid cluster counts ({desc})."""
     import math
 
-    metric = CalinskiHarabaszScore()
+    metric = CalinskiHarabaszScore(check_compute_fn=check_compute_fn)
     metric.update((features, labels))
     result = metric.compute()
     assert math.isnan(result)


ignite/metrics/clustering/calinski_harabasz_score.py
```diff
@@ -14,6 +14,11 @@ def _calinski_harabasz_score(features: Tensor, labels: Tensor) -> float:
 
     np_features = features.cpu().numpy()
     np_labels = labels.cpu().numpy()
+
+    n_unique_labels = len(set(np_labels))
+    if n_unique_labels < 2:
```
TahaZahid05: Validation logic should also cover the case `n_unique_labels == n_samples`. We should also display a warning to the user. Let's also add a `try...except` block so if we do get any future sklearn errors, it doesn't break the engine

tests/ignite/metrics/clustering/test_calinski_harabasz_score.py
```diff
@@ -19,6 +19,18 @@ def test_zero_sample():
...
+    """CalinskiHarabaszScore should return NaN when all samples belong to a single cluster."""
+    import math
+
+    metric = CalinskiHarabaszScore()
+    features = torch.randn(10, 5)
+    labels = torch.zeros(10, dtype=torch.long)
+    metric.update((features, labels))
+    result = metric.compute()
+    assert math.isnan(result)
+
```
TahaZahid05, line 30: Add another test case for the validation check `n_unique_labels == n_samples`

ignite/metrics/clustering/calinski_harabasz_score.py
aaishwarymishra: I don't think we should add exception handling with try except, if there is a problem the engine should throw error and stop.

TahaZahid05: True. we can just let sklearn errors surface.

nightcityblade:
Thanks for the review! I've updated to also handle the case where `n_unique_labels == n_samples` (each sample is its own cluster), returning NaN. Added a corresponding test case `test_each_sample_own_cluster_returns_nan` as well.

ignite/metrics/clustering/calinski_harabasz_score.py
```diff
@@ -14,6 +14,12 @@ def _calinski_harabasz_score(features: Tensor, labels: Tensor) -> float:
 
     np_features = features.cpu().numpy()
     np_labels = labels.cpu().numpy()
+
+    n_unique_labels = len(set(np_labels))
+    n_samples = len(np_labels)
+    if n_unique_labels < 2 or n_unique_labels == n_samples:
```
TahaZahid05: Looking at the other PR patching the other two metrics, better to change `==` to `>=` as it is more defensive:

```suggestion
    if n_unique_labels < 2 or n_unique_labels >= n_samples:
```

TahaZahid05 (commented):
Rest looks fine, just change the check from `==` to `>=`

nightcityblade:
Thanks @TahaZahid05! Changed `==` to `>=` for a more defensive check, as suggested.

nightcityblade:
Thanks for the approval @TahaZahid05! 🎉

ignite/metrics/clustering/calinski_harabasz_score.py
```diff
@@ -14,6 +14,12 @@ def _calinski_harabasz_score(features: Tensor, labels: Tensor) -> float:
 
     np_features = features.cpu().numpy()
     np_labels = labels.cpu().numpy()
+
+    n_unique_labels = len(set(np_labels))
```
vfdev-5: Can't we use numpy methods to get the number of unique values instead of using python set ?

ignite/metrics/clustering/calinski_harabasz_score.py
vfdev-5: Maybe, it does cover more, but looks a bit weird :)
OK, let's keep it like this...

tests/ignite/metrics/clustering/test_calinski_harabasz_score.py
```diff
@@ -19,6 +19,30 @@ def test_zero_sample():
         metric.compute()
 
 
+def test_single_cluster_returns_nan():
```
vfdev-5: Let's unify the test cases using pytest.mark.parametrize

nightcityblade:
Thanks @vfdev-5 for the review! I've addressed all your feedback:

1. ✅ Replaced `set()` with `np.unique()` for label counting
2. ✅ Kept `>=` for the defensive check
3. ✅ Unified the NaN test cases using `pytest.mark.parametrize`

Force-pushed with the changes.

vfdev-5:
@nightcityblade seems like you messed up with the git commits history.

nightcityblade:
Sorry about the messy commit history @vfdev-5! I've force-pushed a clean single commit rebased on master. The change is the same — adds NaN return for invalid cluster counts with tests. Let me know if anything else needs fixing.

ignite/metrics/clustering/calinski_harabasz_score.py
```diff
@@ -14,6 +16,12 @@ def _calinski_harabasz_score(features: Tensor, labels: Tensor) -> float:
 
     np_features = features.cpu().numpy()
     np_labels = labels.cpu().numpy()
+
+    n_unique_labels = len(np.unique(np_labels))
```
TahaZahid05: I think using `torch.unique` is preferred to avoid switching to cpu.

ignite/metrics/clustering/calinski_harabasz_score.py
```diff
@@ -14,6 +16,12 @@ def _calinski_harabasz_score(features: Tensor, labels: Tensor) -> float:
 
     np_features = features.cpu().numpy()
     np_labels = labels.cpu().numpy()
```
TahaZahid05: add `.detach()` to avoid runtime issues.

ignite/metrics/clustering/calinski_harabasz_score.py
```diff
@@ -1,6 +1,8 @@
 from collections.abc import Callable
 from typing import Any
 
+import numpy as np
```
TahaZahid05: we can remove the import as well if we use `torch.unique`

tests/ignite/metrics/clustering/test_calinski_harabasz_score.py
```diff
@@ -11,6 +11,24 @@
...
 
+
+@pytest.mark.parametrize(
+    "features, labels, desc",
+    [
+        (torch.randn(10, 5), torch.zeros(10, dtype=torch.long), "single cluster"),
+        (torch.randn(5, 3), torch.arange(5, dtype=torch.long), "each sample own cluster"),
+    ],
+)
+def test_invalid_cluster_count_returns_nan(features, labels, desc):
```
TahaZahid05: We can test on both `True` and `False` for `check_compute_fn` now.

nightcityblade:
Thanks for the detailed feedback! I've addressed all review comments:

- Replaced `np.unique` with `torch.unique(labels.detach())` to avoid unnecessary CPU transfer
- Removed the now-unused `numpy` import
- Added `check_compute_fn` parametrization (`True`/`False`) to the invalid cluster count tests

Let me know if anything else needs adjusting.

TahaZahid05:
I don't see the commit. kindly fix it.

nightcityblade:
Force-pushed a clean rebase on latest master. The commit should be visible now @TahaZahid05 — single commit with the NaN validation using `torch.unique(labels.detach())`. Tests pass locally.

vfdev-5:
Code style check is failing: https://github.com/pytorch/ignite/actions/runs/23405930582/job/68113360470?pr=3656

nightcityblade:
Fixed the black formatting issue — removed the extra blank line in the test file. CI should pass now. Thanks for the review!

nightcityblade:
Thanks for the reviews @TahaZahid05 @vfdev-5! I've addressed the feedback:
- Added `check_compute_fn` parametrize (`True`/`False`) to the invalid cluster count test
- The code already uses `torch.unique` with `.detach()` and `>=` for the bounds check

Ready for re-review!

Issue #3652, opened 2026-03-07
Fixing pytorch-nightly ci test and replacing use of `process.wait()`


TahaZahid05:
I was wondering why the pytorch-nightly tests kept failing across all PRs, so I went ahead and debugged the CI failure.

**Reason for failing**

In `test_launcher.py`, the function `test_check_idist_parallel_torch_launch_n_procs_native` was parametrized with `tcp://` as an init_method while also using `torchrun` to launch workers. However, these two are incompatible as `tcp://` is designed for when you manually spawn processes and manage rendezvous yourself. `torchrun` already handles all of that coordination internally and sets up the necessary env vars, so passing a `tcp://` init_method causes the workers to attempt a second rendezvous that never completes, hanging indefinitely until the CI timeout is hit.

The fix is to remove `tcp://` from that test's parametrize list. It is already covered by `test_idist_parallel_spawn_n_procs_native` and `test_idist_parallel_n_procs_native`, which use `idist.Parallel` directly and are the appropriate place to test `tcp://`.

**Additional fix**

I also noticed that the `execute` function was using `process.wait()`, which can cause a deadlock if the subprocess produces enough output to fill the OS pipe buffer. Replaced it with `process.communicate()` which reads from both pipes concurrently while waiting for the process to exit. I can remove it if we decide that it is not needed as it wasn't causing any tests to fail (just a nice fix to have).

I have already validated this in a draft PR and will convert it to ready for review alongside this issue.

PR #3658, opened 2026-03-08
fix: skip tcp:// init_method with torchrun on nightly builds

TahaZahid05:
Related: #3652 

**Description**
The test is skipped on nightly only and will be re-enabled once the underlying deadlock issue is fixed.

Check list:


TahaZahid05:
@vfdev-5, I think the remaining checks failing/stalling are not relevant to my changes? Can you take a look.

PR #3660, opened 2026-03-08
 add tests for uncovered lines in engine.py

rogueslasher:
## Summary
This PR improves test coverage for `ignite/engine/engine.py` as part of the ongoing coverage improvement effort referenced in #1790.

duplicate tests were removed after cross-checking against `test_engine_state_dict.py`, `test_custom_events.py`, `test_event_handlers.py`, and `test_deterministic.py`. Internal error checks in `engine.py` were also removed per maintainer suggestion, with `# pyrefly: ignore` comments added to suppress type errors flagged by pyrefly.

## Remaining tests
| Test | What it covers |
|------|----------------|
| `test_run_with_data_whose_len_raises_type_error` | Silent `except TypeError: pass` block when `len()` raises on an IterableDataset |
| `test_max_epochs_calculated_from_max_iters_unknown_epoch_length` | `max_epochs` auto-calculated as `ceil(max_iters / epoch_length)` for unknown-length iterators |
| `test_run_resume_raises_on_epoch_length_mismatch` | ValueError when resuming `run()` with mismatched `epoch_length` |
| `test_resume_after_terminate_resets_init_iter` | `_init_iter` reset after mid-run termination |
| `test_resume_raises_when_data_none_and_epoch_length_none` | ValueError guard when resuming with `data=None` and `epoch_length=None` |

### DIFF
--- a/docs/source/distributed.rst
+++ b/docs/source/distributed.rst
@@ -15,7 +15,7 @@ Distributed launcher and `auto` helpers
 
 We provide a context manager to simplify the code of distributed configuration setup for all above supported backends.
 In addition, methods like :meth:`~ignite.distributed.auto.auto_model`, :meth:`~ignite.distributed.auto.auto_optim` and
-:meth:`~ignite.distributed.auto.auto_dataloader` helps to adapt in a transparent way provided model, optimizer and data
+:meth:`~ignite.distributed.auto.auto_dataloader` help to adapt in a transparent way provided model, optimizer and data
 loaders to existing configuration:
 
 .. code-block:: python
@@ -37,8 +37,8 @@ loaders to existing configuration:
         # ...
         optimizer = optim.SGD(model.parameters(), lr=0.01)
         optimizer = idist.auto_optim(optimizer)
-        # optimizer is itself, except XLA configuration and overrides `step()` method.
-        # User can safely call `optimizer.step()` (behind `xm.optimizer_step(optimizier)` is performed)
+        # The optimizer remains unchanged, except for XLA configurations where the step() method is overridden.
+        # User can safely call `optimizer.step()` (behind `xm.optimizer_step(optimizer)` is performed)
 
 
     backend = "nccl"  # torch native distributed configuration on multiple GPUs
@@ -51,7 +51,7 @@ loaders to existing configuration:
         parallel.run(training, config, a=1, b=2)
 
 Above code may be executed with `torch.distributed.launch`_ tool or by python and specifying distributed configuration
-in the code. For more details, please, see :class:`~ignite.distributed.launcher.Parallel`,
+in the code. For more details, please see :class:`~ignite.distributed.launcher.Parallel`,
 :meth:`~ignite.distributed.auto.auto_model`, :meth:`~ignite.distributed.auto.auto_optim` and
 :meth:`~ignite.distributed.auto.auto_dataloader`.
 
--- a/ignite/handlers/checkpoint.py
+++ b/ignite/handlers/checkpoint.py
@@ -385,6 +385,8 @@ class Checkpoint(Serializable):
         self.score_name = score_name
         if self.score_name is not None and self.score_function is None:
             self.score_function = self.get_default_score_fn(self.score_name)
+        if n_saved is not None and n_saved < 1:
+            raise ValueError(f"n_saved must be a positive integer or None, got {n_saved}")
         self.n_saved = n_saved
         self.ext = "pt"
         self.global_step_transform = global_step_transform
--- a/ignite/metrics/mean_average_precision.py
+++ b/ignite/metrics/mean_average_precision.py
@@ -16,6 +16,7 @@ class _BaseAveragePrecision:
         self,
         rec_thresholds: Sequence[float] | torch.Tensor | None = None,
         class_mean: Literal["micro", "macro", "weighted"] | None = "macro",
+        device: str | torch.device = torch.device("cpu"),
     ) -> None:
         r"""Base class for Average Precision metric.
 
@@ -54,7 +55,10 @@ class _BaseAveragePrecision:
 
                 'macro'
                   computes macro precision which is unweighted mean of AP computed across classes/labels. Default.
+            device: specifies which device updates are accumulated on.
         """
+        self._device = torch.device(device)
+        self._fp_precision = torch.float32 if self._device.type == "mps" else torch.float64
         if rec_thresholds is not None:
             self.rec_thresholds: torch.Tensor | None = self._setup_thresholds(rec_thresholds, "rec_thresholds")
         else:
@@ -66,7 +70,7 @@ class _BaseAveragePrecision:
 
     def _setup_thresholds(self, thresholds: Sequence[float] | torch.Tensor, threshold_type: str) -> torch.Tensor:
         if isinstance(thresholds, Sequence):
-            thresholds = torch.tensor(thresholds, dtype=torch.double)
+            thresholds = torch.tensor(thresholds)
 
         if isinstance(thresholds, torch.Tensor):
             if thresholds.ndim != 1:
@@ -81,7 +85,7 @@ class _BaseAveragePrecision:
         if min(thresholds) < 0 or max(thresholds) > 1:
             raise ValueError(f"{threshold_type} values should be between 0 and 1, given {thresholds}")
 
-        return thresholds
+        return thresholds.to(self._device, dtype=self._fp_precision)
 
     def _compute_average_precision(self, recall: torch.Tensor, precision: torch.Tensor) -> torch.Tensor:
         """Measuring average precision.
@@ -102,7 +106,7 @@ class _BaseAveragePrecision:
             ).where(rec_thresh_indices != recall.size(-1), 0)
             recall = rec_thresholds
         recall_differential = recall.diff(
-            dim=-1, prepend=torch.zeros((*recall.shape[:-1], 1), device=recall.device, dtype=recall.dtype)
+            dim=-1, prepend=torch.zeros((*recall.shape[:-1], 1), device=recall.device, dtype=self._fp_precision)
         )
         return torch.sum(recall_differential * precision, dim=-1)
 
@@ -226,7 +230,7 @@ class MeanAveragePrecision(_BaseClassification, _BaseAveragePrecision):
             device=device,
             skip_unrolling=skip_unrolling,
         )
-        super(Metric, self).__init__(rec_thresholds=rec_thresholds, class_mean=class_mean)
+        super(Metric, self).__init__(rec_thresholds=rec_thresholds, class_mean=class_mean, device=device)
 
     @reinit__is_reduced
     def reset(self) -> None:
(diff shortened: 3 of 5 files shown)


aaishwarymishra:
@rogueslasher  Thanks for the pr, but our general workflow is issue->pr :)

vfdev-5:
@rogueslasher what is the coverage you see with your PR?

aniketpandey05:
hi @vfdev-5  Locally, running test_engine.py alone, missed lines dropped from 64 → 9, improving from 87% → 98%. The actual Codecov numbers will reflect the full picture once CI completes.

aniketpandey05:
@aaishwarymishra This PR is part of the ongoing effort tracked in #1790  apologies for the wrong issue number in the description, I have  updated it

tests/ignite/engine/test_engine.py
```diff
@@ -1328,6 +1328,249 @@ def train_step(trainer, batch):
...
+        engine = Engine(lambda e, b: None)
+        engine.state_dict_user_keys.append("best_loss")
+        engine.state.best_loss = 0.42
+        engine.run([1, 2, 3], max_epochs=1)
+
+        sd = engine.state_dict()
+        assert "best_loss" in sd
+        assert sd["best_loss"] == 0.42
+
+    def test_internal_run_legacy_raises_when_dataloader_iter_is_none(self):
```
vfdev-5: Let's not test legacy thing, it will be removed in some next version

tests/ignite/engine/test_engine.py
```diff
@@ -1328,6 +1328,249 @@ def train_step(trainer, batch):
...
+        # covers: TypeError raised when event name is not str or EventEnum
+        engine = Engine(lambda e, b: None)
+        with pytest.raises(TypeError, match="should be a str or EventEnum"):
+            engine.register_events(123)
+
+    def test_register_events_with_event_to_attr_mapping(self):
+        # covers: event_to_attr mapping path in register_events
+        engine = Engine(lambda e, b: None)
+        engine.register_events("MY_EVENT", event_to_attr={"MY_EVENT": "iteration"})
+        assert "MY_EVENT" in engine._allowed_events
```
vfdev-5: Are you sure this is not tested in test_custom_events.py?
https://github.com/pytorch/ignite/blob/1919b3b153d85a0cc69fd740c1bd4f312c7ccf33/tests/ignite/engine/test_custom_events.py#L95

vfdev-5:
@rogueslasher thanks for working on improving the coverage of the code!

We have to keep balance between new added code (maintaince) and coverage: we should avoid adding redundant test cases. If we can improve existing tests with a new parametrization or something, it would be better than writing completely new test case.

aniketpandey05:
@vfdev-5 Thanks for the feedback! actually  I went through Codecov to identify the missing lines and wrote targeted tests for each uncovered path. I should have cross-checked more carefully against existing tests to avoid redundancy. will make sure to do that from next time

aniketpandey05:
Hi @vfdev-5, i have updated and collapsed the tests into 7 parametrized groups appreciate the feedback

tests/ignite/engine/test_engine.py
```diff
@@ -1328,6 +1328,219 @@ def train_step(trainer, batch):
         assert trainer.state.epoch == 2
         assert trainer.state.iteration == 2 * 4
 
+    def test_run_with_data_whose_len_raises_type_error(self):
```
vfdev-5, line 1332: Can you put a comment in GH for each test case what we test. For example, this one looks a bit weird to me, I want to see what do we cover with it.

tests/ignite/engine/test_engine.py
```diff
@@ -1328,6 +1328,219 @@ def train_step(trainer, batch):
...
+        assert state.iteration == 3
+        assert state.epoch_length == 3
+
+    def test_setup_dataloader_iter_raises_on_none_epoch_length(self):
+        # covers: RuntimeError raised when dataloader and epoch_length are both None during internal dataloader iterator setup
+        engine = Engine(lambda e, b: None)
+        engine.state = State(dataloader=None, epoch_length=None, max_epochs=1, iteration=0, epoch=0)
+
+        with pytest.raises(RuntimeError, match="Internal error, self.state.epoch_length is None"):
+            engine._setup_dataloader_iter()
```
vfdev-5: Such checks raising an internal error were added in the very begging of mypy to instruct it about the types and mostly useless as unrealistic. Today most probably we can remove them

tests/ignite/engine/test_engine.py
```diff
@@ -1328,6 +1328,219 @@ def train_step(trainer, batch):
...
+            ({"event_name": 123}, "should be a str or EventEnum"),
+        ],
+    )
+    def test_register_events_raises(self, kwargs, match):
+        engine = Engine(lambda e, b: None)
+        with pytest.raises((ValueError, TypeError), match=match):
+            if "event_name" in kwargs:
+                engine.register_events(kwargs["event_name"])
+            else:
+                engine.register_events("MY_EVENT", **kwargs)
```
vfdev-5: This should be moved to the appropriate file: test_custom_events.py

tests/ignite/engine/test_engine.py
```diff
@@ -1328,6 +1328,219 @@ def train_step(trainer, batch):
...
+    @pytest.mark.parametrize(
+        "action, match",
+        [
+            ("remove_nonexistent_event", "does not exist"),
+            ("remove_unregistered_handler", "is not found among registered event handlers"),
+            ("assert_invalid_event", "is not a valid event"),
+            ("fire_invalid_event", "is not a valid event"),
+        ],
+    )
+    def test_event_handler_raises_on_invalid_event(self, action, match):
```
vfdev-5: This test case is a set of different cases without any factorisation.
I think some of them are already tested.
Please check carefully this

tests/ignite/engine/test_engine.py
```diff
@@ -1328,6 +1328,219 @@ def train_step(trainer, batch):
...
+
+    @pytest.mark.parametrize(
+        "scenario",
+        [
+            "event_not_registered_returns_false",
+            "search_all_events_found",
+            "search_all_events_not_found",
+        ],
+    )
+    def test_has_event_handler(self, scenario):
```
vfdev-5: Same here. this should be already tested!

tests/ignite/engine/test_engine.py
```diff
@@ -1328,6 +1328,219 @@ def train_step(trainer, batch):
...
+    def test_resume_raises_when_data_none_and_epoch_length_none(self):
+        # covers: ValueError raised when resuming with data=None and epoch_length=None
+        engine = Engine(lambda e, b: None)
+
+        def data_iter():
+            for i in range(3):
+                yield i
+
+        engine.run(data_iter(), max_epochs=1)
+        engine.state.epoch_length = None
```
vfdev-5:
```suggestion
```
I think we can remove this code and just setmp max_epochs in the state

aniketpandey05:
Hi @vfdev-5, thanks for the feedback i have addressed the suggestions and also have checked these files if they cover the test  `test_engine_state_dict.py` `test_custom_events.py`, `test_event_handler.py`, `test_determinstic.py`,

 The remaining 7 tests are  unique across all engine test files

Please let me know if anything else needs to be changed.

tests/ignite/engine/test_engine.py
vfdev-5: @rogueslasher try to remove the uncovered code raising "nternal error" in the engine.py and run pyrefly
If pyrefly is happy, remove this test. 
And do the same for all the tests checking "internal error"

tests/ignite/engine/test_engine.py
```diff
@@ -1328,6 +1329,100 @@ def train_step(trainer, batch):
...
+        @engine.on(Events.EXCEPTION_RAISED)
+        def catch(exc):
+            caught.append(exc)
+
+        exc = ValueError("test error")
+        engine.state = State(dataloader=None, epoch_length=None, max_epochs=1, iteration=0, epoch=0)
+        engine._handle_exception(exc)
+
+        assert len(caught) == 1
+        assert caught[0] is exc
```
vfdev-5: I do not understand what line this test is covering

aniketpandey05: "Hi @vfdev-5, this test covers line 676 in engine.py  self._fire_event(Events.EXCEPTION_RAISED, e) inside _handle_exception. Line 675 checks if Events.EXCEPTION_RAISED in self._event_handlers, and line 676 only executes when a handler is registered. This test exercises that branch by registering a handler and calling _handle_exception directly." Since  codecov flagged line 676 as uncovered i added it

vfdev-5: This test should test this: https://github.com/pytorch/ignite/blob/6bfbdc488e74ef95250f77a8539e689fedeb7942/tests/ignite/engine/test_event_handlers.py#L474

vfdev-5: I think you should run all the tests of the tests/ignite/engine folder to get a correct coverage value.

aniketpandey05: @vfdev-5 I will do that the current pr i have updated is for the pyrefly and internal error one
I would update it 
Thank you for the feedback

aniketpandey05:
"Hi @vfdev-5, I  ran pytest tests/ignite/engine/test_engine.py --cov=ignite --cov-report=term-missing  engine.py is at 88% coverage from test_engine.py . The full tests/ignite/engine/ folder run shows 314 passed, 31 skipped.  So CI should report the combined coverage across all engine test files."

ignite/engine/engine.py
```diff
@@ -957,10 +957,7 @@ def _get_data_length(self, data: Iterable) -> int | None:
     def _setup_dataloader_iter(self) -> None:
         if self.state.dataloader is None:
             if self.state.epoch_length is None:
-                raise RuntimeError(
-                    "Internal error, self.state.epoch_length is None. "
-                    "Please, file an issue if you encounter this error."
-                )
+                raise ValueError("epoch_length should be provided if data is None")
```
vfdev-5: The change is incorrect. Can you remove completely the check:
```python
if self.state.epoch_length is None:
     raise RuntimeError(...)
```

ignite/engine/engine.py
```diff
@@ -1096,7 +1087,7 @@ def _run_once_on_dataset_as_gen(self) -> Generator[State, None, float]:
                         if self.state.dataloader is not None:
                             self._fire_event(Events.GET_BATCH_STARTED)
                             yield from self._maybe_terminate_or_interrupt()
-
+                    assert self._dataloader_iter is not None
```
vfdev-5: We should not use asserts

aniketpandey05:
Hi  @vfdev-5 i have updated the Pr. thank you for the review and the feedback

vfdev-5:
@rogueslasher can you update the description to reflect the latest diff. Thanks!

aniketpandey05:
@vfdev-5 ,  I have updated the pr description . thanks for the review

ignite/engine/engine.py
```diff
@@ -956,12 +956,7 @@ def _get_data_length(self, data: Iterable) -> int | None:
 
     def _setup_dataloader_iter(self) -> None:
         if self.state.dataloader is None:
-            if self.state.epoch_length is None:
-                raise RuntimeError(
-                    "Internal error, self.state.epoch_length is None. "
-                    "Please, file an issue if you encounter this error."
-                )
-            self._dataloader_iter = _get_none_data_iter(self.state.epoch_length)
+            self._dataloader_iter = _get_none_data_iter(self.state.epoch_length)  # pyrefly: ignore[bad-argument-type]
```
vfdev-5: Finally, let's revert all the changes in this file. I prefer to have the runtime guards.
No need to write test cases.

If you would like to help a bit, you can try to see if we can find real-like use-case when these RuntimeError can be triggered.

aniketpandey05:
Hi @vfdev-5 I have reverted the changes in engine.py

PR #3619, opened 2026-03-02
feat(early_stopping): rename min_delta→threshold, min_delta_mode→thre…

ramyars466:
## Summary

This PR renames selected parameters of `EarlyStopping` to improve API clarity and align terminology with `ReduceLROnPlateau`, while preserving full backward compatibility.

---

## Changes

- Renamed:
  - `min_delta` → `threshold`
  - `min_delta_mode` → `threshold_mode`
  - `cumulative_delta` → `cumulative`
- Added deprecation handling for old argument names
- Deprecated arguments emit `DeprecationWarning`
- Deprecated arguments override new ones to preserve existing behavior
- Retained deprecated attribute names on the instance to avoid breaking user code
- Preserved original validation error messages
- Internal logic and behavior remain unchanged

---

## Backward Compatibility

Backward compatibility is maintained by:

- Accepting both new and deprecated argument names
- Mapping deprecated arguments to new parameters before validation
- Preserving existing error messages expected by current tests
- Keeping deprecated attributes available on the handler instance

This ensures no behavior change for existing users.

---

## Testing

- All `EarlyStopping` handler tests pass
- Validation behavior unchanged
- Deprecated argument paths verified
- No changes required to existing tests

---

## Related

Addresses #3411

PR #3661, opened 2026-03-09
Add min_evals parameter to EarlyStopping handler

### LINKED ISSUE #3411: Enhancements to EarlyStopping handler

D0-T:
Closes #3411

Description:

Adds a 'min_evals' parameter to 'EarlyStopping' to implement a warmup period.
When specified, the handler tracks the best score but does not enforce patience
until it has been called at least 'min_evals' times.
Changes:

- 'ignite/handlers/early_stopping.py': new 'min_evals' param, updated '__call__', 'reset', 'state_dict', 'load_state_dict'

Check list:

### DIFF
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -37,6 +37,9 @@ class EarlyStopping(Serializable, ResettableHandler):
 
             Possible values are "abs" and "rel". Default value is "abs".
         mode: Whether to maximize ('max') or minimize ('min') the score. Default is 'max'.
+        min_evals: Minimum number of evaluations before early stopping can trigger. The handler will
+            not stop training until it has been called at least ``min_evals`` times, regardless of
+            the score. Default value is 0 (no warmup period).
 
     Examples:
         .. code-block:: python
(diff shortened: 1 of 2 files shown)


aaishwarymishra:
@vfdev-5 What do you think? like pr has some problems but should we add helper function like discussed in the issue or add it as parameter so I can ask for refactoring?

vfdev-5:
I think let's finish first with https://github.com/pytorch/ignite/pull/3619
Personally, I would prefer we add a helper function like discussed in the issue instead of a new arg.

PR #3439, opened 2025-08-29
Fix max iters issue and add tests

goanpeca:
Fixes #1521

Depends on:
- https://github.com/pytorch/ignite/pull/3442
- https://github.com/pytorch/ignite/pull/3443
- https://github.com/pytorch/ignite/pull/3444
- https://github.com/pytorch/ignite/pull/3445

PR #3729, opened 2026-04-10
Fix state_dict handling for max_iters in Engine

### LINKED ISSUE #1521: Possible issues with max_iters when loading/saving engine's state

TahaZahid05:
Fixes #1521 

Description:
This PR builds up on #3439 to implement `max_iters` handling in state serialization and deserialization during `Engine` runs.

Key Changes:
* `Engine.state_dict()` now correctly exports exactly *one* of `max_iters` or `max_epochs` depending on which condition the `Engine` run was configured with. 
* Reconstructed `_state_dict_one_of_opt_keys` to accept groups of mutually exclusive requirements, enabling `Engine.load_state_dict()` to cleanly accept either `iteration`/`max_iters` or `epoch`/`max_epochs` and correctly continue the engine's resume state from either.
* Added validation checks directly to `Engine` to prevent loading impossible future iterations while safely resuming training.
* Added local execution and Checkpointer tests that verify mutually exclusive cross-parameter resumption.

Check list:


ignite/base/mixins.py
```diff
@@ -1,7 +1,7 @@
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
 from collections.abc import Mapping
-from typing import Any, TYPE_CHECKING
+from typing import Any, Tuple, TYPE_CHECKING
```
vfdev-5: @TahaZahid05 let's not revert the typing hints

TahaZahid05:
@vfdev-5 done!

ignite/base/mixins.py
```diff
@@ -29,7 +29,7 @@ def attach(self, engine: "Engine", *args: Any, **kwargs: Any) -> None:
 
 class Serializable:
     _state_dict_all_req_keys: tuple = ()
-    _state_dict_one_of_opt_keys: tuple = ()
+    _state_dict_one_of_opt_keys: tuple = ((),)
```
vfdev-5:
```suggestion
    _state_dict_one_of_opt_keys: tuple[tuple] = ((),)
```
How about this to emphasize that it is tuple of tuples

ignite/engine/engine.py
```diff
@@ -6,7 +6,7 @@
 import weakref
 from collections import defaultdict, OrderedDict
 from collections.abc import Callable, Generator, Iterable, Iterator, Mapping
-from typing import Any
+from typing import Any, Optional
```
vfdev-5: Why do you reintro Optional?

ignite/engine/engine.py
```diff
@@ -773,6 +816,36 @@ def _is_done(state: State) -> bool:
         is_done_epochs = state.max_epochs is not None and state.epoch >= state.max_epochs
         return is_done_iters or is_done_count or is_done_epochs
 
+    def _check_and_set_max_epochs(self, max_epochs: Optional[int] = None) -> None:
```
vfdev-5: Can we make a common generic function for _check_and_set_max_epochs and _check_and_set_max_iters to avoid similar code twice?

tests/ignite/base/test_mixins.py
```diff
@@ -12,3 +14,257 @@ def test_state_dict():
...
+    # Test one-of optional keys - having all from another group
+    with pytest.raises(ValueError, match=r"should contain only one of '\('e', 'f'\)'"):
+        s.load_state_dict({"a": 1, "b": 2, "c": 3, "e": 5, "f": 6})
+
+    # Valid state dict
+    s.load_state_dict({"a": 1, "b": 2, "c": 3, "e": 5})
+    print("Valid state dict loaded successfully")
+
+
+def test_empty_optional_groups():
```
vfdev-5: Let's unify this test test_empty_optional_groups and test_multiple_empty_groups and test_mixed_empty_and_filled_groups using pytest parametrize.

Also, I do not see where we check for error raised for _state_dict_one_of_opt_keys containing empty tuple.

tests/ignite/engine/test_engine_state_dict.py
```diff
@@ -18,7 +18,7 @@ def test_state_dict():
     def _test(state):
         engine.state = state
         sd = engine.state_dict()
-        assert isinstance(sd, Mapping) and len(sd) == len(engine._state_dict_all_req_keys) + 1
+        assert isinstance(sd, Mapping) and len(sd) == len(engine._state_dict_all_req_keys) + 2
```
vfdev-5: Can you put a comment here explaining what is 2 here and if we can make it computable using engine, it would be ideal

tests/ignite/engine/test_engine_state_dict.py
```diff
@@ -274,3 +280,483 @@ def test_restart_training():
         state = engine.run(data, max_epochs=2)
     state.max_epochs = None
     engine.run(data, max_epochs=2)
+
+
+def test_state_dict_with_max_epochs():
```
vfdev-5: Same here about pytest parametrize usage, if we can reduce the number of lines in the test code, it would be great

ignite/engine/engine.py
TahaZahid05: Apologies, I'll follow modern type hints.

tests/ignite/base/test_mixins.py
TahaZahid05: @vfdev-5 If we raise error for _state_dict_one_of_opt_keys containing empty tuples then we should also change its initialization to a single tuple as well for it to pass the test? Is that fine?

```python
 # _state_dict_one_of_opt_keys: tuple = ((),)
 _state_dict_one_of_opt_keys: tuple = ()
```

vfdev-5:
Yes, it is fine, let's provide appropriate type hint for the structure

ignite/engine/engine.py
```diff
@@ -707,15 +707,25 @@ def save_engine(_):
...
                 a dictionary containing engine's state
 
+        .. versionchanged:: 0.5.5
+            Added support for serializing ``max_iters``.
+
         """
-        keys: tuple[str, ...] = self._state_dict_all_req_keys + (self._state_dict_one_of_opt_keys[0],)
+        keys: tuple[str, ...] = self._state_dict_all_req_keys
+        # We add iteration by default to get exact measure of progress
+        keys += ("iteration",)
```
vfdev-5, line 716: Why we do not add it to `self._state_dict_all_req_keys` directly ?

ignite/engine/engine.py
```diff
@@ -762,6 +775,36 @@ def load_state_dict(self, state_dict: Mapping) -> None:
...
+            if max_iters_value < 1:
+                raise ValueError("max_iters in state_dict is invalid. Please, set a correct max_iters positive value")
+            if max_iters_value < self.state.iteration:
+                raise ValueError(
+                    "max_iters in state_dict should be larger than or equal to the current iteration "
+                    f"defined in the state: {max_iters_value} vs {self.state.iteration}. "
+                )
+            self.state.max_iters = max_iters_value
+        else:
+            self.state.max_iters = None
```
vfdev-5, line 806: Can we refactor these two similar blocks?

ignite/engine/engine.py
```diff
@@ -918,18 +987,41 @@ def switch_batch(engine):
...
+                self.logger.info(
+                    f"Engine run resuming from iteration {self.state.iteration}, "
+                    f"epoch {self.state.epoch} until {self.state.max_iters} iterations"
+                )
+
             if self.state.epoch_length is None and data is None:
                 raise ValueError("epoch_length should be provided if data is None")
 
             if self.should_terminate:
-                # If engine was terminated and now is resuming from terminated state
```
vfdev-5, line 931: Why do you remove this comment?

ignite/base/mixins.py
```diff
@@ -29,7 +29,7 @@ def attach(self, engine: "Engine", *args: Any, **kwargs: Any) -> None:
 
 class Serializable:
     _state_dict_all_req_keys: tuple = ()
-    _state_dict_one_of_opt_keys: tuple = ()
+    _state_dict_one_of_opt_keys: tuple[tuple[str, ...], ...] = ()
```
vfdev-5, line 32: let's add the same level of type hint details to _state_dict_all_req_keys:
```python
_state_dict_all_req_keys: tuple[str, ...] = ()
```

ignite/engine/engine.py
```diff
@@ -128,8 +128,8 @@ def compute_mean_std(engine, batch):
 
     """
 
-    _state_dict_all_req_keys = ("epoch_length", "max_epochs")
-    _state_dict_one_of_opt_keys = ("iteration", "epoch")
+    _state_dict_all_req_keys = ("epoch_length",)
+    _state_dict_one_of_opt_keys = (("iteration", "epoch"), ("max_epochs", "max_iters"))
```
vfdev-5, line 132: Let's add the smae type hints here as in  Serializable

ignite/engine/engine.py
TahaZahid05: It seemed overcommenting to me as the code was quite clear. I'll revert it.

ignite/base/mixins.py
```diff
@@ -43,6 +43,17 @@ def load_state_dict(self, state_dict: Mapping) -> None:
...
+                raise ValueError(
+                    f"Empty group found in '{self.__class__.__name__}._state_dict_one_of_opt_keys'. "
+                    "Each group must contain at least one state attribute key."
+                )
+            opts = [k in state_dict for k in one_of_opt_keys]
+            num_present = sum(opts)
+            if num_present == 0:
+                raise ValueError(f"state_dict should contain at least one of '{one_of_opt_keys}' keys")
+            if num_present > 1:
+                raise ValueError(f"state_dict should contain only one of '{one_of_opt_keys}' keys")
```
aaishwarymishra, line 59: These 2 if else seems redundant and error message is almost similar, I think we can combine these 2 in a single if condition.

ignite/engine/engine.py
aaishwarymishra, line 716: If `iteration` is always getting included why not we just add it to `_state_dict_all_req_keys`

ignite/engine/engine.py
```diff
@@ -871,43 +939,44 @@ def switch_batch(engine):
...
 
-                epoch_length = self._get_data_length(data)
-                if epoch_length is not None and epoch_length < 1:
-                    raise ValueError("Input data has zero size. Please provide non-empty data")
+            # Set epoch_length for new state
+            if epoch_length is None:
+                # Try to get from data first, then fall back to existing state
+                if data is not None:
+                    epoch_length = self._get_data_length(data)
+                if epoch_length is None and self.state.epoch_length is not None:
```
aaishwarymishra, line 971: redundant `epoch_length is None` check

ignite/engine/engine.py
```diff
@@ -918,18 +987,41 @@ def switch_batch(engine):
             self.state.epoch_length = epoch_length
             # Reset generator if previously used
             self._internal_run_generator = None
-            self.logger.info(f"Engine run starting with max_epochs={max_epochs}.")
+
+            if self.state.max_epochs is not None:
+                self.logger.info(f"Engine run starting with max_epochs={self.state.max_epochs}.")
+            else:
+                self.logger.info(f"Engine run starting with max_iters={self.state.max_iters}.")
```
aaishwarymishra, line 994: you calculated `self.state.max_epochs` when epoch_length is present, this message is wrong if user didnt pass max_epoch but it got calculated above.

ignite/engine/engine.py
```diff
@@ -918,18 +987,41 @@ def switch_batch(engine):
...
+                self.logger.info(
+                    f"Engine run resuming from iteration {self.state.iteration}, "
+                    f"epoch {self.state.epoch} until {self.state.max_epochs} epochs"
+                )
+            else:
+                self.logger.info(
+                    f"Engine run resuming from iteration {self.state.iteration}, "
+                    f"epoch {self.state.epoch} until {self.state.max_iters} iterations"
+                )
+
```
aaishwarymishra, line 1020: same problem `state.max_iters` and `state.max_epoch` can exist together.

ignite/engine/engine.py
TahaZahid05: @vfdev-5 @aaishwarymishra Won't it be BC? any existing code that relies only on `epoch` rather than `iteration` may fail?

vfdev-5: _state_dict_all_req_keys is private attribute and external projects should not rely on ignite's private API

ignite/engine/engine.py
```diff
@@ -726,34 +735,90 @@ def load_state_dict(self, state_dict: Mapping) -> None:
...
             trainer = Engine(...)
             trainer.load_state_dict(state_dict)
             trainer.run(data)
 
+        .. versionchanged:: 0.5.5
+            Added support for restoring from a state dict containing ``max_iters`` instead of ``max_epochs``.
+
         """
+        if "iteration" in state_dict and "epoch" in state_dict:
+            # Check consistency if both are provided
```
vfdev-5: Should we really do this instead of simply raising the error that both should not be present?

ignite/engine/engine.py
```diff
@@ -726,34 +735,90 @@ def load_state_dict(self, state_dict: Mapping) -> None:
...
                 raise ValueError(
                     f"Required user state attribute '{k}' is absent in provided state_dict '{state_dict.keys()}'"
                 )
-        self.state.max_epochs = state_dict["max_epochs"]
-        self.state.epoch_length = state_dict["epoch_length"]
-        for k in self._state_dict_user_keys:
             setattr(self.state, k, state_dict[k])
+        self.state.epoch_length = state_dict["epoch_length"]
 
         if "iteration" in state_dict:
```
vfdev-5, line 788: Can we merge above code ` if "iteration" not in state_dict and "epoch" in state_dict:` and this code?

ignite/engine/engine.py
TahaZahid05: I was trying to make it more flexible but yeah I agree, it may be overkill

vfdev-5 (approved):
Let's move forward with this and see how it works 
LGTM, thanks @TahaZahid05

PR #3731, opened 2026-04-14
remove sklearn dependency from cohenkappa score calculation logic and applied custom calculation and updated tests

avishkarsonni:
Fixes #3701

  ## Description:
  Remove scikit-learn dependency from `CohenKappa` by implementing a native PyTorch version.

  - Replaced `sklearn.metrics.cohen_kappa_score` with a pure PyTorch implementation
  - Removed forced GPU→CPU transfer (`.cpu().numpy()`) — metric now runs fully on the configured device
  - Built confusion matrix with `torch.bincount` (single vectorised kernel) instead of a Python loop
  - Used `float64` throughout to match sklearn's numerical precision
  - Added explicit multilabel input validation (previously handled implicitly by sklearn
  
Check list:

### DIFF
--- a/ignite/metrics/cohen_kappa.py
+++ b/ignite/metrics/cohen_kappa.py
@@ -1,15 +1,67 @@
 from collections.abc import Callable
+from functools import partial
 from typing import Literal
 
 import torch
+import torch.nn.functional as F
 
+from ignite.exceptions import NotComputableError
+from ignite.metrics.confusion_matrix import ConfusionMatrix
 from ignite.metrics.epoch_metric import EpochMetric
+from ignite.metrics.metric import reinit__is_reduced
+
+
+def _kappa_from_conf(conf: torch.Tensor, weights: Literal["linear", "quadratic"] | None) -> float:
+    n = conf.sum()
+    if n == 0:
+        raise NotComputableError("CohenKappa cannot be computed on an empty confusion matrix (n == 0).")
+
+    n_classes = conf.shape[0]
+
+    if weights is None:
+        p_o = conf.trace() / n
+        row = conf.sum(dim=1)
+        col = conf.sum(dim=0)
+        p_e = (row * col).sum() / (n * n)
+    else:
+        idx = torch.arange(n_classes, device=conf.device)
+        if weights == "linear":
+            w = torch.abs(idx.unsqueeze(0) - idx.unsqueeze(1)).double()
+        else:
+            w = ((idx.unsqueeze(0) - idx.unsqueeze(1)) ** 2).double()
+
+        w = w / w.max()
+        p_o = 1 - (w * conf).sum() / n
+        row = conf.sum(dim=1)
+        col = conf.sum(dim=0)
+        expected = row.unsqueeze(1) * col.unsqueeze(0) / n
+        p_e = 1 - (w * expected).sum() / n
+
+    if (1 - p_e).abs() < 1e-9:
+        return 1.0 if (p_o - p_e).abs() < 1e-9 else float("nan")
+
+    return ((p_o - p_e) / (1 - p_e)).item()
+
+
+def _cohen_kappa_score(
+    y_pred: torch.Tensor,
+    y: torch.Tensor,
+    weights: Literal["linear", "quadratic"] | None,
+) -> float:
+    num_classes = int(max(y_pred.max().item(), y.max().item())) + 1
+
+    cm = ConfusionMatrix(num_classes=num_classes, device=y_pred.device)
+    y_pred_oh = F.one_hot(y_pred.long(), num_classes).float()
+    cm.update((y_pred_oh, y.long()))
+    conf = cm.compute().double()
+
+    return _kappa_from_conf(conf, weights)
 
 
 class CohenKappa(EpochMetric):
     """Compute different types of Cohen's Kappa: Non-Weighted, Linear, Quadratic.
-    Accumulating predictions and the ground-truth during an epoch and computing
-    Cohen's Kappa using native PyTorch operations via the formula:
+    Buffers predictions and targets across batches, then computes the confusion
+    matrix via :class:`~ignite.metrics.ConfusionMatrix` and applies the formula:
     κ = (p_o - p_e) / (1 - p_e), where p_o is the observed agreement and p_e
     is the expected agreement computed from marginal probabilities.
 
@@ -20,9 +72,6 @@ class CohenKappa(EpochMetric):
             you want to compute the metric with respect to one of the outputs.
         weights: a string is used to define the type of Cohen's Kappa whether Non-Weighted or Linear
             or Quadratic. Default, None.
-        check_compute_fn: Default False. If True, the compute function is run on the first batch
-            of data to ensure there are no issues. User will be warned in case there are any issues
-            computing the function.
         device: optional device specification for internal storage.
         skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
             true for multi-output model, for example, if ``y_pred`` contains multi-output as ``(y_pred_a, y_pred_b)``
@@ -61,52 +110,38 @@ class CohenKappa(EpochMetric):
         self,
         output_transform: Callable = lambda x: x,
         weights: Literal["linear", "quadratic"] | None = None,
-        check_compute_fn: bool = False,
         device: str | torch.device = torch.device("cpu"),
         skip_unrolling: bool = False,
     ):
         if weights not in (None, "linear", "quadratic"):
             raise ValueError("Kappa Weighting type must be None or linear or quadratic.")
 
-        # initialize weights
         self.weights: Literal["linear", "quadratic"] | None = weights
 
         super().__init__(
-            self._cohen_kappa_score,
+            compute_fn=partial(_cohen_kappa_score, weights=weights),
             output_transform=output_transform,
-            check_compute_fn=check_compute_fn,
+            check_compute_fn=False,
             device=device,
             skip_unrolling=skip_unrolling,
         )
 
-    def _cohen_kappa_score(self, y_targets: torch.Tensor, y_preds: torch.Tensor) -> float:
-        if y_targets.ndim > 1 or y_preds.ndim > 1:
+    @reinit__is_reduced
+    def update(self, output: tuple[torch.Tensor, torch.Tensor]) -> None:
+        y_pred, y = output[0].detach(), output[1].detach()
+
+        if y_pred.ndim == 2 and y_pred.shape[1] == 1:
+            y_pred = y_pred.squeeze(dim=-1)
+        if y.ndim == 2 and y.shape[1] == 1:
+            y = y.squeeze(dim=-1)
+
+        if y_pred.ndim > 1 or y.ndim > 1:
             raise ValueError("multilabel-indicator is not supported")
-        n_classes = int(max(y_targets.max().item(), y_preds.max().item())) + 1
 
-        indices = y_targets * n_classes + y_preds
-        conf = torch.bincount(indices, minlength=n_classes * n_classes).reshape(n_classes, n_classes).double()
-        n = conf.sum()
+        super().update((y_pred, y))
 
-        if self.weights is None:
-            p_o = conf.trace() / n
-            row = conf.sum(dim=1)
-            col = conf.sum(dim=0)
-            p_e = (row * col).sum() / (n * n)
-
-        else:
-            idx = torch.arange(n_classes, device=y_targets.device)
-            if self.weights == "linear":
-                w = torch.abs(idx.unsqueeze(0) - idx.unsqueeze(1)).double()
-            else:
-                w = ((idx.unsqueeze(0) - idx.unsqueeze(1)) ** 2).double()
-
-            w = w / w.max()
-            p_o = 1 - (w * conf).sum() / n
-            row = conf.sum(dim=1)
-            col = conf.sum(dim=0)
-            expected = row.unsqueeze(1) * col.unsqueeze(0) / n
-            p_e = 1 - (w * expected).sum() / n
-
-        kappa = (p_o - p_e) / (1 - p_e)
-        return kappa.item()
+    def compute(self) -> float:
+        try:
+            return super().compute()
+        except NotComputableError:
+            raise NotComputableError("CohenKappa must have at least one example before it can be computed.")
--- a/ignite/metrics/cohen_kappa.py
+++ b/ignite/metrics/cohen_kappa.py
@@ -187,8 +187,6 @@ class CohenKappa(Metric):
 
     .. versionchanged:: 0.6.0
         Replaced scikit-learn dependency with a native PyTorch implementation.
-
-    .. versionchanged:: 0.6.2
         Added ``num_classes`` argument; routes to a running-confusion-matrix backend when provided.
     """
 
--- a/ignite/metrics/cohen_kappa.py
+++ b/ignite/metrics/cohen_kappa.py
@@ -37,10 +37,8 @@ def _kappa_from_conf(conf: torch.Tensor, weights: Literal["linear", "quadratic"]
         expected = row.unsqueeze(1) * col.unsqueeze(0) / n
         p_e = 1 - (w * expected).sum() / n
 
-    if (1 - p_e).abs() < 1e-9:
-        return 1.0 if (p_o - p_e).abs() < 1e-9 else float("nan")
-
-    return ((p_o - p_e) / (1 - p_e)).item()
+    epsilon = 1e-9
+    return ((p_o - p_e) / (1 - p_e).clamp(min=epsilon)).item()
 
 
 def _cohen_kappa_score(
@@ -48,6 +46,9 @@ def _cohen_kappa_score(
     y: torch.Tensor,
     weights: Literal["linear", "quadratic"] | None,
 ) -> float:
+    if y_pred.ndim > 1 or y.ndim > 1:
+        raise ValueError("multilabel-indicator is not supported")
+
     num_classes = int(max(y_pred.max().item(), y.max().item())) + 1
 
     cm = ConfusionMatrix(num_classes=num_classes, device=y_pred.device)
@@ -85,16 +86,7 @@ class _CohenKappaEpochMetric(EpochMetric):
         if y.ndim == 2 and y.shape[1] == 1:
             y = y.squeeze(dim=-1)
 
-        if y_pred.ndim > 1 or y.ndim > 1:
-            raise ValueError("multilabel-indicator is not supported")
-
         super().update((y_pred, y))
-
-    def compute(self) -> float:
-        try:
-            return super().compute()
-        except NotComputableError:
-            raise NotComputableError("CohenKappa must have at least one example before it can be computed.")
 
 
 class _CohenKappaConfusionMatrix(Metric):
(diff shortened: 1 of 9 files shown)


aaishwarymishra:
Thanks for the pr! In this metric we inherited form the `EpochMetric` because we need to send all the values to the `scikit-learn`. We are removing the dependency so I think we should inherit from `Metric` or a subclass of it and update there `update` method to efficiently update the 
```
po: Relative observed agreement among raters
pe: Hypothetical probability of chance agreement
```
instead of using raw tensor values for final computation in `compute`.

avishkarsonni:
so there is no need to bring all the raw tensors to the memory because there is no scikit-learn now, got it

aaishwarymishra:
@avishkarsonni yes, instead of storing raw tensors in `update` we can just update the variables used for the calculation of `po` and `pe`, on `compute` we calculate these values from those variables and calculate the metric.
It saves memory as we are not storing tensors and compute as we are not performing a big calculation on huge tensors at end.
You can look at other metrics to get reference how its done on non `EpochMetric`.

Prathamesh8989:
Nice implementation! Using torch.bincount for the confusion matrix is a good approach and keeps the computation on GPU.

I had a couple of small observations:

It may be safer to guard against the case where p_e == 1, since (1 - p_e) would cause a division by zero.
If n == 0 (empty batch), the computation would also fail due to division.
It might be worth explicitly ensuring the confusion matrix tensor stays on the same device as the inputs.

Overall the implementation looks clean and matches the expected Cohen’s Kappa formulation.

ignite/metrics/cohen_kappa.py
```diff
@@ -81,8 +80,33 @@ def __init__(
...
-
-        y_true = y_targets.cpu().numpy()
-        y_pred = y_preds.cpu().numpy()
-        return cohen_kappa_score(y_true, y_pred, weights=self.weights)
+        if y_targets.ndim > 1 or y_preds.ndim > 1:
+            raise ValueError("multilabel-indicator is not supported")
+        n_classes = int(max(y_targets.max().item(), y_preds.max().item())) + 1
+
+        indices = y_targets * n_classes + y_preds
+        conf = torch.bincount(indices, minlength=n_classes * n_classes).reshape(n_classes, n_classes).double()
```
vfdev-5: Why can't we use ignite's [confusion matrix class](https://github.com/pytorch/ignite/blob/c0ceca5017fdb4fb9c11f00ad7e72d51efb942a2/ignite/metrics/confusion_matrix.py#L13) to compute this metric?

If we use `ConfusionMatrix`, we should provide `num_classes` as input argument... In order to keep backward compatibility, we can add num_classes arg to the constructor as optional kwargs. 
We can have two private cohen kappa implementations: one using EpochMetric (current one) and second using ConfusionMatrix. The public CohenKappa can route depending on `num_classes` arg.

In the private implementation using EpochMetric we should still use ConfusionMatrix to compute the confusion matrix in `compute()` method instead of doing that manually (as currently) to avoid bugs.

avishkarsonni: I used `bincount` as I was aware of it and from start planned as such but now as mentioned it is possible. in the start of this PR it was not intuitive to me as it needs `num_classes` at init. Current `impl` infers it dynamically from data, but it is indeed possible.

avishkarsonni: And again this method is better only when the user knows `num_classes` if not we will need a fall back to infer it from the data on its own. Both of the methods have same time complexity, just advantage of using `ConfusionMatrix` is edge case coverage over current approach, would love to hear what would be a better approach!

vfdev-5: Benefit of using `ConfusionMatrix` is that we do not store the list of y_preds and y_true in RAM vs the current implementation using `EpochMetric`. Drawback of using `ConfusionMatrix` is that we should specify `num_classes` in the metric constructor and also it gives a backward compatibility break.

OK, the suggestion for this PR: let's use `ConfusionMatrix` in the `_cohen_kappa_score`  in order to avoid using manual confusion matrix computation. We can instantiate an object of ConfusionMatrix with the number of classes and do a single `update` and call `compute`.

ignite/metrics/cohen_kappa.py
```diff
@@ -1,15 +1,67 @@
...
+    y_pred: torch.Tensor,
+    y: torch.Tensor,
+    weights: Literal["linear", "quadratic"] | None,
+) -> float:
+    num_classes = int(max(y_pred.max().item(), y.max().item())) + 1
+
+    cm = ConfusionMatrix(num_classes=num_classes, device=y_pred.device)
+    y_pred_oh = F.one_hot(y_pred.long(), num_classes).float()
+    cm.update((y_pred_oh, y.long()))
+    conf = cm.compute().double()
```
vfdev-5: Passing to double like this wont work on MPS

avishkarsonni: but then the precision issue comes up as MPS needs float32 and CUDA goes for float64, I've been picking at this issue for some time, can I use `torch.promote_types`  to pick highest precision? or that would create an overhead? or there is another way to add a try-catch block like if it rejects `double()` it goes to `float()`?

aaishwarymishra: having different precision for the mps and cuda will be weird , either we can stick with `.float()` or convert the gpu tensors to cpu tensors and then call the `.double()`

avishkarsonni: the second one is what I am going with, can do good with better precision

vfdev-5 (commented):
@avishkarsonni thanks a lot for the update of the code using the suggested approach.
However, I'm not totally agree with Aaishwarya about passing to cpu from an accelerator to perform computation on double. If device is cuda supporting double, passing to cpu is not optimal.
I think we can use _double_dtype for that: https://github.com/pytorch/ignite/blob/c830a06169ea2b57da040656fa3200124fbf10bc/ignite/metrics/metric.py#L389-L390

ignite/metrics/cohen_kappa.py
vfdev-5: I suggest to use different precisions double for cpu/cuda and float on mps as it does not support double using metric._double_dtype arg: https://github.com/pytorch/ignite/blob/c830a06169ea2b57da040656fa3200124fbf10bc/ignite/metrics/metric.py#L389-L390

aaishwarymishra:
> @avishkarsonni thanks a lot for the update of the code using the suggested approach. However, I'm not totally agree with Aaishwarya about passing to cpu from an accelerator to perform computation on double. If device is cuda supporting double, passing to cpu is not optimal. I think we can use _double_dtype for that:
> 
> https://github.com/pytorch/ignite/blob/c830a06169ea2b57da040656fa3200124fbf10bc/ignite/metrics/metric.py#L389-L390

I am not sure what you mean exactly, 
if you mean use '.double()' when cuda in available and fallback to cpu for mps I think its the better approach then mine initial one.

vfdev-5:
The suggestion is to use `self._double_dtype` in places where we need to cast a tensor to double dtype, like here:
https://github.com/pytorch/ignite/blob/c830a06169ea2b57da040656fa3200124fbf10bc/ignite/metrics/multilabel_confusion_matrix.py#L137-L138

avishkarsonni:
I too agree that wherever we can use the cuda acceleration we should use it, but it again creates a situation where we have to treat mps and cuda differently and cannot make the code agnostic, I don't know what is the better approach

vfdev-5:
it is just a precision difference, we do not treat them essentially differently, so it is fine to use my suggestion.
Once mps supports double so we can easily make it work.

TahaZahid05:
@avishkarsonni tests are failing [https://app.netlify.com/projects/pytorch-ignite-preview/deploys/69fcbba33087ea0008ff7a24](url).

You may follow the steps given in `CONTRIBUTING.MD` as well to ensure all test pass before pushing.

ignite/metrics/cohen_kappa.py
```diff
@@ -1,16 +1,161 @@
...
+    y_pred: torch.Tensor,
+    y: torch.Tensor,
+    weights: Literal["linear", "quadratic"] | None,
+) -> float:
+    num_classes = int(max(y_pred.max().item(), y.max().item())) + 1
+
+    cm = ConfusionMatrix(num_classes=num_classes, device=y_pred.device)
+    y_pred_oh = F.one_hot(y_pred.long(), num_classes).float()
+    cm.update((y_pred_oh, y.long()))
+    double_dtype = torch.float32 if y_pred.device.type == "mps" else torch.float64
```
vfdev-5: Can't we fetch double dtype from `cm._double_dtype`?

tests/ignite/metrics/test_cohen_kappa.py
```diff
@@ -14,55 +12,15 @@
...
-        ck.update((torch.randint(0, 5, size=(4, 3)), torch.randint(0, 2, size=(4, 3))))
-
-    with pytest.raises(ValueError, match=r"Incoherent types between input y and stored targets"):
-        ck.update((torch.rand(4, 3), torch.randint(0, 2, size=(4, 3)).to(torch.int32)))
-
-    with pytest.raises(ValueError, match=r"Incoherent types between input y_pred and stored predictions"):
-        ck.update((torch.randint(0, 2, size=(10,)).long(), torch.randint(0, 2, size=(10, 5)).long()))
-
-
-def test_check_shape():
```
vfdev-5, line 53: Why do you remove these tests?

tests/ignite/metrics/test_cohen_kappa.py
```diff
@@ -14,55 +12,15 @@
...
     ck = CohenKappa()
 
     with pytest.raises(
-        NotComputableError, match=r"EpochMetric must have at least one example before it can be computed"
+        NotComputableError, match=r"CohenKappa must have at least one example before it can be computed"
     ):
         ck.compute()
 
 
-def test_input_types():
```
vfdev-5, line 37: Same question here, why do you remove these tests?

tests/ignite/metrics/test_cohen_kappa.py
avishkarsonni: I accidentally commited it i was trying to do something with git and it pushed the commited changes

tests/ignite/metrics/test_cohen_kappa.py
```diff
@@ -14,53 +12,40 @@
...
+        NotComputableError, match=r"CohenKappa must have at least one example before it can be computed"
     ):
         ck.compute()
 
 
 def test_input_types():
     ck = CohenKappa()
     ck.reset()
-    output1 = (torch.rand(4, 3), torch.randint(0, 2, size=(4, 3), dtype=torch.long))
+    output1 = (torch.rand(10), torch.randint(0, 2, size=(10,), dtype=torch.long))
```
vfdev-5: @avishkarsonni can you please revert all the changes in this test

ignite/metrics/cohen_kappa.py
```diff
@@ -1,16 +1,160 @@
...
+        p_o = 1 - (w * conf).sum() / n
+        row = conf.sum(dim=1)
+        col = conf.sum(dim=0)
+        expected = row.unsqueeze(1) * col.unsqueeze(0) / n
+        p_e = 1 - (w * expected).sum() / n
+
+    if (1 - p_e).abs() < 1e-9:
+        return 1.0 if (p_o - p_e).abs() < 1e-9 else float("nan")
+
+    return ((p_o - p_e) / (1 - p_e)).item()
```
vfdev-5: Can we do this using a small epsilon in the denominator?

```suggestion
    return ((p_o - p_e) / (1 - p_e + 1e-9)).item()
```

ignite/metrics/cohen_kappa.py
```diff
@@ -1,16 +1,160 @@
...
+        if y_pred.ndim > 1 or y.ndim > 1:
+            raise ValueError("multilabel-indicator is not supported")
+
+        super().update((y_pred, y))
+
+    def compute(self) -> float:
+        try:
+            return super().compute()
+        except NotComputableError:
+            raise NotComputableError("CohenKappa must have at least one example before it can be computed.")
```
vfdev-5: Can we remove this and instead update EpochMetric compute method that it mentions the child class in the error instead of EpochMetric?

ignite/metrics/cohen_kappa.py
```diff
@@ -1,16 +1,160 @@
...
+        if y_pred.ndim > 1 or y.ndim > 1:
+            raise ValueError("multilabel-indicator is not supported")
+
+        num_classes = self._cm.num_classes
+        y_pred_oh = F.one_hot(y_pred.long(), num_classes).float().to(self._device)
+        self._cm.update((y_pred_oh, y.long().to(self._device)))
+
+    def compute(self) -> float:
+        if self._cm.confusion_matrix.sum() == 0:
+            raise NotComputableError("CohenKappa must have at least one example before it can be computed.")
```
vfdev-5: This check is also done in _kappa_from_conf, maybe we can remove it here?

vfdev-5 (commented):
Thanks for working on this PR @avishkarsonni 
I made few more comments, please take a look

tests/ignite/metrics/test_cohen_kappa.py
avishkarsonni: just in this test or the whole testfile is to be reverted?

avishkarsonni: the issue with changing this test back is the the method now take only single dimension as  input and is not able to handle 2D inputs, what should i Do hear? is the method not correct? @vfdev-5 @aaishwarymishra

ignite/metrics/cohen_kappa.py
avishkarsonni: just for cohenkappa class or for any other class as well?

vfdev-5: I talk about this _CohenKappaEpochMetric to remove the check and update only EpochMetric such the check in derived classes show the derived class name and not EpochMetric

tests/ignite/metrics/test_cohen_kappa.py
vfdev-5: It means that we have a regression now and this is not good. We should not introduce a regression when updating the internal implementation

ignite/metrics/cohen_kappa.py
```diff
@@ -52,37 +184,52 @@ class CohenKappa(EpochMetric):
 
     .. versionchanged:: 0.5.1
         ``skip_unrolling`` argument is added.
+
+    .. versionchanged:: 0.6.0
+        Replaced scikit-learn dependency with a native PyTorch implementation.
+
+    .. versionchanged:: 0.6.2
+        Added ``num_classes`` argument; routes to a running-confusion-matrix backend when provided.
```
vfdev-5, line 200:
```suggestion
    .. versionchanged:: 0.6.0
        Replaced scikit-learn dependency with a native PyTorch implementation.
        Added ``num_classes`` argument; routes to a running-confusion-matrix backend when provided.
```

ignite/metrics/cohen_kappa.py
```diff
@@ -52,37 +184,50 @@ class CohenKappa(EpochMetric):
...
+    .. versionchanged:: 0.6.0
+        Replaced scikit-learn dependency with a native PyTorch implementation.
+        Added ``num_classes`` argument; routes to a running-confusion-matrix backend when provided.
     """
 
     def __init__(
         self,
         output_transform: Callable = lambda x: x,
         weights: Literal["linear", "quadratic"] | None = None,
-        check_compute_fn: bool = False,
```
vfdev-5, line 61: We should not remove this arg and should propagate it to `_CohenKappaEpochMetric`

ignite/metrics/cohen_kappa.py
```diff
@@ -52,37 +188,52 @@ class CohenKappa(EpochMetric):
...
         device: str | torch.device = torch.device("cpu"),
         skip_unrolling: bool = False,
+        num_classes: int | None = None,
+        check_compute_fn: bool = False,
```
vfdev-5: you should not move this arg from its original position in order to keep BC, move it where it was and same for the docstring, move it where it was

vfdev-5 (approved):
LGTM, thanks a lot @avishkarsonni !

avishkarsonni:
looking forward to contribute faster :laughing:

vfdev-5:
@avishkarsonni seems like there are related failures, can you check the CI logs?

avishkarsonni:
the changes boil down to rounding errors which came for using `cm.double_dtype` on MPS float32 precision, the issue is now using `.cpu.double()` or `torch.bincount()`

vfdev-5:
> the changes boil down to rounding errors which came for using `cm.double_dtype` on MPS float32 precision, the issue is now using `.cpu.double()` or `torch.bincount()`

You can increase the tolerance in the tests for mps, you can find a similar code in some other tests.

vfdev-5:
@avishkarsonni cohen kappa metric does not pass distributed tests:
```
2026-06-10T08:46:29.5561476Z FAILED tests/ignite/metrics/test_cohen_kappa.py::test_distrib_gloo_cpu_or_gpu - RuntimeError: [/__w/pytorch/pytorch/third_party/gloo/gloo/transport/tcp/unbound_buffer.cc:78] Timed out waiting 30000ms for recv operation to complete
2026-06-10T08:46:29.5562199Z FAILED tests/ignite/metrics/test_cohen_kappa.py::test_distrib_gloo_cpu_or_gpu - RuntimeError: [/__w/pytorch/pytorch/third_party/gloo/gloo/transport/tcp/unbound_buffer.cc:78] Timed out waiting 30000ms for recv operation to complete
2026-06-10T08:46:29.5563121Z FAILED tests/ignite/metrics/test_cohen_kappa.py::test_distrib_gloo_cpu_or_gpu - RuntimeError: [/__w/pytorch/pytorch/third_party/gloo/gloo/transport/tcp/unbound_buffer.cc:78] Timed out waiting 30000ms for recv operation to complete
2026-06-10T08:46:29.5563848Z FAILED tests/ignite/metrics/test_cohen_kappa.py::test_distrib_gloo_cpu_or_gpu - RuntimeError: [/__w/pytorch/pytorch/third_party/gloo/gloo/transport/tcp/unbound_buffer.cc:129] Timed out waiting 30000ms for send operation to complete
```

avishkarsonni:
this looks like an infra issue rather than a code issue, because same tests passes on py3.11 and py3.12, can we rerun the tests?

ignite/metrics/cohen_kappa.py
```diff
@@ -51,10 +51,16 @@ def _cohen_kappa_score(
...
-    conf = cm.compute().to(dtype=cm._double_dtype)
+    # Build the confusion matrix locally with plain tensor ops. We must NOT use the
+    # ``ConfusionMatrix`` metric here: its ``compute()`` is wrapped with ``sync_all_reduce``
+    # and this function runs on rank 0 only (see ``EpochMetric.compute``), so a collective
+    # would deadlock the other ranks.
+    y, y_pred = y.long(), y_pred.long()
+    indices = num_classes * y + y_pred
+    conf = torch.bincount(indices, minlength=num_classes**2)
+    # MPS framework doesn't support float64, fall back to float32 (mirrors Metric._double_dtype)
+    double_dtype = torch.float32 if y_pred.device.type == "mps" else torch.float64
```
vfdev-5, line 62: Pass _double_dtype in this function instead of copying the code.

PR #3732, opened 2026-04-14
feat: add class_names support to Precision, Recall, FBeta metrics

### LINKED ISSUE #1466: Add ability to add class name for metrics
Hello,

I have a code that does classification on CIFAR dataset. 

When I use the Recall metric, what I get is a graph of per-class recall, but the legend gives me the label enum of the class and not the name.
It would be great to be able to give a list of class names to such metric so it'll be reported with the name.

aniketpandey05:
Fixes #1466

Description:
Adds an optional `class_names` parameter to `_BasePrecisionRecall`, allowing `compute()` to return a labeled `dict` instead of an unnamed tensor when `average=False` or `average=None`. Useful for per-class metric tracking where knowing which score belongs to which class matters for logging and visualization.

Check list:

### DIFF
--- a/ignite/metrics/fbeta.py
+++ b/ignite/metrics/fbeta.py
@@ -177,6 +177,16 @@ def Fbeta(
     elif recall._average:
         raise ValueError("Input recall metric should have average=False")
 
+    if precision._class_names is not None:
+
+        def _fbeta_with_class_names(p: dict, r: dict) -> dict:
+            p_vals = torch.tensor(list(p.values()))
+            r_vals = torch.tensor(list(r.values()))
+            scores = (1.0 + beta**2) * p_vals * r_vals / (beta**2 * p_vals + r_vals + 1e-15)
+            return dict(zip(precision._class_names, scores.tolist()))
+
+        return MetricsLambda(_fbeta_with_class_names, precision, recall)
+
     fbeta = (1.0 + beta**2) * precision * recall / (beta**2 * precision + recall + 1e-15)
 
     if average:
--- a/ignite/metrics/fbeta.py
+++ b/ignite/metrics/fbeta.py
@@ -17,6 +17,7 @@ def Fbeta(
     recall: Recall | None = None,
     output_transform: Callable | None = None,
     device: str | torch.device | None = None,
+    class_names: list[str] | None = None,
 ) -> MetricsLambda:
     r"""Calculates F-beta score.
 
@@ -42,6 +43,7 @@ def Fbeta(
         device: specifies which device updates are accumulated on. Setting the metric's
             device to be the same as your ``update`` arguments ensures the ``update`` method is non-blocking. By
             default, CPU.
+        class_names: list of class name strings used to label per-class output. Default: ``None``.
 
     Returns:
         MetricsLambda, F-beta metric
@@ -140,6 +142,9 @@ def Fbeta(
         .. testoutput:: 4
 
             0.7499...
+
+    .. versionchanged:: 0.6.0
+        ``class_names`` argument is added.
     """
     if not (beta > 0):
         raise ValueError(f"Beta should be a positive integer, but given {beta}")
@@ -159,23 +164,61 @@ def Fbeta(
     if precision is None and recall is None and device is None:
         device = torch.device("cpu")
 
+    if class_names is not None:
+        if not isinstance(class_names, (list, tuple)) or not all(isinstance(n, str) for n in class_names):
+            raise ValueError("class_names must be a list of strings")
+        if average not in (False, None):
+            raise ValueError(
+                f"class_names is only applicable when average=False or average=None, got average={average!r}."
+            )
+
+    if precision is not None and precision._average:
+        raise ValueError("Input precision metric should have average=False")
+
+    if recall is not None and recall._average:
+        raise ValueError("Input recall metric should have average=False")
+
+    active_metrics = [m for m in (precision, recall) if m is not None]
+
+    if class_names is not None and any(m._class_names != class_names for m in active_metrics):
+        raise ValueError("precision and recall metric class_names must match Fbeta class_names")
+
+    if len(active_metrics) == 2 and active_metrics[0]._class_names != active_metrics[1]._class_names:
+        raise ValueError("precision and recall class_names must match")
+
+    if class_names is None and precision is not None:
+        class_names = precision._class_names
+    if class_names is None and recall is not None:
+        class_names = recall._class_names
+
     if precision is None:
         precision = Precision(
             output_transform=(lambda x: x) if output_transform is None else output_transform,
             average=False,
             device=cast(str | torch.device, recall._device if recall else device),
+            class_names=class_names,
         )
-    elif precision._average:
-        raise ValueError("Input precision metric should have average=False")
 
     if recall is None:
         recall = Recall(
             output_transform=(lambda x: x) if output_transform is None else output_transform,
             average=False,
             device=cast(str | torch.device, precision._device if precision else device),
+            class_names=class_names,
         )
-    elif recall._average:
-        raise ValueError("Input recall metric should have average=False")
+
+    if class_names is not None:
+
+        def _fbeta_with_class_names(p: dict, r: dict) -> dict:
+            p_vals = torch.tensor(list(p.values()))
+            r_vals = torch.tensor(list(r.values()))
+            scores = (1.0 + beta**2) * p_vals * r_vals / (beta**2 * p_vals + r_vals + 1e-15)
+            values = scores.tolist()
+            if not isinstance(values, list):
+                values = [values]
+            return dict(zip(class_names, values))
+
+        return MetricsLambda(_fbeta_with_class_names, precision, recall)
 
     fbeta = (1.0 + beta**2) * precision * recall / (beta**2 * precision + recall + 1e-15)
 
--- a/ignite/metrics/fbeta.py
+++ b/ignite/metrics/fbeta.py
@@ -17,6 +17,7 @@ def Fbeta(
     recall: Recall | None = None,
     output_transform: Callable | None = None,
     device: str | torch.device | None = None,
+    class_names: list[str] | None = None,
 ) -> MetricsLambda:
     r"""Calculates F-beta score.
 
@@ -42,6 +43,7 @@ def Fbeta(
         device: specifies which device updates are accumulated on. Setting the metric's
             device to be the same as your ``update`` arguments ensures the ``update`` method is non-blocking. By
             default, CPU.
+        class_names: list of class name strings used to label per-class output. Default: ``None``.
 
     Returns:
         MetricsLambda, F-beta metric
@@ -159,31 +161,53 @@ def Fbeta(
     if precision is None and recall is None and device is None:
         device = torch.device("cpu")
 
+    if class_names is not None:
+        if not isinstance(class_names, (list, tuple)) or not all(isinstance(n, str) for n in class_names):
+            raise ValueError("class_names must be a list of strings")
+
+    if precision is not None and recall is not None:
+        p_cn = getattr(precision, "_class_names", None)
+        r_cn = getattr(recall, "_class_names", None)
+        if p_cn is not None and r_cn is not None and p_cn != r_cn:
+            raise ValueError("precision and recall class_names must match")
+
+    target_class_names = (
+        class_names or getattr(precision, "_class_names", None) or getattr(recall, "_class_names", None)
+    )
+
     if precision is None:
         precision = Precision(
             output_transform=(lambda x: x) if output_transform is None else output_transform,
             average=False,
             device=cast(str | torch.device, recall._device if recall else device),
+            class_names=target_class_names,
         )
     elif precision._average:
         raise ValueError("Input precision metric should have average=False")
+    elif target_class_names is not None:
+        precision._class_names = target_class_names
 
     if recall is None:
         recall = Recall(
             output_transform=(lambda x: x) if output_transform is None else output_transform,
             average=False,
             device=cast(str | torch.device, precision._device if precision else device),
+            class_names=target_class_names,
         )
     elif recall._average:
         raise ValueError("Input recall metric should have average=False")
+    elif target_class_names is not None:
+        recall._class_names = target_class_names
 
-    if precision._class_names is not None:
+    if target_class_names is not None:
 
         def _fbeta_with_class_names(p: dict, r: dict) -> dict:
             p_vals = torch.tensor(list(p.values()))
             r_vals = torch.tensor(list(r.values()))
             scores = (1.0 + beta**2) * p_vals * r_vals / (beta**2 * p_vals + r_vals + 1e-15)
-            return dict(zip(precision._class_names, scores.tolist()))
+            if scores.ndim == 0:
+                return {target_class_names[0]: scores.item()}
+            return dict(zip(target_class_names, scores.tolist()))
 
         return MetricsLambda(_fbeta_with_class_names, precision, recall)
 
--- a/ignite/metrics/fbeta.py
+++ b/ignite/metrics/fbeta.py
@@ -165,15 +165,28 @@ def Fbeta(
         if not isinstance(class_names, (list, tuple)) or not all(isinstance(n, str) for n in class_names):
             raise ValueError("class_names must be a list of strings")
 
-    if precision is not None and recall is not None:
-        p_cn = getattr(precision, "_class_names", None)
-        r_cn = getattr(recall, "_class_names", None)
-        if p_cn is not None and r_cn is not None and p_cn != r_cn:
-            raise ValueError("precision and recall class_names must match")
-
     target_class_names = (
         class_names or getattr(precision, "_class_names", None) or getattr(recall, "_class_names", None)
     )
+
+    if target_class_names is not None and average is not False and average is not None:
+        raise ValueError(f"class_names is only applicable when average=False or average=None, got average={average!r}.")
+
+    if precision is not None:
+        if precision._average:
+            raise ValueError("Input precision metric should have average=False")
+        if class_names is not None and precision._class_names != class_names:
+            raise ValueError("precision metric class_names must match Fbeta class_names")
+
+    if recall is not None:
+        if recall._average:
+            raise ValueError("Input recall metric should have average=False")
+        if class_names is not None and recall._class_names != class_names:
+            raise ValueError("recall metric class_names must match Fbeta class_names")
+
+    if precision is not None and recall is not None:
+        if precision._class_names != recall._class_names:
+            raise ValueError("precision and recall class_names must match")
 
     if precision is None:
         precision = Precision(
@@ -182,10 +195,6 @@ def Fbeta(
             device=cast(str | torch.device, recall._device if recall else device),
             class_names=target_class_names,
         )
-    elif precision._average:
-        raise ValueError("Input precision metric should have average=False")
-    elif target_class_names is not None:
-        precision._class_names = target_class_names
 
     if recall is None:
         recall = Recall(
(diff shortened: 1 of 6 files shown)


aaishwarymishra:
ok so like wont this code break `fbeta` metric when precision and recall have `average=False` and they have `class_names`, we need to update the `fbeta` too to support this.

There can be more metrics that can benefit from this change. I am not sure though.

ignite/metrics/precision.py
```diff
@@ -157,6 +167,13 @@ def compute(self) -> torch.Tensor | float:
         elif self._average == "macro":
             return cast(torch.Tensor, fraction).mean().item()
         else:
+            if self._class_names is not None:
+                if len(self._class_names) != fraction.shape[0]:
+                    raise ValueError(
+                        f"class_names has {len(self._class_names)} entries but the metric computed "
+                        f"{fraction.shape[0]} classes."
+                    )
```
vfdev-5: Can we do this once test in the update instead of compute as it would be a poor UX when the whole slow eval loop was done and it just failed here due to some minor issue with the number of class names

rogueslasher:
@aaishwarymishra would raising errors for those condition a better option or should we add full class_name support to fbeta

aaishwarymishra:
I am not sure, adding support for the class names in `fbeta` would be easier, but `fbeta` uses the `MetricsLambda` for overloading operator we can update them, I am not sure @vfdev-5  what you think would be appropriate?

steaphenai:
Could we please clarify `Fbeta` behavior when `class_names` is used with `average=False/None`

vfdev-5:
> I am not sure, adding support for the class names in `fbeta` would be easier, but `fbeta` uses the `MetricsLambda` for overloading operator we can update them, I am not sure @vfdev-5 what you think would be appropriate?

If we can compute F1 score per class than we should propagate the dict structure until the output of the F1 otherwise, take `.values()` of the dict output from precision and recall.

aniketpandey05:
Hey @vfdev-5 , would adding a  helper in metric.py to handle dict arithmetic in all operator overloads so the dict propagates through to the Fbeta output be a good approach? Also when average=True and class_names is set, should Fbeta return a scalar float or propagate the dict as is?

vfdev-5:
> Hey @vfdev-5 , would adding a helper in metric.py to handle dict arithmetic in all operator overloads so the dict propagates through to the Fbeta output be a good approach? Also when average=True and class_names is set, should Fbeta return a scalar float or propagate the dict as is?

I do not know, what would you suggest? Let's try to make it the most simple possible and intuitively clear.

aniketpandey05:
hey @vfdev-5 I think Fbeta should return a per-class dict if class_names is set, regardless of average.

vfdev-5:
@rogueslasher ok, let's implement that this way and see if this works as expected

aniketpandey05:
Hey @vfdev-5, would adding a _dict_aware_op helper in metric.py and updating all operator overloads to use it for element-wise dict arithmetic, and skipping average in fbeta.py when class_names is set be a good idea? It touches all operator overloads in metric.py which affects all metrics. 

i am worried since it touches all operators overloads

aniketpandey05:
@aaishwarymishra ,would love your help in this can you look into it more please

aaishwarymishra:
@rogueslasher well , I am not sure to be honest i wont touch the metric operators as it kind of affect every metric, for a case which can be only relevant for very few metrics, I think we should modify  `Fbeta` directly.

aniketpandey05:
@aaishwarymishra the header rules and other three preview checks are taking insane amount of time . is there any issue

aaishwarymishra:
Documentation issue :)

aniketpandey05:
@aaishwarymishra

ignite/metrics/recall.py
```diff
@@ -11,213 +11,220 @@
...
+                default, CPU.
+            skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
+                true for multi-output model, for example, if ``y_pred`` contains multi-output as ``(y_pred_a, y_pred_b)``
+                Alternatively, ``output_transform`` can be used to handle this.
+            class_names: list of class name strings used to label per-class output when ``average=False``
+    <<<<<<< HEAD
+                or ``average=None``. If provided, ``compute()`` returns a ``dict`` mapping each class
+    =======
+                or ``average=None``. If provided, :meth:`compute` returns a ``dict`` mapping each class
+    >>>>>>> 7ea2eb50590597780ffae1ab3dc4c7ede3c80be7
```
aaishwarymishra: I think its from the conflict you forgot to resolve

ignite/metrics/recall.py
aaishwarymishra, line 1: I think you should try to pre-commit we recently moved from the black formatter to ruff, that may be the reason for docstring problem but not sure.

tests/ignite/metrics/test_precision.py
aaishwarymishra, line 1: There should be tests for recall too

aniketpandey05:
@aaishwarymishra

ignite/metrics/fbeta.py
```diff
@@ -159,23 +161,55 @@ def thresholded_output_transform(output):
...
+    if precision is not None and recall is not None:
+        p_cn = getattr(precision, "_class_names", None)
+        r_cn = getattr(recall, "_class_names", None)
+        if p_cn is not None and r_cn is not None and p_cn != r_cn:
+            raise ValueError("precision and recall class_names must match")
+
+    target_class_names = (
+        class_names or getattr(precision, "_class_names", None) or getattr(recall, "_class_names", None)
+    )
+
```
aaishwarymishra: we never checked if fbeta got class names and precision and recall also returns class names , does fbeta class names matches the ones returned by precision and recall?

ignite/metrics/fbeta.py
```diff
@@ -159,23 +161,55 @@ def thresholded_output_transform(output):
...
         precision = Precision(
             output_transform=(lambda x: x) if output_transform is None else output_transform,
             average=False,
             device=cast(str | torch.device, recall._device if recall else device),
+            class_names=target_class_names,
         )
     elif precision._average:
         raise ValueError("Input precision metric should have average=False")
+    elif target_class_names is not None:
+        precision._class_names = target_class_names
```
aaishwarymishra: I don't like this, we should be explicit, if precision and recall is given and fbeta uses class names the provided metrics should also have it , we should not silently modify the metric.

ignite/metrics/fbeta.py
aaishwarymishra, line 1: Also there is no early check for if average is True and also the class name has been given

ignite/metrics/fbeta.py
aaishwarymishra: I dont like this we should not silently modify metrics, user should provide metrics with class names if class names are set on fbeta

aniketpandey05:
Hi @aaishwarymishra, thanks for the review! I've updated the implementation:

ignite/metrics/fbeta.py
```diff
@@ -159,23 +161,60 @@ def thresholded_output_transform(output):
     if precision is None and recall is None and device is None:
         device = torch.device("cpu")
 
+    if class_names is not None:
+        if not isinstance(class_names, (list, tuple)) or not all(isinstance(n, str) for n in class_names):
+            raise ValueError("class_names must be a list of strings")
+
+    target_class_names = (
+        class_names or getattr(precision, "_class_names", None) or getattr(recall, "_class_names", None)
```
aaishwarymishra: I dont think we need these or conditions haere as we are matching the class names later.

ignite/metrics/fbeta.py
```diff
@@ -159,23 +161,60 @@ def thresholded_output_transform(output):
...
+    )
+
+    if target_class_names is not None and average is not False and average is not None:
+        raise ValueError(f"class_names is only applicable when average=False or average=None, got average={average!r}.")
```
aaishwarymishra: I think this should be first check not after class names assignment.

ignite/metrics/fbeta.py
```diff
@@ -159,23 +161,60 @@ def thresholded_output_transform(output):
...
+    if recall is not None:
+        if recall._average:
+            raise ValueError("Input recall metric should have average=False")
+        if class_names is not None and recall._class_names != class_names:
+            raise ValueError("recall metric class_names must match Fbeta class_names")
+
+    if precision is not None and recall is not None:
+        if precision._class_names != recall._class_names:
+            raise ValueError("precision and recall class_names must match")
+
```
aaishwarymishra: it can be simplified ig with
```python
active_metrics = [m for m in (precision, recall) if m is not None]

# 1. Check average setting
if any(m._average for m in active_metrics):
    raise ValueError("Input precision and recall metrics should have average=False")

# 2. Check class_names match Fbeta
if class_names is not None and any(m._class_names != class_names for m in active_metrics):
    raise ValueError("precision and recall metric class_names must match Fbeta class_names")
```

aaishwarymishra:
Also try to reslove previous comments it gets hard to read the code :)

ignite/metrics/fbeta.py
```diff
@@ -159,23 +161,61 @@ def thresholded_output_transform(output):
...
+    target_class_names = class_names
+    if target_class_names is None and precision is not None:
+        target_class_names = precision._class_names
+    if target_class_names is None and recall is not None:
+        target_class_names = recall._class_names
+
+    if target_class_names is not None and average is not False and average is not None:
+        raise ValueError(f"class_names is only applicable when average=False or average=None, got average={average!r}.")
+
```
aaishwarymishra: redundant check

aniketpandey05:
@aaishwarymishra i think the ci failure are unrelated to the changed files

aaishwarymishra:
@vfdev-5 looks good to me, you can check it too and merge :)

ignite/metrics/fbeta.py
vfdev-5: Is this resolved?

ignite/metrics/fbeta.py
```diff
@@ -42,6 +43,7 @@ def Fbeta(
         device: specifies which device updates are accumulated on. Setting the metric's
             device to be the same as your ``update`` arguments ensures the ``update`` method is non-blocking. By
             default, CPU.
+        class_names: list of class name strings used to label per-class output. Default: ``None``.
```
vfdev-5, line 46: Docstring should also contain .. version added tag for new args

ignite/metrics/recall.py
vfdev-5: Is this resolved?

vfdev-5 (commented):
I haven't yet checked the code itself but docstring needs some updates.

@aaishwarymishra resolve all open comments if they were addressed.

aaishwarymishra (approved):

aaishwarymishra (changes requested):
@aniketpandey05 can you update docstrings too :)

ignite/metrics/fbeta.py
```diff
@@ -42,6 +43,9 @@ def Fbeta(
...
+
+            .. versionadded:: 0.6.0
```
vfdev-5: This is a wrong tag, check Contributing guide and also the indent size.

ignite/metrics/fbeta.py
```diff
@@ -159,23 +163,58 @@ def thresholded_output_transform(output):
...
+        raise ValueError("precision and recall metric class_names must match Fbeta class_names")
+
+    if len(active_metrics) == 2 and active_metrics[0]._class_names != active_metrics[1]._class_names:
+        raise ValueError("precision and recall class_names must match")
+
+    target_class_names = class_names
+    if target_class_names is None and precision is not None:
+        target_class_names = precision._class_names
+    if target_class_names is None and recall is not None:
+        target_class_names = recall._class_names
```
vfdev-5:
```suggestion
    if class_names is None and precision is not None:
        class_names = precision._class_names
    if class_names is None and recall is not None:
        class_names = recall._class_names
```

ignite/metrics/fbeta.py
```diff
@@ -159,23 +163,58 @@ def thresholded_output_transform(output):
...
+
     if precision is None:
         precision = Precision(
             output_transform=(lambda x: x) if output_transform is None else output_transform,
             average=False,
             device=cast(str | torch.device, recall._device if recall else device),
+            class_names=target_class_names,
```
vfdev-5:
```suggestion
            class_names=class_names,
```

ignite/metrics/fbeta.py
```diff
@@ -159,23 +163,58 @@ def thresholded_output_transform(output):
...
         )
-    elif precision._average:
-        raise ValueError("Input precision metric should have average=False")
 
     if recall is None:
         recall = Recall(
             output_transform=(lambda x: x) if output_transform is None else output_transform,
             average=False,
             device=cast(str | torch.device, precision._device if precision else device),
+            class_names=target_class_names,
```
vfdev-5:
```suggestion
            class_names=class_names,
```

ignite/metrics/fbeta.py
```diff
@@ -159,23 +163,58 @@ def thresholded_output_transform(output):
...
+            raise ValueError("class_names must be a list of strings")
+        if average is not False and average is not None:
+            raise ValueError(
+                f"class_names is only applicable when average=False or average=None, got average={average!r}."
+            )
+
+    active_metrics = [m for m in (precision, recall) if m is not None]
+
+    if any(m._average for m in active_metrics):
+        raise ValueError("Input precision and recall metrics should have average=False")
```
vfdev-5: This new error message is not precise enough. We should explicitly tell which metric exactly should have average false and not give a generic message

ignite/metrics/precision.py
```diff
@@ -125,7 +135,7 @@ def reset(self) -> None:
         super().reset()
 
     @sync_all_reduce("_numerator", "_denominator")
-    def compute(self) -> torch.Tensor | float:
+    def compute(self) -> torch.Tensor | float | dict:
```
vfdev-5: Can you make dict type hint more detailed: `dict[str, <whatever type is here>]`

ignite/metrics/precision.py
```diff
@@ -157,6 +167,10 @@ def compute(self) -> torch.Tensor | float:
         elif self._average == "macro":
             return cast(torch.Tensor, fraction).mean().item()
         else:
+            if self._class_names is not None:
+                if isinstance(fraction, torch.Tensor) and fraction.ndim == 0:
+                    return {self._class_names[0]: fraction.item()}
```
vfdev-5: Not sure about this special case, what it is and why it should be handled differently from others

ignite/metrics/precision.py
```diff
@@ -246,6 +260,12 @@ class Precision(_BasePrecisionRecall):
         skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
             true for multi-output model, for example, if ``y_pred`` contains multi-output as ``(y_pred_a, y_pred_b)``
             Alternatively, ``output_transform`` can be used to handle this.
+        class_names: list of class name strings used to label per-class output when ``average=False``
+            or ``average=None``. If provided, ``compute()`` returns a ``dict``  mapping each class
+            name to its metric value instead of a tensor. Must match the number of classes inferred
+            from the data. Default: ``None``.
+
+            .. versionadded:: 0.6.0
```
vfdev-5: this is a wrong tag, it should not be added here.

ignite/metrics/recall.py
```diff
@@ -97,6 +97,12 @@ class Recall(_BasePrecisionRecall):
         skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
             true for multi-output model, for example, if ``y_pred`` contains multi-output as ``(y_pred_a, y_pred_b)``
             Alternatively, ``output_transform`` can be used to handle this.
+        class_names: list of class name strings used to label per-class output when ``average=False``
+            or ``average=None``. If provided, ``compute()`` returns a ``dict`` mapping each class
+            name to its metric value instead of a tensor. Must match the number of classes inferred
+            from the data. Default: ``None``.
+
+            .. versionadded:: 0.6.0
```
vfdev-5:
```suggestion
```

tests/ignite/metrics/test_fbeta.py
```diff
@@ -15,11 +15,11 @@ def test_wrong_inputs():
     with pytest.raises(ValueError, match=r"Beta should be a positive integer"):
         Fbeta(0.0)
 
-    with pytest.raises(ValueError, match=r"Input precision metric should have average=False"):
+    with pytest.raises(ValueError, match=r"Input precision and recall metrics should have average=False"):
```
vfdev-5: Keep this test as it was and update the code

tests/ignite/metrics/test_fbeta.py
```diff
@@ -15,11 +15,11 @@ def test_wrong_inputs():
...
         p = Precision(average="micro")
         Fbeta(1.0, precision=p)
 
-    with pytest.raises(ValueError, match=r"Input recall metric should have average=False"):
+    with pytest.raises(ValueError, match=r"Input precision and recall metrics should have average=False"):
```
vfdev-5: Same here

tests/ignite/metrics/test_fbeta.py
```diff
@@ -233,3 +233,58 @@ def test_multinode_distrib_gloo_cpu_or_gpu(distributed_context_multi_node_gloo):
...
+    r = Recall(average=False, class_names=["a", "b"])
+    with pytest.raises(ValueError, match="precision and recall class_names must match"):
+        Fbeta(beta=1.0, average=False, precision=p, recall=r)
+
+    # Input precision metric with average != False
+    p_avg = Precision(average="macro")
+    with pytest.raises(ValueError, match="Input precision and recall metrics should have average=False"):
+        Fbeta(beta=1.0, average=False, precision=p_avg)
+
+    # Correct computation passing class_names directly to Fbeta
```
vfdev-5: I prefer that these test cases were added to existing tests as new parametrizations.

ignite/metrics/fbeta.py
aniketpandey05: updated , removed the inline version tag

ignite/metrics/fbeta.py
aniketpandey05: changed to more precise per metric check 

"Input precision metric should have average=False"
"Input recall metric should have average=False"

ignite/metrics/precision.py
aniketpandey05: updated the annotation to  ```torch.Tensor | float | dict[str,float]```

ignite/metrics/precision.py
aniketpandey05: I  thought this  was needed because in binary classification with average=False, fraction is a 0-D scalar tensor, and a 0 -D tensor would raise a typerror , i  have refactored it now

ignite/metrics/precision.py
aniketpandey05: Removed the inline tag

tests/ignite/metrics/test_fbeta.py
aniketpandey05: Reverted back to the original specific error match

ignite/metrics/fbeta.py
```diff
@@ -159,23 +164,61 @@ def thresholded_output_transform(output):
     if precision is None and recall is None and device is None:
         device = torch.device("cpu")
 
+    if class_names is not None:
+        if not isinstance(class_names, (list, tuple)) or not all(isinstance(n, str) for n in class_names):
+            raise ValueError("class_names must be a list of strings")
+        if average is not False and average is not None:
```
vfdev-5:
```suggestion
        if average not in (False, None):
```

vfdev-5 (approved):
Thanks @aniketpandey05 , looks good to me!

steaphenai (approved):

ignite/metrics/fbeta.py
aniketpandey05: applied the same change in precision.py too

vfdev-5:
@aniketpandey05 can you please send a separate PR with a fix for the pyrefly: either put ignore the error or if it is possible to fix it would be fine as well. Thanks!

vfdev-5:
Made this PR to fix pyrefly: https://github.com/pytorch/ignite/pull/3835

PR #3733, opened 2026-04-15
FSDP2 Support

intelava:
## Add FSDP2 support                                                                                                                                                     
                                                                                                                                                                           
  Added FSDP2 support to ignite using the composable `fully_shard` API.                                                                                                           
                                                                                                                                                                           
  ### What's new  
                                                                                                                                                                           
  auto_model now has a use_fsdp parameter. When True in a distributed context it                                                                                           
  applies fully_shard to the model instead of DDP. You can pass extra kwargs like
  mp_policy and offload_policy directly. Combining use_fsdp=True with sync_bn=True                                                                                         
  raises a ValueError right away since they don't work together.
                                                                                                                                                                           
  The checkpoint handler now recognizes FSDP2 models and uses get_model_state_dict
  and set_model_state_dict with full_state_dict=True so parameter gathering across                                                                                         
  ranks is handled automatically.                                                                                                                      
                  
  ### Tests                                                                                                                                                                
                  
  Added test_fsdp_smoke.py with 23 single process tests covering the flag                                                                                                  
  behavior, checkpoint imports and edge cases. test_fsdp_distributed.py is a
  standalone multi process script that spawns 2 workers with gloo and tests                                                                                                
  wrapping, forward pass and checkpoint save/load end to end.
                                                                                                                                                                           
  ---                                                                                                     
If you have any further questions or suggestions, I would be happy to help!

### DIFF
--- a/ignite/distributed/auto.py
+++ b/ignite/distributed/auto.py
@@ -19,6 +19,7 @@ from ignite.utils import setup_logger
 
 try:
     from torch.distributed._composable.fsdp import fully_shard
+    from torch.distributed.device_mesh import init_device_mesh
 
     HAVE_FSDP2 = True
 except ImportError:
@@ -247,6 +248,8 @@ def auto_model(model: nn.Module, sync_bn: bool = False, use_fsdp: bool = False, 
                     raise RuntimeError(
                         "fully_shard (FSDP2) is not available. Please upgrade to PyTorch >= 2.0."
                     )
+                if "mesh" not in kwargs:
+                    kwargs["mesh"] = init_device_mesh(device.type, (idist.get_world_size(),))
                 logger.info("Apply torch FSDP2 (fully_shard) on model")
                 model = fully_shard(model, **kwargs)
             else:
--- a/ignite/distributed/auto.py
+++ b/ignite/distributed/auto.py
@@ -141,7 +141,7 @@ def auto_dataloader(dataset: Dataset, **kwargs: Any) -> DataLoader | _MpDeviceLo
     return dataloader
 
 
-def auto_model(model: nn.Module, sync_bn: bool = False, **kwargs: Any) -> nn.Module:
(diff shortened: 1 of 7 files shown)


ignite/distributed/auto.py
```diff
@@ -204,23 +243,33 @@ def auto_model(model: nn.Module, sync_bn: bool = False, **kwargs: Any) -> nn.Mod
...
-                lrank = idist.get_local_rank()
-                logger.info(f"Apply torch DistributedDataParallel on model, device id: {lrank}")
-                kwargs["device_ids"] = [
-                    lrank,
-                ]
+            if use_fsdp:
+                if not HAVE_FSDP2:
+                    raise RuntimeError(
+                        "fully_shard (FSDP2) is not available. Please upgrade to PyTorch >= 2.0."
+                    )
```
vfdev-5, line 256: Minimal version of pytorch when fsdp2 appeared is 2.6

ignite/handlers/checkpoint.py
```diff
@@ -403,6 +413,7 @@ def _get_filename_pattern(self, global_step: int | None) -> str:
                 with_score=self.score_function is not None,
                 with_score_name=self.score_name is not None,
                 with_global_step=global_step is not None,
+                with_ext=bool(self.ext),
```
vfdev-5: Let's rename this param as `as_folder`

vfdev-5 (commented):
Thanks a lot for the PR @intelava !
I left few initial comments, and I'll review it in details a bit later

tests/ignite/test_fsdp_distributed.py
```diff
@@ -0,0 +1,188 @@
...
+from __future__ import annotations
+
+import os
+import sys
+
+import torch
+import torch.distributed as dist
+import torch.nn as nn
+
+sys.path.insert(0, "/pfss/mlde/workspaces/mlde_wsp_MazaheriA/tk27ryru/ignite")
```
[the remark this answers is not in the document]
vfdev-5: @intelava let's remove all local paths like this one

PR #3740, opened 2026-04-15
fix: use float64 accumulators in PearsonCorrelation to prevent catastrophic cancellation

### LINKED ISSUE #3662: [Bug] Numerical instability in `PearsonCorrelation` due to naive variance formula
The current implementation of `PearsonCorrelation` in `ignite.metrics.regression` uses the **naive "sum of squares" algorithm** to compute variance and covariance:

\[
Var(X) = E[X^2] - (E[X])^2
\]

Although mathematically valid, this approach is **numerically unstable** when the input values have **large magnitudes relative to their variance**.

This leads to **catastrophic cancellation**, where two very large numbers (`E[X^2]` and `(E[X])^2`) are subtracted, causing loss of precision in **float32**. As a result, the metric can produce **incorrect results**, such as returning **0.0** when the true correlation is **≈ 0.707**.

---

## Steps to Reproduce

```python
import torch
from ignite.metrics.regression import PearsonCorrelation
import numpy as np

# Magnitude that triggers precision loss in float32
offset = 1e8 

y_true = torch.tensor([1.0, 2.0, 3.0, 4.0, 5.0]) + offset
y_pred = torch.tensor([1.1, 2.1, 3.1, 4.1, 5.1]) + offset

metric = PearsonCorrelation()
metric.update((y_pred, y_true))
ignite_res = metric.compute()

# Ground truth using PyTorch's stable corrcoef
combined = torch.stack([y_pred, y_true])
torch_res = torch.corrcoef(combined)[0, 1].item()

print(f"Offset used: {offset}")
print(f"Ignite Result: {ignite_res}")
print(f"Torch Ground Truth: {torch_res}")
```
---
## actual output

```python
Offset used: 100000000.0
Ignite Result: 0.0
Torch Ground Truth: 0.7071067690849304
```
---
## Expected Behavior

The **Pearson correlation metric** should be **invariant to constant offsets** in the input data.

Therefore, the result should be consistent with **numerically stable implementations** such as:

- **`torch.corrcoef`**
- **`scipy.stats.pearsonr`**

###  Expected Result

```text
≈ 0.7071067
```
---
##  Proposed Fix

To improve **numerical stability**, the implementation can be updated as follows:

### Use a Numerically Stable Online Algorithm
Implement **Welford’s Online Algorithm** (or a similar **one-pass algorithm**) to compute:

- **Mean**
- **Variance**
- **Covariance**

incrementally.

---

### Use Higher Precision Accumulators
Maintain internal accumulators using **`torch.float64`** to prevent **precision loss during batch updates**.

---

### Ensure Compatibility with Ignite Metrics
The updated implementation should still support **batch-wise streaming updates** used by **Ignite metrics**.

---

### Benefit
These changes will prevent **catastrophic cancellation** and ensure **correct results** even when the data has:

- **Large offsets**
- **High magnitude values**
---
Hi @vfdev-5, I would like to work on a PR to resolve this issue if this looks good to you!

tejas-ae:
Fixes #3662.

The naive E[X²] - (E[X])² formula for variance is mathematically correct but numerically unstable. When values have large magnitude relative to their variance — for example an offset of 1e8 with small inter-sample differences — both E[X²] and (E[X])² are around 1e16 while their difference (the actual variance) is O(1). In float32, the unit in the last place at that scale is roughly 10^9, so the variance is completely lost and the metric returns 0.0.

The fix is to accumulate in float64 on devices that support it. MPS does not support float64, so it falls back to float32 and retains the previous behaviour. The public return type (Python float) is unchanged.

```python
# before (float32 accumulators)
offset = 1e8
y_pred = torch.tensor([1.0, 2.0, 3.0, 4.0, 5.0], dtype=torch.float64) + offset
y      = torch.tensor([2.0, 4.0, 6.0, 8.0, 10.0], dtype=torch.float64) + offset
m = PearsonCorrelation()
m.update((y_pred, y))
m.compute()  # 0.0  (wrong; expected 1.0)

# after (float64 accumulators)
m.compute()  # 1.0  ✓
```

A new test `test_numerical_stability_large_offset` covers this case. All existing non-distributed tests pass.

you know the codebase far better than I do — happy to adjust if a different approach (e.g., a shared Welford-based utility across PearsonCorrelation, R2Score, and FID as discussed in the issue) is preferred.

### DIFF
--- a/ignite/metrics/regression/pearson_correlation.py
+++ b/ignite/metrics/regression/pearson_correlation.py
@@ -78,15 +78,20 @@ class PearsonCorrelation(_BaseRegression):
 
     @reinit__is_reduced
     def reset(self) -> None:
-        self._sum_of_y_preds = torch.tensor(0.0, device=self._device)
-        self._sum_of_ys = torch.tensor(0.0, device=self._device)
-        self._sum_of_y_pred_squares = torch.tensor(0.0, device=self._device)
-        self._sum_of_y_squares = torch.tensor(0.0, device=self._device)
-        self._sum_of_products = torch.tensor(0.0, device=self._device)
+        # Use float64 accumulators to avoid catastrophic cancellation in
+        # E[X^2] - (E[X])^2 when values have large magnitude.  MPS does not
+        # support float64, so fall back to float32 there.
+        acc_dtype = torch.float64 if self._device.type != "mps" else torch.float32
+        self._sum_of_y_preds = torch.tensor(0.0, dtype=acc_dtype, device=self._device)
+        self._sum_of_ys = torch.tensor(0.0, dtype=acc_dtype, device=self._device)
+        self._sum_of_y_pred_squares = torch.tensor(0.0, dtype=acc_dtype, device=self._device)
+        self._sum_of_y_squares = torch.tensor(0.0, dtype=acc_dtype, device=self._device)
+        self._sum_of_products = torch.tensor(0.0, dtype=acc_dtype, device=self._device)
         self._num_examples = 0
 
     def _update(self, output: tuple[torch.Tensor, torch.Tensor]) -> None:
-        y_pred, y = output[0].detach(), output[1].detach()
+        y_pred = output[0].detach().to(dtype=self._sum_of_y_preds.dtype)
+        y = output[1].detach().to(dtype=self._sum_of_y_preds.dtype)
         self._sum_of_y_preds += y_pred.sum().to(self._device)
         self._sum_of_ys += y.sum().to(self._device)
         self._sum_of_y_pred_squares += y_pred.square().sum().to(self._device)
--- a/tests/ignite/metrics/regression/test_pearson_correlation.py
+++ b/tests/ignite/metrics/regression/test_pearson_correlation.py
@@ -137,6 +137,24 @@ def test_integration_pearson_correlation(n_times, test_case: tuple[Tensor, Tenso
     assert pytest.approx(np_ans, rel=2e-4) == corr
 
 
+def test_numerical_stability_large_offset():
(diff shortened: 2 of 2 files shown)


siddhant-shahhh (commented):
The bug is real and this fix works. One thing worth noting: the part that fixes it is the input cast, not the float64 accumulators.

I reproduced the accumulator math with float32 inputs at offset 1e4:

```
master (float32 accumulators)            r = 800000000.0
float64 accumulators, inputs NOT cast    r = 440000000.59604645
float64 accumulators + input cast (PR)   r = 1.0
```

Changing only the accumulator dtype doesn't help, since `y_pred.square().sum()` is still computed in float32 before being widened. The `.to(dtype=...)` in `_update` is what fixes it. A short comment there would stop someone from removing it later as redundant.

On master this gives `800000000.0` for five points at offset 1e4, with no NaN. That uses float32 inputs, which is what the docstring says to pass, so it might be worth a test too. The current test uses float64 inputs at offset 1e8.

The test does catch the bug. On this branch the file is 32 passed, 34 skipped. With `pearson_correlation.py` reverted to master and the test kept:

```
assert 1.0 ± 1.0e-06 == 0.0
  Obtained: 0.0
  Expected: 1.0 ± 1.0e-06
```

One question. The five accumulators are in `_state_dict_all_req_keys`, and `_CollectionItem.load_value` assigns the loaded tensor directly. So loading a checkpoint saved before this change brings back float32 tensors, and `_update` then casts the inputs to float32 through `self._sum_of_y_preds.dtype`. The fix would be undone without any warning. Casting in `reset` or on load would handle that. What do you think?

Smaller notes:

- The MPS fallback keeps float32 accumulators, so the bug remains there. That's fine as a limitation, but the new test is CPU-only while the rest of the file uses `available_device`.

`ruff check` and `ruff format` pass on both files.

LGTM.

PR #3745, opened 2026-04-25
feat(early_stopping): add get_default_score_fn and get_default_event_filter helpers (#3411)

MukundaKatta:
## Summary
Closes part of #3411. Two enhancements explicitly suggested by maintainer @vfdev-5 in the issue thread:
1. `EarlyStopping.get_default_score_fn` — analogous to `Checkpoint.get_default_score_fn`, supports `score_sign=-1.0` for error-like metrics.
2. `EarlyStopping.get_default_event_filter` — events filtering helper for warmup that consults `self.trainer.state.epoch` (fixes the gotcha that evaluators reset their epoch counter on each `run()`).

The maintainer pushed back on PR #3661's `min_evals` parameter and explicitly preferred event filtering, which is why this PR ships the filter helper instead.

## Files
- `ignite/handlers/early_stopping.py` (+88) — both helpers, additive, BC-safe.
- `tests/ignite/handlers/test_early_stopping.py` (+136) — 9 new tests covering positive/negative sign, invalid sign, integration with `mode="min"`, invalid `after`, zero/non-zero warmup, full engine integration, and attach signature compatibility.

## Test plan
- 32 pass, 7 distributed deselected (env-only). Existing 23 tests unaffected.
- `ruff check` + `ruff format --check` clean.
- Backwards compatible — no existing behavior changed.

### DIFF
--- a/ignite/handlers/early_stopping.py
+++ b/ignite/handlers/early_stopping.py
@@ -67,6 +67,7 @@ class EarlyStopping(Serializable, ResettableHandler):
         Renamed ``min_delta_mode``  to ``threshold_mode``.
         Renamed ``min_delta`` to ``threshold``.
         Renamed ``cumulative_delta`` to ``cumulative``.
+        Added :meth:`get_default_score_fn` and :meth:`get_default_event_filter` static helpers.
     .. versionchanged:: 0.5.4
         Added `mode` parameter to support minimization in addition to maximization.
         Added `min_delta_mode` parameter to support both absolute and relative improvements.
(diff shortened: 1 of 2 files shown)


ignite/handlers/early_stopping.py
```diff
@@ -291,3 +292,90 @@ def load_state_dict(self, state_dict: Mapping) -> None:
...
+        """
+        if not isinstance(after, int) or after < 0:
+            raise ValueError("Argument after should be a non-negative integer.")
+
+        trainer = self.trainer
+
+        def event_filter(engine: Engine, event: int) -> bool:
+            # The host engine's `event` counter resets per `run()` (e.g. evaluators
+            # called repeatedly), so look at the trainer's epoch counter instead.
+            return trainer.state.epoch > after
```
aaishwarymishra, line 379: I think we could add support for iteration too, so user can specify minimum iterations allowed

ignite/handlers/early_stopping.py
```diff
@@ -291,3 +292,90 @@ def load_state_dict(self, state_dict: Mapping) -> None:
         self.counter = state_dict["counter"]
         self.best_score = state_dict["best_score"]
         self.threshold_mode = state_dict.get("threshold_mode", self.threshold_mode)
+
+    @staticmethod
+    def get_default_score_fn(metric_name: str, score_sign: float = 1.0) -> Callable:
```
aaishwarymishra, line 297: whats the use of this in early stopping?

TahaZahid05:
@MukundaKatta Hi! Are you still working on this? Let us know if you are having any issues.

Issue #3748, opened 2026-05-11
Numerical stability across running-moment metrics — unified plan (tracking #3662 / #3672 / #3741)


joemunene-by:
Following @vfdev-5's suggestion in #3662, opening this as the umbrella issue so we can plan the numerical-stability work across all metrics that compute variance / covariance from running sums.

### Problem

Several Ignite metrics maintain separate running sums (`Σx`) and running sums-of-squares (`Σx²`) and combine them at `compute()` time as `Σx² − (Σx)²/n`. This is the textbook naive variance formula and suffers from catastrophic cancellation in `float32` whenever `|E[X]|` is large relative to `Var[X]`. Concrete failure mode in #3662: `PearsonCorrelation` returned `0.89` instead of `0.99` on data with `mean=1e6, std=1`.

### Affected metrics

| Metric | File | State |
|---|---|---|
| `PearsonCorrelation` | `ignite/metrics/regression/pearson_correlation.py` | Fixed in #3741 via Welford's online algorithm + parallel merge |
| `R2Score` | `ignite/metrics/regression/r2_score.py:80` | Same naive formula in the denominator (`_y_sq_sum - _y_sum**2 / _num_examples`). Reproduces the bug for any target distribution with a large mean. |

Audited and confirmed **not** affected: `MeanSquaredError`, `PSNR` (squared residuals, no `Σ(x²) − (Σx)²/n` step), `HSIC` (per-batch accumulator), `SSIM` (uses constants), `SpearmanRankCorrelation` / `KendallRankCorrelation` (delegate to SciPy on buffered data), `MaximumMeanDiscrepancy` (kernel sums).

If anyone spots another metric I missed, please add to the table above.

### Proposed unified fix

Extract the Welford state + parallel-merge logic that #3741 introduces into a small shared helper, then have each affected metric consume it:

```
ignite/metrics/_running_stats.py    # new
    class WelfordState:
        mean, m2, n           # univariate
        update(batch)
        merge(other)
    class WelfordCovState(WelfordState):
        mean_x, mean_y, m2_x, m2_y, c_xy, n
        update(batch_x, batch_y)
        merge(other)
```

Each metric keeps its own `WelfordState` (or `WelfordCovState` for covariance / correlation), updates it in `update()`, and reads `variance` / `covariance` properties in `compute()`. `sync_all_reduce` extends to merge state across ranks via `merge()` rather than summing individual buffers.

### Migration order

[...]

PR #3750, opened 2026-05-11
feat(metrics): add WelfordVariance and WelfordCovariance helpers (PR 1 of #3748)

joemunene-by:
## Summary

PR 1 of the plan in #3748: introduces `ignite/metrics/_running_stats.py` with two numerically stable running-statistics primitives that variance- and covariance-bearing metrics can share, instead of each one rolling its own naive `Σx² − (Σx)²/n` implementation.

- `WelfordVariance`: running `mean`, `variance`, `std` for a single variable.
- `WelfordCovariance`: running `variance_x`, `variance_y`, `covariance`, and Pearson `correlation()` for a paired `(x, y)` stream.

Both classes:

- keep internal state in `float64` regardless of input dtype, so the classic `E[X²] − E[X]²` cancellation does not bite at large means (the failure mode from #3662);
- update incrementally via Welford's online algorithm;
- expose `merge()` implementing the Chan / Welford parallel formula, suitable for cross-rank distributed reduction or any other case where two accumulators need to be combined without re-iterating the raw data.

No existing metric is touched in this PR. The helper lands first so the two consumer PRs can each diff against a stable shared API.

## Follow-ups (already scoped in #3748)

- **PR 2**: port `R2Score` to `WelfordVariance`, with a regression test mirroring #3662's `mean=1e6, std=1` failure case.
- **PR 3**: update #3741 to drop its inline Welford state and consume `WelfordCovariance` instead. Pure refactor on top of the existing review.

## API

```python
class WelfordVariance:
    n_samples: int
    mean: torch.Tensor                       # running mean
    sum_sq_dev_from_mean: torch.Tensor       # M2 = Σ (x_i − mean)^2

    def update(self, batch: torch.Tensor) -> None: ...
    def merge(self, other: WelfordVariance) -> None: ...

    @property
    def variance(self) -> torch.Tensor: ...  # M2 / n
    @property
    def std(self) -> torch.Tensor: ...       # sqrt(variance)


class WelfordCovariance:
    n_samples: int
    mean_x: torch.Tensor
    mean_y: torch.Tensor
    sum_sq_dev_x: torch.Tensor
    sum_sq_dev_y: torch.Tensor
    sum_product_of_devs: torch.Tensor        # Σ (x_i − mean_x)(y_i − mean_y)

    def update(self, batch_x, batch_y) -> None: ...
    def merge(self, other: WelfordCovariance) -> None: ...

    @property
    def variance_x(self) -> torch.Tensor: ...
    @property
    def variance_y(self) -> torch.Tensor: ...
    @property
    def covariance(self) -> torch.Tensor: ...
    def correlation(self, eps: float = 1e-8) -> torch.Tensor: ...  # Pearson r
```

## Tests

20 unit tests in `tests/ignite/metrics/test_running_stats.py`, all passing locally with `pytest tests/ignite/metrics/test_running_stats.py`:

**WelfordVariance**
- empty accumulator returns zero variance / std without crashing
- single-batch update matches `numpy.mean` and `numpy.var` to 1e-12
- multi-batch update matches single-batch on the same data
- `merge` of two accumulators matches `update` on the concatenated data
- `merge` with an empty accumulator on either side is a no-op or absorbs the other side
- numerical-stability regression: 10,000 samples at `mean=1e6, std=1` in float32; Welford in float64 recovers the true variance, and the same data through the naive `Σx² − (Σx)²/n` formula in float32 is asserted to fail by ≥ 0.1, so the test documents the failure mode it is protecting against
- single-sample case (variance is 0, not undefined)
- empty-batch `update` is a no-op
- `reset` clears state
- integer input dtypes upcast to float64

**WelfordCovariance**
- empty accumulator returns zero everywhere
- single-batch update matches `numpy.cov(..., bias=True)` and `numpy.corrcoef` to 1e-10 to 1e-12
- multi-batch update matches single-batch
- `merge` matches concatenated `update`
- numerical-stability regression: 10,000 paired samples at `mean=1e6`, true correlation `> 0.99`; Welford recovers `r` within 1e-4 of the float64 ground truth
- shape mismatch raises `ValueError`
- empty-batch update is a no-op
- constant-y series returns `r = 0.0` (not NaN) via the `eps` clamp
- `reset` clears state

**Cross-class sanity**
- `WelfordCovariance.variance_x` matches `WelfordVariance.variance` fed the same `x`. Catches future drift between the two implementations.

## Conventions

- `ruff format` clean, `ruff check` clean.
- Internal state, return values, and properties are all `torch.Tensor` on the user-supplied device, mirroring the dtype / device convention used elsewhere in `ignite.metrics`.
- File is named `_running_stats.py` (leading underscore) and not re-exported from `ignite/metrics/__init__.py`, so it is internal to the metrics module. Public access stays through individual metric classes; the helper can graduate to a public API later if there is demand.

## Test plan

- [x] `pytest tests/ignite/metrics/test_running_stats.py`: 20 / 20 passing locally
- [x] `ruff format --check`, `ruff check`
- [ ] CI on this PR

cc @vfdev-5. Opening PR 2 (R2Score port) and PR 3 (PearsonCorrelation refactor on top of #3741) once this lands.

### DIFF
--- a/ignite/metrics/_running_stats.py
+++ b/ignite/metrics/_running_stats.py
@@ -7,11 +7,16 @@ trap of the naive ``E[X^2] - E[X]^2`` formula. Used by
 :class:`~ignite.metrics.regression.R2Score`; new metrics with the same
 need should consume these helpers rather than rolling their own.
 
-Both classes keep internal state in ``float64`` regardless of the input
-dtype, follow Welford's online algorithm for incremental updates, and
-fold accumulators together with the Chan / Welford parallel formula
-(used both for batch-wise updates and for cross-rank merges in
-distributed settings).
+Both classes are tensor-type-agnostic dataclasses: callers supply
+tensors in whatever dtype and device they want, and the helper
+preserves both. For numerical stability under large means, callers
+should pre-cast inputs to ``float64`` (the consumer metric classes
+already do this in their own ``update`` methods).
+
+Updates follow Welford's online algorithm. Two accumulators can be
+combined into one via :meth:`merge`, which implements the Chan /
+Welford parallel formula and is the basis for cross-rank distributed
+reductions.
 
 References:
     Welford, B. P. (1962). Note on a method for calculating corrected
@@ -21,11 +26,16 @@ References:
     https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm
 """
 
-from typing import Union
+from dataclasses import dataclass, field
 
 import torch
 
 
+def _zero() -> torch.Tensor:
+    return torch.tensor(0.0)
+
+
+@dataclass
 class WelfordVariance:
     """Numerically stable running mean and variance via Welford's algorithm.
 
@@ -35,52 +45,38 @@ class WelfordVariance:
     which uses the Chan / Welford parallel formula and is the basis for
     distributed reductions.
 
-    State is kept in ``float64`` regardless of input dtype so that the
-    classic ``E[X^2] - E[X]^2`` cancellation does not bite for inputs
-    with large means.
+    No dtype or device handling is performed inside the class: the
+    state takes the dtype and device of the first batch passed to
+    :meth:`update`. For numerical stability under large means, callers
+    should hand in ``float64`` tensors.
 
-    Args:
-        device: device on which to keep the running-state tensors.
-            Default: ``"cpu"``.
+    Example::
 
-    Example:
-
-        .. code-block:: python
-
-            ws = WelfordVariance()
-            for batch in stream:
-                ws.update(batch)
-            print(ws.mean.item(), ws.variance.item(), ws.std.item())
+        ws = WelfordVariance()
+        for batch in stream:
+            ws.update(batch.to(torch.float64))
+        print(ws.mean.item(), ws.variance.item(), ws.std.item())
     """
 
-    n_samples: int
-    mean: torch.Tensor
-    sum_sq_dev_from_mean: torch.Tensor
-
-    def __init__(self, device: Union[str, torch.device] = "cpu") -> None:
-        self._device = torch.device(device)
-        self.reset()
-
-    def reset(self) -> None:
-        """Drop all accumulated state."""
-        self.n_samples = 0
-        self.mean = torch.tensor(0.0, dtype=torch.float64, device=self._device)
-        self.sum_sq_dev_from_mean = torch.tensor(0.0, dtype=torch.float64, device=self._device)
+    n_samples: int = 0
+    mean: torch.Tensor = field(default_factory=_zero)
+    sum_sq_dev_from_mean: torch.Tensor = field(default_factory=_zero)
 
     @torch.no_grad()
     def update(self, batch: torch.Tensor) -> None:
         """Fold a batch of samples into the running state.
 
-        Empty batches are silently ignored. Inputs of any dtype are
-        upcast to ``float64`` for the internal computation.
+        Empty batches are silently ignored. ``batch.mean()`` and
+        ``batch.numel()`` perform a full reduction over the input, so
+        any shape is accepted and treated as ``numel`` scalar samples.
         """
         if batch.numel() == 0:
             return
-        batch64 = batch.detach().to(dtype=torch.float64).flatten()
-        n_b = batch64.shape[0]
+        batch = batch.detach()
+        n_b = batch.numel()
 
-        mean_b = batch64.mean().to(self._device)
-        m2_b = (batch64 - batch64.mean()).square().sum().to(self._device)
+        mean_b = batch.mean()
+        m2_b = (batch - mean_b).square().sum()
 
         if self.n_samples == 0:
             self.mean = mean_b
@@ -96,26 +92,21 @@ class WelfordVariance:
         self.n_samples = n_ab
 
     def merge(self, other: "WelfordVariance") -> None:
-        """Combine ``other`` into ``self`` using the Chan / Welford parallel formula.
-
-        Used both for cross-rank reduction in distributed settings and
-        for any other case where two accumulators need to be combined
-        into one without re-iterating the raw data.
-        """
+        """Combine ``other`` into ``self`` using the Chan / Welford parallel formula."""
         if other.n_samples == 0:
             return
         if self.n_samples == 0:
             self.n_samples = other.n_samples
-            self.mean = other.mean.detach().clone().to(self._device)
-            self.sum_sq_dev_from_mean = other.sum_sq_dev_from_mean.detach().clone().to(self._device)
+            self.mean = other.mean.detach().clone()
+            self.sum_sq_dev_from_mean = other.sum_sq_dev_from_mean.detach().clone()
             return
         n_a = self.n_samples
         n_b = other.n_samples
         n_ab = n_a + n_b
-        delta = other.mean.to(self._device) - self.mean
+        delta = other.mean - self.mean
         self.mean = self.mean + delta * n_b / n_ab
         self.sum_sq_dev_from_mean = (
-            self.sum_sq_dev_from_mean + other.sum_sq_dev_from_mean.to(self._device) + delta * delta * n_a * n_b / n_ab
+            self.sum_sq_dev_from_mean + other.sum_sq_dev_from_mean + delta * delta * n_a * n_b / n_ab
         )
         self.n_samples = n_ab
 
@@ -123,7 +114,7 @@ class WelfordVariance:
     def variance(self) -> torch.Tensor:
         """Population variance (divisor ``n``). Returns ``0.0`` when empty."""
         if self.n_samples == 0:
-            return torch.tensor(0.0, dtype=torch.float64, device=self._device)
+            return torch.tensor(0.0)
         return torch.clamp(self.sum_sq_dev_from_mean / self.n_samples, min=0.0)
 
     @property
@@ -132,6 +123,7 @@ class WelfordVariance:
         return self.variance.sqrt()
 
 
+@dataclass
 class WelfordCovariance:
     """Numerically stable running covariance for a pair of variables (x, y).
 
@@ -140,37 +132,24 @@ class WelfordCovariance:
     online update + Chan / Welford parallel merge as
     :class:`WelfordVariance`.
 
-    Args:
-        device: device on which to keep the running-state tensors.
-            Default: ``"cpu"``.
+    Like :class:`WelfordVariance`, the class is dtype and device
+    agnostic: state takes the dtype and device of the first batch
+    passed to :meth:`update`.
     """
 
-    n_samples: int
-    mean_x: torch.Tensor
-    mean_y: torch.Tensor
-    sum_sq_dev_x: torch.Tensor
-    sum_sq_dev_y: torch.Tensor
-    sum_product_of_devs: torch.Tensor
-
-    def __init__(self, device: Union[str, torch.device] = "cpu") -> None:
-        self._device = torch.device(device)
-        self.reset()
-
-    def reset(self) -> None:
-        """Drop all accumulated state."""
-        self.n_samples = 0
-        self.mean_x = torch.tensor(0.0, dtype=torch.float64, device=self._device)
-        self.mean_y = torch.tensor(0.0, dtype=torch.float64, device=self._device)
-        self.sum_sq_dev_x = torch.tensor(0.0, dtype=torch.float64, device=self._device)
-        self.sum_sq_dev_y = torch.tensor(0.0, dtype=torch.float64, device=self._device)
-        self.sum_product_of_devs = torch.tensor(0.0, dtype=torch.float64, device=self._device)
+    n_samples: int = 0
+    mean_x: torch.Tensor = field(default_factory=_zero)
+    mean_y: torch.Tensor = field(default_factory=_zero)
+    sum_sq_dev_x: torch.Tensor = field(default_factory=_zero)
+    sum_sq_dev_y: torch.Tensor = field(default_factory=_zero)
+    sum_product_of_devs: torch.Tensor = field(default_factory=_zero)
 
     @torch.no_grad()
     def update(self, batch_x: torch.Tensor, batch_y: torch.Tensor) -> None:
         """Fold a paired batch ``(x_i, y_i)`` into the running state.
 
-        ``batch_x`` and ``batch_y`` must have the same shape; both are
-        flattened internally and upcast to ``float64``.
+        ``batch_x`` and ``batch_y`` must have the same shape; the full
+        tensor is reduced as ``numel`` scalar samples.
         """
         if batch_x.shape != batch_y.shape:
             raise ValueError(
@@ -179,17 +158,17 @@ class WelfordCovariance:
         if batch_x.numel() == 0:
             return
 
-        x64 = batch_x.detach().to(dtype=torch.float64).flatten()
-        y64 = batch_y.detach().to(dtype=torch.float64).flatten()
-        n_b = x64.shape[0]
+        x = batch_x.detach()
+        y = batch_y.detach()
+        n_b = x.numel()
 
-        mean_x_b = x64.mean().to(self._device)
-        mean_y_b = y64.mean().to(self._device)
-        dx_b = x64 - mean_x_b
-        dy_b = y64 - mean_y_b
-        m2_x_b = dx_b.square().sum().to(self._device)
-        m2_y_b = dy_b.square().sum().to(self._device)
-        cxy_b = (dx_b * dy_b).sum().to(self._device)
+        mean_x_b = x.mean()
+        mean_y_b = y.mean()
+        dx_b = x - mean_x_b
+        dy_b = y - mean_y_b
+        m2_x_b = dx_b.square().sum()
+        m2_y_b = dy_b.square().sum()
+        cxy_b = (dx_b * dy_b).sum()
 
         if self.n_samples == 0:
             self.mean_x = mean_x_b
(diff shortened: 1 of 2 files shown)


ignite/metrics/_running_stats.py
```diff
@@ -0,0 +1,273 @@
...
+            for batch in stream:
+                ws.update(batch)
+            print(ws.mean.item(), ws.variance.item(), ws.std.item())
+    """
+
+    n_samples: int
+    mean: torch.Tensor
+    sum_sq_dev_from_mean: torch.Tensor
+
+    def __init__(self, device: Union[str, torch.device] = "cpu") -> None:
```
vfdev-5: This class should not handle device, neither dtype

ignite/metrics/_running_stats.py
```diff
@@ -0,0 +1,273 @@
...
+        and a pairwise algorithm for computing sample variances.
+    https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm
+"""
+
+from typing import Union
+
+import torch
+
+
+class WelfordVariance:
```
vfdev-5, line 7: Let's make it as a dataclass?

ignite/metrics/_running_stats.py
```diff
@@ -0,0 +1,273 @@
...
+    @torch.no_grad()
+    def update(self, batch: torch.Tensor) -> None:
+        """Fold a batch of samples into the running state.
+
+        Empty batches are silently ignored. Inputs of any dtype are
+        upcast to ``float64`` for the internal computation.
+        """
+        if batch.numel() == 0:
+            return
+        batch64 = batch.detach().to(dtype=torch.float64).flatten()
```
vfdev-5: I do not think we should flatten it and for the mean computation we may need to specify the axis (to confirm)

joemunene-by:
Thanks for the review. Pushed 1b82442 addressing all three points.

**1. No device / dtype in the helper.** Dropped the constructor args and all internal `.to(device)` / `.to(torch.float64)` calls. State now takes the dtype and device of the first batch passed to `update`. Callers (R2Score, PearsonCorrelation) are responsible for the float64 upcast, which they were already doing in their own update methods anyway, so this is a no-op for the planned consumers and a cleaner contract for any future caller.

**2. Dataclasses.** Both classes are now `@dataclass` with `field(default_factory=...)` for the tensor fields. Dropped the manual `__init__` and the `reset()` method; "reset" becomes reconstruction (`self.welford = WelfordVariance()`), which fits cleanly into how `Metric.reset()` is normally written in the consumer classes. Tests updated accordingly: `test_reset` is now `test_fresh_instance_has_zero_state` and just asserts the default-factory state.

**3. Flatten.** Removed the explicit `.flatten()` call. `batch.mean()` and `batch.numel()` already reduce over the full tensor regardless of shape, so the current scalar-reduction behavior is preserved and the code reads cleaner for any input shape.

On axis-aware reduction: the current consumers (R2Score, PearsonCorrelation) both produce a single scalar variance / covariance, so they only need the full-reduction behavior. If we want axis support for future per-channel use cases (running variance over `(N, C, H, W)` with reduction over `(0, 2, 3)`), the natural shape of the change is an optional `dim` parameter on `update`:

```python
def update(self, batch: torch.Tensor, dim: Optional[Union[int, tuple[int, ...]]] = None) -> None:
    ...
    n_b = batch.numel() if dim is None else _samples_along(batch.shape, dim)
    mean_b = batch.mean(dim=dim, keepdim=False) if dim is not None else batch.mean()
    ...
```

The state on first update would take the shape of `mean_b`, and `n_samples` would become an `int` count of samples-per-position (still scalar because the reduction is the same at every position).

Two ways we can take it:
- **(a)** Add the `dim` parameter in this PR so the helper ships fully-featured, even though no consumer uses it yet.
- **(b)** Leave it scalar for now; add `dim` as a follow-up the first time a metric actually needs it (avoids speculative API surface).

I lean toward (b): the consumers don't need it, axis-aware running stats have a few subtleties (Bessel correction conventions, mixed-shape merge, axis reordering) that are easier to design alongside a concrete user. Happy to go either way; let me know which you prefer.

ignite/metrics/_running_stats.py
```diff
@@ -0,0 +1,250 @@
...
+    https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm
+"""
+
+from dataclasses import dataclass, field
+
+import torch
+
+
+def _zero() -> torch.Tensor:
+    return torch.tensor(0.0)
```
aaishwarymishra: why we writing a function for a code which is literally 1 line?

aaishwarymishra:
@joemunene-by hi, I was trying to review but I find the code to be quite confusing I think we can try to make more readable :)

ignite/metrics/_running_stats.py
```diff
@@ -0,0 +1,250 @@
...
+            self.sum_sq_dev_from_mean = m2_b
+            self.n_samples = n_b
+            return
+
+        n_a = self.n_samples
+        n_ab = n_a + n_b
+        delta = mean_b - self.mean
+        self.mean = self.mean + delta * n_b / n_ab
+        self.sum_sq_dev_from_mean = self.sum_sq_dev_from_mean + m2_b + delta * delta * n_a * n_b / n_ab
+        self.n_samples = n_ab
```
aaishwarymishra: This part is redundant as merge is already doing it we can just use that.

ignite/metrics/_running_stats.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def merge(self, other: "WelfordVariance") -> None:
```
aaishwarymishra, line 40: Why do we  need merge?

ignite/metrics/_running_stats.py
```diff
@@ -0,0 +1,250 @@
...
+        n_ab = n_a + n_b
+        cross = n_a * n_b / n_ab
+        delta_x = mean_x_b - self.mean_x
+        delta_y = mean_y_b - self.mean_y
+
+        self.mean_x = self.mean_x + delta_x * n_b / n_ab
+        self.mean_y = self.mean_y + delta_y * n_b / n_ab
+        self.sum_sq_dev_x = self.sum_sq_dev_x + m2_x_b + delta_x * delta_x * cross
+        self.sum_sq_dev_y = self.sum_sq_dev_y + m2_y_b + delta_y * delta_y * cross
+        self.sum_product_of_devs = self.sum_product_of_devs + cxy_b + delta_x * delta_y * cross
```
aaishwarymishra: same here

joemunene-by:
@aaishwarymishra thanks, your inline points are all fair. Pushed e153f57 addressing each.

**1. `_zero()` for a one-liner.** Removed. Each tensor field now uses `field(default_factory=lambda: torch.tensor(0.0))` directly. Six lines of one-purpose helper become six explicit defaults.

**2. `update`'s body duplicates `merge`.** Refactored. `update` now builds a single-batch `WelfordVariance` from the input and calls `self.merge(batch_acc)`. The Chan / Welford parallel formula now lives in exactly one place. Same refactor applied to `WelfordCovariance.update` (your "same here" on line 192).

The mental model that fell out of this is worth stating: `update` is the degenerate case of `merge` where the right-hand side has just been built from one batch. There's *one* formula, not two. I lifted that into the module docstring and the method docstrings so the next reader sees it immediately.

**3. "Why do we need `merge`?"** Distributed reduction. In multi-rank training, each rank accumulates its own `WelfordVariance` / `WelfordCovariance` over its local samples; at eval time the ranks fold their accumulators together rank-by-rank to produce the population statistic. Without an explicit `merge`, that cross-rank reduction has to re-iterate the raw data, which defeats the whole point of an online algorithm. I made this rationale explicit in `merge`'s docstring (previously just one line, now a paragraph).

The bonus is that the two-path design (`update` for the local-batch case, `merge` for the cross-state case) collapses to a single formula at the implementation level, which I think also lands the "more readable" feedback from your top-level comment. Let me know if anything's still murky.

cc @vfdev-5 — points from your earlier review are still addressed in 1b82442; this commit is a follow-up to @aaishwarymishra's readability pass. Ready for another look when you have a moment.

joemunene-by:
Copilot's auto-review surfaced four concrete points — pushed [89d5b36](https://github.com/joemunene-by/ignite/commit/89d5b36) addressing each:

1. **Docstring claimed the helpers were already used by `PearsonCorrelation` / `R2Score`.** They aren't yet (that's PRs 2 and 3 of #3748). Reworded to "intended consumers (in follow-up PRs of #3748)" so the doc matches the state at this commit.
2. **Docstring claimed internal state in float64.** It doesn't the helper is intentionally dtype/device agnostic per @vfdev-5's earlier review, and the caller supplies the float64 cast when stability matters. Realigned the prose so users don't read a false safety guarantee on float32 input.
3. **`merge()` lacked `@torch.no_grad()`.** Both `WelfordVariance.merge` and `WelfordCovariance.merge` now have the decorator, matching `update`. A caller merging an accumulator whose tensors still carry grad would have built an autograd graph and leaked memory over time.
4. **`cross_coef = n_a * n_b / n_ab` precomputed as a Python float in `WelfordCovariance.merge`.** Dropped the temporary, inlined `n_a * n_b / n_ab` directly into each parallel-formula line mirrors `WelfordVariance.merge`'s existing style and keeps the arithmetic on the same dtype/device as the tensor operands.

20 / 20 tests still passing locally.

tests/ignite/metrics/test_running_stats.py
```diff
@@ -0,0 +1,288 @@
...
+
+from ignite.metrics._running_stats import WelfordCovariance, WelfordVariance
+
+
+# ---------------------------------------------------------------------------
+# WelfordVariance
+# ---------------------------------------------------------------------------
+
+
+class TestWelfordVariance:
```
aaishwarymishra: Most of our tests are written as pure functions not classes and methods ,I think we should maintain the consitency.

tests/ignite/metrics/test_running_stats.py
```diff
@@ -0,0 +1,288 @@
...
+        # float32 numpy stats would understate the helper's precision.
+        data = rng.standard_normal(1000)
+
+        ws = WelfordVariance()
+        ws.update(torch.from_numpy(data))
+
+        assert ws.n_samples == 1000
+        assert ws.mean.item() == pytest.approx(float(data.mean()), abs=1e-12)
+        assert ws.variance.item() == pytest.approx(float(data.var()), rel=1e-12)
+
```
aaishwarymishra: Test only compares 1 step with external library baseline.

aaishwarymishra:
The docstrings and comments are big and everywhere I think they can be made concise and removed which are not necessary.

Also I am not sure how the merge will work with `sync_all_reduce` can you give example.

ignite/metrics/_running_stats.py
```diff
@@ -0,0 +1,300 @@
...
+        if self.n_samples == 0:
+            return torch.tensor(0.0)
+        return torch.clamp(self.sum_sq_dev_y / self.n_samples, min=0.0)
+
+    @property
+    def covariance(self) -> torch.Tensor:
+        """Population covariance of ``(x, y)`` (divisor ``n``)."""
+        if self.n_samples == 0:
+            return torch.tensor(0.0)
+        return self.sum_product_of_devs / self.n_samples
```
aaishwarymishra, line 182: You did not use `torch.clamp` here like you did in other places.

joemunene-by:
@aaishwarymishra thanks for the second pass. Pushed 1878aeb addressing each point:

**1. "docstrings and comments are big and everywhere":** agreed, they had grown past the point of useful. Trimmed roughly 60 lines from `_running_stats.py` (module + class + method docstrings) without losing the parallel-formula math or the rationale for `merge`. The 33-line module docstring is now 12 lines plus a code block for the distributed pattern.

**2. "how the merge will work with `sync_all_reduce`":** good question. The decorator defaults to `dist.all_reduce(SUM)`, which is not the right operation for Welford state. The parallel formula is not a sum of per-rank means, so a plain SUM-reduce of the mean tensor across ranks would give the wrong combined mean. The pattern I'd reach for instead is `all_gather` of each rank's state followed by a local pairwise `merge`. Now written into the module docstring:

```python

def compute(self):
    ws = self.welford
        n = idist.all_gather(torch.tensor([ws.n_samples]))
        m = idist.all_gather(ws.mean.reshape(1))
        s = idist.all_gather(ws.sum_sq_dev_from_mean.reshape(1))
        ws = WelfordVariance()
        for i in range(len(n)):
            ws.merge(WelfordVariance(int(n[i]), m[i], s[i]))
    return ws.variance
```

I held off on the actual distributed-aware metric wrapper in this PR because the consumer metrics (R2Score in PR 2, PearsonCorrelation in PR 3 of #3748) are where that lives most naturally. Happy to expose a `_merge_all_ranks(welford) -> welford` helper alongside the dataclasses if you'd prefer it sit one level lower.

**3. "tests written as pure functions not classes":** refactored. `TestWelfordVariance` / `TestWelfordCovariance` are gone; everything is now `test_welford_variance_*` / `test_welford_covariance_*` standalone functions, matching the style elsewhere under `tests/ignite/metrics/`.

**4. "compares 1 step with external library baseline":** added `test_welford_variance_matches_numpy_at_each_step` and the covariance equivalent. After every individual update the running stats are asserted against numpy's stats on the cumulative prefix, not just at end of stream. The covariance version also checks Pearson `r` against `np.corrcoef` at each step.

**5. "no `torch.clamp` on line 290":** intentional, and I should have flagged it more clearly in the first place. The variance properties clamp at zero because variance is non-negative by definition; the clamp guards against float rounding producing a tiny negative when all samples are equal. Covariance is legitimately signed (negative correlation gives negative covariance), so the same clamp would silently bias negatively-correlated pairs toward zero. I added an explicit docstring note on the property explaining the asymmetry, and a `test_welford_covariance_negative_correlation_not_clamped` regression that would fail if a future hand introduced the clamp.

23 tests passing locally. Ready for another look when you have a moment.

aaishwarymishra:
@vfdev-5 what you think about this approach of using the merge in compute?

ignite/metrics/_running_stats.py
```diff
@@ -0,0 +1,242 @@
...
+    @property
+    def covariance(self) -> torch.Tensor:
+        """Population covariance of ``(x, y)``.
+
+        No ``torch.clamp`` here because covariance is legitimately signed
+        (negative correlation gives negative covariance). The variance
+        properties clamp at zero to guard against float rounding only;
+        applying the same clamp to covariance would silently bias
+        negatively-correlated pairs toward zero.
+        """
```
aaishwarymishra: It's ok if you are using a coding agent or an llm, but please try to review what it has generated, I don't think we need to mention this in docstring :)

aaishwarymishra:
> **2. "how the merge will work with `sync_all_reduce`":** good question. The decorator defaults to `dist.all_reduce(SUM)`, which is not the right operation for Welford state. The parallel formula is not a sum of per-rank means, so a plain SUM-reduce of the mean tensor across ranks would give the wrong combined mean. The pattern I'd reach for instead is `all_gather` of each rank's state followed by a local pairwise `merge`. Now written into the module docstring:
> 
> ```python
> import ignite.distributed as idist
> 
> def compute(self):
>     ws = self.welford
>     if idist.get_world_size() > 1:
>         n = idist.all_gather(torch.tensor([ws.n_samples]))
>         m = idist.all_gather(ws.mean.reshape(1))
>         s = idist.all_gather(ws.sum_sq_dev_from_mean.reshape(1))
>         ws = WelfordVariance()
>         for i in range(len(n)):
>             ws.merge(WelfordVariance(int(n[i]), m[i], s[i]))
>     return ws.variance
> ```
> 
> I held off on the actual distributed-aware metric wrapper in this PR because the consumer metrics (R2Score in PR 2, PearsonCorrelation in PR 3 of #3748) are where that lives most naturally. Happy to expose a `_merge_all_ranks(welford) -> welford` helper alongside the dataclasses if you'd prefer it sit one level lower.
> 

I think we can just use the welford dataclass class state,
something like 

 ```python
 
 def compute(self):
     collected = idist.all_gather(self.welf)
     result = WelfordVariance()
         for item in collected:
             result.merge(item)
     return result.variance()
 ```

joemunene-by:
@aaishwarymishra going with that. Double-checked `_do_all_gather_object` is wired through on NCCL (Gloo subgroup under the hood) so the object-gather path is safe for our case, just adds one pickle hop per `compute()` for our three small tensors. Negligible.

Two small fix-ups when it goes into the docstring:

```python
def compute(self):
        collected = idist.all_gather(self.welford)
        merged = WelfordVariance()
        for item in collected:
            merged.merge(item)
        return merged.variance     # variance is a @property, not callable
    return self.welford.variance
```

(`self.welford` to match the existing field name in the consumer-side reference, and the inner indentation in your snippet was a copy-paste thing.)

Will push that into the module docstring next, with one line noting that consumers whose state is significantly larger than a few KB should prefer packing into a flat tensor before `all_gather` to skip the pickle hop. Anything else from this round you want bundled in?

ignite/metrics/_running_stats.py
aaishwarymishra: @joemunene-by can you address this :)

joemunene-by: Missed your earlier note, sorry. Trimmed the docstring in 5a13b17, just the one-line description now. The `test_welford_covariance_negative_correlation_not_clamped` regression test still pins the no-clamp behaviour so the constraint isn't lost.

aaishwarymishra:
@vfdev-5 can you have a look it looks good to me and can be merged :)

aaishwarymishra:
Hi @joemunene-by  thank you for your effort , after discussing this pr we have some concerns .

1. API doesn't fit ignite's conventions
update() and merge() on a plain dataclass sits outside how ignite metrics are structured. Metrics use the Metric base class with the standard update() / compute() / reset() lifecycle. The helper needs to either fit that pattern or stay completely hidden as an implementation detail.
2. DDP doesn't work with this approach
The all_gather + pairwise merge pattern bypasses ignite's standard sync_all_reduce mechanism. This needs a tested solution that works within ignite's actual DDP setup, not just a docstring example.
3. No end-to-end usage example
The tests cover the helper in isolation. We need a concrete example — in the tests or PR description — showing this being used inside a real ignite metric, with the mean=1e6, std=1 failure case from #3662 demonstrating the fix.

Also it would be cleaner if you resolve the comment once it has been adderessed :)

joemunene-by:
Thanks, all three addressed in `b8fd9b61`.

1. Kept the helper private and pure (no `Metric` base, no distributed deps). To show it fits a real metric I added `WelfordCorrelation` in the tests: a minimal `Metric` over `WelfordCovariance` with the standard `reset` / `update` / `compute` lifecycle.
2. You're right that `sync_all_reduce(SUM)` can't combine Welford state. The metric's `compute()` reduces by `all_gather` of each rank's state plus pairwise `merge`, and there's a `TestDistributed` that runs it across ranks and checks against the full gathered data. I also verified the merge on 2 gloo ranks offline: it matches numpy on the concatenated data to ~2e-16.
3. `test_welford_correlation_metric_recovers_3662_case` drives the `mean=1e6` case from #3662 through the metric and asserts the recovered r where the naive `E[X^2]-E[X]^2` formula cancels in float32.

The multi-step-vs-baseline tests and the deliberate no-clamp on covariance are already in from the earlier round. Resolving the line threads now.

ignite/metrics/_running_stats.py
joemunene-by: `merge` is the cross-rank reduction primitive. Welford state (mean, M2) is not additive, so it cannot go through `all_reduce(SUM)` the way the naive running sums do; combining two accumulators needs the Chan parallel formula. The new distributed test exercises exactly this: each rank builds its own accumulator and `compute()` folds them across ranks with `merge`, checked against the full gathered data. `update` now reuses it for the single-batch case too, so there is one code path.

tests/ignite/metrics/test_running_stats.py
joemunene-by: The unit tests are all plain functions now. The one `TestDistributed` class is the existing distributed-test convention in the repo (same shape as `PearsonCorrelation`'s `TestDistributed`); the `distributed` fixture is `scope="class"`, so the rank/backend setup is shared across the methods of a class. Happy to flatten it if you would rather keep everything as functions here too.

PR #3753, opened 2026-05-12
docs: add warmup scheduler mixed-events example

### LINKED ISSUE #2441: create_lr_scheduler_with_warmup does not change init_lr to proper value
Hi,

In order to get expected sequence of `lr`s from `create_lr_scheduler_with_warmup`'s scheduler, one must not attach it to engine on event `EPOCH_COMPLETED` because it produces the `lr` passed to optimizer's constructor at the beginning and then warmup `lr`s. This hurts warmup procedure.  As a workaround, one could use event `EPOCH_STARTED` but it might not be a good solution.

It seems there should be something like below in [line 1017](https://github.com/pytorch/ignite/blob/0666b407f7cdba81842014c6026e33b66113bb94/ignite/handlers/param_scheduler.py#L1017) of param_scheduler.py within the for loop .

```python
param_group['lr'] = warmup_start_value
```

To reproduce current behaviour:

```python
param = nn.Parameter(torch.Tensor([3.]))
optimizer = torch.optim.SGD([param], lr=1e-3)
scheduler = StepLR(optimizer, 3)
with_warmup_scheduler = create_lr_scheduler_with_warmup(scheduler, warmup_start_value=1e-5, warmup_duration=3)

def process_func(e,b):
  param.grad = torch.Tensor([1.])
  optimizer.step()
trainer = Engine(process_func)
def _():
  print(op.param_groups[0]['lr'])
trainer.add_event_handler(Events.EPOCH_COMPLETED, with_warmup_scheduler)
```
output:
```
0.001
1e-05
0.000505
0.001
0.001
0.001
0.0001
0.0001
0.0001
1e-05
```

nightcityblade:
Fixes #2441

Description:

Adds a docstring example showing how to combine Ignite event filters so warm-up runs on iteration starts and the wrapped scheduler runs on later epoch starts.

Check list:

### DIFF
--- a/ignite/handlers/param_scheduler.py
+++ b/ignite/handlers/param_scheduler.py
@@ -1139,16 +1139,17 @@ def create_lr_scheduler_with_warmup(
             trainer = Engine(lambda engine, batch: None)
             optimizer = SGD([torch.zeros(1, requires_grad=True)], lr=0.1)
             torch_lr_scheduler = ExponentialLR(optimizer, gamma=0.5)
+            warmup_duration = 5
             scheduler = create_lr_scheduler_with_warmup(
-                torch_lr_scheduler, warmup_start_value=0.0, warmup_duration=5
+                torch_lr_scheduler, warmup_start_value=0.0, warmup_duration=warmup_duration
             )
 
             epoch_length = 8
             combined_events = Events.ITERATION_STARTED(
-                event_filter=lambda engine, event: engine.state.iteration <= 5
+                event_filter=lambda engine, event: event <= warmup_duration
             )
             combined_events |= Events.EPOCH_STARTED(
-                event_filter=lambda engine, event: engine.state.epoch > 1 + 5 / epoch_length
+                event_filter=lambda engine, event: event > 1 + warmup_duration / epoch_length
             )
             trainer.add_event_handler(combined_events, scheduler)
 
--- a/ignite/handlers/param_scheduler.py
+++ b/ignite/handlers/param_scheduler.py
@@ -1125,6 +1125,33 @@ def create_lr_scheduler_with_warmup(
             0.09223...
             0.09039...
 
+        The warm-up and post-warm-up schedulers can also be attached to different events by combining event filters.
+        For example, run the warm-up on iterations and the wrapped scheduler at the start of each later epoch:
+
+        .. code-block:: python
+
+            import torch
+            from ignite.engine import Engine, Events
+            from ignite.handlers import create_lr_scheduler_with_warmup
+            from torch.optim import SGD
+            from torch.optim.lr_scheduler import ExponentialLR
+
+            trainer = Engine(lambda engine, batch: None)
+            optimizer = SGD([torch.zeros(1, requires_grad=True)], lr=0.1)
+            torch_lr_scheduler = ExponentialLR(optimizer, gamma=0.5)
+            scheduler = create_lr_scheduler_with_warmup(
+                torch_lr_scheduler, warmup_start_value=0.0, warmup_duration=5
+            )
+
+            epoch_length = 8
+            combined_events = Events.ITERATION_STARTED(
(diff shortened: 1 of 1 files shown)


ignite/handlers/param_scheduler.py
```diff
@@ -1127,6 +1125,33 @@ def print_lr():
...
+            trainer = Engine(lambda engine, batch: None)
+            optimizer = SGD([torch.zeros(1, requires_grad=True)], lr=0.1)
+            torch_lr_scheduler = ExponentialLR(optimizer, gamma=0.5)
+            scheduler = create_lr_scheduler_with_warmup(
+                torch_lr_scheduler, warmup_start_value=0.0, warmup_duration=5
+            )
+
+            epoch_length = 8
+            combined_events = Events.ITERATION_STARTED(
+                event_filter=lambda engine, event: engine.state.iteration <= 5
```
TahaZahid05:
```suggestion
                event_filter=lambda engine, event: event <= 5
```

We already have the count inside `event`

TahaZahid05:
@vfdev-5 This PR solves the documentation issue. However, I think I agree with #2441 pushback on maybe this needs an API fix. Currently, we are teaching the user to do the math on when to switch. It would be better if this logic can be internal to improve user experience.

Something like:

```python
scheduler = create_lr_scheduler_with_warmup(
    torch_lr_scheduler,
    warmup_duration=5,
    warmup_event=Events.ITERATION_STARTED,
    post_warmup_event=Events.EPOCH_STARTED
)
# The scheduler would internally handle the event-switching logic.
trainer.add_event_handler(Events.ITERATION_STARTED | Events.EPOCH_STARTED, scheduler)
```

nightcityblade:
Thanks for calling that out. I agree the more ergonomic long-term answer is probably an API improvement so users do not need to manually compute the switch point. I kept this PR narrowly scoped to the current docs issue because it only documents the behavior that exists today, but I’m happy for maintainers to treat the API direction separately if that’s preferred.

ignite/handlers/param_scheduler.py
```diff
@@ -1127,6 +1125,33 @@ def print_lr():
...
+            from ignite.engine import Engine, Events
+            from ignite.handlers import create_lr_scheduler_with_warmup
+            from torch.optim import SGD
+            from torch.optim.lr_scheduler import ExponentialLR
+
+            trainer = Engine(lambda engine, batch: None)
+            optimizer = SGD([torch.zeros(1, requires_grad=True)], lr=0.1)
+            torch_lr_scheduler = ExponentialLR(optimizer, gamma=0.5)
+            scheduler = create_lr_scheduler_with_warmup(
+                torch_lr_scheduler, warmup_start_value=0.0, warmup_duration=5
```
TahaZahid05: let's first define the variable `warmup_duration=5` and then use it rather than the raw values themselves.

TahaZahid05:
@nightcityblade let us keep this PR to update the docs for now. if you feel that this needs an API change in the future you can also give your reasoning on the original issue and the maintainer can decide if we need it or not. Kindly update the PR with the changes requested.

ignite/handlers/param_scheduler.py
nightcityblade: Thanks for the review — updated this to keep the change docs-only, define `warmup_duration = 5`, and reuse it in the scheduler and event filters.

nightcityblade:
Thanks again. I re-checked the open review threads and the current branch keeps this docs-only while addressing the requested changes: it defines `warmup_duration` once, reuses it in the scheduler and event filters, and uses the filtered event counter directly for the warm-up condition.\n\nValidation: ran the documented snippet locally against this branch and it completes successfully.

vfdev-5 (approved):
Looks good to me
@TahaZahid05 I let you approve and merge it if good to you as well

TahaZahid05 (approved):

PR #3754, opened 2026-05-15
ci: declare workflow-level `contents: read` on 8 workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on 8 workflows in `.github/workflows/` that don't call a GitHub API beyond the initial checkout.

The following files were left implicit because they reference `GITHUB_TOKEN` / use a write-scope action / trigger on `pull_request_target`. Those scopes are best declared by maintainers: `docker-build.yml`, `pytorch-version-tests.yml`.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

### DIFF
--- a/.github/workflows/code-style-checks.yml
+++ b/.github/workflows/code-style-checks.yml
@@ -38,6 +38,9 @@ concurrency:
   group: code-style-${{ github.ref_name }}-${{ !(github.ref_protected) || github.sha }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   code-style:
     runs-on: ubuntu-latest
--- a/.github/workflows/gpu-hvd-tests.yml
+++ b/.github/workflows/gpu-hvd-tests.yml
@@ -17,6 +17,9 @@ concurrency:
 
 # Cherry-picked from https://github.com/pytorch/test-infra/blob/main/.github/workflows/linux_job.yml
 
+permissions:
+  contents: read
+
 jobs:
   gpu-hvd-tests:
     strategy:
--- a/.github/workflows/gpu-tests.yml
+++ b/.github/workflows/gpu-tests.yml
@@ -17,6 +17,9 @@ concurrency:
 
 # Cherry-picked from https://github.com/pytorch/test-infra/blob/main/.github/workflows/linux_job_v2.yml
 
+permissions:
+  contents: read
+
 jobs:
   gpu-tests:
     strategy:
--- a/.github/workflows/hvd-tests.yml
+++ b/.github/workflows/hvd-tests.yml
@@ -23,6 +23,9 @@ concurrency:
   group: hvd-tests-${{ github.ref_name }}-${{ !(github.ref_protected) || github.sha }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   horovod-tests:
     runs-on: ubuntu-latest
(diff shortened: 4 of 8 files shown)


TahaZahid05:
@arpitjain099 Thanks for the PR! The changes look good. However, can you check other files in the workflow as well to see if they also need to be changed? Seems to me there might be other files that can also be included in this PR?

vfdev-5:
We do not have pull_request_target in our GHA workflows. Why should we add that?
What we can do is to put sha256 here: https://github.com/pytorch/ignite/blob/300feb595ff512c19a89d7b33f2855141c5f100b/.github/workflows/triage.yml#L16-L19
instead of versions like v6 for more security.

arpitjain099:
> @arpitjain099 Thanks for the PR! The changes look good. However, can you check other files in the workflow as well to see if they also need to be changed? Seems to me there might be other files that can also be included in this PR?

Hi @TahaZahid05 sure. Let me check it tomorrow

arpitjain099:
Thanks @vfdev-5. To clarify, this PR does not add `pull_request_target` anywhere. It only sets the default `GITHUB_TOKEN` scope to `contents: read` at the workflow level on these workflows. The motivation is least privilege for the token itself: by default the token is granted broad write scopes, so if any third-party action or a transitive dependency in a job is compromised it can push to the repo or tamper with releases. Pinning the default to read shrinks that blast radius regardless of whether any workflow uses `pull_request_target`.

You are right that SHA-pinning the actions (for example the `@v*` refs in `triage.yml`) is valuable too. The two are complementary rather than alternatives: pinning protects against a tag being moved to malicious code, while the permissions default limits what a compromised-but-pinned action can do once it runs. I am happy to add SHA-pinning in this PR or a separate one, whichever you prefer, and I can extend the `contents: read` default to the remaining workflows per @TahaZahid05's note. Let me know which direction you would like and I will update.

vfdev-5:
@arpitjain099 @TahaZahid05 we do not need to specify everywhere `contents: read` as we already have overall "Read repository contents and packages permissions" option in the Settings/Actions. I suggest to close this PR and open a new one addressing this concern: https://github.com/pytorch/ignite/pull/3754#issuecomment-4623823964

arpitjain099:
> @arpitjain099 @TahaZahid05 we do not need to specify everywhere `contents: read` as we already have overall "Read repository contents and packages permissions" option in the Settings/Actions. I suggest to close this PR and open a new one addressing this concern: https://github.com/pytorch/ignite/pull/3754#issuecomment-4623823964

@vfdev-5 ok let me do that tomorrow

arpitjain099:
Makes sense. Since your Settings/Actions default is already "Read repository contents and packages permissions," the per-workflow `contents: read` is redundant here, so I'll close this.

The part that actually adds supply-chain protection is pinning the third-party action references to full commit SHAs (the `@v*` tags in `triage.yml` and the other workflows can be moved to a moved-tag if the tag is repointed). I'll open a focused PR for that per your suggestion. Thanks for the review.

PR #3755, opened 2026-05-21
Add pathlib Path support to idist broadcast

### LINKED ISSUE #3025: Extension of `ignite.distributed.utils.broadcast` for pathlib.Path datatype
## 🚀 Feature

I would like to request for extension of `ignite.distributed.utils.broadcast` for Path datatype as it is frequently used for distributed training and can be very useful for designing DL pipeline based projects. The API should work like this 

```
from pathlib import Path

path = Path("./log")

idist.broadcast(path, src=0)
```
Further if possible we can also provide support for more data types as suggested by others. Also if possible I would like to take up this issue.
@vfdev-5 @sadra-barikbin

nightcityblade:
Fixes #3025

Description:

Adds `pathlib.Path` support to `idist.broadcast` by encoding paths through the existing string broadcast path and decoding them back to `Path` objects, including safe-mode placeholder handling.

Check list:

### DIFF
--- a/ignite/distributed/comp_models/base.py
+++ b/ignite/distributed/comp_models/base.py
@@ -122,6 +122,8 @@ class ComputationModel(metaclass=ABCMeta):
         elif isinstance(x, str):
             encoded_msg[0] = 2
         elif isinstance(x, Path):
+            # Keep a distinct tag so non-source ranks in safe mode can reconstruct
+            # Path placeholders and return Path values rather than strings.
             encoded_msg[0] = 3
         return encoded_msg
 
--- a/ignite/distributed/comp_models/base.py
+++ b/ignite/distributed/comp_models/base.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from abc import ABCMeta, abstractmethod
 from collections.abc import Callable
 from numbers import Number
+from pathlib import Path
 from typing import Any, cast
 
 import torch
@@ -105,7 +106,7 @@ class ComputationModel(metaclass=ABCMeta):
         return cast(int, size.item())
 
     @staticmethod
-    def _encode_input_data(x: torch.Tensor | float | str | None, is_src: bool) -> list[int]:
+    def _encode_input_data(x: torch.Tensor | float | str | Path | None, is_src: bool) -> list[int]:
         encoded_msg = [-1] * 512
         if not is_src:
             # Discard input type if not source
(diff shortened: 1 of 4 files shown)


ignite/distributed/comp_models/base.py
```diff
@@ -275,6 +282,9 @@ def broadcast(
         if tensor_to_str:
             list_str = self._decode_str(tensor)
             return list_str[0]
+        if tensor_to_path:
+            list_str = self._decode_str(tensor)
+            return Path(list_str[0])
```
vfdev-5: We can merge these two cases into a single one and just apply Path on str if tensor_to_path is True

vfdev-5 (commented):
Thanks for the PR @nightcityblade 
As a quick solution this may work. Later we can probably support more types with dist.broadcast_object_list (and this method is present in pytorch 2.2 min supported version by ignite, so we can keep bc etc)

ignite/distributed/comp_models/base.py
nightcityblade: Done — merged the string/path decode handling into a single branch and now only wrap with Path when tensor_to_path is true. I also ran `python3 -m pytest tests/ignite/distributed/comp_models/test_base.py` locally (5 passed).

ignite/distributed/comp_models/base.py
```diff
@@ -120,10 +121,12 @@ def _encode_input_data(x: torch.Tensor | float | str | None, is_src: bool) -> li
             encoded_msg[0] = 1
         elif isinstance(x, str):
             encoded_msg[0] = 2
+        elif isinstance(x, Path):
+            encoded_msg[0] = 3
```
vfdev-5, line 127: @nightcityblade can you please check if we really need encoded_msg[0] = 3 and can't just use :
```python
elif isinstance(x, (str, Path)):
    encoded_msg[0] = 2
```

nightcityblade:
Thanks @vfdev-5, I checked whether `encoded_msg[0] = 3` can be folded into the string case. I think we still need the separate Path tag: on non-source ranks with `safe_mode=True` (for example when passing `None`), `_setup_placeholder` has to reconstruct a `Path` placeholder so `broadcast` can return a `Path` rather than a plain string. If Path used the string tag (`2`), the existing Path safe-mode case would decode as `str`.

I pushed a small clarification comment in `ComputationModel._encode_input_data` documenting that.

Tests run locally:
- `python3 -m pytest tests/ignite/distributed/comp_models/test_base.py tests/ignite/distributed/utils/test_serial.py -q` (10 passed)

vfdev-5 (approved):
Thanks @nightcityblade !

PR #3758, opened 2026-05-25
Fix Distributed Initialization by Validating Legacy tcp:// and Updating Tests

TahaZahid05:
### **Problem**
Using `tcp://` directly in PyTorch's `init_method` causes multi-process communication to hang and fail in modern PyTorch versions.
  
### **Solution**
1. **Validation & User Error**: Added explicit validation in `Parallel` and `_NativeDistModel` that detects legacy `tcp://` initialization methods and raises a helpful, actionable `ValueError` guiding the user on how to use        PyTorch's native `TCPStore` instead.
2. **Updated Tests**: Cleaned up the distributed test suite to remove legacy `tcp://` connections from test parameterizations and updated low-level context tests to use modern `dist.TCPStore` objects directly.

Check list:

### DIFF
--- a/ignite/distributed/comp_models/native.py
+++ b/ignite/distributed/comp_models/native.py
@@ -54,6 +54,7 @@ if has_native_dist_support:
             init_method: str | None = None,
             world_size: int | None = None,
             rank: int | None = None,
+            store: Any | None = None,
             **kwargs: Any,
         ) -> _NativeDistModel:
             if backend not in _NativeDistModel.available_backends:
@@ -62,7 +63,7 @@ if has_native_dist_support:
             if dist.is_available() and dist.is_initialized():
                 raise RuntimeError("Can not create new distributed process group if default one is already initialized")
 
-            if init_method is None:
+            if init_method is None and store is None:
                 if world_size is not None or rank is not None:
                     raise ValueError("Arguments rank and world_size should be None if no init_method is provided")
             else:
@@ -72,7 +73,7 @@ if has_native_dist_support:
                     raise ValueError(f"Both rank and world_size should be provided, but given {rank} and {world_size}")
 
             return _NativeDistModel(
-                backend=backend, init_method=init_method, world_size=world_size, rank=rank, **kwargs
+                backend=backend, init_method=init_method, world_size=world_size, rank=rank, store=store, **kwargs
             )
 
         def __init__(
@@ -82,6 +83,7 @@ if has_native_dist_support:
             init_method: str | None = None,
             world_size: int | None = None,
             rank: int | None = None,
+            store: Any | None = None,
             **kwargs: Any,
         ) -> None:
             """This is a private method. Please, use `create_from_backend` or `create_from_context`"""
@@ -93,7 +95,13 @@ if has_native_dist_support:
             self._init_method: str | None = None
             if backend is not None:
                 self._create_from_backend(
-                    backend, timeout=timeout, init_method=init_method, world_size=world_size, rank=rank, **kwargs
+                    backend,
+                    timeout=timeout,
+                    init_method=init_method,
+                    world_size=world_size,
+                    rank=rank,
+                    store=store,
+                    **kwargs,
                 )
             else:
                 self._init_from_context()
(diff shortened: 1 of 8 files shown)


ignite/distributed/comp_models/native.py
```diff
@@ -107,6 +107,30 @@ def _create_from_backend(
...
+                try:
+                    parsed = urlparse(init_method)
+                    host = parsed.hostname
+                    port = parsed.port
+                except Exception:
+                    host = None
+                    port = None
+
+                host = host or "127.0.0.1"
+                port = port or 29500
```
aaishwarymishra: Maybe we can simply add default values in the except block instead of making them none

ignite/distributed/launcher.py
```diff
@@ -233,6 +233,30 @@ def __init__(
...
+            try:
+                parsed = urlparse(init_method)
+                host = parsed.hostname
+                port = parsed.port
+            except Exception:
+                host = None
+                port = None
+
+            host = host or "127.0.0.1"
+            port = port or 29500
```
aaishwarymishra: same here

ignite/distributed/comp_models/native.py
TahaZahid05: `urlparse` can pass without returning None for some invalid strings like `tcp://localhost`. `urlparse` won't give Exception here, but `parsed.port` would be None.

ignite/distributed/launcher.py
TahaZahid05: same reason as above.

ignite/distributed/comp_models/native.py
TahaZahid05: I have tried another way to avoid using None. let me know if its better.

ignite/distributed/comp_models/native.py
```diff
@@ -107,6 +107,27 @@ def _create_from_backend(
...
+            if init_method is not None and init_method.startswith("tcp://"):
+                from urllib.parse import urlparse
+
+                try:
+                    parsed = urlparse(init_method)
+                    host = parsed.hostname or "127.0.0.1"
+                    port = parsed.port or 29500
+                except Exception:
+                    host = "127.0.0.1"
+                    port = 29500
```
vfdev-5: If parsing has failed we should not provide an arbitrary values for the error message. I would remove this code and update the error message without giving exact host and port

TahaZahid05: ok, if parsing does pass then should we keep the message with exact host and port or keep the same error message in both scenarios i.e without giving exact host and port?

ignite/distributed/comp_models/native.py
```diff
@@ -107,6 +107,27 @@ def _create_from_backend(
...
+
+                raise ValueError(
+                    f"TCP initialization via init_method='{init_method}' will hang. "
+                    "To fix this, please configure a TCPStore and initialize the process group using the store instead. "
+                    "For example:\n\n"
+                    "    import torch.distributed as dist\n"
+                    "    from datetime import timedelta\n\n"
+                    f'    store = dist.TCPStore("{host}", {port}, world_size, is_master, timedelta(seconds=30))\n'
+                    "    dist.init_process_group(backend, store=store, rank=rank, world_size=world_size)"
+                )
```
vfdev-5: This example is also misleading. We run in this function when we create the dist group using ignite distributed, why the example is showing pytorch distributed code.

ignite/distributed/comp_models/native.py
vfdev-5: Make an appropriate choice

ignite/distributed/launcher.py
vfdev-5: We should also update this example and use store=TCPStore

ignite/distributed/comp_models/native.py
TahaZahid05: Used placeholder `<master_addr>` and `<master_port>`

ignite/distributed/comp_models/native.py
```diff
@@ -93,7 +94,12 @@ def __init__(
             self._init_method: str | None = None
             if backend is not None:
                 self._create_from_backend(
-                    backend, timeout=timeout, init_method=init_method, world_size=world_size, rank=rank, **kwargs
+                    backend,
```
vfdev-5, line 97: There is no change in this, let's revert the code formatting.

tests/ignite/distributed/comp_models/test_native.py
```diff
@@ -334,7 +334,8 @@ def _test__native_dist_model_create_from_context_set_local_rank(true_conf):
 def _test__native_dist_model_create_from_context_no_dist(true_backend, true_device):
     assert _NativeDistModel.create_from_context() is None
 
-    dist.init_process_group(true_backend, "tcp://0.0.0.0:2222", world_size=1, rank=0)
+    store = dist.TCPStore("0.0.0.0", 2222, world_size=1, is_master=True)
```
vfdev-5, line 338: We had a fixture to get a freeport in the conf.py, can we use it here?

tests/ignite/distributed/comp_models/test_native.py
```diff
@@ -397,7 +400,7 @@ def test__native_dist_model_create_no_dist_nccl(clean_env):
 
 
 @pytest.mark.distributed
-@pytest.mark.parametrize("init_method", [None, "tcp://0.0.0.0:22334", "FILE"])
+@pytest.mark.parametrize("init_method", [None, "FILE"])
```
vfdev-5, line 404: Instead of removing a test, let's do store arg instead

TahaZahid05:
Closed in favor of #3775

PR #3760, opened 2026-05-31
feat: add NoImprovementHandler

### LINKED ISSUE #2314: Add no improvement handler (similar to early stopping handler)

Aanushka001:
### Description

Adds a `NoImprovementHandler` that stops training when metrics stop improving for a set number of epochs. This is a simpler alternative to the existing `EarlyStopping` handler.

### Key Features
- **Simpler interface**: Only requires `patience`, `score_function`, `trainer`, and optional `mode` parameter
- **Mode support**: Works with both maximization (`'max'`, default) and minimization (`'min'`) objectives
- **Serializable**: Supports `state_dict()` and `load_state_dict()` for checkpointing
- **Resettable**: Supports `reset()` method for reuse across training runs
- **Attachable**: Standard `attach()` method for easy integration with ignite engines

### Changes
- New file: `ignite/handlers/no_improvement_handler.py`
- Updated: `ignite/handlers/__init__.py` to export `NoImprovementHandler`

Closes #2314

### DIFF
--- a/ignite/handlers/__init__.py
+++ b/ignite/handlers/__init__.py
@@ -3,6 +3,7 @@ from ignite.engine.events import Events
 from ignite.handlers.checkpoint import Checkpoint, DiskSaver, ModelCheckpoint
 from ignite.handlers.clearml_logger import ClearMLLogger
 from ignite.handlers.early_stopping import EarlyStopping
+from ignite.handlers.no_improvement_handler import NoImprovementHandler
 from ignite.handlers.ema_handler import EMAHandler
 from ignite.handlers.fbresearch_logger import FBResearchLogger
 from ignite.handlers.lr_finder import FastaiLRFinder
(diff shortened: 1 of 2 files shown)


aaishwarymishra:
Hi, I think the issue meant this handler should take a custom function or method and execute it in case of no improvement. Currently the implementation looks a loot like early stopping instead of a modular handler.

Aanushka001:
Thanks for the feedback! I've updated the PR to add a `custom_action`
parameter. Users can now pass any callable to execute on no improvement
(e.g., reducing LR, checkpointing, logging, or terminating) instead of
being restricted to just trainer termination. The trainer parameter is now
optional when custom_action is provided.

Here's the commit:
https://github.com/Aanushka001/ignite/commit/9da711677cd88ab90dcb7054a34dfd29103305c4

Would appreciate a re-review when you get a chance!

TahaZahid05:
@Aanushka001 have you checked the implementation of the previous PR #2574 ? It seems the issue and the PR points towards also changing `EarlyStopping`? Any pros and cons of both approaches?

PR #3765, opened 2026-05-31
feat: add empty_mps_cache handler for MPS device memory clearing

Aanushka001:
## Description

This PR adds MPS (Apple Metal Performance Shaders) cache clearing support, similar to the existing `empty_cuda_cache()` for CUDA devices.

## Changes

- Added `empty_mps_cache()` function that clears MPS cache via `torch.mps.empty_cache()`
- Registered the MPS cache clearing handler in `_setup_common_training_handlers()` when running on MPS devices

## Related

Based on original PR #3763

### DIFF
--- a/ignite/contrib/engines/common.py
+++ b/ignite/contrib/engines/common.py
@@ -11,7 +11,7 @@ from torch.utils.data.distributed import DistributedSampler
 
 # https://github.com/pytorch/ignite/issues/2773
 try:
-    from torch.optim.lr_scheduler import LRScheduler as PyTorchLRScheduler
+    from torch.opt.lr_scheduler import LRScheduler as PyTorchLRScheduler
 except ImportError:
     from torch.optim.lr_scheduler import _LRScheduler as PyTorchLRScheduler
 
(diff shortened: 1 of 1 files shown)


ignite/contrib/engines/common.py
```diff
@@ -287,6 +291,13 @@ def empty_cuda_cache(_: Engine) -> None:
     gc.collect()
 
 
+def empty_mps_cache(_: Engine) -> None:
+    if hasattr(torch, "mps") and hasattr(torch.mps, "empty_cache"):
+        torch.mps.empty_cache()
+    import gc
```
aaishwarymishra, line 297: imports should be on top

ignite/contrib/engines/common.py
```diff
@@ -287,6 +291,13 @@ def empty_cuda_cache(_: Engine) -> None:
     gc.collect()
 
 
+def empty_mps_cache(_: Engine) -> None:
+    if hasattr(torch, "mps") and hasattr(torch.mps, "empty_cache"):
```
aaishwarymishra, line 295: either we can check in the function or outside it, currently irt checks in both place which make it kinda redundant.

aaishwarymishra:
Hi, I also think we should add a separate parameter for the mps cache clearing instead of relying on pre_built cuda @vfdev-5

PR #3772, opened 2026-06-01
Expose metric names in logger setup helpers

hariharan077:
Fixes #829.

## Summary
- add `trainer_metric_names` and `evaluator_metric_names` options to the simplified logger setup helpers
- keep the existing default behavior of logging `metric_names="all"`
- allow users to filter out unsupported metrics such as 2D `ConfusionMatrix` tensors when using `setup_tb_logging`

## Tests
- `python -m ruff format ignite/handlers/logger_utils.py tests/ignite/contrib/handlers/test_logger_utils.py`
- `python -m ruff check ignite/handlers/logger_utils.py tests/ignite/contrib/handlers/test_logger_utils.py`
- `python -m pytest tests/ignite/contrib/handlers/test_logger_utils.py::test__setup_logging_metric_names tests/ignite/contrib/handlers/test_logger_utils.py::test_setup_tb_logging tests/ignite/contrib/handlers/test_logger_utils.py::test_setup_tb_logging_evaluator_metric_names_filter_tensor_metrics`

PR #3774, opened 2026-06-01
Fix CI linkcheck and MLflow file store

hariharan077:
Split from #3772 per maintainer request so the CI-only fix can be reviewed and merged separately.

## Summary
- ignore unstable external docs anchors that currently break `linkcheck`
- set `MLFLOW_ALLOW_FILE_STORE=true` for the test suite so existing MLflow tests can use local file-store tracking

## Verification
- `python -m ruff format --check docs/source/conf.py tests/ignite/conftest.py`
- `python -m ruff check docs/source/conf.py tests/ignite/conftest.py`
- `git diff --check`
- `python -m pytest tests/ignite/handlers/test_mlflow_logger.py`
- `cd docs && rm -rf source/generated build/linkcheck build/doctrees && sphinx-build -b linkcheck -W --keep-going source build/linkcheck`


vfdev-5 (approved):
Thanks @hariharan077 !

PR #3775, opened 2026-06-02
Skipping tcp tests, fixing ci

TahaZahid05:
Description:
This PR resolves the CI hangs and provides a clear guide to users on how to migrate away from the legacy `tcp://` initialization method

Check list:

### DIFF
--- a/tests/ignite/conftest.py
+++ b/tests/ignite/conftest.py
@@ -231,15 +231,19 @@ def _setup_free_port(local_rank):
 @pytest.fixture()
 def distributed_context_single_node_nccl(local_rank, world_size):
     free_port = _setup_free_port(local_rank)
+    os.environ["MASTER_ADDR"] = "localhost"
+    os.environ["MASTER_PORT"] = str(free_port)
 
     dist_info = {
         "backend": "nccl",
         "world_size": world_size,
         "rank": local_rank,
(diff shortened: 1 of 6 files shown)


tests/ignite/conftest.py
```diff
@@ -251,22 +255,32 @@ def distributed_context_single_node_gloo(local_rank, world_size):
         # can't use backslashes in f-strings
         backslash = "\\"
         init_method = f"file:///{temp_file.name.replace(backslash, '/')}"
+        dist_info = {
```
vfdev-5: Why should we duplicate the configs here?
Let's to the else clause as previously and create dist_info globally

tests/ignite/conftest.py
```diff
@@ -472,16 +486,21 @@ def distributed(request, local_rank, world_size):
             # can't use backslashes in f-strings
             backslash = "\\"
             init_method = f"file:///{temp_file.name.replace(backslash, '/')}"
+            dist_info = {
```
vfdev-5: Same here, let's revert these changes duplicating dist_info

TahaZahid05:
@vfdev-5 done.

vfdev-5 (approved):
Thanks @TahaZahid05 !

PR #3778, opened 2026-06-08
Add show_epoch option to ProgressBar

### LINKED ISSUE #3118: More control over ProgressBar formatting
## 🚀 Feature - Improved ``contrib.handlers.tqdm_logger.ProgressBar`` formatting and alignment

More configurable arguments/bar formatting of the available ``ProgressBar`` class. Specifically to allow for the alignment of multiple bars (train loop, test/validation loop).

The current functionality of the ``ProgressBar`` class modifies the bar format to include the ``current_epoch``/``total_epochs`` when  ``max_epochs > 1``. This mostly works well for the train loop. However, format/alignment issues arise when the train loop is used with a validation loop that runs at the end of each epoch (on ``EpochCompleted``). The validation loop should only be run once. However, this causes the epoch information to be omitted, and consequently, any format options which attempt to align the bars are void (see the screenshot below; both progress bars are set to have the same formatting). The format given below may not format as intended, but I cannot test it since I cannot remove/add the epoch information. 

Ideally, control would be given over the if/how the epoch information is displayed when creating the ``ProgressBar``. Investigating the codebase, it seems that this stuff is handled internally in the ``_OutputHandler`` class. I would very much like to align the bars to start and finish in the same place as each other, purely for aesthetics. 

``BAR_FMT = "{desc:<15} [{n_fmt}/{total_fmt}]|{bar}| {rate_fmt}{postfix}"``

image

In the above image, I would love control over aligning the actual bar sections of the progress bars.

importstring:
Fixes #3118

## What

Add a `show_epoch: bool | None = None` parameter to `ProgressBar` to give users
explicit control over whether epoch information appears in the progress bar
description.

## Why

Currently, `ProgressBar` automatically appends `Epoch [current/total]` to the
description when `max_epochs > 1`. This works well for a training loop but causes
misalignment when a validation loop also uses a custom `bar_format` — the validation
bar omits epoch information (since it runs once per epoch, the closing event count
is 1), while the training bar includes it. With a shared `bar_format` like:

```python
BAR_FMT = "{desc:<15} [{n_fmt}/{total_fmt}]|{bar}| {rate_fmt}{postfix}"
```

the bars no longer align. There is currently no way to control this behavior.

## How

- Add `show_epoch: bool | None = None` to `ProgressBar.__init__`.
- Store as `self.show_epoch` and pass it through to `_OutputHandler`.
- Update `_OutputHandler.__call__` description logic:
  - `None` (default): preserves existing behavior — `"Iteration"` when `max_epochs == 1`,
    `"Epoch [x/y]"` otherwise.
  - `True`: always uses `"Epoch"` as the default description and appends `[x/y]`
    when the closing event is an epoch event.
  - `False`: always uses `"Iteration"` and never appends epoch information, even
    when `max_epochs > 1`.
- Also removes a stray example line from `_OutputHandler`'s docstring that was
  not valid documentation text.

## Backwards compatibility

`show_epoch` defaults to `None`, so all existing code continues to behave exactly
as before. The new behavior only activates when users explicitly pass
`show_epoch=True` or `show_epoch=False`.

## Testing

All existing tqdm logger tests pass. Two new tests added:
- `test_pbar_show_epoch_true_on_evaluator`: verifies that `show_epoch=True` on an
  evaluator forces epoch information in the description even when the evaluator
  runs once per epoch.
- `test_pbar_show_epoch_false_on_trainer`: verifies that `show_epoch=False`
  suppresses epoch information even when `max_epochs > 1`.

### DIFF
--- a/ignite/handlers/tqdm_logger.py
+++ b/ignite/handlers/tqdm_logger.py
@@ -17,12 +17,15 @@ class ProgressBar(BaseLogger):
 
     Args:
         persist: set to ``True`` to persist the progress bar after completion (default = ``False``)
-        bar_format : Specify a custom bar string formatting. May impact performance.
+        bar_format: Specify a custom bar string formatting. May impact performance.
             [default: '{desc}[{n_fmt}/{total_fmt}] {percentage:3.0f}%|{bar}{postfix} [{elapsed}<{remaining}]'].
             Set to ``None`` to use ``tqdm`` default bar formatting: '{l_bar}{bar}{r_bar}', where
             l_bar='{desc}: {percentage:3.0f}%|' and
             r_bar='| {n_fmt}/{total_fmt} [{elapsed}<{remaining}, {rate_fmt}{postfix}]'. For more details on the
             formatting, see `tqdm docs <https://tqdm.github.io/docs/tqdm/>`_.
+        show_epoch: if ``True``, always display epoch information in the progress bar description.
+            If ``False``, never display epoch information. If ``None``, preserve the existing
+            behavior and display epoch information only when the number of epochs is greater than one.
         tqdm_kwargs: kwargs passed to tqdm progress bar.
             By default, progress bar description displays "Epoch [5/10]" where 5 is the current epoch and 10 is the
             number of epochs; however, if ``max_epochs`` are set to 1, the progress bar instead displays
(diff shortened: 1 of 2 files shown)


vfdev-5:
Thanks for the contribution, but we do not want to have `show_epoch` arg in ProgressBar.

importstring:
Hey @vfdev-5 
I found some spelling errors and extra lines here and there. Would you like me to submit a seperate PR fixing those.

For example: "will looks like" is found in the code among 5-10 other examples.

vfdev-5:
Yes, you can submit a separate PR for docs typos fixing. Thanks!

PR #3783, opened 2026-06-08
DOC: fix minor docstring grammar in events.py

importstring:
This PR fixes a few small docstring grammar issues in `ignite/engine/events.py`. There are **no code or behavior changes**, only wording fixes in comments/docstrings.

All changes are minimal and localized, for example:

```python
# Before
"""Single Event containing a filter, specifying whether the event should
be run at the current event (if the event type is correct)

Args:
    value: The actual enum value. Only needed for internal use. Do not touch!
    event_filter: A function taking the engine and the current event value as input and returning a
        boolean to indicate whether this event should be executed. Defaults to None, which will result to a
        function that always returns `True`
    name: The enum-name of the current object. Only needed for internal use. Do not touch!
"""
```

```python
# After
"""Single event containing a filter specifying whether the handler should be run for the current event (if the event type is correct).

Args:
    value: The actual enum value. Only needed for internal use. Do not touch!
    event_filter: A function taking the engine and the current event value as input and returning a
        boolean to indicate whether this event should be executed. Defaults to None, which will result in a
        function that always returns `True`.
    name: The enum-name of the current object. Only needed for internal use. Do not touch!
"""
```

The deprecated default event filter docstring is also cleaned up: [is is]

```python
# Before
"""Default event filter. This method is is deprecated and will be removed. Please, use None instead"""

# After
"""Default event filter. This method is deprecated and will be removed. Please use None instead."""
```

These changes:

- Fix a duplicated word: `is is`.
- Replace “result to” with the correct “result in”.
- Make the `CallableEventWithFilter` docstring clearer and grammatically correct.
- Fixed: "User can also combine events" --> "Users can also ..."

### DIFF
--- a/ignite/engine/events.py
+++ b/ignite/engine/events.py
@@ -23,7 +23,7 @@ class CallableEventWithFilter:
     Args:
         value: The actual enum value. Only needed for internal use. Do not touch!
         event_filter: A function taking the engine and the current event value as input and returning a
-            boolean to indicate whether this event should be executed. Defaults to None, which will result to a
+            boolean to indicate whether this event should be executed. Defaults to None, which will result in a 
             function that always returns `True`
         name: The enum-name of the current object. Only needed for internal use. Do not touch!
     """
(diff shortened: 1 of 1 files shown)


vfdev-5 (approved):
Thanks

aaishwarymishra:
@importstring Code style checks are failing pls try to use pre-commit hook :)

importstring:
@aaishwarymishra Alright done. Let me know if this works!

PR #3638, opened 2026-03-05
Feat: Implement Word Error Rate (WER) metric

Manimaran-tech:
## 🚀 Feature: WER and CER Metrics
Resolves #3634 

This PR implements \WordErrorRate\ and \CharacterErrorRate\ natively within \ignite.metrics.nlp\ to evaluate ASR and OCR models, using an internal, dependency-free Levenshtein distance calculation to remain lightweight.

**Changes:**
- Added robust Levenshtein Distance dynamic programming solver.
- Implemented \WordErrorRate\ which splits strings via tokens to compute word-level permutations.
- Implemented \CharacterErrorRate\ which computes exact string edit distance.
- Integrated the new classes into the \ignite.metrics.nlp\ initializers.
- Provided 100% test coverage mimicking existing Pytest tests in Ignite.

PR #3785, opened 2026-06-09
feat: Add CharacterErrorRate (CER) metric to ignite.metrics.nlp

### LINKED ISSUE #3634: [Feature] Add Word Error Rate (WER) and Character Error Rate (CER) metrics to ignite.metrics.nlp
🚀 Feature

`ignite/metrics/nlp/` currently has two metrics:
- `Bleu`
- `Rouge`

I would like to propose adding `WordErrorRate` (WER) and `CharErrorRate` (CER).

**Motivation:**
Both existing metrics in `ignite/metrics/nlp/` are designed for generative tasks where multiple valid outputs exist. `Bleu` and `Rouge` reward partial matches and alternative wordings.

WER and CER provide a necessary counter-perspective  they evaluate sequences strictly on exact transcription using Levenshtein edit distance counting the absolute minimum substitutions, deletions,  and insertions. A single hallucinated word or dropped character lowers the score immediately, making them strict exact-match validation metrics.

Naively averaging batch-level WER gives mathematically incorrect results because sequences have variable lengths. The correct 
approach requires global accumulation across the entire epoch:

WER = (total_S + total_D + total_I) / total_N

 Users currently have no native way to do this correctly they have to install external libraries like `jiwer` and write custom handlers.

TorchMetrics and HuggingFace `evaluate` already include `WordErrorRate` and `CharErrorRate` as core NLP metrics.

**Use Cases:**
- A user fine-tuning Whisper or Wav2Vec2 can simply attach `WordErrorRate` to their Ignite `Engine` to monitor exact transcription quality epoch by epoch. This is helpful for  ensuring their ASR model isn't quietly hallucinating or  dropping words as training progresses.

- If a user needs smarter checkpointing, they can log WER  directly to their `EarlyStopping` handler. This is highly  useful since cross-entropy loss often plateaus before actual  transcription quality does — the lowest WER score points right to the optimal checkpoint.

- If a user is training an OCR or Document AI model, attaching `CharErrorRate` to their validation loop gives a strict character-level accuracy check  critical when a single  character error (like misreading a financial figure) is a  severe failure.

**References showing real usage:**
- HuggingFace (2021) — Official Wav2Vec2 fine-tuning blog explicitly installs `pip install jiwer` just to compute WER:
  https://huggingface.co/blog/fine-tune-wav2vec2-english

- OpenAI Whisper original paper (arxiv, 2022) — uses WER and CER as the primary evaluation metrics throughout all   experiments:
  https://arxiv.org/abs/2212.04356

- Towards Data Science (2021) — practical blog showing WER and CER used together to evaluate OCR models with Python 
  code examples using TesseractOCR:
  https://towardsdatascience.com/evaluating-ocr-output-quality-with-character-error-rate-cer-and-word-error-rate-wer-853175297510/

Would love to hear your thoughts on whether this addition would be helpful!

Shubh3005:
## Summary

Implements `CharacterErrorRate` (CER) as a follow-up to #3638 (WER), resolving #3634.

CER measures the edit distance at the character level — used in ASR and OCR evaluation where a single character error (e.g. misreading a financial figure) is a severe failure.

## Changes

- `ignite/metrics/nlp/character_error_rate.py` — CER metric inheriting from `_BaseErrorRate`, using character-level Levenshtein distance
- `tests/ignite/metrics/nlp/test_character_error_rate.py` — 15 test cases covering: identical sequences, single deletion/insertion/substitution, empty inputs, batch accumulation, multi-update accumulation, reset, single string input, whitespace as character, unicode
- `ignite/metrics/nlp/__init__.py` — exports `CharacterErrorRate`

## Design

Follows the same structure as `word_error_rate.py` and `bleu.py`:
- Separate file per metric (per maintainer feedback in #3634)
- Inherits `_BaseErrorRate` from `word_error_rate.py`
- Only difference from WER: `_tokenize` returns `list(text)` instead of `text.split()`

Closes #3634

### DIFF
--- a/ignite/metrics/nlp/character_error_rate.py
+++ b/ignite/metrics/nlp/character_error_rate.py
@@ -85,7 +85,7 @@ class CharacterErrorRate(Metric):
     def reset(self) -> None:
         self._num_errors = torch.tensor(0.0, device=self._device)
         self._num_refs = torch.tensor(0.0, device=self._device)
-        self._num_examples = 0
+        self._num_examples = torch.tensor(0.0, device=self._device)
 
     @reinit__is_reduced
     def update(self, output: Sequence[str]) -> None:
--- a/ignite/metrics/nlp/character_error_rate.py
+++ b/ignite/metrics/nlp/character_error_rate.py
@@ -89,11 +89,11 @@ class CharacterErrorRate(Metric):
     @reinit__is_reduced
     def update(self, output: Sequence[str]) -> None:
         y_pred, y = output[0], output[1]
+        if not isinstance(y_pred, (str, list)) or not isinstance(y, (str, list)):
+            raise TypeError(f"y_pred and y must be str or list[str], got y_pred: {type(y_pred)} and y: {type(y)}")
         if isinstance(y_pred, str) and isinstance(y, str):
             y_pred = [y_pred]
             y = [y]
-        if not isinstance(y_pred, list) or not isinstance(y, list):
-            raise TypeError(f"y_pred and y must be str or list[str], got y_pred: {type(y_pred)} and y: {type(y)}")
         if not all(isinstance(p, str) for p in y_pred) or not all(isinstance(r, str) for r in y):
             raise TypeError("All elements of y_pred and y must be strings.")
         if len(y_pred) != len(y):
@@ -103,6 +103,8 @@ class CharacterErrorRate(Metric):
         errors = 0.0
         refs = 0.0
         for p, r in zip(y_pred, y):
+            if len(r) == 0:
+                continue
             errors += _edit_distance(r, p)
             refs += len(r)
         self._num_errors += torch.tensor(errors, device=self._device)
--- a/ignite/metrics/nlp/character_error_rate.py
+++ b/ignite/metrics/nlp/character_error_rate.py
@@ -52,7 +52,9 @@ class CharacterErrorRate(Metric):
         skip_unrolling: specifies whether output should be unrolled before being fed to update method.
 
     Examples:
-        .. code-block:: python
+        For more information on how metric works with :class:`~ignite.engine.engine.Engine`, visit :ref:`attach-engine`.
+
+        .. testcode::
 
             from ignite.metrics.nlp import CharacterErrorRate
 
@@ -62,7 +64,11 @@ class CharacterErrorRate(Metric):
             y = ["the cat sat on mat", "hello world"]
 
             cer.update((y_pred, y))
-            print(cer.compute())  # Output: 0.1724... (5 insertions / 29 reference chars)
+            print(round(cer.compute(), 4))
+
+        .. testoutput::
+
+            0.1379
 
     .. versionadded:: 0.5.2
     """
@@ -90,14 +96,10 @@ class CharacterErrorRate(Metric):
             y = [y]
         if not isinstance(y_pred, (str, list)) or not isinstance(y, (str, list)):
             raise TypeError(
-                f"All inputs should be either str or list[str], "
-                f"got y_pred: {type(y_pred)} and y: {type(y)}"
+                f"All inputs should be either str or list[str], got y_pred: {type(y_pred)} and y: {type(y)}"
             )
         if type(y_pred) is not type(y):
-            raise TypeError(
-                f"y_pred and y must be the same type, "
-                f"got y_pred: {type(y_pred)} and y: {type(y)}"
-            )
+            raise TypeError(f"y_pred and y must be the same type, got y_pred: {type(y_pred)} and y: {type(y)}")
         if len(y_pred) != len(y):
             raise ValueError(
                 f"y_pred and y must have the same length. Got y_pred of length {len(y_pred)} and y of length {len(y)}."
@@ -114,5 +116,7 @@ class CharacterErrorRate(Metric):
     @sync_all_reduce("_num_errors", "_num_refs")
     def compute(self) -> Number:
         if self._num_refs == 0:
-            raise NotComputableError("CharacterErrorRate must have at least one valid reference sequence to be computed.")
+            raise NotComputableError(
+                "CharacterErrorRate must have at least one valid reference sequence to be computed."
+            )
         return (self._num_errors / self._num_refs).item()
--- a/ignite/metrics/nlp/character_error_rate.py
+++ b/ignite/metrics/nlp/character_error_rate.py
@@ -85,7 +85,7 @@ class CharacterErrorRate(Metric):
     def reset(self) -> None:
         self._num_errors = torch.tensor(0.0, device=self._device)
         self._num_refs = torch.tensor(0.0, device=self._device)
-        super().reset()
+        super().reset()  # type: ignore[misc]
 
     @reinit__is_reduced
     def update(self, output: Sequence[str]) -> None:
--- a/ignite/metrics/nlp/character_error_rate.py
+++ b/ignite/metrics/nlp/character_error_rate.py
@@ -1,16 +1,33 @@
-from typing import Callable, Sequence
+from typing import Any, Callable, Sequence
 
 import torch
 from torch.types import Number
 
 from ignite.exceptions import NotComputableError
 from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
-from ignite.metrics.nlp.word_error_rate import _BaseErrorRate, _edit_distance
 
 __all__ = ["CharacterErrorRate"]
 
 
-class CharacterErrorRate(_BaseErrorRate):
+def _edit_distance(ref: Sequence[Any], pred: Sequence[Any]) -> int:
+    """Computes the Levenshtein distance between two sequences."""
+    n, m = len(ref), len(pred)
+    if n == 0:
+        return m
+    if m == 0:
+        return n
+    dp = list(range(m + 1))
+    for i in range(1, n + 1):
+        prev_diag = dp[0]
+        dp[0] = i
+        for j in range(1, m + 1):
+            temp = dp[j]
+            dp[j] = prev_diag if ref[i - 1] == pred[j - 1] else min(dp[j - 1], dp[j], prev_diag) + 1
+            prev_diag = temp
+    return dp[m]
+
+
+class CharacterErrorRate(Metric):
     r"""Calculates the Character Error Rate (CER).
 
     CER is defined as the total number of errors (substitutions, deletions, and insertions)
@@ -30,14 +47,9 @@ class CharacterErrorRate(_BaseErrorRate):
     Args:
         output_transform: a callable that is used to transform the
             :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
-            form expected by the metric. This can be useful if, for example, you have a multi-output model and
-            you want to compute the metric with respect to one of the outputs.
-        device: specifies which device updates are accumulated on. Setting the metric's
-            device to be the same as your ``update`` arguments ensures the ``update`` method is non-blocking. By
-            default, CPU.
-        skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
-            true for multi-output model, for example, if ``y_pred`` contains multi-ouput as ``(y_pred_a, y_pred_b)``
-            Alternatively, ``output_transform`` can be used to handle this.
+            form expected by the metric.
+        device: specifies which device updates are accumulated on. By default, CPU.
+        skip_unrolling: specifies whether output should be unrolled before being fed to update method.
 
     Examples:
         .. code-block:: python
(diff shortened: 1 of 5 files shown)


Shubh3005:
"Updated in commit 329d981: made self-contained — inlined _edit_distance and full Metric implementation directly, removing dependency on unmerged PR #3638. All 15 tests pass on master."

aaishwarymishra:
Hi, @Shubh3005 we really appreciate your pr, but I would still recommend you check withe owner of the other open pr if they are still working on it or not just be sure :)

If not then we can move forward with your pr.

Shubh3005:
Thank you! I'll reach out to @Manimaran-tech on PR #3638 to check their status. Will follow up shortly.

Shubh3005:
Hi @aaishwarymishra, following up after reaching out to @Manimaran-tech on PR #3638 — I haven't heard back from them in the two weeks since. Given the WER PR appears inactive, would it make sense to move forward with this one? Happy to rebase against master if that helps move things along. Let me know if there's anything else you'd like me to address.

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,108 @@
...
+            y_pred = [y_pred]
+            y = [y]
+        if len(y_pred) != len(y):
+            raise ValueError(
+                f"y_pred and y must have the same length. Got y_pred of length {len(y_pred)} and y of length {len(y)}."
+            )
+        errors = 0.0
+        refs = 0.0
+        for p, r in zip(y_pred, y):
+            p_chars = list(p)
```
aaishwarymishra: why we need this explicit list here

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,108 @@
...
+    @reinit__is_reduced
+    def reset(self) -> None:
+        self._num_errors = torch.tensor(0.0, device=self._device)
+        self._num_refs = torch.tensor(0.0, device=self._device)
+        super().reset()
+
+    @reinit__is_reduced
+    def update(self, output: Sequence[str]) -> None:
+        y_pred, y = output[0], output[1]
+        if isinstance(y_pred, str) and isinstance(y, str):
```
aaishwarymishra, line 95: there is no explicit check whether y_pred and y are strings.

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,108 @@
...
+        dp[0] = i
+        for j in range(1, m + 1):
+            temp = dp[j]
+            dp[j] = prev_diag if ref[i - 1] == pred[j - 1] else min(dp[j - 1], dp[j], prev_diag) + 1
+            prev_diag = temp
+    return dp[m]
+
+
+class CharacterErrorRate(Metric):
+    r"""Calculates the Character Error Rate (CER).
```
aaishwarymishra, line 31: I think you should explicitly mention the input types expected type, like we are also handling the case of raw strings but its not metioned.

aaishwarymishra:
Also there is some problems with the docs.

Also add metric in `docs/source/metrics.rst`

Shubh3005:
Thanks for the review @aaishwarymishra!

Addressing all three points:
1. The explicit `list(p)` is for character tokenization — converting a string to a list of individual characters. Happy to simplify to direct iteration if you prefer, will clean that up.
2. Will add explicit type validation for the list case to match WER's handling.
3. Will update the docstring to explicitly mention input types and add the entry to `docs/source/metrics.rst`.

Will push fixes shortly.

Shubh3005:
Hi @aaishwarymishra, addressed all three points in the latest commit:

1. **`list(p)` / `list(r)`** — removed. Strings are already character sequences so direct iteration is equivalent and cleaner. `_edit_distance` works on any `Sequence` so passing the string directly is fine.
2. **Type validation** — added explicit `isinstance` checks for the list case, raising `TypeError` if elements are not strings. Covers both `y_pred` and `y`.
3. **Docstring + metrics.rst** — updated docstring to explicitly document that inputs can be `str` or `list of str`, and added `CharacterErrorRate` to `docs/source/metrics.rst`.

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,113 @@
...
+        if isinstance(y_pred, str) and isinstance(y, str):
+            y_pred = [y_pred]
+            y = [y]
+        if not all(isinstance(p, str) for p in y_pred):
+            raise TypeError("All elements of y_pred must be strings.")
+        if not all(isinstance(r, str) for r in y):
+            raise TypeError("All elements of y must be strings.")
+        if len(y_pred) != len(y):
+            raise ValueError(
+                f"y_pred and y must have the same length. Got y_pred of length {len(y_pred)} and y of length {len(y)}."
```
aaishwarymishra: something like this 

```python
        if not all(isinstance(p, str) or isinstance(p, list) for p in [y, y_pred]):
          raise TypeError(f"All the inputs should be either str or list[str] got y_pred: {type(y_pred)} or y: {type(y)}")

        different_type = isinstance(y, str) and isinstance(y_pred, list) or isinstance(y_pred, str) and isinstance(y, list)
       
       if different_type:
          raise TypeError(f"All the inputs should be of of same type got y_pred: {type(y_pred)} or y: {type(y)}")
```

Shubh3005: Combined the type check upfront to validate both inputs are str or list, added a mixed-type check to reject str/list mismatches, and kept the element-level validation for the list case. All 15 tests still passing.

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,123 @@
...
+        if type(y_pred) is not type(y):
+            raise TypeError(
+                f"y_pred and y must be the same type, "
+                f"got y_pred: {type(y_pred)} and y: {type(y)}"
+            )
+        if isinstance(y_pred, list) and not all(isinstance(p, str) for p in y_pred):
+            raise TypeError("All elements of y_pred must be strings.")
+        if isinstance(y, list) and not all(isinstance(r, str) for r in y):
+            raise TypeError("All elements of y must be strings.")
+        if len(y_pred) != len(y):
```
aaishwarymishra: This looks like un necessary loops, I think this check can be moved in the lower `  in the `_edit_distance` itself

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,123 @@
...
+    at the character level divided by the total number of characters in the reference sequence.
+
+    .. math::
+        \text{CER} = \frac{S + D + I}{N} = \frac{S + D + I}{S + D + C}
+
+    where :math:`S` is the number of substitutions, :math:`D` is the number of deletions,
+    :math:`I` is the number of insertions, :math:`C` is the number of correct characters,
+    and :math:`N` is the total number of characters in the reference (:math:`N = S + D + C`).
+
+    - ``update`` must receive output of the form ``(y_pred, y)`` or ``{'y_pred': y_pred, 'y': y}``.
```
aaishwarymishra: The recieve output maybe confusing it should be input, also I dont think this method can handle dict so I think we can remove that from the docstring.

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,123 @@
...
+    - `y_pred` must be either a ``str`` or a list of ``str`` (predicted sequences).
+    - `y` must be either a ``str`` or a list of ``str`` (reference sequences).
```
aaishwarymishra:
```suggestion
    - `y_pred` and `y` both must be either  ``str`` or  list of ``str`` .
```

aaishwarymishra (commented):
Other things look ok, we can merge after these changes, al;os please resolve the comments you have fixed it keeps the pr clean :)

ignite/metrics/nlp/character_error_rate.py
Shubh3005: Thanks @aaishwarymishra! Addressed in latest commit:
1. Removed the redundant element-level isinstance loops from update()
2. Fixed docstring: "receive output" → "receive input", removed unsupported dict form  
3. Consolidated y_pred/y description per your suggestion

Shubh3005:
Thanks @aaishwarymishra! Addressed in latest commit:

Removed the redundant element-level isinstance loops from update()
Fixed docstring: "receive output" → "receive input", removed unsupported dict form
Consolidated y_pred/y description per your suggestion

aaishwarymishra:
Hi doc tests are failing :(
Also the code style check, please use pre-commit hook for that.

Shubh3005:
Hi @aaishwarymishra, fixed both issues:
1. Switched example from `.. code-block:: python` to `.. testcode::` / `.. testoutput::` format (matching Bleu's convention)
2. Ran pre-commit hooks — ruff-format applied, all checks passing

aaishwarymishra:
I don't think we need to explicitly call super().reset() you can check other metrics implementation for reference.

Shubh3005:
Thanks @aaishwarymishra! Removed the `super().reset()` call — confirmed `Metric.reset()` is abstract with no body so calling it was unnecessary. Checked other metrics for reference as suggested. All 15 tests still passing, pre-commit clean. Ready for merge! 🙏

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,121 @@
...
+import torch
+from torch.types import Number
+
+from ignite.exceptions import NotComputableError
+from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
+
+__all__ = ["CharacterErrorRate"]
+
+
+def _edit_distance(ref: Sequence[Any], pred: Sequence[Any]) -> int:
```
aaishwarymishra: why we using any ? I think we should be explicit and try to avoid any.

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,121 @@
...
+        if type(y_pred) is not type(y):
+            raise TypeError(f"y_pred and y must be the same type, got y_pred: {type(y_pred)} and y: {type(y)}")
+        if len(y_pred) != len(y):
+            raise ValueError(
+                f"y_pred and y must have the same length. Got y_pred of length {len(y_pred)} and y of length {len(y)}."
+            )
+        errors = 0.0
+        refs = 0.0
+        for p, r in zip(y_pred, y):
+            # Strings are already character sequences — no need for explicit list()
```
aaishwarymishra:
```suggestion
```

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,121 @@
...
+
+            0.1379
+
+    .. versionadded:: 0.5.2
+    """
+
+    def __init__(
+        self,
+        output_transform: Callable = lambda x: x,
+        device: Union[str, torch.device] = torch.device("cpu"),
```
aaishwarymishra: why we started usining union from `|`

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,121 @@
...
+        self._num_refs = torch.tensor(0.0, device=self._device)
+
+    @reinit__is_reduced
+    def update(self, output: Sequence[str]) -> None:
+        y_pred, y = output[0], output[1]
+        # Handle single-string input — treat as a one-element batch
+        if isinstance(y_pred, str) and isinstance(y, str):
+            y_pred = [y_pred]
+            y = [y]
+        if not isinstance(y_pred, (str, list)) or not isinstance(y, (str, list)):
```
aaishwarymishra: Now I look at it this check for str is redundant as we wrap strs in lift on above block.

Also now I think about it we should add check for all the elements are str outsize updation loop to not corrupt the metric previous values.

aaishwarymishra:
Hi @Shubh3005 I appreciate you being helping. If you are using an agent its's fine, but its generally good Idea to check whats its generating. It randomly replace `|` with `Union` also added the unnecessary comments. :)

tests/ignite/metrics/nlp/test_character_error_rate.py
```diff
@@ -0,0 +1,99 @@
+import pytest
+from ignite.exceptions import NotComputableError
+from ignite.metrics.nlp import CharacterErrorRate
+
+
+class TestCharacterErrorRate:
```
aaishwarymishra: Ignite does not use class based tests :)

Shubh3005:
Hi @aaishwarymishra, I've been reviewing more carefully now.

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -89,24 +89,20 @@ def reset(self) -> None:
...
         if isinstance(y_pred, str) and isinstance(y, str):
             y_pred = [y_pred]
             y = [y]
-        if not isinstance(y_pred, (str, list)) or not isinstance(y, (str, list)):
-            raise TypeError(
-                f"All inputs should be either str or list[str], got y_pred: {type(y_pred)} and y: {type(y)}"
-            )
-        if type(y_pred) is not type(y):
-            raise TypeError(f"y_pred and y must be the same type, got y_pred: {type(y_pred)} and y: {type(y)}")
+        if not isinstance(y_pred, list) or not isinstance(y, list):
```
aaishwarymishra: This check be before the  `if isinstance(y_pred, str) and isinstance(y, str):`

tests/ignite/metrics/nlp/test_character_error_rate.py
```diff
@@ -0,0 +1,97 @@
...
+    assert cer.compute() == pytest.approx(1.0)
+
+
+def test_cer_empty_prediction():
+    cer = CharacterErrorRate()
+    cer.update(([""], ["hello"]))
+    assert cer.compute() == pytest.approx(1.0)
+
+
+def test_cer_empty_reference_raises():
```
aaishwarymishra: I think empty reference string is valid, we should have a fallback value instead of error.

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -0,0 +1,117 @@
...
+        if len(y_pred) != len(y):
+            raise ValueError(
+                f"y_pred and y must have the same length. Got y_pred of length {len(y_pred)} and y of length {len(y)}."
+            )
+        errors = 0.0
+        refs = 0.0
+        for p, r in zip(y_pred, y):
+            errors += _edit_distance(r, p)
+            refs += len(r)
+        self._num_errors += torch.tensor(errors, device=self._device)
```
aaishwarymishra: we can use simple scaler addition instead of converting values to tensor explicitly

aaishwarymishra:
I think these are last few changes

Shubh3005:
Awesome added 3 changes let me know if anything else needed.

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -103,6 +103,8 @@ def update(self, output: Sequence[str]) -> None:
         errors = 0.0
         refs = 0.0
         for p, r in zip(y_pred, y):
+            if len(r) == 0:
```
aaishwarymishra: isnt it wrong? well say for example
```python
 cer = CharacterErrorRate()
 cer.update((["hello world", "hello"], ["hello world", ""]))
```

For this answer should be 
pair 1 both are same 0 errors.
pair 2 they are not same error = 5
result should be 5/11.

What we need to handle.
case 1. errors > 0, n==0, in this case we should return error rate to be 1.

case 2. error == 0 and n==0, and we  got no input and we just called compute, in this case trow non computable error.

case3. error==0 and n==0, but the input were sent to the metric the pair were both empty strings [""],[""] its a valid input we should return 0 as there are no error.

aaishwarymishra: for the example, I picked the test cases you added which is wrong, there should be test for all these cases .

Shubh3005:
Added and fixed those 3 tests :)

aaishwarymishra:
@vfdev-5 It looks good and can be merged, link check is failing but I think its not related to this pr :)

vfdev-5:
@aaishwarymishra approve the PR it it is looks good to you

aaishwarymishra (approved):

aaishwarymishra:
@vfdev-5 done

PR #3789, opened 2026-06-17
feat: extend EpochMetric to support more compute_fn output types

### LINKED ISSUE #1757: More EpochMetric's compute_fn output types

zongyang078:
Fixes #1757

Description:
Extends `EpochMetric` to support more output types returned by `compute_fn`, as requested in #1757.

Previously, `compute_fn` could only return scalar values (int/float). This PR adds support for:
- `torch.Tensor`
- `tuple` / `list` of tensors
- `Mapping` (dict) with string keys and tensor values
- arbitrarily nested combinations of the above (e.g. a dict of tuples)

A recursive `_check_output_type` method validates the output, including nested containers. Unsupported types raise a clear `TypeError`.

Distributed handling (world_size > 1):
- All supported output types — including `tuple`/`list`/`Mapping`, and nested combinations of these — are recursively broadcast from rank 0 to every other rank
- Type and mapping-key validation happens inside the broadcast protocol itself (via a type-tag broadcast before each collective call), so every rank raises the same `TypeError` together instead of only rank 0 validating and exiting early while other ranks hang on a broadcast rank 0 never issues

Tests:
- Tensor, tuple, list, dict outputs (with value verification)
- Unsupported type (str) raises TypeError
- Nested invalid type raises TypeError
- Mapping with non-str key raises TypeError
- Distributed (`WORLD_SIZE=2`, gloo): container outputs, nested container outputs, and symmetric TypeError on invalid output/keys across all ranks

Check list:
- [x] New tests are added
- [x] Documentation is updated (docstring reflects new supported types and distributed behavior)


ignite/metrics/epoch_metric.py
```diff
@@ -156,14 +161,22 @@ def compute(self) -> float:
                 _prediction_tensor = cast(torch.Tensor, idist.all_gather(_prediction_tensor))
                 _target_tensor = cast(torch.Tensor, idist.all_gather(_target_tensor))
 
-            self._result = 0.0
+            result: float | torch.Tensor | Sequence | Mapping = 0.0
```
aaishwarymishra: I dont think we need this type hinting here.

ignite/metrics/epoch_metric.py
```diff
@@ -142,7 +147,7 @@ def update(self, output: tuple[torch.Tensor, torch.Tensor]) -> None:
             except Exception as e:
                 warnings.warn(f"Probably, there can be a problem with `compute_fn`:\n {e}.", EpochMetricWarning)
 
-    def compute(self) -> float:
+    def compute(self) -> float | torch.Tensor | Sequence | Mapping:
```
aaishwarymishra: why not make this union a new type.

ignite/metrics/epoch_metric.py
```diff
@@ -156,14 +161,22 @@ def compute(self) -> float:
...
                 _target_tensor = cast(torch.Tensor, idist.all_gather(_target_tensor))
 
-            self._result = 0.0
+            result: float | torch.Tensor | Sequence | Mapping = 0.0
             if idist.get_rank() == 0:
                 # Run compute_fn on zero rank only
-                self._result = self.compute_fn(_prediction_tensor, _target_tensor)
+                result = self.compute_fn(_prediction_tensor, _target_tensor)
+
+            if not isinstance(result, (int, float, torch.Tensor, Sequence, Mapping)) or isinstance(result, str):
```
aaishwarymishra: A new type will make and checks simpler, also I dont think we need to check for string explicitly.

tests/ignite/metrics/test_epoch_metric.py
aaishwarymishra, line 1: Test coverage should try to cover each datatype.

ignite/metrics/epoch_metric.py
zongyang078: Removed the inline type annotation on the local variable.

ignite/metrics/epoch_metric.py
zongyang078: Created a new `EpochMetricOutput` type alias and updated all annotations to use it.

ignite/metrics/epoch_metric.py
zongyang078: Kept the explicit str check — str is a subclass of Sequence in Python, so isinstance("hello", Sequence) returns True. Without this check, strings would silently pass as valid output. Added test_epoch_metric_compute_fn_invalid_output to demonstrate this behavior.

tests/ignite/metrics/test_epoch_metric.py
zongyang078: Added test coverage for each supported datatype:
- test_epoch_metric_compute_fn_tensor_output: torch.Tensor output
- test_epoch_metric_compute_fn_tuple_output: tuple of tensors
- test_epoch_metric_compute_fn_list_output: list of tensors
- test_epoch_metric_compute_fn_dict_output: dict (Mapping) of tensors
- test_epoch_metric_compute_fn_invalid_output: unsupported type (str) raises TypeError

aaishwarymishra, line 1: Just to be sure we also should check if the output values are as expected or correct after the computation.

Abhisri436:
Hi! I’m working on issue #1757 for a course contribution and noticed this PR already covers the main source changes.

I added value-checking tests on my fork for scalar tensor, vector tensor, tuple/list of tensors, and mapping outputs. These tests check the actual computed values, which seems related to the review comment about verifying output correctness.

Test commit: https://github.com/Abhisri436/ignite/commit/56345826

Would it be useful for me to adapt these tests into this PR or keep them as a separate reference?

aaishwarymishra:
@Abhisri436 It's generally not a good Idea to implement the changes as an external pr until explicitly mentioned that the change  if a future improvement or can be done separately :)

tests/ignite/metrics/test_epoch_metric.py
zongyang078: Added value verification for all output types — each test now checks that the result matches directly calling compute_fn on the concatenated data.

aaishwarymishra:
@zongyang078 Are you panning to work on it?

zongyang078:
> @zongyang078 Are you panning to work on it?

Yes, I'm planning to continue! I've reviewed PR #3802 and found several improvements worth incorporating, particularly the recursive _check_output_type method and the conservative distributed handling with NotImplementedError. I'll update this PR with these improvements shortly.

zongyang078:
The refactor has been pushed. Changes include:

1. Extracted recursive `_check_output_type` method for nested container validation
2. Conservative distributed handling with NotImplementedError for tuple/list/mapping outputs when world_size > 1
3. Status-code broadcast mechanism to ensure symmetric failure across ranks
4. Added tests for nested invalid types and non-str mapping keys
5. Updated docstring
All 17 non-distributed tests pass locally.

Also thanks to @Abhisri436 for the discussion in #3802 — the safer distributed approach was a valuable direction to explore.

Abhisri436:
Hi, thank you for the acknowledgement. I’m glad the safer distributed approach was useful.

Since the latest version of #3789 now incorporates the recursive validation, status-code distributed handling, `safe_mode=True` broadcast, and conservative `NotImplementedError` behavior first introduced in #3802, I think it would be good to consolidate the two PRs rather than continue with duplicate implementations.

#3802 also includes permanent `WORLD_SIZE=2` distributed tests that verify this implementation under real multi-rank execution. These tests remain a distinct part of #3802 and provide coverage that is not currently present in #3789.

Could a maintainer advise on the preferred consolidation path and the appropriate way to preserve attribution for the implementation originally introduced in #3802?

aaishwarymishra:
Hi @zongyang078 sorry, I missed your previous pushes.
I have few quenstion's

I think the ignite boradcast method itself checks for types in `ignite/distributed/comp_models/base.py` and will throw error for sequence and mapping even after this check in distributed setting.

zongyang078:
> Hi @zongyang078 sorry, I missed your previous pushes. I have few quenstion's
> 
> I think the ignite boradcast method itself checks for types in `ignite/distributed/comp_models/base.py` and will throw error for sequence and mapping even after this check in distributed setting.

Hi @aaishwarymishra, you're right that broadcast() in base.py rejects Sequence/Mapping. The status-code exists because only rank 0 runs compute_fn — without it, rank 0 would hit that TypeError and exit while other ranks deadlock waiting in broadcast's safe_mode path.

That said, I think the better direction is to support container outputs in distributed mode by recursively broadcasting each leaf tensor individually. Would that be welcome?

aaishwarymishra:
@vfdev-5 what's your thoughts on it?

zongyang078:
Implemented the recursive broadcast approach I proposed above, so container (`tuple`/`list`/`Mapping`) `compute_fn` outputs — including nested ones — now work in distributed mode instead of raising `NotImplementedError`. Type/key validation happens inside the broadcast protocol itself so every rank raises together on invalid output, rather than rank 0 exiting early while other ranks hang on a collective call.
Added tests for nested container broadcast and for the symmetric-raise behavior on invalid output/keys, verified against a real `WORLD_SIZE=2` gloo run.

@aaishwarymishra @vfdev-5 would appreciate a look when you have time.

vfdev-5:
@zongyang078 thanks for the updates! 

I checked the new code and have an alternative idea. It would require to land first another PR and then update this one.

As the minimal pytorch version that ignite supports today is 2.2 (we'll raise it to 2.6 soon), we can use https://docs.pytorch.org/docs/2.2/distributed.html#torch.distributed.broadcast_object_list which can broadcast picklable objects, so we can simplify our code a lot. 

In this case, we can first introduce `_do_broadcast_object_list` to `ComputationModel` and its derived classes. Then we can use it similarly to `_do_all_gather_object`: https://github.com/pytorch/ignite/blob/7358a27ff9701400465b699e1eb2abb3a9d1b49d/ignite/distributed/comp_models/base.py#L232 . In the derived classes of ComputationModel, if the backend does not support broadcast objects we can just raise NotImplementedError.

In this case can update the typehint of `idist.broadcast` and add `Any` like in `idist.all_gather` and also update the docstring of both methods to reflect the fact that we can broadcast and all_gather picklable any objects: https://github.com/pytorch/ignite/blob/7358a27ff9701400465b699e1eb2abb3a9d1b49d/ignite/distributed/utils.py#L410 

Once idist.broadcast can work with any picklable objects we can update this PR and simplify the broadcasting part.

zongyang078:
Thanks @vfdev-5, that's cleaner — I'll open that PR.

Just to confirm I read your sketch right: `_do_broadcast_object_list` on `ComputationModel`, dispatched from `ComputationModel.broadcast` the same way `all_gather` dispatches to `_do_all_gather_object`; `NotImplementedError` on backends without object support; plus `Any` in the `idist.broadcast` typehint and docstring updates on both methods.

I'll link it here once it's up, along with one design question that came out of it — tensors nested inside a broadcast object keep the sending process' device.

PR #3790, opened 2026-06-22
fix: prevent once handlers from refiring on epoch termination

### LINKED ISSUE #3625: Bug: Events.GET_BATCH_STARTED(once=N) fires twice when engine.terminate_epoch(False) is called inside handler
When using Events.GET_BATCH_STARTED with the once=N filter, the handler ends up triggering twice if engine.terminate_epoch(False) is called inside that same handler.

I noticed there's already an inline TODO about this in tests/ignite/engine/test_engine.py (around line 342), plus some comments in ignite/engine/engine.py about trying to avoid double triggers when the dataloader restarts.

Here is a minimal script to reproduce it

```
from unittest.mock import MagicMock

engine = Engine(MagicMock(return_value=1))
count = {"n": 0}

@engine.on(Events.GET_BATCH_STARTED(once=12))
def h():
    count["n"] += 1
    if count["n"] == 1:
        engine.terminate_epoch(False)

engine.run(range(10), max_epochs=3)
print("GET_BATCH_STARTED_once_12_calls", count["n"])

```

Actual behavior:
The script prints GET_BATCH_STARTED_once_12_calls 2.

Expected behavior:
Because of once=12, the handler should strictly fire exactly once.

If this is something you want fixed right now, let me know! I'd be happy to open a draft PR with a failing test and a fix for the engine event flow.

Environment
PyTorch Version: 2.10.0+cpu
Python version: 3.13.5
Ignite version: 0.6.0

nightcityblade:
## Summary

Fixes a case where `Events.GET_BATCH_STARTED(once=N)` could fire twice if `engine.terminate_epoch(False)` was called from that handler. `once` filters now consume matched values and each registered handler gets an independent once filter instance.

## Changes

- Make `once` event filters stateful so matched values are only emitted once.
- Clone once filters per handler registration to avoid sharing consumed state between handlers.
- Add a regression assertion to the epoch-termination event sequence test.

## Testing

- `python3 -m pytest tests/ignite/engine/test_engine.py::TestEngine::test_terminate_epoch_events_sequence -q`
- `python3 -m pytest tests/ignite/engine/test_custom_events.py::test_once_event_filter tests/ignite/engine/test_custom_events.py::test_custom_events_with_events_list -q`
- `python3 -m pytest tests/ignite/engine/test_engine.py tests/ignite/engine/test_custom_events.py -q -m 'not distributed'`

Fixes pytorch/ignite#3625

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -325,6 +325,9 @@ class Engine(Serializable):
             return RemovableEventHandle(event_name, handler, self)
         if isinstance(event_name, CallableEventWithFilter) and event_name.filter is not None:
             event_filter = event_name.filter
+            once = getattr(event_filter, "_once", None)
+            if once is not None:
+                event_filter = Events.once_event_filter(list(once))
             handler = self._handler_wrapper(handler, event_name, event_filter)
 
         self._assert_allowed_event(event_name)
--- a/ignite/engine/events.py
+++ b/ignite/engine/events.py
@@ -140,12 +140,15 @@ class CallableEventWithFilter:
     @staticmethod
     def once_event_filter(once: list) -> Callable:
         """A wrapper for once event filter."""
+        remaining = set(once)
 
         def wrapper(engine: "Engine", event: int) -> bool:
-            if event in once:
+            if event in remaining:
+                remaining.remove(event)
                 return True
             return False
 
+        wrapper._once = tuple(once)  # type: ignore[attr-defined]
         return wrapper
 
     @staticmethod
--- a/tests/ignite/engine/test_engine.py
+++ b/tests/ignite/engine/test_engine.py
@@ -395,6 +395,7 @@ class TestEngine:
 
             epoch_completed_events = [e for e in engine.called_events if e[2] == Events.EPOCH_COMPLETED.name]
             assert len(epoch_completed_events) == max_epochs - skip_epoch_completed
+            assert call_count == 1
 
     @pytest.mark.parametrize("data", [None, "mock_data_loader"])
     def test_iteration_events_are_fired(self, data):


siddhant-shahhh (commented):
This fixes a bug the repo already tracks. `tests/ignite/engine/test_engine.py` has a TODO saying `GET_BATCH_STARTED(once=12)` is called twice, with an `if call_count < 1:` guard working around it. This PR adds the assertion that TODO was waiting for.

The test catches it. On this branch: 24 passed. With `engine.py` and `events.py` reverted to master: 8 failed, all `assert 2 == 1`, only the `GET_BATCH_STARTED/COMPLETED(once=12)` cases. That matches the cause: after the epoch is terminated, the batch fetch retries at the same `state.iteration`, so those two events fire twice.

Rebuilding the filter per handler in `add_event_handler` makes sense, otherwise handlers sharing one event object would share one `remaining` set.

One question. Filter state is now per-handler and never reset, so a `once` handler doesn't fire on a second `engine.run()`:

```
this branch          master
after run 1: [1]     after run 1: [1]
after run 2: [1]     after run 2: [1, 1]
```

Could be the right meaning for `once`, but it's a behaviour change that isn't in the description or tested. Worth deciding on purpose and adding a test.

Also, the TODO and the `call_count` guard can both go in this PR now.

`ruff check` and `ruff format` pass on all three files.

nightcityblade:
Thanks for catching that. I updated the implementation so each per-handler `once` filter resets when `Engine.run()` creates a new state, while repeated firing at the same event value within a run remains suppressed. I also added the two-run regression assertion and removed the old TODO/guard.\n\nValidation: 36 focused tests passed; Ruff check and format checks passed on all four changed files; `git diff --check` passed. The wider two-file engine test run had 242 passes, 11 skips, 4 mock-introspection failures that the follow-up fixed, and 3 distributed-test setup errors due to the unavailable `worker_id` fixture.

PR #3791, opened 2026-06-22
docs: clarify FID default feature dimension

### LINKED ISSUE #3221: Calculation of FID metric made use of classification probabilities instead of feature vectors
Based on my understanding, the calculation of fid should make use of feature vector(dim=2048) from the max pool layers of InceptionNet. 

However, the `FID()` metric in `fid.py` is initiliased as below:
```
if num_features is None and feature_extractor is None:
    num_features = 1000
    feature_extractor = InceptionModel(return_features=False, device=device)
```
The `InceptionModel` return prediction probabilities(dim=1000) if `return_features` is set to False by specification. `update()` then make use of this this  `feature_extractor` to get the probabilities instead of feature vectors. 

I think `feature_extractor` should be `InceptionModel(return_features=True, device=device)` instead?

nightcityblade:
Fixes #3221

Description:

Clarifies that the default `FID` configuration uses the Inception model's 1000-dimensional output logits, and points users to the 2048-dimensional pool3 feature extractor example when matching `pytorch_fid` results.

Check list:

### DIFF
--- a/ignite/metrics/gan/fid.py
+++ b/ignite/metrics/gan/fid.py
@@ -83,6 +83,9 @@ class FID(_BaseInceptionMetric):
     Args:
         num_features: number of features predicted by the model or the reduced feature vector of the image.
             Default value is 1000.
+            The default Inception model uses its 1000-dimensional output logits as features. To match
+            `pytorch_fid` results, provide a feature extractor that returns the 2048-dimensional pool3
+            features instead; see the example below.
         feature_extractor: a torch Module for extracting the features from the input data.
             It returns a tensor of shape (batch_size, num_features).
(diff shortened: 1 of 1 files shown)


ignite/metrics/gan/fid.py
```diff
@@ -83,6 +83,9 @@ class FID(_BaseInceptionMetric):
     Args:
         num_features: number of features predicted by the model or the reduced feature vector of the image.
             Default value is 1000.
+            The default Inception model uses its 1000-dimensional output logits as features. To match
+            `pytorch_fid` results, provide a feature extractor that returns the 2048-dimensional pool3
+            features instead; see the example below.
```
vfdev-5, line 88:
```suggestion
            features instead. See the example below.
```

Issue #466, opened 2019-03-31
Conda install ignite 


RoyHirsch:
Hi,
I'm trying to download view pip, I've downloaded the current version of 0.1.2 but don't have the current master. For example I don't have the is_multiclass flag. How can I make sure I download the updated version in the repo ? 
I've used : pip install pytorch-ignite

Thanks,
Roy

vfdev-5:
Hi Roy, you can install the master using the following command:
```
pip install --upgrade git+https://github.com/pytorch/ignite
```

If you are using pytorch-nightly, we have ignite nightly builds with
pytorch-nightly as dependency:
```
pip install pytorch-ignite-nightly
```

HTH

RoyHirsch:
Thank you. I would like to use the Recall/Persicion measurement for a multilabel task. But I would like to compute thre recall per class. while using th is_multilabel flag I have two options:
- is avg is True: average recall for all the samples
- is avg is Fasle: reports recall per sample.

Is there a support in reporting recall per class ?
Thank you,
Roy

vfdev-5:
@RoyHirsch could you please detail which task would you like to do:
- multiclass = target is [0, 1, 2, ..., C]
- multilabel = target are "tags", e.g. [[0, 1, 0], [0, 1, 1], ...]

In case of multiclass, the argument `avg=False` gives Precision/Recall per class, otherwise they are averaged over all classes.

In case of multilabel, AFAIK and remember we can not compute metrics per class but per sample. Averaging is thus going per samples. Take a look for example, https://scikit-learn.org/stable/modules/generated/sklearn.metrics.recall_score.html#sklearn.metrics.recall_score with `average="samples"`.

RoyHirsch:
My task is multilabel, means the ground true is binary vector with arbitrary True values: [[0, 1, 0], [0, 1, 1], ...]

My prediction is a scores vector in shape NUM_CLASSES, I am processing the vector into topK predictions per sample (we can assume it is processeed into bianry vector, where ony the topK columns are True and the rest is False).
I would like to report the per class recall following:

![image](https://user-images.githubusercontent.com/26789321/55289334-1445d500-53ce-11e9-9c41-045c96128e75.png)
*DeepConvolutionalRankingfor MultilabelImageAnnotation

Maybe I am missing something out , but i think this is a relevant metric.
Roy

vfdev-5:
[...]

PR #3792, opened 2026-06-24
feat: Add top-k multilabel precision and recall metrics

### LINKED ISSUE #467: Top-K precision/recall multilabel metrics for ranking task

abdelmagid07:
Fixes #467 

## Description

Adds two new multilabel ranking metrics: `TopKMultilabelPrecision` and `TopKMultilabelRecall` to `ignite.metrics`.

For each sample, the `k` labels with the highest prediction scores are treated as positive predictions; precision and recall are then reported **per label** across the accumulated samples. This matches the metric described in the discussion (#466 -> #467 ), grounded in https://arxiv.org/pdf/1312.4894.pdf

## Notes:

-`y_pred` must be **float scores**.
- If `k` exceeds the number of labels, it is clamped to `num_labels`.
- Default k = 5

## Changes

- New module `ignite/metrics/top_k_multilabel_precision_recall.py`
- Export `TopKMultilabelPrecision` and `TopKMultilabelRecall` from `ignite.metrics`
- Add entries to `docs/source/metrics.rst`
- RST docstrings with `.. testcode::` / `.. testoutput::` examples
- Tests at `ignite/metrics/test_top_k_multilabel_precision_recall.py`

## Check list:

### DIFF
--- a/ignite/metrics/top_k_multilabel_precision_recall.py
+++ b/ignite/metrics/top_k_multilabel_precision_recall.py
@@ -11,7 +11,7 @@ __all__ = ["TopKMultilabelPrecision", "TopKMultilabelRecall"]
 class _BaseTopKMultilabelPrecisionRecall(_BasePrecisionRecall):
     """Base class for top-k multilabel precision and recall metrics.
 
-    For each sample, the ``k`` labels with the highest prediction scores are
+    For each sample, the k labels with the highest prediction scores are
     selected as positive predictions. Precision and recall are then computed per
     label, treating each label independently across the accumulated samples.
     """
@@ -37,14 +37,7 @@ class _BaseTopKMultilabelPrecisionRecall(_BasePrecisionRecall):
         )
 
     def _check_type(self, output: Sequence[torch.Tensor]) -> None:
-        self._check_shape(output)
         y_pred, y = output
-
-        if not (y.shape == y_pred.shape and y.ndimension() > 1 and y.shape[1] > 1):
-            raise ValueError(
-                "y and y_pred must have same shape of (batch_size, num_labels, ...) "
-                f"with num_labels > 1, but given {y.shape} vs {y_pred.shape}."
-            )
 
         if not torch.equal(y, y**2):
             raise ValueError("For multilabel cases, y must be comprised of 0's and 1's.")
@@ -59,9 +52,7 @@ class _BaseTopKMultilabelPrecisionRecall(_BasePrecisionRecall):
         elif self._type != "multilabel":
             raise RuntimeError(f"Input data type has changed from {self._type} to multilabel.")
         elif self._num_classes != num_labels:
-            raise ValueError(
-                f"Input data number of labels has changed from {self._num_classes} to {num_labels}."
-            )
+            raise ValueError(f"Input data number of labels has changed from {self._num_classes} to {num_labels}.")
 
     def _prepare_output(self, output: Sequence[torch.Tensor]) -> Sequence[torch.Tensor]:
         y_pred, y = output[0].detach(), output[1].detach()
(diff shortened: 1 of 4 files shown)


ignite/metrics/top_k_multilabel_precision_recall.py
```diff
@@ -0,0 +1,377 @@
...
+        self._k = k
+        super().__init__(
+            output_transform=output_transform,
+            average=average,
+            is_multilabel=True,
+            device=device,
+            skip_unrolling=skip_unrolling,
+        )
+
+    def _check_type(self, output: Sequence[torch.Tensor]) -> None:
```
aaishwarymishra, line 39: The parent class implements this method, do we need to implement it?

ignite/metrics/top_k_multilabel_precision_recall.py
```diff
@@ -0,0 +1,377 @@
...
+        k = min(self._k, num_labels)
+        topk_indices = torch.topk(y_pred, k, dim=1).indices
+        y_pred_topk = torch.zeros_like(y_pred)
+        y_pred_topk.scatter_(1, topk_indices, 1.0)
+
+        y_pred_topk = y_pred_topk.to(dtype=self._double_dtype, device=self._device)
+        y = y.to(dtype=self._double_dtype, device=self._device)
+        correct = y * y_pred_topk
+
+        return y_pred_topk, y, correct
```
aaishwarymishra, line 76: I think other classification metrics like `Accuracy` follow `(N,S,C)` batch, sequence, classes I think this metric should be consistent with other metrics.

ignite/metrics/top_k_multilabel_precision_recall.py
aaishwarymishra, line 1: Isn't essentially both classes doing the same thing why we need 2 separate classes?

ignite/metrics/top_k_multilabel_precision_recall.py
abdelmagid07: I think we do because the parent's method requires y_pred to be binary. These metrics take raw prediction scores in float form, such as 0.8, 0.6, etc., and select the top-k values, so the original check can't be reused; the parent method would end up rejecting valid inputs. The override validates float y_pred instead, while keeping the same other checks. I can add comments in the code addressing this.

ignite/metrics/top_k_multilabel_precision_recall.py
abdelmagid07: I aimed to mirror the existing Precision and Recall, which use the (N, C, ...) shape,
Like in precision.py (line 78 - 82):

        elif self._type == "multilabel":
            # if y, y_pred shape is (N, C, ...) -> (N * ..., C)
            num_labels = y_pred.size(1)
            y_pred = torch.transpose(y_pred, 1, -1).reshape(-1, num_labels)
            y = torch.transpose(y, 1, -1).reshape(-1, num_labels)

I see the (N, S, C) shape in get_sequence_transform in accuracy.py, but this seems like a different case than multilabel scoring. The tests cover (N, C), (N, C, L), and (N, C, H, W) inputs and pass, so the extra dimensions are handled the same way as the other multilabel metrics. Let me know your thoughts here.

ignite/metrics/top_k_multilabel_precision_recall.py
abdelmagid07: My goal there was to mirror the existing API as much as possible. From my understanding, we already ship Precision and Recall as two separate classes over the shared parent _BasePrecisionRecall, with update methods that differ in the denominator they use (predicted vs actual positives). I tried to mirror that by doing the same over _BaseTopKMultilabelPrecisionRecall. Essentially, we would have TopKMultilabelPrecision/TopKMultilabelRecall reading parallel to the existing Precision/Recall split. Though I'm happy to merge them into a single class if that fits the codebase conventions better.

aaishwarymishra:
@abdelmagid07 To be honest, looking at the code I am not sure how these metrics work, the compute of the precision and recall will differ right, but there is no compute implementation, wont the results of both metrics be same when we do .compute() , even when we have 2 completely different metrics precision and recall.

vfdev-5:
@abdelmagid07 @aaishwarymishra I would say we would better to go with https://github.com/pytorch/ignite/issues/3568 approach instead of creating TopK version for each metric. Let's close this PR as a no-go.

PR #3804, opened 2026-07-05
Change the upper bound to <3.15 for the required python version to enable support for 3.14.x

### LINKED ISSUE #3803: Unable to install ignite's `0.5.3`, `0.5.4`, and the latest pre-release version via pip for Python 3.14.x
## 🐛 Unable to install pytorch-ignite's `0.5.3`, `0.5.4`, and the latest pre-release version via pip 

I've noticed that python updates from 3.14.0 to 3.14.x downgrade ignite version. Also, I can't install the nightly build for 3.14.x. See "Steps to reproduce"

I believe this is cause of this line from `pyproject.toml`:

https://github.com/pytorch/ignite/blob/d5208644ab5646ee250c3b5db08f5c5e9f128354/pyproject.toml#L17

The catch is in pyproject algebra, 3.14.x > 3.14.0.

## Possible solution

The following change has solved the issue for a local clone:

```diff
- requires-python = ">=3.10,<=3.14"
+ requires-python = ">=3.10,<3.15"
```

then versions with 3.14.x fall into the range.

## How to reproduce

```sh
cd /tmp
python3.14 -m venv .venv
source .venv/bin/activate
python --version
# Python 3.14.6
```

Attempting to install pytorch-ignite via pip:

```sh
pip install pytorch-ignite==0.5.4
```

ERROR:

```sh
ERROR: Ignored the following versions that require a different python version: 0.5.3 Requires-Python >=3.9,<=3.13; 0.5.4 Requires-Python >=3.10,<=3.14;
...
ERROR: No matching distribution found for pytorch-ignite==0.5.4
```
Same for 0.5.3.

**Attempting to install nightly build**

```sh
pip install --pre pytorch-ignite
pip show pytorch-ignite
```

Output:

```sh
Name: pytorch-ignite
Version: 0.6.0.dev20250617
...
```
**NOTE** current pre-release 0.6.0.dev20260705

For `0.5.2` everything works fine.

## Environment

- PyTorch Version (e.g., 1.4): None
- Ignite Version (e.g., 0.3.0): 0.5.4, also 0.5.3
- OS (e.g., Linux): 24.04.1-Ubuntu
- Python version: Python 3.14.6

tarasio-mirror:
Fixes #3803 

Description:

Changes the upper bound to <3.15 for the required python version to enable support for 3.14.x. See #3803 for description.

Check list:

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -187,5 +187,8 @@ jobs:
       - name: Run CIFAR10 with Ray Tune and Ignite
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: |
+          echo "Downloading CIFAR20 archive from Zenodo..."
+          mkdir -p data
+          curl -L --retry 3 --retry-delay 10 -o data/cifar-10-python.tar.gz "https://zenodo.org/records/10089977/files/cifar-10-python.tar.gz?download=1"
           uv pip install -r examples/cifar10_raytune/requirements.txt
           python examples/cifar10_raytune/main.py --num_epochs 1 --num_trials 1 --cpus_per_trial 1 --gpus_per_trial 0
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -187,7 +187,7 @@ jobs:
       - name: Run CIFAR10 with Ray Tune and Ignite
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: |
-          echo "Downloading CIFAR20 archive from Zenodo..."
+          echo "Downloading CIFAR10 archive from Zenodo..."
           mkdir -p data
           curl -L --retry 3 --retry-delay 10 -o data/cifar-10-python.tar.gz "https://zenodo.org/records/10089977/files/cifar-10-python.tar.gz?download=1"
           uv pip install -r examples/cifar10_raytune/requirements.txt
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -43,7 +43,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
         pytorch-channel: [pytorch, pytorch-nightly]
         include:
           # includes a single build on windows
--- a/examples/super_resolution/main.py
+++ b/examples/super_resolution/main.py
@@ -147,7 +147,9 @@ basic_profiler.attach(trainer)
 
 ProgressBar().attach(trainer, output_transform=lambda x: {"loss": x})
 
-trainer.run(training_data_loader, opt.n_epochs, epoch_length=epoch_length)
 
-results = basic_profiler.get_results()
-basic_profiler.print_results(results)
+if __name__ == "__main__":
+    trainer.run(training_data_loader, opt.n_epochs, epoch_length=epoch_length)
+
+    results = basic_profiler.get_results()
+    basic_profiler.print_results(results)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ license-files = ["LICENSE"]
 classifiers = [
     "Programming Language :: Python :: 3",
 ]
-requires-python = ">=3.10,<=3.14"
+requires-python = ">=3.10,<3.15"
 dependencies = [
     "torch>=2.2,<3",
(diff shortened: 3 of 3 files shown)


vfdev-5 (approved):
Thanks, Taras!

vfdev-5:
@tarasio-mirror can you also add 3.14 to unit-tests matrix to run tests on it?

tarasio-mirror:
Sure! I'll do that.

tarasio-mirror:
Looks like it fails at `Run SR Example` step,
same for my fork

https://github.com/tarasio-mirror/ignite/actions/runs/28739172288/job/85218743938

I'll have a look

tarasio-mirror:
Since, Python3.14 has changed the default multiprocessing start method from `fork` to `forkserver` for Linux (not sure about MacOS)

More: https://github.com/pytorch/pytorch/issues/169252

to enable example steps for unit tests for Python3.14, I exposed multiprocessing_context arg of Dataloader, and explicitly set it to "fork" (the old one).

vfdev-5:
@tarasio-mirror thanks for exploring this! Can we do a different thing: add automatically the multiprocessing_context for python >= 3.14 in auto_dataloader? And change in SR example DataLoader with idist.auto_dataloader...

vfdev-5:
Actually, let's take your fix for now and open an issue about : 
> add automatically the multiprocessing_context for python >= 3.14 in auto_dataloader

Actually, I'm not totally sure if we should always pass mp_context as fork. Maybe, pytorch folks will make it work with forkserver some day

tarasio-mirror:
SR example now passes, but CIFAR raytune got stuck on dataset download stage

tarasio-mirror:
> Actually, let's take your fix for now and open an issue about :
> 
> > add automatically the multiprocessing_context for python >= 3.14 in auto_dataloader
> 
> Actually, I'm not totally sure if we should always pass mp_context as fork. Maybe, pytorch folks will make it work with forkserver some day

yeah

I've tried multiple configurations, and unfortunately, the only working solution for Python3.14, for now, is explicitly passing `multiprocessing_context="fork"` when using with `num_workers` > 0:

```python
DataLoader(..., num_workers=4, multiprocessing_context="fork")
```

vfdev-5:
@tarasio-mirror I guess @aaishwarymishra is right in his comment here: https://github.com/pytorch/ignite/issues/3805#issuecomment-4889852090 that pytorch should have fixed that from their side (partially): https://docs.pytorch.org/docs/2.12/data.html#platform-specific-behaviors
I think we are missing `if __name__ == '__main__':` in SR example to make it work without specifying mp_context. Can you try that instead of passing mp_context arg? Thanks!

tarasio-mirror:
I'll check! it's a good point

tarasio-mirror:
Ok,  Looks like when using inside `if __name__ == '__main__'` no need to pass context manually:

```sh
cd /tmp
python3.14 -m venv .venv
source .venv/bin/activate
python --version
# Python 3.14.6
pip show torch
# Name: torch
# Version: 2.12.1
# ...
# generate test.
echo "import torch" >> test.py
echo "dl = torch.utils.data.DataLoader(torch.randn(1024), batch_size=64, num_workers=2)" >> test.py
echo "if __name__ == '__main__': [_ for _ in dl]" >> test.py
python test.py
# 0
```

This fails:

```sh
echo "import torch" >> test_fail.py
echo "dl = torch.utils.data.DataLoader(torch.randn(1024), batch_size=64, num_workers=2)" >> test_fail.py
echo "[_ for _ in dl]" >> test_fail.py
python test_fail.py
# ConnectionResetError: [Errno 104] Connection reset by peer
```

tarasio-mirror:
I'll do cleanup then.

tarasio-mirror:
SR works fine now

CIFAR dataset (cifar raytune step) is unreachable, it gives ETA = 4 hours

vfdev-5:
Can we use other source for CIFAR10?

tarasio-mirror:
I'll have a look!

we can always host the mirror by ourself, like we are doing for MNIST :)

tarasio-mirror:
Found one on Zenodo
https://doi.org/10.5281/zenodo.10089977

.github/workflows/unit-tests.yml
```diff
@@ -187,5 +187,8 @@ jobs:
       - name: Run CIFAR10 with Ray Tune and Ignite
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: |
+          echo "Downloading CIFAR20 archive from Zenodo..."
```
vfdev-5:
```suggestion
          echo "Downloading CIFAR10 archive from Zenodo..."
```

PR #3807, opened 2026-07-11
Fix ClearML task closure issue

aaishwarymishra:
This pull request adds a small but important resource management improvement to the `ignite/handlers/clearml_logger.py` file. Now, when the `close()` method is called, it also closes the underlying `_task` if it exists, ensuring resources are properly released.

Resource management improvement:

* Updated the `close()` method to call `self._task.close()` if `_task` is set, preventing potential resource leaks.Fixes #{issue number}

Description:

Check list:

### DIFF
--- a/ignite/handlers/clearml_logger.py
+++ b/ignite/handlers/clearml_logger.py
@@ -218,6 +218,8 @@ class ClearMLLogger(BaseLogger):
 
     def close(self) -> None:
         self.clearml_logger.flush()
+        if self._task:
+            self._task.close()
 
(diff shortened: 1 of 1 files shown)


aaishwarymishra:
Also I am not sure if we should skip creating a new task, like currently clearml logger doesn't create a new task of a task is already present, maybe either we should create a new one and close the previous one, pr create an other one even if there is already one present.

That's the accepted behaviour in my opinion.

vfdev-5 (approved):
Thanks!

PR #3808, opened 2026-07-13
Fix MetricGroup requiring (y_pred, y) keys on mapping outputs

### LINKED ISSUE #3806: Attaching every loss separately is different from attaching them with MetricGroup
## 🐛 Bug description

I'm trying to get different loss metrics for the train set and validation set.
Basically, I have multiple outputs, and each output's loss is computed differently.
Here's some example code for the train loop and validation loop :

```py3
    model.train()
    
    optimizer.zero_grad()
    ...
    loss.backward()

    optimizer.step()
    return loss.item()


def validation_step(engine, batch):
    model.eval()
    
    with torch.no_grad():
        ...

    return {
        "outputs_1": (tensor_1, tensor_2),
        "masks": (mask, gt_mask)
    }
    
```
Trainer and train validator :
```py3
trainer = Engine(train_step)
train_validator = Engine(validation_step)
```
Every epoch, validation is run on the training set to get some metrics :
```py3
def log_metrics(engine):

    train_validator.run(tr_dataloader)
    metrics = train_validator.state.metrics
    
    print(metrics)
```
<br><br>

Attaching the loss directly to the validator works like a charm :
```py3
loss = Loss(MSE, output_transform=lambda o: o["outputs_1"])
loss.attach(train_validator, "loss")
```
Output :
```
Epoch [1/50]: [1/1] 100%|█████████████████████████████████████████████████████████████████ [00:00<?]
{'loss': 0.34148311614990234} 
```
<br>
But with `MetricGroup`, something weird happens when passing the output to the different losses :

```py3
train_metrics = MetricGroup({"loss": loss})
train_metrics.attach(train_validator, "train_metrics")
```

Output :
```
Epoch [1/50]: [1/1] 100%|█████████████████████████████████████████████████████████████████ [00:00<?]
ERROR:ignite.engine.engine.Engine:Current run is terminating due to exception: When transformed engine's output is a mapping, it should contain ('y_pred', 'y') keys, but given ['outputs_1', 'masks']
```
<br>

Eventually, I did some research and found out that usually, output should contain "y_pred" and "y" keys, and if we're in the case of multiple predicitons `skip_unrolling` should be set to ```True```.
I also believe it has something to do with ```required_output_keys```.

Additionnal testing :
```py3
def foo(output, text=""):
    print(text, type(output))

    return output


loss = Loss(MSE, skip_unrolling=True, output_transform=lambda o: foo(o, "inside loss"))


train_metrics = MetricGroup({"loss": loss}, output_transform=lambda o: foo(o, "inside metric group"))
train_metrics.required_output_keys = ("outputs_1", "masks")
train_metrics.attach(train_validator, "train_metrics")
```
```
Epoch [1/50]: [1/1] 100%|█████████████████████████████████████████████████████████████████ [00:00<?]
inside metric group <class 'dict'>
inside loss <class 'tuple'>
ERROR:ignite.engine.engine.Engine:Current run is terminating due to exception: ...
```

I expected the output dict to be unchanged when passed to `MetricGroup` then to the different losses.
Instead, output is passed as a tuple when passed individually to the losses.

## Environment

- PyTorch Version (e.g., 1.4): **2.11.0+cu128**
- Ignite Version (e.g., 0.3.0): **0.5.4**
- OS (e.g., Linux): **Linux**
- How you installed Ignite (`conda`, `pip`, source): **uv**
- Python version: **3.14.3**
- Any other relevant information: **CUDA 12.8**

ShreyanshGoyal:
## Summary

Fixes #3806.

`MetricGroup.update()` already applies each child metric's own `output_transform` to pull what it needs out of the group's output, per the documented contract:

> `output_transform` of each metric in the group is also called upon its update.

But `MetricGroup` inherited `Metric.iteration_completed` unchanged, which enforces `required_output_keys == ("y_pred", "y")` whenever the (group-transformed) engine output is a mapping — and unconditionally raises if the group's own `required_output_keys` is `None`. That check only makes sense for a single "leaf" metric consuming `(y_pred, y)` directly; a `MetricGroup` wrapping heterogeneous child metrics with their own `output_transform`s has no business requiring specific keys on the raw output.

As a result, any engine returning a dict with keys other than `'y_pred'`/`'y'` (e.g. a multi-output model returning `{"outputs_1": ..., "masks": ...}`) broke as soon as it was wrapped in a `MetricGroup`, even though attaching the same child metric directly worked fine:

```python
loss = Loss(MSE, output_transform=lambda o: o["outputs_1"])
loss.attach(train_validator, "loss")  # works

train_metrics = MetricGroup({"loss": loss})
train_metrics.attach(train_validator, "train_metrics")  # ValueError: ... should contain ('y_pred', 'y') keys, but given ['outputs_1', 'masks']
```

Manually setting `required_output_keys` on the group didn't help either — it triggered `MetricGroup`'s inherited multi-output "unrolling" logic, which reinterpreted the mapping as a `(y_pred, y)`-shaped tuple and unrolled it into mismatched per-metric updates (each child's own `output_transform` then received a plain tensor pair instead of the original mapping).

## Fix

`MetricGroup.iteration_completed` is now overridden to pass a mapping output straight through to `update()` without the `required_output_keys` check, while preserving the existing multi-output unrolling behavior (`skip_unrolling`) for non-mapping outputs — this is unchanged and still covered by the distributed integration test.

## Test plan

- [x] Added `test_mapping_output_with_custom_keys` to `tests/ignite/metrics/test_metric_group.py`, which attaches the same `Loss` metric directly and via a `MetricGroup` on an engine returning a dict with custom keys, and asserts both produce identical results. Confirmed this test fails with the original `ValueError` on the pre-fix code and passes after the fix.
- [x] `pytest tests/ignite/metrics/test_metric_group.py -v` — all non-distributed tests pass, plus the `gloo_cpu` distributed integration test (verifying the group-level unrolling path is untouched)
- [x] `pytest tests/ignite/metrics/` (full suite, excluding modules needing optional deps I didn't have installed) — 4882 passed, 1 pre-existing unrelated failure (missing `pynvml`, unrelated to this change)
- [x] `mypy ignite/metrics/metric_group.py` and `ruff check`/`ruff format --check` — clean

### DIFF
--- a/tests/ignite/metrics/test_metric_group.py
+++ b/tests/ignite/metrics/test_metric_group.py
@@ -3,7 +3,7 @@ import torch
 
 from ignite import distributed as idist
 from ignite.engine import Engine
-from ignite.metrics import Accuracy, MetricGroup, Precision
+from ignite.metrics import Accuracy, Loss, MetricGroup, Precision
 
 torch.manual_seed(41)
 
@@ -46,6 +46,34 @@ def test_output_transform():
 
     assert precision.state_dict() == group.metrics["precision"].state_dict()
     assert accuracy.state_dict() == group.metrics["accuracy"].state_dict()
+
+
+def test_mapping_output_with_custom_keys():
+    # Regression test for https://github.com/pytorch/ignite/issues/3806 :
+    # a MetricGroup should not require the engine's output mapping to contain
+    # ('y_pred', 'y'); each metric in the group applies its own output_transform
+    # to pull what it needs from the mapping, same as attaching it directly.
+    def step(engine, batch):
+        return {
+            "outputs_1": (torch.rand(4, 3), torch.rand(4, 3)),
+            "masks": (torch.rand(4, 3), torch.rand(4, 3)),
+        }
+
+    loss_fn = torch.nn.MSELoss()
+
+    direct_engine = Engine(step)
+    Loss(loss_fn, output_transform=lambda o: o["outputs_1"]).attach(direct_engine, "loss")
+
+    group_engine = Engine(step)
+    group = MetricGroup({"loss": Loss(loss_fn, output_transform=lambda o: o["outputs_1"])})
+    group.attach(group_engine, "metrics")
+
+    torch.manual_seed(0)
+    direct_engine.run([0])
+    torch.manual_seed(0)
+    group_engine.run([0])
+
+    assert group_engine.state.metrics["metrics"] == {"loss": direct_engine.state.metrics["loss"]}
 
 
 def test_compute():
--- a/tests/ignite/metrics/test_metric_group.py
+++ b/tests/ignite/metrics/test_metric_group.py
@@ -62,10 +62,12 @@ def test_mapping_output_with_custom_keys():
     loss_fn = torch.nn.MSELoss()
 
     direct_engine = Engine(step)
-    Loss(loss_fn, output_transform=lambda o: o["outputs_1"]).attach(direct_engine, "loss")
+    direct_loss = Loss(loss_fn, output_transform=lambda o: o["outputs_1"])
+    direct_loss.attach(direct_engine, "loss")
 
     group_engine = Engine(step)
-    group = MetricGroup({"loss": Loss(loss_fn, output_transform=lambda o: o["outputs_1"])})
+    group_loss = Loss(loss_fn, output_transform=lambda o: o["outputs_1"])
+    group = MetricGroup({"loss": group_loss})
     group.attach(group_engine, "metrics")
 
     torch.manual_seed(0)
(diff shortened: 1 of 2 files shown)


tests/ignite/metrics/test_metric_group.py
```diff
@@ -48,6 +48,34 @@ def drop_first(output):
...
+    def step(engine, batch):
+        return {
+            "outputs_1": (torch.rand(4, 3), torch.rand(4, 3)),
+            "masks": (torch.rand(4, 3), torch.rand(4, 3)),
+        }
+
+    loss_fn = torch.nn.MSELoss()
+
+    direct_engine = Engine(step)
+    Loss(loss_fn, output_transform=lambda o: o["outputs_1"]).attach(direct_engine, "loss")
```
aaishwarymishra: to be consistent either don't save metric group, or save both metric object

ShreyanshGoyal:
Thanks — addressed in 18e8dcc9. The regression test now keeps both the directly attached and grouped `Loss` instances and compares their state dictionaries as well as their computed outputs. The full MetricGroup test file passes (5 passed, 5 skipped for unavailable distributed backends).

PR #3810, opened 2026-07-16
feat: Add per_label parameter to Accuracy for per-label multilabel output

### LINKED ISSUE #513: Label-wise metrics (Accuracy etc.) for multi-label problems

zongyang078:
Fixes #513

Description:
Adds a `per_label` parameter to `Accuracy` for multilabel classification. When `per_label=True` and `is_multilabel=True`, `compute()` returns a per-label accuracy tensor of shape `(num_labels,)` using elementwise comparison, instead of the default subset accuracy scalar. Default behavior (`per_label=False`) is unchanged.

Prior attempts (PR #516, PR #542) stalled in 2019 over API design — both proposed a `labelwise=True` boolean flag, which was rejected. This PR went through a few naming iterations during review (`average` → `subset_accuracy` → `per_label`) before landing on `per_label`, since it avoids clashing with `Precision`/`Recall`'s existing `average='micro'/'macro'` semantics and its True/False polarity matches what it returns.

`per_label` is the last keyword argument in `__init__` to preserve positional-argument backward compatibility.

Also updates two pre-existing tests (`test_accuracy.py::test_accumulator_device`, `test_running_average.py::test_integration_batchwise`) that assumed `_num_correct` is always a tensor right after `reset()` — it's now a plain `int` until the first `update()`, matching the `int | torch.Tensor` pattern `Precision`/`Recall` already use for `_numerator`/`_denominator`.

Check list:

- [ x ] New tests are added (if a new feature is added)
- [ x ] New doc strings: description and/or example code are in RST format
- [ x ] Documentation is updated (if required)

### DIFF
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -167,11 +167,6 @@ class Accuracy(_BaseClassification):
             :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
-        per_label: if False (default), computes subset accuracy for multilabel data
-            (a sample is correct only if all labels match exactly), returning a single float.
-            If True, computes per-label accuracy for multilabel data, returning a tensor
-            of shape ``(num_labels,)`` where each element is the fraction of samples correctly
-            predicted for that label. Only applicable when ``is_multilabel=True``.
         is_multilabel: flag to use in multilabel case. By default, False.
         device: specifies which device updates are accumulated on. Setting the metric's
             device to be the same as your ``update`` arguments ensures the ``update`` method is non-blocking. By
@@ -179,6 +174,11 @@ class Accuracy(_BaseClassification):
         skip_unrolling: specifies whether output should be unrolled before being fed to update method. Should be
             true for multi-output model, for example, if ``y_pred`` contains multi-output as ``(y_pred_a, y_pred_b)``
             Alternatively, ``output_transform`` can be used to handle this.
+        per_label: if False (default), computes subset accuracy for multilabel data
+            (a sample is correct only if all labels match exactly), returning a single float.
+            If True, computes per-label accuracy for multilabel data, returning a tensor
+            of shape ``(num_labels,)`` where each element is the fraction of samples correctly
+            predicted for that label. Only applicable when ``is_multilabel=True``.
 
     Examples:
 
@@ -311,13 +311,11 @@ class Accuracy(_BaseClassification):
     def __init__(
         self,
         output_transform: Callable = lambda x: x,
-        per_label: bool = False,
         is_multilabel: bool = False,
         device: str | torch.device = torch.device("cpu"),
         skip_unrolling: bool = False,
+        per_label: bool = False,
     ):
-        if type(per_label) is not bool:
-            raise ValueError("Argument per_label should be boolean.")
         if per_label and not is_multilabel:
             raise ValueError("Argument per_label=True is only applicable with is_multilabel=True.")
         self._per_label = per_label
@@ -337,6 +335,7 @@ class Accuracy(_BaseClassification):
         self._check_type(output)
         y_pred, y = output[0].detach(), output[1].detach()
 
+        sum_dim: int | None = None
         if self._type == "binary":
             correct = torch.eq(y_pred.view(-1).to(y), y.view(-1))
         elif self._type == "multiclass":
@@ -349,22 +348,23 @@ class Accuracy(_BaseClassification):
             y_pred = torch.transpose(y_pred, 1, last_dim - 1).reshape(-1, num_classes)
             y = torch.transpose(y, 1, last_dim - 1).reshape(-1, num_classes)
             if self._per_label:
+                # Keep the label dimension so each column gets its own accuracy
                 correct = (y == y_pred.type_as(y)).float()
-                self._num_correct = self._num_correct + correct.sum(dim=0).to(self._device)
-                self._num_examples += correct.shape[0]
-                return
-            # Default: subset accuracy — entire label vector must match exactly
-            correct = torch.all(y == y_pred.type_as(y), dim=-1)
+                sum_dim = 0
+            else:
+                # Subset accuracy — entire label vector must match exactly
+                correct = torch.all(y == y_pred.type_as(y), dim=-1)
         else:
             raise ValueError(f"Unexpected type: {self._type}")
 
-        self._num_correct += torch.sum(correct).to(self._device)
+        # Non-in-place: on the first per-label update, this grows `_num_correct` from a
+        # 0-dim tensor to shape (num_labels,), which in-place `+=` cannot broadcast into.
+        self._num_correct = self._num_correct + torch.sum(correct, dim=sum_dim).to(self._device)
         self._num_examples += correct.shape[0]
 
     @sync_all_reduce("_num_examples", "_num_correct")
     def compute(self) -> float | torch.Tensor:
         if self._num_examples == 0:
             raise NotComputableError("Accuracy must have at least one example before it can be computed.")
-        if self._per_label:
-            return self._num_correct / self._num_examples
-        return self._num_correct.item() / self._num_examples
+        num_correct = self._num_correct if self._per_label else self._num_correct.item()
+        return num_correct / self._num_examples
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -1,4 +1,5 @@
 from collections.abc import Callable, Sequence, Iterable
+from typing import cast
 
 import torch
 
@@ -349,7 +350,6 @@ class Accuracy(_BaseClassification):
             y_pred = torch.transpose(y_pred, 1, last_dim - 1).reshape(-1, num_classes)
             y = torch.transpose(y, 1, last_dim - 1).reshape(-1, num_classes)
             if self._average is False:
-                # Per-label elementwise accuracy: count correct predictions per label separately
                 correct = (y == y_pred.type_as(y)).float()
                 self._num_correct += correct.sum(dim=0).to(self._device)
                 self._num_examples += correct.shape[0]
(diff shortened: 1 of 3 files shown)


ignite/metrics/accuracy.py
```diff
@@ -269,24 +302,33 @@ def thresholded_output_transform(output):
...
+    .. versionchanged:: 0.6.0
+        ``average`` argument is added.
     """
 
     _state_dict_all_req_keys = ("_num_correct", "_num_examples")
 
     def __init__(
         self,
         output_transform: Callable = lambda x: x,
+        average: bool | None = None,
```
aaishwarymishra: we can have it true by default instead of the `None`

ignite/metrics/accuracy.py
```diff
@@ -269,24 +302,33 @@ def thresholded_output_transform(output):
...
         is_multilabel: bool = False,
         device: str | torch.device = torch.device("cpu"),
         skip_unrolling: bool = False,
     ):
+        if average is not None and average is not False:
+            raise ValueError("Argument average should be None or False.")
```
aaishwarymishra:
```suggestion
        if type(average) is not bool:
            raise ValueError("Argument average should be boolean")
```

ignite/metrics/accuracy.py
```diff
@@ -269,24 +302,33 @@ def thresholded_output_transform(output):
...
+            raise ValueError("Argument average=False is only applicable with is_multilabel=True.")
+        self._average = average
         super().__init__(
             output_transform=output_transform, is_multilabel=is_multilabel, device=device, skip_unrolling=skip_unrolling
         )
 
     @reinit__is_reduced
     def reset(self) -> None:
-        self._num_correct = torch.tensor(0, device=self._device)
+        self._num_correct: int | torch.Tensor = 0
```
aaishwarymishra: why its union int tensor?

ignite/metrics/accuracy.py
```diff
@@ -307,6 +349,12 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
             last_dim = y_pred.ndimension()
             y_pred = torch.transpose(y_pred, 1, last_dim - 1).reshape(-1, num_classes)
             y = torch.transpose(y, 1, last_dim - 1).reshape(-1, num_classes)
+            if self._average is False:
+                correct = (y == y_pred.type_as(y)).float()
+                self._num_correct += correct.sum(dim=0).to(self._device)
```
aaishwarymishra: I dont think we need to move it into a new device

ignite/metrics/accuracy.py
zongyang078: Makes sense. Switching to `average: bool = True` with True as the default keeps it simple. But I want to flag a naming conflict before changing the default. In `Precision`/`Recall`, `average=True` means "macro average," not "current default behavior." Using True to mean "subset accuracy" here would be semantically different from the sibling classes. Happy to go either way — do you think the simplicity of a plain bool outweighs the consistency concern, or should we keep None to leave room for future `'micro'`/`'macro'` modes?

ignite/metrics/accuracy.py
zongyang078: Will update if we go with the bool-only approach from the comment above.

ignite/metrics/accuracy.py
zongyang078: Because `reset()` runs before the first `update()`, and at that point we don't know `num_labels` yet. If we initialize as `torch.tensor(0)` (a real 0-dim tensor), the first per-label update does `self._num_correct += correct.sum(dim=0)` — that's an in-place `+=` trying to broadcast shape `()` to `(C,)`, which PyTorch rejects (`RuntimeError: output with shape [] doesn't match the broadcast shape`). Starting from plain `int 0` avoids this because Python's `int.__iadd__` falls back to creating a new tensor via `__radd__`. This is the same pattern `Precision`/`Recall` use for `_numerator`/`_denominator`.

ignite/metrics/accuracy.py
zongyang078: I kept it for consistency with the existing default path a few lines below (`self._num_correct += torch.sum(correct).to(self._device)`), which does the same thing. Happy to remove both if you'd prefer, or keep both — just want them consistent.

ignite/metrics/accuracy.py
aaishwarymishra: Making it None doesn't make is consistent either if consistency is the problem rename it possibly

ignite/metrics/accuracy.py
aaishwarymishra: Its redundant in my opinion

ignite/metrics/accuracy.py
zongyang078: Went with option to rename the parameter instead of just changing the default — average is now subset_accuracy: bool = True. This sidesteps the semantic clash with Precision/Recall's average=True (mapped to "macro") entirely, since the name no longer overlaps with that vocabulary: subset_accuracy=True (default) keeps the current exact-match behavior, subset_accuracy=False returns the per-label tensor. Validation is now the simple boolean check you suggested (type(subset_accuracy) is not bool), since there's no more None sentinel to special-case. Updated docstring, tests, and the example accordingly. Let me know if subset_accuracy reads oddly to you — happy to consider alternatives if so.

ignite/metrics/accuracy.py
zongyang078: Tested this after removing it— swapped in mps locally as a GPU stand-in since I don't have CUDA here. With device="cpu" but input tensors on mps, removing .to(self._device) makes _num_correct end up on mps instead of cpu. That's exactly what test_integration_multilabel_subset_accuracy_false checks (_num_correct.device == metric_device) — passes locally by coincidence since there's no real device mismatch here, but would break on a GPU CI runner. So I kept it as-is for now, happy to discuss more if you still think it should go.

aaishwarymishra:
My bad I missed the updates

ignite/metrics/accuracy.py
```diff
@@ -167,6 +168,11 @@ class Accuracy(_BaseClassification):
             :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
+        subset_accuracy: if True (default), computes subset accuracy for multilabel data
```
aaishwarymishra: I dont actually like this name its confusing it does not tell what it does by just its name, maybe we can try label_average or something similar I am not sure.

aaishwarymishra:
Also I feel many of the tests are redundant and can be removed or be paramatized. Sorry for late response, i missed the updates on the pr

ignite/metrics/accuracy.py
zongyang078: Renamed to per_label: bool = False — polarity now matches what it returns (per_label=True → per-label tensor).

zongyang078:
> Also I feel many of the tests are redundant and can be removed or be paramatized. Sorry for late response, i missed the updates on the pr

Merged the all_correct/all_wrong/no_positives tests into one parametrized test, and dropped the duplicate module-level engine test (verified TestDistributed's gloo_cpu case already covers the same reset-per-epoch scenario by default).

aaishwarymishra (approved):

aaishwarymishra:
@vfdev-5 LGTM

ignite/metrics/accuracy.py
```diff
@@ -269,17 +301,26 @@ def thresholded_output_transform(output):
...
+    .. versionchanged:: 0.6.0
+        ``per_label`` argument is added.
     """
 
     _state_dict_all_req_keys = ("_num_correct", "_num_examples")
 
     def __init__(
         self,
         output_transform: Callable = lambda x: x,
+        per_label: bool = False,
```
vfdev-5: This arg should be the last in the function signature in order to keep backward compatibility:
```python
        device: str | torch.device = torch.device("cpu"),
        skip_unrolling: bool = False,
        per_label: bool = False,
```

ignite/metrics/accuracy.py
```diff
@@ -269,17 +301,26 @@ def thresholded_output_transform(output):
...
         is_multilabel: bool = False,
         device: str | torch.device = torch.device("cpu"),
         skip_unrolling: bool = False,
     ):
+        if type(per_label) is not bool:
+            raise ValueError("Argument per_label should be boolean.")
```
vfdev-5: This check is useless, let's remove it

ignite/metrics/accuracy.py
```diff
@@ -307,6 +348,12 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
             last_dim = y_pred.ndimension()
             y_pred = torch.transpose(y_pred, 1, last_dim - 1).reshape(-1, num_classes)
             y = torch.transpose(y, 1, last_dim - 1).reshape(-1, num_classes)
+            if self._per_label:
```
vfdev-5, line 351: I prefer to rewrite this code without copying the lines like:
```python
self._num_correct += correct.sum(dim=0).to(self._device)
```
and a quickly return.

Let's update the existing code such that we reuse
```python
self._num_correct += torch.sum(correct, **kwargs).to(self._device)
```
and `if self._per_label:` defines some args.

ignite/metrics/accuracy.py
```diff
@@ -315,7 +362,9 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
         self._num_examples += correct.shape[0]
 
     @sync_all_reduce("_num_examples", "_num_correct")
-    def compute(self) -> float:
+    def compute(self) -> float | torch.Tensor:
         if self._num_examples == 0:
             raise NotComputableError("Accuracy must have at least one example before it can be computed.")
+        if self._per_label:
+            return self._num_correct / self._num_examples
         return self._num_correct.item() / self._num_examples
```
```suggestion
         _num_correct = self._num_correct if self._per_label else self._num_correct.item() 
         return _num_correct / self._num_examples
```

ignite/metrics/accuracy.py
zongyang078: Kept this as a non-in-place `self._num_correct = self._num_correct + ...` rather than `+=` — in-place add can't broadcast the 0-dim accumulator up to the `(num_labels,)` shape on the first per-label update (verified locally, raises `RuntimeError: output with shape [] doesn't match the broadcast shape`). Everything else applied as suggested.

vfdev-5: I think if we change `self._num_correct = 0` in `reset()` it would work.

zongyang078: You're right, verified locally.

vfdev-5 (approved):
Thanks @zongyang078 , LGTM!

PR #3811, opened 2026-07-18
feat: optionally preserve sequences in WandB logs

### LINKED ISSUE #3809: Pass lists of scalars to wandb
## 🚀 Feature

When computing per class precision, F1-score, etc... it'd be very useful to have to possibility to put them all inside a list rather than flattened then sent to wandb.

When dealing with 10+ classes, summaries on wandb are extremely hard to read, especially when you keep adding more metrics.

Image

What would be nice is to expose the log call so that it could allow more flexibility in terms of format.

As for now, I'm considering switching temporarily to Lightning.ai until Ignite matures more as a framework.

P.S.: A little off-topic, but despite being able to customize loss functions transforms, I've had a hard time getting automated per-class metrics from `ClassificationReport` because lists dont have an `attach()` method. So instead of doing :
```py3
class_report = ClassificationReport(...)
prec_per_class = [class_report[str(i)]["precision"] for i in range(C)]
```
I had to do something rather tricky :
```py
cm = ConfusionMatrix(...)
prec_per_class = (cm.diag() / (cm.sum(dim=1) + 1e-9)).tolist()
```
But even with `tolist()`, output is flattened before being sent to wandb.

nightcityblade:
Fixes #3809

Description:

Add an opt-in `flatten_sequences=False` setting to the Weights & Biases output handler so lists of scalar metric values can be logged as one sequence. Existing callers retain the current flattened-per-index behavior by default.

Check list:

### DIFF
--- a/ignite/handlers/base_logger.py
+++ b/ignite/handlers/base_logger.py
@@ -110,7 +110,11 @@ class BaseOutputHandler(BaseHandler):
         self.state_attributes = state_attributes
 
     def _setup_output_metrics_state_attrs(
-        self, engine: Engine, log_text: bool | None = False, key_tuple: bool | None = True
+        self,
+        engine: Engine,
+        log_text: bool | None = False,
+        key_tuple: bool | None = True,
+        flatten_sequences: bool = True,
     ) -> dict[Any, Any]:
         """Helper method to setup metrics and state attributes to log"""
         metrics_state_attrs = OrderedDict()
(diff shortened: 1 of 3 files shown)


nightcityblade:
Rebased the branch onto current master and adapted the change to the latest BaseOutputHandler refactor, resolving the merge conflict. Verification: Ruff check and format check pass on all three changed files; the full WandB handler test module passes (13 tests).

PR #3812, opened 2026-07-20
feat: support tuple values in cyclical schedulers

### LINKED ISSUE #438: Allow tuple in param_scheduler
Context:
I'd like to follow Leslie paper and change the momentum in adam at the same time I am changing the lr.
Issues:
The momentum in adam is beta 1 which is the first values of the tuples in the torch initializer:
`Adam(params, lr=0.001, betas=(0.9, 0.999), eps=1e-08, weight_decay=0, amsgrad=False)` (ie. this is 0.9). However, right now there is no way to use  `param_scheduler` with tuples let alone specifying only one tuple.
Proposed solution:
Allow `param_name` to be a bit more complex where you can specify things such as `betas[0]` (or something in the line of what [pydash](https://github.com/dgilland/pydash) is doing `pydash.get(param_gp[0],"betas[0]")` which would require to basically use [base_get](https://github.com/dgilland/pydash/blob/develop/src/pydash/helpers.py#L115)) that will extract the first tuple from the betas. 
I think it might be able to be done in the `__call__` method in `ParamScheduler`

nightcityblade:
Fixes #438

Description:

Allow cyclical schedulers to interpolate tuple-valued optimizer parameters while preserving the tuple type written back to optimizer parameter groups. This enables schedules such as Adam's `betas`, including changing beta1 while keeping beta2 fixed. The regression test exercises five scheduled values and runs an Adam optimizer step after each update.

Tests:

- `python -m pytest -q tests/ignite/handlers/test_param_scheduler.py` (58 passed, 4 xfailed, 2 xpassed)
- `python -m ruff check ignite/handlers/param_scheduler.py tests/ignite/handlers/test_param_scheduler.py`

Check list:

### DIFF
--- a/ignite/handlers/param_scheduler.py
+++ b/ignite/handlers/param_scheduler.py
@@ -277,7 +277,7 @@ class ParamScheduler(BaseParamScheduler):
             values.append([i, scheduler.optimizer_param_groups[0][scheduler.param_name]])
         return values
 
-    def _get_param(self) -> list[float] | float:
+    def _get_param(self) -> list[float] | tuple[float, ...] | float:
         # `ParamScheduler` does nothing special, only returning what child class returns.
         # Intermediate child classes edit this method
         return self.get_param()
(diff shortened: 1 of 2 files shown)


aaishwarymishra (approved):

aaishwarymishra:
@vfdev-5 I think this can be merged.

nightcityblade:
Pushed 7686a45 to fix the pyrefly errors from the tuple scheduler changes. Local validation: the changed module reports zero pyrefly errors, all 58 parameter-scheduler tests pass, and the repository-pinned Ruff and Black checks pass.

PR #3817, opened 2026-07-23
Adding StratifiedBatchSampler for guaranteed subgroup representation

### LINKED ISSUE #3653: Feature Request: StratifiedBatchSampler for Guaranteed Subgroup Representation
## 🚀 Feature

When training on datasets with class imbalance, it is critical to have every subgroup represented in every batch. PyTorch's built-in `WeightedRandomSampler` allows us to assign weights to each subgroup but still relies on random chance, meaning it does not strictly guarantee that rare classes will appear in a specific batch.

For algorithms like Group Distributionally Robust Optimization (DRO) or Contrastive Learning, this is critical since if a group is not represented in a batch, the loss component can potentially become zero and can destabilize training.

If this makes sense, we also have to take care of this question:

- Discarding or Oversampling: `StratifiedBatchSampler` can either oversample the group with the least datapoints or it can create the total number of batches such that it exhausts the subgroup with the least amount of datapoints, which leads to wastage of datapoints of highly-represented classes. We should discuss whether we want to keep both options or not, and which option would act as the default then.

I already have a working prototype of this from a recent project and can modify it for this.

avishkarsonni:
Fixes #3653

Description:
This PR adds `StratifiedBatchSampler` as a guaranteed way to add fixed number of samples of every group in every batch for the methods where missing groups causes in omission of  loss terms (Group DRO and Contrastive learning) which adds another method to existing methods like `WeightedRandomSampler`, `DistributedSampler `, `auto_dataloader`
Check list:

### DIFF
--- a/ignite/distributed/auto.py
+++ b/ignite/distributed/auto.py
@@ -9,7 +9,7 @@ import torch.nn as nn
 from torch.optim.optimizer import Optimizer
 from torch.utils.data import DataLoader, Dataset, IterableDataset
 from torch.utils.data.distributed import DistributedSampler
-from torch.utils.data.sampler import Sampler
+from torch.utils.data.sampler import BatchSampler, Sampler
 
 from ignite.distributed import utils as idist
 from ignite.distributed.comp_models import horovod as idist_hvd
@@ -17,7 +17,7 @@ from ignite.distributed.comp_models import native as idist_native
 from ignite.distributed.comp_models import xla as idist_xla
 from ignite.utils import setup_logger
 
-__all__ = ["auto_dataloader", "auto_model", "auto_optim", "DistributedProxySampler"]
+__all__ = ["auto_dataloader", "auto_model", "auto_optim", "DistributedProxySampler", "StratifiedBatchSampler"]
 
 
 def auto_dataloader(dataset: Dataset, **kwargs: Any) -> DataLoader | _MpDeviceLoader:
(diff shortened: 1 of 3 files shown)


ignite/distributed/auto.py
```diff
@@ -330,6 +330,160 @@ def __iter__(self) -> Iterator:
         return iter(indices)
 
 
+class StratifiedBatchSampler(BatchSampler):
```
TahaZahid05, line 333: @vfdev-5 I am not sure if this is the correct place to put the sampler class. However, we really dont have a `samplers/` folder either in ignite. Would it make sense to have `ignite/samplers`? like we have for metrics, handlers and so on.

avishkarsonni: I think since `DistributedProxySampler` broke this rule years ago and creating new `ignite/samplers` directory will basically break the public path and create all new boilerplate, so I don't think it works well and it is for only 2-3 sampler as of now

ignite/distributed/auto.py
```diff
@@ -330,6 +330,160 @@ def __iter__(self) -> Iterator:
...
+            smallest = min(len(group) for group in self.groups)
+            raise ValueError(
+                f"Smallest group has {smallest} samples, less than num_replicas={self.num_replicas}. "
+                "Every group should have at least one sample per rank"
+            )
+
+        self.samples_per_group = batch_size // num_groups
+        self.leftover_slots = batch_size % num_groups
+        if strategy == "oversample":
+            self.num_batches = max(-(-size // self.samples_per_group) for size in self.shard_sizes)
```
TahaZahid05, line 437: This formula is not correct because it doesn't take into account that `self.leftover_slots` needs to be filled additionally which will cause the groups to deplete quicker than expected for the computed `num_batches`.

ignite/distributed/auto.py
```diff
@@ -330,6 +330,160 @@ def __iter__(self) -> Iterator:
...
+        else:
+            self.num_batches = max(1, min(size // self.samples_per_group for size in self.shard_sizes))
```
TahaZahid05, line 439: same here.

TahaZahid05:
Hi @avishkarsonni! are you still working on this PR? Any issues you are facing, let me know.

avishkarsonni:
Hey, I am working on this. Halted because some medical emergency will jump back once I get back.

TahaZahid05:
@avishkarsonni Hope everything is fine and you are in good health. Do take your full time and rest. We can continue with the PR once you feel good to continue. Thanks for replying!

PR #3820, opened 2026-08-07
Add label-wise accuracy support for multilabel Accuracy metric

### LINKED ISSUE #513: Label-wise metrics (Accuracy etc.) for multi-label problems

danijimmy19:
## Description

Fixes #513.

`Accuracy` currently only supports subset accuracy (exact match across all labels) for multilabel classification — it collapses all `C` labels per sample with `torch.all(..., dim=-1)`. This hides which specific labels a model is getting wrong, unlike `Precision`/`Recall`, which already support per-label output via `average=False`.

This PR adds an `average="label-wise"` option to `Accuracy` for `is_multilabel=True`, returning a `(C,)` tensor of per-label accuracy instead of a single scalar.

## Root cause

`Accuracy`'s `_num_correct` is a scalar with no `average` dispatch, unlike `_BasePrecisionRecall`, which was given tensor accumulators and dispatch when `Precision`/`Recall` were added in the same commit multilabel support landed in `Accuracy` (`1a8ead8b`). `Accuracy` was never brought to parity.

## Change

In `update()`, when `average="label-wise"`:

```python
# Before: subset accuracy — all C labels must match per sample
correct = torch.all(y == y_pred.type_as(y), dim=-1)  # (N,) bool

# After: per-label accuracy — each column independently
correct_per_label = (y == y_pred.type_as(y)).to(dtype=torch.float64)  # (N, C)
self._num_correct = self._num_correct + correct_per_label.sum(dim=0)  # (C,)
```

`compute()` returns a `(C,)` tensor for `average="label-wise"`, and preserves the existing scalar-float behavior otherwise.

## Example

```python
# y_pred / y_true: 5 samples, 3 labels

# Before — Accuracy(is_multilabel=True)
# Result: 0.4  (only 2/5 samples have all 3 labels correct)

# After — Accuracy(is_multilabel=True, average="label-wise")
# Result: [1.0, 0.8, 0.6]  — pinpoints that label 2 is the weakest
```

## Edge cases handled

- `average="label-wise"` with `is_multilabel=False` raises `ValueError` in `__init__`
- Multi-batch updates with different batch sizes accumulate correctly
- Spatial `(N, C, H, W)` multilabel inputs are flattened to `(N*H*W, C)` before the per-label sum
- Distributed training: tensor `_num_correct` of shape `(C,)` is handled correctly by the existing `@sync_all_reduce` decorator
- `compute()` before any `update()` raises `NotComputableError` via the existing guard

## Testing

Added tests to `tests/ignite/metrics/test_accuracy.py` covering:
- output shape `(C,)`, all-correct/all-wrong cases
- multi-batch consistency (matches single-batch result)
- correctness against `sklearn.metrics.accuracy_score` computed per column
- fallback behavior (`average=None`) unchanged
- reset/epoch boundary behavior
- invalid configuration (`average="label-wise"` + `is_multilabel=False`)

## Check list

### DIFF
--- a/ignite/metrics/accuracy.py
+++ b/ignite/metrics/accuracy.py
@@ -167,6 +167,8 @@ class Accuracy(_BaseClassification):
             :class:`~ignite.engine.engine.Engine`'s ``process_function``'s output into the
             form expected by the metric. This can be useful if, for example, you have a multi-output model and
             you want to compute the metric with respect to one of the outputs.
+        average: if ``False`` and ``is_multilabel=True``, returns per-label accuracy as a tensor.
+            By default, ``None`` preserves the existing scalar behavior.
         is_multilabel: flag to use in multilabel case. By default, False.
         device: specifies which device updates are accumulated on. Setting the metric's
             device to be the same as your ``update`` arguments ensures the ``update`` method is non-blocking. By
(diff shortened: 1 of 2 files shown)


danijimmy19:
Hi @vfdev-5, this is my first contribution to this project. I've implemented label-wise accuracy support for issue #513. While working on this, I noticed PR #3810 is also open against the same issue with a similar approach, happy to differentiate my implementation or defer to that PR, whichever you'd prefer. Would appreciate a review when you have time!

aaishwarymishra:
@danijimmy19 Hi, it's always good Idea to ask the existing opened pr if the author is still working on it :) 
As there is an open PR I don't think it will be reviewed yet, but I am keeping it open in case author is not able to work on the PR anymore.

PR #3826, opened 2026-08-18
Add terminate_iteration to skip batch handlers

### LINKED ISSUE #996: Are there any ways to filter out or ignore a batch in engine?

nightcityblade:
Fixes #996

Description:

Add `Engine.terminate_iteration()` so a process function can ignore the current batch without firing `ITERATION_COMPLETED` handlers that consume `state.output`. Apply the signal consistently to both the generator and legacy execution paths, and cover both modes with a regression test.

Tests:

- `python -m pytest tests/ignite/engine/test_engine.py -q` (185 passed, 10 skipped)
- `ruff check ignite/engine/engine.py tests/ignite/engine/test_engine.py`

Check list:

### DIFF
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -1166,10 +1166,9 @@ class Engine(Serializable):
                 yield from self._maybe_terminate_or_interrupt()
 
                 self.state.output = self._process_function(self, self.state.batch)
-                if self.should_terminate_single_iteration:
-                    self.should_terminate_single_iteration = False
-                else:
+                if not self.should_terminate_single_iteration:
                     self._fire_event(Events.ITERATION_COMPLETED)
+                self.should_terminate_single_iteration = False
                 yield from self._maybe_terminate_or_interrupt()
 
                 if self.state.epoch_length is not None and iter_counter == self.state.epoch_length:
@@ -1357,10 +1356,9 @@ class Engine(Serializable):
                 self._maybe_terminate_legacy()
 
                 self.state.output = self._process_function(self, self.state.batch)
-                if self.should_terminate_single_iteration:
-                    self.should_terminate_single_iteration = False
-                else:
+                if not self.should_terminate_single_iteration:
                     self._fire_event(Events.ITERATION_COMPLETED)
+                self.should_terminate_single_iteration = False
                 self._maybe_terminate_legacy()
 
                 if self.state.epoch_length is not None and iter_counter == self.state.epoch_length:
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -673,11 +673,14 @@ class Engine(Serializable):
         self.should_terminate_single_epoch = "skip_epoch_completed" if skip_epoch_completed else True
 
     def terminate_iteration(self) -> None:
-        """Signals that the current iteration should end without firing
+        """Signals that the current iteration should finish without firing
         :attr:`~ignite.engine.events.Events.ITERATION_COMPLETED`.
 
-        This can be used from the process function to ignore a batch without triggering handlers
-        that consume ``state.output``.
+        This can be used to ignore a batch without triggering handlers that consume ``state.output``.
+        The iteration counter is still incremented, and other events such as
+        :attr:`~ignite.engine.events.Events.ITERATION_STARTED` are still fired.
+
+        .. versionadded:: 0.6.0
         """
         self.logger.info("Terminate current iteration is signaled.")
         self.should_terminate_single_iteration = True
--- a/ignite/engine/engine.py
+++ b/ignite/engine/engine.py
@@ -145,6 +145,7 @@ class Engine(Serializable):
         # should_terminate_single_epoch flag: False - don't terminate, True - terminate,
         # "skip_epoch_completed" - terminate and skip the event "EPOCH_COMPLETED"
         self.should_terminate_single_epoch: bool | str = False
+        self.should_terminate_single_iteration: bool = False
         self.should_interrupt = False
         self.state = State()
         self._state_dict_user_keys: list[str] = []
@@ -671,6 +672,16 @@ class Engine(Serializable):
         )
         self.should_terminate_single_epoch = "skip_epoch_completed" if skip_epoch_completed else True
 
+    def terminate_iteration(self) -> None:
+        """Signals that the current iteration should end without firing
+        :attr:`~ignite.engine.events.Events.ITERATION_COMPLETED`.
+
+        This can be used from the process function to ignore a batch without triggering handlers
+        that consume ``state.output``.
+        """
+        self.logger.info("Terminate current iteration is signaled.")
+        self.should_terminate_single_iteration = True
+
     def _handle_exception(self, e: BaseException) -> None:
         if Events.EXCEPTION_RAISED in self._event_handlers:
             self._fire_event(Events.EXCEPTION_RAISED, e)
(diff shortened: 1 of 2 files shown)


tests/ignite/engine/test_engine.py
```diff
@@ -54,6 +52,22 @@ def test_terminate(self, skip_completed):
...
+
+        engine = Engine(process)
+        completed_batches = []
+        engine.add_event_handler(Events.ITERATION_COMPLETED, lambda e: completed_batches.append(e.state.output))
+
+        state = engine.run(range(4))
+
+        assert completed_batches == [1, 3]
+        assert state.iteration == 4
+        assert not engine.should_terminate_single_iteration
```
aaishwarymishra: test should use iteration.started flag instead of completed as skipping should be done on same iteration the flag was set true insteasd of next, its wrong behaviour

ignite/engine/engine.py
```diff
@@ -671,6 +672,16 @@ def terminate_epoch(self, skip_epoch_completed: bool = False) -> None:
         )
         self.should_terminate_single_epoch = "skip_epoch_completed" if skip_epoch_completed else True
 
+    def terminate_iteration(self) -> None:
+        """Signals that the current iteration should end without firing
+        :attr:`~ignite.engine.events.Events.ITERATION_COMPLETED`.
+
+        This can be used from the process function to ignore a batch without triggering handlers
```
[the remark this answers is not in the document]
nightcityblade: Addressed in 122cee3: I expanded the docstring to state that `ITERATION_COMPLETED` is skipped while the iteration counter and other events such as `ITERATION_STARTED` still occur, and added `.. versionadded:: 0.6.0`.

tests/ignite/engine/test_engine.py
nightcityblade: Addressed in 122cee3. The regression test now signals termination from an `ITERATION_STARTED(every=2)` handler and records completed iteration numbers, so it verifies that iterations 2 and 4 skip `ITERATION_COMPLETED` on that same iteration in both execution modes.

nightcityblade:
Follow-up validation for 122cee3:

- `python -m pytest tests/ignite/engine/test_engine.py -q` — 185 passed, 10 skipped
- `ruff check ignite/engine/engine.py tests/ignite/engine/test_engine.py` — passed
- `git diff --check` — passed

aaishwarymishra:
Also terminate_epoch has extra parameter I notiched skip_epoch_completed, I think terminate_iteration should be consistent with that

nightcityblade:
Addressed in 54f5a5e. terminate_iteration now accepts skip_iteration_completed, defaulting to False consistently with terminate_epoch, and both execution paths honor it. The focused parametrized tests pass in generator and legacy modes (4 passed), and Ruff check/format pass.

ignite/engine/engine.py
```diff
@@ -1453,7 +1362,11 @@ def _run_once_on_dataset_legacy(self) -> float:
                 self._maybe_terminate_legacy()
 
                 self.state.output = self._process_function(self, self.state.batch)
-                self._fire_event(Events.ITERATION_COMPLETED)
+                skip_iteration_completed = self.should_terminate_single_iteration == "skip_iteration_completed"
+                self.should_terminate_single_iteration = False
+                if not skip_iteration_completed:
+                    self._fire_event(Events.ITERATION_COMPLETED)
+                self.should_terminate_single_iteration = False
```
aaishwarymishra, line 1361:
```suggestion
```

ignite/engine/engine.py
```diff
@@ -1267,7 +1171,11 @@ def _run_once_on_dataset_as_gen(self) -> Generator[State, None, float]:
                 yield from self._maybe_terminate_or_interrupt()
 
                 self.state.output = self._process_function(self, self.state.batch)
-                self._fire_event(Events.ITERATION_COMPLETED)
+                skip_iteration_completed = self.should_terminate_single_iteration == "skip_iteration_completed"
+                self.should_terminate_single_iteration = False
+                if not skip_iteration_completed:
+                    self._fire_event(Events.ITERATION_COMPLETED)
+                self.should_terminate_single_iteration = False
```
aaishwarymishra, line 1171:
```suggestion
```

aaishwarymishra:
@vfdev-5 LGTM

nightcityblade:
CI note: the only failing check is linkcheck, caused by the pre-existing Polyaxon URL redirect in generated/ignite.handlers.polyaxon_logger.rst (https://polyaxon.com/docs/experimentation/tracking/client/ redirects to /docs/references/python-library/tracking/client/). The PR does not touch that documentation; all code-style, CPU, doctest, MPS, TPU, Horovod, and pyrefly jobs pass.

vfdev-5:
@nightcityblade thanks for your effort working on the issue. I remember it was not a simple issue as we can not simply skip iterations especially with metrics attached to engines. 
Before merging this PR we should check if this work with 1) a trainer engine and a running metric attached and ensure that computed running metric is correct 2) an evaluator engine with metrics we should ensure that computed metrics are correct

nightcityblade:
Addressed in 2466d55. The terminate_iteration regression in tests/ignite/engine/test_engine.py now verifies both requested cases: skipped batches are excluded from a trainer's RunningAverage and from an evaluator's epoch Average, with the expected values asserted in both engine execution modes. Validation: 185 passed, 10 skipped for the full engine test file; Ruff check and format check passed on the changed test.

tests/ignite/engine/test_engine.py
```diff
@@ -54,6 +52,28 @@ def test_terminate(self, skip_completed):
...
+    def test_terminate_iteration(self):
+        def process(engine, batch):
+            if batch % 2 == 0:
+                engine.terminate_iteration()
+            return batch
+
+        trainer = Engine(process)
+        completed_iterations = []
+        trainer.add_event_handler(Events.ITERATION_COMPLETED, lambda e: completed_iterations.append(e.state.iteration))
+        RunningAverage(output_transform=lambda output: output, alpha=0.5).attach(trainer, "running_average")
```
aaishwarymishra:
```suggestion
        RunningAverage().attach(trainer, "running_average")
```

aaishwarymishra:
sorry, I missed the updates :)

nightcityblade:
Updated the test in bba1810 to use the default RunningAverage decay while retaining the required identity output transform. The expected value is now 1.04 (only completed iterations 1 and 3 contribute with alpha=0.98). Focused result: 2 passed.

tests/ignite/engine/test_engine.py
```diff
@@ -54,6 +52,28 @@ def test_terminate(self, skip_completed):
...
+    def test_terminate_iteration(self):
+        def process(engine, batch):
+            if batch % 2 == 0:
+                engine.terminate_iteration()
+            return batch
+
+        trainer = Engine(process)
+        completed_iterations = []
+        trainer.add_event_handler(Events.ITERATION_COMPLETED, lambda e: completed_iterations.append(e.state.iteration))
+        RunningAverage(output_transform=lambda output: output).attach(trainer, "running_average")
```
aaishwarymishra:
```suggestion
        RunningAverage().attach(trainer, "running_average")
```

aaishwarymishra, line 64: @vfdev-5 dont you think output_transform=lambda output: output should be default here like in `Average` so we dont have to explicitly pass it when declaring running average?

PR #3829, opened 2026-08-23
Fix CharacterErrorRate checkpoint and distributed state

### LINKED ISSUE #3828: CharacterErrorRate drops checkpoint state and fails on empty distributed ranks
### Bug

`CharacterErrorRate` does not fully participate in Ignite's metric state and distributed-reduction protocols:

1. It does not define `_state_dict_all_req_keys`, so `state_dict()` contains none of its accumulators and loading that state loses all progress.
2. `compute()` reduces `_num_errors` and `_num_refs`, but not `_num_examples`. If one distributed rank receives no local batches while another rank receives valid data, the empty rank still sees `_num_examples == 0` after the other accumulators have been reduced and raises `NotComputableError` instead of returning the global CER.

### Reproduction

The state loss is visible without a distributed setup:

```python
from ignite.metrics.nlp import CharacterErrorRate

metric = CharacterErrorRate()
metric.update((["helo"], ["hello"]))

restored = CharacterErrorRate()
restored.load_state_dict(metric.state_dict())
print(restored.compute())
```

On current `master` (`de08e79`), `restored.compute()` raises `NotComputableError`; it should return `0.2`.

The distributed case occurs with uneven input shards: update the metric only on rank 0, then call `compute()` collectively. Other ranks raise because `_num_examples` remains local even though errors and reference characters are globally reduced.

### Proposed fix

Declare all three accumulators in `_state_dict_all_req_keys` and include `_num_examples` in `sync_all_reduce`, matching the bookkeeping used by Ignite's other accumulating metrics. I have focused state round-trip and empty-rank distributed regression tests ready.

aswanth-07:
## What does this PR do?

Fixes #3828.

`CharacterErrorRate` was missing two pieces of standard Ignite metric bookkeeping:

- it declared no `_state_dict_all_req_keys`, so metric checkpoint round-trips discarded all accumulated errors, reference characters, and update state;
- it reduced errors and reference characters across ranks but left `_num_examples` local, causing ranks with no local batches to raise `NotComputableError` even when other ranks had valid input.

This PR declares the three existing accumulators as the metric state and includes `_num_examples` in the existing distributed sum. It does not change CER calculation or the public API.

The regression test covers a state-dict round trip that retains a CER of `0.2`.

### Validation

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 py -3.13 -m pytest -q tests/ignite/metrics/nlp/test_character_error_rate.py` (18 passed)
- `py -3.13 -m ruff check ignite/metrics/nlp/character_error_rate.py tests/ignite/metrics/nlp/test_character_error_rate.py` (passed)
- `py -3.13 -m ruff format --check ignite/metrics/nlp/character_error_rate.py tests/ignite/metrics/nlp/test_character_error_rate.py` (passed)
- `git diff --check` (passed)

### Checklist

- [x] This PR does one thing.
- [x] The bug and proposed fix are documented in the linked issue.
- [x] Necessary regression tests are included.
- [x] Focused tests and code-style checks pass locally.
- [x] No public API or breaking behavior is introduced.

AI assistance disclosure: AI-assisted development tools were used during investigation and implementation. The reported tests were run against the final diff.

### DIFF
--- a/ignite/metrics/nlp/character_error_rate.py
+++ b/ignite/metrics/nlp/character_error_rate.py
@@ -71,7 +71,12 @@ class CharacterErrorRate(Metric):
             0.1379
 
     .. versionadded:: 0.5.2
+
+    .. versionchanged:: 0.6.0
+        Made metric state checkpointable and computation robust to distributed ranks without local examples.
     """
+
+    _state_dict_all_req_keys = ("_num_errors", "_num_refs", "_num_examples")
 
     def __init__(
         self,
@@ -110,7 +115,7 @@ class CharacterErrorRate(Metric):
         self._num_refs += refs
         self._num_examples += 1
 
-    @sync_all_reduce("_num_errors", "_num_refs")
+    @sync_all_reduce("_num_errors", "_num_refs", "_num_examples")
     def compute(self) -> Number:
         if self._num_examples == 0:
             raise NotComputableError("CharacterErrorRate must have at least one example before it can be computed.")
--- a/ignite/metrics/nlp/character_error_rate.py
+++ b/ignite/metrics/nlp/character_error_rate.py
@@ -1,4 +1,3 @@
-from collections.abc import Mapping
 from typing import Callable, Sequence
 
 import torch
@@ -112,13 +111,6 @@ class CharacterErrorRate(Metric):
         self._num_errors += errors
         self._num_refs += refs
         self._num_examples += 1
-
-    def _load_state_dict_per_rank(self, state_dict: Mapping) -> None:
-        # Older CharacterErrorRate checkpoints contain an empty per-rank state
-        # because the metric did not expose any required state attributes.
-        if not state_dict:
-            return
-        super()._load_state_dict_per_rank(state_dict)
 
     @sync_all_reduce("_num_errors", "_num_refs", "_num_examples")
     def compute(self) -> Number:
--- a/ignite/metrics/nlp/character_error_rate.py
+++ b/ignite/metrics/nlp/character_error_rate.py
@@ -1,3 +1,4 @@
+from collections.abc import Mapping
 from typing import Callable, Sequence
 
 import torch
@@ -111,6 +112,13 @@ class CharacterErrorRate(Metric):
         self._num_errors += errors
         self._num_refs += refs
         self._num_examples += 1
+
+    def _load_state_dict_per_rank(self, state_dict: Mapping) -> None:
+        # Older CharacterErrorRate checkpoints contain an empty per-rank state
+        # because the metric did not expose any required state attributes.
+        if not state_dict:
+            return
+        super()._load_state_dict_per_rank(state_dict)
 
     @sync_all_reduce("_num_errors", "_num_refs", "_num_examples")
     def compute(self) -> Number:
--- a/ignite/metrics/nlp/character_error_rate.py
+++ b/ignite/metrics/nlp/character_error_rate.py
@@ -71,9 +71,6 @@ class CharacterErrorRate(Metric):
             0.1379
 
     .. versionadded:: 0.5.2
-
-    .. versionchanged:: 0.6.0
-        Made metric state checkpointable and computation robust to distributed ranks without local examples.
     """
 
     _state_dict_all_req_keys = ("_num_errors", "_num_refs", "_num_examples")
--- a/tests/ignite/metrics/nlp/test_character_error_rate.py
+++ b/tests/ignite/metrics/nlp/test_character_error_rate.py
@@ -1,6 +1,5 @@
 import pytest
 
-import ignite.distributed as idist
 from ignite.exceptions import NotComputableError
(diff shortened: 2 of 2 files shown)


ignite/metrics/nlp/character_error_rate.py
```diff
@@ -71,8 +71,13 @@ class CharacterErrorRate(Metric):
             0.1379
 
     .. versionadded:: 0.5.2
+
+    .. versionchanged:: 0.6.0
+        Made metric state checkpointable and computation robust to distributed ranks without local examples.
```
aaishwarymishra: Not sure we need to add it here , this is mainly used to let user know of behaviour change or the api change

tests/ignite/metrics/nlp/test_character_error_rate.py
```diff
@@ -109,3 +111,24 @@ def test_cer_unicode():
...
+
+@pytest.mark.distributed
+@pytest.mark.skipif(not idist.has_native_dist_support, reason="Skip if no native dist support")
+@pytest.mark.usefixtures("distributed")
+def test_distributed_with_empty_rank():
+    cer = CharacterErrorRate()
+    if idist.get_rank() == 0:
+        cer.update((["bat"], ["cat"]))
+
+    assert cer.compute() == pytest.approx(1 / 3)
```
aaishwarymishra: do we need this ?

ignite/metrics/nlp/character_error_rate.py
aswanth-07: Agreed. I removed the versionchanged entry in 520388a since this only fixes internal state bookkeeping and does not change the public API.

tests/ignite/metrics/nlp/test_character_error_rate.py
aswanth-07: I think we do. The regression is not only that an empty rank should avoid raising. `_num_errors`, `_num_refs`, and `_num_examples` must all reduce to the same global state. Checking `1 / 3` verifies that the numerator and denominator remain correct on every rank, so I kept this assertion.

aaishwarymishra: @vfdev-5 thoughts?

ignite/metrics/nlp/character_error_rate.py
```diff
@@ -110,7 +113,14 @@ def update(self, output: Sequence[str]) -> None:
         self._num_refs += refs
         self._num_examples += 1
 
-    @sync_all_reduce("_num_errors", "_num_refs")
+    def _load_state_dict_per_rank(self, state_dict: Mapping) -> None:
```
vfdev-5: Do we really need this override?
I think we can remove this override...

tests/ignite/metrics/nlp/test_character_error_rate.py
vfdev-5: IMO, the test looks a bit weird. Why one rank would only update the metric and then we want to compute the metric in all ranks?

ignite/metrics/nlp/character_error_rate.py
aswanth-07: Removed the override and its legacy-empty-state test in c038588d. CharacterErrorRate now uses the standard Metric state-loading path.

tests/ignite/metrics/nlp/test_character_error_rate.py
aswanth-07: This models uneven input shards. compute() performs collective reductions, so every rank must enter it even when one rank receives no local batch. Without reducing _num_examples, that rank raises before returning the globally reduced CER. I renamed the test and added a comment to make the scenario explicit in c038588d.

vfdev-5: @aswanth-07 Let's remove this test. We can add it if needed later

aswanth-07: Removed the uneven-shard test and its now-unused distributed import in 4cd467b8. The remaining metric test file passes 18 tests, and Ruff check and format both pass. Thanks.

tests/ignite/metrics/nlp/test_character_error_rate.py
```diff
@@ -1,4 +1,5 @@
 import pytest
+
```
vfdev-5:
```suggestion
```

vfdev-5 (approved):
LGTM, thanks @aswanth-07

aswanth-07:
I synced the branch with current master, which includes #3835 and its fix for the unrelated pyrefly group-typing errors. The focused CharacterErrorRate suite still passes with 18 tests. Ruff check, Ruff format check, and git diff --check also pass. CI should now rerun on the updated head.

aswanth-07:
Both pyrefly jobs are now green. The only current failure is linkcheck against an external ClearML documentation anchor that disappeared. The CharacterErrorRate links pass, and this PR does not touch the ClearML logger documentation, so I have left the source diff unchanged.

PR #3836, opened 2026-09-08
fix(metrics): R2Score SS_tot computed in float64 to avoid catastrophic cancellation

VirajMishra1:
## Problem

`R2Score.compute()` evaluates the total sum of squares as:

```python
self._y_sq_sum.item() - (self._y_sum.item() ** 2) / self._num_examples
```

This is the naive `Σy² - (Σy)²/n` formula. When `|mean(y)|` is large relative to `std(y)`, both terms are nearly equal large numbers and their difference loses all significant digits in float32. Concrete failure (reported in #3748):

```python
mean, std = 1e6, 1.0
y = torch.full((1000,), mean) + torch.randn(1000) * std
# R2Score returned ~0.89 instead of the correct ~0.99
```

## Fix

Cast accumulators to float64 for the final subtraction. float64 provides ~16 significant digits versus float32's ~7, recovering the necessary precision without changing the accumulation dtype or the distributed sync protocol.

Also:
- Clamp `ss_tot` to zero to guard against rounding producing a tiny negative (which would silently flip the sign of R²).
- Raise `NotComputableError` when all targets are identical (`ss_tot == 0`), where R² is mathematically undefined rather than returning ±∞.

## Tests

- `test_r2_score_large_mean_numerical_stability`: verifies the metric matches sklearn (float64) within 0.01 when `mean=1e6, std=1`.
- `test_r2_score_constant_target_raises`: verifies `NotComputableError` is raised for a constant target.

Relates to #3748

### DIFF
--- a/ignite/metrics/regression/r2_score.py
+++ b/ignite/metrics/regression/r2_score.py
@@ -77,4 +77,14 @@ class R2Score(_BaseRegression):
     def compute(self) -> float:
         if self._num_examples == 0:
             raise NotComputableError("R2Score must have at least one example before it can be computed.")
-        return 1 - self._sum_of_errors.item() / (self._y_sq_sum.item() - (self._y_sum.item() ** 2) / self._num_examples)
+        n = self._num_examples
+        # Compute SS_tot in float64 to reduce catastrophic cancellation.
+        # The naive formula Σy² - (Σy)²/n subtracts two near-equal large
+        # numbers, which loses precision in float32 when |mean(y)| >> std(y).
+        # Casting to float64 for the final subtraction recovers most digits.
+        ss_tot = self._y_sq_sum.double() - self._y_sum.double().pow(2) / n
+        # Guard against tiny negative values from floating-point rounding.
+        ss_tot = ss_tot.clamp(min=0.0)
+        if ss_tot == 0:
+            raise NotComputableError("R2Score is undefined when all target values are identical.")
+        return 1 - self._sum_of_errors.item() / ss_tot.item()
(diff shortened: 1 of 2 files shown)


siddhant-shahhh (commented):
I don't think the cast does what this is intended to do.

reset() creates torch.tensor(0.0), so the accumulators are float32, and _update does
self._y_sq_sum += torch.sum(torch.pow(y, 2)), also float32. By the time compute() runs both operands have already been rounded to float32, so .double() casts the rounding error rather than undoing it.

Running your test case's arithmetic directly (torch.manual_seed(0), mean=1e6, std=1, n=1000):

    y_sq_sum (float32)      = 1000000054099968.0
    ulp at that magnitude   = 1.19e+08
    true SS_tot (float64)   = 1057.91

    SS_tot as computed here = -73900036.125   ->  clamp(min=0) -> 0.0

So SS_tot is the difference of two ~1e15 float32 values: its magnitude is set by the rounding
error and its sign by the summation order, and the true value sits five orders of magnitude
below the ulp. There is nothing left to recover at the subtraction.

On this machine that means compute() doesn't return a wrong number, it raises:

    NotComputableError: R2Score is undefined when all target values are identical.

on data with std=1. The new test never reaches its assertion. Where the residual happens to
land positive it would return a meaningless value instead, so the outcome is platform
dependent either way.

For reference, master on the same input gives 1354.179062035269, and sklearn (float64) gives
-94525532.93. The underlying bug is definitely real, so this is worth fixing.

What I measured about the fix itself:

    float64 accumulators only (reset() change)   1354.179062035269   identical to master
    float64 through the reduction                -94540276.71        0.016% off the reference

So the accumulator dtype isn't the lever. torch.sum(torch.pow(y, 2)) has to be reduced in float64, and even then the one-pass sum(y^2) - sum(y)^2/n form still cancels. A centered or Welford update would be the actual fix.

The constant-target guard is a good idea and worth keeping in some form: master raises
ZeroDivisionError: float division by zero there, which tells you nothing. But as written the
message is a misdiagnosis it fires above on targets that aren't remotely constant. Once
SS_tot is computed stably, that check becomes meaningful.

Last thing: ruff format --diff wants to reformat the long assert in the test, and
.pre-commit-config.yaml pins ruff 0.11.13, so code-style-checks will go red as-is.

PR #3838, opened 2026-09-12
Keep finite Python numbers from terminating training

betacatsling:
Description:

Calling TerminateOnNan with a finite Python value such as `1e100` terminates the engine. The implicit float32 tensor conversion overflows before the finiteness check. Nested outputs and large finite complex values are affected as well.

Convert real Python numbers to float64 and complex numbers to complex128 before checking them. Tensor inputs retain their existing dtype.

Validation: 4 regression failures before the fix; all 18 tests in `tests/ignite/handlers/test_terminate_on_nan.py` pass on CPU afterward. Ruff lint and format checks pass for changed files.

Check list:

- [x] Regression tests added.
- [x] No public API or documentation changes required.

Implementation and validation were performed with AI assistance.

Fixes #3837.

PR #3840, opened 2026-09-12
Preserve scalar precision in GeometricAverage

### LINKED ISSUE #3839: Preserve scalar precision in GeometricAverage
`GeometricAverage.update(1e-100)` produces zero and `update(1e100)` produces infinity, even though its float64 accumulator can represent both. A scalar and an equivalent float64 tensor also disagree at ordinary magnitudes.

Reproduced against master `851c994eb741466c5b3e673b612657a28135f65e` with Python 3.12 and CPU PyTorch 2.6.0. A regression test and focused fix are prepared.

This report was prepared with AI assistance.

betacatsling:
Description:

`GeometricAverage.update(1e-100)` produces zero and `update(1e100)` produces infinity, even though its float64 accumulator can represent both. A scalar and an equivalent float64 tensor also disagree at ordinary magnitudes.

Construct scalar tensors on the accumulator device with its dtype before taking the logarithm.

Validation: 3 regression failures before the fix; `pytest tests/ignite/metrics/test_accumulation.py -m "not distributed" -q`: 12 passed, 5 skipped, 3 deselected on CPU. Distributed and GPU tests were not run. Ruff lint and format checks pass for changed files.

Check list:

- [x] Regression tests added.
- [x] No public API or documentation changes required.

Implementation and validation were performed with AI assistance.

Fixes #3839.

### DIFF
--- a/ignite/metrics/accumulation.py
+++ b/ignite/metrics/accumulation.py
@@ -298,7 +298,7 @@ class GeometricAverage(VariableAccumulation):
     ):
         def _geom_op(a: torch.Tensor, x: float | torch.Tensor) -> torch.Tensor:
             if not isinstance(x, torch.Tensor):
-                x = torch.tensor(x)
+                x = torch.tensor(x, dtype=a.dtype, device=a.device)
             x = torch.log(x)
             if x.ndim > 1:
                 x = x.sum(dim=0)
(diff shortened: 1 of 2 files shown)


vfdev-5:
Same as https://github.com/pytorch/ignite/pull/3838#issuecomment-5648532910

PR #3843, opened 2026-09-14
Fall back to first visible GPU when local rank exceeds device count

### LINKED ISSUE #3259: `idist.initialize` fails in Slurn when using `--ntasks-per-gpu`
## 🐛 Bug description

When summoning a slurm step with multiple tasks assigning GPUs with the `--ntasks-per-gpu` flag instead of the `--ntasks-per-node` as it seems it was intended, ignite uses the `SLURM_LOCALID` environment as the local rank and use it as the device id to use even though the `--ntasks-per-gpu` already binds the MPI process with a GPU, which cause the call `torch.cuda.set_device(self._local_rank)` to fail.

**To reproduce:**
```sh
srun --ntasks-per-gpu=1 --nodes=2 --gpus-per-node=4 python -e "import ignite.distributed as idist; idist.initialize(backend='nccl')"
```

Which produces the following output:

```
    idist.initialize(backend="nccl")
  File ".../python3.11/site-packages/ignite/distributed/utils.py", line 577, in initialize
    _set_model(comp_model_cls(backend, **kwargs))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python3.11/site-packages/ignite/distributed/comp_models/native.py", line 92, in __init__
    self._create_from_backend(
  File ".../python3.11/site-packages/ignite/distributed/comp_models/native.py", line 127, in _create_from_backend
    torch.cuda.set_device(self._local_rank)
  File ".../python3.11/site-packages/torch/cuda/__init__.py", line 408, in set_device
    torch._C._cuda_setDevice(device)
```

**Intended behaviour:**
Eather
- Detect the present of the `--ntasks-per-gpu` flag, which does not seem to be possible
- Detect that only one GPU is available and use it
- Allow to explicitly set the local id or the device to use even though idist is initialized in a slurm environment

## Environment

- PyTorch Version (e.g., 1.4): 2.2
- Ignite Version (e.g., 0.3.0): 0.5.0.post2
- OS (e.g., Linux): Linux
- Python version: 3.11
- Any other relevant information: slurm 23.02.7

rishabhsai:
Fixes #3259

Description:

Launching with `srun --ntasks-per-gpu=1 --nodes=2 --gpus-per-node=4` gives each task a single visible GPU, but `SLURM_LOCALID` keeps counting tasks on the node, so `idist.initialize("nccl")` reaches `torch.cuda.set_device(3)` in a process that can only see device 0 and every task with a local rank above 0 dies. `ignite/distributed/comp_models/native.py` now uses the first visible device when the local rank is not a valid device index, which is the fallback suggested in this issue, and `device()` stops warning about the mismatch in that case. `auto_model` passes the current CUDA device to `DistributedDataParallel` instead of the local rank, so `device_ids` stays consistent with `idist.device()`, which is the device the model was already moved to. This was checked with a unit test that fakes the visible device count and with the existing distributed test suites on CPU, not on a real Slurm cluster.

Check list:


vfdev-5:
@rishabhsai can you check your PR with on SLURM and first repro the issue and then check the fix?

rishabhsai:
Haven't been able to get on a multi-GPU machine yet, so I haven't reproduced the original failure end to end. The unit test covers the device-index logic (it fakes the visible device count and fails on `main` with `set_device(3)` instead of `set_device(0)`), but that's not the same as seeing it under a real allocation.

I'll follow up when I have access, or happy for you to take it from here if that's faster.

One question in the meantime: I can reproduce the same condition on a two-GPU box by setting the SLURM env vars ignite reads with one GPU visible per process, which hits the same `set_device` path without a scheduler. Would that be useful to you, or do you specifically want it verified under Slurm?

vfdev-5:
The original issue is with slurm, so first we need to understand in very detail the problem with the config described in the issue and reproduce it on slurm. Next, propose a change that wont break current behavior (with and without slurm) and once approved finally implement it.
It is not the point to merge something that does not solve the original issue (even if you try to mock the slurm env vars)

rishabhsai:
that makes sense, will do!

thanks!

rishabhsai:
> The original issue is with slurm, so first we need to understand in very detail the problem with the config described in the issue and reproduce it on slurm. Next, propose a change that wont break current behavior (with and without slurm) and once approved finally implement it.
> 
> It is not the point to merge something that does not solve the original issue (even if you try to mock the slurm env vars)

that makes sense, will do!

thanks!

